PR #58, opened 2014-09-16
Add `\forall` and `\exists`

btwael:
I added `\forall` and `\exists`, so KaTeX is now able to render this:
```tex
\forall x \in X, \quad \exists y \leq \epsilon
```
I don't now exactly the group that I should use for this two arrows, I use "textord" maybe it should be "op".


spicyj:

PR #73, opened 2014-09-17
unicode support (including ∑∏∐ ∫ and √)

kasperpeulen:


gameguy43:
Why were this and #59 closed?

jmeas:
@gameguy43, I messed up my ref in the above comment. See #126

flying-sheep:
#126 doesn’t do what this is about though.

i want `σ ∈ ℝ` to work just as well as `\sigma \in \mathbb R`. this in combination with #132 would fix that, but not #126 (not that #126 wasn’t great work!)

kasperpeulen:
@flying-sheep @gameguy43 

I felt back then that the katex was not interested in this functionality, but I've seen now that it is a project. Are you interested in this? @kevinbarabash

Accept of supporting `σ ∈ ℝ` I think I also made it work to support `∑_n=0^n` render like `\sum_n=0^n`. Note that `∑` is the unicode symbol exactly made for this purpose, and so is not same as the Sigma symbol. 

Combined with something that provides quick unicode typing, see this project of me for example:
http://kasperpeulen.github.io/PressAndHold/Unicode/

This can be a nice way of writing math on the web.

gameguy43:
Thanks for this. FWIW, I think one of our issues with this at the time was the "cent sign" unicode character:
http://www.fileformat.info/info/unicode/char/00A2/index.htm

We probably switched to the html entity or just put it _outside_ the katex blocks. Though to be honest my memory is fuzzy now, this was like 2 years ago!

Curious to know: will this PR allow non-math unicode chars like the "cent sign" as well?

I think that'd be helpful. Sometimes you want to say "5c * 5c = 25c," for example.

Thanks!

kevinbarabash:
@kasperpeulen we're definitely interested in this functionality.  Thanks for re-opening this.  There are some characters that are mapped to multiple commands, see https://github.com/Khan/KaTeX/pull/261#issuecomment-114415474.  Could you leave those out of this pull request?  At some point in the future we'll figure out a good mapping, but for now I'd like to avoid the contentious bits of supporting unicode math symbols.

gagern:
Even without conflicting mappings at the moment, https://github.com/Khan/KaTeX/pull/261#issuecomment-115169955 shows how conflicts in the future might be an issue, so https://github.com/Khan/KaTeX/pull/261#issuecomment-114952458 suggested to make unicode-to-TeX opt-in unless *very* carefully vetted. See also #243 and https://github.com/Khan/KaTeX/labels/Unicode for some more discussion on unicode support in general.

micrub:
+1

edemaine:
@ronkok I believe you have by now added most of this functionality using the new Unicode symbol system, so this old PR can now be closed. I do notice a few extra symbols here, e.g., all the double-struck letters instead of just the typical ones, C/H/N/P/Q/R/Z.  I assume you followed that approach because of what LaTeX supports?  That may be the right approach...

ronkok:
@edemaine I do intend to continue with a few more small Unicode PRs. The choices yet to be made are harder ones and the PRs will come more slowly. I may never get to quite all the characters in this PR #73.

Of course, other are welcome to write their own PR if I have omitted a desired character. I may never get to the one you need.

Regarding double-struck characters, the ones I added are: ℂ ℍ ℕ ℙ ℚ ℝ ℤ, from the BMP. I don't intend to write PRs for any surrogate pairs.

edemaine:
Cool. I think this approach is abandoned, so I'll close the issue, but it will of course remain here for glancing at its list of characters.

PR #75, opened 2014-09-17
Added some favicon links from the KA webpage.

n370:


spicyj:
Thanks for the PR. I don't think we want to do this for now.

PR #80, opened 2014-09-19
Fix alignment of squareroot bar

hollingberry:
Before this change:
screen shot 2014-09-18 at 21 10 26

After this change:
screen shot 2014-09-18 at 21 10 19

Tested in Chrome and Safari

### DIFF
--- a/static/katex.less
+++ b/static/katex.less
@@ -328,7 +328,7 @@
                 border-bottom-width: 0.04em;
                 content: "";
                 display: block;
-                margin-top: -1px;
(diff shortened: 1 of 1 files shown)


PeterBocan:
Did anyone merge this thing since? This thing drives me mad :smile:

kevinb7:
There's other rendering issues with `\sqrt` that we have to address.  See https://github.com/Khan/KaTeX/pull/268#issuecomment-115293652.

PR #81, opened 2014-09-19
Add ominus symbol to symbols.js

kmeurer:
I did not change any functional code whatsoever.  I just added support for the ominus symbol(u2296) to complement the oplus symbol that had already been added.

I also had some trouble finding psi, so I moved it up right next to Psi for maximum legibility.  It's a small change, but I hope it helps!


spicyj:
These should be in master now already; planning to do a release soon.

PR #82, opened 2014-09-19
Added support for \ell and \amalg symbols.

wburgo999:
Added support for two new standard latex symbols.
1. \ell, unicode: 2113
2. \amalg, unicode: 2210


spicyj:

PR #83, opened 2014-09-19
Added ability to render a specific class automatically

matthaywardwebdesign:


spicyj:
Since this can easily be done outside of KaTeX proper and we're trying to keep the API small, I'm going to mark this as won't fix.

PR #57, opened 2014-09-16
Add more arrows such as `\mapsto`, `\leadsto`

simon04:
The following arrows are added:
`\Leftarrow`, `\longleftarrow`, `\leftrightarrow`, `\longleftrightarrow`,
`\Rightarrow`, `\longrightarrow`, `\leadsto`, `\mapsto`, `\longmapsto`

PR #126, opened 2014-09-25
The Big Symbols PR

### LINKED ISSUE #62: Add support for \forall, \exists, \subset, and logical symbols.
Fantastic project. It has been working very well, however it doesn't render quantifiers, logical symbols such as \Leftrightarrow and \wedge, and set notation such as \subseteq, \varnothing, and \cap. These symbols are frequently used so their addition would be greatly appreciated. Thanks.

jmeas:
#### [View the symbols here](http://jmeas.com/katex-symbols/new/src/new.html)

– Adds [lots of new symbols](https://gist.github.com/jmeas/4e2b9df04b2e3a0dfe05)
– Organizes the new symbols
– Only new math symbols have been added
– Some metrics are missing (for now)

###### Future work

Right now it is not easy to add new symbols. I would recommend moving to a system where the symbols list is automatically generated from data lists, which is how I'm doing this PR.

======

`\P` is not be supported. [Ref.](https://github.com/mathjax/MathJax/issues/198#issuecomment-4261384) I tested in a browser and it came out as not-the-right-symbol.

=====

Resolves #62
Supersedes #57, #58, #81, #99, and part of #92

### DIFF
--- a/src/symbols.js
+++ b/src/symbols.js
@@ -17,6 +17,1760 @@
 
 var symbols = {
     "math": {
+        // Non-ASCII
+        "\\aa": {
+            font: "main",
+            group: "textord",
+            replace: "\u00e5"
+        },
+        "\\AA": {
+            font: "main",
+            group: "textord",
+            replace: "\u00c5"
+        },
+        "\\AE": {
+            font: "main",
+            group: "textord",
+            replace: "\u00c6"
+        },
+        "\\ae": {
+            font: "main",
+            group: "textord",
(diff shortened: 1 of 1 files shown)


xymostech:
Hey @jmeas. Sorry for the long absence, I've been working on other projects for a while, and I haven't had time to look into stuff. Thanks for your hard work in putting this together. I'm going to merge this in and fix some straggling problems after the fact.

kevinb7:
:+1:

PR #131, opened 2014-09-27
Add \LaTeX function

drewsberry:
Add function that produces the `\LaTeX` symbol. Only thing different from `\KaTeX` symbol is the `A`.

Values tweaked to most closely represent this LaTeX symbol from Wikimedia: Wikimedia LaTeX symbol

Screenshot:

latex symbol

### DIFF
--- a/src/buildTree.js
+++ b/src/buildTree.js
@@ -672,6 +672,32 @@ var groupTypes = {
             ["katex-logo"], [k, a, t, e, x], options.getColor());
     },
 
+    latex: function(group, options, prev) {
+        // The LaTeX logo. Taken directly from KaTeX symbol, with K
+        // replaced with L.
+        var l = makeSpan(
+            ["l"], [buildCommon.mathrm("L", group.mode)]);
+        var a = makeSpan(
+            ["a"], [buildCommon.mathrm("A", group.mode)]);
+
+        a.height = (a.height + 0.2) * 0.75;
+        a.depth = (a.height - 0.2) * 0.75;
+
+        var t = makeSpan(
+            ["t"], [buildCommon.mathrm("T", group.mode)]);
+        var e = makeSpan(
+            ["e"], [buildCommon.mathrm("E", group.mode)]);
+
+        e.height = (e.height - 0.2155);
+        e.depth = (e.depth + 0.2155);
+
+        var x = makeSpan(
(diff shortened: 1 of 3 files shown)


gagern:
This is the box structure, using [this approach](https://github.com/Khan/KaTeX/wiki/Examining-TeX#boxes):

```
\hbox(6.83331+2.15277)x25.66368
.\OT1/cmr/m/n/10 L
.\kern -3.6
.\vbox(6.83331+0.0)x5.90282, glue set 2.04997fil
..\hbox(4.78334+0.0)x5.90282
...\OT1/cmr/m/n/7 A
..\glue 0.0 plus 1.0fil minus 1.0fil
.\kern -1.49994
.\OT1/cmr/m/n/10 T
.\kern -1.66702
.\hbox(6.83331+0.0)x6.80557, shifted 2.15277
..\OT1/cmr/m/n/10 E
.\kern -1.25
.\OT1/cmr/m/n/10 X
```

And the definition in `latex.ltx` which contains more readable line breaks and indentation:

```tex
\DeclareRobustCommand{\LaTeX}{L\kern-.36em%
        {\sbox\z@ T%
         \vbox to\ht\z@{\hbox{\check@mathfonts
                              \fontsize\sf@size\z@
                              \math@fontsfalse\selectfont
                              A}%
                        \vss}%
        }%
        \kern-.15em%
        \TeX}
```

So we typeset the L and go back 0.36em. Then we save a T in box 0, and construct a vbox with the same height. In that vbox we do some font changing magic, the result of which is a 7pt roman font as can be seen from the boxe structure. The `\vss` makes that letter flush to the top of the vbox. So essentially this is “take a 7pt A but make its top line up with the top of the 10pt letter T”. Then comes the TeX command we already know, with -0.15em space between.

Perhaps the `\TeX` command should be added as well, while we are at it?

*By the way, @xymostech thanks for pointing out [that article](http://tex.stackexchange.com/q/17886/16923) about how to see the definition of a protected command. I incorporated that into [the wiki](https://github.com/Khan/KaTeX/wiki/Examining-TeX#macro-definitions).*

@drewsberry are you interested in adapting your code to this algorithm?

kevinbarabash:
Closing this due to lack of activity.  @drewsberry please re-open if you have time to address the concerns raised above.

edemaine:
Perhaps `\raisebox` would be a good way to go for these macros, once #685 is complete.

Issue #42, opened 2014-09-15
Add support for alternative math fonts


daturkel:
LaTeX has support for quite a few math fonts/alphabets. #19 references the need for \mathrm (roman fonts), but there are quite a few, as we can see [here](http://tex.stackexchange.com/questions/58098/what-are-all-the-font-styles-i-can-use-in-math-mode). The high priority ones, which don't require any nonstandard packages, would probably be:
\mathrm
\mathcal
\mathbb (requires amsfonts/amssymb)
\mathbf

PeterBocan:
+1 
what can I do to implement this feature? (cc @xymostech @spicyj )

raichu:
Please consider adding [esstix](https://www.ctan.org/pkg/esstix) [font](https://en.wikipedia.org/wiki/STIX_Fonts_project#ESSTIX) ([ttf](http://www.w3.org/Amaya/Distribution/esstix.tar.gz)).

spicyj:
Merging into #19.

Issue #19, opened 2014-08-30
Add font styles like \mathrm, \mathcal, etc.


xymostech:

delip:
@xymostech can you look into the PR submitted by @kevinb7? Please let us know if you need help.

kevinb7:
Implemented in https://github.com/Khan/KaTeX/pull/290, https://github.com/Khan/KaTeX/pull/291, and https://github.com/Khan/KaTeX/pull/292.

delip:
Just awesome!

PR #132, opened 2014-09-27
Support math font commands

kevinb7:
This addresses issue https://github.com/Khan/KaTeX/issues/19 and partially addresses issue https://github.com/Khan/KaTeX/issues/42.

TODO:
- [x] ```\mathrm```
- [x] ```\mathit```
- [x] ```\mathbf```
- [x] ```\mathsf```
- [x] ```\mathtt```
- [x] ```\mathbb```
- [x] ```\mathfrak```
- [x] ```\mathcal```
- [x] ```\mathscr```
- [x] test HTML output
- [x] test MathML output
- [x] get rid of fontStack
- [x] fallback to appropriate font, e.g. ```\mathbb{x}``` should look like ```\mathit{x}```
- [ ] ~~font aliases, e.g. ```\Bbb``` => ```\mathbb```~~their behavior is different so I'd like to do this in a separate pull request after the other font commands land.
- [x] \imath \jmath
- [x] more unit tests
- [x] screenshotter tests for everything
- [x] rebase
- [ ] ~~accents use the correct font~~requires exporting shift metrics from mapping.pl and using the combining glyphs for each accent instead of the non-combining glyphs we're using right now.


spelufo:
Very nice. I pulled your `mathrm_mathbb` branch just so I could use this. `\mathbb{R}` for real numbers is essential. Is there a reason why this isn't merged yet? Can I help in any way?

mici-:
Are you considering implementing this? I think mathbb would be a great addition.

underyx:
:1234: :+1:

maxdumas:
I agree this should be merged. This feature is absolutely critical for me to use KaTeX as I work with sets a lot.

akof1314:
I agree this should be merged.

PeterBocan:
@xymostech :+1:

justjanne:
:+1: Only found out KaTeX had no mathbb today, because I assumed it’d just have it – this is a must have!

src/buildMathML.js
```diff
@@ -23,6 +23,16 @@ var makeText = function(text, mode) {
     return new mathMLTree.TextNode(text);
 };
 
+// Keep track of the current font specified by /mathbb, /mathit, or /mathrm
+var fontStack = [];
```
kevinb7: Unfortunately, we can't use the same approach used for color b/c the mathvariant attribute is only valid on <mi>, <mn>, <mo>, <ms>, <mtext> so we have to set it on each of them individually.  This means we have to keep track of the current font.

kevinb7:
I've rebased the code and added support for MathML.  I need to update the screenshots though using the new screenshotter.

kevinb7:
The screenshots have been updated using the new screenshotter.  @xymostech This is ready for review now.

src/buildMathML.js
```diff
@@ -23,6 +23,18 @@ var makeText = function(text, mode) {
...
+
+// A map between tex font commands an MathML mathvariant attribute values
+var fontMap = {
+    "mathbb": "double-struck",
+    "mathit": "italic",
+    "mathrm": "normal",
+    "mathcal": "script",
+    "mathfrak": "fraktur"
+};
```
kevinb7: these values are based on MathML produced by MathJax

gagern:
I should have looked at this before starting my own work on this issue, as I did in #258. Nice work already. Your MathML construction is very elegant with that math-variant thingy. It avoids those exotic unicode symbols which I had been using.

One thing that I think I got right is the list of supported characters for each font. With your code, if someone drops a lower-case letter in any of these functions (except for `k` in `\mathbb`), it will use the KaTeX_Main font and display a warning message about missing metrics. I think it would be better to fall back to normal italic symbols in those cases. That's what MathJax is doing there, and it feels right to me. LaTeX just renders whatever symbols happen to be in the AMS font in these locations, but I'd consider this a bug, not a feature. I haven't yet decided if the MathML should do such a fall-back as well, i.e. whether it should be closer to the input or closer to the output if these don't agree.

The global variable `fontStack` in `buildMathML` is really ugly. In particular, if one rendering run aborts due to an exception, subsequent runs may start with a non-empty stack. Perhaps we should add an `options` object, as it's used in `buildHTML`. Or perhaps we should make all of these functions methods of some object which can be used to hold such state. Or we should at least clean out the stack at the top level invocation for `buildMathML`.

Do you want to add `\mathscr` as well? Can this be distinguished from `\mathcal` in the MathML variant? How about the obsolete but shorter `\Bbb` synonym to `\mathbb`? I believe that there may be similar synonyms for the other fonts as well.

kevinb7:
It should probably fallback to whatever font is appropriate.  We don't have double struck numbers so we should use KaTeX_Main for those and KaTeX_Math for lowercase letters (and Greek) as suggested.  MathJax's MathML output matches its fallbacks.

I will do something about the ```fontStack```.  Thanks for pointing out the exception issue.

I can add ```\mathscr```.  Its mathvariant would also be ```script```.  MathJax adds an additional ```class``` attribute to make the distinction in its MathML rendering.

If ```\Bbb``` is obsolete what's the motivation in supporting it?

gagern:
> If \Bbb is obsolete what's the motivation in supporting it?

Because people are till using it. In my opinion it's the same reason which made you jump through hoops to get `\over` working despite the existence of `\frac`, except that there are no serious hoops involved for `\Bbb`. OK, `\over` is plain TeX, right, which hasn't really been “obsoleted” by LaTeX, but that's just a matter of terminology. Both have more modern replacements, both are being used nonetheless, and both are supported on the (La)TeX side, even if a notice gets displayed.

kevinb7:
Good enough for me.

kevinb7:
I got rid of ```fontStack``` and replaced it with an ```options``` parameter mirroring what was done in buildHTML.js.  I added some tests that verify that the mathvariant attribute is correct in the MathML output.  @gagern are there any other font aliases I should add?

gagern:
> are there any other font aliases I should add?

From my `amstex.sty`:

```tex
\def\frak{\mathfrak}
\def\Bbb{\mathbb}
\def\bold{\mathbf}
```

And from my `article.cls`:

```tex
\DeclareOldFontCommand{\rm}{\normalfont\rmfamily}{\mathrm}
\DeclareOldFontCommand{\sf}{\normalfont\sffamily}{\mathsf}
\DeclareOldFontCommand{\tt}{\normalfont\ttfamily}{\mathtt}
\DeclareOldFontCommand{\bf}{\normalfont\bfseries}{\mathbf}
\DeclareOldFontCommand{\it}{\normalfont\itshape}{\mathit}
\DeclareOldFontCommand{\sl}{\normalfont\slshape}{\@nomath\sl}
\DeclareOldFontCommand{\sc}{\normalfont\scshape}{\@nomath\sc}
\DeclareRobustCommand*\cal{\@fontswitch\relax\mathcal}
\DeclareRobustCommand*\mit{\@fontswitch\relax\mathnormal}
```

Looking at my `amsldoc.pdf`:

> For more comprehensive information on font use in LaTeX, see the LaTeX font guide (`fntguide.tex`) or *The LaTeX Companion* [\[4\]](http://www.amazon.com/Latex-Companion-Addison-Wesley-Techniques-Computer/dp/0201541998). The basic set of math font commands in LaTeX includes `\mathbf`, `\mathrm`, `\mathcal`, `\mathsf`, `\mathtt`, `\mathit`. Additional math alphabet commands such as `\mathbb` for blackboard bold, `\mathfrak` for Fraktur, and `\mathscr` for Euler script are available through the packages `amsfonts` and `euscript` (distributed separately).

So if you are aiming for completeness in terms of supported font commands (I'm not saying that you should be, if that means merging this feature will take considerably longer), then `\mathbf`, `\mathsf` and `\mathtt` should perhaps be added to your TODO list.

Another thing to consider is `\boldsymbol`, declared in `amsmath.sty` and `amstex.sty`. We probably could reproduce `\pmb` as well, but if we do, we should do it in the same “poor” way as `\pmb` does, i.e. by typesetting the same box repeatedly, with slight offsets. Or perhaps using the `text-shadow` CSS property to achieve something similar. I'd say that's a different issue, and not one for which I envision urgent need.

metrics/parse_tfm.py
```diff
@@ -64,11 +64,14 @@ def __init__(self, start_char, end_char, char_info, width_table,
...
-    def get_char_metrics(self, char_num):
+    def get_char_metrics(self, char_num, zero_indexed=False):
         if char_num < self.start_char or char_num > self.end_char:
             raise RuntimeError("Invalid character number")
 
-        info = self.char_info[char_num + self.start_char]
+        if zero_indexed:
+            info = self.char_info[char_num - self.start_char]
+        else:
+            info = self.char_info[char_num + self.start_char]
```
kevinb7: The script font, rsfs, is zero_indexed meaning that although the start_char is 65, the first glyph index is 0.  TODO: update the mapping so that we don't have to do ```- self.start_char``` here.

kevinb7: I tried implementing the TODO, but I could get it to work.  I'm probably misunderstanding something about how the mapping works.  I'm going to leave this as is and add some comments.

kevinb7:
Implemented ```\imath``` and ```\jmath``` as per https://github.com/Khan/KaTeX/issues/191.

src/buildCommon.js
```diff
@@ -308,10 +424,110 @@ var spacingFunctions = {
...
+            // TODO(kevinb) update this when we implement unicode
+            return /[a-zA-Z0-9]/.test(value.charAt(0)) ||
+                utils.contains(greekCapitals, value);
+        }
+    }
+};
+
 module.exports = {
     makeSymbol: makeSymbol,
+    fontMap: fontMap,
```
kevinb7: the fontMap feels common because it's used in both buildHTML.js and builtMathML.js

src/buildCommon.js
```diff
@@ -308,10 +424,110 @@ var spacingFunctions = {
...
     mathit: mathit,
     mathrm: mathrm,
+    mathbf: mathbf,
+    mathbb: mathbb,
+    mathcal: mathcal,
+    mathfrak: mathfrak,
+    mathtt: mathtt,
+    mathscr: mathscr,
+    mathsf: mathsf,
+    mathsym: mathsym,
```
kevinb7: ```mathit```, ```mathrm```, etc. are not common b/c they're only used in buildHTML.js and should be moved into that file.

src/buildCommon.js
```diff
@@ -308,10 +424,110 @@ var spacingFunctions = {
...
+ * - test: function acception mathord/textord value, returns true if the
+ *      variant or non-default font should be used.
+ */
+// A map between tex font commands an MathML mathvariant attribute values
+var fontMap = {
+    // styles
+    "mathbf": {
+        variant: "bold",
+        test: function(value) {
+            return /[a-zA-Z0-9]/.test(value.charAt(0));
```
gagern: Why the `charAt(0)`? Shouldn't this be a single-symbol group in any case?

src/buildCommon.js
gagern: Then `makeVList`, `makeSpan` and so on should be moved as well, I think. Except if you consider these things *common* to different groupTypes, in which case the file name is just a bit confusing. In any case, I'd say that's a separate issue, and I'd leave `mathit` in here for now.

gagern:
Personally, I'd combine all the screenshot tests into a single one, e.g. using the array environment for multi-line support. But that's just my taste, I guess, trying to keep the number of image files in the repository small.

I guess I'd also include `\mathbb k` in the mathbb screenshot. And perhaps the KaTeX symbol in one of them, to verify it's not affected by font changes.

gagern:
The old font commands (those declared using `\DeclareOldFontCommand`) don't affect the single following group, but instead the current group to the end of that group, the way sizing commands work. So they are not exact synonyms. I guess that we generally should have a screenshot to verify the scope of each of these commands: affecting deep levels not just the top level, but ending after the argument (for the modern notation) resp. at the end of the current group (for the old notation).

kevinb7:
> Personally, I'd combine all the screenshot tests into a single one, e.g. using the array environment for multi-line support.

I was thinking about doing this, but I'm worried about having dependencies between tests.

> The old font commands (those declared using \DeclareOldFontCommand) don't affect the single following group, but instead the current group to the end of that group, the way sizing commands work.

Thanks for highlighting the distinction.  I'll save those "alias" commands for another pull request.

> I guess that we generally should have a screenshot to verify the scope of each of these commands

Agreed.

kevinb7:
> And perhaps the KaTeX symbol in one of them, to verify it's not affected by font changes.

Interestingly LaTeX doesn't support ```\LaTeX``` inside math mode.

kevinb7:
The plan is to break this PR into 3 separate PRs, namely:
- [ ] font metrics + CSS changes
- [ ] HTML implementation + screenshot tests
- [ ] MathML implementation + unit tests
Once each of these have been merged, I'll close this PR.

kevinb7:
Depends on:
- https://github.com/Khan/KaTeX/pull/290
- https://github.com/Khan/KaTeX/pull/291
- https://github.com/Khan/KaTeX/pull/292

kevinb7:
Closing because this has been implemented in https://github.com/Khan/KaTeX/pull/290, https://github.com/Khan/KaTeX/pull/291, and https://github.com/Khan/KaTeX/pull/292.

PR #134, opened 2014-09-28
Added a thinspace in the middle of \limsup and \liminf

kevinb7:
This address issue https://github.com/Khan/KaTeX/issues/111.


kevinb7:
We'll do this with macros.

Issue #85, opened 2014-09-19
Add support for \cfrac


murkle:
As a quick fix would be nice to have it as a synonym for \frac

For later here's an example of the difference between \frac and \cfrac
http://malinc.wikidot.com/insert-formulas

prlbr:
It seems that `\cfrac` isn’t in KaTeX v0.7.1. It would be great if it found its way into the next release. Thank you guys for all the wonderful work.

kevinbarabash:
@prlbr see https://github.com/Khan/KaTeX/pull/135.  The code is pretty stale now though.

prlbr:
Unfortunately I have no idea how to help or else I would offer to.
Is there a chance to handle `\cfrac` as alias of `\dfrac` until `\cfrac` is fully supported?

edemaine:
@prlbr You can definitely add a custom macro for yourself that maps `\cfrac` to `\dfrac`. I'll have to look at adding  proper support...

prlbr:
> @prlbr You can definitely add a custom macro for yourself that maps \cfrac to \dfrac.

Thanks for the hint, it works like a charm!

edemaine:
`\cfrac` is defined as follows in [`amsmath.dtx`](http://tug.ctan.org/tex-archive/macros/latex/required/amslatex/math/amsmath.dtx):

```latex
%    Continued fractions. The optional arg l or r controls horizontal
%    placement of the numerators. The |\kern-\nulldelimiterspace|
%    is needed in the definition if we want the right-hand sides of the
%    fraction rules to line up. The \cs{strut} keeps the numerator of
%    a subsidiary cfrac from coming too close to the fraction rule above
%    it.
%    \begin{macrocode}
\newcommand{\cfrac}[3][c]{{\displaystyle\frac{%
  \strut\ifx r#1\hfill\fi#2\ifx l#1\hfill\fi}{#3}}%
  \kern-\nulldelimiterspace}
%    \end{macrocode}
```

LaTeX defines `\nulldelimiterspace=1.2pt`.

Ignoring the optional `[l]`/`[r]` argument, and the `\strut` which is not yet supported, a more accurate macro for `\cfrac` would thus be `"\\cfrac": "\\dfrac{#1}{#2}\kern-1.2pt".  However, I find that this still doesn't right-align the fraction lines, unlike LaTeX.  Any ideas why there's a discrepancy here?

[...]

PR #135, opened 2014-09-29
Added support for \cfrac

kevinbarabash:
This addresses issue https://github.com/Khan/KaTeX/issues/85.

It needs a huxley test.  I've attached some screenshots of `\cfrac` vs `\dfrac`.  I also did some tests with LaTeX to make sure that I was getting the proper spacing.

@xymostech If you happy with the rendering do you want me to try and update the huxley tests?

cfrac:
screen shot 2014-09-28 at 5 59 24 pm

dfrac:
screen shot 2014-09-28 at 5 59 52 pm

cfrac vs dfrac in LaTeX
screen shot 2014-09-28 at 6 05 56 pm

### DIFF
--- a/src/buildCommon.js
+++ b/src/buildCommon.js
@@ -41,6 +41,23 @@ var makeSymbol = function(value, style, mode, color, classes) {
 };
 
 /**
+ * Makes a strut
+ */
+var makeStrut = function() {
+    return new domTree.symbolNode("", 0.85, 0.35);
+};
+
+/**
+ * Inserts kerning where value is specified in ems.
+ * @param value
+ */
+var makeKern = function(value) {
+    var span = new domTree.span(["kern"], [new domTree.symbolNode("\u200b")]);
+    span.style.marginRight = value + "em";
+    return span;
+};
+
+/**
  * Makes a symbol in the italic math font.
  */
 var mathit = function(value, mode, color, classes) {
@@ -267,5 +284,7 @@ module.exports = {
     mathrm: mathrm,
     makeSpan: makeSpan,
     makeFragment: makeFragment,
-    makeVList: makeVList
+    makeVList: makeVList,
+    makeStrut: makeStrut,
+    makeKern: makeKern
 };
--- a/src/buildTree.js
+++ b/src/buildTree.js
@@ -169,6 +169,14 @@ var groupTypes = {
             group.value, group.mode, options.getColor(), ["mord"]);
     },
 
+    strut: function(group, options, prev) {
+        return buildCommon.makeStrut();
+    },
+
+    kern: function(group, options, prev) {
+        return buildCommon.makeStrut(group.value);
+    },
+
     bin: function(group, options, prev) {
         var className = "mbin";
         // Pull out the most recent element. Do some special handling to find
@@ -471,6 +479,11 @@ var groupTypes = {
         // Rule 15e
         var innerChildren = [makeSpan(["mfrac"], [frac])];
 
+        var nullDelimiter = fontMetrics.metrics.nullDelimiter;
+        if (group.value.continued) {
+            innerChildren.push(buildCommon.makeKern(-nullDelimiter));
+        }
+
         var delimSize;
         if (fstyle.size === Style.DISPLAY.size) {
             delimSize = fontMetrics.metrics.delim1;
@@ -484,6 +497,9 @@ var groupTypes = {
                     group.value.leftDelim, delimSize, true,
                     options.withStyle(fstyle), group.mode)
             );
+        } else {
+            // create a null delimiter
+            innerChildren.unshift(buildCommon.makeKern(nullDelimiter));
         }
         if (group.value.rightDelim != null) {
             innerChildren.push(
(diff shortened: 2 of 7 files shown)


src/buildTree.js
```diff
@@ -169,6 +169,14 @@ var groupTypes = {
             group.value, group.mode, options.getColor(), ["mord"]);
     },
 
+    strut: function(group, options, prev) {
+        return buildCommon.makeStrut();
+    },
+
+    kern: function(group, options, prev) {
+        return buildCommon.makeStrut(group.value);
```
gagern: Shouldn't this be `makeKern`?

src/fontMetrics.js
```diff
@@ -106,7 +106,10 @@ var metrics = {
             return sigma21ScriptScript;
         }
         throw new Error("Unexpected style size: " + style.size);
-    }
+    },
+
+    // \nulldelimiterspace=1.2pt, pg. 344 TeXbook
+    nulldelimiterspace: 1.2 / ptPerEm
```
gagern: Does this really belong with the font metrics? I think we're getting the value from `ltplain.dtx`.

Also note that as far as I can tell, it's *always* in `pt`, i.e. always the same size, so you should take the *current* fint size into account when emitting an `em`-based measurement, instead of scaling the space with the font, right?

src/buildTree.js
kevinb7: It definitely should.  Thanks.

src/fontMetrics.js
spicyj: For all pt-based metrics, we assume 10 pt per em. This means we don't need to know the font size in order to render the expressions, like in server mode when we have absolutely no way of knowing. We might want to bump this up a little to 12 or something but there's only a few place where it matters and it makes only a minor difference.

kevinb7: There's lots of other measurements like this that appear in fontMetrics.js, that's why I put it there.  I think you're right about it being always in ```pt```.

src/buildCommon.js
```diff
@@ -41,6 +41,23 @@ var makeSymbol = function(value, style, mode, color, classes) {
...
+
+/**
+ * Inserts kerning where value is specified in ems.
+ * @param value
+ */
+var makeKern = function(value) {
+    var span = new domTree.span(["kern"], [new domTree.symbolNode("\u200b")]);
+    span.style.marginRight = value + "em";
+    return span;
+};
```
kevinb7: makeKern should probably take a "units" parameter.

src/fontMetrics.js
gagern: @spicyj I didn't mean to refer to *absolute* font size, I'm fine with the 10 pt per em approach. What I'm worried about is *relative* size: if you go to scriptscriptstyle, then the em in the inner box will be only half as big as it would have been in the outer box. But if you follow LaTeX, then the spaces should not scale by that factor 0.5 but instead should remain fixed. So I'd say that if you have 1.2/ptPerEm outside, then in scriptscriptstyle you should have twice that. Not that it really matters much in this situation here, it's more a matter of principle. Or is there a deliberate decision to also scale such things down, contrary to what LaTeX does?

spicyj: Oh, I see! Yes, I agree.

src/buildCommon.js
```diff
@@ -41,6 +41,23 @@ var makeSymbol = function(value, style, mode, color, classes) {
 };
 
 /**
+ * Makes a strut
+ */
+var makeStrut = function() {
+    return new domTree.symbolNode("", 0.85, 0.35);
```
gagern: I'm probably being pedantic, but shouldn't this be 0.84 and 0.36? After all, we have a `\baselineskip` of `1.2em` and a strut with height `0.7\baselineskip` and depth `0.3\baselineskip`.

I also wonder: do we need to create a new node every time, or would it be all right to insert the same leaf in multiple places of the internal DOM representation?

kevinb7: It should be 0.84 and 0.36.  I'll change this to be ```0.7 * metrics.baselineSkip``` and ```0.3 * metrics.baselineSkip``` so that people know where those numbers are coming from.  I guess we could re-use the ```domTree.symbolNode``` because it's not a real DOM node.

src/fontMetrics.js
gagern: In #295 I've added a `makeNullDelimiter` function which adds all the classes to change back to textstyle size5 so that the width given relative to that will actually be correct even if the surrounding text is scaled.

edemaine:
@kevinbarabash Here's another blast from the past (though maybe should be re-reviewed after your `\not` PR, #140).  I guess the hardest part will be rebasing to address conflicts.  Also need to implement the few small comments that you already responded to, and the missing tests.  I guess this also predates `texcmp` so we could compare with that to see how the rendering is (but it looks like you followed LaTeX's computations, so it's probably pretty good).

kevinbarabash:
@edemaine since this PR is so old I'd probably cherry-pick or manually port the changes.  I'm going to close this for now and will open a new PR when I have time.

Issue #29, opened 2014-09-07
Add \not


xymostech:

konn:
Why is this yet unimplemented?

edemaine:
@konn It's still being implemented; you can see #140 for the current state of things.

kevinbarabash:
#140 has been merged.  There are some slight difference from LaTeX output in certain situations.  If anyone runs into those situations using LaTeX code you care about, please open an issue for the specific issue.

PR #140, opened 2014-10-01
Added support for \not

kevinbarabash:
Addresses issue https://github.com/Khan/KaTeX/issues/29.  Requires a Huxley test which I will try to add tomorrow.  Here's a screenshot:
screen shot 2014-09-30 at 9 31 54 pm

TODO:
- [x] rebase
- [x] switch the order so the solidus comes first in the span and is shifted over with CSS
- [x] figure out how to make firefox render \not{abc} correctly


kevinb7:
We'll handle this using macros.

gagern:
I don't see how this would be handled with macros, particularly the interaction with the atom type. Can you illustrate that idea? What would the `\not` macro look like?

kevinb7:
I called the symbol the wrong thing, but I added a symbol for the long solidus.  The ```not``` method in buildTree.js places one the long solidus after the ```inner``` symbol.  It seems like a we could have a macro like `\def\not#1{\longSolidus #1}`.  I just made up `\longSolidus` so maybe this isn't possible.

gagern:
`\def\not#1{\longSolidus #1}` could almost be abbreviated to `\let\not=\longSolidus` except for the fact that `\not{ab}` is essentially `\longSolidus{a}b` which shouldn't make a difference. In LaTeX the definition of `\not` is `\not=\mathchar"3236` and it is of type `\mathrel`. Looking at it this way, just your modifications to `symbols.js` should be sufficient to emulate LaTeX. Anything more fancy might be well intended, but actually contrary to how LaTeX does things, leading to incompatibilities.

kevinb7:
In that case I'm going to re-open this.

spicyj:
Maybe-stupid question: why is this its own command type instead of just being one of the supported symbols?

kevinb7:
@spicyj the `\not` glyph gets rendered _after_ the `rel` that follows it in the source code so there needs to be some processing in buildTree.js (now buildHTML.js) to handle this.  The macro I was suggesting in a previous comment should've been `\def\not#1{#1 \longSolidus}`.  Sorry about the confusion.

spicyj:
Oh! I see.

gagern:
But if `\longSolidus` overlaps the preceding character, then it behaves differently from the `\not` in LaTeX. In particular it will be aligned to the right edge of the group it affects, not the left edge. Perhaps we should add kernings so that the glyph is shifted to affect the following character? Or edit the font to achieve that?

kevinb7:
@gagern because the `\longSolidus` occurs after the `inner` in the span it should already be right aligned.  That glyph has funky metrics which shift it to the left by its width.

gagern:
> … it should already be right aligned.

Yes, and that's not consistent with TeX. In LaTeX, `\not{abc}` will essentially be `{\not a}bc`, while your `{abc}\longSolidus` would look like `ab{\not c}`. Sure, with multi-character groups we could split up the group and make this `a\longSolidus bc`, but the same problem holds for symbols of different widths, and for more complicated things in that group following the `\not`.

kevinb7:
@gagern sorry, reading too quickly.  I can flip the order and shift it forward.

src/buildMathML.js
```diff
@@ -454,6 +454,11 @@ groupTypes.phantom = function(group, options, prev) {
     return new mathMLTree.MathNode("mphantom", inner);
 };
 
+groupTypes.not = function(group, options) {
+    // TODO: it's an operator first and find the correct "not" version of that operator
+    return buildGroup(group.value.body, options);
```
kevinbarabash: In the case where there isn't a "not" version of whatever's in the body, I'll add the solidus in manually.  This is what MathJax does:
```
<math xmlns="http://www.w3.org/1998/Math/MathML">
  <mrow class="MJX-TeXAtom-REL">
    <mpadded width="0">
      <mtext>&#x29F8;</mtext>
    </mpadded>
  </mrow>
  <mrow class="MJX-TeXAtom-ORD">
    <mi>a</mi>
    <mi>b</mi>
    <mi>c</mi>
  </mrow>
</math>
```

kevinbarabash:
The order is correct now.  It's too bad firefox doesn't want to render it correctly.

xymostech:
@kevinbarabash do you know why that is? Is it just failing to render the font correctly or something?

kevinbarabash:
@xymostech I'm not sure.  It renders everything correctly except for `\not{abc}`.  For some reason the solidus is missing in this case.

sthoch:
The solidus is there in my Firefox 44.0.2 (OS X 10.9.5).

kevinbarabash:
@sthoch this pull request is to add it as a combining mark.

sthoch:
Sorry, let me clarify: If I take your PR and use it to render `\not{abc}` with my Firefox, the solidus is there and not missing. I thought that was the problem.

kevinbarabash:
@sthoch I didn't realize that you had tried this pull request out.  I'll give it another try.  It might have been a problem with the specific version of Firefox that I was using.  What version are you using?

sthoch:
> Firefox 44.0.2 (OS X 10.9.5).

kevinbarabash:
I can no longer reproduce the issue I was seeing in Firefox.  I've rebased this PR and updated the MathML code to produce something closer to what MathJax produces.

`\not{abc}+1` produces the following MathML:
```
<math>
  <semantics>
    <mrow>
       <mrow>
         <mpadded width="0">&amp;#x29F8;</mpadded>
         <mrow>
            <mi>a</mi>
            <mi>b</mi>
            <mi>c</mi>
          </mrow>
       </mrow>
       <mo>+</mo>
       <mn>1</mn>
     </mrow>
     <annotation encoding="application/x-tex">\not{abc}+1</annotation>
  </semantics>
</math>
```

kevinbarabash:
New Firefox issues.  :(

Running `make screenshots` produces the same output.
screen shot 2017-01-15 at 1 00 09 am
Removing some of the `\not` commands will "fix" others, e.g. removing `\not=` fixes the `\not>` that appears after it.  The whole thing is very strange.

kevinbarabash:
I was able fix the Firefox rendering by using `position: absolute` and a CSS transform.  I'll update the PR in the morning.
screen shot 2017-01-15 at 1 18 02 am

src/buildHTML.js
```diff
@@ -1491,6 +1491,27 @@ groupTypes.mclass = function(group, options) {
...
+    // 0.778 = -1 * ptPerEm * -778
+    // -778 is the horizontal correction that has been added to glyph.
+    // TODO(kevinb) update the font to remove this correction so that we don't
+    // have to reverse its effects here.
+    solidus.style.position = 'absolute';
+    const transform = 'translate(0.778em, 0em)';
+    solidus.style.transform = transform;
+    solidus.style.MozTransform = transform;
+    solidus.style.MsTransform = transform;
+    solidus.style.WebkitTransform = transform;
```
gagern: Why use a transform instead of `solidus.style.left = '0.778em'`? Shouldn't that do the same thing?

src/buildHTML.js
```diff
@@ -1491,6 +1491,27 @@ groupTypes.mclass = function(group, options) {
...
+
+    // The reason for flipping the order is that TeX put the solidus at the
+    // start of a group, not the end.
+    return makeSpan(["mord mrel relpos"], [solidus, ...inner]);
```
gagern: This commit increases the size of the gzipped file by about 4.5k, and I guess it is most likely due to this command here. Not that I'm objecting – switching to ES6 is of little value unless we actually start using some of its features. But we should be aware of this cost.

src/buildMathML.js
```diff
@@ -515,6 +515,14 @@ groupTypes.mclass = function(group, options) {
     return new mathMLTree.MathNode("mstyle", inner);
 };
 
+groupTypes.not = function(group, options) {
+    const solidus = new mathMLTree.TextNode('&#x29F8;');
```
gagern: Why an XML escape sequence instead of a JavaScript one? I disabled the CSS rules which ide our MathML, and sure enough, Firefox renders the escape sequence as is, instead of the symbol you meant.

src/symbols.js
```diff
@@ -492,6 +492,7 @@ defineSymbol(math, main, rel, "\u2190", "\\gets");
 defineSymbol(math, main, rel, ">", "\\gt");
 defineSymbol(math, main, rel, "\u2208", "\\in");
 defineSymbol(math, main, rel, "\u2209", "\\notin");
+defineSymbol(math, main, rel, "\u0338", "\\not");
```
gagern: So `\not` is a symbol *and* a function? Somewhat confusing, I think. Better add a comment there, indicating that this line isn't all there is to that command.

Why do you use the combining version of the glyph, instead of the plain `/`? After all, you're not combining with anything, and I'm not sure whether placing a combining character without a base character is sufficiently well specified to ensure consistent rendering across systems.

static/katex.less
```diff
@@ -553,4 +553,7 @@
         }
     }
 
+    .relpos {
+        position: relative;
+    }
```
gagern: Just a though: Here the name describes *how* the group works, but not its *role*. Perhaps `positioning-root` or something to that effect would better characterize what this is about. On the other hand, it's overly verbose, so perhaps not.

One thing worries me (and has been worrying me for quite some time now): you introduce yet another class with specific meaning, and if the site into which we embed this class has other ideas of what that class name might signify, our element might receive some strange styling. This name in particular isn't too likely to get styled in a strange way, but I'd still be careful with every new class we introduce, as each such change might break some site.

Perhaps we should write a [less plugin](http://www.tutorialspoint.com/less/less_plugins.htm) to add some prefix to all our class names. That way we could use these same prefixes when building our trees, and simply tell page admins not to use any classes starting with that prefix for thir own purposes. `kt-` would be my prefix of choice, I think.

src/buildHTML.js
kevinbarabash: I'll give that a try.

src/buildHTML.js
kevinbarabash: I'll have a look at the output to see what's going on there.

src/buildMathML.js
kevinbarabash: d'oh, thanks :)

static/katex.less
kevinbarabash: Maybe using a negative margin will work better with the non-combining character since it'll have a non-zero width.  Regardless, adding a prefix is probably a good idea.

edemaine:
Just checking in on this.  It looks like most of @gagern's comments were implemented already, but @kevinbarabash probably has a better idea of what remains.  How about we make "add prefix to CSS classes" a separate issue/PR?  I agree that that's important, but shouldn't hold back this PR (IMO).

kevinbarabash:
I started looking into this again and after rebasing there's now a space between `x` and `y` in `\not{x}y` when there shouldn't be so I'll have to figure that out.  Also, Safari handles the zero width solidus incorrectly.  I'm going to see if I can find a solidus with a real width and see that works in Safari.

kevinbarabash:
After doing some more investigation I've come to the realization that the behavior of `\not` is actually quite strange.  Give the following input:
```
ab\not{x}y + \not{x}y + \left(\not{x}y\right) \\

\not=\not>\not\geq\not\in\not<\not\leq\not{abc}
```
LaTeX renders:
screen shot 2017-06-29 at 6 26 14 pm

Notice the spacing around the first `\not{x}y`.  There no space between the `y` and `+` which seems to indicate that both the `x` and `y` should be `mrel`s even though the `y` isn't inside the `\not` group.  :worried:

kevinbarabash:
Also note that the original solution had incorrect spacing between operations.  The spacing is tighter in the LaTeX rendering above.

edemaine:
To my surprise, LaTeX defines `\not` as a symbol, not a function with an argument:

```latex
\DeclareMathSymbol{\not}{\mathrel}{symbols}{"36}
```

So `\not{x}` is enclosing `x` in a group.  `y` isn't doing anything special in `\not{x}y`.  The issue is that `x+\not` renders without a space before the `+` because `\not` is a mathrel.

edemaine:
I tested just having your line

```javascript
defineSymbol(math, main, rel, "\u0338", "\\not");
```

and here's what I got in `make serve`:

not

This actually seems pretty close to what we want -- the only issue is that the `\not` is rendering to the left instead of to the right.  I assume this is a bug in the font, given that it works in LaTeX but not here?

kevinbarabash:
> defineSymbol(math, main, rel, "\u0338", "\\not");

I'll give that a try and play around with the positioning.

src/symbols.js
kevinbarabash: The plain `/` is not as slanted as the combining version of the glyph.  I've updated the code so that `\not` is just a symbol and not a function.   buildHTML.js now does some post-process on `\not`s in `buildExpression`.

src/buildMathML.js
```diff
@@ -639,6 +639,9 @@ const buildExpression = function(expression, options) {
         const group = expression[i];
         groups.push(buildGroup(group, options));
     }
+
+    // TODO(kevinb): combine \\not with mrels and mords
```
kevinbarabash: MathJax combines `\\not` with certain relation operators and outputs the combined HTML entity inside an `<mo>` and for mords, it puts the HTML entity for `not` after the identifier inside an `<mi>`.

kevinbarabash:
I've changed the approach significantly.  Instead having a specific `not` group type and `\\not` function, `buildExpression` processes `\\not` groups and makes necessary changes to the following group.

I left a number of TODOs for subsequent PRs.

kevinbarabash:
not

kevinbarabash:
The handling of `\\not` with ords can definitely be improved, but it seems like a bit of an edge case as `\\cancel` is a better option for ords.

src/buildHTML.js
```diff
@@ -120,6 +120,40 @@ const buildExpression = function(expression, options, isRealGroup) {
         }
     }
 
+    // Process \\not commands within the group.
+    // TODO(kevinb): Handle multiple \\not commands in a row.
+    // TODO(kevinb): Handle \\not{abc} correctly.  The \\not should appear over
+    // the 'a' instead of the 'c'.
```
kevinbarabash: I'll take care of these in a separate PR.

kevinbarabash:
@gagern could you have another look at this?

kevinbarabash:
Looks like I'll have to update the screenshots for this.

src/buildHTML.js
```diff
@@ -117,6 +117,40 @@ const buildExpression = function(expression, options, isRealGroup) {
...
+    for (let i = 0; i < groups.length; i++) {
+        if (groups[i].value === "\u0338" && i + 1 < groups.length) {
+            const children = groups.slice(i, i + 2);
+
+            children[0].classes = ["mainrm"];
+            // \u0338 is a combining glyph so we could reorder the children so
+            // that it comes after the other glyph.  This works correctly on
+            // most browsers except for Safari.  Instead we absolutely position
+            // the glyph and set it's right side to match that of the other
```
edemaine: it's -> its

Can this same approach be used for #802 and #735?

kevinbarabash: It could almost be used for those but those accents are centered horizontally over their contents which complicates matters.

edemaine:
For some reason I'm unable to actually merge this.  Feel free to do so if you can...

kevinbarabash:
> For some reason I'm unable to actually merge this. Feel free to do so if you can...

@edemaine the branch needed to be updated.  Do you see an "Update branch" in your UI?

edemaine:
I did, but it wasn't clickable, which I hadn't seen before.  Presumably a temporary glitch.  Anyway, glad you fixed it.

PR #144, opened 2014-10-01
Added \forall, \exists, \nexists, \implies, \iff, and \span

rguthrie3:
Added these symbols and tested.  I will continue to add symbols as I think of them, but I went ahead and submitted a pull request because these symbols (with the exception of span) are very important and they don't seem to be present.

### DIFF
--- a/src/functions.js
+++ b/src/functions.js
@@ -241,7 +241,7 @@ var duplicatedFunctions = [
             "\\arcsin", "\\arccos", "\\arctan", "\\arg", "\\cos", "\\cosh",
             "\\cot", "\\coth", "\\csc", "\\deg", "\\dim", "\\exp", "\\hom",
             "\\ker", "\\lg", "\\ln", "\\log", "\\sec", "\\sin", "\\sinh",
-            "\\tan","\\tanh"
+            "\\span", "\\tan", "\\tanh"
         ],
         data: {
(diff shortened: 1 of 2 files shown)


spelufo:
I think `\implies` also goes by `\then`, but perhaps there is some subte difference.

spicyj:
Thanks for sending this in. These should all be in master now except \implies and \iff which we'll add soon.

spicyj:
Actually, I'm going to close this out but we have #190 now to track those two.

correnson:
Great job ! Although, it would be nice to update the Wiki accordingly, since this is the primary document found when using GitBook, which could be a _great_ environment for co-authoring math. And having logic connectors is a _strong_ requirement here.

kevinbarabash:
@correnson which part of the wiki needs to be updated?

correnson:
https://github.com/Khan/KaTeX/wiki/Function-Support-in-KaTeX

kevinbarabash:
I see, we're missing \forall, \exists, \nexists, and \span.  Good catch.

kevinbarabash:
After double checking it appears that `\span` is not in master.

kevinbarabash:
@correnson I've updated the wiki.

correnson:
Nice!

PR #149, opened 2014-10-04
Add support for basic comparison operands

SeMeKh:
It is a common practice to use `\lt`, `\gt`, `\eq` instead of `<`, `>`
and `=` in formulas. This also helps in achieving MathJax compatibility.

### DIFF
--- a/src/symbols.js
+++ b/src/symbols.js
@@ -441,6 +441,11 @@ var symbols = {
             group: "rel",
             replace: "\u2245"
         },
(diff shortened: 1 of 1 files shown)


spicyj:
These aren't present in TeX or LaTeX so we're not sure if we want these. This might be a case where we ask you to define your own symbols if you want support for these (though we don't yet let you do that).

kevinb7:
Let's revisit this once we support macros.

PR #151, opened 2014-10-07
Added relation mathematical symbols

peterolson:
Added the following relation, binary operation, and set/logic symbols, and other symbolds listed in the link below that were not already present.

http://en.wikibooks.org/wiki/LaTeX/Mathematics#List_of_Mathematical_Symbols

Added relation symbols:

`\ll\gg\sqsubset\sqsubseteq\sqsupset\sqsupseteq\prec\preceq\succ\succeq\doteq\equiv\simeq\sim\parallel\nparallel\perp\asymp\vdash\smile\frown\bowtie\dashv\ni\mid`

Added binary operation symbols:

`\mp\ominus\uplus\sqcap\ast\sqcup\bigcirc\bullet\ddagger\wr\amalg`

Added set/logic notation symbols:

`\exists\nexists\forall\mapsto\implies\Rightarrow\leftrightarrow\iff\Leftrightarrow`

Added Greek letters:

`\varkappa`

Other symbols:

`\hbar\imath\jmath\ell\Re\Im\wp\nabla`

I tried to add `\eth\aleph\beth\gimel` with the following code, but they did not display how I expected them to, so I'm leaving them out for now.

	"\\eth": {
		font: "main",
		group: "textord",
		replace: "\u00f0"
	},
	"\\aleph": {
		font: "ams",
		group: "textord",
		replace: "\u05d0"
	},
	"\\beth": {
		font: "main",
		group: "textord",
		replace: "\u05d1"
	},
	"\\gimel": {
		font: "main",
		group: "textord",
		replace: "\u05d2"
	},

I don't completely understand when to use the `bin`, `textord` or `mathord` groups, could somebody explain that?

### DIFF
--- a/src/symbols.js
+++ b/src/symbols.js
@@ -366,6 +366,61 @@ var symbols = {
             group: "bin",
             replace: "\u2228"
         },
+        "\\mp": {
+            font: "main",
+            group: "bin",
+            replace: "\u2213"
+        },
+        "\\ominus": {
+            font: "main",
+            group: "textord",
+            replace: "\u2296"
+        },
+        "\\uplus": {
+            font: "main",
+            group: "textord",
+            replace: "\u228e"
+        },
+        "\\sqcap": {
+            font: "main",
+            group: "bin",
+            replace: "\u2293"
+        },
+        "\\ast": {
+            font: "main",
+            group: "bin",
+            replace: "\u2217"
+        },
+        "\\sqcup": {
+            font: "main",
+            group: "bin",
+            replace: "\u2294"
+        },
+        "\\bigcirc": {
+            font: "main",
+            group: "bin",
+            replace: "\u25cb"
+        },
+        "\\bullet": {
+            font: "main",
+            group: "textord",
+            replace: "\u2219"
+        },
+        "\\ddagger": {
+            font: "main",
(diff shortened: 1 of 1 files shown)


walkerrandolphsmith:
I would love to see this feature too.

drdozer:
I really need these symbols. What is blocking this pull request? Is it something I can fix?

spicyj:
I need someone to go through and test that all of the symbols are mapped correctly from the command name to the corresponding unicode character and to make sure that the right character appears. It is also important that the `group` for each symbol is correct, or else the spacing will be wrong.

It's possible that we can do an automated conversion from the MathJax source's symbol tables which uses the same fonts but I haven't looked into what that would involve.

spicyj:
(As my inline comment indicates, at least one of these symbols is miscategorized. I haven't gone through to see how many there are.)

spicyj:
#126 added almost all of these symbols, with the exception of \implies, \iff, \imath, \jmath.

\implies and \iff should be easy to add; \imath and \jmath are harder at the moment due to some font problems.

spicyj:
Actually, I'm going to close this out but we have #190 and #191 now to track those.

PR #156, opened 2014-10-16
Automate screenshotter

kevinb7:
Replacing the original pull request which was against master with this one which is against the branch making it easier to merge.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,3 +2,21 @@ language: node_js
 node_js:
 - "0.11"
 - "0.10"
+addons:
+  firefox: "28.0"
+before_install:
+- wget http://mirrors.kernel.org/ubuntu/pool/main/f/freetype/libfreetype6_2.5.2-1ubuntu2.2_amd64.deb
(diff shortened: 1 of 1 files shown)


gagern:
There is a known problem with the `Lap` example, wich sometimes yields different answers. See https://github.com/Khan/KaTeX/pull/264#issuecomment-114272707. Would it make sense to exclude that example in this test run?

kevinb7:
I'd like to compare both renderings against LaTeX's rendering before taking any action here.

gagern:
I'm making good progress with automatically generating PNGs from LaTeX as well. I guess I'll describe a docker for that, too, to fix the texlive version we're using as well as the ghostscript driver which does the rasterization for imagemagick.

kevinb7:
This is exciting.  I think this will help a lot with adding new features and finding bugs.  
+1 for imagemagick.

kevinb7:
Closing because automating docker on travis-ci is better than having a separate implementation.

PR #161, opened 2014-10-29
Add arrows

pesca:
I added a bunch of arrows that weren't supported.


spicyj:

PR #170, opened 2014-11-27
 Added 14 Symbols

stevenolsen:
Hello Emily and Ben,

Apologies for not emailing you back Emily. After reading the contribution file, we decided to add a few symbols. If the pull request is approved we get a bonus 5 lab marks. It's not worth a lot but anything helps. Hopefully we helped KaTeX a little bit.

The symbols added are: ⊎ ⊓ ⊔ ⊖ ◯ ‡ ⨿ ∃ ∄ ∀ ⇒ ⇔ ∍

### DIFF
--- a/src/symbols.js
+++ b/src/symbols.js
@@ -326,6 +326,11 @@ var symbols = {
             group: "bin",
             replace: "\u00b1"
         },
+	"\\mp": {
+	    font: "main",
+	    group: "bin",
+	    replace: "\u2213"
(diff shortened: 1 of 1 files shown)


spicyj:
Thanks for sending this in. These should all be in master now except \implies and \iff which we'll add soon.

spicyj:
Actually, I'm going to close this out but we have #190 now to track those two.

PR #174, opened 2014-12-18
Added \impliedby and \thinspace

Gupocca:
Adds `\thinspace`, which is simply an alias of `\,`, and the `\impliedby` arrow.


kevinb7:
Could you rebased this an move the definition for ```\impliedby``` so that it comes after ```\implies```?

kevinb7:
We'll tackle this once we have macros.

PR #178, opened 2015-01-07
Add double arrows and preference symbols for economics

cmakler:
Added symbols \LeftArrow and \Rightarrow (double arrows), as well as \prec, \sim, \nsim, and \succ (preference symbols for economics) to symbols.js.

Tested in browser to ensure font compatibility (dropped \precsim and \succsim, which I had originally included, because they raised font errors). Also ran make test. 

Please let me know if you would like me to change something! I anticipate the possibility of needing additional symbols in the future.

Thanks for creating this amazing resource!

Chris

UPDATE: found some equivalent symbols for weak preferences that don't cause font problems; added those in.

### DIFF
--- a/src/symbols.js
+++ b/src/symbols.js
@@ -436,6 +436,11 @@ var symbols = {
             group: "rel",
             replace: "\u2248"
         },
+        "\\sim": {
+            font: "main",
+            group: "rel",
+            replace: "\u223c"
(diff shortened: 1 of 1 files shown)


spicyj:

cmakler:
Awesome, thanks!

Issue #38, opened 2014-09-15
Make KaTeX accessible


dylanb:
Make KaTeX accessible to people with disabilities in particular screen reader users as they are most affected by the current limitations

xymostech:
Hello everyone, thanks for all the feedback here. I've been distracted by other projects, but I'm going to be making a pull request soon about making KaTeX accessible, taking a hybrid MathML + aria-hidden HTML approach. I'd love to get some feedback and a hand in testing it out in different environments if anyone would like to help out.

NSoiffer:
I'm glad to hear you are getting back to this. You probably should take a look at what wikipedia has done to deal with accessible math:  if the user has selected the option  "MathML with SVG or PNG fallback (recommended for modern browsers and accessibility tools)", they render the TeX on their server and then send down MathML and the fallbacks.  To set the options, login, choose appearances, and scroll to the bottom for math options.
For more info, see this [wikimedia page](https://www.mediawiki.org/wiki/Extension:Math). The more sites that do things the same way, the more impetus there is for accessibility tools to handle that.

Hywan:
\o/

fred-wang:
@NSoiffer You can try it on http://katex.xymostech.com/

NSoiffer:
That works with the new MathPlayer+NVDA. I haven't tried anything else, but
that's a positive sign.

brucemiller:
I'm not sure what "works" should mean here: on Firefox it is not using MathML, but span-soup.  Ultimately, you'd want to send MathML to browsers that support it, right?   It looks OK, though, although it's all quite huge, but I'm guessing from the CSS that that is intentional.

NSoiffer:
If you use the develop tools in Firefox to look at the DOM, you'll see
MathML is in there. NVDA finds the MathML and gives it to MathPlayer to
speak.

As for who/what is doing the rendering, that I'm not sure about. For screen
reader accessibility, that doesn't matter. When you need to do sync
highlighting, potentially that might matter. E.g., if some code modifies
the MathML to change the background color on the mfrac, would the display
change correctly?

     Neil

fred-wang:
[...]

PR #189, opened 2015-02-25
Add MathML rendering to improve accessibility

xymostech:
Summary: This adds support for rendering KaTeX to both HTML and MathML
with the intent of improving accessibility. To accomplish this, both
MathML and HTML are rendered, but with the MathML visually hidden and
the HTML spans aria-hidden. Hopefully, this should produce much better
accessibility for KaTeX.

Should fix/improve #38

Test plan:
 - Ensure all the tests, and the new tests, still pass.
 - Ensure that for each of the group types in `buildHTML.js`, there is a
   corresponding one in `buildMathML.js`.
 - Ensure that the huxley screenshots didn't change (except for
   BinomTest, which changed because I fixed a bug in `buildHTML` where
   `genfrac` didn't have a `groupToType` mapping).
 - Run ChromeVox on the test page, render some math. (for example,
   `\sqrt{x^2}`)
   - Ensure that a mathy-sounding expression is read. (I hear "group
     square root of x squared math").
   - Ensure that nothing else is read (like no "x" or "2").
 - Ensure that MathML markup is generated correctly and is interpreted
   by the browser correctly by running
   `document.getElementById("math").innerHTML =
   katex.renderToString("\\sqrt{x^2}");` and seeing that the same speech
   is read.


xymostech:
@spicyj @jeresig Here's the change, would love a review.

I set up a testing site at http://katex.xymostech.com/ If anyone with a screenreader would like to test it out, that would be great!

fred-wang:
> I set up a testing site at http://katex.xymostech.com/ If anyone with a screenreader would like to test it out, that would be great!

The first minus sign is incorrectly converted into an mi element (instead of an mo element).

xymostech:
Hmm, you are correct. That's probably because when we build the HTML tree, we change some of the parse tree in accordance with how TeX works. I'll switch the order of how we build things, but we should probably stop doing that some time in the future...

fred-wang:
I just got a reply from Chris Fleizach (Apple, VoiceOver):
> I did some basic testing and it seems to be working as expected, so it’s looking good!

xymostech:
> note the default value on a ∑ mo is largeop="true"

@fred-wang Interesting. Is there a place where that is documented? [This table](http://www.w3.org/TR/MathML/chapter3.html#presm.mo.dict.attrs) seems to indicate that the default is always `largeop="false"`.

I also found [this stackoverflow question](http://stackoverflow.com/questions/4697100/accessibility-recommended-alt-text-convention-for-svg-and-mathml) about adding alt text for math nodes. I'm a little hesitant about using the `<semantics>` tag because it doesn't look like it's supported anywhere but in Firefox. I guess MathML itself isn't that well supported either in other browsers (:disappointed:), but I'd worry about breaking support on things that work now. What do you think?

> > I did some basic testing and it seems to be working as expected, so it’s looking good!

Sounds great!

fred-wang:
> Interesting. Is there a place where that is documented? This table seems to indicate that the default is always largeop="false".

The table has "set by dictionary (false)" so this means we pick the value from the dictionary or otherwise fallback to "false". The MathML3 dictionary is http://www.w3.org/TR/MathML/appendixc.html and is used in Gecko-based, WebKit-based browsers, MathJax etc

> I'm a little hesitant about using the <semantics> tag because it doesn't look like it's supported anywhere but in Firefox. I guess MathML itself isn't that well supported either in other browsers (:disappointed:), but I'd worry about breaking support on things that work now. What do you think?

I'm not sure what you mean by "not supported". Gecko-based and WebKit-based browsers have MathML support (including for the semantics tag). Even recent Blink-based browsers have some support for the semantics element: the TeX source is shown to workaround the lack of native MathML support. But semantics support should be irrelevant for accessibility? Or maybe Chrome's semantics implementation is breaking support in ChromeVox? (in that case they will probably fix that soon, since Wikipedia uses the semantics element too and I ask them to test it... and it's just a matter of overriding the default CSS stylesheet).

Also, I personally think people should use and encourage standards so "Chrome does not support MathML rendering" is never an excuse ;-)

> I also found this stackoverflow question about adding alt text for math nodes. 

As I recall, altext is more to provide a readable description equivalent to alt (the page you mention indicates "spoken math" not source) while semantics is to provide alternative representation of the formula (TeX source, image etc) that can be accessed by copy and paste etc I know that tools like LaTeXML use the TeX source for alttext and put it in a semantics annotation too... Wikipedia uses the semantics annotation. And I believe MathJax has an option to expose the TeX as a semantics annotation.

xymostech:
Ah, I hadn't seen that table. Very interesting!

These are good points. I'll add in the semantics tag!

PR #214, opened 2015-03-29
Add \# and \& to AMS misc symbols

fkling:
Those symbols are listed in
ftp://ftp.ams.org/pub/tex/doc/amsmath/short-math-guide.pdf, section 3.6

I don't know if I have done this correctly, or if adding support for these symbols conflicts with something else. The tests run fine though. Please let me know!

### DIFF
--- a/src/symbols.js
+++ b/src/symbols.js
@@ -152,6 +152,16 @@ var symbols = {
         },
 
         // Misc Symbols
+        "\\#": {
+          font: "main",
+          group: "textord",
+          replace: "\u0023"
+        },
+        "\\&": {
+          font: "main",
+          group: "textord",
+          replace: "\u0026"
+        },
         "\\aleph": {
             font: "main",
             group: "textord",


src/symbols.js
```diff
@@ -736,6 +736,16 @@ var symbols = {
         },
 
         // AMS Misc
+        "\\#": {
+          font: "ams",
```
spicyj: These should actually be `font: "main"` I believe – you'll notice on the test page (`make serve`) that with that change, these look like they match the other font (instead of falling back to Helvetica) and don't produce warnings in the console:

image

spicyj:
Thanks @fkling! :) If you can fix that up and sign the Khan Academy CLA at https://www.khanacademy.org/r/cla I'll merge this in.

src/symbols.js
fkling: I should have read https://github.com/Khan/KaTeX/blob/master/CONTRIBUTING.md first :-/ I just picked that font because the other symbols used it as well...

fkling:
Fixed and signed! Thanks!

src/symbols.js
```diff
@@ -736,6 +736,16 @@ var symbols = {
         },
 
         // AMS Misc
+        "\\#": {
```
spicyj: Oops, sorry – can you also move these up to the "// Misc symbols" section with the other main-font ones?

fkling: :+1:

fkling:
Moved!

spicyj:
Thanks!

PR #215, opened 2015-03-30
Partly fixes #48: added optional arguments to sqrt

### LINKED ISSUE #48: Cubic (and others) roots are not supported
Example: \sqrt[3]{x} should be rendered as

azotova:
I changed functions.js and buildHTML.js files to support optional arguments to the sqrt function. This partly fixes #48 .

The buildMathML.js file still needs to be changed.

The test fail because I did not remove the following test: "Expected '\sqrt[3]{2}' to fail parsing, but it succeeded".

### DIFF
--- a/src/buildHTML.js
+++ b/src/buildHTML.js
@@ -812,7 +812,26 @@ var groupTypes = {
             ], "firstBaseline", null, options);
         }
 
-        return makeSpan(["sqrt", "mord"], [delim, body]);
+        if(!group.value.optional) {
+            return makeSpan(["sqrt", "mord"], [delim, body]);            
+        } else {
+            var optionInner = buildGroup(group.value.optional,
+                options.withStyle(options.style.sup()));
+            var optionInnerSpan;
+            optionInnerSpan = makeSpan(
+                [options.style.reset(), options.style.sup().cls()], [optionInner]);
+
+            var toShift; //determines vertical shift
+            if (options.style === Style.DISPLAY) {
+                toShift = fontMetrics.metrics.sup1;
+            } else if (options.style.cramped) {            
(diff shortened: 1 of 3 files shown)


xymostech:
Hey there! Thanks for trying to get this working. I'm not sure where you came up with the calculation for how much to shift the exponent up. I looked up the definition of `\root` in the TeX source (which we would like to base the calculations on), and it says this:

```tex
\def\root#1\of{%
  \setbox\rootbox
  \hbox{%
    $\m@th\scriptscriptstyle{#1}$%
  }%
  \mathpalette\r@@t%
}
\def\r@@t#1#2{%
  \setbox\z@\hbox{%
    $\m@th#1\sqrt{#2}$%
  }%
  \dimen@\ht\z@
  \advance\dimen@-\dp\z@
  \mkern5mu%
  \raise.6\dimen@\copy\rootbox %
  \mkern-10mu%
  \box\z@%
}
```

Also, could you change the failing test to make sure that the optional arguments work? Thanks!

azotova:
Hi Emily,

I've changed the failing test.  However, I am afraid that I cannot make the vertical shift fully compliant with  the TeX source. I assumed that the optional argument should be displayed at the same height as a usual superscript (e.g. in 2^3), so I copied the code from 'supsub' group type (rule 18c) to determine the vertical shift (my 'toShift' variable). I am attaching a screenshot (from Chrome) to compare the result with that from jqMath and MathJax.

texcomparison

If it is important to be fully compliant with the TeX source, please close my pull request.  I would be happy if any part of my code could help to fix this issue as I really need it for my personal project.

xymostech:
This was added completely in c48de165e8795dc6fc0c932d83c75b44ff7dc2d9.

PR #218, opened 2015-04-08
Protect elements inside .katex from inheriting text-indent.

### LINKED ISSUE #217: Protect from text-indent
If an element has the `text-indent` property set, then any math that appears within it is spaced out ridiculously. If you go to the KaTeX page, for example, and edit the CSS to add a rule like `body { text-indent: 1em}`, you can see it for yourselves. 

The problem, of course, is that `text-indent` is being inherited and applied to elements displayed as blocks. The fix should be as simple as setting `.katex { text-indent: 0}` in your CSS.

ruricolist:
This fixes #217.

### DIFF
--- a/static/katex.less
+++ b/static/katex.less
@@ -15,6 +15,9 @@
     line-height: 1.2;
     white-space: nowrap;
 
+    // Protect elements inside .katex from inheriting text-indent.
+    text-indent: 0;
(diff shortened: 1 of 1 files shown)


xymostech:
Thanks! Looks great.

PR #225, opened 2015-04-30
buildHTML's buildExpression is destructive, changing the tree object.

jeresig:
This clones the object first before manipulating it, ensuring that the tree object will remain the same (right now if buildHTML was run before buildMathML, buildMathML would have weird failures).

I was able to reproduce this by running `\vec{A}^2` through KaTeX. Without this change `tree` becomes:

```json
[
    {
        "type": "supsub",
        "value": {
            "base": {
                "type": "ordgroup",
                "value": [
                    {
                        "type": "mathord",
                        "value": "A",
                        "mode": "math"
                    }
                ],
                "mode": "math"
            },
            "sup": {
                "type": "textord",
                "value": "2",
                "mode": "math"
            }
        },
        "mode": "math"
    }
]
```

With this change it's left as the correct:

```json
[
    {
        "type": "supsub",
        "value": {
            "base": {
                "type": "accent",
                "value": {
                    "type": "accent",
                    "accent": "\\vec",
                    "base": {
                        "type": "ordgroup",
                        "value": [
                            {
                                "type": "mathord",
                                "value": "A",
                                "mode": "math"
                            }
                        ],
                        "mode": "math"
                    }
                },
                "mode": "math"
            },
            "sup": {
                "type": "textord",
                "value": "2",
                "mode": "math"
            }
        },
        "mode": "math"
    }
]
```

### DIFF
--- a/src/buildHTML.js
+++ b/src/buildHTML.js
@@ -1157,6 +1157,10 @@ var buildGroup = function(group, options, prev) {
  * nodes.
  */
 var buildHTML = function(tree, settings) {
+    // buildExpression is destructive, so we need to make a clone
+    // of the incoming tree so that it isn't accidentally changed
+    tree = JSON.parse(JSON.stringify(tree));
+
     var startStyle = Style.TEXT;
     if (settings.displayMode) {
         startStyle = Style.DISPLAY;


jeresig:
I should note that I couldn't think of a good way to test this, as it would require inspection of the contents at runtime. I'm open to suggestions, though!

jeresig:
cc: @ariabuckles @xymostech

spicyj:
This is known (hence this comment https://github.com/Khan/KaTeX/blob/v0.3.0/src/buildTree.js#L9-L10) and I think our eventual goal was to just make buildHTML not mutate the tree. Curious why this is causing you problems.

jeresig:
@spicyj Ah, I missed that comment! Perhaps this can be a stop-gap, at least, to hopefully provide some consistency to buildHTML. Presumably if the method is ever updated to not be destructive this logic can just be removed.

As to why I encountered this: I'm writing a KaTeX plugin which will support exporting math formulas as english strings (e.g. "x^2" becomes "x to the power of 2"). We're continuing to have problems with MathML in screen readers (sometimes they don't read it, many browsers read it incorrectly, sometimes it's double-read) whereas a pure-English (hopefully i18n, as well!) string seems to work everywhere. This is a technique that is used by other entities, as well.

spicyj:
Is `tree` always valid JSON? I don't recall – maybe @xymostech knows. If it is, I'd rather do `JSON.parse(JSON.stringify(tree))` to do a deep clone here which tends to be fast and has very clear semantics. Generic deep clone functions are generally hard to reason about – what are their semantics when passed a function? an array? a self-referential object? a DOM node? an object with nonenumerable keys? Or maybe the constructor throws when you call it with no arguments (like your code does here). There's really no way to do write a deep clone function that works for all cases since you need to understand the data and object model.

…but if the data is just JSON, it's pretty clear how it should work. :)

jeresig:
@spicyj I was totally going to do the `JSON.parse(JSON.stringify(tree))` technique but then I saw that KaTeX supports IE 8, thus I went with the manual solution :) I don't /think/ it contains non-JSON stuff but yeah, would be good to verify that.

...also @xymostech and I just chatted and we may be able to table this for now as I might be able to pursue my modifications from another direction which won't make this necessary (at least not for now).

xymostech:
@jeresig I'm fairly sure that the tree is pure JSON, so that solution should work. Also, it looks like IE 8 does support both `JSON.parse` and `JSON.stringify` (http://caniuse.com/#feat=json) , so that should be fine?

I think it's pretty reasonable to just merge this in. I've been bitten by this mutating the tree before anyway.

jeresig:
Oh wow, ok! Sigh, I have such outdated JS knowledge in my brain. Ok, I'll update to use the JSON technique.

spicyj:
:+1:

xymostech:
Great! Can you squash the commits?

jeresig:
@xymostech done!

PeterBocan:
I wonder, if this hits performance or not (for large amount of expressions)... 

(Excuse my obsession with performance, but I am "the C guy". :+1: )

jeresig:
It's absolutely inefficient. The buildHTML code should be changed to not be
destructive -- at which point this can just be removed. Unfortunately I
don't know the buildHTML sub-systems as well but if someone is willing to
tackle this then that'd be fantastic!

spicyj:
@PeterBocan We care about perf too. JSON.parse and JSON.stringify are actually very fast in most browsers and this is unlikely to take a significant cost, especially in comparison to the parsing itself and DOM tree-building. Still would be nice to get rid of it though.

PR #227, opened 2015-05-04
Expose a new .__parse() method for generating a parse tree from a math expression

jeresig:
Explicitly marked with tons of warnings, both in name and in comments. To be used for some internal projects, for the time being. Added a simple test, just as a sanity check.

### DIFF
--- a/test/katex-spec.js
+++ b/test/katex-spec.js
@@ -1116,6 +1116,31 @@ describe("A markup generator", function() {
     });
 });
 
+describe("A parse tree generator", function() {
+    it("generates a tree", function() {
+        var tree = katex.__parse("\\sigma^2");
+        expect(JSON.stringify(tree)).toEqual(JSON.stringify([
+            {
+                "type": "supsub",
+                "value": {
+                    "base": {
+                        "type": "mathord",
+                        "value": "\\sigma",
+                        "mode": "math"
+                    },
+                    "sup": {
+                        "type": "textord",
+                        "value": "2",
+                        "mode": "math"
+                    },
+                    "sub": undefined
+                },
+                "mode": "math"
+            }
+        ]));
+    });
+});
+
 describe("An accent parser", function() {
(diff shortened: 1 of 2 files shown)


test/katex-spec.js
```diff
@@ -1116,6 +1116,31 @@ describe("A markup generator", function() {
     });
 });
 
+describe("A parse tree generator", function() {
+    it("generates a tree", function() {
+        var tree = katex.__parse("\\sigma^2");
+        expect(JSON.stringify(tree)).toEqual(JSON.stringify([
```
jeresig: Surprisingly just doing `expect(tree).toEqual(...)` failed! Doing a string comparison seemed to work, though!

jeresig:
cc: @ariabuckles @xymostech

test/katex-spec.js
xymostech: I think that `toEqual` just does an `===` compare. You could try `expect(_.isEqual(tree, ...)).toBe(true)`?

xymostech: No, never mind, `toEqual` does do deep compare: [jasmine's deep `eq` function](https://github.com/jasmine/jasmine/blob/33641578e6f7d90be6155ea6dcb262b11a326fea/src/core/matchers/matchersUtil.js#L53). The only thing I can think of is that `"sub": undefined` should be `"sub": null`?

jeresig: Yeah - that was my thought as well. The method generates a `"sub": undefined`. I tried leaving it out and adding it in -- it failed in both cases. Not sure :(

jeresig:
I'm going to merge this so that I can finish up my react-components changes -- let me know if you want me to make any additional changes to the tests!

xymostech:
Oops, forgot this hadn't gotten merged in. Looks good to me! If the tests break at some point or something, I'll maybe change them, but if they're working locally and on Travis then :+1:

PR #228, opened 2015-05-11
Add quick syntax guide to README.md

dt1973:


kirbyfan64:
I'd like it more if there was something that said, "Hey! Not all of MathJax is supported!"

kevinb7:
It would be better to link to https://github.com/Khan/KaTeX/wiki/Function-Support-in-KaTeX so that people know what's supported and what isn't.  Also, stating that "KaTeX uses MathJax syntax" is a little weird.  It would be more precise to say that KaTeX uses TeX syntax, but I feel like that's already covered in the first line of the README which states: "KaTeX is a ... library for TeX math rendering"

kevinb7:
I added a link to the wiki page with the list of commands.  If people are looking for information on syntax then we could link to some `TeX` and/or `LaTeX` resources.

kevinbarabash:
I'm closing this b/c I think it's been addressed.  @dt1973 please reopen if you think we should add more information to the README.md.

PR #232, opened 2015-05-13
unnecessary left (

pkuphy:
Maybe the left most `(` is unnecessary.


spicyj:
It's useful to compare to the adjacent parenthesis so I think we'll keep it. Thanks for sending this in though!

Issue #206, opened 2015-03-16
Add support for \begin \end environments


rymohr:
Found a couple related issues and just wanted to add a central one to tie everything together:

- align support https://github.com/Khan/KaTeX/issues/61
- array support https://github.com/Khan/KaTeX/issues/47
- matrix support https://github.com/Khan/KaTeX/issues/43

PeterBocan:
Is there somebody working on it?

rymohr:
Not that I know of, though @kevinb7 and @xymostech have given it some thought:
https://github.com/Khan/KaTeX/issues/43#issuecomment-55658982
https://github.com/Khan/KaTeX/issues/61#issuecomment-57091390

gagern:
Can we close this, now that #246 has been merged?

kevinb7:
I think we can close this because the infrastructure for ```\begin``` and ```\end```.  We'll leave the other issues open that deal with specific environments.

Issue #47, opened 2014-09-16
Array (or Tabular)  support is missing


lidel:
One of less known use cases for TeX on a webpage is [ability to display data in tabular form][2].

The Other Library does not implement tabular but they [suggest][1] array environment as a replacement:

```tex
\begin{array}{c|cc}
\hline A & B & C\\
\hline D & E & F\\
\hline
\end{array}
```

which renders:

![](https://meta-discourse.global.ssl.fastly.net/uploads/default/34517/98bd1768b9e42654.png)

Sadly, I was unable to render the `array` above at http://khan.github.io/KaTeX/

btw: there is a related issue about matrices: #43

[1]: https://github.com/mathjax/MathJax-docs/wiki/LaTeX-Tabular-environment
[2]: https://meta.discourse.org/t/editing-an-insert-table-option/19719/5?u=lidel

gagern:
My #246 which has just been merged provides the core functionality for arrays. But it doesn't do lines yet, neither horizontal nor vertical. Since lines were an essential part of the original post here, I'd leave this ticket open until lines are implemented as well. I have quite a number of things on my agenda, so if someone wants to take a stab at adding support for lines, feel free to do so.

xymostech:
@gagern I'm actually going to close this and add a new issue to specifically add horizontal and vertical lines. I think that will make it clearer why this issue is open.

Issue #43, opened 2014-09-15
add support for matrices


vks:
It would be great to support things like `\begin{pmatrix} 1 & 0 \\ 0 & 1 \end{pmatrix}` like MathJax does.

hobson:
+1 for {matrix} and {bmatrix}. Really need it for linear algebra on slides.com and reveal.js presentations

paxidently:
+1

slametapr:
+1 for matrix

karoltarasiuk:
oh yeah +1

PeterBocan:
waiting for that day and then I'll move my ass from MathJax to KaTeX. +1!

RodrigoCard:
I am allowed to say +1 :+1:  here too? :)

gagern:
Can we close this, now that #246 has been merged?

xymostech:
Yes! Fixed by #246

PeterBocan:
Uhm... I'm feeling the next upcoming release gonna be soon with a whole lotta of sweets.

NateZimmer:
+1 ... was this ever implemented?

gagern:
@NateZimmer: yes, #246 implemented this, and KaTeX releases since [0.4.0](https://github.com/Khan/KaTeX/releases/tag/v0.4.0) include that code. Simply give it a try on https://khan.github.io/KaTeX/. If something about matrices doesn't work as expected, that would likely be a separate issue.

PR #246, opened 2015-06-12
Matrices, arrays, environments

gagern:
This is a first step towards support for matrices, and it addresses issues raised in

* #206 – support for `\begin`…`\end` environments
* #47 – `\begin{array}` (with argument) in particular
* #43 – various matrix flavours like `matrix`, `bmatrix`…

In particular, I've implemented parser support for environments, made environments available for `array`, `matrix`, `pmatrix`, `bmatrix`, `vmatrix`, `Vmatrix`, and added a proof-of-concept renderer for these.

This branch is really just a first step, one day of work put into this. This pull request is not meant as “please pull this *now*” but rather as a platform where future development can be discussed and coordinated, to result in a pull eventually.

My branch doesn't make much effort to do spacing the same way TeX does, mainly because I don't own a copy of Knuth's TeX Book. It doesn't work well with small font sizes. For really large numbers of rows, the delimiters appear too small in my opinion. But before I continue on these details, I'd like to know whether others have already put in some effort here. Perhaps someone has all the spacing calculations already worked out, and was just waiting for parser support to hook this up? I'd also like to know whether the way I tackled the problem, extending the parser in particular, is going in the right direction.

The way I see it, KaTeX has some knowledge of how *high* its boxes are, both above and below the baseline, but none about how *wide* they are. Is this correct? Seeing this, I decided that the most feasible way to implement HTML output for matrices would be column-wise: each column is a vlist, with items positioned according to the overall height and depth of the corresponding matrix row. Perhaps this will even allow matrices to line-wrap in certain environments. I've got some doubts whether the same approach would be reasonable for `align*` and similar environments. Should we stick to this approach, or should aligned equations and/or matrices be encoded to a `<table>` instead?

I can't make any guarantees how far I'll be able to follow this up. I'm doing this for a project at TU München, and once it satisfies the requirements we have there, continued development would likely have to be in my spare time, which is a rare commodity. How feature-complete does the thing have to be before it can get merged into the master?


kevinb7:
@gagern Thank you for this pull request.  It looks like you have some lint errors.  Please have a look at those.  You should also write some automated tests as well as image comparison tests.  Please have a look the /dockers/Screenshotter folder for instructions.  Also, you'll need to sign a CLA.  See https://github.com/Khan/KaTeX/blob/master/CONTRIBUTING.md#cla for more details.  All of these things are required before merging.

We would like the output of KaTeX to match TeX as much as possible.  Here's a link to the TeXBook: http://www.ctex.org/documents/shredder/src/texbook.pdf.  

@xymostech should be able to answer your question about whether to output tables column-wise vs. row-wise.  As for knowing about the height and/or width of its boxes, it should have enough information to calculate both because we have glpyh metrics for everything.

spicyj:
@xymostech and I came to the same conclusion as you: that each column should be a vlist.

src/environments.js
```diff
@@ -0,0 +1,104 @@
...
+    greedingess: 1,
+    allowedInText: false,
+    numOptionalArgs: 0,
+    begin: function() { /* default is no operation, end does the work. */ }
+};
+
+function array() { }
+array.prototype = new EnvironmentBase();
+array.prototype.numArgs = 1;
+array.prototype.end = function() {
```
kevinb7: Putting props on the prototype is going to mess up nested environments.  These should be member variables that are defined in the constructors.

src/Parser.js
```diff
@@ -199,6 +200,37 @@ Parser.prototype.handleInfixNodes = function (body, mode) {
...
+ * Turn an array-like data structure, with rows delimited by \\ and
+ * columns delimited by &, into nested lists of groups.  This method
+ * isn't directly called by the parser, but may be called by some
+ * environments.
+ *
+ * @param {string} mode       the mode to assign to newly created groups
+ * @param {ParseResult} body  the overall parse result to be decomposed
+ * @param {Array}             decomposed data in row major mode
+ */
+Parser.prototype.decomposeArray = function(mode, body) {
```
kevinb7: This is only used in environments, maybe it should go in that file.

src/environments.js
gagern: The properties in question are read-only, and even if some instance were to assign to them, those assignments would be local to the instance, since the values are immutable. I thought about having them as properties of the constructor instead of the prototype, but elements of the instance allow for two similar environments to share one implementation even if their number of arguments differs. Have I convinced you that this use here is all right?

src/Parser.js
gagern: That is where I had them first, but it needs the `ParseNode` constructor, which I hadn't made part of `Parser`'s prototype at that time. Since I later on needed `ParseNode` for the `leftright` group, and moving that to the parser felt even more wrong, I can move this decomposition to the environments. I guess it would be cleaner to export `ParseNode` without reference to the instance, preferably by moving it to a separate file.

gagern:
I wrote some parser test cases for environment parsing. A more specific test case for array doesn't fit in well with what's currently in `katex-spec.js`. I wonder whether it would be a good idea to test input string against resulting parse trees, saved in a big JSON list. Of course noone writes down the expected parse tree manually, but one may well verify it before saving the file. What do you think?

Screenshot tests seem a bit premature. Since I know that I'll probably be modifying the layout, I'd rather postpone that till shortly before merging. Regarding CLA I'll have to check with my university.

src/environments.js
kevinb7: I read through this a little to quickly on the first take.  I see that you're creating new EnvironmentBase() objects so this should be okay.

src/Parser.js
kevinb7: I think it make sense to have ParseNode in a separate file.

src/Parser.js
```diff
@@ -371,10 +403,54 @@ Parser.prototype.parseImplicitGroup = function(pos, mode) {
...
+                this.lexer, begin.result.value.namepos);
+        }
+        // Build the environment object. Arguments and other information will
+        // be made available to the begin and end methods using properties.
+        var env = new environments[envName](envName, this);
+        env.parser = this;
+        env.lexer = this.lexer;
+        env.mode = mode;
+        env.beginNode = begin;
+        env.args = [];
```
kevinb7: If all of these things are needed by an environment, why not just pass them as arguments to the constructor.  That way people won't forget to set these values.

kevinb7:
@gagern I think it's a good start for the katex-specs.  Can you add tests for nested matrices as well as embedded fractions using ```\over``` within the matrices?  Parse tree comparisons are a bit finicky because if something changes then it's a lot of work to redo them.  There's a few tests that look at some of the trees.  I don't think anything more comprehensive is worth the effort.

The nice thing about the screenshot tests is that if you change things in the tree but it still renders the same then it still passes.  Only create screenshot tests when you're happy with the layout, although I wouldn't mind seeing a couple screenshots here in this thread.

src/Parser.js
gagern: Not needed by most environments, but offered to them. The most important of these are `args`, `argPositions` and of course `body`. These can't be passed to the constructor. But even if they could, I'd rather not do this. There are potentially many environments, but there is only one place where instances of these are created. So having that code just in this one place avoids code duplication. Particularly since environments, the way I implemented them, don't call a base class constructor for common functionality.

kevinb7:
@gagern I'm going to do a more thorough review this weekend.  In the meantime, if you have time, focus on the layout.

gagern:
Up till today I had assumed that delimiter parentheses would scale smoothly between different sizes. Now I know that they grow in increments, both in LaTeX and in KaTeX. However, the points at which they grow seem to be very different:

Screenshot comparing KaTeX with LaTeX

This is simply `$$\begin{pmatrix}1\\[1.52ex]2\end{pmatrix}$$` rendered but with different values plugged into that extra row space argument. How much effort should we put into delimiters to get the sizes just as LaTeX uses them?

Apart from this, the spacing of my arrays should be pretty close to what LaTeX does now, since I've put in some effort to reproduce those algorithms. But if delimiters are going to change (and I believe that we should be switching to bigger delimiters a bit earlier), then I'd rather see that handled before creating screenshots for automated testing.

gagern:
I can't work out why the delimiters are smaller than they are in LaTeX. The computation from Rule 19, with `\delimiterfactor=901` and `\delimitershortfall=5pt` is reproduced in `makeLeftRightDelim`. The `\left`…`\right` documentation on page 292 doesn't seem to add any space either. (Note: page 359 lists more delimiters than KaTeX supports.) Page 442 suggests

> See the METAFONT manual or the system documentation of `tfm` files for further information about successors and extensible characters.

I don't have that manual, and I don't know what file this tfm documentation could be. Perhaps either of these contains some more details on how extensible characters should be composed from their parts? Or perhaps the font we use doesn't exactly match what LaTeX uses? Right now I don't know how to proceed.

kevinb7:
Please have a look at https://github.com/xymostech/KaTeX/commit/422c77b147e81785ab5d47d52b4dcf6b7e33135e, in particular buildHTML.js and environment.js.  The reason why we're not merging that branch in is because we discovered a bad interaction with infix operators such as ```\over``` and @xymostech hasn't had time to work on it.

That's unfortunate that the TeXBook isn't more elucidating in this situation.  There's a reference in the commit to page 153 of "TeX for the Impatient" that may be more helpful.  Here's a link that book (hopefully it's the same version): http://www.gnu.org/software/teximpatient/.

kevinb7:
I just saw your screenshots (I should've scrolled up sooner).  We definitely want to match the delimiter sizes.

gagern:
My code won't work with infix operators either (so far). I haven't yet decided how to tackle this. I could either have `parseImplicitGroup` return `null` for every `&`, `\\` or `\cr`, and have some outer function to collect the implicit groups of the alignment. Or I could try to parse the list after the fact, including handling for the alignment commands in `handleInfixNodes`. The latter looks closer to what I already have, but I fear that other stuff is also affected by the implicit grouping, e.g. style commands. `\small a & b` must get parsed as `{\small a} & b` not `\small {a & b}`. So I tend towards the former.

Page 153 of TeX for the Impatient describes things like `\lineskiplimit`. But as far as I can tell, that only affects text mode vboxes. In `$$\begin{matrix}\dfrac88\\\dfrac88\end{matrix}$$` the two matrix elements will always overlap, no even if `\lineskiplimit` and `\lineskip` are set to `100pt` each. So I'd say that was a bit too impatient here.

kevinb7:
@gagern I cloned your repo as well as the other repo and have the two branches in separate working folders so I can compare them and see if I can help figure out the cause of the layout differences.

kevinb7:
screen shot 2015-06-13 at 8 27 17 pm

The correct rendering is on the right.  On the left, the vlist is too high and there's not enough space between the 1 and the 2.  As for actually positions, using the default font size the vlist child containing the ```1``` should be at ```top: -0.52778em;``` as opposed to ```top: -0.61em;``` and the ```2``` should be at ```top: 0.67222em;``` instead of ```top: 0.59em;```.

Another interesting thing that probably doesn't matter is that the order that vlist children appear in the DOM is reverse.

gagern:
There is how LaTeX sets this `\begin{pmatrix}1\\2\end{pmatrix}`. Display style or not makes no difference.

```
.\mathon
.\hbox(14.5001+9.50012)x19.7223
..\hbox(0.39998+23.60025)x7.36115, shifted -14.10013
...\OMX/cmex/m/n/10 ^^R
..\glue -5.0
..\vbox(14.5+9.5)x15.00002
...\hbox(8.39996+3.60004)x15.00002
....\glue(\tabskip) 0.0
....\hbox(8.39996+3.60004)x15.00002
.....\rule(8.39996+3.60004)x0.0
.....\glue 5.0
.....\glue 0.0 plus 1.0fil
.....\mathon
.....\OT1/cmr/m/n/10 1
.....\mathoff
.....\mathon
.....\hbox(0.0+0.0)x0.0
.....\mathoff
.....\glue 0.0 plus 1.0fil
.....\glue 5.0
....\glue(\tabskip) 0.0
...\glue(\lineskip) 0.0
...\hbox(8.39996+3.60004)x15.00002
....\glue(\tabskip) 0.0
....\hbox(8.39996+3.60004)x15.00002
.....\rule(8.39996+3.60004)x0.0
.....\glue 5.0
.....\glue 0.0 plus 1.0fil
.....\mathon
.....\OT1/cmr/m/n/10 2
.....\mathoff
.....\glue 0.0 plus 1.0fil
.....\glue 5.0
....\glue(\tabskip) 0.0
..\glue -5.0
..\hbox(0.39998+23.60025)x7.36115, shifted -14.10013
...\OMX/cmex/m/n/10 ^^S
.\mathoff
```

So each line has a height of 8.39996pt+3.60004pt=12pt, with zero space between them. And with ptPerEm=10 this makes 12pt=1.2em=0.61em+0.59em. So I'd say that my rendering is more correct here. It also agrees with what LaTeX does:

Comparison of three two-row vectors

I'm not worries about the internal spacing of my array box, I have much confidence in that. What I am worries about is the delimiters, since as you can see in my previous screenshot, they don't agree with what LaTeX does. And the individual parts of the delimiters are not represented in the box structure either, so we can't compare at that level.

What we *can* see is that the size of the delimiters, 0.39998+23.60025=24.00023, is slightly bigger than the total height of the vbox, 14.5+9.5=24.0. This leads to 14.5001+9.50012=24.00022 as the height of the whole box with delimiters. If anyone can explain where one of these two bigger sums comes from, then we're making progress here.

gagern:
I [asked for help on the TeX Stack Exchange](http://tex.stackexchange.com/q/250147/16923). Perhaps someone there knows where these numbers come from.

kevinb7:
@gagern I'm surprised that LaTeX's rendering isn't centered.  I just assumed that is should be centered.

I was able to modify your branch to reproduce @xymostech's layout.  Getting the delimiters to be the right size is accomplished by increasing the height and depth of the span returned by ```buildHTML::array(array, group, options)``` by adding ```baselineSkip - 1.0``` to each.  

It's interesting that LaTeX overlaps the ```\dfrac{8}{8}```s.  Personally, I feel like increasing the gap would make for better rendering, but let's stick with replicating LaTeX's behavior.  I'll make a branch with the contents centered and the other spacing (for posterity).

Here are a couple of screenshots with the increased delimiter size (1.52ex, and 2.17ex):
screen shot 2015-06-13 at 10 12 44 pm screen shot 2015-06-13 at 10 12 52 pm

gagern:
Now that I think of it, perhaps the difference is simply due to the way the next suitable delimiter size gets chosen. But it's not that easy. With `$\begin{pmatrix}1\\[2.16ex]2\end{pmatrix}$` I get

```
.\hbox(19.15+14.14998)x20.83339
..\hbox(0.39998+29.60031)x7.91669, shifted -17.10016
...\OMX/cmex/m/n/10  
..\glue -5.0
..\vbox(19.15+14.14998)x15.00002
```

with 19.15+14.14998=33.29998 outer and inner size but 0.39998+29.60031=30.00029 delimiter size. So the assumption “the delimiter is no smaller than the content” is incorrect. And I was wrong: one *can* see the parts of a big delimiter, the examples so far just weren't big enough. With `$\begin{pmatrix}1\\[2.17ex]2\end{pmatrix}$` LaTeX suddenly produces

```
.\hbox(20.50017+15.50017)x22.50005
..\vbox(0.39998+35.60036)x8.75002, shifted -20.10019
...\hbox(0.39998+17.60019)x8.75002
....\OMX/cmex/m/n/10 0
...\hbox(0.39998+17.60019)x8.75002
....\OMX/cmex/m/n/10 @
..\glue -5.0
..\vbox(19.17151+14.17151)x15.00002
```

i.e. a paren split in two parts. It takes 3.1ex till KaTeX uses two parts there. At that point, it reports delimiters with a size of 1.155+0.64502=1.80002em, which is a good match for the 0.39998+17.60019=18.00017pt in LaTeX. So it knows how big its delimiters are. But what about the content? Still with 3.1ex, `makeLeftRightDelim` sees 19.2249+12.5805pt while LaTeX reports 
21.1736+16.1736 for the inner vbox. So the outside height of that vbox is wrong.

gagern:
Found my mistake. Now I'll only have to get docker up and running, and then I'll be able to take some snapshots. But not today.

kevinb7:
I pulled the ```baselineSkip - 1.0``` number from environments.js on xymostech's branch.  Note: this should only be added for the ```matrix``` environment and not the ```array``` or ```align``` environment.  We'd also like to support the ```align``` environment because it's used heavily at Khan Academy, but I wouldn't worry about it b/c there's enough other stuff going on in this pull request.

As for the issue of infix operators, I think maybe ```parseExpression``` could be modified to accept multiple tokens to break on.  Also, it seems like instead of calling ```parseExpression``` after we see ```\begin{array}``` we should probably have a ```parseArray``` function so that it can recursively call ```parseExpression``` using ```&``` and ```\\``` as tokens to break on as opposed to processing the array later in ```decomposeArray```.

kevinb7:
> Found my mistake. Now I'll only have to get docker up and running, and then I'll be able to take some snapshots. But not today.

Excellent!

gagern:
>  I think maybe parseExpression could be modified to accept multiple tokens to break on.

Do we need an argument list for this, or can we hard-wire this? Is there ever a situation where we want `parseExpression` to continue past one of `}`, `\right`, `\end`, `&`, `\\`, `\cr`, `\egroup`? I can think of no use case where this would make sense, so I'd terminate on all of these, and the caller can then ensure that the current lexer token is the one (or among those) it expects.

kevinb7:
The mode and environment dictate which tokens to break on, e.g. when inside ```\text``` we shouldn't break on ```&``` and when we're in "math" mode but not inside an "array" environment ```&``` should cause an error.  

Hard coding this inside ```parseExpression()``` would require us to put logic to deal with a particular environment and/or mode inside ```parseExpression()```.  Changing ```breakOnToken``` to ```breakOnTokens``` gets us what we need with minimal modifications to the rest of the codebase.

gagern:
When we are inside `\text`, we break on all the things I mentioned, and report an error if it's not `}`. After all, a lone `&` or `\end` inside a `\text` is an error, and `expect` can be used to report that. So instead of letting `parseExpression` decide whether to break or error, I'd let it break in all cases and leave generating the error to the caller.

gagern:
Come to think of it, we probably need this behavior in any case. When `parseImplicitGroup` has to handle a size or style function, that implicit body has to end at `&` if we are inside a matrix, and has to end at `\end` if we are inside an environment, and so on. So if we don't want to pass the list of break symbols all the way down to the `parseImplicitGroup` call, then we have to make sure that we don't need such a list in the first place.

kevinb7:
I tried hardcoding it, but there's a few errors that appear in the tests.  Here's the code that I used:
```
Parser.prototype.parseExpression = function(pos, mode, breakOnInfix) {
    var body = [];
    while (true) {
        var lex = this.lexer.lex(pos, mode);
        if (["}", "]", "&", "\\right", "\\end"].indexOf(lex.text) !== -1) {
            break;
        }
        var atom = this.parseAtom(pos, mode);
        if (!atom) {
            break;
        }
        if (breakOnInfix && atom.result.type === "infix") {
            break;
        }
        body.push(atom.result);
        pos = atom.position;
    }
    return new ParseResult(this.handleInfixNodes(body, mode), pos);
};
```

Here are the resulting errors which I haven't had time to look into:
```
Failures:

  1) A close parser should not fail
    Message:
      ')]?!' failed parsing with error: KaTeX parse error: Expected 'EOF', got ']' at position 2: )]̲?!


    Stacktrace:
      Error: ')]?!' failed parsing with error: KaTeX parse error: Expected 'EOF', got ']' at position 2: )]̲?!
    at Object.<anonymous> (/Users/kevin/KaTeX/test/katex-spec.js:229:28)


  2) A close parser should build a list of closes
    Message:
      ')]?!' failed parsing with error: KaTeX parse error: Expected 'EOF', got ']' at position 2: )]̲?!


    Stacktrace:
      Error: ')]?!' failed parsing with error: KaTeX parse error: Expected 'EOF', got ']' at position 2: )]̲?!
    at getParsed (/Users/kevin/KaTeX/test/katex-spec.js:26:18)
    at Object.<anonymous> (/Users/kevin/KaTeX/test/katex-spec.js:233:21)


  3) A close parser should build a list of closes
    Message:
      ParseError: KaTeX parse error: Expected 'EOF', got ']' at position 2: )]̲?!


    Stacktrace:
      ParseError: KaTeX parse error: Expected 'EOF', got ']' at position 2: )]̲?!
  at new ParseError (/Users/kevin/KaTeX/src/ParseError.js:29:16)
  at Parser.expect (/Users/kevin/KaTeX/src/Parser.js:86:15)
  at Parser.parseInput (/Users/kevin/KaTeX/src/Parser.js:112:10)
  at Parser.parse (/Users/kevin/KaTeX/src/Parser.js:100:22)
  at parseTree (/Users/kevin/KaTeX/src/parseTree.js:14:19)
  at getParsed (/Users/kevin/KaTeX/test/katex-spec.js:28:12)
  at Object.<anonymous> (/Users/kevin/KaTeX/test/katex-spec.js:233:21)
  at Timer.listOnTimeout [as ontimeout] (timers.js:112:15)



  4) An optional argument parser should not fail
    Message:
      '\frac[1]{2}{3}' failed parsing with error: KaTeX parse error: Expected 'EOF', got ']' at position 8: \frac[1]̲{2}{3}


    Stacktrace:
      Error: '\frac[1]{2}{3}' failed parsing with error: KaTeX parse error: Expected 'EOF', got ']' at position 8: \frac[1]̲{2}{3}
    at Object.<anonymous> (/Users/kevin/KaTeX/test/katex-spec.js:1343:35)

```

kevinb7:
@gagern please note, these failures are for the code without your changes.  I'm on my personal computer and I don't have a copy of your repo, but I assume that the failures would be similar.

kevinb7:
Using ```parseExpression``` might be more difficult that than I originally thought.  The problem is that we we want each cell in the array to be parsed as a separate expression.  These are delimited by ```&``` (and rows by ```\\```).  At the same time we want ```parseExpression``` to return when it it hits a ```\end```.  The only problem is that we can't tell the difference between ```ParseResult```s in the middle of the array and ```ParseResult```s at the end.  In order to get ```parseExpression``` working correctly in this situation we'd have to have it return which token it's halting on.

I suppose another approach would be to modify ```handleInfix``` to handle the ```&``` and ```\\``` when inside certain environments.  It would be nice if we could call ```decomposeArray``` array first, then we could call ```handleInfix``` on each of the cells.

gagern:
@kevinb7 

> `if (["}", "]", "&", "\\right", "\\end"].indexOf(lex.text) !== -1) {`

`]` is a special case, since it's an ordinary symbol except when parsing an optional argument. So we should keep the breakOnToken approach for that, in addition to the hard-wired expression endings.

> In order to get `parseExpression` working correctly in this situation we'd have to have it return which token it's halting on.

I've let `parseExpression` return the token it's halting on as a `peek` property. But that's just for the sake of performance. Theoretically we could lex again starting at the returned position to see what token comes next. I think tat on the whole we might be lexing a lot of stuff more than once. Might be worthwhile to investigate whether this is really the case, and if so decide on how to mitigate the issue. But that would be a separate ticket.

kevinb7:
> ```]``` is a special case, since it's an ordinary symbol except when parsing an optional argument.

That makes sense.  Removing it and keeping breakOnToken for that case fixed the failing test cases.

I had a look at your last commit and it seems like that should work with infix operators.  Can you write some tests to verify that this is now the case?

I like how ```parseImplicitGroup``` cleaned up.

gagern:
I testes one infix fraction, and it worked. Will write test cases and commit them when I find the time to do the snapshots, to add both in one go. I'm mostly on OS X at work, while I assume that docker only runs on Linux. Had to reconfigure my kernel till it should run, haven't tried it yet. Have you ever considered using [PhantomJS](http://phantomjs.org/) to [create snapshots](http://phantomjs.org/screen-capture.html)? I don't know whether having different versions of libraries instaled could lead to different results with this, but it might be woth investigating.

kevinb7:
I'm glad to hear that infix is working.   Thank you for all of your hard work on this.  

I work on OS X as well and was able to use the docker.  If you have an old install of docker, please upgrade.  It fixes some issues when doing network communication between the host and the vm.  I'm not sure why we went docker and not phantomjs.  I do know that different versions of libraries leads to different results.  I spent some time trying to automate the snapshotting using travis-ci and I had to install a very specific version of freetype to get the results to match.  I would definitely be in favor of testing harness that requires less setup as long as we can get consistent results across platform b/c not all devs are working on OS X.

gagern:
Finally got docker up and running in my VirtualBox. That's one huge chunk of software for just a few snapshots! And on *some* runs the `Lap` example gave a different result. I don't know what has me more worried: the fact that it had a different result, or the fact that even after all this effort, the result wasn't perfectly reproducible.

Now I'm trying to design one good example for arrays and matrices. And in the process I realized that I'm not strictly behaving the way LaTeX does. So far I've inherited the size and style from the surrounding environment. In LaTeX an inner matrix set in a cell with `\scriptstyle` is nevertheless full-sized. Reducing matrices in size is a very tricky problem, which either requires the `smallmatrix` package, or executing a size command outside math mode, i.e. in a nested text box. Currntly KaTeX can't handle mode changes (it doesn't know about `$` or `\mbox`). On the other hand, it does allow size changes in math mode, which are forbidden in LaTeX.

So I wonder, how much should we support the anti-features of LaTeX? Should I change the style back to normal when building the matrix cells? Should I do something about the size? Should I capture any of these aspects in my main screenshot for arrays, or should I rather demonstrate undisputed behavior there, and leave such details to separate screenshots which can be changed if we change our opinion? Or not create a screenshot at all, to leave things unspecified?

kevinb7:
We've had trouble with the reproducibility of  ```Lap```.  If some of the character positions are off vertically by a couple of pixels please ignore that test and don't commit that file.  Whenever I run that test it's slightly different from the reference file.  That's something we need to look into.

> In LaTeX an inner matrix set in a cell with \scriptstyle is nevertheless full-sized. 

You're talking about nested matrices, that might be used for tensors, right?  I don't think we're too concerned about this particular use case right now.  Let's put a pin in this for now and we can deal with this in a future change set.  Once we merge this in maybe you can create an issue for the non-compliant behavior you've described.

Create screenshots for the undisputed behavior as suggested.  I assume by this you mean non-nested arrays and matrices and an example of infix operators being rendered correctly.

gagern:
> That's something we need to look into.

If I had to guess, I'd say the HTML engine hasn't settled down by the time the screenshot is taken. But that's just guessing. Would be interesting to try whether a 2 seconds pause before snapshotting will lead to a reproducible behavior there…

> You're talking about nested matrices, that might be used for tensors, right?

Not neccessarily. When writing inline math, using a smaller font makes sense for even very small matrices like 2×2.

> Once we merge this in maybe you can create an issue for the non-compliant behavior you've described.

Fine with me.

> I assume by this you mean non-nested arrays and matrices and an example of infix operators being rendered correctly.

I can nest them all right, I just can't scale them down. Will commit that tomorrow, have that work begun on my other computer.

xymostech:
Just leaving some small comments, this is looking great!

>  I think tat on the whole we might be lexing a lot of stuff more than once.

This is indeed the case, we haven't tried to do any caching or anything to help the performance of this at all. If just re-lexing is easier, I think that's a totally fine approach at this point.

> Currntly KaTeX can't handle mode changes (it doesn't know about $ or \mbox). On the other hand, it does allow size changes in math mode, which are forbidden in LaTeX.

This is something that we've had varying ideas about. I think generally, the sizing commands like `\small` must work outside of the environments, and should probably do something reasonable inside.

> And on some runs the Lap example gave a different result.

Like @kevinb7 says, this is just a funky result due to how rendering works sometimes. You're not doing anything wrong, it's due to some funkiness in how firefox renders stuff (if you delay taking the screenshot for any amount of time, the problem goes away).

gagern:
I added a screenshot which covers many of the aspects of how arrays and matrices are rendered. Here is a comparison of that screenshot against what LaTeX renders for me:

Comparison of renderings

The black parts are black in KaTeX but white in LaTeX, the white parts the other way round. As you can see, most parts match pretty good. Fonts in KaTeX tend to be a bit thicker. Both LaTeX and LaTeX have tiny gaps in the right delimiter, but KaTeX has two of them. The biggest differences are with the positioning of the fraction, and the spacing around the solidus. I still have to investigate the causes for these discrepancies. Since I don't know how long this investigation might take, feel free to merge this once you consider it good enough, and I can follow up any remaining issues in separate tickets.

kevinb7:
That looks pretty good.  I was talking with @xymostech and she says that we forget to put some space around fractions.  It's a separate issue so don't worry about it.  Could you check what happens when sizing commands are used inside the environment?  Even if the output doesn't match LaTeX's output for all different sizes, as longs as the rendering does something reasonable it should be okay.

xymostech:
That image diff looks pretty great! I think that the discrepencies around the solidus are probably just font differences, almost certainly unrelated to the matrix stuff.

Also, indeed KaTeX fonts are thicker, they're designed that way (to match what MathJax does, and because it's easier to read at smaller font sizes): https://github.com/Khan/MathJax-dev/blob/master/fonts/OTF/TeX/Makefile#L28

gagern:
> Also, indeed KaTeX fonts are thicker, they're designed that way

Thought so, makes a lot of sense on screen, since we can't expect the kind of resolution a printer would have.

> we forget to put some space around fractions. It's a separate issue so don't worry about it.

I noticed you opened #249 for that. Nice. (It might have been even nicer if that post had given me some kind of credit, for creating the image or for finding the issue.) If you have an idea of where this is coming from, I'll not investigate this myself any further just now.

> Could you check what happens when sizing commands are used inside the environment? Even if the output doesn't match LaTeX's output for all different sizes, as longs as the rendering does something reasonable it should be okay.

The example with the 1/2 in solidus notation is using an explicit `\scriptstyle`. If you don't count that as a sizing command (I guess in TeX parlance it is not) then I can certainly downsize another entry as well, using `\scriptsize` for a change. But that would mean we're testing for some deliberate deviation from LaTeX, and I thought that I should restrict the screenshot to the undisputed behavior. Or is it an undisputed decision that sizing in math mode should do reasonable things, even if doing so deviates from LaTeX? I'd have assumed this to be controversial. Long story short: It does work reasonably, for sizing as well as styling, but I'd rather not test for it unless you insist.

If you don't insist, can this be merged? A CLA for me should have arrived by now.

(By the way: I'm also working on rendering KaTeX to HTML5 canvas. Is that of interest to the project? If so, I'll post a pull request once I'm reasonably satisfied with things.)

bryevdv:
> (By the way: I'm also working on rendering KaTeX to HTML5 canvas. Is that of interest to the project? If so, I'll post a pull request once I'm reasonably satisfied with things.)

Whoa whoa whoa this would be an ***amazing*** development for Bokeh (https://github.com/bokeh/bokeh, https://github.com/bokeh/bokeh) Do you have any rough estimated timeframe?

spicyj:
I'd be interested in seeing the canvas rendering.

spicyj:
@bryevdv Curious: why would KaTeX canvas rendering be important to Bokeh?

bryevdv:
LaTeX labels is one of the longest and most requested features for Bokeh plots, and BokehJS renders in HTML5 canvas. I was about to give up and look at absolutely positioning transparent divs above the plot canvas but native rendering would be 1000 times preferable.

kevinb7:
> (It might have been even nicer if that post had given me some kind of credit, for creating the image or for finding the issue.)

Sorry about that.  I've amended the issue to give credit where credit is due.  I'm curious about how you managed to line things up.  Was there a lot of manual work creating that image?  I'm curious b/c it would be nice to have an easy way to compare KaTeX output against LaTeX output for a variety of things.

> If you have an idea of where this is coming from, I'll not investigate this myself any further just now.

I don't think this was introduced by your code nor is fixing it within the scope of this pull request.  It's a separate issue.

> But that would mean we're testing for some deliberate deviation from LaTeX, and I thought that I should restrict the screenshot to the undisputed behavior.

The screenshot test is for undisputed behavior and I think it's fine as is.  I was interested in determining whether or not using ```\small``` or ```\Large``` returns something reasonable, i.e. each cell is basically where it should be relative to the other cells, or whether it blows up and cells appear opposite of where they should be or they're intersecting the delimiters (or something crazy like that).  Sorry I haven't had time to play around with the recent changes that's why I was asking.

> If you don't insist, can this be merged? A CLA for me should have arrived by now.

I checked our spreadsheet and we did indeed receive a CLA from you.  Thank you.

I understand your eagerness to get this merged.  We're excited about these changes as well.  There are some code style changes that should be made before merging.  The style guide encourages writing code the blends in with the surrounding code.  This makes it easier for other people working on the project to grok your code which makes it more maintainable in the long run.

Environments are essentially the same as the entries in functions.js, except their handler methods should also some parsing.  See how ```duplicatedFunctions``` are defined in functions.js.  In particular look at the ```handler``` method for ```\\begin``` and ```\\end```.  It's ```this``` is bound to a ```Parser``` instance so it has access to the ```lexer``` as well as any parsing methods needs.  Lastly, look at ```parseFunction``` in Parser.js, specifically this line: ```var result = functions.funcs[func].handler.apply(this, args);```.  This is how the handler gets access to the ```Parser``` instance.

Please follow this pattern.  I have to admit that using ```apply(this, ...)``` is a little cryptic, but once it's in place, it will be very easy for people to add other environments because it will look similar to how functions are implemented.  If you have any questions about how to make these changes let me know.

There are some other more minor nits, but this is the big one.

> (By the way: I'm also working on rendering KaTeX to HTML5 canvas. Is that of interest to the project? If so, I'll post a pull request once I'm reasonably satisfied with things.)

Awesome!  I was thinking of doing myself at some point.  I'm not sure how useful it would be to Khan Academy, but I can see why people would want this.  I have some ideas on this but I think that the discussion would be better served in a separate issue.

kevinb7:
I tried out some different size commands inside a matrix and the output seems reasonable.
```\begin{pmatrix} \Huge{1} & \Large{2} \\ \small{3} & \tiny{4} \end{pmatrix}``` produces:
screen shot 2015-06-17 at 8 37 41 pm

**Comment by gagern** [comment]
> I'm curious about how you managed to line things up. Was there a lot of manual work creating that image? I'm curious b/c it would be nice to have an easy way to compare KaTeX output against LaTeX output for a variety of things.

What I did was this:
1. Render LaTeX to `article` PDF, crop that and import it into GIMP at high resolution.
2. Open screenshot from KaTeX and copy the imported PDF to that as a layer.
3. Resize that until the positions mostly match. This was a lot of trial and error.
4. Set mode to [grain extract](http://docs.gimp.org/en/gimp-concepts-layer-modes.html#idp14580272) to show differences.
5. Tweak position to minimize differences.

If you'd like to automate generation of images like these, I'd suggest an alternate approach.
1. Render to PDF to an as-small-as-you-can-make-it PDF. There is a class for that, don't remember its name just now.
2. Rasterize that PDF at the correct resolution so that we'd expect sizes to match. We know how big our font is in LaTeX, we know how big a TeX point is, and we know what resolution we use for screenshots. So we should be able to work out the correct resolution to use here.
3. Align images using an FFT-based correlation computation. I've done stuff like this before.
4. Compute differences by iterating over the pixels, to avoid dependency on GIMP.

> There are some code style changes that should be made before merging. …

I'll have a look. My main reason for not following the `functions` pattern in the first place was that I expected most environments to carry information from `begin` to `end`. Now that that is under the control of a single function, there might be no strong reasons left for this approach.

>  I have some ideas on this but I think that the discussion would be better served in a separate issue.

So let's continue the render-to-canvas discussion on #251. I'm looking forward to hearing your ideas.

**Comment by fpliger** [comment]
Wow! I second @bryevdv enthusiasm on this! :+1:

**Comment by gagern** [comment]
@kevinb7 wrote:

> Environments are essentially the same as the entries in functions.js, except their handler methods should also some parsing.

I implemented things this way, but I'm not happy about it.

* Abusing `this` to refer to the parser feels a bit counter-intuitive. What makes this argument different from all the others, that it has to get passed another way? If anything deserves the name `this` in this context, it's the name of the function resp. environment.
* Stuffing the arguments to `begin` in among all the other fixed arguments like `pos` and `mode` feels a bit confusing, I think.
* Having everything in one big array means 8 spaces more of indentation, i.e. 1/10 of the allowed width wasted before the handler even starts.
* Having everything in one big array means it's harder to plug in helper functions (like `parseArray`, but perhaps imagine something with a more narrow scope) close to the code that's using them.
* It also means helper data (like the map from environment name to delimiter pair) can't simply be a static variable associated with the prototype.

But I agree that it makes the code more similar. Just wanted my objections noted. I'd be glad to change both `environments` and `functions` together to mitigate some of these aspects.

**Comment by kevinb7** [comment]
@gagern I agree with a lot of what you're saying here.  It might be good to revisit some of these design decisions.  Could you open a separate issue so that we can continue this discussion there?

Everything looks good to go.  We'd like to have a clean history.  Can you squash this into a single commit?

Thanks so much for this pull request.

File: src/Parser.js [hunk 4]
```diff
@@ -127,11 +114,16 @@ Parser.prototype.parseInput = function(pos, mode) {
...
 Parser.prototype.parseExpression = function(pos, mode, breakOnInfix, breakOnToken) {
     var body = [];
+    var lex = null;
     // Keep adding atoms to the body until we can't parse any more atoms (either
     // we reached the end, a }, or a \right)
     while (true) {
-        var lex = this.lexer.lex(pos, mode);
-        if (breakOnToken != null && lex.text === breakOnToken) {
+        lex = this.lexer.lex(pos, mode);
+        if (["}", "\\end", "\\right", "&", "\\\\", "\\cr"]
```
xymostech: Could you move this list out into a constant somewhere?

src/Parser.js
```diff
@@ -127,11 +114,16 @@ Parser.prototype.parseInput = function(pos, mode) {
...
+            .indexOf(lex.text) !== -1) {
+            break;
+        }
+        if (breakOnToken !== null && lex.text === breakOnToken) {
```
xymostech: If you make this `breakOnToken != null`, then you can just make the last argument optional instead of having to pass `null` in everywhere.

xymostech:
Hi! Thank you so much for this! Everything looks great! I'm just going through and making some small javascript nits, but nothing major. Thanks!

src/delimiter.js
```diff
@@ -231,28 +231,24 @@ var makeStackedDelim = function(delim, heightTotal, center, options, mode) {
     var repeatHeightTotal = repeatMetrics.height + repeatMetrics.depth;
     var bottomMetrics = getMetrics(bottom, font);
     var bottomHeightTotal = bottomMetrics.height + bottomMetrics.depth;
-    var middleMetrics, middleHeightTotal;
+    var middleMetrics = 0, middleHeightTotal = 0, middleFactor = 1;
```
xymostech: `middleMetrics` isn't a number, it's an object, so it probably shouldn't start out initialized to `0`.  Also, when we initialize and declare variables at the same time, we like to put them on separate `var` lines, so

```
var middleMetrics;
var middleHeightTotal = 0;
...
```

src/functions.js
```diff
@@ -517,6 +529,31 @@ var duplicatedFunctions = [
...
+            numArgs: 1,
+            argTypes: ["text"],
+            handler: function(func, nameGroup, positions) {
+                if (nameGroup.type !== "ordgroup") {
+                    throw new ParseError(
+                        "Invalid environment name",
+                        this.lexer, positions[1]);
+                }
+                var name = "";
+                for (i = 0; i < nameGroup.value.length; ++i) {
```
xymostech: this should be `var i`

test/katex-spec.js
```diff
@@ -880,6 +880,43 @@ describe("A left/right parser", function() {
     });
 });
 
+describe("A begin/end parser", function() {
+
+    it("should parse a simple environment", function() {
+        expect("\\begin{matrix}a&b\\\\c&d\\end{matrix}").toParse();
+    });
```
xymostech: Can you add a test that makes sure the `\cr` command also works?

gagern:
I squashed everything into one commit, except the improvements to the stacked delimiters, since these are actually a separate thing and not directly linked to all the rest. I had worked on that since I had assumed the reason for the spacing discrepancies somewhere in there, but in the end it was somewhere else. Hope that's OK for you.

kevinb7:
@gagern Thank you for addressing all of @xymostech's concerns.  I think it makes sense to separate the commits like you've done.  LGTM.

xymostech:
+1 on keeping the stacked delimiters stuff separate. Thanks for cleaning that code up!

Issue #45, opened 2014-09-16
Some characters have higher baseline than others


uranusjr:
![screen shot 2014-09-16 at 11 39 30](https://cloud.githubusercontent.com/assets/605277/4282229/9b44b0c6-3d53-11e4-997d-41183de6e48e.png)

Top: http://khan.github.io/KaTeX/ in iframe. Bottom: custom rendering.

The smaller characters have a higher baseline than others. The `k`s even look like superscript.

[HTML file](https://gist.github.com/uranusjr/2882e927d055d02891f7). I used the CSS/JS/font files directly from the v0.1.0 release. Tested on OS X 10.9.4. The above was rendered with Safari 7.0.6, but other browsers also yield the same result.

stefman2:
On a similar vein, I had noticed that, in my setup, which uses katex.renderToString(), KaTex was not positioning any subscripts lower at all. The CSS generated by KaTex seems to assume everything is already positioned, and has ...style="top:0.15em..." for positioning subscripts. So, to get it to work for me, I needed to change this to style="position:relative; top:0.15em..." in the output CSS and now it works.  (Explicitly setting a position attribute to "relative" in <span> tags containing the entire KaTex expression did not seem to work either.)
Any idea on what is causing KaTex elements to lose their position attributes, and whether Katex ought to rectify this in katex.renderToString()?

xymostech:
@stefman2 Are you including the katex CSS file on your website? You need the main KaTeX CSS (which should have `position: relative` rules) as well as the generated HTML from `katex.renderToString`.

stefman2:
Yes, I am including katex.min.css.  All the typesetting looks fine except for subscripts, and that minor change seems to do the trick as far as I can tell. I will look into this some more and try to isolate the problem; perhaps there is an issue with other CSS I have on the pages in question.

mbourne:
Hello again

I'm still seeing subscripts rendered as superscript (almost) in IE v11, Win 7 with KaTeX 0.5.0 js and css.

![image](https://cloud.githubusercontent.com/assets/7541940/9375113/69e3aef6-472f-11e5-9c3b-2bf8f6a74d2e.png)

And on the same browser, superscripts end up too high 

![image](https://cloud.githubusercontent.com/assets/7541940/9375147/ca15c50c-472f-11e5-96fb-e9ed77282aab.png)

Here's the same portion on Chrome:

[...]

PR #251, opened 2015-06-18
Render to HTML5 canvas

gagern:
Since my [announcement of canvas rendering](/Khan/KaTeX/pull/246#issuecomment-112945759) was met with considerable interest (by KaTeX developers as well as for bokeh/bokeh#647), I decided to share my work early on. At the time I'm opening this request, this branch is in a very early stage of development. But others are welcome to contribute, be it by commenting, by suggesting ideas, or by providing code in the form of pull requests against this branch.

After launching the local development server, see [http://localhost:7936/canvas.html](http://localhost:7936/canvas.html) for an example of rendering to canvas.


bryevdv:
This is very exciting, I am at a conference this weekend but I should be able to take a closer look next week. I am happy to help out in any way I can. For some context about the Bokeh use-case, we'd like to be able to render math text into an existing canvas at a specified locations (similar to ``ctx.fillText``).

gagern:
This can now reproduce most of the formulas used for the snapshot tests. `\rule` is the only exception that I'm aware of. I'm surprised by how quickly this is progressing and nearing feature completeness. So if you notice anything which doesn't work besides `\rule`, please add a comment here and I'll investigate.

If anyone has experience with the [CSS font loading module](http://www.w3.org/TR/css-font-loading-3/) and/or the [Web Font Loader](https://github.com/typekit/webfontloader) project, I'd be glad if anyone would have a look at hooking up our fonts so that we don't have to guess and hope when they'll be available. It would be best if we could introduce two more options. One to control whether the formula should be rendered with incorrect fonts if the correct ones aren't available, as opposed to not rendering at all, and the other to register a callback which gets called if all (or perhaps even if more) fonts have become available, to trigger a redraw of the canvas.

kevinb7:
It's awesome to see how quickly this coming along.  I checked out the branch and gave it try.  The rendering seems blurry compared to the HTML.  It might have something to with being on a retina device.  Also, there was some clipping going on, but I think adjust the size of the canvas to match the bounding box of the output should fix that.

Canvas:
screen shot 2015-06-18 at 12 12 13 pm

HTML:
screen shot 2015-06-18 at 12 17 49 pm

gagern:
Yes, the canvas so far is simply fixed size and pixel density 1. For the latter, [this tutorial by Paul Lewis](http://www.html5rocks.com/en/tutorials/canvas/hidpi/) points the way towards a solution. Have done that in another project alreary, works like a charm. Haven't integrated it with the former, i.e. with auto-resizing canvas, yet, but it's certainly doable.

bryevdv:
I was just about to link that, we used the same article to develop Bokeh's support for HiDPI modes. @gagern do you think that the capability to render at a specified location onto an already existing, user-supplied canvas is feasible for this work? If not, it should be easy enough for us to maintain small private canvases for annotations, that can be quickly blitzed onto the main canvas, but I thought it would be worth asking about.

spicyj:
For retina support, I write this last night: https://gist.github.com/spicyj/1d9145ab46190782260f

gagern:
@bryevdv: Sure you can render to an existing canvas. Have a look at [these lines](https://github.com/gagern/KaTeX/blob/8651a784ed98a92ed99eb6b0ad5aa8f994773529/static/canvas.js#L32-44) for an example application. There I compute the box up front, then draw something else (in this case the bounding box and its base line) and then draw the box I computed before. If all you care about is render at a given position, you can call [renderToCanvas](https://github.com/gagern/KaTeX/blob/8651a784ed98a92ed99eb6b0ad5aa8f994773529/katex.js#L93) instead of [canvasBox](https://github.com/gagern/KaTeX/blob/8651a784ed98a92ed99eb6b0ad5aa8f994773529/katex.js#L94). It will prepare a box and [immediately](https://github.com/gagern/KaTeX/blob/4441413006696c91b567033e8b4b8aceed16a3ff/src/canvasRenderer.js#L558) render it at the given position. You can still control horizontal alignment using the [halign property of the options object](https://github.com/gagern/KaTeX/blob/4441413006696c91b567033e8b4b8aceed16a3ff/src/canvasRenderer.js#L544).

bryevdv:
@gagern that's great. We have a new release coming up in a few weeks but immediately after that we will start looking into integrating this as a high priority!

kevinb7:
@gagern I'd like to be able to do hit testing when using the canvas renderer.  This is for a personal project, but I think that other people might find that ability useful as well.  The code generates a flat data structure which is okay for select individual atoms, but makes it impossible to select groups.  I'm going to take a stab at refactoring the code this weekend so that ```CanvasRenderer.protoype.prepare``` returns a nested data structure as opposed to appending to ```this.outList```.

gagern:
@kevinb7 My main reason for this flat data structure is performance. The prepared box can be typeset fairly quickly as many times as you want to. Keep this in mind when refactoring, by either doing performance tests on different browsers (perhaps via jsperf?) to show that the cost of a nested data structure is negligible, or by keeping the flat structure as an option.

kevinb7:
@gagern I think I'll create a second renderer base on your implementation.  This will make it easy to compare the two approaches.

On another note, I tried using webfontloader with some success.  It doesn't recognize the KaTeX_Size fonts as being loaded.  The browser shows that they're loading fine.  I'll file a bug for this tomorrow. 

I added this script tag to canvas.html:
```
<script src="https://ajax.googleapis.com/ajax/libs/webfont/1.5.10/webfont.js"></script>
```
and this code at the bottom of canvas.js:
```
var activeCount = 0;

WebFont.load({
    custom: {
        families: ['KaTeX_AMS:n4', 'KaTeX_Main:n4,i4', 'KaTeX_Math:n4,i4', 
            'KaTeX_Size1:n4', 'KaTeX_Size2:n4', 'KaTeX_Size3:n4'],
    },
    fontloading: function(familyName, fvd) {
        console.log("loading: " + familyName + " " + fvd);
    },
    fontactive: function(familyName, fvd) {
        console.log("active: " + familyName + " " + fvd);
        activeCount++;
        if (activeCount === 8) {
            init();
        }
    },
    fontinactive: function(familyName, fvd) {
        console.log("inactive: " + familyName + " " + fvd);
        activeCount++;
        if (activeCount === 8) {
            init();
        }
    }
});
```
Here's what dev tools shows in the network tab.
screen shot 2015-06-26 at 12 19 48 am

kevinb7:
Found the issue:
http://stackoverflow.com/questions/6879469/google-typekit-webfont-loader-does-not-detect-non-ascii-fonts
> I'm one of the developers of the webfontloader. We are currently using an ASCII string to detect whether or not a font has loaded. So if your font does not include the characters "BESbswy" the webfontloader won't be able to detect it. We're working on adding support for configurable test strings (so you can include some characters from the Unicode range your font supports.) I've opened an issue for this on the webfontloader repository (https://github.com/typekit/webfontloader/issues/104).

Problem solved with this minor change:
```
    custom: {
        families: ['KaTeX_AMS:n4', 'KaTeX_Main:n4,i4', 'KaTeX_Math:n4,i4',
            'KaTeX_Size1:n4', 'KaTeX_Size2:n4', 'KaTeX_Size3:n4', 'KaTeX_Size4:n4'],
        testStrings: {
            'KaTeX_Size1': '()[]',
            'KaTeX_Size2': '()[]',
            'KaTeX_Size3': '()[]',
            'KaTeX_Size4': '()[]'
        }
    },
```

gagern:
@kevinb7: Oh, thanks! I've been using the WebFontLoader myself as well, but I've had my machinery tied in with the `active` callback, not the `fontactive`. And the `active` callback would apparently fire if at least one font was loaded successfully, giving up on the others after some timeout. So I got everything displayed, but it took longer than I had expected. I guess it would have taken me quite some time to find out that this is due to wrong test strings. With your configuration, the even fires much sooner now, which is great!

gagern:
With all the fonts we have at our disposal once #132 gets merged, web font loader is quite busy:

                families: [
                    "KaTeX_Main:n7,i4,n4",
                    "KaTeX_Math:i7,i4,n4",
                    "KaTeX_AMS:n4",
                    "KaTeX_Size1:n4",
                    "KaTeX_Size2:n4",
                    "KaTeX_Size3:n4",
                    "KaTeX_Size4:n4",
                    "KaTeX_Caligraphic:n4,n7",
                    "KaTeX_Fraktur:n4,n7",
                    "KaTeX_SansSerif:n4",
                    "KaTeX_Script:n4",
                    "KaTeX_Typewriter:n4",
                ]

This causes lots of fonts to get loaded unneccessarily, which delays page loads considerably. I think I'll have to come up with a way to load these fonts on demand. Either my `PreparedBox` objects will contain a list of all the fonts that have been used during typesetting, or the `options` object can contain some function which gets called whenever a font is required, and which can cause the rendering to abort if said font is not available yet. Something like this.

kevinb7:
I was just thinking that it might be good to use this as a fallback for browsers with rendering bugs.  See https://github.com/Khan/KaTeX/issues/45 for some examples.

gagern:
Might make sense in some scenarios, yes. And now that #264 got merged, I can look about tweaking that to process the same set of test cases using canvas rendering. Will need to employ [`executeAsyncScript`](http://seleniumhq.github.io/selenium/docs/api/javascript/module_selenium-webdriver_class_WebDriver.html#executeAsyncScript) so that the browser can call back to Selenium once all its fonts have been loaded and the rendering should be stable.

bryevdv:
ping @canavandl

kevinb7:
So I added @spicyj's fix for retina displays and it looks good on Firefox and Safari
screen shot 2015-07-25 at 3 26 20 pm
but Chrome is messed up
screen shot 2015-07-25 at 3 26 31 pm
I tried rendering "(" from KaTeX_Size2 and KaTeX_Size3 in a scale canvas and it just gets them wrong.  Not sure why Chrome doesn't like these fonts.  :disappointed:

spicyj:
Weird, I think it worked fine for me when I tried.

gagern:
How tightly do we want to integrate KaTeX with WebFontLoader? Since 5b182fc99b02143f390d4c0d2ba28c09aa859594 I'm exporting the set of used fonts, so the application can trigger required font loads on demand, and then repaint the canvas once the fonts are available. I have that logic implemented in my own application. Should we move some of this into KaTeX? And if so, what should the API look like?

kevinb7:
I think we want to keep KaTeX pretty lean so canvas related code and any use of WebFontLoader would probably end up in contrib/.

gagern:
My render-to-canvas is building on the katex-internal dom representation, so if you'd prefer having render-to-canvas outside the core code, then we'd need some API to expose that layer instead.

Would it be acceptable to have contrib code covered by travis-verified screenshots? In the long run I'd like to run the full set of screenshot tests through render-to-canvas as well, to ensure that noone forgets about that when modifying the dom structure, in particular when adding new CSS classes.

kevinb7:
It should be relatively easy to return the internal DOM representation.

gagern:
Returning the internal DOM is technically easy, yes. It might however be considered a promise to keep that API stable in the future, so we should be aware of that commitment. Or there should be some kind of big red warning sign stating that even though the representation is accessible from the outside, it is only intended to be used by code which evloves in lockstep with KaTeX.

On current master, merging in my canvas branch increases the gzipped size from 29,432 bytes to 32,009 bytes, so one might consider canvas support to cost approximately 2.5kB and increase the overall size by 9%. That is a noticable increase, so I can see how having this separate would be preferable.

kevinb7:
It would definitely need to include a warning that this is an unstable internal data structure, use at your own risk.

I can't see us include canvas support as part of the default build.  I think it makes sense as an optional module.

spicyj:
I can easily see us including support for canvas if it's faster or otherwise better than the default rendering. It may well be more efficient, though I'm worried about the perf of measureText. Would be nice to do some benchmarks.

The implementation here takes the DOM tree as input, which is too low-level. It shouldn't need to parse out class names to decide what to do. We should have an intermediate stage that's made up of hlists, vlists, boxes, glue, etc. which can then be rendered to either HTML (DOM) or Canvas.

kevinb7:
If we replace DOM rendering with canvas rendering then it makes sense to include in the build, but I think there's a lot of questions to answer before we get to that state.  The perf concerns of measureText could be addressed by including the horizontal glyph metrics.  I'm also concerned about the wonky rendering that I was seeing in Chrome (version 44).  It would be nice to have an intermediate layout rep.  We kind of already do with the groups, but parts of it (like the vlists) get specified at the same time that our internal DOM rep gets created.

kevinb7:
@spicyj what was the original reason for DOM rendering over Canvas rendering?

spicyj:
Seemed easier at the time, and IE8 support.

lukeburns:
@gagern, is it possible to render server-side using node-canvas?

gagern:
@lukeburns Should be possible. Not sure what steps would be required to make the relevant fonts available to Cairo, but installing them in a suitable directory should be enough I think. Server-side rendering in this fashion stirs some unpleasant memories of the server-side latex calls being used to render GIFs, though.

lukeburns:
@gagern Some immediate trouble with node-canvas: http://cdn.img42.com/322b7c639f5c3a539305106b9ced9dcc.png. The code: https://gist.github.com/lukeburns/9dab651d886c494ff694.

pnsaevik:
Rendering to canvas is great. But I'm curious to why no one has mentioned SVG, which seems to be the better choice over HTML5 canvas. SVG can contain <text> tags with absolutely positioned glyphs, which allows full control. Unlike canvas, SVG is scalable and searchable. And it's pretty cross-platform, at least within the modern browsers. Actually, there is a tool called dvisvgm, which can generate such SVG files directly from latex (server-side). See this [stackexchange question](http://tex.stackexchange.com/questions/282340/how-to-create-non-outlined-svg-files-from-latex-formulae) for more.

gagern:
@pnsaevik: When using just the formulas all by themselves, then SVG is probably better than canvas, and HTML probably better than both. So the main reason to choose canvas output is to integrate math rendering with an already existing canvas-based rendering pipeline. And there may be many reasons to use canvas. One might be performance, another a focus on raster-based image sources and operations. And there is also compatibility with other output devices to consider. So there are good resons for a web app to choose canvas, and those who do might benefit from integrating KaTeX like this. Those apps who already are using SVG would of course benefit from KaTeX exporting SVG, so it might make sense to have both in the long run.

canavandl:
hi @gagern 

So I was able to render LaTeX onto an already existing canvas element. It works as expected, except that modifying the `context.font` and have the change be reflected in the rendered text. Is that not implemented or am I missing something?

Demonstration of working `katex.renderToCanvas` method using a Bokeh plot:

(the lower text is the canvas branch, the upper is an overlaid div using KaTex)

screen shot 2016-09-02 at 8 38 52 pm


Thanks again for working on this.

gagern:
@canavandl: In our implementation we are [extracting the size from the context font](https://github.com/CindyJS/CindyJS/blob/61286fd6b480f47df466b9745a6ad4b379d510b0/plugins/katex/src/js/katex-plugin.js#L152-L153) and then [passing that](https://github.com/CindyJS/CindyJS/blob/61286fd6b480f47df466b9745a6ad4b379d510b0/plugins/katex/src/js/katex-plugin.js#L163) as an option to [the `canvasBox` invocation](https://github.com/CindyJS/CindyJS/blob/61286fd6b480f47df466b9745a6ad4b379d510b0/plugins/katex/src/js/katex-plugin.js#L181). As far as I can tell, the size is the only thing worth passing along, since KaTeX will use its own fonts and styles. Perhaps I should support the `textAlign` property one day, though. Having that set to anything but `'left'` will severely break rendering, I guess.

By the way: [this presentation](http://cindyjs.org/pub/2016-icms/plugins/#/3) (which is part of [a group of talks](http://cindyjs.org/pub/2016-icms/) on [CindyJS](http://cindyjs.org/)) demonstrates KaTeX to Canvas in an interactive environment, with a comparison to MathJax, the outcome of which you all can easily guess.

canavandl:
@gagern : your CindyJS example resolved my issue (I'd misread the regex and was passing '16px' instead of '16' as the fontSize).

My last question is about the text dimensions. Calling `canvasBox` returns a `PreparedBox` obj with some dimensions attached. The box.depth and box.height values are correct for "single level text", but not correct for quotients or similar. Is there a way to more correctly get the output dimensions (we'd like to support drawing background boxes). Again, this is a "nice to have" and not "must have".

Here's an example:

screen shot 2016-09-05 at 3 12 09 pm

gagern:
@canavandl: I fail to reproduce your problem. Running `make serve` and then opening [http://localhost:7936/canvas.html?size=32](http://localhost:7936/canvas.html?size=32) in my browser, I obtained the following picture:

canvas content

This is using a tweak I just added to this branch here to allow changing the font size; otherwise you'll be at the default size of `74 * 1.21 = 89.54` which is generally used for the development server. The cyan box is drawn using the dimensions from the prepared box, and it looks pretty much correct to me. Could it be that by the time you drew the background box, the fonts were not available yet? If that's not it, can you reproduce my cyan box as described above?

By the way, I just added some changes to ensure that merging my branch with the current master will not only be free of conflicts, but will actually work up to the point where `make serve` displays what it should.

marcianx:
Out of curiosity, what's blocking progress on this? Is it a lack of interest? Or prioritization?

I'd noticed, for example, that some diacritics weren't rendering properly with this renderer even though they are supported by the standard renderer.
I also noticed that this PR is based off of a time when the regular renderer didn't yet support \underline as I tried to use regular rendering off of this PR.

kevinbarabash:
@marcianx there are a couple of things that I'm concerned about with this and don't have good answers for at the moment:
- although there aren't a lot of changes, this does increase the size of the KaTeX bundle, it would be preferable for this to live in contrib (not a big change in and of itself)
- it relies on knowing about the internal structure returned by `buildHTML`... ideally we'd have a more display agnostic representation of the layout

gagern:
> Out of curiosity, what's blocking progress on this? Is it a lack of interest? Or prioritization?

Knowledge of the hackish nature of this approach, and lack of time for a proper solution.

> I'd noticed, for example, that some diacritics weren't rendering properly with this renderer even though they are supported by the standard renderer.

I expect there are several things which the main renderer can handle but this code can not. Essentially what this code does is read some HTML-like representation, then use the CSS classes from that and turn that into its own metrics information. Essentially a bit like a very lightweight CSS engine, but not based off the KaTeX css file but with parts of that CSS duplicated in JavaScript program logic.

So this will miss classes I forgot, or nesting constucts I forgot, or I got wrong. It will also miss classes added to the CSS since I started this branch. I haven't revisited that file in quite a while. And it will miss positioning expressed through means other than these classes, e.g. new ways of per-node positioning, or box borders, or combining characters. All of this is the reason why we consider this more of a hack and proof of concep implementation than a fully supported feature: with the current approach, keeping things in sync is rather hard. We'd prefer some intermediate representation which makes positioning explicit in a way that can be handled cleanly both by CSS and canvas.

> I also noticed that this PR is based off of a time when the regular renderer didn't yet support \underline as I tried to use regular rendering off of this PR.

You are right, underline isn't supported as it wasn't in core when I started this. It should be possible to add it, if someone (might include me) finds the time. In the meantime, I've resolved some conflicts so this branch merges with master again. That way, people can use master with all its current features to see which of these are supported by the canvas renderer.

edemaine:
Out of curiousity, have you thought about going straight from the parse to HTML canvas, a `renderCanvas.js` alongside `renderHTML.js` instead of this specialized HTML -> Canvas approach?  Is that what you mean by "a proper solution"?

spicyj:
Right now, buildHTML handles both metrics/styling decisions (for example, the supsub function that handles superscript/subscript vertical positioning) as well as how to encode those in HTML with CSS classes. I suppose the ideal is something akin to another intermediate representation (at least conceptually) which has sufficient positioning information and can be used to output to either HTML or Canvas.

gagern:
I agree. Writing a canvas-specific version of `buildHTML.js` would cause a lot of code duplication for things that essentially work out how a given high-level construct should get positioned at the lower level, before turning that low level representation into acutual dom nodes. A proper solution would identify the appropriate cut point between the code that would need to be duplicated and the code that is HTML-specific, introduce some clean interface there and allow different backends to get plugged in. I envision something like a box model representation, but making that sufficiently expressive to cover all our existing CSS use cases will require some work.

marcianx:
Firstly, thank you very much for rebasing this PR!
Secondly, is abstracting out the rendering backend from the general layout computation via an intermediate representation a priority? Is there a tracking bug? There seems to be a general desire for plugging in SVG and canvas backends.

rrandallcainc:
Hello, I'd be interested in contributing to this PR. I'm just curious...what is required to take this over the finish line and ready to be merged.

I'd be happy to assist with any work that's needed.

kevinbarabash:
@marcianx I've closed this PR for now but the rebased code lives on in the `canvas` branch.  #376 is tracking an intermediate representation.

rrandallcainc:
@gagern Thank you for this PR. It's really exciting to see Latex rendering being brought to the canvas!

With this PR, is it possible to be able to get the appropriate width/height and translate each individual character?

kevinbarabash:
@rrandallcainc the PR uses the Canvas API to measure the glyphs to get the width of each as it lays them out.  Another approach which is may be performant is to grab widths for glyphs beforehand.  Running `make extended_metrics` will generate a version of `src/fontMetricsData.js` that contains widths for all of the glyphs in the KaTeX fonts.  We'd still need to measure at runtime the width of any unicode characters we don't have metrics for.

rrandallcainc:
FWIW, I am working on modernizing this solution, and already have most of the features implemented. I reached out to @gagern but received no response :(. I'd like to open-source the solution in a separate package to allow others to benefit. I don't think any LaTeX Canvas solutions exist at the moment.

If anyone is interested in seeing progress, lmk. I can see what I can do.

kevinbarabash:
@rrandallcainc cool.  I'm looking forward to seeing what you come up with.

kevinbarabash:
I'm going to close this PR since whenever we get to supporting canvas it will require substantial rework.  It's linked to in a number of other issues.

joshProAssistant (commented):
cool

kortenkamp:
> FWIW, I am working on modernizing this solution, and already have most of the features implemented. I reached out to @gagern but received no response :(. I'd like to open-source the solution in a separate package to allow others to benefit. This would still leverage KaTeX's parsing/buildTree.
> 
> If anyone is interested in seeing progress, lmk. I can see what I can do.

@ry-randall – @gagern did this canvas rendering in order to KaTeX with CindyJS, meanwhile he left the project, and we are a bit lost, as we are using an ancient version of KaTeX now (see https://github.com/CindyJS/CindyJS/issues/829). Is there any way I could help? I could try to contact @gagern through other channels, but this only makes sense if this would help you…

gagern:
Sorry for lack of response to @ry-randall in https://github.com/KaTeX/KaTeX/pull/251#issuecomment-360329560. I find I've trouble keeping up with many comms channels. @kortenkamp I'll provide you with a phone number just in case. If there is any plan to get render-to-canvas support into KaTeX I'd be interested to contribute, but I only have limited time available besides day job.

Alternatives that I can think of:

1. Update my ancient branch to work with current KaTeX.

    I'd prefer to only do that if it actually gets merged in and maintained. Either into KaTeX proper, or into a separate fork which receives integrates from KaTeX on a regular basis. Probably should have some automated tests on the rendered images, similar to the `screenshotter` tests KaTeX uses for its HTML rendering. I believe I had `screenshotter` tests for canvas created at some point, but I don't see them in this PR here so perhaps not. I don't think I can promise maintaining such a fork on an ongoing basis. So unless someone can promise to take care of things going forward, we would be in the same spot again in a few years.

2. Invest in a proper design, refactor code.

    This sounds like a lot more work, but has the highest chances of getting accepted by KaTeX. Would require someone from KaTeX involved in discussions as to what kinds of design direction would be acceptable.

    Perhaps the first step would be to inspect [tree.js](https://github.com/KaTeX/KaTeX/blob/master/src/tree.js) and figure out a way how each `VirtualNode` could translate itself (and its children) into canvas draw instructions. I imagine the API would take something like the `CanvasState` from this PR here as argument, and produce something like the `CanvasRenderer.outList` list of atoms.

3. Make this a separate package.

    One problem is that the current renderer in this PR here is based off the intermediate `VirtualNode` representation from [tree.js](https://github.com/KaTeX/KaTeX/blob/master/src/tree.js), which is not part of the public API of KaTeX, and probably shouldn't be because it's not stable enough. Furthermore, you would run into issues by people trying to use incompatible versions. Trying to support more than one version of KaTeX would quickly turn into a nightmare.

    I think forking KaTeX and re-integrating regularly would stand a better chance, since that would give you access to all the internals, and would allow you to keep things well aligned in terms of versions, since you'r merge from upstream, address all issues then do your own release on a well-tested combination.

Random thought: It might make sense to also have SVG in mind when doing all of this. On the one hand, SVG should have all the primitives needed for rendering, so translating between the two should be really simple. On the other hand, that would allow using resulting SVG files as test cases. Processing those in automated tests should be a lot easier than processing screenshots, so one could build a really exhaustive collection of test cases and ensure they keep rendering to the same SVG files over time. Of course, the internal structure of those SVG might occasionally change with no real impact on the rendering, but if we avoid deeply nested grouping constructs in SVG, and instead render each primitive symbol on its own, then there shouldn't be a lot of variation there.

PR #258, opened 2015-06-20
Implement \mathbb

gagern:
Since blackboard letters (all uppercase and lower case k) are contained in the AMS font, this requires no additional work on the font support side. The mechanism has been designed to support `\mathcal` and `\mathscr` as well, but these need additional font metrics.

Furthermore, there appears to be no distinction between caligraphic and script letters in Unicode, which will affect the MathML rendering. At least on my system, “𝒜ℬ𝒞𝒟ℰℱ𝒢ℋℐ𝒥𝒦ℒℳ𝒩𝒪𝒫𝒬ℛ𝒮𝒯𝒰𝒱𝒲𝒳𝒴𝒵” looks like a strange mix between caligraphic and script letters, altghough they are all called “script” by Unicode. Some of them are from the BMP while others are not, which might lead to different fonts being used for them.

In https://github.com/Khan/KaTeX/commit/5e127ba6febe0d0850718c7142863a280c42fb1a I've included a little tool which I used to figure out what symbols are in what places in the KaTeX fonts. I'm not sure whether you'd want something like this included in the final merge. There are probably tools to achieve the same, I don't know. I was a bit surprised to find that the blackboard letters are in the normal alphabetic range, not associated with their Unicode codepoints. I wonder whether it would be possible to add a mapping for these Unicode codepoints in addition to the alphabetic range. KaTeX probably won't need that, though.

There is quite a bit of code duplication between `buildHTML` and `buildMathML`. Should I try to move that to `functions`, i.e. perform as much of the common functionality at parse time? Or should I stuff some of that functionality into a function in `buildCommon`, to which the two builders can delegate?

Is it acceptable to have unescaped unicode literals in the sources? I think that in this case, it makes the code a lot more readable (if your editor plays along), and the `ascii_only=true` argument to uglify will cause these literals to be escaped in the minified version (which I assume is what most people will be using).

For some reason, jshint isn't happy with my code. It complains about `Unexpected use of '&'` in the code which detects an UTF-16 surrogate pair. Do you have any idea of why this message is triggered? Should I experiment with alternate formulations, or just suppress the warning?

### DIFF
--- a/server.js
+++ b/server.js
@@ -65,6 +65,7 @@ app.use(express["static"](path.join(__dirname, "static")));
 app.use(express["static"](path.join(__dirname, "build")));
 app.use("/test", express["static"](path.join(__dirname, "test")));
 app.use("/contrib", express["static"](path.join(__dirname, "contrib")));
+app.use("/tools", express["static"](path.join(__dirname, "tools")));
 
 app.use(function(err, req, res, next) {
     console.error(err.stack);
--- a/src/buildCommon.js
+++ b/src/buildCommon.js
@@ -308,6 +308,28 @@ var spacingFunctions = {
     }
 };
 
+var specialFonts = {
+    "\\mathbb": {
+        font: "AMS-Regular",
+        support: "ABCDEFGHIJKLMNOPQRSTUVWXYZk",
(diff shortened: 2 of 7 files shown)


gagern:
Oh, I realize that I had missed #132. Will have a look atthat and compare.

gagern:
#132 is superior, let's continue there.

PR #261, opened 2015-06-20
Add unicode symbols to symbols table

### LINKED ISSUE #243: Symbol unicode replacement doesn’t work
[test case](http://jsbin.com/manevavibi/2/edit)

the trivial `katex.renderToString('σ')` throws `ParseError: KaTeX parse error: Unexpected character: 'σ' at position 0: ̲σ`

flying-sheep:
fixes #243 once it works, but it doesn’t:

```
  1) A subscript and superscript tree-builder should not fail when there is no nucleus
    Message:
      '^3' failed building with unknown error: Cannot read property 'charCodeAt' of undefined


    Stacktrace:
      Error: '^3' failed building with unknown error: Cannot read property 'charCodeAt' of undefined
    at Object.<anonymous> (/home/phil/Dev/Web/KaTeX/test/katex-spec.js:356:22)


  2) A subscript and superscript tree-builder should not fail when there is no nucleus
    Message:
      '^3_2' failed building with unknown error: Cannot read property 'charCodeAt' of undefined


    Stacktrace:
      Error: '^3_2' failed building with unknown error: Cannot read property 'charCodeAt' of undefined
    at Object.<anonymous> (/home/phil/Dev/Web/KaTeX/test/katex-spec.js:358:24)


  3) A TeX-compliant parser should fail when sup/subscripts are interspersed with arguments
    Message:
      Expected '\sqrt^23' to fail parsing, but it succeeded


    Stacktrace:
      Error: Expected '\sqrt^23' to fail parsing, but it succeeded
    at Object.<anonymous> (/home/phil/Dev/Web/KaTeX/test/katex-spec.js:1042:29)


  4) A TeX-compliant parser should fail when sup/subscripts are interspersed with arguments
    Message:
      Expected '\frac^234' to fail parsing, but it succeeded


    Stacktrace:
      Error: Expected '\frac^234' to fail parsing, but it succeeded
    at Object.<anonymous> (/home/phil/Dev/Web/KaTeX/test/katex-spec.js:1043:30)


  5) A TeX-compliant parser should fail when sup/subscripts are interspersed with arguments
    Message:
      Expected '\frac2^34' to fail parsing, but it succeeded


    Stacktrace:
      Error: Expected '\frac2^34' to fail parsing, but it succeeded
    at Object.<anonymous> (/home/phil/Dev/Web/KaTeX/test/katex-spec.js:1044:30)
```

anyone an idea?


gagern:
Can you wrap all that symbol-adding code in a function? That way, variables used inside the function won't spill into the module scope. (I'm no owner of this project, so this is just a suggestion from a fellow contributor.)

flying-sheep:
this is CommonJS, so only the `module.exports = ...` line defines what ends up in the module.

and if we pollute the module-building scope or the scope of the only function defined during module building doesn’t matter. (in the latter case we have all the loop vars, and in the former we have those and “symbols”.)

kevinb7:
I think it's fine the way it is.  There are already other module level variables we're defining like ```mathTextSymbols```, etc. and it's not like we're exporting these variables.

kevinb7:
@flying-sheep In terms of the tests you could probably just verify the parse tree of one against the other.  Have a look at some of the tests that use ```getParsed```.

gagern:
I just remember [this instance](https://github.com/Khan/KaTeX/pull/246#discussion_r32770177) where I accidentially used a module-level variable and jshint didn't catch that. But there are ferer functions in this file, mostly data, so I guess it's indeed fine. (Hopefully we'll be able to switch to ES6 one day, and make all these issues go away using `let` instead of `var`.)

gagern:
Before writing tests, we'll need lexer support first. Which made me look at the lexer, and wonder why we had a bunch of distinct regular expressions, instead of building a single one which matches any of the given ones. Any reason we shouldn't combine them?

flying-sheep:
yeah, so how to proceed? merge this and add lexer support in another PR?

gagern:
Does this PR here add anything, without lexer support? I'd merge both together, including tests. If I had the power to make that decision…

gagern:
`Parser.prototype.parseSymbol` will take whatever the lexer has to offer, then check that against `functions` and `symbols`. Which in my opinion means that the lexer can be generous in what it accepts. Whenever none of its input matches, it can simply take the next character, all by itself, and let the parser complain if it isn't a function or symbol it knows about. If that is indeed true, all those `mathNormals` and `textNormals` seem superfluous, duplication information already available elsewhere. Can someone please confirm that I'm not missing something here?

flying-sheep:
i rewrote the history and added lexer support and tests.

can anyone review this?

kevinb7:
I can review this later today.

gagern:
Looks good to me. From my past experiences here I'd assume the maintainers will want you to squash this into a single commit. If you want to change anything, I personally might avoid the `utils` import by writing `syntax.indexOf(input[pos]) !== -1`. But that's mostly a matter of taste (and imagined performance), no need to change the code just for that.

It could theoretically be the case that multiple TeX commands lead to the same Unicode symbol, but with different characteristics. Perhaps it would be good to have a test to verify that this is not the case. I.e. some test that does `require("../src/symbols")`, iterates over the symbols and checks that if the replacement matches, everything else matches as well. Should this go into one of the existing test files, or a new one? Should this go into the same merge?

spicyj:
I want to be careful here – ideally we match the behavior of `unicode-math`; I don't want to accidentally support more than that package does or have differing behavior.

kevinb7:
@spicyj It looks like support for unicode code points in ```unicode-math``` depends on the font being used with LaTeX.  http://mirror.its.dal.ca/ctan/macros/latex/contrib/unicode-math/unimath-symbols.pdf shows the symbols that are in each font when using unicode.  Since we don't support using different fonts would it make sense to restrict ourselves to those entries that are marked with either a (p), (l), or (a)?  These indicate support in pure TeX, LaTeX, and ```amssymb```.

kevinb7:
@flying-sheep I originally thought the parse tree would be enough, but it isn't until later in buildHTML and buildMathML that the command gets replaced with the symbol so that's where we to do the check.  Here's what checking the HTML would look like:
```
describe("Unicode symbol output", function () {
    it("should be match for unicode greek symbols and commands", function () {
        var builtCommand = getBuilt("\\sigma")[0];
        var builtUnicode = getBuilt("\u03c3")[0];
        expect(builtCommand.value).toBe(builtUnicode.value);
        expect(builtCommand.classes).toBe(builtUnicode.classes);
    });
})
```
Compare classes here should address @gagern's concerns about having the "characteristics" match.  We should also check the MathML.

Once we've decided which symbols we'd like to support we should probably have a test that verifies all of them.  There should also be a test that verifies that symbols we don't support don't work.

gagern:
> I want to be careful here – ideally we match the behavior of unicode-math; I don't want to accidentally support more than that package does or have differing behavior.

What's the motivation here? At least as long as the meaning is unambiguous, which I believe it is for the symbols contained in our fonts, I think the more support the better. They apparently try to support anything their fonts support, so shouldn't we do the same with our fonts, even if that is different from any font configuration in LaTeX?

spicyj:
Two reasons:

1. I don't trust us (mostly don't trust myself) to make good decisions here and if we choose to make our own interpretations for certain commands/symbols then it'll be difficult to change our minds later if we realize we messed up because people (e.g., KA) will have built up a large swath of expressions that depend on it.

2. I'd like to continue to be MathJax- and LaTeX-compatible to the extent we can so people don't learn non-standard KaTeX extensions and then get confused when they don't work in other environments. More concretely, KA uses KaTeX but falls back to MathJax for things that KaTeX doesn't support. This works well when KaTeX supports a subset of MathJax's functionality but doesn't work as well if they each support some things that the other doesn't. You might have an expression that uses a KaTeX-only command and then want to add a `\begin{align*}` (or whatever) to it which KaTeX doesn't support, but now neither KaTeX nor MathJax can render the combined expression. Obviously this becomes less of a concern as we make our way through the remaining missing features, like the valuable matrix support you recently added.

Let me know if these make sense. Basically, I just like to err on the side of safety, which in this case means adding fewer new things to support.

gagern:
@spicyj: I don't share your doubts about 1. But 2. makes sense for the specific KA application you describe. In that case, I think MathJax should be the reference to compare against, not the unicode-math package.

flying-sheep:
hmm, @spicyj is right in that there are nontrivial mappings!

you can test with this code:

```
var reverse = {}

// We let unicode symbols represent themselves
var syntax = "^_";
for (var mode in symbols) {
    for (var symbol in symbols[mode]) {
        var spec = symbols[mode][symbol];
        if (spec.replace && !utils.contains(syntax, spec.replace)) {
            if (!reverse[spec.replace])
                reverse[spec.replace] = [];
            reverse[spec.replace].push({ s: symbol, f: spec.font, g: spec.group });
            
            symbols[mode][spec.replace] = {
                font: spec.font,
                group: spec.group
            };
        }
    }
}

for (var rep in reverse) {
    var equal = reverse[rep].every(function(s) {
        return s.f == reverse[rep][0].f && s.g == reverse[rep][0].g;
    });
    
    if (reverse[rep].length > 1 && !equal) {
        var stringified = reverse[rep].map(function(s) {
            return '“' + s.s + '” (' + s.f + ';' + s.g + ')';
        });
        console.log('' + rep + ' →\t' + stringified.join(', '));
    }
}
```

which will print:

>∼ →	“\sim” (main;rel), “\thicksim” (ams;rel)
>⊥ →	“\perp” (main;rel), “\bot” (main;textord)
>∣ →	“\mid” (main;rel), “\shortmid” (ams;rel), “\lvert” (main;open), “\rvert” (main;close), “|” (main;textord), “\vert” (main;textord)
>∥ →	“\parallel” (main;rel), “\shortparallel” (ams;rel), “\|” (main;textord), “\Vert” (main;textord)
>⌣ →	“\smile” (main;rel), “\smallsmile” (ams;rel)
>⌢ →	“\frown” (main;rel), “\smallfrown” (ams;rel)
>∝ →	“\propto” (main;rel), “\varpropto” (ams;rel)
>⋅ →	“\cdotp” (main;punct), “\centerdot” (ams;bin), “\cdot” (main;bin)
>ℏ →	“\hbar” (main;textord), “\hslash” (ams;textord)
>† →	“\dag” (main;textord), “\dagger” (main;bin)
>‡ →	“\ddag” (main;textord), “\ddagger” (main;bin)
>⊴ →	“\unlhd” (ams;bin), “\trianglelefteq” (ams;rel)
>⊵ →	“\unrhd” (ams;bin), “\trianglerighteq” (ams;rel)
>△ →	“\vartriangle” (ams;rel), “\triangle” (main;textord), “\bigtriangleup” (main;bin)
>▽ →	“\triangledown” (ams;textord), “\bigtriangledown” (main;bin)
>⊲ →	“\vartriangleleft” (ams;rel), “\lhd” (ams;bin)
>⊨ →	“\vDash” (ams;rel), “\models” (main;rel)
>≈ →	“\thickapprox” (ams;rel), “\approx” (main;rel)
>⊳ →	“\vartriangleright” (ams;rel), “\rhd” (ams;bin)
>∖ →	“\smallsetminus” (ams;bin), “\setminus” (main;bin)
>∅ →	“\emptyset” (main;textord), “\varnothing” (ams;textord)

gagern:
Interesting! Taking one example, namely

> ⊥ → “\perp” (main;rel), “\bot” (main;textord)

I find that MathJax will interpret ⊥ on input as `\perp`, while `unicode-math` with XeLaTeX interprets it as `\bot`. This establishes that there is no clear standard to follow, and that compatibility with `unicode-math` doesn't ensure compatibility with MathJax. Do we want to exclude all the ambiguous symbols? Do we want to prepare a document with all the remaining ones, and verify that XeLaTeX with `unicode-math` and MathJax both render the unicode version visually equivalent to the macro version? Do we want to verify that the generated group types are the same as well? If so, how?

flying-sheep:
well that’s an error in the mapping. we map both to the first one:

* `⊥` is U+22A5 “UP TACK” (Mathematical Operators)
* `⟂` is U+27C2 “PERPENDICULAR” (Miscellaneous Mathematical Symbols-A)

we should map `\perp` to the latter, than this one will go away. maybe there are more problems like this.

flying-sheep:
also interesting is `|` and `∣`. the former is the normal bar on everyone’s keyboard and the latter is the “divides” mathematical operator.

gagern:
> that’s an error in the mapping.

Then it's an error in the font. The KaTeX_Main font contains a glyph called “perpendicular” and maps unicode codepoint U+22A5 to that. We can only emit HTML for symbols which we know our fonts support, otherwise the output would become arbitrary. The glyph associated with U+2223 “DIVIDES” is apparently unnamed in the font, as far as I can tell from the auto-generated name TTX produces for it.

flying-sheep:
> Then it's an error in the font.

then it is: [U+27C2 prependicular](http://www.fileformat.info/info/unicode/char/27c2/index.htm) ≠ [U+22A5 up tack](http://www.fileformat.info/info/unicode/char/22a5/index.htm)

flying-sheep:
U | normal | alt | rel | alt rel. | open | close | punctuation | bin | alt. bin
---|---|---|---|---|---|---|---|---|---
∼ | `\sim` | `\thicksim`
⊥ | `\bot` | | `\perp`
∣ | `|`, `\vert` | | `\mid` | `\shortmid` | `\lvert` | `\rvert` 
∥ | `||`, `\Vert` | | `\parallel` | `\shortparallel` | 
⌣X | `\smile` | `\smallsmile` 
⌢X | `\frown` | `\smallfrown`
∝ | `\propto` | `\varpropto`
⋅ |  `\cdot` | `\centerdot` | | | | | `\cdotp`
ℏ | `\hbar` | `\hslash`
† | `\dag` | | | | | | | `\dagger`
‡ | `\ddag` | | | | | | | `\ddagger`
⊴ | | | `\trianglelefteq` | | | | | `\unlhd`
⊵ | | | `\trianglerighteq` | | | | | `\unrhd`
△ | `\triangle` | | `\vartriangle` | | | | | `\bigtriangleup`
▽ | `\triangledown`  | | | | | | | `\bigtriangledown`
⊲ | | | `\vartriangleleft` | | | | | `\lhd`
⊨ | | | `\models` | `\vDash`
≈ | | | `\approx` | `\thickapprox`
⊳ | | | `\vartriangleright` | | | | | `\rhd`
∖ | | | | | | | | `\setminus` | `\smallsetminus`
∅ | `\emptyset` | `\varnothing`

spicyj:
I hope this shows that there's at least some variation here and careful thought is required. For example, `|` should definitely be spaced as `\vert` is. (Did you notice that I asked that tex.stackexchange question? :))

When MathJax and LaTeX differ slightly, we tend to follow LaTeX's behavior. (Our long-term plan is not to follow MathJax forever…)

kevinb7:
@flying-sheep excellent detective work.  I love the table you've created.  With regards to the UP TACK vs PERPENDICULAR debate, I had a look at the font's repetoire in the Font Book utility on Mac OS X and it reports U+22A5 as UP TACK, but it doesn't show PERPENDICULAR.  To get to the bottom of this I'd probably use freetype to look up the glyph for U+22A5 and U+27C2.  I'm pretty sure they map to the same glyph.  I'll give this a try tonight.

In terms of mappings I think U+27C2 should map to ```\perp``` while U+22A5 should map to ```\bot```.
Reference: https://en.wikipedia.org/wiki/Up_tack.

@spicyj which tex.stackexchange question are you referring to?

kevinb7:
So I was wrong.  I thought the font would map both code points to the same glyph.  Although it is possible for fonts to do so, this particular font does not.  It only contains one mapping and that mapping is for U+22A5.

Here's a gist of how I determined this: https://gist.github.com/kevinb7/fa5e9ed342c928a0f31f.

flying-sheep:
[this question](http://tex.stackexchange.com/a/506/2867) which i linked above. and indeed i didn’t notice @spicyj asked it :laughing:

i think a good approach would be special casing all ambiguous mappings, and then implementing a check that makes tests fail if there are unhandled ambiguous cases left.

spicyj:
That doesn't catch cases where we currently have only one of the symbols but might add another with the same output symbol in the future.

It would also mean that you can only type UP TACK and not PERPENDICULAR to get `\perp`, where it should probably be the other way around. The character used for the output font doesn't necessarily correspond to something that makes sense semantically as input.

flying-sheep:
> That doesn't catch cases where we currently have only one of the symbols but might add another with the same output symbol in the future.

of course it would. i specifically addressed this with this sentence:

> […] implementing a check that makes tests fail if there are unhandled ambiguous cases left.

spicyj:
I meant, we might realize later that we used the wrong symbol now. I'd really rather avoid adding symbols and then changing them later.

flying-sheep:
so what do you propose to fix “i can’t type ‘σ’ without KaTeX breaking on me” NOW?

and what better strategy do you have to vet all thousands of symbols?

spicyj:
If there's no good solution NOW, then we won't fix the problem NOW. Sorry.

I'd be happy to add all greek letters. I'm sure there are other important things but we don't need to add all of the thousands today.

If I knew a magic solution I'd tell you. I don't.

flying-sheep:
i use greek, arithmetic symbols (·×÷≠≤≥±∓), logical symbols (∧∨¬), proof stuff (∃∀∎), ∑, ∫, set theory (∈∉∅), numeric classes (ℝℕℤ), and more stuff i can’t remember right now but i will remember the compose key sequence one i need them.

this is why i want a comprehensive solution: i copy that stuff from ConTeXt documents into Reveal.js presentations and want them to work, not to have to search for a LaTeX macro replacing the symbol.

---

so what’s the process you created the mapping in symbols.js with? how can we fix e.g. \perp not pointing at the UP TACK codepoint but the right symbol and the font still displaying a glyph it has.

kevinb7:
Why don't we just avoid the ones that have duplicates for now?  That still covers most of them.  As for the number classes we don't support ```\mathbb``` yet, but when we do (hopefully soon), we can add those as well.  

I agree we should eventually add the U+27C2 code point to the fonts (it should be included in all KaTeX_Main variants), but that feels like a separate pull request.  I'm not sure how the fonts were created in the first place.  We could use FontForge to edit the .ttf files and then convert those files to .woff, .woff2, .eot.

gagern:
How about: make this an optional feature, and make the option enabling it a version number. So you'd e.g. pass `{unicodeInput:1}` as an option to the top level KaTeX interfaces, and enable the mapping as described above: unambiguous things get mapped, ambiguous don't. If at some point we find that we are behaving different than *‹your favorite Unicode-aware math typesetting configuration here›* then we would encourage users to update the version number, and implement the new behavior for that setup while maintaining backwards compatibility for the old. And if you ever feel confident enough to enable this by default, the you can do so without being encumbered by past choices.

You wouldn't even need the version-handling code just now; just having the enabled-or-disabled switch is enough. And if everything goes well, then you only ever add symbols which either are not in conflict to existing ones, or where the ambiguity should definitely be resolved in favor of the existing ones. In that case, you'll never actually have to handle these version numbers, they were just a safety precaution.

flying-sheep:
> Why don't we just avoid the ones that have duplicates for now?

i’d like to see that as well, but @spicyj said he doesn’t want to have mappings change over time. e.g. if we have, say `\foobar` mapped to U+3333, and then we find out U+3333 is actually wrong and we change the mapping to `\foobar` → U+4318, then suddenly U+3333 will no longer map to anything.

that’s what @spicyj wants to avoid. (right?) so it has nothing to do with duplicates.

gagern:
> that’s what @spicyj wants to avoid. (right?)

I don't think that was his point. The problem you describe could easily be addressed by manually adding a mapping for U+3333 as input. We could even add tests to ensure that whenever a symbol gets added, that fact is recoded in some list, and whenever it gets removed again, something will complain loudly.

The way I understood it, the problem is more like this: Suppose we now have a mapping for `\foobar`to U+3333 and later on we find that there is another symbol `\qux` which a) looks like U+3333 as well, but b) matches the semantic description of U+3333 even better and c) some TeX setup renders U+3333 on input like `\qux`, not like `\foobar` in terms of group type and spacing. In that case, it would have been correct for us to map U+3333 to `\qux` from the beginning. Changing it later on breaks backwards compatibility, while not changing it breaks compatibility with other implementations. Both of this should be avoided, ergo his conclusion that we should not map U+3333 to anything unless we can establish that doing so is right in terms of compatibility with something authoritative.

kevinb7:
Providing unicode support via an option sounds like a good way to roll this out.

> We could even add tests to ensure that whenever a symbol gets added, that fact is recoded in some list, and whenever it gets removed again, something will complain loudly.

It's always good when tests complain if something gets removed.

spicyj:
> The way I understood it, the problem is more like this: Suppose we now have a mapping for `\foobar`to U+3333 and later on we find that there is another symbol `\qux` which a) looks like U+3333 as well, but b) matches the semantic description of U+3333 even better and c) some TeX setup renders U+3333 on input like `\qux`, not like `\foobar` in terms of group type and spacing. In that case, it would have been correct for us to map U+3333 to `\qux` from the beginning. Changing it later on breaks backwards compatibility, while not changing it breaks compatibility with other implementations. Both of this should be avoided, ergo his conclusion that we should not map U+3333 to anything unless we can establish that doing so is right in terms of compatibility with something authoritative.

That's exactly my stance, thanks.

I'm not against having an option that enables experimental unicode support that's likely to change in future versions, but if we go down that route I'd like to also have some path for getting a stable subset into the core features too, or else for clients who need stability this feature won't give any benefit.

gagern:
> if we go down that route I'd like to also have some path for getting a stable subset into the core features too

How? Quoting [myself](#issuecomment-114401008):

> Do we want to prepare a document with all the remaining ones, and verify that XeLaTeX with `unicode-math` and MathJax both render the unicode version visually equivalent to the macro version? Do we want to verify that the generated group types are the same as well? If so, how?

That's still the best idea I have to base our decisions on some reference implementation, and I'd like a second opinion on this.

kevinb7:
@flying-sheep noticed this when running ```make metrics``` today.  
```
The glyph named Delta is mapped to U+0394.
But its name indicates it should be mapped to U+2206.
The glyph named Omega is mapped to U+03A9.
But its name indicates it should be mapped to U+2126.
```
We should probably handle those mappings manually.  I think we probably want both U+0394 and U+2206 to map ```\Delta``` and both U+03A9 and U+2126 to map to ```\Omega```.

gagern:
@kevinb7: I don't follow you. The one glyph is named “Delta” and it maps to [U+0394  Greek capital letter Delta](http://www.fileformat.info/info/unicode/char/0394/index.htm). Why should it be mapped to [U+2206 Increment](http://www.fileformat.info/info/unicode/char/2206/index.htm)? The same holds for Omega. I consider the greek letter meaning the canonical in both cases, while the other use is a more derived form.

On input, U+0394 should be treated like a variable name, while I'd say U+2206 should be treated the same as `\partial`, since it has a similar meaning. Luckly, the two are both `textord`, so treating them the same is justified. [U+2126 Ohm sign](http://www.fileformat.info/info/unicode/char/2126/index.htm) should be treated like a unit, so perhaps we should have a look at how packages like `siunitx` space such things?

kevinb7:
Sorry I misread the message.  It's doing the right thing.  U+2126 Ohm sign probably needs a gap between it and a preceding number.  Looking at ```siunitx``` sounds like a good idea.

kevinb7:
Any work to implement parts of ```siunitx``` should be under a new pull request.

kevinb7:
> I'm not against having an option that enables experimental unicode support that's likely to change in future versions, but if we go down that route I'd like to also have some path for getting a stable subset into the core features too, or else for clients who need stability this feature won't give any benefit.

Would this be an acceptable path:
 1. put this behind a flag and merge it
 2. move uncontested symbols from behind the flag
 3. where there are sensible mappings from multiple code points to multiple commands, make those mappings and move them from behind the flag, e.g.
   - U+27C2 => \perp and U+22A5 => \bot
   - U+03B5 => \varepsilon and U+03F5 => \epsilon (https://en.wikipedia.org/wiki/Epsilon)
 4. do some more research into the implementation of `unicode-math`
   - http://mirror.its.dal.ca/ctan/macros/latex/contrib/unicode-math/unicode-math.pdf
 5. apply new learnings to move some or all of the remaining symbols from behind the flag

Question: is there a difference between "ord" and "ord alt" or "rel" and "rel alt"?

gagern:
Do you want to merge this in its current form, or is there any chance of getting #132 dealt with soon, so that we can follow [the roadmap](https://trello.com/b/P5DWbSuR/katex-roadmap) and eventually reach [the symbols part](https://github.com/gagern/KaTeX/commit/c5d7bf85f9d015d4d695031c6debf53d50277e06#diff-58a03b51243f0041858a5394b0a40090) from #262 and the lexer change from 8496a1e40f8dfbfc8dd7d2e7b5096bd5c2b01c02 scheduled for #266? Because those commits would mean a complete rewrite of the code here, and make many things implemented here superfluous. Furthermore, they'd suggest a different data representation for the unicode mapping data tables we'll be needing.

The way things look at the moment, I fear all of this is pretty far into the future. In that case I'm all for merging this here as it stands, and adapting those other things when I rebase them.

> move uncontested symbols from behind the flag

How do we come up with such a list of uncontested symbols?

kevinb7:
There's that much code here so a rewrite wouldn't be the worst thing, but it'd be nice to get this merged sooner rather than later.  I recently discovered that some of the content creators at KA are using unicode in their content so it would be nice not to have to fallback to MathJax just because there's a unicode theta in the math code.  I'd like to hear @spicyj's thoughts on the path I proposed first.

> Furthermore, they'd suggest a different data representation for the unicode mapping data tables we'll be needing.

Couldn't we iterate through `module.exports` in the new version of symbols?  Why would the data representation need to change?

spicyj:
How do we come up with a list of uncontested symbols?

kevinb7:
> How do we come up with a list of uncontested symbols?

I was thinking that would just be everything that wasn't in the table that @flying-sheep posted.

gagern:
> Couldn't we iterate through `module.exports` in the new version of symbols? Why would the data representation need to change?

Firstly, instead of iterating over all the symbols, I'd integrate this with the [`defineSymbol`](https://github.com/gagern/KaTeX/blob/b8c0343378a453a0aa9850878064d2eb95b0e151/src/symbols.js#L23-L29) function I'd introduce, since any symbol definition passes through that in any case and I'd rather avoid code outside of functions due to scoping concerns.

Secondly, we'd need extra data: which symbols are automatically enabled, which ones only if a flag is given? Which ones change meaning depending on whether that flag is given, or perhaps even depending on the value of the flag? That kind of data would in the current structure most likely be encoded as some big JSON dict, to stay in style with the existing data base. But if that data base gets turned into a bunch of function invocations, then having the Unicode information in such a format would be more consistent.

> t'd be nice to get this merged sooner rather than later.

Then by all means do so.

> How do we come up with a list of uncontested symbols?

Perhaps [this tool of mine](https://github.com/gagern/KaTeX/commit/ac1117fb0b0acea059cecf78fa3bb6ecd5686d50) from #294 can be used to verify chosen atom types against some LaTeX package like `unicode-math`. And we could of course generate test cases to be fed to the `texcmp` toolchain we have since #268, to perform visual comparisons.

kevinb7:
@flying-sheep can you hide your changes behind a flag?

kevinb7:
> Firstly, instead of iterating over all the symbols, I'd integrate this with the defineSymbol function

This makes sense instead of doing a second pass.

> Secondly, we'd need extra data: which symbols are automatically enabled, which ones only if a flag is given?

At least at the start of the proposed path forward, it should be pretty simple.  I think it's going to take some time to work our through 1–5.  Merging in 1 on its own should have too much of an effect on the pending refactorings.

qbolec:
Perhaps my comment is obvious and does not move things forward, but my use case is that I want to migrate from MathJax to KaTeX, as your library is much faster. I already have a massive amount of content which was prepared for MathJax. Now, every discrepacy between MathJax and KaTeX means that I have to choose between:
a) running some find&replace through all the content 
b) not migrating to KaTeX and continue to use MathJax
c) do what KhanAcademy allagedly does: try{KaTeX}catch{MathJax}
d) ???
How about having a switch (or dedicated build) for MathJax-compatible behaviour?
For example lodash library has a mode compatible with underscore.
This made transition from one lib to another smooth.
This buys the developer some time, to adjust the app and content for finally switching to non-compatible mode.
I realize that this path sounds like it leads to never-ending story of supporting compatibility with another library, but:
a) you seem to do that anyway
b) that seems to be what people want (in order to migrate)
c) at least the migration/compatibility code can be clearly separated/marked/documented, leaving the core clean and reasonable
d) perhaps some volunteers would like to contribute to maintenance of this mathjax-compatible version - once there is no volunteers you have a clear signal that this support is no longer needed

I'm not a latex or unicode expert, but I suspect that there can be other discrepancies between libraries than handling of unicode - perhaps MathJax syntax differs slightly from MikTeX or whatever - the idea of having a configuration option called `unicodeInput:1` for me is semantically different than saying `MathJaxCompatibile:true` in that first assumes that the problem is in the encoding and that there is a single good solution (or even: a progression of better and better solutions which can be ordered and numbered by versions) while I think that there could be just many different dialects.
Another example would be various libraries out there for Markdown, some of which give you an option to decide which dialect you want to use, or even define your own.
Perhaps an option like `dialect:"MathJax"` or even `dialect:{encoding:{"σ":'\\sigma"}}` is what we need?

flying-sheep:
OK so what of https://github.com/Khan/KaTeX/pull/261#issuecomment-121511602 is done?

do you all agree that it’s enought to have a special case override using this table https://github.com/Khan/KaTeX/pull/261#issuecomment-114415474 and an unit test checking that no new ambiguous reverse mappings exist that aren’t handled in the override table?

kevinbarabash:
@flying-sheep that sounds good to me.

flying-sheep:
@spicyj might be opposed to it. i think he said before in this thread that wrong mappings could slip through and would break backwards compat if changed. (right?)

but i think it’s the most correct approach that isn’t extremely incomplete or extremely much work. adding reverse mappings manually while checking every single mapping for correctness is invariantly one of both. or a compromise that is both incomplete and much work.

kevinbarabash:
@flying-sheep in re-reading parts of the discussion it probably makes sense to see which unicode symbols MathJax supports.

gagern:
@flying-sheep: While I agree that checking every single mapping *manually* would be extremely much work, I believe that it should be possible to automatically verify a large portion of the candidate mappings. This holds for compatibility with MathJax as well as unicode-math. If we get that set up, we could even identify the portion on which both agree, and enable those without any flag or switch or whatever.

If I find the time, I'll try to come up with something to check things against xelatex and unicode-math within a week. If I don't report back by then, I didn't find the time. If someone else wants to tackle MathJax, that would be great. The things to verify would be essentially that entering the Unicode symbol and the command sequence yield the same output, including spacing. See page 170 of [The TeX Book](http://www.ctex.org/documents/shredder/src/texbook.pdf) for how atom types affect spacing. Ideally I'd say the symbol in question should be both preceded and followed by one symbol from each of these types, and all the spaces should be compared.

qbolec:
Not sure if this helps anyway, but here are the places where the sgima unicode character encoded as `\u03C3` occurs in the MathJax source: https://github.com/mathjax/MathJax/search?utf8=%E2%9C%93&q=%5Cu03C3 
These seem to be places concerned with outputing UTF-8, not reading it, but still provide some mappings between commands and UTF-8 characters.
This one seems to be for input: 
https://github.com/mathjax/MathJax/blob/ba9afeb5a743249acdea17540b48b14ebc95dbe4/unpacked/jax/input/TeX/jax.js#L471
Not sure what the code does, but I guess they use unicode as the internal representation and try to convert commands to unicode - not the other way around.
And the search for `sigma` gives a lot of results, many of them seem to be some font metrics:
https://github.com/mathjax/MathJax/search?utf8=%E2%9C%93&q=sigma

gagern:
It could well be that the MathJax is simply forwarding sigma and other things as letter-like symbols to the font, without containing any explicit list of the things it forwards. Perhaps some code point ranges.

ronkok:
I realize that this discussion is stale, but I have an question.

Consider two functions: `\land` and `\wedge`. These two functions are *synonyms*, and by that I mean:
1. their MathML output is identical, i.e., `<mo>∧</mo>`
2. their HTML output is identical, i.e., they are both math symbols of the same atom type (`BIN`), displaying the same character (∧), from the same font (KaTeX Main).
3. their sameness goes deeper than the final output. They are also identical while resident in the DOM tree and in the MathML tree.

When we consider the Unicode character ∧ and map it to a TeX function, it doesn't matter which synonym we choose. To the author or reader, either choice has no percievable difference in input, output, or tree. If we choose one now and change the choice later, nothing breaks.

Can we agree on that?

Also, let's set a criterion such that we prefer to map to a simple atom as compared to a function with an argument. For example, → should map to `\rightarrow` or `\to`, rather than, say, `\vec`. `\overrightarrow`, or `\xrightarrow`.

Can we agree on that, too?

If we are agreed, then I can begin to start writing PR's in the style of PR #410. My intent is to write some scripts that will gather mapping data from KaTeX [symbols.js](https://github.com/Khan/KaTeX/blob/master/src/symbols.js), from Carol JVF Burns's [MathJax page](http://www.onemathematicalcat.org/MathJaxDocumentation/TeXSyntax.htm), from [unicode-math](http://ctan.sharelatex.com/tex-archive/macros/latex/contrib/unicode-math/unicode-math-table.tex), and from John Cook's [Unicode/LaTeX Converter](https://www.johndcook.com/unicode_latex.html). I may also try to get background data from [Detexify](http://detexify.kirelabs.org/classify.html).

If all those sources agree on a mapping for one character, or at least agree on synonyms, then I will feel free to include that character in a PR. I'll save the hard cases for later discussion.

Acceptable?

edemaine:
@ronkok followed through on this plan, and we now have much more nuanced/careful support for most Unicode symbols.  Closing this PR which represents a simpler approach, and the epic discussion that resulted.  If you have more symbols to add, I suggest opening new issues or PRs, in a style similar to @ronkok's recent several PRs.

ronkok:
The epic discussion was very helpful. I wrote my Unicode PRs in a way that tried to avoid the potential pitfalls pointed out here.  So thanks!

Issue #256, opened 2015-06-18
Break up array of function declarations


gagern:
In `functions.js`, the functions are stored in two big arrays, namely `functions` and `duplicateFunctions`. This makes it hard to mix code which is not a LaTeX function in with the rest. To prove my point, see how the `delimiters` constant array is located just before `duplicateFunctions`, while in fact it belongs to the implementation of the delimiter functions which is quite a bit down from that. This makes reading those functions harder.

I'd prefer a way where each function definition occurs at the top level, so that there is no motivation to separate things location-wise which belong together conceptually. Ideally, we'd design that in such a way that the distinction between single-name and multi-name functions becomes less pronounced, so that if a function grows a second name, it doesn't have to get moved around, thus making the diff smaller and easier to read.

How about something like this:

```js
declareFunction(["\\name1", "\\name2"], {
    numArgs: 2
}, function(…) {
    …
});
```

I guess that if there is only a single name, the first argument could be simplified from an array to a single string. And if the number of arguments is the only option, the second argument could be simplified to a single number. The `declareFunction` could take care of all the defaults currently handled in the loop at the end of the file, which has the added benefit of restricting any variables it uses to the scope of that function. And finally, having the code indented like this means the body of the handler has only four spaces of indentation, not sixteen as is currently the case for `duplicateFunctions`. Similar changes could be made for environments.

*This is a spin-off from https://github.com/Khan/KaTeX/pull/246#issuecomment-113141856, one of two.*

kevinb7:
I think this makes a lot of sense.  We'd probably want to do this for symbols as well.  I'd like to hear @xymostech's and @spicyj's thoughts on this.

gagern:
Come to think of it, `groupTypes` in `buildHTML.js` is suffering from a mild case of the same: one big array, where you can't simply insert auxiliary functions. There my preferred notation would be changing the `genfrac: function(…) {` to `groupTypes.genfrac = function(…) {`, thus reducing the nesting depth by one level and allowing for other code to get mixed with the group rendering code.

kevinb7:
Why use a different pattern in this case?

gagern:
Mainly to get rid of quotes. `funcs["\\frac"] = { … }` would not work well with several functions sharing one implementation (i.e. `duplicateFunctions`), and would still have to contain quotation marks due to the name which is not a valid identifier. However,

```js
groupType("genfrac", function(…) { … });
```

looks more verbose and obscure than

```js
[...]

Issue #255, opened 2015-06-18
Function argument restructuring


gagern:
Currently the arguments to a function handler are first the name of the function, then the arguments of the function as separate arguments, then the positions of these arguments, with `this` pointing at the parser. All of this feels pretty mixed-up. Things become even worse for the newly merged environments, where we need two more arguments, `pos` and `mode`.

I can think two possible solutions. Either make the number of arguments fixed. I.e. always pass `parser, func, args, positions` in a fixed order, with `args` being an array of arguments, and `this` unset (i.e. equal to `window` in non-strict mode). Or let the function arguments in TeX and JS agree completely, wrapping everything else in `this`. So you'd have a custom-made `this` object (some small JSON constructed for the purpose), with `this.parser`, `this.lexer`, `this.name` and so on.

For the positions, an alternative to the `positions` array might be blurring the distinction between `ParseNode` and `ParseResult`, so that the object which holds the type also holds the position. That would mean keeping position information throughout the tree, but I believe that it would greatly simplify things in a number of places.

*This is a spin-off from https://github.com/Khan/KaTeX/pull/246#issuecomment-113141856, one of two.*

kevinb7:
I feel like ```ParseNode``` and ```ParseResult``` should probably be merged into one.  The main goal of ```ParseResult``` is to add position information to each node.  Libraries like esprima and acorn which parse JavaScript into an AST add location metadata to each node as code is being parsed.  It seems to work well.  Having the start and end locations of each node might open up some interesting use cases.  It would have the more immediate effect of removing the need for positions.  It also removes an extra layter of indirection in many places.

I like the idea of having ```parser```, ```lexer```, ```name```, and ```mode``` on ```this``` and passing in the args from the latex function as is... with optional arguments coming at the end?

Thoughts @xymostech @spicyj ?

xymostech:
> make the number of arguments fixed. I.e. always `pass parser, func, args, positions` in a fixed order, with `args` being an array of arguments, and `this` unset

I prefer this much more than the custom `this` value. Abusing `.apply` just to tack on extra values seems really strange. Maybe we could instead pass in an `options` parameter as the first argument that contains `parser`, `lexer`, `name`, `mode`, `positions`, etc?

gagern:
[...]

Issue #250, opened 2015-06-18
add support for macros via \def and \newcommand


kevinbarabash:
In the process of helping @gagern with https://github.com/Khan/KaTeX/pull/246 I came to the realize that the environments are essential functions that do some extra parsing and do that parsing in a slightly different way.  In the case of the array/matrix environments they recognize `&` as a special token to separate cells.

Macros are similar to environments except instead of simply parsing the incoming tokens differently, they also generate a new input stream from their definition by inserting the arguments they parsed into the macro definition.  Then they parse this new input stream in place of the command and its arguments.

Why not create a new Parser/Lexer instance whenever we run into a macro, expand the macro and parse the new input stream on it's own.  The parse tree from that expansion can then be returned to the parent Parser instance.  Neither the parser or the lexer has much state so it shouldn't consume too much memory to new up additional Parsers/Lexers when needed.

cc: @xymostech @spicyj

kevinb7:
I think https://github.com/Khan/KaTeX/issues/37 could be handled in a similar fashion.

I just noticed my note about ```\def\oneover{\frac{1}}```.  To handle this situation, if temporary Parser is expecting more tokens when it comes to the end of the expanded input stream it switches over to its parent's lexer.

spicyj:
I think it would might be a good idea to first figure out a solid API for defining new commands in JS. Something like

```
defineCommand("\\limsup", function() {
  return hlist(text("lim"), spaces.thinspace, text("sup"));
});
```

and then we can implement `\newcommand` on top of that. I think most people using KaTeX (such as KA) won't want to support def/newcommand for user-defined expressions but people might want to define some site-specific commands.

(Incidentally, we don't have limsup or liminf because I want something like this first.)

gagern:
Site-specific commads are on my agenda as well. I'd like to see single-string shorthands for simple cases:

```js
katex.render(formula, element, {
    macros: {
        limsup: "\\operatorname{lim}\\,\\operatorname{sup}",
        RR: "\\mathbb{R}",
        set: "\\left\\{#1\\,\\middle\\vert\\,#2\\right\\}"
    }});
```

You could construct frunctions in the way KaTeX implements them from these string representations. For more complicated things, we could still let the user provide a function. But then compatibility is a major issue, so I'd make this something like this:

```js
katex.render(formula, element, {
    macros: {
        limsup: defineMacro(1, function(api) {
[...]

PR #262, opened 2015-06-21
Restructure functions, environments, symbols

gagern:
This addresses #255 and #256, currently one commit for each. Since both of these modify a pretty large portion of the code, and will likely conflict with many other things which are currently work in progress, I believe that it makes sense to keep these two commits together, so that people will only have to manually resolve conflicts on a larger scale once. The command

    git merge -s recursive -X ignore-all-space origin/master

or similar can be used to avoid conflicts which are solely due to changes in indentation. The resulting code might require reindentation, though. By the way, on my system these changes reduced the minified code size from 147496 to 135148 bytes. Mostly due to `symbols`, I guess. Makes me wonder whether `fontMetrics` should be switched to something which doesn't need names in the literals.


kevinb7:
Even though I think this is moving in the right direction, I was hoping to have more discussion in https://github.com/Khan/KaTeX/issues/255 and https://github.com/Khan/KaTeX/issues/256 before getting a PR.  Sorry for not explicitly stating that.

> Since both of these modify a pretty large portion of the code, and will likely conflict with many other things which are currently work in progress, I believe that it makes sense to keep these two commits together, so that people will only have to manually resolve conflicts on a larger scale once.

I was hoping to merge https://github.com/Khan/KaTeX/pull/132 before doing any refactoring to avoid exactly this problem.  In the future please let us know before hand if you are planning to submit a large PR that might conflict with other peoples' work so that we can coordinate better moving forward.  

Thank you for suggesting ```git merge -s recursive -X ignore-all-space origin/master```.  Hopefully it will make resolving merge conflicts easier.

As for the changes themselves, I have a couple of comments:

There should probably be some convenience functions that set up the context objects and call the handlers.  If we had a ```Function``` class, each of the function objects it could have an ```invoke()``` method which the convenience methods would call.  Since we can't have a class called ```Function``` we could use ```Command``` which I think makes sense seeing that's what they're called in TeX.  Here's a sketch of what that might look like (some details have been omitted for the sake of brevity):
```
function Command(name, props, handler) {
    this.name = name;
    // extract props and handle defaults
    this.handler = handler;
}

Command.prototype.invoke = function(context, args, positions) {
   context.func = this.name;  // should probably probably rename func => name (or cmdName)
   this.handler.apply(context, args, positions);
};

function defineCommand(names, props, handler) {
   for (var i = 0; i < names.length; ++i) {
      module.exports[names[i]] = new Command(name, props, handler);
   }

Parser.prototype.invokeCommand = function(cmd, args, positions) {
   var context = {
       lexer: this.lexer,
       parser: this
   }
   cmd.invoke(context, args, positions);
};
```
If we combine ```ParseNode``` and ```ParseResult``` then the ```positions``` parameter would disappear.

I saw that you re-ordered some of the methods in buildHTML.js and buildMathML.js.  It would make it easier to review if the order was maintained when changing how they're defined.

I prefer ```define``` to ```declare``` because it seems to fit more with TeX vocab, i.e. ```\def```.  Also ```sym``` should be ```defineSymbol``` so that it follows the pattern.  It's longer but it should still be uglified to a single char.

It's nice to see a decrease in our minified code size.  With regards to shrinking fontMetrics, if we convert each entry to an array we should see an appreciable decrease in size as well.
```
{"8672":[-0.064,0.437,0,0]}
```
instead of 
```
{"8672":{"depth":-0.064,"height":0.437,"italic":0,"skew":0}
```

gagern:
> Even though I think this is moving in the right direction, I was hoping to have more discussion in #255 and #256 before getting a PR. Sorry for not explicitly stating that.

I had assumed as much. But I didn't see a lot of discussion happening so far, so I wanted to make things more tangible. I expect that I'll probably be redoing the changes I did here one more time, after you've decided it's ready for merging. The equivalent of a rebase. But now you have something explict to discuss.

> was hoping to merge #132 before doing any refactoring

By all means, don't block other merges due to this. #132 in particular I'm very interested in. I need that more than I need this internal restructuring here.

I'm not sure I agree with the `Command` class you suggest. On the one hand, the step `context.func = this.name` means that you need a new `Command` instance for every command even if they share implementation, which currently isn't required with my approach. But, more severely, having the handler become part of a `Command` class and *still* have its `this` method not refer to that object, that feels even more weird than the `this` for current handlers. So I fear we might be adding confusion by making the handler look like a method when it is not.

On the other hand, I wholeheartedly agree with `Parser.prototype.invokeCommand` to avoid code duplication there. I also agree with your other suggestions. Regarding `sym` I'm indifferent. My main concern, why I chose such a short name, was that a longer name might cause problems with the 80 columns limit. I guess we're still a good distance away from that, though.

kevinb7:
> But now you have something explict to discuss.

I always find talking about code easier.

> But, more severely, having the handler become part of a Command class and still have its this method not refer to that object, that feels even more weird than the this for current handlers.

Point taken.  The problem I was attempting to solve with this suggestion was getting rid of the need to pass information that the objects in funcs already have, e.g. name.  Here's how that might be accomplished without introducing a class.

```
function defineCommand(names, props, handler) {
   var data = { ... };
   for (var i = 0; i < names.length; ++i) {
      data.name = names[i];
      module.exports[names[i]] = data;
   }
}
```
```
Parser.prototype.invokeCommand = function(cmd, args, positions) {
   var context = {
       lexer: this.lexer,
       parser: this,
       name: cmd.name
   }
   cmd.handler.apply(context, args, positions);
};
```

gagern:
@kevinb7 No, absolutely not! `data` is a *single* object, which gets passed around *by reference*. So each iteration of the loop would overwrite the name setting from the previous iteration, and in the end you have just the last name stick. My point is that currently, and with my setup, the function does *not* know its name, and it's cheaper that way since the parser knows that name just fine.

kevinb7:
@gagern whoops... and fixing it would be kind of gross.

kevinb7:
@gagern sounds good.  Can you make the change suggested in https://github.com/Khan/KaTeX/issues/255#issuecomment-114224682?

gagern:
> Thank you for suggesting `git merge -s recursive -X ignore-all-space origin/master`. Hopefully it will make resolving merge conflicts easier.

It did so in a similar situation for me in the past. While we are at it, `git diff -b` (or even `git show -b`) will make reviewing this easier. (And the fact that we are not talking Python or CoffeeScript or some other language with semantic whitespace is really saving a lot of hassle here right now.)

gagern:
> I was hoping to merge #132 before doing any refactoring

Any other in-flight pull requests you'd like to see merged before this one?

* The idea from #261 would benefit from the changes here, I think, but the actual implementation would need to be completely rewritten, so I see no urgent need to pull #261 in first. Even though that view might be slightly frustrating to the author.
* #174 you postponed past #250 which in turn I'd postpone past #266 which would be better off building on this here. The same probably holds for #149.
* #146 is probably already in conflict with #246 so it will require rebasing any way, might as well get rebased to take this one here into account as well.
* #140 sounds about ready for inclusion, so if I have to rebase this on #132, I might as well take that along. But #140 itself needs to rebased to use screenshotter instead of huxley. And it hasn't seen any review afaics.
* I don't know how close #135 is to completion. Should we wait on that, too?

Of course, as long as you agree with the general direction of the code changes here, I can build my next improvements on this, and then rebase them onto the version that eventually makes it into master.

kevinb7:
Thanks for listing these out:
- I don't think we've reached consensus on what needs to be done with https://github.com/Khan/KaTeX/pull/261.  There's also not a lot of code there so rebasing shouldn't be too hard.
- I've closed https://github.com/Khan/KaTeX/pull/146.  They were minor refactorings and not critical.
- https://github.com/Khan/KaTeX/pull/140 can be done better with macros.  Adding a special ```not``` group handler to ```buildTree.js``` is not ideal.  I'm closing it as well.
- I think there's only a minor change I have to make with https://github.com/Khan/KaTeX/pull/135, but we can let this one sit.  If the codebase shifts underneath it not a big deal b/c there's not much code to rebase.
- I've already asked @xymostech to prioritize review of https://github.com/Khan/KaTeX/pull/132.  The plan is to finish the checklist this weekend and have her look at it next week.

> Of course, as long as you agree with the general direction of the code changes here, I can build my next improvements on this, and then rebase them onto the version that eventually makes it into master.

I would strongly urge against this approach.  If there's a refactoring that's necessary for a particular feature then that refactoring should be part of the pull request for that feature.  Large refactorings which make the code generally easier to understand or easier to work with should be done independently and if possible should be split into smaller logical pull requests so that they're easier to review.

gagern:
> if possible should be split into smaller logical pull requests so that they're easier to review.

In that spirit, I suggest you review the two commits here separately, since each is a distinct change for different issue. As I wrote in https://github.com/Khan/KaTeX/issues/266#issuecomment-115748090 I'm willing to post them separately when the time for doing so is right, i.e. after #132.

kevinb7:
@gagern sounds good.  I've responding in more detail in https://github.com/Khan/KaTeX/issues/266#issuecomment-115933756.

gagern:
When I one day repost this here as 5 distinct pull requests, would you want me to write each as two distinct commits, one doing the actual modifications and one with just the reindentation? If you can do a `git diff -w` locally, there is nothing to be gained by separating stuff, but if you review here in GitHub, then having the whitespace changes out of the way might make things easier to review. At the cost of a longer history, of course.

gagern:
Superseded by #345, #346, #347, #348, #356.

Issue #263, opened 2015-06-21
Snapshot alternatives


gagern:
In https://github.com/Khan/KaTeX/pull/246#issuecomment-112148381 I asked about whether PhantomJS could/should be used to create snapshots, instead of Selenium. Today [I gave that a try](https://github.com/gagern/KaTeX/commits/phantomjs-experiments), using the phantomjs 1.9.0 package that comes with Ubuntu 14.04 (though 15.04 is still at 1.9.0). I stayed with Docker for now, since that (with Boot2Docker) appears to be the most reliable option for cross-platform reproducibility, but I envisioned packaging phantomjs and its libs into a tarball which could be used with a `LD_LIBRARY_PATH` override, on Linux without Docker and Ubuntu and all that stuff along. [Here is the Dockerfile.](https://github.com/gagern/KaTeX/blob/phantomjs-experiments/dockers/Phantom/Dockerfile)

I used this to execute a script which was accessed through the `/KaTeX` mount and named on the command line. The docker I created is in no way KaTeX-Specific and therefore highly reusable. (I'll likely upload it to the DockerHub if I ever create an account there.) I had the script access the KaTeX development server running outside the docker, on my development machine, thus avoiding the need for node and other stuff inside the docker.

So far, so good. But when I got [the first results](https://github.com/gagern/KaTeX/commit/7a1b0158f95ca06cf52a3f9742cf4020301ae3d3), things looked *really bad.* Quite a number of test cases had severe problems with vertical positioning. So what should we learn from this? One thing to learn is that the QtWebKit version used by that version of PhantomJS doesn't seem particularly well suited to rendering our output. If that browser is affected, how many other browsers might be affected as well? Should we diagnose the bug, make sure that QtWebKit is violating some spec to render stuff the way it does, then file a bug report? Or should we try to work around the issue, assuming that others might be using QtWebKit or something related in one way or another, so we should support it?

[...]

PR #264, opened 2015-06-22
Selenium images

gagern:
In #263 I suggested (among other things) the use of standard selenium docker images instead of the custom docker image currently used by KaTeX. Reasons for this approach include avoiding the build time associated with the custom docker, using the perfect reproducibility of a fixed image, and the chance of easily obtaining images from Chrome as well. My suggestion received at least one positive response, so here I'm following this up with an actual pull request.

Except for `Lap`, all the other examples can be reproduced *exactly* by the new setup, which is more than I had dared to hope for. I tried to use this opportunity to work around the strange non-determinism of the `Lap` example as well. Assuming that it had to be some kind of race condition, I added this code:

```diff
     driver.get(url);
+    if (key === "Lap") {
+        driver.controlFlow().timeout(200, "Deal with strange non-determinism");
+    }
     driver.takeScreenshot().then(function haveScreenshot(img) {
```

Unfortunately, I still got different results on different machines. Very strange!

Note that I replaced the Python script controlling the Selenium driver from inside the docker with a JavaScript doing so from the outside. The language switch in my opinion makes things more consistent, since the rest of the project is in JavaScript. It also means we can use `npm` to drag in the required dependency, and won't need any install instructions for the Python bindings.

As far as I understand it, the same repository and tag may identify different images over time. Although I don't expect this to happen if the tag is a specific version number, you never know. Just in case, here are the image ids for the images I'm using for this:

    selenium/standalone-chrome:2.46.0   7aa0b75ca012
    selenium/standalone-firefox:2.46.0  7f9e0fc69088


test/screenshotter/test.html
```diff
@@ -8,6 +8,9 @@
       #math, #pre, #post {
         font-size: 4em;
       }
+      body {
+        font-family: "DejaVu Serif",serif;
```
xymostech: Is this font guaranteed to be installed in the selenium docker?

xymostech:
> Unfortunately, I still got different results on different machines. Very strange!

Hmm, are there consistent differences between the images?

test/screenshotter/test.html
gagern: Looks like: with this setting (or even just the first half of it), the Firefox screenshots look no different than without, but the Chrome screenshot switch from some sans-serif default to something that looks just like it does in Firefox.

gagern:
> Hmm, are there consistent differences between the images?

There are essentially two possible outcomes for that test:
case A with SHA-1 d1aca790423e87b83b8465c61aa33d755c680c4f and
case B with SHA-1 8569c87d185597d39d6f70956cf9b7112a54fd7a.

Case A is what we currently have in the repo. B is what I get *most* of the time. But on both OS X and Linux I get A occsionally, though still rarer than B. Case A has the *f* and *g* a pixel or so lower than B. The Chrome rendering so far always was visually indistinguishable from B, even though its SHA-1 was 642f9e6693c2ee5634f4279086c29231d876580f so there must be some very slight differences.

kevinb7:
> Case A is what we currently have in the repo. B is what I get most of the time. But on both OS X and Linux I get A occsionally, though still rarer than B.

It might be worth investigating using webfontloader instead of an arbitrary timeout.  If there are two SHAs that are very similar there are a few options:
- except either
- re-run a number of times to get the one we want
- ignore tests that are not determinant and have a human do the comparison

I'd rather have some automation than no automation.

> Chrome rendering so far always was visually indistinguishable from B, even though its SHA-1 was 642f9e6693c2ee5634f4279086c29231d876580f so there must be some very slight differences.

Could be some minor anti-aliasing difference.

gagern:
> It might be worth investigating using webfontloader instead of an arbitrary timeout.

Feels like overkill, and if even such a long timeout is not sufficient, I doubt webfontloader will be of any help. If you want to try it, feel free to do so, though.

> accept either

I'd prefer that. The first difference is at offset 0x32, where A has the byte 0xC0 while B has 0xF8. We could e.g. have both versions in the repository, and choose the output file name based on that byte.

Come to think of it, and in the light of https://github.com/Khan/KaTeX/pull/268#issuecomment-115712664, perhaps it would be desirable to only store some kind of checksum for all the images, and report checksum violations? That way, your test cases will take *much* less data, which in turn means you can have more test cases without worrying about repo size.

gagern:
What direction should this pull request take? As it is, I'd say this request is fairly self-contained and can be merged, moving any future improvements to separate requests. Possible follow-ups:

* Deal with `Lap`, one way or another
* Let Travis take and compare screenshots (#156)
* Remove images out of repository, just keep their hashes

Of course, if you worry about repository size (as I tend to do), then adding the chrome screenshots just to remove them later on would be unwise. If so, should I move this request into that direction, or just rewrite it to drop the chrome screenshots from their commit for now?

dockers/Screenshotter/screenshotter.sh
```diff
@@ -0,0 +1,21 @@
...
+# directly.
+
+for browserTag in firefox:2.46.0 chrome:2.46.0; do
+    browser=${browserTag%:*}
+    image=selenium/standalone-${browserTag}
+    echo "Starting container for ${image}"
+    container=$(docker run -d -P ${image})
+    [[ ${container} ]] || continue
+    echo "Container ${container:0:12} started, creating screenshots..."
+    res=Failed
```
kevinb7: Why is res failed here?

kevinb7:
I think we should do the simplest thing start with for ```Lap``` and just warn if there are differences with a message directing people to have a look at the image.  We can maybe post the two acceptable images somewhere online.  I'm not sure how to exclude files when using ```git diff --exit-code```. 

It would be nice if we could use docker on travis-ci because I'd rather not have to maintain a separate way of generating screenshots and all the issues that entails.  Good news is that someone got it working so maybe we can too: https://github.com/moul/travis-docker.

I like having the images even if they take up space because if they do change for whatever reason it's pretty easy to look at the before and after images.  If we just have the hash, getting the before image takes a little work.

dockers/Screenshotter/screenshotter.sh
gagern: Because it's changed to `Done` one line down, *if* the run succeeds. If you consider this confusing, I can change it to be the other way round, but that probably wouldn't be any clearer either: the test isn't done or failed at that point.

kevinb7: Since ```res``` isn't being used for anything else wouldn't it be simpler to remove it and just echo "Done taking screenshots..."?

gagern: Why not leave it as it is? Do you consider that line to be so confusing? Personally, I rather get confused if some process tells me it's `Done` when in fact it had some errors and couldn't complete as intended.

gagern:
I [dealt](https://github.com/gagern/KaTeX/commit/fdbb82c98f5352b49b01c3d3a49f432c843f2429) with the `Lap` example by storing [both](https://github.com/gagern/KaTeX/blob/fdbb82c98f5352b49b01c3d3a49f432c843f2429/test/screenshotter/images/Lap-firefox.png) [versions](https://github.com/gagern/KaTeX/blob/fdbb82c98f5352b49b01c3d3a49f432c843f2429/test/screenshotter/images/Lap_alt-firefox.png) in the repository, and accepting either.

dockers/Screenshotter/screenshotter.sh
kevinb7: I'm confused about the behaviour when it fails.  If ```node "$(dirname "$0")"/screenshotter.js ${browser} ${container}``` fails, does it print ```Failed taking screenshots, stopping and removing ${container:0:12}```?  Sorry if this a silly question, I don't do much shell programming.

gagern: Exactly. Currently it doesn't cause the script as a whole to fail (i.e. return non-zero), which would have to be changed at least if you want to run this from Travis.

kevinb7:
@gagern thanks for dealing with ```Lap```.  I finally had a good look through the code.  I think it could be made simpler and more maintainable by using some libraries.  Please have a look at:
- https://github.com/apocas/dockerode
- https://github.com/niegowski/node-pngjs
- https://github.com/substack/minimist (options parser)

For options parsing, could you use explicit switches (minimist handles this really well).  Something like:
```
node screenshotter.js --browser firefox --container ss --exclude file1.png, file2.png
```

kevinb7:
I noticed you used ```pngparse``` in https://github.com/Khan/KaTeX/pull/268, maybe that could work here too.

kevinb7:
I'd like to start using ```minimist``` for our cli app as well.

gagern:
I'm not sure what we'd gain from using `dockerode`, and how. Do you want to replace the outer shell script, `screenshotter.sh`? That should be feasible, but ensuring that we *always* clean up the container, even if the screenshotter itself throws an exception or gets interrupted by Ctrl+C, is far from trivial.

I'm also not sure about what we'd gain from parsing the file at the PNG level. The way I see it, the task of the `Lab` workaround is dealing with the file that while all other results are *exactly* the same file all the time, this one results in a different file from time to time. Whether that file is an image, or a video, or whatever, doesn't really matter to the logic of the script. So doing some image-specific parsing just to decide between the alternatives based on some pixel value seems just a lot of overhead for zero gain. We have two possible files, they differe in some bytes, so look at the first differing byte to decide what situation we're in.

Using some more elaborate command line interface sounds sensible, and minimist would be a good library to use for that. I can add changes along these lines.

kevinb7:
> I'm not sure what we'd gain from using dockerode, and how. Do you want to replace the outer shell script, screenshotter.sh?

I didn't look closely enough at the docker commands in screenshotter.js.  After taking a closer look I realized that you're communicate with selenium running inside docker using screenshotter.js which lives outside docker.  Why not run screenshotter.js from inside docker using the same technique that was used for screenhotter.py?  I think this would simplify things quite a bit.

> I'm also not sure about what we'd gain from parsing the file at the PNG level. 

I guess parsing the whole .png just to get the width and height is a little overkill.

dockers/Screenshotter/screenshotter.js
```diff
@@ -0,0 +1,188 @@
...
+        }
+        var sw = buf.readUInt32BE(16);
+        var sh = buf.readUInt32BE(20);
+        if (sw === width && sh === height) {
+            process.nextTick(takeScreenshots);
+            return;
+        }
+        if (++attempts > 5) {
+            throw new Error("Failed to set window size correctly.");
+        }
```
kevinb7: Does it actually ever take more than two tries?

dockers/Screenshotter/screenshotter.js
```diff
@@ -0,0 +1,188 @@
...
+var fs = require("fs");
+var path = require("path");
+var net = require("net");
+var selenium = require("selenium-webdriver");
+
+var data = require("../../test/screenshotter/ss_data.json");
+
+var dstDir = path.normalize(
+    path.join(__dirname, "..", "..", "test", "screenshotter", "images"));
+var todo = Object.keys(data);
```
kevinb7: Can you make this more descriptive, maybe ```taskQueue``` or ```jobList```?

dockers/Screenshotter/screenshotter.js
```diff
@@ -0,0 +1,188 @@
...
+        var sw = buf.readUInt32BE(16);
+        var sh = buf.readUInt32BE(20);
+        if (sw === width && sh === height) {
+            process.nextTick(takeScreenshots);
+            return;
+        }
+        if (++attempts > 5) {
+            throw new Error("Failed to set window size correctly.");
+        }
+        return setSize(width + w - sw, height + h - sh);
```
kevinb7: There's lots of different heights and widths being used here.  It'd be good to give them more descriptive names, maybe:
- ```width```, ```height``` => ```targetW```, ```targetH```
- ```sw```, ```sh``` => ```actualW```, ```actualH```
- ```w```, ```h``` => ```reqW```, ```reqH```

dockers/Screenshotter/screenshotter.js
```diff
@@ -0,0 +1,188 @@
...
+    return driver.manage().window().setSize(w, h).then(function() {
+        return driver.takeScreenshot();
+    }).then(function(img) {
+        var buf = new Buffer(img, "base64");
+        var ihdr = buf.readUInt32BE(12);
+        if (ihdr !== 0x49484452) {
+            throw new Error("PNG IHDR not in expected location.");
+        }
+        var sw = buf.readUInt32BE(16);
+        var sh = buf.readUInt32BE(20);
```
kevinb7: Pretty much the same code is repeated below.  Could you create a helper function that takes ```img``` and returns an object containing ```width```, ```height```, and ```buffer```?

dockers/Screenshotter/screenshotter.js
```diff
@@ -0,0 +1,188 @@
...
+    url = baseURL + url.substr(toStrip.length);
+    driver.get(url);
+    driver.takeScreenshot().then(function haveScreenshot(img) {
+        var buf = new Buffer(img, "base64");
+        var width = buf.readUInt32BE(16);
+        var height = buf.readUInt32BE(20);
+        if (width !== 1024 || height !== 768) {
+            throw new Error("Excpected 1024x768, got " + width + "x" + height);
+        }
+        if (key === "Lap" && browser === "firefox" && buf[0x32] === 0xf8) {
```
kevinb7: Could you add a comment about why ```buf[0x32]``` is special?

gagern:
> Why not run screenshotter.js from inside docker using the same technique that was used for `screenhotter.py`?

Because the selenium images don't contain node. They contain enough stuff to run a selenium server which can be reached from the outside; everything else I'd consider internal and subject to changes. Furthermore, creating files from within the docker can cause problems with access permissions, as recently discussed.

Having the control logic outside the docker makes us more flexible, since we could as well connect to any other Selenium server, be it on a local Windows machine, a virtual machine or some web-based Selenium service. So I see this as a benefit.

> I guess parsing the whole .png just to get the width and height is a little overkill.

Ah, *that's* what you had been thinking about. Granted, parsing PNG at that point makes more sense than it does to distinguish the two `Lap` alternatives, which is what I had in mind. I still think it would be overkill, though.

dockers/Screenshotter/screenshotter.js
gagern: I don't think so, but the Python code allowed for five tries, so I sticked to that.

dockers/Screenshotter/screenshotter.js
gagern: Good suggestion.

dockers/Screenshotter/screenshotter.js
gagern: Yes, that makes a lot of sense.

gagern:
When writing the code, I felt that [minimist](https://www.npmjs.com/package/minimist) was a bit too minimalistic for my taste. I like to see a clear list of all allowed options somewhere, with name, description, default value. And if I include that in any case, I might as well make it available on the command line via `--help`. Furthermore, multi-word command line options are usually hyphenized lower case, while the corresponding JavaScript identifiers should be camel cased to avoid ugly syntax. I experimented a bit with [nomnom](https://www.npmjs.com/package/nomnom) and found it to be both clear and flexible. I hope you agree with my choice.

kevinb7:
```nomnom``` looks really nice, go with it.  I like how you can give it help strings.  We should adopt it in cli.js as well.

> Because the selenium images don't contain node.

There are selenium images that contain node:
- https://registry.hub.docker.com/u/selenium/node-chrome/
- https://registry.hub.docker.com/u/selenium/node-firefox/

> Furthermore, creating files from within the docker can cause problems with access permissions, as recently discussed.

Good point.  Can you give the ```-p``` option a try?  It maps a host port to a port in the docker image.
```
docker run -d -p 4444:4444 ${image}
```
This should allow you to connect to selenium via localhost:4444 without having to look up the IP address etc.  I tried it a while when I was trying different things to automate taking screenshots.  I ended up not using it because the images were the wrong size.  I like how your code adjusts the size of the browser window until the screenshots are the right size.

**Comment by gagern** [comment]
> There are selenium images that contain node…

I think you are mistaken. Those images result in containers which can be used as a node in a selenium grid, attached to some hub. So this is the term “node” as used by Selenium, which is pretty far from “node” as in node.js.

> Can you give the -p option a try? It maps a host port to a port in the docker image.

Problem with this setup is that the port number 4444 might already be taken on the current machine. With the randomized port numbers, we are independent of such things, and could in theory even run tests for different browsers in parallel one day. In any case, this would only simplify my code by about two lines (simplifying [three lines](https://github.com/gagern/KaTeX/blob/854dc7b0f9072fd731f60eee63fd2443e3a3a651/dockers/Screenshotter/screenshotter.js#L119-L121) to one). The by far more complex part of the whole setup is figuring out our own IP address with respect to the Selenium container, i.e. the address to enter into the Selenium-controlled browser to load content from the KaTeX development server running outside Docker. That wouldn't be avoided as easily, since so far I haven't seen a way to bind a certain port on the outside machine to a local port inside the container. As it stands I see too little benefit in using `-p` just now.

> I like how your code adjusts the size of the browser window until the screenshots are the right size.

This is simply ported from the [`set_driver_size`](https://github.com/Khan/KaTeX/blob/2f552af02d34911a180bc882df9c40736dc846f8/dockers/Screenshotter/screenshotter.py#L18-L35) function of `screenshotter.py`.

**Comment by kevinb7** [comment]
> I think you are mistaken.

Yes I am.

> The by far more complex part of the whole setup is figuring out our own IP address with respect to the Selenium container, i.e. the address to enter into the Selenium-controlled browser to load content from the KaTeX development server running outside Docker.

Couldn't we forward port 7936 and have the browser inside docker connect to localhost:7936?

> That wouldn't be avoided as easily, since so far I haven't seen a way to bind a certain port on the outside machine to a local port inside the container.

Isn't that what ```-p``` does?

> Problem with this setup is that the port number 4444 might already be taken on the current machine. With the randomized port numbers, we are independent of such things, and could in theory even run tests for different browsers in parallel one day.

I've had issues with this in the past.  It'd be interesting to run the tests in parallel.  Maybe we could use https://github.com/joeferner/node-openport to solve the open port issue and ```-p``` to avoid having to figure out the IP address.  Note: I haven't actually used ```node-openport```.

**Comment by gagern** [comment]
`-p` opens a listening port on the bridge device and forwards any connection on that to a daemon listening at a specified port of the container in question. What we need for access to the development server is the other way round: a listening port on the container which forwards any connection to our development server running outside docker. At least that's the way I understand it.

**Comment by kevinb7** [comment]
@gagern you're correct.

I wanted to see what things would look like using dockerode so I coded it up in https://github.com/kevinb7/KaTeX/commit/3b545125d6af1e891ec07d5cce4e79a3c9c6d7e6.  Let me know what you think.  I still have to handle the case where port 4444 is already in use, but that should be pretty easy using `node-openport`.

**Comment by kevinb7** [comment]
Please dis regard that last message.  Using a random port is a superior solution.  I tried getting this to work with `dockerode` by failed.  I kept getting `ECONNREFUSED` when trying to talk to selenium.  I assume that's the problem that `tryConnect()` overcomes.

**Comment by gagern** [comment]
Yes, that's what the comment “Wait for container to become ready” was meant to convey. But that aspect is independent of the port number. As long as you use a newly started container, as opposed to one started by the user a while ago, you have to give it time till the Selenium server inside the container is actually ready to accept requests.

I think dockerode might become interesting if at some point we decide to get rid of my `screenshotter.sh`, and use JavaScript for that logic as well. But right now I'm worried about cleanup after an interrupt, so I won't take that route just now.

**Comment by kevinb7** [comment]
It's weird because when I specified a specific port to open with `dockerode` I was able to get selenium to connect in the callback to `docker.start` without having to wait.

> But right now I'm worried about cleanup after an interrupt, so I won't take that route just now.

Good point.  We'd have to add some extra signal catching code.  Right now if you do a ctrl-c it exits `screenshotter.js` and `screenshot.sh` continues to run and stop the container?

File: dockers/Screenshotter/screenshotter.js [hunk 8]
```diff
@@ -0,0 +1,241 @@
...
+        }
+        if (++attempts > 5) {
+            throw new Error("Failed to set window size correctly.");
+        }
+        return setSize(targetW + reqW - actualW, targetH + reqH - actualH);
+    }, check);
+}
+
+function imageDimensions(img) {
+    var buf = new Buffer(img, "base64");
```
xymostech: Could we use a png parsing library like [png.js](https://github.com/devongovett/png.js) to do this? Then this can be simplified to:
```
var png = new PNG(img);
return {
    width: png.width,
    height: png.height
};
```

gagern: Sure we could do this, but at some performance penalty since we'd parse the whole image just to find its dimensions. We discussed this [above](https://github.com/Khan/KaTeX/pull/264#issuecomment-118710680) and came to the conclusion that this might be overkill.

If you insist, I'll have to compare [png-js](https://www.npmjs.com/package/png-js) against [node-png](https://www.npmjs.com/package/node-png) which I tried first but which can't handle the 4bit depth required for texcmp, and against [pngparse](https://www.npmjs.com/package/pngparse) which is the replacement I'm now using in texcmp. I think it would make sense to stick to *one* library for all our PNG parsing purposes.

gagern: If you want to simplify things, we could abandon the ihdr check which is just a sanity check in any case since any valid PNG file must have `IHDR` as its first chunk. So we could just as well write

```js
var buf = new Buffer(img, "base64");
return {
    buf: buf,
    width: buf.readUInt32BE(16),
    height: buf.readUInt32BE(20)
};
```

which is even shorter than the version with the png parsing step.

kevinb7: > I think it would make sense to stick to one library for all our PNG parsing purposes.

Agreed.

kevinb7: @gagern let's go with png module like @xymostech suggested.  May as well go with `pngparse` seeing as it's already being used by `texcmp`.

kevinb7: If we find that this becomes a bottleneck we can always re-evaluate.

gagern:
I've started using pngparse as [requested](https://github.com/Khan/KaTeX/pull/264#discussion_r34209881), but even I'm surprised by the complexity of this change: pngparse (and png-js as well) only offer asynchroneous interfaces, which we'll have to turn into a promise to make them fit in with Selenium. A very quick search on npmjs didn't show any synchroneous png buffer parsing library, at least none which advertises this fact.

Honestly, I'd like to rather *not* include feba25adc54bfbf808573e8482fe3c46dcd9d691 in this pull, since it adds complexity to the code, it adds complexity to the computation, it makes things harder to maintain due to the more complicated flow of control, and all that for almost no gain that I can see. I'd prefer to omit that commit and either do something along the lines of [my comment above](https://github.com/Khan/KaTeX/pull/264#discussion_r34211147) or leave things as they were.

kevinb7:
@gagern thanks for trying that change out. I didn't realize that the these libraries only have async interfaces.  All of this would be much easier if we could use async/await, but since we can't (yet) let's go with the simplified version that you suggested in your comment above.

gagern:
Let me know when you are satisfied and I should squash this. Also let me know whether you'd prefer a single commit, or whether I should separate Chrome from the rest, i.e. pick dd4304227962d78b9ee37ebb05c106cecb26606b separately during rebase, perhaps together with [this line in the README](https://github.com/gagern/KaTeX/commit/b5ba2739d904494de062b14dc7981611f2ff32a4#diff-c5c7df152a98a847a98add9b82b94340R34).

kevinb7:
Looks good.  Putting the Chrome changes in a separate diff sounds good.

gagern:
@kevinb7: Squashed and ready for merge.

kevinb7:
Awesome!  Thanks for the pull request.

PR #265, opened 2015-06-23
fixed perpendicular symbol

flying-sheep:
as discussed in #261

### DIFF
--- a/src/symbols.js
+++ b/src/symbols.js
@@ -41,7 +41,7 @@ var symbols = {
         "\\perp": {
             font: "main",
             group: "rel",
-            replace: "\u22a5"
+            replace: "\u27C2"
         },
         "\\preceq": {
             font: "main",


gagern:
No. This by itself would cause us to request a symbol which isn't present in the font in question. Unless you also modify the font, this is not an option since we don't know what replacement font or character might be substituted instead.

flying-sheep:
the font has to be modified as it contains a wrong mapping (as said in 261). [U+27C2 perpendicular](http://www.fileformat.info/info/unicode/char/27c2/index.htm) ≠ [U+22A5 up tack](http://www.fileformat.info/info/unicode/char/22a5/index.htm), and the font maps U+22A5 to “perpendicular”

gagern:
Well, the font uses the same symbol for `\bot` as well. It doesn't really matter what that glyph is called, I think. The relevant question, in my opinion, is whether we can have two unicode codepoints refer to the same glyph, both according to the specs of the various font formats and in real world implementations. If not, then either choice is as wrong as the other.

flying-sheep:
true of course!

i just referred to you saying

> The KaTeX_Main font contains a glyph called “perpendicular” and maps unicode codepoint U+22A5 to that.

which is wrong. either it’s able to map both to the glyph (in which case both names would be equally incomplete namings) or it’d only map one codepoint to it, which should then correspond to the glyph name.

gagern:
I don't know whether glyph names have any significance except to the font creator. At least in the TTF world. On the PostScript level, I believe things might be somewhat different, and symbol names are important. But we are not in that world. So while I tend to agree that the font is slightly inconsistent, I see no urgent need to fix that inconsistency.

kevinb7:
I checked and the font we're using doesn't actually have a code point for U+27C2.  What ends up happen (at least on my computer) is that the browser defaults to a font that has this endpoint.  The result is the following rendering which is not what we want.
screen shot 2015-06-23 at 10 51 33 pm

PR #268, opened 2015-06-25
Generate reference images from LaTeX

gagern:
In https://github.com/Khan/KaTeX/pull/156#issuecomment-114572831, @kevinb7 wrote that he'd like to see TeX renderings of our examples before proceeding. Here they are. The docker may still be improved upon, I guess, in terms of size and similar, but it works. I also added a small tool to compute visual diffs, but haven't committed the results of this. One thing I notice is that in TeX, `\llap` and `\rlap` apparently take an argument in text mode.

### DIFF
--- a/dockers/texcmp/Dockerfile
+++ b/dockers/texcmp/Dockerfile
@@ -1,0 +1,11 @@
+# convert from PDF to PNG with -flatten looks really bad on 14.04 LTS
+FROM ubuntu:15.04
+
+MAINTAINER Martin von Gagern <gagern@ma.tum.de>
+
+# Disable regular updates, but keep security updates
+RUN sed -i 's/^\(deb.*updates\)/#\1/' /etc/apt/sources.list && apt-get update
+
+# Install all required packages, but try not to pull in TOO much
+RUN apt-get -qy --no-install-recommends install \
+  texlive-latex-base etoolbox imagemagick ghostscript nodejs
--- a/dockers/texcmp/README.md
+++ b/dockers/texcmp/README.md
@@ -1,0 +1,82 @@
+# How to compare against LaTeX
+
+The tools in this directory can be used to create reference images
+using LaTeX, and to compare them against the screenshots taken from a
+browser.
+
+## Execution environment
+
+### Docker environment
+
+If you don't want to ensure the presence of all required tools, or
+want to make sure that you create reproducible results, simply run
+
+    dockers/texcmp/texcmp.sh
+
+from the root of your KaTeX directory tree.
+This will build a suitable docker image unless such an image already
+exists.  It will then use a container based on that image to generate
+all the images described below.
+
+Note that the files and directories created in the source tree from
+within the docker will be owned by root, so you might have trouble
+deleting them later on.  Be sure you can obtain superuser permissions
+on your computer or know someone who can, just to be safe.
+
+### Native environment
+
+If you want to avoid the overhead of creating a docker container, or
+the even larger overhead of setting up docker and creating the initial
+image, then you may instead execute the commands
+
+    cd dockers/texcmp
+    npm install
+    node texcmp.js
+
+from the root of your KaTeX directory tree.  Required tools include the
+`pdflatex` tool of a standard TeX distribution as well as the
+`convert` tool from ImageMagick.
+
+Note that this approach will use `/tmp/texcmp` as a temporary directory.
+The use of a single directory name here can lead to conflicts if
+multiple developers on the same machine try to use that directory.
+
+Also note that different software configurations can lead to different results,
+so if reproducibility is desired, the Docker approach should be chosen.
+
+## Generated files
+
+After running either of the above commands, you will find two
+(possibly new) subdirectories inside `test/screenshotter`,
+called `tex` and `diff`.
+
+### Rasterized documents
+
+`test/screenshotter/tex` will contain images created by `pdflatex` by
+plugging the test case formula in question into the template
+`test/screenshotter/test.tex`.  This is essentially our reference of
+how LaTeX renders a given input.
+
+### Difference images
+
+`test/screenshotter/diff` will contain images depicting the difference
+between the LaTeX rendering and the Firefox screenshot.  Black areas
+indicate overlapping print.  Green areas are black in LaTeX but white
+in Firefox, while it's the other way round for red areas.  Colored
+input is first converted to grayscale, before being subject to the
+coloring just described.  The pictures will be aligned in such a way
+as to maximize the overlap between the two versions (i.e. the amount
+of black output).  The result will then be trimmed so it can easily be
+pasted into bug reports.
+
+## Command line arguments
+
+Both `texcmp.sh` and `texcmp.js` will accept the names of test cases
+on the command line.  This can be useful if one particular test case
+is affected by current development, so that the effects on it can be
+seen more quickly.
+
+Examples:
+
+    dockers/texcmp/texcmp.sh Sqrt SqrtRoot
+    node dockers/texcmp/texcmp.js Baseline
--- a/dockers/texcmp/texcmp.js
+++ b/dockers/texcmp/texcmp.js
@@ -1,0 +1,265 @@
+"use strict";
+
+var querystring = require("querystring");
+var childProcess = require("child_process");
+var fs = require("fs");
+var path = require("path");
+var Q = require("q"); // To debug, pass Q_DEBUG=1 in the environment
+var pngparse = require("pngparse");
+var fft = require("ndarray-fft");
+var ndarray = require("ndarray-fft/node_modules/ndarray");
+
+var data = require("../../test/screenshotter/ss_data");
+
+// Adapt node functions to Q promises
+var readFile = Q.denodeify(fs.readFile);
+var writeFile = Q.denodeify(fs.writeFile);
+var mkdir = Q.denodeify(fs.mkdir);
+
+// ignore some tests, since they contain commands not supported by LaTeX
+var blacklist = {
+    Colors: "Color handling differs",
+    DeepFontSizing: "\\Huge inside \\dfrac doesn't work for some reason",
+    KaTeX: "Custom command, doesn't exist in LaTeX"
+};
+var todo;
+if (process.argv.length > 2) {
+    todo = process.argv.slice(2);
+} else {
+    todo = Object.keys(data).filter(function(key) {
+        return !blacklist[key];
+    });
+}
+
+// Dimensions used when we do the FFT-based alignment computation
+var alignWidth = 2048; // should be at least twice the width resp. height
+var alignHeight = 2048; // of the screenshots, and a power of two.
+
+// Compute required resolution to match test.html. 16px default font,
+// scaled to 4em in test.html, and to 1.21em in katex.css. Corresponding
+// LaTeX font size is 10pt. There are 72.27pt per inch.
+var pxPerEm = 16 * 4 * 1.21;
+var pxPerPt = pxPerEm / 10;
+var dpi = pxPerPt * 72.27;
+
+var tmpDir = "/tmp/texcmp";
+var ssDir = path.normalize(
+    path.join(__dirname, "..", "..", "test", "screenshotter"));
+var imagesDir = path.join(ssDir, "images");
+var teximgDir = path.join(ssDir, "tex");
+var diffDir = path.join(ssDir, "diff");
+var template;
+
+Q.all([
+    readFile(path.join(ssDir, "test.tex"), "utf-8"),
+    ensureDir(tmpDir),
+    ensureDir(teximgDir),
+    ensureDir(diffDir)
+]).spread(function(data) {
+    template = data;
+    // dirs have been created, template has been read, now rasterize.
+    return Q.all(todo.map(processTestCase));
+}).done();
+
+// Process a single test case: rasterize, then create diff
+function processTestCase(key) {
+    if (blacklist[key]) {
+        return;
+    }
+    var url = data[key];
+    var query = url.replace(/^.*?\?/, ""); // extract query string
+    query = query.replace(/\+/g, "%2B"); // plus doesn't mean space here
+    query = querystring.parse(query);
+    var tex = "$" + query.m + "$";
+    if (query.display) {
+        tex = "$$" + query.m + "$$";
+    }
+    if (query.pre) {
+        tex = query.pre.replace("<br>", "\\\\") + tex;
+    }
+    if (query.post) {
+        tex = tex + query.post.replace("<br>", "\\\\");
+    }
+    tex = template.replace(/\$.*\$/, tex.replace(/\$/g, "$$$$"));
+    var texFile = path.join(tmpDir, key + ".tex");
+    var pdfFile = path.join(tmpDir, key + ".pdf");
+    var pngFile = path.join(teximgDir, key + "-pdflatex.png");
+    var browserFile = path.join(imagesDir, key + "-firefox.png");
+    var diffFile = path.join(diffDir, key + ".png");
+
+    // Step 1: write key.tex file
+    var fftLatex = writeFile(texFile, tex).then(function() {
+        // Step 2: call "pdflatex key" to create key.pdf
+        return execFile("pdflatex", [
+            "-interaction", "nonstopmode", key
+        ], {cwd: tmpDir});
+    }).then(function() {
+        console.log("Typeset " + key);
+        // Step 3: call "convert ... key.pdf key.png" to create key.png
+        return execFile("convert", [
+            "-density", dpi, "-units", "PixelsPerInch", "-flatten",
+            pdfFile, pngFile
+        ]);
+    }).then(function() {
+        console.log("Rasterized " + key);
+        // Step 4: apply FFT to that
+        return readPNG(pngFile).then(fftImage);
+    });
+    // Step 5: apply FFT to reference image as well
+    var fftBrowser = readPNG(browserFile).then(fftImage);
+
+    return Q.all([fftBrowser, fftLatex]).spread(function(browser, latex) {
+        // Now we have the FFT result from both 
+        // Step 6: find alignment which maximizes overlap.
+        // This uses a FFT-based correlation computation.
+        var x, y;
+        var real = createMatrix();
+        var imag = createMatrix();
+
+        // Step 6a: (real + i*imag) = latex * conjugate(browser)
+        for (y = 0; y < alignHeight; ++y) {
+            for (x = 0; x < alignWidth; ++x) {
+                var br = browser.real.get(y, x);
+                var bi = browser.imag.get(y, x);
+                var lr = latex.real.get(y, x);
+                var li = latex.imag.get(y, x);
+                real.set(y, x, br * lr + bi * li);
+                imag.set(y, x, br * li - bi * lr);
+            }
+        }
+
+        // Step 6b: (real + i*imag) = inverseFFT(real + i*imag)
+        fft(-1, real, imag);
+
+        // Step 6c: find position where the (squared) absolute value is maximal
+        var offsetX = 0;
+        var offsetY = 0;
+        var maxSquaredNorm = -1; // any result is greater than initial value
+        for (y = 0; y < alignHeight; ++y) {
+            for (x = 0; x < alignWidth; ++x) {
+                var or = real.get(y, x);
+                var oi = imag.get(y, x);
+                var squaredNorm = or * or + oi * oi;
+                if (maxSquaredNorm < squaredNorm) {
+                    maxSquaredNorm = squaredNorm;
+                    offsetX = x;
+                    offsetY = y;
+                }
+            }
+        }
+
+        // Step 6d: Treat negative offsets in a non-cyclic way
+        if (offsetY > (alignHeight / 2)) {
+            offsetY -= alignHeight;
+        }
+        if (offsetX > (alignWidth / 2)) {
+            offsetX -= alignWidth;
+        }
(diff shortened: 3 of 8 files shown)


kevinb7:
This is awesome!  I was trying to the square root rendering again and started comparing against LaTeX manually and noticed that the amount space under the hrule differs quite a bit between KaTeX and LaTeX.  I'll post some screenshots later, but I was wondering if you noticed any differences for the square root test case?

gagern:
Which test case, `Sqrt` or `SqrtRoot`? I noticed considerable differences for both:

sqrt

sqrtroot

The horizontal spacing in the second one seems to be largely affected by the proportions of the digits in the root. I guess this is because LaTeX has a separate font for smaller sizes, while KaTeX simply scales the 10pt font down, right?

gagern:
> I also added a small tool to compute visual diffs, but haven't committed the results of this. 

I just committed and pushed these diffs to [a separate branch](https://github.com/gagern/KaTeX/tree/diffs/test/screenshotter/diff). I'm not sure if you want this in repo. I don't think I'd want it there, since they are easily reproduced as long as you have imagemagick around. But it might be convenient to have a look now.

spicyj:
> I guess this is because LaTeX has a separate font for smaller sizes, while KaTeX simply scales the 10pt font down, right?

This certainly causes some deviations. We tend to just ignore it.

kevinb7:
@gagern I was look at ```Sqrt``` thanks for posting ```SqrtRoot``` too.  Thanks for posting all of the screenshots.  It's nice to know where were at in terms of LaTeX compliance for current features.  I agree we can just regenerate these as needed.  When we improve any of the renderings we'll update the screenshotter files.  I'm not sure we need to store the -pdflatex.png images either.

gagern:
> I'm not sure we need to store the -pdflatex.png images either.

I'm not sure how different those are from the `*-firefox.png` files. Both can be recreated exactly from the sources with the help of a docker, which requires a working docker setup and considerable amounts of HDD space for the relevant images. Is there any reason to keep the KaTeX snapshots but not their reference images? Is it for the sake of documenting changes to the code base which affect screenshots?

kevinb7:
The difference is that we use ```*-firefox.png``` as reference files to identify unintended changes in the rendering.  We do this by running the snapshotter and seeing if there are any differences in the files according to git.  This process should be automated so that we can fail the build if there are difference, but right now it's a manual process.

If there are, either the code should be fixed so the rendering matches or the reference file should be updated because we fixed/improved the rendering.

On the other hand ```*-pdflatex.png``` images aren't being compared against old version of themselves so we don't need to have them in the repo.

kevinb7:
I was looking through some of the ```-pdflatex.png``` renderings and some of them are clipped.  Do you know why this happening?

gagern:
> some of them are clipped.

That's because I'm aligning the images to maximize overlap, but using the canvas from the Firefox images for reference. So if after that overlap-maximizing alignment, the TeX version extends beyond the margin of the Firefox screenshot, it will get clipped. I fear that without that alignment step, many of the images would have some constant offset over large parts of the canvas, making relative changes harder to see.

kevinb7:
@gagern I would like merge this in.  I'll try to do a better review next week.  Could you add some comments in texcmp.js to describe what's going on?  I assume that you're using the fft to align images.  

I was looking at https://github.com/gagern/KaTeX/blob/diffs/test/screenshotter/diff/SizingBaseline.png and realized that the reason why it's so off is because LaTeX doesn't support sizing inside math.  And with respect to https://github.com/gagern/KaTeX/blob/diffs/test/screenshotter/diff/NestedFractions.png I see that it's lining things up with the bigger fractions and b/c our spacing is wrong the LaTeX screenshot gets clipped.

Could you make it so that the bounding box of the diffs is the union of the bounding boxes of the images being compared?

Also, if there are things that we know are not going to yield invalid comparisons, i.e. sizing commands inside math, can skip those examples?

kevinb7:
I think the reason why the exponents are a bit off is b/c LaTeX actually stretches them horizontally a bit (or uses different glyphs).

spicyj:
It uses different glyphs. Same as https://github.com/Khan/KaTeX/pull/268#issuecomment-115293652.

gagern:
> I would like merge this in.

Without the actual images, right? Will rewrite when I find the time.

> Could you make it so that the bounding box of the diffs is the union of the bounding boxes of the images being compared?

The way I implemented things now, I rasterize the PDF to a temporary file, then align that to produce the `*-pdflatex.png`. So at that point, the clipping has already occurred. The reason I do it this way is because I sometimes want to flip between two renderings, and see what changes. If I move the aligning from the code running inside the docker to the diff, that will no longer be possible that easily. What do you think?

> can skip those examples?

Sure, we can add those to the blacklist.

spicyj:
Can we just add a little padding to the katex screenshots so that the pdflatex ones don't later get cut off?

gagern:
Some of them are nearly exhausting their 1024×768 screenshot size. E.g. `OpLimits` and `Spacing`. If we add margin or padding to the `test.html`, these might start scrolling unless we increase the screenshoit size. Is that desirable?

kevinb7:
> Without the actual images, right? Will rewrite when I find the time.

That's right.  Whenever you have time.

I think @spicyj meant expanding the bounds of the KaTeX images.  After alignment we should know the difference in bounds and be able to pad either image accordingly.  We probably want to keep the existing reference screenshots as is so this would be another set of temp images.  These images could be used for flipping between and generate visualize diffs without clipping any of the renderings.

The clipping doesn't need to be perfect for the images to be useful as indicated by the number of new issues opened due to rendering differences.  I think we should merge it and improve the clipping later.

spicyj:
Well, whichever works. We could probably reduce the font size and be okay if we're running out of room.

gagern:
OK, rewrote things, incorporating your requests. The rasterized images are now not artificially aligned, but use the dimensions of the underlying document. The alignment is computed but used only for the diffs. So these are indeed rendered to the union of both pictures, then trimmed to the area that's actually being used. I added some comments about what's happening in [`texcmp.js`](https://github.com/gagern/KaTeX/blob/texcmp/dockers/texcmp/texcmp.js). And I omitted the resulting images, although I added them [on another branch](https://github.com/gagern/KaTeX/tree/diffs/test/screenshotter/tex). Hope this helps.

I also implemented some fake math sizing commands (although they still [don't work with fractions](http://tex.stackexchange.com/q/252815/16923)), a `\blue` command and something to handle display mode, pre and post text chunks. This makes the images resemble the ones from Firefox much better. The blacklist now contains only three items, namely `Colors`, `DeepFontSizing` and `KaTeX`.

kevinb7:
@gagern Awesome, thanks!  I'll try to review this evening.

dockers/texcmp/Dockerfile
```diff
@@ -0,0 +1,5 @@
+FROM ubuntu:15.04
+MAINTAINER Martin von Gagern <gagern@ma.tum.de>
+RUN sed -i 's/^\(deb.*updates\)/#\1/' /etc/apt/sources.list && apt-get update
```
xymostech: What does this do?

dockers/texcmp/texcmp.js
```diff
@@ -0,0 +1,244 @@
...
+
+Q.all([
+    Q.nfcall(fs.readFile, path.join(ssDir, "test.tex"), "utf-8"),
+    mkdir(tmpDir),
+    mkdir(teximgDir),
+    mkdir(diffDir)
+]).spread(function(data) {
+    template = data;
+    // dirs have been created, template has been read, now rasterize.
+    return Q.all(todo.map(processTestCase));
```
xymostech: If you do something like `processTestCase.bind(null, data)` and add an argument to `processTestCase`, you won't have to make the global `template` variable.

dockers/texcmp/Dockerfile
gagern: Disable normal updates, so we don't receive any bug fix releases or similar, just security relevant fixes. The benefit is that we don't have to pin the versions of the packages we install (which are quite numerous, with all the indirect deps pulled in), and still rely on the same versions getting installed in subsequent runs unless something drastic forces a security update to one of the packages. It's one more step towards better reproducibility.

dockers/texcmp/texcmp.js
gagern: True. I considered the module level variable to be easier to understand and maintain than some `bind` magic.

gagern: At some point in the past, `template` was the result of a call to `readFileSync`, so this is partly historic. But we might get back to something like this one day, for some reason I can#t imagine right now, and then the global variable is more flexible because we don't have to care where it gets set, as long as it does get set by the time processTestCase is invoked.

dockers/texcmp/Dockerfile
kevinb7: Can you add a comment here to that effect?

dockers/texcmp/README.md
```diff
@@ -0,0 +1,73 @@
...
+
+from the root of your KaTeX directory tree.
+This will build a suitable docker image unless such an image already
+exists.  It will then use a container based on that image to generate
+all the images described below.
+
+Note that the files and directories created inside the docker will be
+owned by root, so you might have trouble deleting them later on.  Be
+sure you can obtain superuser permissions on your computer or know
+someone who can, just to be safe.
```
kevinb7: These are files on the host system created by docker, correct?

dockers/texcmp/README.md
```diff
@@ -0,0 +1,73 @@
...
+
+### Native environment
+
+If you want to avoid the overhead of creating a docker container, or
+the even large roverhead of setting up docker and creating the initial
```
kevinb7: large roverhead => larger overhead

dockers/texcmp/README.md
```diff
@@ -0,0 +1,73 @@
...
+    node texcmp.js
+
+from the root of your KaTeX directory tree.  Required tools include the
+`pdflatex` tool of a standard TeX distribution as well as the
+`convert` tool from ImageMagick.
+
+Note that this approach will use `/tmp/texcmp` as a temporary directory.
+The use of a single directory name here can lead to conflicts if
+multiple developers on the same machine try to use that directory.
+
```
kevinb7: It would be good to remind people that the results from this method will differ from machine to machine.

dockers/texcmp/README.md
```diff
@@ -0,0 +1,73 @@
...
+## Generated files
+
+After running either of the above commands, you will find two
+(possibly new) subdirectories.
```
kevinb7: It would be good to list the subdirectories here.

dockers/texcmp/README.md
```diff
@@ -0,0 +1,73 @@
...
+
+`test/screenshotter/diff` will contain images depicting the difference
+between the LaTeX rendering and the Firefox screenshot.  Black areas
+indicate overlapping print.  Green areas are black in LaTeX but white
+in Firefox, while it's the other way round for red areas.  Colored
+input is first converted to grayscale, before being subject to the
+coloring just described.  The pictures will be aligned in such a way
+as to maximize the overlap between the two versions (i.e. the amount
+of black output).  The result will then be trimmed so it can easily be
+pasted into bug reports.
```
kevinb7: Nice!

dockers/texcmp/README.md
```diff
@@ -0,0 +1,73 @@
...
+
+## Command line arguments
+
+Both `texcmp.sh` and `texcmp.js` will accept the names of test cases
+on the command line.  This can be useful if one particular test case
+is affected by current development, so that the effects on it can be
+seen more quickly.
```
kevinb7: It would be good to have an example here.

dockers/texcmp/texcmp.js
```diff
@@ -0,0 +1,244 @@
...
+// Read given file, and apply FFT transformation to it.
+function fftPNG(file) {
+    var deferred = Q.defer();
+    var onerror = deferred.reject.bind(deferred);
+    var stream = fs.createReadStream(file);
+    stream.on("error", onerror);
+    pngparse.parseStream(stream, parsed);
+    return deferred.promise;
+
+    function parsed(err, image) {
```
kevinb7: Could you extract the body of the function (everything below the ```if (err) ...``` check and put it in a function called ```calcAlignment``` (or something like that)?

dockers/texcmp/texcmp.js
```diff
@@ -0,0 +1,244 @@
...
+            deferred.reject(err);
+        } else {
+            deferred.resolve(stdout);
+        }
+    });
+    return deferred.promise;
+}
+
+// Read given file, and apply FFT transformation to it.
+function fftPNG(file) {
```
kevinb7: Can you break this up into separate functions, one for reading the PNG and one for computing the FFT on it?

dockers/texcmp/texcmp.js
```diff
@@ -0,0 +1,244 @@
...
+            }
+        }
+
+        // Step 6b: (real + i*imag) = inverseFFT(real + i*imag)
+        fft(-1, real, imag);
+
+        // Step 6c: find position where the (squared) absolute value is maximal
+        var ox = 0;
+        var oy = 0;
+        var ov = -1;
```
kevinb7: ```ox``` == ```overlapX```?  What's ```ov``` and why is it ```-1```?

kevinb7: ```offsetX```?

dockers/texcmp/texcmp.js
```diff
@@ -0,0 +1,244 @@
...
+            real: real,
+            imag: imag,
+            width: image.width,
+            height: image.height
+        });
+    }
+}
+
+// Create a new matrix of preconfigured dimensions, initialized to zero
+function newMatrix() {
```
kevinb7: Can you switch this to ```createMatrix```? ```newMatrix()``` is too close to ```new Matrix()```.

dockers/texcmp/texcmp.js
kevinb7: `offsetX`?

dockers/texcmp/texcmp.js
```diff
@@ -0,0 +1,244 @@
...
+        }
+        console.log("Positioned " + key + ": " + ox + ", " + oy);
+
+        // Step 7: use these offsets to compute difference illustration
+        var uw = Math.max(browser.width, latex.width) + Math.abs(ox);
+        var uh = Math.max(browser.height, latex.height) + Math.abs(oy);
+        var bx = Math.max(ox, 0);
+        var by = Math.max(oy, 0);
+        var lx = Math.max(-ox, 0);
+        var ly = Math.max(-oy, 0);
```
kevinb7: Can you make the labels a bit more verbose?

dockers/texcmp/texcmp.js
```diff
@@ -0,0 +1,244 @@
...
+        return execFile("convert", [
+            "(", pngFile, "-grayscale", "Rec709Luminance",
+            "-extent", uw + "x" + uh + "-" + lx + "-" + ly,
+            ")",
+            "(", browserFile, "-grayscale", "Rec709Luminance",
+            "-extent", uw + "x" + uh + "-" + bx + "-" + by,
+            ")",
+            "(", "-clone", "0-1", "-compose", "darken", "-composite", ")",
+            "-channel", "RGB", "-combine", "-trim", diffFile
```
kevinb7: template strings would really help here ^_^

dockers/texcmp/texcmp.js
```diff
@@ -0,0 +1,244 @@
...
+var tmpDir = "/tmp/texcmp";
+var ssDir = path.normalize(
+    path.join(__dirname, "..", "..", "test", "screenshotter"));
+var imagesDir = path.join(ssDir, "images");
+var teximgDir = path.join(ssDir, "tex");
+var diffDir = path.join(ssDir, "diff");
+var template;
+
+Q.all([
+    Q.nfcall(fs.readFile, path.join(ssDir, "test.tex"), "utf-8"),
```
kevinb7: you might want to look at denodeify instead of using ```Q.nfcall```

dockers/texcmp/texcmp.js
```diff
@@ -0,0 +1,244 @@
...
+        console.log("Compared " + key);
+    });
+};
+
+// Create a directory, but ignore error if the directory already exists.
+function mkdir(dir) {
+    return Q.nfcall(fs.mkdir, dir).catch(function(err) {
+        if (err.code !== "EEXIST")
+            throw err;
+    });
```
kevinb7: Can you use ```mkdirp``` instead?  https://github.com/substack/node-mkdirp

dockers/texcmp/texcmp.js
kevinb7: why the parens?

dockers/texcmp/texcmp.js
```diff
@@ -0,0 +1,244 @@
...
+// Execute a given command, and return a promise to its output.
+function execFile(cmd, args, opts) {
+    var deferred = Q.defer();
+    child_process.execFile(cmd, args, opts, function(err, stdout, stderr) {
+        if (err) {
+            console.error("Error executing " + cmd + " " + args.join(" "));
+            console.error(stdout + stderr);
+            err.stdout = stdout;
+            err.stderr = stderr;
+            deferred.reject(err);
```
kevinb7: denodeify would help here too

dockers/texcmp/texcmp.js
gagern: My personal policy is that I go with `nfcall` if I use it in only one or perhaps two source lines, and `denodeify` if I call it in more different source lines, or as part of a tight loop. Can I go with that here, or should I `denodeify` everything for the sake of consistency?

dockers/texcmp/texcmp.js
gagern: `ox` and `oy` stand for offset, yes. `ov` is the value associated with the best offset, or to be more precise, the squared absolute value. Since that will never be negative, initializing it to a negative number ensures that the first actual pixel will overwrite that value.

dockers/texcmp/texcmp.js
gagern: The parenthesized parts describe a sequence of operations which is applied to some input and the result is then used as the input on the outer scope. So the first parenthesis reads the LaTeX file, converts that to grayscale and pads it to the union. The second does the same for the browser screenshot. The third takes the first two and takes the minimal gray value for each pixel. These three then form the red, green and blue channel of the final image. See http://www.imagemagick.org/script/command-line-processing.php#stack for details.

dockers/texcmp/texcmp.js
gagern: Sure. I don't need to create parent directories, but if you prefer extra dependencies, I can do that.

dockers/texcmp/texcmp.js
gagern: I don't think I can use `denodeify` to get access to the `stderr` content. If the promise gets rejected, all information from the accepting path is lost, afaik.

dockers/texcmp/texcmp.js
gagern: I can. In that case, the promise would return the parsed image, and almost everything in `parsed` would be a separate function.

dockers/texcmp/texcmp.js
gagern: To me, it feels conceptually close, so this was intentional. But I have no strong opinion there, `createMatrix` is fine with me, too.

dockers/texcmp/README.md
gagern: Right. I wonder whether we should be streaming out those files via a TCP connection instead of a `--volume` mount. Or pre-create the files outside. Or anything else like this. But I see that as a separate pull request, once the core functionality is in place.

dockers/texcmp/texcmp.js
gagern: Can I add comments instead? I'd prefer to keep the `-extent` arguments in one line.

kevinb7: that works too

dockers/texcmp/texcmp.js
kevinb7: As long as we keep katex.js free from extra dependencies. ;)  This is a very minor detail so we can leave it as is.

dockers/texcmp/texcmp.js
kevinb7: Good point.

dockers/texcmp/texcmp.js
kevinb7: Let's go with ```denodeify``` for consistency.

dockers/texcmp/texcmp.js
kevinb7: This was really helpful, can you add it as a comment?

dockers/texcmp/texcmp.js
kevinb7: Maybe ```ov``` => ```maxSquaredNorm``` and ```sq``` => ```squaredNorm```?

dockers/texcmp/texcmp.js
kevinb7: That's okay.  Could you instead inline ```parsed``` so that we're not relying on hoisting for this work?  It's weird to have code appear after a ```return```.

dockers/texcmp/README.md
kevinb7: Fair enough.  Thanks for calling it out in the docs.  Pre-creating the files seems like it would be pretty easy to implement.

gagern:
If you are happy with all the new changes, I'd squash everything into a single commit before merging it.

kevinb7:
@gagern the new changes look great!  Squash it and I'll merge it.

kevinb7:
Thanks for the pull request.

gagern:
Good to have this in. Thanks!

Issue #278, opened 2015-06-29
Add support for the cases environment


azotova:
While #246 added support for the `\begin`...` \end` environment (which is terrific!), it would be also great to support `cases` environment in order to create systems of simultaneous equations (examples here: https://kogler.wordpress.com/2008/03/21/latex-multiline-equations-systems-and-matrices/).

kevinb7:
The ```cases``` environment looks really useful, especially for piecewise functions.

gagern:
Relevant part of `amsmath.sty`:

```tex
\renewenvironment{cases}{%
  \matrix@check\cases\env@cases
}{%
  \endarray\right.%
}
\def\env@cases{%
  \let\@ifnextchar\new@ifnextchar
  \left\lbrace
  \def\arraystretch{1.2}%
  \array{@{}l@{\quad}l@{}}%
}
```

The most important here seems to be the non-default `\arraystretch`, which the current `array` implementation [doesn't support yet](https://github.com/Khan/KaTeX/blob/2f7a54877a412058fe2a134d973aa6fb0d448f50/src/buildHTML.js#L514).

kevinb7:
Implemented by @gagern.  This is now on master, but isn't yet part of an official release.

Issue #269, opened 2015-06-25
More complete support of the array environment


gagern:
While #246 added basic support for the `array` environment, that support is not complete yet. It would be nice to get it (more) complete one day, so let's keep track of what's still missing, since filing a bug for each of these things feels inappropriate just now.

In the format description:
- [ ] `|` – vertical lines between columns (#267)
- [ ] `||` – i.e. repeated vertical lines
- [ ] `*{7}{cc}` – repetition in format description
- [ ] `@{=}` – i.e. custom coumn separators
- [ ] `\extracolsep` for use in `@{…}`
- [ ] `p{2em}` – inserts `\parbox` of given width

In the body:
- [ ] `\hline` – horizontal lines (#267)
- [ ] `\hline\hline` – repeated horizontal lines
- [ ] `\cline[1-3]` – lines spanning some of the columns
- [ ] `\multicolumn{1}{c}{Foo}` – to change the formatting of a single cell
- [ ] `\multicolumn{2}{c}{Foo}` – content spanning multiple columns

In individual cells:
- [ ] `\hfill` at the ends to change the alignment of a single cell
- [ ] `\hfill` within the content to split it into a left-aligned and a right-aligned part
- [ ] `\vline` – vertical line as high as the cell

Format descriptions from the `array` package:
- [ ] `<{foo}` and `>{bar}` – add content to all cells
- [ ] `m{1em}` and `b{3em}` – like `p` but with different vertical alignment
- [ ] `!{=}` – like `@` but keep the space

I'm not sure we'll ever want to support all of this.

azotova:
It would be also great to support {cases} environment to create systems of simultaneous equations (like here: https://kogler.wordpress.com/2008/03/21/latex-multiline-equations-systems-and-matrices/).  I am sorry if this is not the right thread to bring this up, but it seems related to the {array} environment.

kevinb7:
@azotova It uses {array}, but {cases} is a separate environment.  Could you open a new issue for it.

kevinb7:
Request for {cases} environment in https://github.com/Khan/KaTeX/issues/278.

edemaine:
[...]

PR #279, opened 2015-06-30
Implement cases environment

gagern:
See issue #278.

Although the official definition makes use of `@{…}` notation, we use custom spacing instead, since that seems easier for now.


gagern:
The rendering of this code looks pretty close to what LaTeX does. The spacing within `\text{…}` seems to be somewhat different, but that should be an unrelated issue, and not a serious one imho.

diff

kevinb7:
I think the spacing issues are an existing issue as evidenced by https://github.com/gagern/KaTeX/blob/diffs/test/screenshotter/diff/Text.png and should be dealt with separately.

kevinb7:
Most of our code uses camelCase.  Can you update this to match?

```utils.def``` makes me think we're defining something, can you change to something like this?
```
var defaultIfUndefined = function(value, defaultValue) {
   return value !== undefined ? value : defaultValue;
}
```

```\arraystretch``` is a global macro.  Within the scope of an environment defining ```\arraystretch``` shadows the outer global AFAICT.  We probably should have a unified global lookup table at some point, but in the meantime can you add a TODO to do something better for ```\arraystretch``` once we have macros?

**Comment by kevinb7** [comment]
Thanks for the pull request.  Despite the issue with ```\text```, the rendering looks really good!  Can you add a parse test for the new environment?

File: src/environments.js [hunk 1]
```diff
@@ -66,10 +66,12 @@ var environmentDefinitions = [
...
             // Currently only supports alignment, no separators like | yet.
             colalign = colalign.value.map ? colalign.value : [colalign];
-            colalign = colalign.map(function(node) {
+            var cols = colalign.map(function(node) {
                 var ca = node.value;
                 if ("lcr".indexOf(ca) !== -1) {
-                    return ca;
+                    return {
+                        align: ca
+                    };
```
kevinb7: Do you forsee more props being added to this object?

gagern: Yes. Many of the things mentioned in #269 could go in there. `numVerticalLinesAfter` for `|` or `parboxWidth` for `p{…}` or `separatorAfter` for `@{…}` or `intro` and `outro` for `>{…}` and `<{…}`, to name a few ideas.

src/environments.js
```diff
@@ -111,8 +113,36 @@ var environmentDefinitions = [
...
+    {
+        names: ["cases"],
+        handler: function(pos, mode, envName) {
+            var res = {
+                type: "array",
+                arraystretch: 1.2,
+                cols: [{
+                    align: "l",
+                    pregap: 0,
+                    postgap: 1
```
kevinb7: ```fontMetrics.metrics.quad``` instead of ```1```

gagern:
`defaultIfUndefined` is a pretty long name, imo. If you can think of something shorter which is sufficiently clear as well, I'd be happy about it. If not, I'll go with this for now.

src/environments.js
kevinb7: It seems like ```@{...}``` can come before or after a column specifier, the same goes for ```|```.  Maybe instead of grouping these things with each column, ```cols``` could become ```colSpec``` which could contain separate entries for alignment, spacing, and vertical bar specifiers, e.g. ```{|r|@{\quad}c@{\quad}}``` would be represented by:
```
var colSpec = [
   { type: 'sep', value: '|' },
   { type: 'align', value: 'r' },
   { type: 'space', value: quad },
   { type: 'align', value: 'c' },
   { type: 'space', value: quad }
];
```
Thoughts?

gagern: I've thought about similar solutions myself. I think `value` is over-used in the codebase already, so I'd have `{type:"content", align:"c", index:0}` and `{type:"space", width:quad}` and so on. I just started on this and realized that if I want to avoid excessive code duplication, this will require some major work, so it would delay the `{cases}` environment considerably. Shall we postpone this?

kevinb7: I wouldn't worry about it right now.  Whenever we tackle the "format description" part of https://github.com/Khan/KaTeX/issues/269 would be a good time to refactor this.  I like ```width``` instead of ```value``` b/c it conveys more meaning.  Is the ```index``` field necessary for content types?

gagern: Having an `index` around makes it easier to link entries from the `colSpec` back to the actual content from the `body`. Sure, one could have that index in some other variable, and increment that. It's more a matter of convenience.

test/screenshotter/ss_data.json
```diff
@@ -4,6 +4,7 @@
     "Baseline": "http://localhost:7936/test/screenshotter/test.html?m=a+b-c\\cdot d/e",
     "BasicTest": "http://localhost:7936/test/screenshotter/test.html?m=a",
     "BinomTest": "http://localhost:7936/test/screenshotter/test.html?m=\\dbinom{a}{b}\\tbinom{a}{b}^{\\binom{a}{b}+17}",
+    "Cases": "http://localhost:7936/test/screenshotter/test.html?m=f%28a%2Cb%29%3D\\begin{cases}a%2B1%26\\text{if%20}b\\text{%20is%20odd}\\\\a%26\\text{if%20}b%3D0\\\\a-1%26\\text{otherwise}\\end{cases}",
```
xymostech: You don't need to escape all of these characters. For instance, you can put literal spaces instead of `%20`, literal parentheses, etc. I think that'll make this test make a bit more sense. You can even put literal `=` signs.

gagern: That was copy&paste from my browser, I can certainly simplify this. I often think about whether we should have some non-JSON data structure here, so we can just write the formula without any escaping for URL or JavaScript. I'm not sure how to include the other parameters in such a setup, but we could certainly invent a suitable format. Perhaps something that looks like a simple TeX file, but gets parsed by JavaScript? In any case, that parsing imho only makes sense after #264 landed, to avoid the Python components in the mix. What do you think?

kevinb7: In the short term it might be easiest to add a button to index.html that shows the JSON string in an alert or something.  I wouldn't worry about it for this PR.  Here's the string without the URL encoded chars:
```
f(a,b)=\\begin{cases}a+1&\\text{if }b\\text{ is odd}\\\\a&\\text{if }b=0\\\\a-1&\\text{otherwise}\\end{cases}
```

kevinb7:
> ```defaultIfUndefined``` is a pretty long name, imo. If you can think of something shorter which is sufficiently clear as well, I'd be happy about it. If not, I'll go with this for now.

I haven't thought of anything better so let's stay with ```def```.

kevinb7:
@gagern I checked out the branch and tried a couple of different piece-wise functions with it.  I've never used the ```case``` environment before.  I mentioned it before, but the output looks really good.  This is almost ready to merge, just a few things left:
- [ ] The names ```def```'s parameters should be more meaningful
- [ ] JSON string in ss_data.json without URL encoding
- [ ] A test to make sure the new environment parses
- [ ] ```postgap: 1``` should use ```fontMetrics.metrics.quad```

gagern:
It seems I can't tick boxes in your comment, but it's all done now.

kevinb7:
Awesome, thanks!

gagern:
Thanks for the merge.

PR #280, opened 2015-07-01
Add two AMS symbols: checkmark and circledR

JeffEverett:


kevinb7:
Looks good.  Thanks for adding these.

kevinb7:
Updated the wiki to include the new symbols.

PR #282, opened 2015-07-01
package non minified versions to simplify debugging

tokiloki:


kevinb7:
@tokiloki are you running into any issues in particular?

tokiloki:
Kevin, I was getting some error on released version then I switched to github version and now I tried can not reproduce it on release version) Maybe we do not need this request in this case)

kevinb7:
I'm going to close this for now.

kevinb7:
I'm going to close this for now.  Please re-open if you run into issues that require debugging the source.

PR #283, opened 2015-07-02
Update CONTRIBUTING.md

kevinb7:
added "Pull Requests" section with some guidelines, plus some addition code style guidelines.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -100,8 +100,17 @@ Code
  - 80 character line length
  - commas last
  - declare variables in the outermost scope that they are used
+ - camelCase for variables in JavaScript
+ - snake_case for variables in Python
 
 In general, try to make your code blend in with the surrounding code.
+
+## Pull Requests
+ 
+ - link back to the original issue(s) whenever possible
+ - new commands should be added to the [wiki](https://github.com/Khan/KaTeX/wiki/Function-Support-in-KaTeX)
+ - commits should be squashed before merging
+ - large pull requests should be broken into separate pull requests (or multiple logically cohesive commits), if possible
 
 ## CLA
 


CONTRIBUTING.md
```diff
@@ -100,9 +100,17 @@ Code
...
  - declare variables in the outermost scope that they are used
+ - camelCase for variables in JavaScript
+ - snake_case for variables in Python
 
 In general, try to make your code blend in with the surrounding code.
 
+## Pull Requests
+ 
+ - link back to the original issue(s) whenever possible
+ - commits should be squash before merging
```
gagern: Typo: it should be “squash**ed**”

CONTRIBUTING.md
```diff
@@ -100,9 +100,17 @@ Code
...
+ - large commits should be broken into multiple logically cohesive commits if possible
```
gagern: Within a single pull request?

kevinb7: I think either works.  There was that one pull request that you split into two commits.  That worked well, but people may want to submit separate requests.

CONTRIBUTING.md
kevinb7: Thanks.

CONTRIBUTING.md
```diff
@@ -100,9 +100,17 @@ Code
...
+ - camelCase for variables in JavaScript
+ - snake_case for variables in Python
 
 In general, try to make your code blend in with the surrounding code.
 
+## Pull Requests
+ 
+ - link back to the original issue(s) whenever possible
+ - commits should be squashed before merging
+ - large pull requests should be broken into multiple logically cohesive commits (or separate pull requests) if possible
```
xymostech: I'd rather say `broken into separate pull requests (or multiple logically cohesive commits)`, since we'd rather have multiple pull requests.

xymostech:
Looks good to me!

gagern:
While you are at it, do you want to state any preference about the base of the commit? Is it OK for us to squash things to the point where we started diverging, or should we make an effort to rebase to current master? How about bugs, do you want those based on current master or rather on the point in time which introduced the bug? Always barring merge conflicts with current master, of course.

xymostech:
I think it's always preferential to base things off of the current master, because it makes it easier to merge in (and thus it's easier when reviewing to see the actual code that will be committed).

PR #284, opened 2015-07-02
Updated Makefile so that it "make dist" works on OS X

kevinb7:
```cp --recursive``` isn't supported on Mac OS X so use ```cp -R``` instead.


xymostech:
:+1: (sad that macs and linux use different versions of `cp` though)

kevinb7:
It is sad.  I would've preferred ```--recursive```.

PR #285, opened 2015-07-02
Added a link to the list of commands on the wiki.

kevinb7:


xymostech:
Looks good to me! Looks like you've been keeping it up to date too, awesome! I'll try to keep it up to date in the future, too!

kevinb7:
> Looks like you've been keeping it up to date too, awesome!

I've been trying too.  It's still missing stuff.  At this point it'll take me less time to write some that generates a page with all the commands that it will to find and add the remaining missing pieces.

Issue #288, opened 2015-07-03
replace fontforge with freetype-py or fonttools


kevinb7:
We only use fontforge to extract font metrics, but it's big and requires extra dependencies like XQuartz on Mac OS X.  I found installing fontforge on Mac OS X difficult because it was compiled with a different version of Python so I had to install that version of Python alongside the official one and change one of the scripts to use that version.  

freetype-py on the other hand is smaller and easier to install.  I ran the following two commands and was up and running in minutes:
- brew install freetype
- pip install freetype-py

It's also pretty easy to use:
```
import freetype
face = freetype.Face("static/fonts/KaTeX_Main-Regular.ttf")
face.load_char('S')
print face.glyph.metrics.horiAdvance
...
```
This should make it easier for people who want to add even more font commands after https://github.com/Khan/KaTeX/issues/132 lands like ```\mathsfbf```, etc. or use alternative fonts like AMS Euler or STIX as requested in https://github.com/Khan/KaTeX/issues/123.

gagern:
Have you also considered [fonttools](https://github.com/behdad/fonttools)? As far as I can tell, it is pure Python, with no dependency on freetype, which should ease portability quite a bit. In particular, it is to be hoped that it works with the OS X version of Python, with no need to install custom Python versions.

So far I've only been using the `ttx` command line tool from fonttools to extract the `cmap` tables from the fonts, see https://github.com/gagern/KaTeX/commit/86030f7ca43ea3966c5027dafdf44d2ad5c86f03. But it can parse many other TTF tables as well, and I don't think one would have to go via the XML file but could use it directly as a Python package.

gagern:
> `print face.glyph.metrics.horiAdvance`

That could be read from the [`hmtx` table](http://www.microsoft.com/typography/otspec/hmtx.htm) using fonttools, but we don't actually store horizontal advance, do we?

[...]

PR #290, opened 2015-07-04
Adds font metrics to support font commands.

kevinb7:
This is part 1 of 3.  Rendering, screenshots, MathML, and unit tests will
follow in susbequent pull requests.


metrics/parse_tfm.py
```diff
@@ -64,11 +64,14 @@ def __init__(self, start_char, end_char, char_info, width_table,
         self.ligkern_program = LigKernProgram(ligkern_table)
         self.kern_table = kern_table
 
-    def get_char_metrics(self, char_num):
+    def get_char_metrics(self, char_num, zero_indexed=False):
```
xymostech: What does this flag do?

metrics/mapping.pl
```diff
@@ -7,6 +7,8 @@
 # accurately extract the metrics from the corresponding .tfm (TeX font metric)
 # files
 
+# Source: https://github.com/Khan/MathJax-dev/blob/master/fonts/OTF/TeX/makeFF
```
xymostech: It already says this 7 lines up!

metrics/extract_tfms.py
```diff
@@ -31,7 +31,11 @@ def main():
         'cmsy10.tfm',
         'cmti10.tfm',
         'msam10.tfm',
-        'msbm10.tfm'
+        'msbm10.tfm',
+        'eufm10.tfm',
+        'cmtt10.tfm',
+        'rsfs10.tfm',
+        'cmss10.tfm',
```
xymostech: :) thanks for the trailing commas

metrics/mapping.pl
kevinb7: I thought having the full URL was more convenient.  I'll take this out.

metrics/parse_tfm.py
kevinb7: There was one font `rsfs` where glyph indices were messed up.  This corrects them.  I'll add a docstring.

metrics/README.md
```diff
@@ -19,3 +19,21 @@ Once you have these things, run
...
+
+Install the JSON perl module:
+    
+    sudo cpan JSON
+
+Install fontforge with python extensions (this may install python as well):
+
+    brew install fontforge --enable-pyextension
+    
+If you get a "Segmentation fault: 11" when running `make metrics` you'll have
```
xymostech: I think it maybe makes more sense to add `/usr/local/Cellar/python/2.7.9/bin/` to your path than to edit the shebang in the file?

kevinb7: It still picks up the system python.  I'd like to make this go away eventually by replacing `fontforge` with `freetype-py`.  See https://github.com/Khan/KaTeX/issues/288 for details.

gagern: The shebang has to contain a full path, the `PATH` environment variable doesn't have any effect. One common solution is writing it as

    #!/usr/bin/env python2

hoping that that's the name which is being used, and that it's on the path.

kevinb7: The system version of python is 2.7.5.  I tried `#!/usr/bin/env python2`, but it still used 2.7.5 instead of 2.7.9.

gagern:
How shall we deal with this commit here in relation to #303? On the one hand, this commit here is the older one.

On the other hand, reviewing this here might be a lot easier if #303 got merged and this one here rebased onto that one. And there are other benefits as well. With @1f8610cebe2c99223db4082a834fb814afd2b697 you could suggest `make PYTHON=/usr/local/Cellar/python/2.7.9/bin/python` in the README, without danger of a changed shebang getting committed accidentially. You could also drop the install instructions for fontforge.

Perhaps we can simply disregard these issues about python version and fontforge installation here, merge things as they stand knowing that it's a temporary solution, then rebase #303 onto this and add a commit to that which cleans those things up. If this request here is about ready to be merged, then that solution would be preferable, since I'm really waiting for this font stuff to hit master so I can gain some traction with my subsequent restructuring plans.

kevinb7:
I'll rebase this on to https://github.com/Khan/KaTeX/pull/303.

kevinb7:
@xymostech this is ready for review.

xymostech:
(Sorry for taking so long!) This all looks good now! Thanks for the docstring about rsfs, that's very strange!

kevinb7:
Hopefully it's not just my install of TeX.

PR #291, opened 2015-07-04
Adds math commands, HTML rendering, and screenshotter tests.

kevinb7:
This is part 2 of 3.  Part 1 added new fonts metrics.  Part 3 will add MathML support and unit tests.

Depends on https://github.com/Khan/KaTeX/pull/290.

TODO:
- [x] add aliases for ```\frak```, ```\bold```, and ```\Bbb```
- [x] test that fonts don't leak past the end of the group when nesting


gagern:
The commit message writes “Part 2 will add…” which should read “Part 3…”. The screenshot tests don't cover the question of the scope of the new commands, discussed in https://github.com/Khan/KaTeX/pull/132#issuecomment-117103983.

kevinb7:
I decided to leave out the commands defined using ```\DeclareOldFontCommand```.  I'd like to land the basic font commands first and tackle ```rm```, etc. later.

gagern:
I think the scoping checks, or at least one scoping check, makes sense for `\mathrm` and friends as well. To check that they affect nested groups but the effect does not leak past the end of the group.

kevinb7:
> To check that they affect nested groups but the effect does not leak past the end of the group.

I meant to do that and forgot.  Thanks for reminding me.

src/buildCommon.js
```diff
@@ -63,6 +57,65 @@ var mathrm = function(value, mode, color, classes) {
...
+        value = symbols[mode][value].replace;
+    }
+
+    var classes = ["mord"];
+    var color = options.getColor();
+
+    var font = options.font;
+    if (font) {
+        if (font === "mathit" || utils.contains(["\u0131", "\u0237"], value)) {
+            return mathit(value, mode, color, classes.concat(["mathit"]));
```
xymostech: Won't this add the "mathit" class twice sometimes?

src/buildCommon.js
```diff
@@ -308,13 +361,72 @@ var spacingFunctions = {
...
+        variant: "monospace",
+        fontName: "Typewriter-Regular"
+    }
+};
+
 module.exports = {
     makeSymbol: makeSymbol,
-    mathit: mathit,
-    mathrm: mathrm,
+    fontMap: fontMap,
```
xymostech: Why is this exported?

src/buildHTML.js
```diff
@@ -161,13 +161,11 @@ var isCharacterBox = function(group) {
  */
 var groupTypes = {
     mathord: function(group, options, prev) {
-        return buildCommon.mathit(
-            group.value, group.mode, options.getColor(), ["mord"]);
+        return buildCommon.makeOrd(group, options, "mathord");
```
xymostech: I don't understand github's diffing algorithm.

src/buildCommon.js
```diff
@@ -63,6 +57,65 @@ var mathrm = function(value, mode, color, classes) {
...
+        throw new Error("unexpected type: " + type + " in mathDefault");
+    }
+};
+
+/**
+ * Makes a symbol in the italic math font.
+ */
+var mathit = function(value, mode, color, classes) {
+    if (/[0-9]/.test(value.charAt(0)) ||
+        utils.contains(["\u0131", "\u0237"], value) ||
```
xymostech: What are these magic symbols? Can you pull these out into a list somewhere and document why they're different?

xymostech: Also can you indent the tests in the if statement an extra indent so it doesn't line up with the body?

```
if (test &&
        test) {
    body;
}
```

gagern:
The `Lap` example got incorrectly included in this commit, right? I'm really looking forward to getting that non-determinism dealt with automatically one day soon…

src/buildCommon.js
```diff
@@ -308,13 +361,72 @@ var spacingFunctions = {
     }
 };
 
+var greekCapitals = [
```
xymostech: Can you put this above all of the functions that use it?

src/buildCommon.js
```diff
@@ -308,13 +361,72 @@ var spacingFunctions = {
...
+var fontMap = {
+    // styles
+    "mathbf": {
+        variant: "bold",
+        fontName: "Main-Bold"
+    },
+    "mathrm": {
+        variant: "normal",
+        fontName: "Main-Regular"
+    },
```
xymostech: What about "mathit"?

src/buildHTML.js
```diff
@@ -997,6 +995,11 @@ var groupTypes = {
         return makeSpan([options.style.reset(), newStyle.cls()], inner);
     },
 
+    font: function(group, options, prev) {
+        var font = group.value.font;
+        return buildGroup(group.value.body, options.withFont(font), prev);
```
xymostech: This change is added in part 3. Maybe pull it back into this part?

src/buildCommon.js
kevinb7: I left it out because there's specific checks for "mathit" such that the entry here would never be used.  I'll add for completeness.

src/buildCommon.js
kevinb7: It's used by buildMathML.js in the next diff.  I'll take it out of this one.

src/buildHTML.js
kevinb7: I don't think I'll be able to test it without this.  I'll pull the changes to options.js in instead.

src/buildCommon.js
kevinb7: Good catch!

kevinb7:
I'll have to wait until part 1 gets merged before I can tackle "test that fonts don't leak past the end of the group when nesting".

src/buildCommon.js
xymostech: Ah, okay. :+1:

src/functions.js
```diff
@@ -213,6 +213,12 @@ var delimiters = [
     "."
 ];
 
+var fontAliases = {
+    "\\Bbb": "\\mathbb",
+    "\\bold": "\\mathbf",
+    "\\frak": "\\mathfrak"
+};
```
xymostech: Do these font aliases work in LaTeX?

xymostech:
This looks good! I was a little confused about the spacing between the `+` and the `\KaTeX`, but it looks like we just forgot to add an `"mord"` class there. I'll go sneak that in...

kevinb7:
@xymostech let me know when you've sneaked it in and I'll rebase and redo the screenshots.

src/functions.js
kevinb7: I tested these and they do.

src/buildHTML.js
```diff
@@ -819,26 +817,26 @@ var groupTypes = {
...
         e.height = (e.height - 0.2155);
         e.depth = (e.depth + 0.2155);
 
         var x = makeSpan(
-            ["x"], [buildCommon.mathrm("X", group.mode)]);
+            ["x"], [buildCommon.mathsym("X", group.mode)]);
 
         return makeSpan(
-            ["katex-logo"], [k, a, t, e, x], options.getColor());
+            ["katex-logo", "mord"], [k, a, t, e, x], options.getColor());
```
kevinb7: snuck in the "mord"

kevinb7:
A couple of the screenshots are differently, although they don't look different when using github's onion skinning thingy.  In particular these are LimitControls-firefox.png and UnsupportedCmds-firefox.png.  They're probably suffering from the subpixel differences fixed in https://github.com/Khan/KaTeX/pull/324.

kevinb7:
@xymostech this is ready for review now.

src/buildHTML.js
xymostech: :+1:

xymostech:
This all looks good! Sorry it took me so long to review this!

kevinb7:
@xymostech no worries.  Thank you.  I'll rebase later today.

gagern:
I'm so happy that this is going forward now!

PR #292, opened 2015-07-04
Adds MathML support for math font commands.

kevinb7:
This is part 3 or 3.  The first two pull requests added font metrics, HTML rendering, and screenshot tests.

Depends on https://github.com/Khan/KaTeX/pull/291.

### DIFF
--- a/src/Options.js
+++ b/src/Options.js
@@ -6,7 +6,7 @@
  */
 
 /**
- * This is the main options class. It contains the style, size, color and font
+ * This is the main options class. It contains the style, size, color, and font
  * of the current parse level. It also contains the style and size of the parent
  * parse level, so size changes can be handled efficiently.
  *
--- a/src/buildMathML.js
+++ b/src/buildMathML.js
@@ -5,11 +5,14 @@
  */
 
 var buildCommon = require("./buildCommon");
+var fontMetrics = require("./fontMetrics");
 var mathMLTree = require("./mathMLTree");
 var ParseError = require("./ParseError");
 var symbols = require("./symbols");
+var utils = require("./utils");
 
 var makeSpan = buildCommon.makeSpan;
+var fontMap = buildCommon.fontMap;
 
 /**
  * Takes a symbol and converts it into a MathML text node after performing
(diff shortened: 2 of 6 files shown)


src/buildMathML.js
```diff
@@ -24,27 +30,69 @@ var makeText = function(text, mode) {
...
+    return null;
+};
+
+/**
  * Functions for handling the different types of groups found in the parse
  * tree. Each function should take a parse group and return a MathML node.
  */
 var groupTypes = {
-    mathord: function(group) {
+    mathord: function(group, options) {
```
xymostech: Oh boy, adding options to MathML. I guess this day was going to come eventually. :P

src/buildMathML.js
```diff
@@ -298,6 +346,12 @@ var groupTypes = {
         return node;
     },
 
+    font: function(group, options) {
+        var font = group.value.font;
+        var node = buildGroup(group.value.body, options.withFont(font));
```
xymostech:
```
return buildGroup(group.value.body, options.withFont(font));
```
?

src/buildMathML.js
```diff
@@ -443,7 +499,20 @@ var buildGroup = function(group) {
...
+    var startStyle = Style.TEXT;
+    if (settings.displayMode) {
+        startStyle = Style.DISPLAY;
+    }
+
+    // Setup the default options
+    var options = new Options({
+        style: startStyle,
+        size: "size5"
+    });
```
xymostech: Since all of this is exactly the same as for `buildHTML`, and we don't modify `options`, do you want to pull this build into `buildTree` and pass down the `options` instead?

kevinb7:
@xymostech this is ready for review.

src/Options.js
```diff
@@ -6,7 +6,7 @@
  */
 
 /**
- * This is the main options class. It contains the style, size, color and font
+ * This is the main options class. It contains the style, size, color, and font
```
xymostech: Yay grammar

test/katex-spec.js
```diff
@@ -1269,6 +1298,152 @@ describe("An HTML font tree-builder", function () {
...
+        // reverse the order of the commands
+        tex = "\\mathbb{\\color{blue}{R}}";
+        tree = getParsed(tex);
+        markup = buildMathML(tree, tex, defaultOptions).toMarkup();
+        node = "<mstyle mathcolor=\"blue\">" +
+            "<mi mathvariant=\"double-struck\">R</mi>" +
+            "</mstyle>";
+        expect(markup).toContain(node);
+    });
+});
```
xymostech: :O so many tests :+1:

test/katex-spec.js
```diff
@@ -19,7 +25,19 @@ var getBuilt = function(expr, settings) {
...
+    var startStyle = Style.TEXT;
+    if (usedSettings.displayMode) {
+        startStyle = Style.DISPLAY;
+    }
+
+    // Setup the default options
+    var usedOptions = new Options({
+        style: startStyle,
+        size: "size5"
+    });
```
xymostech: :( Having this duplicated all over here makes me sad. Is there a good place this could be consolidated?

kevinb7: I'll see what I can do.

kevinb7:
@xymostech I cleared out some of the repeated code.  We're probably parsing/build trees more often then we have to in the tests, but since the tests run quick I'm not overly concerned.

xymostech:
This looks good to me!

kevinb7:
@xymostech awesome, thanks!

PR #293, opened 2015-07-06
Matrices should be of type ord

gagern:
I've gained some more experience examining the internal workings of LaTeX, documented my findings in [a wiki page about Examining TeX](https://github.com/Khan/KaTeX/wiki/Examining-TeX) and found out that so far I had been using the wrong atom type for arrays, leading to too much space in some situations, e.g. if preceded or followed by a `\mathord`.

I've written my changes as three distinct commits, so you can decide on each of them whether you want it or not. The central one is probably non-controversial. Do you want comments referring to wiki pages? Do you want to keep m wiki page in this form, or change something fundamental about it? Do you want to have a screenshot test for the corrected spacing? Now it compares nicely against what LaTeX produces.


kevinb7:
I think the links to wikipages are good.  Sometimes it's hard to know where to look for information so having the link right there is nice.  If we forget to update a link and it becomes stale then someone will complain.  

The TeX Book makes reference to "Vcent atom", does this correspond the vertical centering of content within cells?

gagern:
> The TeX Book makes reference to "Vcent atom", does this correspond the vertical centering of content within cells?

Not so much within the cells but rather the whole matrix against its surroundings.
When I run `\showlists` on `a\begin{array}{c}b\end{array}c` I see this:

```
\mathord
.\fam1 a
\vcenter
.\vbox(8.39996+3.60004)x14.29166
..\hbox(8.39996+3.60004)x14.29166
...\glue(\tabskip) 0.0
...\hbox(8.39996+3.60004)x14.29166
....\rule(8.39996+3.60004)x0.0
....\glue 5.0
....\glue 0.0 plus 1.0fil
....\mathon
....\OML/cmm/m/it/10 b
....\mathoff
....\glue 0.0 plus 1.0fil
....\glue 5.0
...\glue(\tabskip) 0.0
\mathord
.\fam1 c
```

So the matrix as a whole is a `\vcenter` which is of type Vcent. But the spacing table in chapter 18 doesn't mention Vcent any more. So we need Rule 8 from Appendix G to know that this box will have its center aligned with the math axis, which is σ<sub>22</sub> above the baseline. And the resulting box will become an atom of type Ord. So that's why an `{array}` should be of type Ord.

kevinb7:
I see you're example is testing the spacing before and after the array.  Can you rebase this?

gagern:
Rebased, Travis is happy, ready to go.

kevinb7:
Thanks for the pull request.

PR #294, opened 2015-07-07
Fix symbol group types

gagern:
After learning about `\showlists` and how it can be used to derive the group type (ord, bin, rel, …) of a symbol, I decided to subject our currently defined symbols to some automated tests, verifying the correct type for each. It turns out that I apparently found some incorrectly typed ones. The list of reported differences still isn't empty after this, but that's mostly due to `spacing` becoming `glue` and so on. Note that *many* symbols are called `"textord"` in `symbols.js` but `\mathord` in LaTeX. Is this deliberate? I excluded all these cases from the report so far, since there are so many of them.

### DIFF
--- a/test/symgroups.js
+++ b/test/symgroups.js
@@ -1,0 +1,151 @@
+"use strict";
+
+var fs = require("fs");
+var childProcess = require("child_process");
+
+var opts = require("nomnom")
+    .option("spacing", {
+        flag: true,
+        help: "Print mismatches involving spacing commands"
+    })
+    .parse();
+
+var symbols = require("../src/symbols");
+var keys = Object.keys(symbols.math);
+keys.sort();
+var types = [
+    "mathord", "op", "bin", "rel", "open", "close", "punct", "inner",
+    "spacing", "accent", "textord"
+];
+
+process.nextTick(writeTexFile);
+
+function writeTexFile() {
+    var tex = fs.createWriteStream("symgroups.tex");
+    tex.on("finish", typeset);
+    tex.write("\\documentclass{article}\n" +
+              "\\usepackage{textcomp,amsmath,amssymb,gensymb}\n" +
+              "\\begin{document}\n" +
+              "\\showboxbreadth=\\maxdimen\\showboxdepth=\\maxdimen\n\n");
+    keys.forEach(function(key, idx) {
+        var sym = symbols.math[key];
+        var type = types.indexOf(sym.group) + 1;
+        tex.write("$" + idx + "+" + key + "+" + type + "\\showlists$\n\n");
+    });
+    tex.end("\\end{document}\n");
+}
+
+function typeset() {
+    var proc = childProcess.spawn(
+        "pdflatex", ["--interaction=nonstopmode", "symgroups"],
+        {stdio: "ignore"});
+    proc.on("exit", function(code, signal) {
+        if (signal) {
+            throw new Error("pdflatex terminated by signal " + signal);
+        }
+        fs.readFile("symgroups.log", "ascii", evaluate);
+    }).on("error", function(err) {
+        throw err;
+    });
+}
+
+/* Consider the symbol "\sim" as an example. At the time of this
+ * writing, it has index 431 in our list, and is of group "rel" which
+ * is the fourth of the types listed above. So we construct an input line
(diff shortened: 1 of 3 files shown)


kevinb7:
@gagern I started contributing to this project after first public release so I don't have have context for this, but @xymostech should be able to answer your question.

test/symgroups.js
```diff
@@ -0,0 +1,106 @@
...
+        process.exit(4);
+    }
+}
+
+function extractDigits(str) {
+    var match, res = "";
+    while ((match = reDigit.exec(str)) !== null) {
+        res += match[1];
+    }
+    return +res;
```
kevinb7: Can you use `parseInt(res,10)` instead, just that it's more obvious what's going on?

test/symgroups.js
```diff
@@ -0,0 +1,106 @@
...
+
+    var match, nextidx = 0;
+    while ((match = reMM.exec(log)) !== null) {
+        var list = match[1];
+        match = reParts.exec(list);
+        if (!match) {
+            match = reEmpty.exec(list);
+            if (match) {
+                console.log("Empty: ", nextidx, keys[nextidx]);
+                nextidx++;
```
kevinb7: camelCase please

kevinb7:
@gagern could you add some comments to symgroups.js?  In particular an example of what the output from pdflatex looks like and which pieces of information the regexes are extracting from it.  It doesn't have to be the whole output, just the pertinent bits.

test/symgroups.js
spicyj: `parseInt('12hello', 10)` gives 12 so I tend to prefer unary `+` like this has. If you want to be more explicit, `Number(res)` is equivalent.

kevinb7: I think I'm going to have to give up using `parseInt`.

test/symgroups.js
```diff
@@ -0,0 +1,140 @@
...
+        if (!match) {
+            console.error("Failed to find atom type");
+            console.error(atom);
+            console.error(list);
+            process.exit(3);
+        }
+        var actual = match[1];
+        if (expected !== actual && "math" + expected !== actual &&
+            (expected !== "textord" || actual !== "mathord")) {
+            console.log(expected, actual, idx, keys[idx]);
```
kevinb7: It'd be nice if the output included "expected: " + expected + " + "actual: " + actual + ... or something like.

kevinb7:
Sorry for taking so long to have a serious look at this.  I checked out the branch an ran it.  I was a little confused by the output until I looked at the source.  The output should be labelled more clearly.  Also, it would be nice if there was some indication of what action the operator should take.  After running it looks like it's telling me that we're not defining the spacing commands correctly, but I can't really do anything about those unless we define `\glue`, `\mskip`, and `\kern`.  In the case of symbols with the incorrect type it's very clear that those can and should be fixed immediately.  Maybe the default could be to only show incorrect symbol types and with a certain option it would also show incorrect definitions.

gagern:
> The output should be labelled more clearly.

Agreed. Done that.

> After running it looks like it's telling me that we're not defining the spacing commands correctly.

Perhaps not even that. Perhaps we're just using a different vocabulary. I'm not sure right now how spacing fits in with the whole atom type framework, whether spacing commands are even supposed to have a type.

> Maybe the default could be to only show incorrect symbol types and with a certain option it would also show incorrect definitions.

Maybe we should ignore any mismatch if the KaTeX type is `spacing`. After all, I'm not verifying definitions at all, just the atom types. I took that approach in f2cf08569fc49e1aa35b19d53ee9d198bc22d574. This leaves us with three reported mismatches: `\colon` which should be handled in #307, and `\space` and `\omicron` which don't produce any output. The former makes a lot of sense, since spaces in math mode get ignored. And `\omicron` is undefined, so I'm surprised that this didn't cancel the typesetting, but not surprised that I don't see it in the output.

kevinb7:
This is looking pretty good.  Could you make one minor change and then squash it?  The variables `expected` and `actual` are reversed from what I would expect.  Could you rename them to `katexType` and `latexType` or something like that?

gagern:
I kept the actual fix as a separate commit, distinct from the tool used to identify the problems.

kevinb7:
I like how you split it up.  Thanks!

PR #299, opened 2015-07-08
Provide more delimiters

gagern:
This adds `\lgroup`, `\rgroup`, `\lmoustache` and `\rmoustache`, provides `\lVert` and `\rVert` with the correct type for each, and makes `\lvert`, `\rvert`, `\lVert` and `\rVert` available through `\left`…`\right`.

### DIFF
--- a/src/delimiter.js
+++ b/src/delimiter.js
@@ -217,6 +217,26 @@ var makeStackedDelim = function(delim, heightTotal, center, options, mode) {
         bottom = "\u23ad";
         repeat = "\u23aa";
         font = "Size4-Regular";
+    } else if (delim === "\\lgroup") {
+        top = "\u23a7";
+        bottom = "\u23a9";
+        repeat = "\u23aa";
+        font = "Size4-Regular";
+    } else if (delim === "\\rgroup") {
+        top = "\u23ab";
+        bottom = "\u23ad";
+        repeat = "\u23aa";
+        font = "Size4-Regular";
+    } else if (delim === "\\lmoustache") {
+        top = "\u23a7";
+        bottom = "\u23ad";
+        repeat = "\u23aa";
+        font = "Size4-Regular";
+    } else if (delim === "\\rmoustache") {
+        top = "\u23ab";
+        bottom = "\u23a9";
+        repeat = "\u23aa";
+        font = "Size4-Regular";
     } else if (delim === "\\surd") {
         top = "\ue001";
         bottom = "\u23b7";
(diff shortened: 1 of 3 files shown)


xymostech:
I'm not positive about the `\lgroup`, `\rgroup`, `\lmoustache`, and `\rmoustache` additions, which I think is why they weren't added with the rest of the delimiters. If you render these in TeX or LaTeX, you get funky things:

`\lgroup \rgroup \lmoustache \rmoustache` in LaTeX:

image

`\lgroup \rgroup \lmoustache \rmoustache` in TeX:

image

MathJax does render them the way you might expect, just using their small forms:

image

LaTeX and TeX do seem to show them correctly when used as delimiters, but do so slightly differently

`a \left \lgroup \right . \left \rgroup \right. \left \lmoustache \right. \left \rmoustache \right.` in LaTeX:

image

`a \left \lgroup \right . \left \rgroup \right. \left \lmoustache \right. \left \rmoustache \right.` in TeX:

image

So I'm not really sure what to do. The other change looks good though!

gagern:
I'd say the group and moustache delimiters are primarily intended as delimiters, i.e. with `\big` and friends, or `\left` and `\right`. If TeX and LaTeX differ in size, I'd say we should aim for the LaTeX rendering, since most users will be more familiar with LaTeX than plain TeX I guess.

The fact that TeX and LaTeX renders these things strangely when they are not used as delimiters doesn't seem to be a desirable feature to me. This is in my opinion similar to how in https://github.com/Khan/KaTeX/pull/132#issuecomment-113748502 I argued for some kind of fallback: symbols not supported by a selected font should render in some default font. This is contrary to how TeX deals with the situation, namely printing any symbol which accidentially happens to be in the corresponding location of the selected font. I'd say that kind of behavior is unspecified in TeX so we should be free to do something more sensible. Likewise here: TeX prints just some parts of the delimiters in question, which doesn't look like intentional behavior to me, so we should be allowed to do something more sensible and print the complete glyphs we have at our disposal.

gagern:
Travis just made me aware that the small group and moustache symbols were already available. It's just their use as delimiters which was not available. So since the controversial part was already included in the code base, I hope that the part about using them as delimiters is non-controversial.

kevinb7:
If no one's filed this bug against TeX, maybe we should.  Is Knuth still giving out rewards?

On a more serious note, my preference would be to allow the delimiters in question.  Hopefully this encourage push TeX and LaTeX to fix this issue.

kevinb7:
@gagern thanks for the pull request.  Can you add the new delimiters to the wiki?

gagern:
Added them in [this change](https://github.com/Khan/KaTeX/wiki/Function-Support-in-KaTeX/_compare/60e6f2dbd1a40900385b25eca87dafc116f5bc60...1ccd8a1e05885e9c91d3b4a76fab39eb54052f4e). Thanks for merging!

Issue #301, opened 2015-07-09
Split fontMetrics in manually edited part and auto-generated part


gagern:
I'm slightly conderned about the way `fontMetrics.js` contains a large amount of data (even more the more fonts we deal with) which is auto-generated, but also has some parts which are to be edited manually. For one thing, the whole diff usually doesn't show up on git. It would in my opinion be preferable to have one file which is *purely* auto-generated, and have another which contains all the hand-written parts. The auto-generated part could even be some JSON, I think. So you'd say

```js
var metricMap = require("fontMetricsData");
```

and let the generation process write to `fontMetricsData.json` (simply overwriting its content) instead of modifying `fontMetrics.js`. I haven't tested whether browserify can deal with JSON, but if not, adding a simple `module.exports =` to the front and calling it `fontMetricsData.js` isn't too hard.

Personally, I'd like to have the JSON formatted with one glyph per line, so that modifications can be seen more easily in the diff. On the other hand, we might want to think about ways to compactify that data, changing the field names into something position-dependent.

kevinb7:
I was going to open an issue for this.  Thanks for doing so.  We should dump the data into file with a .json extension, then we don't have to worry about ```module.exports = ```.

gagern:
Solved by #303.

PR #303, opened 2015-07-09
Improve font metrics creation

### LINKED ISSUE #288: replace fontforge with freetype-py or fonttools

gagern:
This is a huge request of various things regarding font metrics creation. I think it is best if you review this one commit at a time, since each commit is a more or less self-contained step in that chain of changes.

* @1f8610cebe2c99223db4082a834fb814afd2b697 solves a build problem if the `python` binary on the `PATH` is Python 3.
* @c20f06da510b228c0a83b6fdb561534581181cea and @ae24b8a4665f9f32659cba9485cc608a3086b17a reformat the metrics data according to #301.
* @8be3b579c009a27ae941d7ec283b1aa85341cc8d fixes a bug I found along the way; here you have to verify some glyphs, comparing TTF content against [these charts](http://www.math.union.edu/~dpvc/jsmath/symbols/welcome.html). Discussion about this happened on #288.
* @db7e46d8c235b3e5ad0c8e271813469bc2e12983 resolves #288 by switching from [fontforge](https://fontforge.github.io/) to [fonttools](https://github.com/behdad/fonttools).

I know you prefer distinct pull requests for distinct issues, but I want to get this out of my head for now, and I think the history will benefit from having a direct relationship between these commits. So please accept my suggestion of reviewing one commit at a time. If at some point you decide that things up to a certain point make sense, and you don't have time to review past that or even decide against later commits, feel free to merge parts of this earlier.


gagern:
I'll commit modifications to existing commits as squashes and fixups, so that once you're satisfied, I can `git rebase -i --autosquash` everything to keep the logical units intact.

kevinb7:
It looks good.  I have an appointment soon so I'll have another look this afternoon in case I missed something, but I don't foresee any big changes here.

kevinb7:
@gagern other than a few minor changes it looks good.

xymostech:
This looks great! Thanks for this! All of this infrastructure was set up in the week before release, and hasn't been really touched since. :) 

Also, I actually think that it makes more sense to leave all of these commits separate, since they're all self-contained, and they make it easier to figure out what changed. Squashing commits makes sense if you break things and then fix them later (because then doing a git bisect might land you in a broken state) but these all seem fine to merge in. (But maybe the last commit could be merged into the second one, which is something you can do with rebase -i?)

gagern:
> But maybe the last commit could be merged into the second one, which is something you can do with rebase -i?

Even better with `git rebase -i --autosquash`, as [I already said](https://github.com/Khan/KaTeX/pull/303#issuecomment-120048942). I'll wait on a decision about my most recent fixup, whether that should be included or not (discussion [here](https://github.com/gagern/KaTeX/commit/8be3b579c009a27ae941d7ec283b1aa85341cc8d#comments) or on that commit), and if that's resolved I'll rebase and you can merge, right?

kevinb7:
@gagern thanks for the fixup and comments.  Autosquash the fixups and I'll merge the commits.

kevinb7:
@gagern for some reason github isn't notifying me when commits change.  Can you post a quick message when things are ready to merge after rebasing?

gagern:
@kevinb7 Sure, I can do that. Thanks for merging!

kevinb7:
Thanks, and thanks for the pull request.

spicyj:
Yes, GitHub doesn't notify when PRs are updated. It sucks.

PR #306, opened 2015-07-14
Update development server for new browserify version

gagern:
Current browserify no longer accepts an options hash to the bundle method.

Seems I broke the dev server in 1603162267d5a3217ff567847581732220ba7151 for #303.


kevinb7:
I hadn't even noticed it was broken.  Thanks for fixing it.  :+1:

PR #308, opened 2015-07-14
Automatically start a development server

gagern:
This avoids one of the few requirements we have left: you no longer have to start a KaTeX development server, the script will do it for you, using a random port number.

To reproduce the old behaviour, explicitely state `--katex-port=7936`.

Note that in contrast to all the find-an-open-port libraries I looked at, this one will open the server and keeping it open, avoiding the small time window where the open port might get re-used by some other process.

### DIFF
--- a/dockers/Screenshotter/screenshotter.js
+++ b/dockers/Screenshotter/screenshotter.js
@@ -2,10 +2,12 @@
 
 var childProcess = require("child_process");
 var fs = require("fs");
+var http = require("http");
 var path = require("path");
 var net = require("net");
 var selenium = require("selenium-webdriver");
 
+var app = require("../../server");
 var data = require("../../test/screenshotter/ss_data.json");
 
 var dstDir = path.normalize(
@@ -49,7 +51,6 @@ var opts = require("nomnom")
     })
     .option("katexPort", {
         full: "katex-port",
-        "default": 7936,
         help: "Port number of the KaTeX development server"
     })
     .option("include", {
@@ -76,10 +77,11 @@ if (opts.exclude) {
 }
 
 var seleniumURL = opts.seleniumURL;
-var katexURL = opts.katexURL;
 var seleniumIP = opts.seleniumIP;
 var seleniumPort = opts.seleniumPort;
+var katexURL = opts.katexURL;
 var katexIP = opts.katexIP;
+var katexPort = opts.katexPort;
 
 //////////////////////////////////////////////////////////////////////
 // Work out connection to selenium docker container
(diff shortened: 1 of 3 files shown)


dockers/Screenshotter/screenshotter.js
```diff
@@ -128,17 +130,53 @@ if (seleniumURL) {
...
+//////////////////////////////////////////////////////////////////////
+// Start up development server
+
+var devServer = null;
+var minPort = 32768;
+var maxPort = 61000;
+
+function startServer() {
+    if (katexURL || katexPort) {
+        process.nextTick(tryConnect);
```
kevinb7: I'll have to start using this.  It's way nicer than the `timeout(callback, 0)` trick that I'm used to.

dockers/Screenshotter/screenshotter.js
```diff
@@ -128,17 +130,53 @@ if (seleniumURL) {
...
+        if (++attempts > 50) {
+            throw new Error("Failed to start up dev server");
+        }
+        process.nextTick(startServer);
+    };
+    var port = Math.floor(Math.random() * (maxPort - minPort)) + minPort;
+    var server = http.createServer(app).listen(port);
+    server.once("error", onerror);
+    server.once("listening", function() {
+        server.removeListener("error", onerror);
```
kevinb7: why remove it?

dockers/Screenshotter/screenshotter.js
```diff
@@ -128,17 +130,53 @@ if (seleniumURL) {
...
 
-var attempts = 0;
-process.nextTick(seleniumIP ? tryConnect : buildDriver);
 function tryConnect() {
+    if (!katexURL) {
+        katexURL = "http://" + katexIP + ":" + katexPort + "/";
+        console.log("KaTeX URL is " + katexURL);
+    }
+    if (!seleniumIP) {
+        process.nextTick(buildDriver);
```
kevinb7: I just notice that you use `process.nextTick` here but `setTimeout` for the socket handling.  Any reason for the difference?

dockers/Screenshotter/screenshotter.js
```diff
@@ -231,7 +271,20 @@ function takeScreenshot(key) {
...
+            } else {
+                deferred.fulfill();
+            }
+        });
+        return deferred.promise;
+    }).then(function() {
         console.log(key);
+        if (--countdown === 0) {
+            // devServer.close(cb) will take too long.
+            process.exit(0);
```
kevinb7: how long does `devServer.close` take?

dockers/Screenshotter/screenshotter.js
```diff
@@ -202,6 +240,8 @@ function imageDimensions(img) {
 //////////////////////////////////////////////////////////////////////
 // Take the screenshots
 
+var countdown = listOfCases.length;
```
kevinb7: doing things in series with promises is such a pain.

server.js
```diff
@@ -7,7 +7,9 @@ var less = require("less");
 
 var app = express();
 
-app.use(express.logger());
+if (require.main === module) {
+    app.use(express.logger());
+}
```
kevinb7: cool, I didn't realize that you could do that

kevinb7:
This is really nice.  Thanks for doing this.

dockers/Screenshotter/screenshotter.js
gagern: I had somehow assumed that the `app` would take care of error handling from that point onward, but that's probably an incorrect assumption. On the other hand, once the server is started, we don't want any more startup attempts, so I'll have to add that distinction to the error handling function.

dockers/Screenshotter/screenshotter.js
gagern: Not really, might as well make that all `nextTick`. I guess that when initially writing the code, I might have been thinking about keeping things simpler by just using one function, but since I'm using `nextTick` in a lot of cases now, it's certainly the more appropriate choice.

dockers/Screenshotter/screenshotter.js
gagern: It's not done in series, or at least not from our point of view. The tasks all get scheduled in parallel, and it's the Selenium implementation which serializes things again. Anyway, something like `Q.all` would be quite useful here, but introducing a second promise library into the game feels like a bad idea. Although it was just when I wrote this code that I first found out that the Selenium library even had a `Deferred` object. I had completely missed the “Types” section of [the API reference](http://seleniumhq.github.io/selenium/docs/api/javascript/index.html).

dockers/Screenshotter/screenshotter.js
gagern: Feels like forever, but realistically it's probably the one minute commonly configured as [`TIME_WAIT`](https://en.wikipedia.org/wiki/Transmission_Control_Protocol#Protocol_operation) for a TCP connection.

server.js
gagern: I had very much hoped that there would be something like [Python's `__name__ == "__main__"` check](http://stackoverflow.com/q/419163/1468366) and was very glad when I found [this](https://nodejs.org/api/modules.html#modules_accessing_the_main_module) in the Node.js docs.

By the way, this is the same approach I'd like to take in Python when merging all the metrics manipulation modules into a single process: have each file be both a module and an executable script.

gagern:
@kevinb7: Tweaked code to incorporate your comments. Let me know when you consider this ready for squashing.

dockers/Screenshotter/screenshotter.js
kevinb7: Right.  I was kind of surprised that Selenium allows you schedule stuff in parallel.  It's nice that it queues the screenshot requests for us.

kevinb7:
This is ready to squash and merge.  Thanks for the minor tweaks.

gagern:
@kevinb7 squashed and ready. Travis happy.

kevinb7:
Awesome.  Thanks for the pull request.

PR #310, opened 2015-07-15
Switch Screenshotter data from JSON to YAML

gagern:
I'd like to move the screenshotter data from JSON to YAML. I think the old data format is bad for a number of reasons:

* Escaping TeX in JSON is a pain: you have to double all the `\`, you can't easily include line breaks in the source form for readability, and staying below the 80 chars line length limit is almost impossible.
* Escaping TeX as a query string is a pain, too, since you have to URL-encode all the `%` and `#` at the very least.
* We are already stripping away the host and port part of the URL in every single application that uses that data, substituting custom values instead in the case of the screenshotter.
* When we extend the screenshotter towards render-to-canvas, we'll have to replace the name of the HTML file as well, so a complete URL would make even less sense than it does now.
* For the texcmp tool we're already clumsily decoding the encoded query string
* The more combinations of browser, rendering technique and so on we have, the more structure we want in our data. E.g. it makes sense to keep blacklists with the test cases, I think, as done in this commit.

[YAML](http://www.yaml.org/) is a reasonably well-established standard, and easy to read and write. It can express complex structures if the need arises. The main remaining problem with YAML is that if a value starts with `{`, it will get interpreted as a mapping, which can be avoided by making use of `>`-introduced multi-line text blocks which will always be read as text. Single-quoted strings would be a viable alternative as well. I think that slight drawback doesn't yet justify inventing a custom format here.


kevinb7:
The .yaml file is much cleaner nice.

test/screenshotter/ss_data.js
```diff
@@ -0,0 +1,14 @@
+var fs = require("fs");
+var querystring = require("querystring");
+var jsyaml = require("js-yaml");
```
xymostech: nit: sort these

test/screenshotter/ss_data.js
```diff
@@ -0,0 +1,14 @@
...
+
+var dict = fs.readFileSync(require.resolve("./ss_data.yaml"));
+dict = jsyaml.safeLoad(dict);
+for (var key in dict) {
+    var itm = dict[key];
+    if (typeof itm === "string") {
+        itm = dict[key] = { m: itm };
```
xymostech: Could we maybe rename this `m` parameter to `tex` or something? Now that it's easier to see what's going on, we might as well name things reasonably. :)

test/screenshotter/ss_data.js
```diff
@@ -0,0 +1,14 @@
+var fs = require("fs");
```
xymostech: Could you add a comment here describing what this file does?

dockers/texcmp/texcmp.js
```diff
@@ -16,18 +15,12 @@ var readFile = Q.denodeify(fs.readFile);
 var writeFile = Q.denodeify(fs.writeFile);
 var mkdir = Q.denodeify(fs.mkdir);
 
-// ignore some tests, since they contain commands not supported by LaTeX
-var blacklist = {
```
xymostech: I think I'd rather leave this in this file, since it's not related to the other stuff, and gets left in the query parameter?

test/screenshotter/ss_data.yaml
```diff
@@ -0,0 +1,85 @@
+Accents: \vec{A}\vec{x}\vec x^2\vec{x}_2^2\vec{A}^2\vec{xA}^2
```
xymostech: This looks much better!

How did you generate this? Was it done automatically?

Now that this is in yaml, could you add some comments at the top describing the options that we can have in here?

gagern: No, that was manually converted, with lots of search & replace, some macros and manual line wrapping. Will add comments.

dockers/texcmp/texcmp.js
gagern: I feel like this is more closely related to the test case than to the application running it. Or, perhaps phrased better, like this is more like configuration data not executable code, so it should go into the configuration file not the implementation file. When someone creates a new test case as a copy of an existing one, it makes sense to copy any special behavior as well. It is true that these settings get included in the query parameter, but having more data in there doesn't hurt anyone. I also like the idea of using the value of the mapping for a human-readable explanation of the reason why this should be blacklisted, and having it in JS entails many of the drawbacks of JSON, as exemplified by the doubled backslashes below. Simply writing that reason with little concern for syntax, perhaps even over multiple lines, makes it more readable in my opinion.

Have I convinced you, @xymostech?

test/screenshotter/ss_data.js
gagern: Sure.

test/screenshotter/ss_data.js
gagern: I tend to sort built-in modules before explicitely installed ones, but that's personal preference and I'll gladly adapt this.

dockers/texcmp/texcmp.js
kevinb7: I agree with @xymostech that we should put extraneous stuff in the query string, but I also agree with @gagern that we should keep all of the data about test cases together.  Why not have ss_data.js export a dictionary where each entry is an object with a `query` property and a `nolatex` property.  Best of both worlds.

gagern: @kevinb7 I assume taht we should *not* put extraneous stuff in the query, right? I can make `ss_data.js` just include selected keys into its `query` property, that's easy enough. Will do so in my next commit, soon.

gagern:
Now the query string only contains the bare minimum it needs to contain. But I strongly feel that this code adds complexity for no actual gain, so I'm including that last commit against my better judgement.

gagern:
Merging current master, I had to add two more keys to the whitelist of things which are to be added to the `query` property. So this kind of change would require modifying one more file in the future, which I consider bad. I could of course change the whitelist to a blacklist, but I still maintain that not filtering at all is the best solution, since it doesn't hurt anyone and keeps things simple.

test/screenshotter/ss_data.js
kevinb7: `fs` is part of the node standard library and contains a bunch file handling functions.

kevinb7:
We could just not have a whitelist and then we don't have to worry about updating the whitelist.  I'm going to merge this as is, we can always remove the whitelist later if it's cumbersome.

kevinb7:
@gagern thanks for the pull request.

PR #313, opened 2015-07-20
Allow for the contrib folder to be included

franciscop:
If you include the `/contrib` folder in bower, then we can use its files with gulp/grunt easily. In my case I need `auto-render` and I am including it as a separate file and then joining them, but I think it'd be easier for everyone who wants to use it if it was only a dependency than if it is a dependency and a stray file (that we have to include in other ways).

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -27,7 +27,6 @@
     "/build",
     "/test",
     "/src",
-    "/contrib",
     "/dockers",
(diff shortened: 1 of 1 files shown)


spicyj:
dist/contrib/auto-render.min.js should already be available in bower. Let me know if that's not what you're seeing.

Issue #198, opened 2015-03-08
KaTeX MathML should use munderover for scripted large operators


fred-wang:
For example

http://katex.xymostech.com/?text=\displaystyle%20\sum_a^b

```
  <math>
    <semantics>
      <mrow>
        <mstyle scriptlevel="0" displaystyle="true">
          <msubsup>
            <mo>&sum;</mo>
            <mi>a</mi>
            <mi>b</mi>
          </msubsup>
        </mstyle>
      </mrow>
      <annotation encoding="application/x-tex">\displaystyle \sum_a^b</annotation>
    </semantics>
  </math>
```

Here msubsup should be replaced with munderover, so that the scripts are correctly placed in displaystyle mode.

kevinbarabash:
Fixed by #788.

PR #314, opened 2015-07-21
Add support for \limits and \nolimits controls

### LINKED ISSUE #35: Add \limits and \nolimits

JeffEverett:
Resolves #35 

Works by altering the `limits` property of the base node while superscripts and subscripts are being considered in the `parseAtom` function.


kevinb7:
@JeffEverett thanks for the pull request.  Nice unit tests.  `\limits` changes how certain things are rendered so you'll have to update buildHTML.js to handle the rendering.  Changes may also be necessary in buildMathML.js to differentiate between the two styles.  Since this is a new rendering you should add a screenshot test (or modify an existing one).  See dockers/Screenshotter/README.md for details.

JeffEverett:
@kevinb7 Here's my understanding: in KaTeX, all operator functions have a `limits` property that  determines whether the superscripts and subscripts are placed to the right of the base node or above and below the base node. Functions like `int` have a default `limits` property of `false` and functions like `sum` have a default `limits` property of `true`. `\limits` and `\nolimits` have the power to alter these defaults and the existing rendering system takes care of the rest. I see no difference between the TeX output of an expression like `\int\limits^2_3 3` and the KaTeX output of the same expression while using my changes. If there are subtleties of the `\limits` and `\nolimits` functions that I'm unaware of, I'd be happy to correct my implementation.

If the above is correct and the new limit-controlling functions add no functionality to the actual renderer, should I still add a screenshot test?

kevinb7:
@JeffEverett just checked out your branch.  The rendering looks good.  I didn't realize that there was already special handling in place for base nodes with a `limits` property set to `true`.  It looks like it was there to handle `\displaystyle\sum_a^b` but in fact handles all `op`s.

We should still have screenshot tests for this in place to prevent future regressions.  The screenshot tests should be simpler to run now that we're using pre-made docker images.

The MathML output should use `<munderover>` instead of `<msubsup>`, but since this is an existing issue (see https://github.com/Khan/KaTeX/issues/198) don't worry about it.

JeffEverett:
@kevinb7 I have added a screenshot test for the limit controls.

kevinb7:
@JeffEverett awesome, thanks for the screenshots and thanks for the pull request.

Issue #273, opened 2015-06-27
render unsupported commands and symbols as red source text


kevinb7:
![screen shot 2015-06-26 at 11 56 31 pm](https://cloud.githubusercontent.com/assets/1044413/8390672/ff416ac0-1c5e-11e5-8715-8412e23c350c.png)

This provides feedback for users as to what going wrong without having to look at the console.  This is especially useful in interactive systems used by content creators.

spicyj:
As an option, presumably.

kevinb7:
You bet.  There could be a color parameter too.

spicyj:
If you're feeling crafty, it could be like

```
handleError: (text) => `\color{red}{${text}}`
```

or something.

kevinb7:
Showing other errors might be useful, but it might be annoying to see stuff like ```Expected '}', got 'EOF'``` at the end of rendering as you're typing.

spicyj:
MathJax seems to just shows your input in a box for errors like that. `$\x{$` displays as:

![image](https://cloud.githubusercontent.com/assets/6820/8766367/1a15b2ba-2dea-11e5-9c2a-944dd7cf7780.png)

kevinb7:
That shouldn't be too difficult.

cben:
Input in a box easy enough with current API (catch error, put text in a span).  (Though a built-in option with default style could be handy.)
But accepting unknown \foo and going on with parsing would be a real advance.
(Of course it has to be optional — if I'm trying KaTeX and falling back to MathJax, I wan't an error for \foo in case MJ does handle it)

PR #317, opened 2015-07-27
Added support for visual depiction of unsupported commands

JeffEverett:
Addresses issue #273.

This patch adds two new options to the parser: `breakOnUnsupportedCmds` and `errorColor`. By default, `breakOnUnsupportedCmds` is `true` and `errorColor` is `#cc0000`. 

When `breakOnUnsupportedCmds` is `true`, the parser and renderer remain unchanged by the patch and unsupported commands in the input expression continue to result in `ParseError`s. When `breakOnUnsupportedCmds` is manually set to `false`, however, the command string will render as textual output in the color given by `errorColor`. This is achieved by the parser generating an `error` node for each unsupported command with a `value` property equal to an array of `textord`s corresponding to the command string. Having the `value` property equal to an array of `textord`s is useful because it allows the renderer to piggyback on the existing rendering capabilities of `text` nodes, which behave similarly.

Jasmine test and screenshotter tests have been added for this patch. In order to successfully create the screenshotter test for this addition, support for changing the settings used in the screenshotter for the two new options have been added. `breakOnUnsupportedCmds` can be set to `false` by setting `doNotBreak` in the query string to a non-empty string and `errorColor` can be set by including it as a name-value pair in the query string.


kevinb7:
@JeffEverett thanks adding this.  I think this will be a huge benefit for anyone using KaTeX, especially for interactive typesetting.  Love the screenshots.  There are a few changes that I'd like to see.  Could you wrap the error node in a color node in the Parser?  That way we won't have to add a "settings" param to everything in buildHTML and buildMathML.  Also, if the "error" node is wrapped in a color node it could just be a "text" node and then we won't need a special "error" group type handler in buildHTML and buildMathML.

JeffEverett:
@kevinb7 Good suggestions. That should make this PR more compact. I'll get on this soon.

JeffEverett:
@kevinb7 I implemented your suggestions. I think the patch is a lot cleaner now.

I also added the two new properties to the readme file.

kevinb7:
Thanks for the pull request.  Can't wait to start using this.

spicyj:
Can we not use abbreviations like "Cmds", especially in public API?

kevinb7:
I think that's a good idea.

JeffEverett:
@spicyj I'll avoid using abbreviations in future contributions. Would you like me to change the option name for this one as well?

spicyj:
@JeffEverett Yes please! I don't love the name `breakOnUnsupportedCommands` because it is pretty long but I don't have any better ideas.

Issue #318, opened 2015-07-27
`\int\limits` in textstyle doesn't correctly produce limits


xymostech:
In KaTeX, `\int\limits_x^y z` produces:

![image](https://cloud.githubusercontent.com/assets/217900/8913817/4c42d788-3450-11e5-8a60-caf5c3195380.png)

whereas in TeX, it produces:

![image](https://cloud.githubusercontent.com/assets/217900/8913858/809f44bc-3450-11e5-88cc-1215edd0f0b2.png)

`\nolimits` and `\limits` was recently introduced in #314

kevinb7:
@xymostech good catch.  @JeffEverett could you have a look into this if you have some time?

PR #319, opened 2015-07-27
Fixed limit controls in textstyle

JeffEverett:
Addresses issue #318.

Sorry, I should have caught this issue. I just didn't realize the limit controls were supposed to work in textstyle (probably because the operators that normally have limits don't in textstyle unless you explicitly specify that they should using `\limits`).

I updated the `LimitControls` screenshot test with the use of a `\limits` function in textstyle.

As an aside, MathML output still does not support limits in both dipslaystyle and textstyle (as documented in #198). I think I will tackle this issue soon to complete KaTeX's limit support.


src/Parser.js
```diff
@@ -263,6 +263,7 @@ Parser.prototype.parseAtom = function(pos, mode) {
             else {
                 var limits = lex.text == "\\limits";
```
xymostech: Can you change this to `===` while you're here? :)

src/Parser.js
```diff
@@ -263,6 +263,7 @@ Parser.prototype.parseAtom = function(pos, mode) {
...
                 base.result.value.limits = limits;
+                base.result.value.alwaysHandleSupSub = true;
```
xymostech: I'm a little sad that we're adding new properties here, but this is fine.

xymostech:
:+1: Thanks for fixing this so quick! Just one little nit, then this looks great!

JeffEverett:
@xymostech No problem. Just changed the `==` to `===`.

Issue #322, opened 2015-08-05
When building katex using webpack an additional loader is needed for json.


blaiprat:
Hi, with the 0.5.0 release the big object from `fontMetrics.js` has been moved to their own JSON file `fontMetricsData.js`. That forces me to add the `json-loader` in my project configuration just for this specific file. 

Not a terrible problem I know, but I was wondering if it can be saved as a javascript file, prepending the `module.exports =` to the file when generating it and saving it with the `js` extension.

kevinb7:
I don't see a problem with doing this.  Would you be able to submit a pull request?

blaiprat:
I've been checking the scripts, I'm not fluent with python, but I'll give it a try.

blaiprat:
Thanks for merging the pull request

PR #323, opened 2015-08-05
Generate fontMetricsData as JavaScript

blaiprat:
As described in issue #322, webpack needs additional configuration to parse JSON.
This pull request includes changing the generator script and `fontMetricsData` converted to javascript. 

(That was easier than expected 😊 )

### DIFF
--- a/metrics/format_json.py
+++ b/metrics/format_json.py
@@ -13,4 +13,4 @@ for font in sorted(data):
       sys.stdout.write(json.dumps(data[font][glyph], sort_keys=True))
       sep = ",\n  "
   sep = "\n},\n"
-sys.stdout.write("\n}}\n");
+sys.stdout.write("\n}};\n");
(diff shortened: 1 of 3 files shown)


metrics/format_json.py
```diff
@@ -4,7 +4,7 @@
 import json
 
 data = json.load(sys.stdin)
-sep = "{\n"
+sep = "module.exports = {\n"
```
xymostech: Looks like you forgot to add the corresponding semicolon at the end of the file, which is why the linter is failing? That should go on line 16, but github isn't letting me add a comment there.

blaiprat:
Ok, added the missing semicolon and now it's passing the checks. Thanks for pointing me to the right direction @xymostech :)

kevinb7:
@blaiprat thanks for the pull request.

PR #324, opened 2015-08-06
Replace screenshots by ones which don't use subpixel rendering

gagern:
While updating #310 I noticed that two screenshots didn't reproduce exactly when I ran `screenshotter.sh` on my system. The differences were almost imperceptible at first, but with the help of GIMP I found out that the screenshots currently included in the master branch use [subpixel rendering](https://en.wikipedia.org/wiki/Subpixel_rendering), while the ones the selenium docker images created for me do not. At large magnifications you'll notice this as the presence resp. absence of a slightly colored boundary along non-horizontal edges.

@JeffEverett, how did you create your screenshots? Did you run the latest `screenshotter.sh` to create them? If you run `screenshotter.sh` on the current master, can you reproduce your results or mine?

If Jeff can reproduce my results instead of his, then I'd ask the project owners to merge this branch so that we all compare against those. If not, then we need a more thourough investigation as to why the same docker image sometimes employs subpixel rendering and sometimes does not.


kevinb7:
@gagern I noticed there were difference between the chrome screenshots that are in the repo and the ones that I generate locally w/o any changes to source code.  I didn't have time to investigate the different and ended up disabling the chrome screenshots because of this.  Maybe it's the same issue as this.

gagern:
@kevinb7 I hadn't noticed 0dc9eed1c42ab111b1ba1d795270e579d4b423cc before, thanks for pointing this out. Can you upload your variants of these screenshots somewhere, e.g. as a separate branch (with no pull request)?

I don't think that this is the same issue, though: subpixel rendering would affect pretty much all of the screenshots, so if you can reproduce at least some screenshots, then you should have the same subpixel behavior. So perhaps we should open a new issue for the Chrome divergence.

Which version of the two Firefox screenshots in question here can you reproduce on that setup with the diverging Chrome results?

kevinb7:
@gagern I've created an issue for it: https://github.com/Khan/KaTeX/issues/325.

gagern:
PR #340 supersedes this one: if you merge that, there is no need to merge this here as well. If you merge this here first, then that, then the history will be clearer but it would mean more work for me.

kevinb7:
Closed in preference of https://github.com/Khan/KaTeX/pull/340.

Issue #296, opened 2015-07-07
How to deal with absolute dimensions


gagern:
In KaTeX, all dimensions are given in `em`, which is relative to the current font size. However, there are some dimensions in the TeX world which are given in `pt` or some similar absolute value. Currently, expressing such dimensions in the DOM is either pretty complicated (manually resetting the size and style up front) or not accurate (if the current font is not the 10pt default).

The prime example which comes to my mind here is the `\nulldelimiterspace=1.2pt` discussed in https://github.com/Khan/KaTeX/pull/135#discussion-diff-32872033 and https://github.com/Khan/KaTeX/issues/249#issuecomment-119094270. If things are to look the way they do in TeX, then we have to make that box always 1.2/ptPerEm times the size of the unsized text style font size.

But how? I can think of several alternatives.
1. Add classes to reset to `textstyle` and `size5`, so that the size is always relative to the default font size. This is what I did in [@7dc8b68](https://github.com/gagern/KaTeX/blob/7dc8b6809239fe9be820e838b271ec274dc0a4cd/src/buildHTML.js#L158-L164).
2. Express the spacing as kerning (using `makeKern`) and then provide some function which takes the current `options` and the size in `pt` and does all the neccessary computation. #135 is using `makeKern`, but so far the resulting size is still relative to the current size and style.
3. Deliberately deviate from LaTeX and declare that all dimensions should scale with the font size.

kevinbarabash:
This has been addressed in #732.

Issue #267, opened 2015-06-23
Add horizontal and vertical line support in environments


xymostech:
Add support for `\hline` and the `|` column specifier for environments. This was mentioned in #47.

btwael:
+1

brunovcosta:
+1

santhosh-tekuri:
+1

PR #330, opened 2015-08-19
Add | column separators to arrays.

xymostech:
This adds the ability to add `|` to a column description and have
vertical separators be added. I added type tags to the column descriptions
and added some logic to handle the separators when building the vertical
lists of the array.

The LimitControls and UnsupportedCmds images also changed, which is worrying... Maybe my computer just doesn't like rendering images the same as everyone else.

Test plan:
 - See the Arrays screenshot looks good.
 - `make test`

texcmp diff: 
arrays

(I'm not at all positive about the 0.2em used as the spacer between the two separators on the right in that image, I just guessed; anyone have an idea where to find a source for that number?)


src/buildHTML.js
```diff
@@ -575,12 +579,48 @@ var groupTypes = {
...
+        for (c = 0, colDescrNum = 0;
+             c < nc || colDescrNum < coldescriptions.length;
+             ++c, ++colDescrNum) {
+            var coldescr = coldescriptions[colDescrNum] || {};
+
+            var lastWasSeparator = false;
+            while (coldescr.type === "separator") {
+                if (lastWasSeparator) {
+                    colsep = makeSpan(["arraycolsep"], []);
+                    colsep.style.width = "0.2em";
```
xymostech: Here's the 0.2em I'm not sure about.

kevinb7: Is there a sigma that's close to 0.2em that we could use?

kevinb7:
Awesome!  Looking forward to being able to augment matrices – reduced row echelon here we come!  I'll have a closer look this evening.

src/buildHTML.js
kevinb7: I found a quote in http://get-software.net/macros/latex/required/tools/array.pdf that might shed some light on the matter:
> In the standard array/tabular implementation this error is not so noticeable
as that version contains negative spacing so that each | takes up no horizontal
space. But since in this package the vertical lines take up their natural
width one sees two lines if two are specified.

I'm sure if we got the "width" metric for `\vert` it would turn out to be close `0.2em`.

kevinb7: Please disregard this, after looking at `.vertical-separator` I can see that `0.2em` would be too thick.

spicyj: Yeah, these look different.

image

kevinb7: I see, this is for the space between the double vertical lines.

src/buildHTML.js
```diff
@@ -575,12 +579,48 @@ var groupTypes = {
...
         var coldescriptions = group.value.cols || [];
         var cols = [];
         var colsep;
-        for (c = 0; c < nc; ++c) {
-            var coldescr = coldescriptions[c] || {};
+        var colDescrNum;
+        for (c = 0, colDescrNum = 0;
+             c < nc || colDescrNum < coldescriptions.length;
+             ++c, ++colDescrNum) {
+            var coldescr = coldescriptions[colDescrNum] || {};
```
kevinb7: camelCase `colDescr` and `colDescriptions`

src/buildHTML.js
```diff
@@ -575,12 +579,48 @@ var groupTypes = {
...
+
+            var lastWasSeparator = false;
+            while (coldescr.type === "separator") {
+                if (lastWasSeparator) {
```
kevinb7: Since we're only going to have separators within this while loop, it'd probably be easier to grok if `lastWasSeparator` was renamed to `firstSeparator` and set it to `true` initially and then check `if (!firstSeparator) { // add a gap }`.

src/environments.js
```diff
@@ -65,14 +65,19 @@ var environmentDefinitions = [
...
         handler: function(pos, mode, envName, colalign, positions) {
             var parser = this;
-            // Currently only supports alignment, no separators like | yet.
             colalign = colalign.value.map ? colalign.value : [colalign];
             var cols = colalign.map(function(node) {
                 var ca = node.value;
                 if ("lcr".indexOf(ca) !== -1) {
                     return {
+                        type: "align",
                         align: ca
```
kevinb7: I forgot how terse some of these variables names were.

test/katex-spec.js
```diff
@@ -1354,7 +1354,18 @@ describe("An array environment", function() {
...
         var parse = getParsed("\\begin{array}r1\\\\20\\end{array}");
         expect(parse[0].type).toBe("array");
-        expect(parse[0].value.cols).toEqual([{align:"r"}]);
+        expect(parse[0].value.cols).toEqual([
+            { type: "align", align: "r" }
+        ]);
+    });
+
+    it("should accept vertical separators", function() {
+        var parse = getParsed("\\begin{array}{c|}\\end{array}");
```
kevinb7: It would be good to have tests that verify that it handles separators at initial position as well as multiple separators in a row.  I know that the screenshot test cover those cases, but the screenshot tests don't get run as often.

src/buildHTML.js
xymostech: Ah ha! I figured it out! (through a lot of guesswork and blind searching through the TeX source).

The expansion of `\array` (which is expanded to when you write \being{array}` eventually calls `\@mkpream` to generate the preamble (which is the analog of the column definitions when using `halign` in plain TeX). In here, it calls `\@testpach`, which looks at the next character, and if said next character is `|`, it sets `\@chclass` to `1`. Then we look at `\@chclass`, and if it is `1` we call `\@classi`, which looks at `\@lastchclass` (presumably the previous `\@chclass`), and if that is also `1`, it adds `\hskip \doublerulesep` before adding the vrule. Observing `\doublerulesep`, we find that it is `2pt`, which ends up being `.2em` with our current `1em=10pt` measurement.

I'll go add this to our font metrics file...

test/katex-spec.js
xymostech: Good idea!

src/buildHTML.js
spicyj: Nice.

kevinb7:
LGTM.  Can you squash this?

gagern:
> The LimitControls and UnsupportedCmds images also changed, which is worrying... Maybe my computer just doesn't like rendering images the same as everyone else.

That's #324. I'd say that when you rebase this to squash it, those two image changes should be kept out since they are unrelated. The `screenshotter.js` modifications are unrelated as well, so personally I'd keep them separate, too. What was the motivation for those, by the way? Had you actually encountered trouble with the window size at some point? Reproducible?

Since the `\doublerulesep` of `2pt` doesn't scale with current style, #296 applies here as well, I think. We should have a test case for that situation, a matrix in scriptscript style with a double vertical line.

Apart from these really minor concerns, nice work! Two boxes to tick in #269 and half of #267 done as well. Looking forward to seeing this included.

xymostech:
@gagern

> Had you actually encountered trouble with the window size at some point? Reproducible?

I did encounter trouble with the window size, but I believe it was being caused by the image being too big! I think I added those lines because it kept looking like it was the previous test that was failing, hence the changes. I'll remove them here, but maybe make a separate diff (It also fixes a typo).

> those two image changes should be kept out since they are unrelated

Can do!

kevinb7:
@xymostech thanks for the pull request.  @gagern thanks for reviewing.

WChargin:
Yay! Thank you!

PR #333, opened 2015-08-22
Add a Gitter chat badge to README.md

gitter-badger:
### Khan/KaTeX now has a Chat Room on Gitter

@kevinb7 has just created a chat room. You can visit it here: [https://gitter.im/Khan/KaTeX](https://gitter.im/Khan/KaTeX?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&content=body_link).

This pull-request adds this badge to your README.md:

[Gitter](https://gitter.im/Khan/KaTeX?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=body_badge)

If my aim is a little off, please [let me know](https://github.com/gitterHQ/readme-badger/issues).

Happy chatting.

PS: [Click here](https://gitter.im/settings/badger/opt-out) if you would prefer not to receive automatic pull-requests from Gitter in future.

PR #334, opened 2015-08-22
Revert "Add a Gitter chat badge to README.md"

kevinb7:
Reverts Khan/KaTeX#333


kevinb7:
Reverting so that we can discuss the merits of gitter vs freenode.

Issue #325, opened 2015-08-08
Chrome screenshots have diverged from a recent commit


kevinb7:
I disabled Chrome screenshots in https://github.com/Khan/KaTeX/commit/0dc9eed1c42ab111b1ba1d795270e579d4b423cc.  If someone else could locally re-enable Chrome screenshots and re-generate them to see if they're seeing differences that would be much appreciated.  I will try to do the same and post images in a branch this weekend.

gagern:
The title is incorrect: at least for me, chrome screenshots have *not* diverged from the *initial* commt, or at least not all of them. However, the *have* diverged from the most recent commit, namely ce99abd1f209b49680910fb9a93a7a45c0025f25. That commit by @xymostech needs some investigation. Taking `Arrays-chrome.png` as an example, I can reproduce the state from b7e1581869ea4f5d20a6b7d7f0b0dc7bb4ae6f64 but not from ce99abd1f209b49680910fb9a93a7a45c0025f25. The PNG files from these two commits are the same in terms of pixel values (i.e. `pngtopam` yields the same result), but not in terms of binary encoding. Surprisingly, the differences are not in metadata, nor in the compression efficiency. The files have exactly the same size, the individual `IDAT` chunks have equal size, but still the pixels are encoded somewhat differently.

Unless @xymostech or someone else has some idea as to *why* the image got encoded differently, my best bet would be addressing the symptom by parsing and then recompressing the PNG file. One way would be using some JavaScript-based PNG encoder/decoder. [pngparse](https://www.npmjs.com/package/pngparse) suggested in the discussion following [this comment](https://github.com/Khan/KaTeX/pull/264#discussion_r34209881) seems to only decode. Ideally, I'd use [optipng](http://optipng.sourceforge.net/) which has the added benefit of reducing the file size. But requiring a specific version of optipng to be installed doesn't feel right. Perhaps I can use emscripten to compile optipng (and its dependency libpng) to JavaScript and provide that as a Node module.

kevinb7:
https://github.com/Khan/KaTeX/commit/90f9e12df3639633911f651ab8c3e20ff9f66081 contains all of the chrome screenshots which differ ever so slightly which is all of them.  This is on the https://github.com/Khan/KaTeX/commits/chrome_screenshot_images branch.

gagern:
[...]

PR #340, opened 2015-08-30
Use jspngopt and pako to create reproducible PNG files for Chrome as well

gagern:
The combination of [jspngopt](https://github.com/gagern/jspngopt) (which I just developed for this task) and [pako](http://nodeca.github.io/pako/) should eliminate possible causes for different PNG encodings, although the core reason for #325 remains unknown.
Pako has poorer compression rates than native libz, but optimization can counter that effect, and actually reduce the size of the screenshots (from 1736 to 1120 kB).

The screenshots for `LimitControls` and `UnsupportedCmds` on Firefox used to exhibit subpixel rendering before, for reasons unknown.  The regenerated versions don't exhibit this.  See #324 for a discussion. `Arrays-chrome.png` has changed as well, since the generation of Chrome screenshots was disabled during the work on #330. Apart from this, all screenshots remain the same with regard to their image content, which I verified by running `optipng` on the directory from these two branches before comparing them.


kevinb7:
Thanks for digging into this and fixing issue.  It's weird that only some images would exhibit subpixel rendering.  Here are some screenshots so that people can the before/after for LimitControls.  The reason why all the images changed is b/c the compression changed.

w/ subpixel rendering
screen shot 2015-08-29 at 6 52 55 pm

w/o subpixel rendering (still antialiased though)
screen shot 2015-08-29 at 6 53 09 pm

kevinb7:
@gagern did you notice an appreciable difference in run times?

gagern:
Perhaps @JeffEverett created those two screenshots using Selenium directly, without the docker, therefore taking shots from his local Firefox setup. This is the most likely explanation I can think of.

This patch here increases the time to take screenshots from 45s to 168s. A factor of two in that is due to the fact that we're now using two different browsers, so the optimization causes another factor two increase in processing time, approximately. I know that jspngopt could be made faster using asynchroneous zlib calls, but since here we're using pako not zlib, that won't help for this use case. What I probably could do is work out the parameter combinations which actually yield the smallest results for our inputs, and then restrict the search space to that.

PR #341, opened 2015-08-30
(Re-)Create some more screenshots in a reproducible way

gagern:
Sorry, I forgot something in #340. Things would be really a lot easier if Travis could check screenshots…

When rebasing for 2e002ff37aea45979a715a0f8015bf60baba4bfb I forgot to re-create the screenshots based on the new parent.  As a consequence, the font testing images from fd2d58fd8011a1ce1931b42ae53bdae8acfeba1f were not updated for Firefox and not even included for Chrome.

We still have the strange issue that Lap can result in one of two possible screenshots, and while the previous commit recreated one of them, this one here recreates the other.


kevinb7:
> Things would be really a lot easier if Travis could check screenshots…

Agreed.  I'm going to be busy moving this weekend and next, but after that I should have some time to tackle this.

Thanks for adding the missing images.

PR #342, opened 2015-09-01
Rename breakOnUnsupportedCmds to throwOnError.

xymostech:
Also, the MathBb-chrome test changed, to what I believe is the correct
result? Not sure why it looked wrong before.

Test plan:
 - `make test`
 - take screenshots, see nothing changed.

@spicyj @kevinb7 @JeffEverett


spicyj:
Not sure on the screenshot, but otherwise good.

kevinb7:
The screenshot is good.  I missed that that screenshot was missing glyphs in https://github.com/Khan/KaTeX/commit/9c2a391ff65a09bec6b82b6b7dad9e67d490cd55.

PR #345, opened 2015-09-10
Split up functions list into calls to defineFunction

gagern:

Having one long array literal to contain the code of all function implementations is problematic.  It makes it difficult to add auxiliary functions or data close to the function inside the list where it is needed.

Now the functions are no longer defined using such a literal, but instead using calls to a `declareFunction` function which receives all the relevant data.  Since each function call is independent from the others, anything can go in between.

I've deliberately submitted this as three distinct commits. The first gets rid of one level of indirection, replacing `functions.funcs` by just `functions`. The second does the actual work, but avoids reindentation, while the third does the reindentation but has no other changes. Hopefully this will make reviewing easier.

### DIFF
--- a/src/Parser.js
+++ b/src/Parser.js
@@ -173,7 +173,7 @@ Parser.prototype.handleInfixNodes = function (body, mode) {
             }
             overIndex = i;
             funcName = node.value.replaceWith;
-            func = functions.funcs[funcName];
+            func = functions[funcName];
         }
     }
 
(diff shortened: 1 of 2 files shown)


kevinb7:
I will have a closer look at this this weekend.  Looking at the combined diff isn't too bad with `?w=1` https://github.com/Khan/KaTeX/pull/345/files?w=1

kevinb7:
It's good to have the re-indent as a separate commit for people look at this use a different git tool.  @gagern thanks for splitting up the commits like this.

kevinb7:
Looks good.

PR #346, opened 2015-09-10
Reformat symbol table

gagern:

Using function calls instead of one big object literal for the symbols makes the notation far more concise and readable.  Having the actual symbol name in the last position helps aligning the preceding columns, making the list easier to read.

Another benefit is that all symbol definitions now pass through a single function, where additional processing (e.g. [for Unicode input](https://github.com/Khan/KaTeX/pull/261)) might take place in a future commit.

### DIFF
--- a/src/symbols.js
+++ b/src/symbols.js
@@ -8,7 +8,7 @@
  * - group (required): the ParseNode group type the symbol should have (i.e.
      "textord", "mathord", etc).
      See https://github.com/Khan/KaTeX/wiki/Examining-TeX#group-types
- * - replace (optional): the character that this symbol or function should be
+ * - replace: the character that this symbol or function should be
  *   replaced with (i.e. "\phi" has a replace value of "\u03d5", the phi
  *   character in the main font).
  *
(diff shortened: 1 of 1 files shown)


kevinb7:
Nothing to squash here.  Okay, I'm going to have look at this one over a couple of sessions b/c I want to make sure nothing got changed accidentally.

gagern:
If you want to automatically verify that essentially nothing changed, you can use code along these lines:

```js
"use strict";

var child_process = require("child_process");

var newSymbols = {rev: "414f9dd248b2822d6f46da529942035dff9960be"};
var oldSymbols = {rev: newSymbols.rev + "^"};
load(oldSymbols);
load(newSymbols);

function load(obj) {
    child_process.execFile("git", ["show", obj.rev + ":src/symbols.js"], step2);
    function step2(err, stdout, stderr) {
        if (err) throw err;
        obj.code = stdout;
        var f = new Function("module", obj.code);
        var m = {exports: {}};
        f(m);
        obj.data = m.exports;
        if (newSymbols.data && oldSymbols.data)
            process.nextTick(compare);
    }
}

function compare() {
    compareMode(oldSymbols.data.text, newSymbols.data.text);
    compareMode(oldSymbols.data.math, newSymbols.data.math);
}

function compareMode(oldDict, newDict) {
    var key;
    for (key in oldDict) {
        if (!(key in newDict)) throw new Error(key);
        compareSymbol(key, oldDict[key], newDict[key]);
    }
    for (key in newDict) {
        if (!(key in oldDict)) throw new Error(key);
    }
}

function compareSymbol(key, oldSym, newSym) {
    if (oldSym.font !== newSym.font) throw new Error(key);
    if (oldSym.group !== newSym.group) throw new Error(key);
    var oldReplace = oldSym.replace;
    if (oldReplace === undefined) {
        if (oldSym.group === "spacing")
            oldReplace = null;
        else
            oldReplace = key;
    }
    if (oldReplace !== newSym.replace) throw new Error(key);
}
```

If you look at the magic around `oldReplace`, you see that there *are* some changes. There are entries where `replace` used to be `undefined` in the old code, but will be equal to `null` or `key` in the new code. The latter is essentially a no-op according to [`makeSymbol`](https://github.com/gagern/KaTeX/blob/414f9dd248b2822d6f46da529942035dff9960be/src/buildCommon.js#L38) and [`makeOrd`](https://github.com/gagern/KaTeX/blob/414f9dd248b2822d6f46da529942035dff9960be/src/buildCommon.js#L120). And spacing has its own code paths in [`buildHTML`](https://github.com/gagern/KaTeX/blob/414f9dd248b2822d6f46da529942035dff9960be/src/buildHTML.js#L669-L686) and [`buildMathML`](https://github.com/gagern/KaTeX/blob/414f9dd248b2822d6f46da529942035dff9960be/src/buildMathML.js#L301-L316) which don't seem to use the `replace` property at all..

kevinbarabash:
@gagern sorry for taking so long to review this.  Thanks for posting code to automatically compare the symbols from the two commits.

PR #347, opened 2015-09-10
Split up environments list into calls to defineEnvironment

gagern:

Having one long array literal to contain the code of all environment implementations is problematic.  It makes it difficult to add auxiliary functions or data close to the function inside the list where it is needed.

Now the functions are no longer defined using such a literal, but instead using calls to a `defineEnvironment` function which receives all the relevant data.  Since each function call is independent from the others, anything can go in between.

I've deliberately submitted this as two distinct commits. The first does the actual work, but avoids reindentation, while the second does the reindentation but has no other changes. Hopefully this will make reviewing easier. Of course I can squash that once you're satisfied.

### DIFF
--- a/src/environments.js
+++ b/src/environments.js
@@ -81,98 +81,98 @@ function defineEnvironment(names, props, handler) {
     }
 }
 
-    // Arrays are part of LaTeX, defined in lttab.dtx so its documentation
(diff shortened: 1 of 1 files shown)


kevinb7:
Hmm... I probably should've had you squash those.  I'll remember for the next one.

PR #348, opened 2015-09-10
Split groupType literals in buildHTML and buildMathML

gagern:

Having long object literals containing the code is problematic. It makes it difficult to add auxiliary functions or data close to the function inside the map where it is needed. Building the map in several steps, repeating the map name at each step, avoids that problem since it makes the definitions independent from one another, so anything can go between them.

I've deliberately submitted this as two distinct commits. The first does the actual work, but avoids reindentation, while the second does the reindentation but has no other changes. Hopefully this will make reviewing easier. Of course I can squash that once you're satisfied.


kevinb7:
Looks good to me.

gagern:
Thanks! One down, three to go…

PR #349, opened 2015-09-10
Make texcmp work again, including KaTeX logo

gagern:
Since all the math font test cases use the KaTeX logo, we need that. I started with the definition of the logo from katex.less, but tweaked that until it gave a good visual match, in particular a very similar logo width, no matter the actual numbers.

KaTeX

With that logo, most tests can be compiled again, with the exception of the one containing illegal functions to test visual error reporting. That one needs to be explicitely disabled.

In particular, we can now for the first time see texcmp output from all the font test cases. Some fonts simply use a lot of unsuitable symbols in TeX, since they only have very limited font support and TeX doesn't fall back the way KaTeX does:
* MathBb: mathbb
* MathCal: mathcal
* MathFrak: mathfrak
* MathScr: mathscr
  That last one looks like a poor match even for the letter A up front. But I guess I'd need to use some other package in LaTeX, not sure which one. Suggestions?

Other fonts have more complete support, though, so we can actually learn something from these:
* MathBf: mathbf
* MathDefaultFonts: mathdefaultfonts
* MathIt: mathit
* MathRm: mathrm
* MathSf: mathsf
* MathTt: mathtt

One thing I notice is that the dotless i is positioned somewhat different in most of these tests. Another thing is that the breve diacritic over the a apparently matches the font in LaTeX: it's italic for italic, bold for bold and teletype-style for teletype. The spacing difference in the italic font looks like it might be an issue, too. Perhaps this has something to do with italic or slant correction, one way or another? Don't know much about those, unfortunately.

Feel free to open separate bug requests to keep track of any such issues you consider relevant, so we can keep track of them even after this merge request here got accepted.

### DIFF
--- a/test/screenshotter/ss_data.yaml
+++ b/test/screenshotter/ss_data.yaml
@@ -52,9 +52,7 @@ Exponents: a^{a^a_a}_{a^a_a}
 FractionTest: \dfrac{a}{b}\frac{a}{b}\tfrac{a}{b}\;-\dfrac12\;1\tfrac12
 Functions: \sin\cos\tan\ln\log
 GreekLetters: \alpha\beta\gamma\omega
-KaTeX:
-    tex: \KaTeX
-    nolatex: There is no LaTeX command for this (yet)
+KaTeX: \KaTeX
 Lap: ab\llap{f}cd\rlap{g}h
 LeftRight: \left( x^2 \right) \left\{ x^{x^{x^{x^x}}} \right.
 LeftRightListStyling: a+\left(x+y\right)-x
@@ -108,6 +106,7 @@ UnsupportedCmds:
     tex: \err\,\frac\fracerr3\,2^\superr_\suberr\,\sqrt\sqrterr
     noThrow: 1
     errorColor: "#dd4c4c"
+    nolatex: deliberately does not compile
 VerticalSpacing:
     pre: potato<br>blah
     tex: x^{\Huge y}z
--- a/test/screenshotter/test.tex
+++ b/test/screenshotter/test.tex
@@ -1,6 +1,8 @@
 \documentclass[10pt]{article}
 
 \usepackage{amsmath,amssymb}
+\usepackage[mathscr]{eucal}
+\usepackage{eufrak}
 \usepackage[papersize={133pt,100pt},margin=0.5pt]{geometry}
 \usepackage{color}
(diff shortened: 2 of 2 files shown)


kevinb7:
I was sure I had the accents matching the font.  It would be nice to have the spacing matching for the dotless i.  I'll open bugs for these.

kevinb7:
Where did the numbers for the KaTeX logo come from?

gagern:
> Where did the numbers for the KaTeX logo come from?

As I wrote in the commit message, I started with the definition of the logo from [`katex.less`](https://github.com/Khan/KaTeX/blob/master/static/katex.less#L328-L346), but tweaked that until it gave a good visual match, in particular a very similar logo width, no matter the actual numbers. So at first I had

```tex
\DeclareRobustCommand{\KaTeX}{\mbox{%
K\kern-.32em%
{\sbox0 T\vbox to\ht0{\hbox%
  {\fontsize{.75em}{1em}\selectfont A}%
\vss}}%
\kern-.23em%
\TeX}}
```

but in that setup, the A was too high, and the K and the A were too close together. The former I fixed by introducing some vertical kerning before the `\hbox`, chosen pretty much arbitrarily but the result looked good. The latter I tweaked manually as well. Now that I think about it, the 0.32em in `katex.less` probably refer to the font size of the A, so that should be 0.32∙0.75em=0.24em. Nevertheless, the 0.26em negative kerning made the left side of the K align better in the texcmp output, so that's what I chose. The 1em line height was arbitrary as well, and shouldn't matter since we're not typesetting more than one letter.

```diff
 \DeclareRobustCommand{\KaTeX}{\mbox{%
-K\kern-.32em%
-{\sbox0 T\vbox to\ht0{\hbox%
+K\kern-.26em%
+{\sbox0 T\vbox to\ht0{\kern.05em\hbox%
   {\fontsize{.75em}{1em}\selectfont A}%
 \vss}}%
 \kern-.23em%
 \TeX}}
```

kevinb7:
> As I wrote in the commit message

Sorry.  I usually only look at "files".  I'll start looking at the individual "commits" too.  Thanks for the detailed explanation.

edemaine:
I had forgotten how `\KaTeX` must be hard-coded in texcmp. Given the recent change in logo, we probably need to change this.

Issue #344, opened 2015-09-09
Parser error with mathbf in superscript


weidendo:
Using current master (Sep 9, 15), I get the following:

KaTeX parse error: Got function '\mathbf' with no arguments as superscript at position 15: \mathbf{G} = e^̲\mathbf{F} =

There is an argument, but it probably is not parsed as part of the superscript?

kevinb7:
@weidendo definitely a bug.  Thanks for reporting this.  I'll have a look at it this evening.

kevinb7:
Looks like the greediness should be set to 2 for this.

gagern:
Is greediness a concept in TeX as well? Or is it simply that any expandable macro gets expanded before the math formula gets assembled, so the `\mathbf{F}` will produce a single group which will then be placed into the exponent as is? If so, should we try to systematically assign greediness based on whether the function in question is an expandable macro or an unexpandable command in TeX?

kevinb7:
> Or is it simply that any expandable macro gets expanded before the math formula gets assembled

I'm not sure if there are an exceptions to that rule, but it seems like that would be reasonable default behavior.

> should we try to systematically assign greediness based on whether the function in question is an expandable macro or an unexpandable command in TeX?

Definitely something worth looking into.

kevinbarabash:
I'm not seeing this bug on master which was just released in https://github.com/Khan/KaTeX/releases.  Please re-open if you can repro this.

PR #352, opened 2015-09-11
Set the greedines of font functions to 2 so that e^\mathbf{x} will parse

kevinbarabash:
Fixes issue https://github.com/Khan/KaTeX/issues/344


xymostech:
LGTM! Looks like this needs to be rebased though.

PR #356, opened 2015-09-15
New calling convention for functions and environments

gagern:
Fixes issue #255.

Mixing the variable number of arguments a function receives from TeX code with the fixed arguments which the parser provides can cause some confusion. After this change, a handler will receive exactly two arguments: one is a context object from which things provided by the parser can be accessed by
name, which allows for simple extensions in the future.  The other is the list of TeX arguments, passed as an array.

If we ever switch to EcmaScript 2015, we might want to use its destructuring features to name the elements of the args array in the function head.  Until then, destructuring that array manually immediately at the beginning of the function seems like a useful convention to easily find the meaning of these arguments.

Since @kevinb7 kindly merged some of the things this dependet upon, I could cherry-pick and adapt this changeset here from #262. Thanks a lot for all those reviews!

### DIFF
--- a/src/Parser.js
+++ b/src/Parser.js
@@ -195,7 +195,8 @@ Parser.prototype.handleInfixNodes = function (body, mode) {
             denomNode = new ParseNode("ordgroup", denomBody, mode);
         }
 
-        var value = func.handler(funcName, numerNode, denomNode);
+        var value = this.callFunction(
+            funcName, [numerNode, denomNode], null);
         return [new ParseNode(value.type, value, mode)];
     } else {
         return body;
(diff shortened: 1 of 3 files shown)


src/Parser.js
```diff
@@ -506,6 +514,18 @@ Parser.prototype.parseFunction = function(pos, mode) {
     }
 };
 
+/**
+ * Call a function handler with a suitable context and arguments.
+ */
+Parser.prototype.callFunction = function(name, data, args, positions) {
```
kevinbarabash: Instead of passing in "data" we could use "name" to do the lookup from the "functions" array.

kevinbarabash: We do this with parseArguments too where we do the lookup before passing in "funcData".

src/environments.js
```diff
@@ -85,8 +88,10 @@ function defineEnvironment(names, props, handler) {
 // is part of the source2e.pdf file of LaTeX2e source documentation.
 defineEnvironment("array", {
     numArgs: 1
-}, function(pos, mode, envName, colalign, positions) {
-    var parser = this;
+}, function(context, args) {
+    var colalign = args[0];
+    var lexer = context.lexer;
+    var positions = context.positions;
```
kevinbarabash: Since `lexer` and `positions` are only used once we may as well use `context.lexer` and `context.positions` directly.

src/functions.js
```diff
@@ -58,14 +58,17 @@ var ParseError = require("./ParseError");
...
  * It is called to handle these functions and their arguments.
- * Its own arguments are:
- *  - func: the text of the function
- *  - [args]: the next arguments are the arguments to the function,
- *            of which there are numArgs of them
+ * It receives two arguments:
+ *  - context contains information and references provided by the parser
+ *  - args is an array of arguments obtained from TeX input
+ * The context contains the following properties:
+ *  - func: the text (i.e. name) of the function, including \
```
kevinbarabash: This should be "funcName" so that it's consistent with the other uses of "func" and "funcName".

kevinbarabash:
@gagern looks good for the most part.  Just a few minor changes and rebasing and this will be good to go.

gagern:
@kevinbarabash included your two suggestions, rebased to current master (resolving one conflict). Ready to merge, I think.

kevinbarabash:
Thanks for making those changes.

PR #359, opened 2015-09-22
Remove 3px border around rules.

xymostech:
Summary: For some reason, adding `border-style: solid` also adds a 3px
border around elements, which means that all of the rules that we
created are 3px too large. This sets the default size to 0 for all the
edges, which makes them correct.

Test plan:
 - See that `a\rule{0em}{0em}b` produces no visible rule.
 - See the new screenshots look reasonable.


spicyj:
:o

PR #361, opened 2015-09-28
Add \gt and \lt

crepererum:
This makes it easier to typeset math, because it avoids using `<` and `>` as special HTML characters.

(CLA was signed)


kevinbarabash:
@crepererum thank you for the pull request.  Although these aren't standard TeX commands, MathJax supports this probably for the same reasons that you've outlined.  I'm okay with merging this, but I'd like to hear @xymostech's thoughts first just in case she has any concerns.

xymostech:
We already talked about these and decided against it here: #149 but now I'm less sure about that decision. Getting a macro expansion of `\lt` into `<` for things like `\left\lt` looks like it works in TeX, so we might try that option once macros come around. Dealing with `<` and `>` inside HTML is a nightmare though, and if someone's using our auto-render plugin then they don't have a chance to do a replacement before running it through KaTeX.

@crepererum How are you using KaTeX? Would doing something like adding an option to the auto-render extension to turn `\lt` into `<` and `\gt` into `>` before sending the input to KaTeX, would that work for you?

crepererum:
My current setup is to type some text using Markdown which sometimes contains math, rendered by the auto-renderer. Escaping backslashes in Markdown is not a problem, but dealing with `<`/`>` is a nightmare, so the proposed PR would simplify this a lot.

Sure you could extend the Markdown parser to handle this step, but for my app (Ghost) there is currently no way to do this without running your own fork (ref https://github.com/TryGhost/Ghost/issues/1724).

So the auto-render extension might work, but I think that the HTML-parser of the browser will already be a problem. What I can do instead is using `&lt;`/`&gt;` when typesetting the math, because the auto-renderer uses `textContent` to get the data (https://github.com/Khan/KaTeX/blob/master/contrib/auto-render/auto-render.js#L55) which will included the transformed `<`/`>`.

If the `&lt;`/`&gt;`-workaround becomes somehow official, we might want to add this to the documentation of the auto-renderer so that users don't have to search through the source code to find this implementation detail.

kevinbarabash:
@crepererum could you verify that `\lt` and `\gt` work as delimiters, e.g. that `\left\lt\frac{1}{x}\right\gt` produces the same rendering as `\left<\frac{1}{x}\right>`?

crepererum:
@kevinbarabash done, added tests for that.

kevinbarabash:
@crepererum looks good except one of the lines in the test is too long.  Can you split it so it's on two lines so that we stay within the 80 column limit?

crepererum:
@kevinbarabash sry, fixed now.

kevinbarabash:
@crepererum awesome, thanks!

PR #362, opened 2015-09-30
Use an array of values instead of object literals to save space in fontMetricsData.js

kevinbarabash:
I ran `make test` and did a quick smoke test by running `make serve` and typesetting a few things.


kevinbarabash:
The size of the uncompressed files goes from ~120KB to 50KB.  After zipping them the files were 9KB and 7KB respectively.

metrics/format_json.py
```diff
@@ -6,11 +6,17 @@
...
-  for glyph in sorted(data[font], key=int):
-      sys.stdout.write(sep + json.dumps(glyph) + ": ")
-      sys.stdout.write(json.dumps(data[font][glyph], sort_keys=True))
-      sep = ",\n  "
-  sep = "\n},\n"
-sys.stdout.write("\n}};\n");
+    sys.stdout.write(sep + json.dumps(font))
+    sep = ": {\n  "
+    for glyph in sorted(data[font], key=int):
+        sys.stdout.write(sep + json.dumps(glyph) + ": ")
```
spicyj: nit: you shouldn't need to quote these

metrics/format_json.py
```diff
@@ -6,11 +6,17 @@
...
-  sep = ": {\n  "
-  for glyph in sorted(data[font], key=int):
-      sys.stdout.write(sep + json.dumps(glyph) + ": ")
-      sys.stdout.write(json.dumps(data[font][glyph], sort_keys=True))
-      sep = ",\n  "
-  sep = "\n},\n"
-sys.stdout.write("\n}};\n");
+    sys.stdout.write(sep + json.dumps(font))
+    sep = ": {\n  "
+    for glyph in sorted(data[font], key=int):
```
spicyj: I don't think this key does anything?

spicyj:
This seems fine, I think. If you wanted to reduce the size a lot further, you could change

```js
module.exports = {
"AMS-Regular": {
  "65": [0, 0.68889, 0, 0],
  "66": [0, 0.68889, 0, 0],
  "67": [0, 0.68889, 0, 0],
  "68": [0, 0.68889, 0, 0],
  "69": [0, 0.68889, 0, 0],
  "70": [0, 0.68889, 0, 0],
  "71": [0, 0.68889, 0, 0],
  "72": [0, 0.68889, 0, 0],
  "73": [0, 0.68889, 0, 0],
  "74": [0.16667, 0.68889, 0, 0],
```

to something like

```js
var metrics = [
  [0, 0.68889, 0, 0],
  [0.16667, 0.68889, 0, 0],
];

module.exports = {
"AMS-Regular": {
  "65": metrics[0],
  "66": metrics[0],
  "67": metrics[0],
  "68": metrics[0],
  "69": metrics[0],
  "70": metrics[0],
  "71": metrics[0],
  "72": metrics[0],
  "73": metrics[0],
  "74": metrics[1],
```

(or even just export the metrics array and the `{65: 0, 66: 0, ..., 73: 0, 74: 1, ...}` separately and have getCharacterMetrics reconstitute it). Might not make a huge difference post-gzip, though.

metrics/format_json.py
kevinbarabash: I tried it without and it uses string comparison to order the entries so "100" becomes before "2".

kevinbarabash:
> Might not make a huge difference post-gzip, though.

I was thinking the same.

metrics/format_json.py
spicyj: Does that matter?

kevinbarabash: Probably doesn't make a huge different, but I could see how maybe more people are using to looking things in numeric order when dealing with numbers so if someone needs to verify something about a particular glyph this ordering might make it easier.

spicyj: Oops, sorry! I totally misread this. This makes sense.

Issue #266, opened 2015-06-23
More internal state for parser


gagern:
I'm thinking of modifying the parser so it has more internal state and less method arguments. In particular, the `pos` and `mode` arguments to the methods should be replaced by `this.currentPos` and `this.mode`. I'd also introduce `this.peek` or `this.nextToken` to describe the lexer result for `currentPos`. We could have a new method `consumeToken` or similar to update the current position to the end of the next token, then lex the next token at that new position.

Here are the benefits I envision:
* **More efficiency.** We'd avoid repeatedly lexing at the same position, which is what we currently do as @xymostech confirmed in https://github.com/Khan/KaTeX/pull/246#issuecomment-112604386.
* **Easier communication with environments.** Environments, which call to the parser to read their body, could simply say “hey, parser, give me the next expression”, instead of “hey, parser, give me the expression at the location you gave me and in the mode you gave me”.
* **Ensure linear operation.** If the normal mode of operation is reading one token after the other, then I believe (although this is a bit of a gut feeling) that it might help in implementing macros (#250). We could just make sure that the next few tokens we read are those coming from the definition of the macro. Doing so might mean modifying the `consumeToken` method or changing the lexer, and if you change the lexer, then the `pos` alone isn't sufficient to locate a token any more. I feel (still a gut feeling) that it also might help in annotating each parse node with a begin and end position, which in turn will help getting rid of the `positions` array for function invocations (#255).

Considering possible downsides, I believe the most likely downside is that changing the mode now means ensuring that you change it back again as well. Since the only place where we change the mode is in `parseSpecialGroup`, I think it should be possible to get that right in that one location.

Would you be interested in a modification along these lines? If so, I'll write that, but I'd like to hear whether it has a chance of getting merged before I start on this endeavor.

kevinb7:
> Environments, which call to the parser to read their body, could simply say “hey, parser, give me the next expression”, instead of “hey, parser, give me the expression at the location you gave me and in the mode you gave me”.

I'm all for simplicity and "get the next expression" is definitely simpler.

> Considering possible downsides, I believe the most likely downside is that changing the mode now means ensuring that you change it back again as well.  Since the only place where we change the mode is in parseSpecialGroup, I think it should be possible to get that right in that one location.

That would be my concern as well.  Good to know we only do that in one location.

> If the normal mode of operation is reading one token after the other, then I believe (although this is a bit of a gut feeling) that it might help in implementing macros (#250).

[...]

PR #363, opened 2015-10-02
Make mode part of the internal state of the parser

gagern:
This is the first step towards #266.

### DIFF
--- a/src/Parser.js
+++ b/src/Parser.js
@@ -20,10 +20,11 @@ var ParseError = require("./ParseError");
  * individual tokens are needed at a position, the lexer is called to pull out a
  * token, which is then used.
  *
- * The main functions also take a mode that the parser is currently in
- * (currently "math" or "text"), which denotes whether the current environment
- * is a math-y one or a text-y one (e.g. inside \text). Currently, this serves
- * to limit the functions which can be used in text mode.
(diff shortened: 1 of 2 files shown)


src/Parser.js
```diff
@@ -638,21 +641,25 @@ Parser.prototype.parseSpecialGroup = function(pos, mode, outerMode, optional) {
...
             false);
-    } else if (mode === "text") {
+    } else if (innerMode === "text") {
         // text mode is special because it should ignore the whitespace before
         // it
         var whitespace = this.lexer.lex(pos, "whitespace");
         pos = whitespace.position;
     }
 
+    this.mode = innerMode;
```
kevinbarabash: The reason for doing the assignment down here is that `innerMode` can only be `text` or `math` by the time we get here.  Is that correct?  This could probably use a comment.

gagern: Right. Will augment my commit.

gagern:
@kevinbarabash Augmented my commit to include [that comment](https://github.com/gagern/KaTeX/blob/cabc08598bfe7a8f12ec9a60ebd8f3e0dabddc5a/src/Parser.js#L654-L655). Any other concerns?

kevinbarabash:
@gagern that was it.  Thanks for the refactoring.

PR #366, opened 2015-10-04
Add a build step to generate extended metrics that additionally contain glyph widths

kevinbarabash:
The ability to use pre-determined character widths will benefit alternative layout engines such as gagern's canvas layout engine.  I would also like to experiment would using CSS transforms to absolutely position each glyph.  This diff adds a new make rule, make extended_metrics, which generates metrics that also containing glyph widths.

### DIFF
--- a/src/fontMetrics.js
+++ b/src/fontMetrics.js
@@ -123,7 +123,10 @@ var metricMap = require("./fontMetricsData");
 
 /**
  * This function is a convenience function for looking up information in the
- * metricMap table. It takes a character as a string, and a style
+ * metricMap table. It takes a character as a string, and a style.
+ * 
+ * Note: the `width` property may be undefined if fontMetricsData.js wasn't
+ * built using `Make extended_metrics`.
  */
 var getCharacterMetrics = function(character, style) {
     var metrics = metricMap[style][character.charCodeAt(0)];
@@ -132,7 +135,8 @@ var getCharacterMetrics = function(character, style) {
             depth: metrics[0],
             height: metrics[1],
             italic: metrics[2],
-            skew: metrics[3]
+            skew: metrics[3],
+            width: metrics[4]
         };
     }
 };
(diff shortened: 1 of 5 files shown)


kevinbarabash:
@gagern I was wondering what your thoughts were about this pull request with respect to https://github.com/Khan/KaTeX/pull/251.

gagern:
Imo #251 doesn't need widths since it can measure them on the fly.

But there are applications where having a width would be great. One thing that comes to my mind would be big horizontal constructs, like underbraces. Horizontal boxes with various amounts of stretchable glue distributed over them would be another. Automatic wrapping of equations comes to mind as well, even though TeX doesn't do that. And of course there is always output to other formats with no interactive font measurement facilities, like SVG.

What has me somewhat worried is the fact that different typesetting engines might make different choices with respect to kerning and so on, and over longer stretches of glyphs, these might amount to considerable errors. So mixing automatic and manual horizontal positioning is perhaps a bad idea. And doing it all manually would mean not only having all the widths, but all the kerning pairs as well, wouldn't it?

kevinbarabash:
I was thinking it might improve the performance if you don't have to measure the width; sounds like that isn't a concern.

With respect to differences in kerning pairs, I was thinking of absolutely positioning everything.  Not sure if that's a good idea or not.

src/fontMetrics.js
```diff
@@ -125,15 +125,25 @@ var metricMap = require("./fontMetricsData");
...
-            skew: metrics[3]
-        };
+        if (includeWidth) {
+            return {
+                depth: metrics[0],
+                height: metrics[1],
+                italic: metrics[2],
+                skew: metrics[3],
+                width: metrics[4]
+            };
```
xymostech: Why not just always return width, with the caveat that without the extended metrics it will be `undefined`?

xymostech:
This LGTM! Remember to squash before pushing though :)

src/fontMetrics.js
kevinbarabash: Sounds good.

Issue #103, opened 2014-09-22
Auto generate supported function list


utensil:
Awesome project!

Currently KaTeX doesn't support all the symbols that LaTeX supports and people use. The supported function list is important for people who are evaluating KaTeX.

This list is:
- currently manually maintained and it seems to be not up-to-date, e.g. symbols added by #72 are not listed
- not mentioned in [Contributing to KaTeX](https://github.com/Khan/KaTeX/blob/master/CONTRIBUTING.md) as a required step, so people would forget to update this Wiki page after adding symbols or functions.

It's nice if KaTeX auto generates supported function list based on [symbols.js](https://github.com/Khan/KaTeX/blob/master/src/symbols.js) and [functions.js](https://github.com/Khan/KaTeX/blob/master/src/functions.js), post it on Wiki or just on http://khan.github.io/KaTeX/ .

Two extensions to this idea:
- to provide an API to output all supported symbols and functions
- it's better if all symbols and functions comes with a example, ready to be rendered by this supported function list, preferably grouped and ordered.

utensil:
I've implemented/handcrafted a proof-of-concept preview page showing [Symbols and functions in KaTeX](http://utensil-site.github.io/available-in-katex/) (not all, until I also add functions from https://github.com/Khan/KaTeX/wiki/Function-Support-in-KaTeX ) .

So one migrating from MathJax can see how are the MathJax-supported symbols and function supported by KaTeX. This could also give potential contributors a sense about what is still missing(thus needs to be added).

kevinbarabash:
@utensil this is really cool.  Could you add a link to it in CONTRIBUTING.md somewhere where it makes sense?

utensil:
@kevinbarabash Thanks. I would create a PR to add a link to it and improve it.

kevinbarabash:
Sounds good.

utensil:
PR submitted as #369 , also added the contents in https://github.com/Khan/KaTeX/wiki/Function-Support-in-KaTeX . In future, there should be a better way to keep them up to date(though I believe the symbols/functions supported by MathJax could keep it sufficiently complete for a while).

PR #369, opened 2015-10-16
Add a link to symbol/function support preview page

utensil:
As suggested by @kevinbarabash in https://github.com/Khan/KaTeX/issues/103#issuecomment-148118466 , I've added a link to a symbol/function support preview page that I automatically generated and handcrafted in CONTRIBUTING.md.

This PR is not meant to be directly merged(though that's also possible), but opens a discussion about how to integrate this preview page into contributing guide.

The preview page currently lists all symbols and function in KaTeX wiki and MathJax document, and attempts to render them, and falls into 3 kind of results:

* successfully rendered by just the symbol itself, presented in a green cell
* can't be rendered by itself(because it's function), but can be successfully rendered by providing an proper example,  presented in a blue cell, along with the tooltip of the original TeX code of the example, that's the main handcraft part
* failed to render  (because KaTeX doesn't support them in the release yet), presented in a red cell

The page is quite self-explanatory and also includes statistics of the three kind of results.

The question left to be discussed are:

* is the preview page proper/good enough to be added to CONTRIBUTING.md ?
* should the preivew page be somewhere more official, like in a rep of KaTeX organization? (the code is not necessary good enough, because it's just a prototype, though it works nicely)
* should the introduction of the 3 places that contributors to check for symbol/function support improve wording and organization?

P.S. I've already signed the SLA.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -10,10 +10,13 @@ If you'd like to contribute, try contributing new symbols or functions that
 KaTeX doesn't currently support. The wiki has a page which lists [all of the
 supported
 functions](https://github.com/Khan/KaTeX/wiki/Function-Support-in-KaTeX) as 
-well as a page that describes how to [examine TeX commands and where to find
-rules](https://github.com/Khan/KaTeX/wiki/Examining-TeX) which can be quite
-useful when adding new commands. You
-can check there to see if we don't support a function you like, or try your
+well as a page that describes how to [examine TeX commands and where to find 
+rules](https://github.com/Khan/KaTeX/wiki/Examining-TeX) which can be quite 
+useful when adding new commands. There's also a user-contributed [preview page]
+(http://utensil-site.github.io/available-in-katex/)
+showing how KaTeX would render a series of symbols/functions (including the ones
+MathJax listed in their documentation and the extra ones supported by KaTeX). You
+can check them to see if we don't support a function you like, or try your
 function in the interactive demo at
 [http://khan.github.io/KaTeX/](http://khan.github.io/KaTeX/).
 


CONTRIBUTING.md
```diff
@@ -10,10 +10,13 @@ If you'd like to contribute, try contributing new symbols or functions that
...
 functions](https://github.com/Khan/KaTeX/wiki/Function-Support-in-KaTeX) as 
-well as a page that describes how to [examine TeX commands and where to find
-rules](https://github.com/Khan/KaTeX/wiki/Examining-TeX) which can be quite
-useful when adding new commands. You
-can check there to see if we don't support a function you like, or try your
+well as a page that describes how to [examine TeX commands and where to find 
+rules](https://github.com/Khan/KaTeX/wiki/Examining-TeX) which can be quite 
+useful when adding new commands. There's also a user-contributed [preview page]
+(http://utensil-site.github.io/available-in-katex/)
+showing how would KaTex render a series of symbols/functions(including the ones
```
kevinbarabash: "showing how KaTeX would"... and "symbols/functions (including"

CONTRIBUTING.md
```diff
@@ -10,10 +10,13 @@ If you'd like to contribute, try contributing new symbols or functions that
...
+MathJax listed in their documentation and the extra ones supported by KaTeX). You
+can check them to see if we don't support a function you like, or try your
```
kevinbarabash: "we don't support" => "we support"

kevinbarabash:
@utensil use the katex.min.js and katex.min.css served off of khan.github.io is an interesting way to stay current.   I think this should be okay unless we change were these files are located which seems unlikely, but it is a potential failure point.  I think having the code for this in your repo is fine for the time being.  If there are changes we want to make we'll fork it at that point.

One thought about the layout... instead of having three colors, it might be better to have three columns: one for the command, one for the code use to generate the rendering, and then the rendering itself (or a a failure message like you have now).

Thanks for signing the CLA.

CONTRIBUTING.md
utensil: I get the "showing how KaTeX would" part, but what does "symbols/functions (including" part mean? I've checked it letter by letter, failed to find the difference.

CONTRIBUTING.md
utensil: This is the original wording, unchanged. I guess it means that the contributor would look for what's missing and add them. Should I change the meaning to the other way around(the users looking for what's already at hand)?

utensil:
> @utensil use the katex.min.js and katex.min.css served off of khan.github.io is an interesting way to stay current. I think this should be okay unless we change were these files are located which seems unlikely, but it is a potential failure point.

I totally agree. I started with the CDN version in README.md, but it seems to be not loading as fast as the github.io version, so I went with it for the time being.

> I think having the code for this in your repo is fine for the time being. If there are changes we want to make we'll fork it at that point.

OK, that's point of open source and Github, no central point of failure, :wink:  And if it becomes somehow official later, there would always be a better way to stay current.

> One thought about the layout... instead of having three colors, it might be better to have three columns: one for the command, one for the code use to generate the rendering, and then the rendering itself (or a a failure message like you have now).

Yeah. The thought crossed my mind too. I didn't use it because it's a little verbose for the symbols and I planed to do something similar when I find a proper way to separate symbols and functions(and meta characters).

CONTRIBUTING.md
kevinbarabash: Sorry, I didn't look at the original wording.  We can leave it as is.

CONTRIBUTING.md
kevinbarabash: There should be a space between before the "(".

utensil: Oh I see.

utensil:
Fixed. Should I squash my commits in this branch into one and `git push --force`? Not quite did that before.

And travis-ci got this weird error: `can't resolve github.com`, seems to be temporary.

kevinbarabash:
> Should I squash my commits in this branch into one and `git push --force`?

Yes please.

utensil:
Is it OK now?

kevinbarabash:
@utensil looks good.  Thanks for the pull request.

PR #371, opened 2015-10-19
Readme - set recent version (0.5.1) for CDN links

stared:
AS in the topic. I had some issues with testing arrays on JSFiddle. Only later I discovered that the CDN links were to 0.3.0 - a version to supporting environments.


kevinbarabash:
@stared cool.  Thanks for the PR.

PR #381, opened 2015-11-06
Check screenshots on Travis using Docker and Selenium

### LINKED ISSUE #300: Automate screenshotter on travis-ci
https://github.com/moul/travis-docker allows you to run docker on travis-ci.  We can use it to automate the screenshotter so that it runs on every commit and fails the build if there differences in the images.

Depends on: https://github.com/Khan/KaTeX/pull/264

gagern:
Thanks to the [docker service provided by Travis CI](http://docs.travis-ci.com/user/docker/), we should be able to download and use the Selenium docker images in order to run our screenshotter and check whether all the screenshots match the images from the repository.

This fixes #300.

The Selenium bindings for JavaScript [appear to use some ES6 syntax](https://travis-ci.org/gagern/KaTeX/jobs/89607826#L549-L562) in their [latest version (2.48.2)](https://travis-ci.org/gagern/KaTeX/jobs/89607826#L264), so I bumped the node version to support this. If you disagree, we'll have to restrict the version of `selenium-webdriver` some more.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,5 +9,4 @@ before_script:
   - docker pull selenium/standalone-chrome:2.46.0
 script:
   - npm test
-  - dockers/Screenshotter/screenshotter.sh
-  - git diff --name-only --exit-code
+  - dockers/Screenshotter/screenshotter.sh --verify
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,13 @@
 language: node_js
 node_js:
-- "0.11"
-- "0.10"
+  - stable
+sudo: required
+services:
+  - docker
+before_script:
+  - docker pull selenium/standalone-firefox:2.46.0
+  - docker pull selenium/standalone-chrome:2.46.0
+script:
+  - npm test
+  - dockers/Screenshotter/screenshotter.sh
+  - git diff --name-only --exit-code
--- a/dockers/Screenshotter/screenshotter.js
+++ b/dockers/Screenshotter/screenshotter.js
@@ -63,6 +63,10 @@ var opts = require("nomnom")
         abbr: "x",
         help: "Comma-separated list of test cases to exclude"
     })
+    .option("verify", {
+        flag: true,
+        help: "Check whether screenshot matches current file content"
+    })
     .parse();
 
 var listOfCases;
@@ -204,7 +208,9 @@ function buildDriver() {
         builder.usingServer(seleniumURL);
     }
     driver = builder.build();
-    setSize(targetW, targetH);
+    driver.manage().timeouts().setScriptTimeout(3000).then(function() {
+        setSize(targetW, targetH);
+    });
 }
 
 //////////////////////////////////////////////////////////////////////
@@ -243,6 +249,9 @@ function imageDimensions(img) {
 
 var countdown = listOfCases.length;
 
+var exitStatus = 0;
+var listOfFailed = [];
+
 function takeScreenshots() {
     listOfCases.forEach(takeScreenshot);
 }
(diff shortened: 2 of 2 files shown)


gagern:
There seem to be some sporadic issues with some of the test cases. While [the Travis run on this pull request here](https://travis-ci.org/Khan/KaTeX/builds/89611573) went through ok, [the one for my latest push](https://travis-ci.org/gagern/KaTeX/builds/89610799) failed, even though the two trees should be identical. The reported differences are

```
test/screenshotter/images/MathDefaultFonts-chrome.png
test/screenshotter/images/MathSf-chrome.png
```

so I guess it might have something to do with font loading. Possible ideas how to address this:

1. Introduce some slight unconditional [`sleep`](http://seleniumhq.github.io/selenium/docs/api/javascript/class_webdriver_WebDriver.html#sleep) between `get` and ` takeScreenshot`.
2. Do a client-side wait in the scope of the page using [`executeAsyncScript`](http://seleniumhq.github.io/selenium/docs/api/javascript/class_webdriver_WebDriver.html#executeAsyncScript) to make sure the wait doesn't get parallelized with the page load.
3. Take more than one snapshot if the outcome doesn't match the expectation.
4. Run some asynchroneous script to check that all fonts have been loaded. This might hide some accidential dependency between webfontloader and KaTeX, so I'd rather not do this.

Right now 3 sounds the most reasonable to me.

gagern:
OK, this looks better now. I've implemented the retry approach, and I'm already seeing it in action in the test output, so this works as intended. Is it OK to have this in two commits, or should I squash it into one?

kevinbarabash:
I left a couple of comments in the second commit.  Other than that it looks good.  I feel like using async/await for some of these scripts would make this much more readable.  That's maybe something to consider for the future.

kevinbarabash:
@gagern thanks for automating screenshot.  I love that it retries, should make it much more robust.

gagern:
I stayed with promises because the Web Driver uses them, and I didn't want to introduce yet another async paradigm. But it might be worth considering, particularly since recent selenium seems to be depending on ES6-capable node versions in any case.

gagern:
@kevinbarabash: Updated my second commit. Factored that wait into a separate method, added some comments and also adjusted the special case handling for the `Lap` example.

kevinbarabash:
@gagern really happy to see this addition.  Thanks for the pull request.

gagern:
The approach seems to be somewhat fragile: https://travis-ci.org/Khan/KaTeX/builds/90289211 failed on me, and I don't see why yet. I haven't observed such hangs in local test runs. The last lines of the build log are:

```
Starting container for selenium/standalone-chrome:2.46.0
Container eb6b12b103e2 started, creating screenshots...
Selenium driver at http://172.17.42.1:32769/wd/hub
KaTeX URL is http://172.17.42.1:55029/

No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.

The build has been terminated
```

Looking at [the script](https://github.com/Khan/KaTeX/blob/21a26b807cb1c3b411257e9eddd581a7fe48b2b6/dockers/Screenshotter/screenshotter.js#L179), we know that the HTTP server serving KaTeX started up correctly, and that we haven't received any snapshot yet. `tryConnect` might loop for up to 50 × 0.2s = 10s so I'd assume that we got a successful connection to the selenium server, at least at the TCP level. So presumably the script got hung someplace waiting for some feedback from Selenium. This might be hard to debug.

kevinbarabash:
This is unfortunate.  Thankfully travis-ci let's you re-run a build.  Hopefully this won't happen too often.

kevinbarabash:
So I re-ran the the failed build that you linked to and got an actual error this time, see https://travis-ci.org/Khan/KaTeX/builds/90289211.

gagern:
I [augmented the script](https://github.com/gagern/KaTeX/commits/debugSelenium) with some more output, and tried to reproduce the problem with that. So far without success in reproducing either the hang or the exception you got. And that exception doesn't look like something under our control either.

gagern:
```
SessionNotCreatedError: session not created exception
from chrome not reachable
  (Session info: chrome=44.0.2403.125)
  (Driver info: chromedriver=2.16.333243 (0bfa1d3575fc1044244f21ddb82bf870944ef961),platform=Linux 3.19.0-30-generic x86_64)
```

`chrome not reachable` might indicate an error in the communication between the selenium server and the chrome application. I've just [asked about this on Stack Overflow](https://stackoverflow.com/q/33672274/1468366), perhaps we can get some idea as to what's happening here. We could also try updating our [selenium images](https://hub.docker.com/u/selenium/); we're using 2.46.0 while the latest is 2.48.2.

kevinbarabash:
Updating the selenium images is worth a try.  Thanks for looking into this.

PR #385, opened 2015-11-09
Added unit tests for error messages

### LINKED ISSUE #297: Systematic testing of error messages
We have a lot of error messages a.k.a. `ParseError` exceptions all over the place, and I feel like many of them don't receive any testing at all. Would it make sense to systematically add unit tests for all of them, ensuring that the right message gets generated, with the right error position indication? Should that go into `katex-spec.js` or some other file instead?

Come to think of it, such a systematic testing of rare code paths might well be the first step towards some code coverage control, e.g. using [istanbul](https://www.npmjs.com/package/istanbul). But that's a different issue, so let's stick to exceptions for now.

gagern:
This is an attempt to actually exercise all the code paths which can lead to a `ParserError` exception (from malformed user input, without tinkering with any KaTeX internals or exploiting a KaTeX bug).  It documents the current state of affairs, without changing any error messages.  Comments indicate future work, particularly with respect to the position often associated with these error messages.

This should fix #297.

### DIFF
--- a/test/errors-spec.js
+++ b/test/errors-spec.js
@@ -1,0 +1,345 @@
+/* global beforeEach: false */
+/* global jasmine: false */
+/* global expect: false */
+/* global it: false */
+/* global describe: false */
+
+var parseTree = require("../src/parseTree");
+var Settings = require("../src/Settings");
+
+var defaultSettings = new Settings({});
+
+beforeEach(function() {
+    jasmine.addMatchers({
+        toFailWithParseError: function(util, customEqualityTesters) {
+            var prefix = "KaTeX parse error: ";
+            return {
+                compare: function(actual, expected) {
+                    try {
+                        parseTree(actual, defaultSettings);
+                        return {
+                            pass: false,
+                            message: "'" + actual + "' parsed without error"
+                        };
+                    } catch (e) {
+                        if (expected === undefined) {
+                            return {
+                                pass: true,
+                                message: "'" + actual + "' parsed with error"
+                            };
+                        }
+                        var msg = e.message;
+                        var exp = prefix + expected;
+                        if (msg === exp) {
+                            return {
+                                pass: true,
+                                message: "'" + actual + "'" +
+                                    " parsed with error '" + expected + "'"
+                            };
+                        } else if (msg.slice(0, 19) === prefix) {
+                            return {
+                                pass: false,
+                                message: "'" + actual + "'" +
+                                    " parsed with error '" + msg.slice(19) +
+                                    "' but expected '" + expected + "'"
+                            };
+                        } else {
+                            return {
+                                pass: false,
+                                message: "'" + actual + "'" +
+                                    " caused error '" + msg +
+                                    "' but expected '" + exp + "'"
+                            };
+                        }
+                    }
+                }
+            };
+        }
+    });
+});
+
+describe("Parser:", function() {
+
+    describe("#handleInfixNodes", function() {
+        // TODO: The position information here is broken, should be fixed.
+        it("rejects repeated infix operators", function() {
+            expect("1\\over 2\\over 3").toFailWithParseError(
+                   "only one infix operator per group at position -1: " +
+                   "1\\over 2\\over ");
+        });
+        it("rejects conflicting infix operators", function() {
+            expect("1\\over 2\\choose 3").toFailWithParseError(
+                   "only one infix operator per group at position -1: " +
+                   "1\\over 2\\choos");
+        });
+    });
+
+    describe("#handleSupSubscript", function() {
+        it("rejects ^ at end of group", function() {
+            expect("{1^}").toFailWithParseError(
+                   "Expected group after '^' at position 3: {1^̲}");
+        });
+        it("rejects _ at end of input", function() {
+            expect("1_").toFailWithParseError(
+                   "Expected group after '_' at position 2: 1_̲");
+        });
+        it("rejects \\sqrt as argument to ^", function() {
+            expect("1^\\sqrt{2}").toFailWithParseError(
+                   "Got function '\\sqrt' with no arguments as superscript" +
+                   " at position 2: 1^̲\\sqrt{2}");
+        });
+    });
+
+    describe("#parseAtom", function() {
+        // TODO: The positions in the following error messages appear to be
+        // off by one, i.e. they should be one character later.
+        it("rejects \\limits without operator", function() {
+            expect("\\alpha\\limits\\omega").toFailWithParseError(
+                   "Limit controls must follow a math operator" +
+                   " at position 6: \\alpha̲\\limits\\omega");
+        });
+        it("rejects \\limits at the beginning of the input", function() {
+            expect("\\limits\\omega").toFailWithParseError(
+                   "Limit controls must follow a math operator" +
+                   " at position 0: ̲\\limits\\omega");
+        });
+        it("rejects double superscripts", function() {
+            expect("1^2^3").toFailWithParseError(
+                   "Double superscript at position 3: 1^2̲^3");
+            expect("1^{2+3}_4^5").toFailWithParseError(
+                   "Double superscript at position 9: 1^{2+3}_4̲^5");
+        });
+        it("rejects double subscripts", function() {
+            expect("1_2_3").toFailWithParseError(
+                   "Double subscript at position 3: 1_2̲_3");
+            expect("1_{2+3}^4_5").toFailWithParseError(
+                   "Double subscript at position 9: 1_{2+3}^4̲_5");
+        });
+    });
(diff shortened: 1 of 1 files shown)


kevinbarabash:
Cool.  I have a look at this later today.

xymostech:
Awesome! Also, thanks for splitting these out into a separate test file. `katex-spec.js` should probably be split up some time as well.

Makefile
```diff
@@ -76,7 +76,7 @@ serve:
 	node server.js
 
 test:
-	./node_modules/.bin/jasmine-node test/katex-spec.js
+	./node_modules/.bin/jasmine-node test/*-spec.js
```
xymostech: I usually do something like `find test/ | grep -- '-spec.js' | xargs ./node_modules/.bin/jasmine-node` for this, so if we ever split things up enough that they need directories then this Just Works. This is fine for now, though!

test/errors-spec.js
```diff
@@ -0,0 +1,339 @@
...
+/* global describe: false */
+
+var parseTree = require("../src/parseTree");
+var Settings = require("../src/Settings");
+
+var defaultSettings = new Settings({});
+
+beforeEach(function() {
+    jasmine.addMatchers({
+        toParseWithError: function(util, customEqualityTesters) {
```
xymostech: I think I would call this `toFailWithError`, since the other tests use a `toParse` matcher that checks that the parse succeeded.

gagern: Hmmm. I wonder whether we one day might have errors other than parse errors. If so, having the word `Parse` in that name would be a useful distinction. I guess the most correct and unambiguous form would be `toFailWithParseError`. I'll go for that.

Makefile
gagern: That adds a dependency on findutils. Should be no problem on \*nix, but I'm less sure about what one can reasonably expect on Windows. And passing `test/*/*-spec.js` isn't a problem either, once we have an extra level. Even if we had to choose the recursive approach, I'd prefer a single command instead of three: `find test/ -name \*-spec.js -exec ./node_modules/.bin/jasmine-node '{}' +`. But let's worry about these if we ever need to.

What I found confusing during my tests was that a spec file with broken syntax apparently prevents jasmin from processing that file, but doesn't result in a non-zero exit code. So a commit might accidentially break a whole test file and a glance at the travis status wouldn't tell us about this. Will investigate and file a bug there if I can confirm this.

gagern: That jasmine issue can indeed be reproduced, see https://github.com/mhevery/jasmine-node/issues/379#issuecomment-155170552. It seems many projects I've seen prefer mocha over jasmine, which would handle this situation more elegantly. And jasmine-node doesn't seem to have a lot of activity lately. In fact zero commits to the main branches in the past year. If you feel like switching, let me know.

xymostech: Switching to mocha sounds fine to me! It might take some work to convert our matchers, but sounds like it'll be worth it.

gagern: Looking at [some assertion libraries supported by mocha](http://mochajs.org/#assertions), it seems to me as if [chai](http://chaijs.com/) could model the current assertion syntax most closely. So we'd have to replace the matcher implementation, but could leave the tests in place. On the other hand, [unexpected](http://unexpected.js.org/) sounds like it could express our semantics quite well, and in an even more readable and flexible fashion. I'm leaning slightly towards the latter. But in any case I'd like to see this PR here landed first.

test/errors-spec.js
```diff
@@ -0,0 +1,345 @@
...
+    });
+
+});
+
+describe("Parser.expect calls:", function() {
+
+    describe("#parseInput expecting EOF", function() {
+        it("complains about extra }", function() {
+            expect("{1+2}}").toFailWithParseError(
+                   "Expected 'EOF', got '}' at position 6: {1+2}}̲");
```
kevinbarabash: It seems like the position on this one is off by one as well.  Could you add a TODO here?

test/errors-spec.js
```diff
@@ -0,0 +1,345 @@
...
+        });
+        it("complains about extra \\end", function() {
+            expect("begin{matrix}x\\end{matrix}").toFailWithParseError(
```
kevinbarabash: It might be better to leave the `begin{matrix}` out completely.  I was initially confused why it didn't start with `\\`.

test/errors-spec.js
```diff
@@ -0,0 +1,345 @@
...
+    });
+
+    describe("#_innerLexSize", function() {
+        it("reject size without unit", function() {
+            expect("\\rule{0}{2em}").toFailWithParseError(
+                   "Invalid size at position 6: \\rule{̲0}{2em}");
+        });
+        it("reject size with bogus unit", function() {
+            expect("\\rule{1au}{2em}").toFailWithParseError(
+                   "Invalid unit: 'au' at position 6: \\rule{̲1au}{2em}");
```
kevinbarabash: `au` nice :)

test/errors-spec.js
```diff
@@ -0,0 +1,345 @@
...
+                                    "' but expected '" + expected + "'"
+                            };
+                        } else {
+                            return {
+                                pass: false,
+                                message: "'" + actual + "'" +
+                                    " caused error '" + msg +
+                                    "' but expected '" + exp + "'"
+                            };
+                        }
```
kevinbarabash: I like how provide different output depending on whether it's a parse error or not.

test/errors-spec.js
```diff
@@ -0,0 +1,345 @@
...
+            expect("\\alpha\\limits\\omega").toFailWithParseError(
+                   "Limit controls must follow a math operator" +
+                   " at position 6: \\alpha̲\\limits\\omega");
+        });
+        it("rejects \\limits at the beginning of the input", function() {
+            expect("\\limits\\omega").toFailWithParseError(
+                   "Limit controls must follow a math operator" +
+                   " at position 0: ̲\\limits\\omega");
+        });
+        it("rejects double superscripts", function() {
```
kevinbarabash: "rejects ambiguous superscripts"

test/errors-spec.js
```diff
@@ -0,0 +1,345 @@
...
+            expect("1^2^3").toFailWithParseError(
+                   "Double superscript at position 3: 1^2̲^3");
+            expect("1^{2+3}_4^5").toFailWithParseError(
+                   "Double superscript at position 9: 1^{2+3}_4̲^5");
+        });
+        it("rejects double subscripts", function() {
```
kevinbarabash: "rejects ambigiuous subscripts"

kevinbarabash:
@gagern a couple of minor issues, but the whole this is a very nice test suite.

test/errors-spec.js
gagern: How do you figure that? We got an unexpected `}`, and that is the character indicated by the position and the underline. Looks good to me.

test/errors-spec.js
gagern: To me, “ambiguous” is less specific than “double”. And since the error message literally mentions double superscripts, I see no harm but great benefit in sticking to the wording of the error message I'm testing here.

test/errors-spec.js
gagern: I guess I felt like including some “real world” error scenarios. Sure, I can change that.

gagern:
@kevinbarabash I addressed one of your concerns, and commented on the others.

Makefile
xymostech: Chai has an [`expect` assertion style](http://chaijs.com/guide/styles/#expect), which looks more usable than unexpected and matches other assert libraries better. I think switching to mocha+chai sounds good, I'm not so sure about using expect vs assert, but I'd be willing to try it!

gagern: As per #387, it's only jasmine-node which is broken and unmaintained, not jasmine itself. Running tests in browser might be easier with jasmine as well. So the need to switch doesn't seem as urgent now as it did at first.

test/errors-spec.js
kevinbarabash: Let's leave it as is.  The thing is that you have can have double subscripts if you use groups.  In a separate diff we could always look at changing the error message, but since nobody's reported being confused by it, let's just leave it.

test/errors-spec.js
kevinbarabash: Shouldn't the underline be under the second `}` which is at position `5`, not the `"`at position `6` which to me indicates that it's reached the end of input.

kevinbarabash: After copying and pasting this I realize that the `_` is in the right place even though github (or Chrome) is rendering it in the wrong place.

kevinbarabash:
@gagern okay, this is good to go, can you rebase this?

gagern:
Rebased.

PR #386, opened 2015-11-10
Avoid re-lexing, move position to internal state

gagern:
Instead of passing around the current position as an argument, we now have a parser property called `pos` to keep track of that.  Instead of repeatedly re-lexing at the current position we now have a property called `nextToken` which contains the token beginning at the current position.  We may need to
re-lex if we switch mode.  Since the position is kept in the parser state, we don't need to return it from parsing methods, which obsoletes the `ParseResult` class.

This file contains two hacks (search for `TODO` in the diff) aimed at keeping error messages exactly as they were. Once #385 gets merged, we can improve these locations and adjust the error message expectations accordingly. Right now, this PR won't conflict with #385 but doesn't depend on it either. In the long run, we might even go one step further and switch from reporting an error *position* to reporting an error *token* or *group*. We could underline these ad a whole.

This is the last step in addressing #266. It prepares the way for implementing macros.

### DIFF
--- a/src/Parser.js
+++ b/src/Parser.js
@@ -34,11 +34,9 @@ var ParseError = require("./ParseError");
  * There are also extra `.handle...` functions, which pull out some reused
  * functionality into self-contained functions.
  *
- * The earlier functions return `ParseResult`s, which contain a ParseNode and a
- * new position.
- *
+ * The earlier functions return ParseNodes.
  * The later functions (which are called deeper in the parse) sometimes return
- * ParseFuncOrArgument, which contain a ParseResult as well as some data about
+ * ParseFuncOrArgument, which contain a ParseNode as well as some data about
  * whether the parsed object is a function which is missing some arguments, or a
  * standalone object which can be used as an argument to another function.
  */
@@ -54,11 +52,10 @@ function Parser(input, settings) {
 }
 
 var ParseNode = parseData.ParseNode;
-var ParseResult = parseData.ParseResult;
 
 /**
  * An initial function (without its arguments), or an argument to a function.
- * The `result` argument should be a ParseResult.
+ * The `result` argument should be a ParseNode.
  */
 function ParseFuncOrArgument(result, isFunction) {
     this.result = result;
@@ -69,14 +66,29 @@ function ParseFuncOrArgument(result, isFunction) {
 /**
  * Checks a result to make sure it has the right type, and throws an
  * appropriate error otherwise.
+ *
+ * @param {boolean=} consume whether to consume the expected token,
+ *                           defaults to true
  */
-Parser.prototype.expect = function(result, text) {
-    if (result.text !== text) {
+Parser.prototype.expect = function(text, consume) {
+    if (this.nextToken.text !== text) {
         throw new ParseError(
-            "Expected '" + text + "', got '" + result.text + "'",
-            this.lexer, result.position
+            "Expected '" + text + "', got '" + this.nextToken.text + "'",
+            this.lexer, this.nextToken.position
         );
     }
+    if (consume !== false) {
+        this.consume();
+    }
+};
+
+/**
+ * Considers the current look ahead token as consumed,
+ * and fetches the one after that as the new look ahead.
+ */
+Parser.prototype.consume = function() {
+    this.pos = this.nextToken.position;
+    this.nextToken = this.lexer.lex(this.pos, this.mode);
 };
 
 /**
@@ -84,21 +96,23 @@ Parser.prototype.expect = function(result, text) {
  *
  * @return {?Array.<ParseNode>}
  */
-Parser.prototype.parse = function(input) {
+Parser.prototype.parse = function() {
     // Try to parse the input
     this.mode = "math";
-    var parse = this.parseInput(0);
-    return parse.result;
+    this.pos = 0;
+    this.nextToken = this.lexer.lex(this.pos, this.mode);
+    var parse = this.parseInput();
+    return parse;
 };
 
 /**
  * Parses an entire input tree.
  */
-Parser.prototype.parseInput = function(pos) {
+Parser.prototype.parseInput = function() {
     // Parse an expression
-    var expression = this.parseExpression(pos, false);
+    var expression = this.parseExpression(false);
     // If we succeeded, make sure there's an EOF at the end
-    this.expect(expression.peek, "EOF");
+    this.expect("EOF", false);
     return expression;
 };
 
@@ -114,25 +128,25 @@ var endOfExpression = ["}", "\\end", "\\right", "&", "\\\\", "\\cr"];
  * @param {?string} breakOnToken The token that the expression should end with,
  *                  or `null` if something else should end the expression.
  *
- * @return {ParseResult}
+ * @return {ParseNode}
  */
-Parser.prototype.parseExpression = function(pos, breakOnInfix, breakOnToken) {
+Parser.prototype.parseExpression = function(breakOnInfix, breakOnToken) {
     var body = [];
-    var lex = null;
     // Keep adding atoms to the body until we can't parse any more atoms (either
     // we reached the end, a }, or a \right)
     while (true) {
-        lex = this.lexer.lex(pos, this.mode);
+        var lex = this.nextToken;
+        var pos = this.pos;
         if (endOfExpression.indexOf(lex.text) !== -1) {
             break;
         }
         if (breakOnToken && lex.text === breakOnToken) {
             break;
         }
-        var atom = this.parseAtom(pos);
+        var atom = this.parseAtom();
         if (!atom) {
             if (!this.settings.throwOnError && lex.text[0] === "\\") {
-                var errorNode = this.handleUnsupportedCmd(lex.text);
+                var errorNode = this.handleUnsupportedCmd();
                 body.push(errorNode);
 
                 pos = lex.position;
@@ -141,15 +155,15 @@ Parser.prototype.parseExpression = function(pos, breakOnInfix, breakOnToken) {
 
             break;
         }
-        if (breakOnInfix && atom.result.type === "infix") {
+        if (breakOnInfix && atom.type === "infix") {
+            // rewind so we can parse the infix atom again
+            this.pos = pos;
+            this.nextToken = lex;
             break;
         }
-        body.push(atom.result);
-        pos = atom.position;
+        body.push(atom);
     }
-    var res = new ParseResult(this.handleInfixNodes(body), pos);
-    res.peek = lex;
-    return res;
+    return this.handleInfixNodes(body);
 };
 
 /**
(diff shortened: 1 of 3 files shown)


kevinbarabash:
Thanks for reminding me about this diff.  I will try to look at it this evening.  Sorry about the delay.

gagern:
I'm glad you don't mind being reminded this way. :grinning:

src/Parser.js
```diff
@@ -70,13 +69,21 @@ function ParseFuncOrArgument(result, isFunction) {
  * Checks a result to make sure it has the right type, and throws an
  * appropriate error otherwise.
  */
-Parser.prototype.expect = function(result, text) {
-    if (result.text !== text) {
+Parser.prototype.expect = function(text, consume) {
```
kevinbarabash: Could update the docstring to include a description of `consume`?

src/Parser.js
```diff
@@ -70,13 +69,21 @@ function ParseFuncOrArgument(result, isFunction) {
...
+            "Expected '" + text + "', got '" + this.nextToken.text + "'",
+            this.lexer, this.nextToken.position
         );
     }
+    if (consume !== false) {
+        this.consume();
+    }
+};
+
+Parser.prototype.consume = function() {
```
kevinbarabash: Could you add brief description of this function?

src/Parser.js
```diff
@@ -87,18 +94,20 @@ Parser.prototype.expect = function(result, text) {
 Parser.prototype.parse = function(input) {
```
kevinbarabash: Could you get rid of `input`?

src/Parser.js
```diff
@@ -141,15 +150,15 @@ Parser.prototype.parseExpression = function(pos, breakOnInfix, breakOnToken) {
...
             break;
         }
-        body.push(atom.result);
-        pos = atom.position;
+        body.push(atom);
     }
-    var res = new ParseResult(this.handleInfixNodes(body), pos);
-    res.peek = lex;
-    return res;
+    return this.handleInfixNodes(body);
```
kevinbarabash: Looking at this code again, I realize that I could've done this in a better way.

src/Parser.js
```diff
@@ -211,31 +220,30 @@ var SUPSUB_GREEDINESS = 1;
...
-        if (!this.settings.throwOnError && lex.text[0] === "\\") {
-            return new ParseResult(
-                this.handleUnsupportedCmd(lex.text),
-                lex.position);
+        if (!this.settings.throwOnError && this.nextToken.text[0] === "\\") {
+            return this.handleUnsupportedCmd();
         } else {
             throw new ParseError(
-                "Expected group after '" + symbol + "'", this.lexer, pos);
+                "Expected group after '" + symbol + "'", this.lexer, this.pos);
```
kevinbarabash: Could you change `this.pos` to `symPos`?

src/parseData.js
```diff
@@ -7,17 +7,7 @@ function ParseNode(type, value, mode) {
...
 }
 
-/**
- * A result and final position returned by the `.parse...` functions.
- * 
- */
-function ParseResult(result, newPosition, peek) {
-    this.result = result;
-    this.position = newPosition;
-}
```
kevinbarabash: Yay!

src/Parser.js
```diff
@@ -278,76 +288,62 @@ Parser.prototype.handleSupSubscript = function(pos, symbol, name) {
  *
  * @return {?ParseResult}
```
kevinbarabash: {?ParseNode}

src/Parser.js
```diff
@@ -392,52 +386,47 @@ var styleFuncs = [
...
- * @return {?ParseResult}
+ * @return {?ParseNode}
  */
-Parser.prototype.parseImplicitGroup = function(pos) {
-    var start = this.parseSymbol(pos);
+Parser.prototype.parseImplicitGroup = function() {
+    var start = this.parseSymbol();
 
-    if (!start || !start.result) {
+    if (start === null) {
```
kevinbarabash: Can we use `start == null` instead?

src/Parser.js
```diff
@@ -488,12 +474,14 @@ Parser.prototype.parseImplicitGroup = function(pos) {
  *
  * @return {?ParseResult}
  */
-Parser.prototype.parseFunction = function(pos) {
-    var baseGroup = this.parseGroup(pos);
+Parser.prototype.parseFunction = function(baseGroup) {
+    if (!baseGroup) {
+        baseGroup = this.parseGroup();
+    }
```
kevinbarabash: It feels a little weird having this as an optional parameter, but I don't have a good suggestion of how to make it less weird.  Could you update the jsdocs to reflect that `baseGroup` is optional?

src/Parser.js
```diff
@@ -534,77 +518,73 @@ Parser.prototype.callFunction = function(name, args, positions) {
...
- * @param {Array} args  list of arguments to which new ones will be pushed
- * @return the position after all arguments have been parsed
+ * @return the array of arguments, with the list of positions as last element
  */
-Parser.prototype.parseArguments = function(pos, func, funcData, args) {
+Parser.prototype.parseArguments = function(func, funcData) {
     var totalArgs = funcData.numArgs + funcData.numOptionalArgs;
     if (totalArgs === 0) {
-        return pos;
+        return [[this.pos]];
```
kevinbarabash: If the return is an array of arguments with the last element being an array of positions, I would expect that we should be returning `[[]]` here.

kevinbarabash: I hope in the future that the args can contain position data so that we don't need this extra element in the array.

src/Parser.js
```diff
@@ -624,43 +604,46 @@ Parser.prototype.parseSpecialGroup = function(pos, innerMode, optional) {
...
+        this.mode = outerMode;
         var data;
         if (innerMode === "color") {
             data = inner.text;
         } else {
             data = inner.data;
         }
-        var closeBrace = this.lexer.lex(inner.position, outerMode);
-        this.expect(closeBrace, optional ? "]" : "}");
+        this.consume();
```
kevinbarabash: Could you add a comment about what this is consuming?

src/Parser.js
```diff
@@ -624,43 +604,46 @@ Parser.prototype.parseSpecialGroup = function(pos, innerMode, optional) {
...
-                closeBrace.position),
+            new ParseNode(innerMode, data, outerMode),
             false);
     } else if (innerMode === "text") {
         // text mode is special because it should ignore the whitespace before
         // it
-        var whitespace = this.lexer.lex(pos, "whitespace");
-        pos = whitespace.position;
+        var whitespace = this.lexer.lex(this.pos, "whitespace");
+        this.pos = whitespace.position;
```
kevinbarabash: Could probably make `_innerLex` always ignore leading whitespace and then we wouldn't need to this special casing.  This is something that should probably go in a separate commit or PR.

src/Parser.js
```diff
@@ -670,23 +653,21 @@ Parser.prototype.parseSpecialGroup = function(pos, innerMode, optional) {
  *
  * @return {?ParseFuncOrArgument}
  */
-Parser.prototype.parseGroup = function(pos) {
-    var start = this.lexer.lex(pos, this.mode);
+Parser.prototype.parseGroup = function() {
+    var start = this.nextToken;
     // Try to parse an open brace
     if (start.text === "{") {
```
kevinbarabash: Could just test against `this.nextToken.text` directly instead of introducing `start`

src/Parser.js
```diff
@@ -695,19 +676,17 @@ Parser.prototype.parseGroup = function(pos) {
  *
  * @return {?ParseFuncOrArgument}
  */
-Parser.prototype.parseOptionalGroup = function(pos) {
-    var start = this.lexer.lex(pos, this.mode);
+Parser.prototype.parseOptionalGroup = function() {
+    var start = this.nextToken;
     // Try to parse an open bracket
     if (start.text === "[") {
```
kevinbarabash: Same comment as in `parseGroup`

kevinbarabash:
Looks pretty good.  I'll do another pass of the weekend just to make sure I'm not missing anything.  I'm glad we have the tests that we do otherwise I'd be much less confident about accepting pull requests that make changes like this one.

src/Parser.js
gagern: Why? Most projects I work with prefer `===` over `==` for its cleaner semantics and its better performance.

src/Parser.js
gagern: No, `positions` is [defined](https://github.com/gagern/KaTeX/blob/21a26b807cb1c3b411257e9eddd581a7fe48b2b6/src/functions.js#L68-L69) as

> the positions in the overall string of the function and the arguments.

So even without arguments, the position of the function itself is required. This is in line with how `positions` is [initialized](https://github.com/gagern/KaTeX/blob/21a26b807cb1c3b411257e9eddd581a7fe48b2b6/src/Parser.js#L548) in the case of a non-zero argument count. And in any case, this is preserving existing functionality, we shouldn't change anything here as part of the new position handling.

Moving positions into e.g. parse nodes and the likes will be greatly simplified with all these changes here, I think.

src/Parser.js
gagern: What's your definition of “leading” here? Right now I don't see how this could be possible while still getting correct spacing inside `\text{…}` constructs.

gagern:
I adressed most of your concerns; those I didn't comment on. Also fixed one off-by-one error in one error message position. If this looks fine to you, let me know and I'll squash and rebase it.

src/Parser.js
kevinbarabash: Sorry I confused whitespace before `\text{}` with whitespace within `\text{}` before non-whitespace text within `\text{}`.

src/Parser.js
kevinbarabash: I forgot about that, thanks.

src/Parser.js
kevinbarabash: At KA we prefer using `==` (or `!=`) when comparing against `null` because it will produce the same results if `start` is `undefined`.  In this case it's pretty easy to verify that `start` is either a `ParseNode` or `null` so it's not a big deal, but in other cases it may not be easy to make this determination.

gagern: OK, I changed this occurrence. Using `==` for `null` only is certainly better than always using it. For a moment I feared I might have to change many other strict comparisons as well.

kevinbarabash:
@gagern the changes look good for the most part.  I made a couple of comments on the most recent commit.  After those are address please squash and rebase.

gagern:
Fixed and squashed. The docstring for the `consume` parameter now reads like this:

```
 * @param {boolean=} consume whether to consume the expected token,
 *                           defaults to true
```

kevinbarabash:
Looks good.  Thanks for the quick response.

PR #387, opened 2015-11-10
Switch from jasmine-node to jasmine itself

gagern:
[Jasmine](http://jasmine.github.io/) [supports node](http://jasmine.github.io/2.3/node.html) these days, so there is no longer a need to use a [separate](https://github.com/mhevery/jasmine-node) (and unmaintained) package to provide such bindings. Particular since that package is buggy, as discussed in https://github.com/Khan/KaTeX/pull/385#discussion_r44322060.

Making the switch exposed several misuses of the [`toMatch`](http://jasmine.github.io/2.3/introduction.html#section-Included_Matchers) assertion in the existing specification.  Most of them were converted to `toEqual`, since `toMatch` is only for matching against regular expressions.


test/jasmine.json
```diff
@@ -0,0 +1,10 @@
+{
+  "spec_dir": ".",
+  "spec_files": [
+    "test/**/*[sS]pec.js",
+    "contrib/**/*[sS]pec.js"
+  ],
+  "helpers": [
+    "helpers/**/*.js"
+  ]
+}
```
kevinbarabash: This is really nice.

gagern:
I've got a commit here to get rid of the `test/jasmine` directory in our repository, using a devDependency on `jasmine-core` instead. I also modified the server to include `test/*-spec.js` in its browserified spec file, so that the tests from #385 will get included in http://0.0.0.0:7936/test/test.html as well. Do you want me to include that in this PR here, or file another request once this here gets merged?

xymostech:
I'd say do that in separate pull requests. :)

gagern:
OK, I'll wait for this one here to land. By the way, the issue reported by travis appears to be spurious. Hopefully #388 will avoid them in the future.

kevinbarabash:
All of the unit tests passed so I think we're okay.

gagern:
More importantly, all of the unit tests were actually executed.

PR #388, opened 2015-11-12
Update Selenium Docker images from 2.46.0 to 2.48.2

gagern:
Update Selenium as per https://github.com/Khan/KaTeX/pull/381#issuecomment-156146605 in the hope of reducing the chances of spurious CI errors. The Chrome snapshots have changed some at the anti-aliased fringes, no serious changes though.


kevinbarabash:
It seems like things have been more stable lately with tests.  I'm going to leave this open for a bit and if we start seeing issues again we can apply it then.

kevinbarabash:
If there's another reason to upgrade let me know and I can merge it sooner.

gagern:
I can reproduce the FF window size problems here. Adding a bit of debug output, I get

```
Requested 1024 x 768 but got 1008 x 889
Requested 1040 x 647 but got 1024 x 849
Requested 1040 x 566 but got 1024 x 849
Requested 1040 x 485 but got 1024 x 849
Requested 1040 x 404 but got 1024 x 849
Requested 1040 x 323 but got 1024 x 849
…/KaTeX/node_modules/selenium-webdriver/lib/goog/async/nexttick.js:41
  goog.global.setTimeout(function() { throw exception; }, 0);
                                      ^

Error: Failed to set window size correctly.
```

The [build history](https://hub.docker.com/r/selenium/standalone-firefox/builds/) suggests that there may have been modifications to this image since I filed this pull request here. Unfortunately I can't see the digests of the images we used back then in the build log. Perhaps we should add `docker images --no-trunc` to the `before_script` section of `.travis.yml`.

I just got myself the [standalone-firefox-debug image](https://hub.docker.com/r/selenium/standalone-firefox-debug/) and had a look at the operations using VNC. The window becommes pretty small, in line with the figures on the left of the above quotation. But apparently the screenshot is still larger, probably as large as the underlying HTML page, not just the visible area. And why is that underlying HTML page so large? Because it's some page displayed by Mozilla to welcome users after upgrading Firefox. So we need to suppress that, then we should be good to go.

PR #389, opened 2015-11-12
Drop local copy of jasmine, multiple spec files in browser tests

gagern:
Instead of having our own copy of jasmine in the repository, we use
jasmine-core as an npm dependency and load it from there.  That reduces the
size of the repository and helps keeping up to date.  We're not using the
transitive dependency on jasmine-core via jasmine, since the jasmine package
might change its dependency any day (although unlikely).

The katex-spec.js shipped from the server now includes all
`test/*[Ss]pec.js` (as matched via glob) so that additional spec files can
be created and will automatically get included in the browser-side test
suite.  The contrib specs are not included at this point.

Visit http://0.0.0.0:7936/test/test.html while running server.js to see this
in action and verify the lack of failures.


kevinbarabash:
Thanks for the PR.  I like how requesting katex-spec.js returns all of the test suites.

Issue #254, opened 2015-06-18
need support for euro symbol


Schaan:
Hi, I couldn't find a way to type euro symbol like this `€`. With a european keyboard-layout this can be typed directly, but it can't be rendered within KaTeX, even if I put this € in a `\text`. It can be rendered in mathjax without problem.

I must say I'm not sure either if euro would exist in the world for a long time in the future... But I really type this a lot in my notes at the time.. I hope someone could help on this.

henzestplus:
if someone could add the eurosign in one of the font files, I can add the eurosign, hopefully according to \eurosym with \euro and \EUR{}

henzestplus:
Just for the heads up, I managed to add the Euro Sign to the regular fonts, and have \euro working in KaTeX, now I am working on \EU{}.

kevinbarabash:
@henzestplus which .tfm files did you use?

henzestplus:
What pull request is this about? The one that got closed because there should be a euro sign in the fonts, or the one that got closed because there is no euro sign in the fonts? The first I just copied from existing TeX fonts where I guessed the fonts where based on. The other just the ones included. It does not exist in them as I stated before, and it shows a "missing metrics" message, but it shows the sign.

kevinbarabash:
I had a look at the pull requests and realized what a silly question that was to ask.  Sorry about that.

There's some info https://github.com/Khan/KaTeX/issues/15 on how to go about updating the fonts.  I'm going to be looking into later this week to see if I can add ß, æ, Æ, œ, Œ, ø, Ø to our fonts.  I'll see if there's a way to add the euro symbol as well.  Once we get the symbol into the font the rest should be pretty straight forward.

henzestplus:
no problem. the code part is ready, so when yo got them in, let me know when I can do a pull request again.

kevinbarabash:
I'm going to close this in favor of #1107 which covers missing monetary indicators in general.

PR #394, opened 2015-11-19
added euro sign with \euro

henzestplus:
issue #254 is addressed here. Currently it' s just the \euro symbol as described in http://www.theiling.de/eurosym.html. I am willing to create a \EUR command, but since we don't really need it, I leave this for now. Maybe someone else can pick this up and continue my work.


spicyj:
Why did the font files change/how did you generate them?

henzestplus:
The euro sign was not embedded in any of the font files, since KaTeX uses them for display it only 'knows' what's in the font files. I used fontForge to,  basically,  copy the euro sign from another font (where this one was, presumably,  based on), and an online converter to get the other formats as well.

spicyj:
The generation of the fonts is completely automated using these scripts:

https://github.com/Khan/KaTeX/tree/master/dockers/MathJaxFonts

We don't change them by hand. I am not sure if there is any euro symbol in the fonts (I find it hard to believe that there isn't) but we can't take this pull request in its current form.

henzestplus:
There isn't any in the result files located in static/fonts. If generated through that, how can I ensure it will be in there? Do I have to add something to a script?

Issue #286, opened 2015-07-03
Provide access to all symbols


gagern:
All symbols which are included in the fonts should be accessible by some TeX command. I wrote [some code](https://gist.github.com/gagern/bef7cae7d19cef32b575) to compare the list of symbols (from `symbols.js`) against the fonts (as parsed by [TTX](https://github.com/behdad/fonttools/), resulting in [these files](https://github.com/gagern/KaTeX/commit/86030f7ca43ea3966c5027dafdf44d2ad5c86f03)). Here are my findings:

* [ ] pund sterling sign *£* (in Main_Italic font only)
* [ ] degree sign ° (see https://github.com/gagern/KaTeX/commit/07131ecac9ef88f6be9a6282e4ecb2db77d16243)
* [ ] various combining or modifier diacritics: acute, grave, tilde, dot, diaresis, macron, breve, ring, solidus, caron
* [ ] circumflex ^ and tilde ~ stand-alone
* [ ] em-dash — and en-dash – (will require changes to the lexer, I guess)
* [ ] typographic quotation marks “” and ‘’
* [ ] some misassociated typographic quotes at `\u07e2` and `\u07e3` in the Typewriter font
* [ ] division slash ∕? Not sure we need that, would be interesting to know how that came to be in the font, i.e. whether TeX is using this
* [ ] Maltese cross ✠ in AMS

I ignored some findings found by the script which sounded like something already covered by KaTeX, probably due to some other font using the semantically same symbol with a different code point. I also aggregated similar stuff quite freely, and in the process might have missed some important distinctions. I also ignored stuff which I know is already in progress, the dotless i and j in particular. So the above list is not guaranteed to be complete. Feel free to have a look at the [raw output](https://gist.github.com/gagern/bef7cae7d19cef32b575#file-result-txt) yourself and see whether you can spot anything I missed.

kevinb7:
Nice detective work!

I'm surprised we don't support ```\degree``` yet.  I think the low hanging fruit here is ```\pounds```, ```\degree``` and ```\maltese```.  Could you create a pull request for these?

In terms of the combining marks we support the math mode commands but not the text mode, see http://tex.stackexchange.com/tags/accents/info.  We could probably have a separate issue for accent support in text mode.  From the raw output it looks like the combining marks are being used by KaTeX_Main, but not the other fonts.  I'll update the test cases in https://github.com/Khan/KaTeX/pull/132 to include accents.

[...]

PR #395, opened 2015-11-21
Stplus pullrequest

henzestplus:
- when executing make build, I had an error due to a script in the dockers/texcmp/texcmp.js. this fixes the problem.
- added the euro the right way (apparently I made a mistake in symbols, letting me think there was no euro sign, couldn't find it reviewing the font files in my fontviewer on kubuntu 14.04 though, and there are no metrics, so that error pops in the console, adding it to the python script crashes said script) \euro and  \€ is possible (#254)
- added the ... symbol as \dots.

### DIFF
--- a/dockers/texcmp/texcmp.js
+++ b/dockers/texcmp/texcmp.js
@@ -6,7 +6,7 @@ var path = require("path");
 var Q = require("q"); // To debug, pass Q_DEBUG=1 in the environment
 var pngparse = require("pngparse");
 var fft = require("ndarray-fft");
-var ndarray = require("ndarray-fft/node_modules/ndarray");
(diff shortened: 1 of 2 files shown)


gagern:
There is no € in the fonts. See #286 for things for which we *have* symbol but no way to enter them.

The `\dots` symbol is tricky, since it sometimes behaves as `\ldots` and sometimes as `\cdots`.

```tex
$(1,2,3,\dots,16)$ % like \ldots
$1+2+3+\dots+16$ % like \cdots
```

I'm surprised that the builds on Travis haven't caught the issue in `texcmp.js`.

In any case, it's better to write distinct pull requests for different issues.

kevinbarabash:
Closing for the reasons stated by @gagern.

Issue #61, opened 2014-09-16
feature request: align / align* support


Pomax:
for the simple reason that it's one of the most used environments (along with `matrix` and `array`) =)

SirVer:
+1. I am missing this feature too, otherwise KaTex works amazingly well.

gagern:
Since #246 got merged, we have support for environments, and with `array` and `matrix` environments also some experience on how to do this kind of alignment. So the main task here, the way I see it, is working out all the spacing rules. How does LaTeX typeset `align*`? How much space is there between rows? How much gap between columns? I guess the latter might be some `\hfill`, at least between pairs of columns, but is there some minimum or default? Where in the TeX, LaTeX or amsmath sources can all of these settings be found? Once we have the dimensions, implementing them should be easy.

kevinb7:
I found this in amstex.sty:
```
\newenvironment{aligned}{%
  \relax\ifmmode\else\nonmatherr@{\begin{aligned}}\fi
  \null\,\new@ifnextchar[{\aligned@}{\aligned@[c]}%
}{%
  \crcr\egroup\egroup
}
\def\aligned@[#1]{%
  \if #1t\vtop \else \if#1b\vbox \else \vcenter \fi\fi
  \bgroup
   \Let@\restore@math@cr
   \default@tag
   \ifinany@\else\openup\jot\fi
   \ialign\bgroup\hfil\strut@$\m@th\displaystyle{##}$&%
    $\m@th\displaystyle{{}##}$\hfil\crcr}
```
In "TeX for the Impatient" it states the following:
> \ialign
This command behaves just like \halign, except that it first sets the \tabskip glue to zero and sets \everycr empty.

I take that to mean that there is no space between the columns or rows.

Pomax:
Note that that is not `align`/`align*`, but `aligned`. in the same `tex/latex/abs/math/amstex.sty` there is also
```
\def\align{\ifingather@
  {\ifnum0=`}\fi
  \DN@{\@nameuse{align (in \string\gather)}}%
  \else \ifmmode\nomath@env
    \DN@{\let\endalign\relax\@gobble}\else
  $$\stepcounter{equation}%
  \st@rredfalse\let\next@\align@\fi\fi
 \collect@body\next@}
\@namedef{align*}{\ifingather@
  {\ifnum0=`}\fi
  \DN@{\@nameuse{align* (in \string\gather)}}\else
   \ifmmode\nomath@env
[...]

PR #398, opened 2015-11-23
Introduce an aligned environment

gagern:
This is almost like the `align*` environment, but it starts out in math mode, so we don't have to worry about the fact that we have no real surrounding text mode in KaTeX.  This is the first step towards `align*` and `align` as requested in #61.


kevinbarabash:
@gagern thank you for implementing this.  A lot of people will be happy to see this addition to KaTeX.

spicyj:
Hard to imagine how someone could accidentally screw it up, but it might be nice to have a screenshot test with a non-rel split (e.g., `a&+b` or `a&b`).

Nice work.

PR #399, opened 2015-11-24
JSLint: A constructor name should start with an uppercase letter

henzestplus:
When building using make, this error shows up. This should fix the error without changing the scripts functionality

### DIFF
--- a/dockers/texcmp/texcmp.js
+++ b/dockers/texcmp/texcmp.js
@@ -6,7 +6,7 @@ var path = require("path");
 var Q = require("q"); // To debug, pass Q_DEBUG=1 in the environment
 var pngparse = require("pngparse");
 var fft = require("ndarray-fft");
-var ndarray = require("ndarray-fft/node_modules/ndarray");
+var Ndarray = require("ndarray-fft/node_modules/ndarray");
 
 var data = require("../../test/screenshotter/ss_data");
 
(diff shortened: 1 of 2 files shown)


kevinbarabash:
I was under the impression that the linter would fail the build if there was any lint.  Looks like that is not the case.  I had a quick look at the docs for `ndarray` and it looks like `ndarray` can be called as a regular function.  @gagern is there a reason that we're using `new` with `ndarray`?

gagern:
I've [seen such a thing recently](https://github.com/strobelm/CindyJS/commit/118f4e30216f1bcd16b72d55e21ab423aae1e6ec#commitcomment-14521520): jshint suddenly complaining about things it didn't complain about before. In the case I saw, this was due to jshint 2.9.1-rc1. Usually npm doesn't install prerelease versions, but older npm versions apparently do. In this case, the npm from node 0.10 was to blame. @henzestplus, what versions of node and npm are you running?

> is there a reason that we're using `new` with `ndarray`?

The call felt like a constructor to me, so I guess I untuitively used it as such. Reading [the ndarray docs](https://github.com/scijs/ndarray#readme) I find that I'm wrong. Feel free to drop the `new` call, it appears to only cause unneccessary overhead and confusion, and will trouble jshint for more people once 2.9 gets released for real.

henzestplus:
npm version: 1.3.10
node version: v0.10.25
I can't remember doing anything special to get a prerelease running.

gagern:
Yes, that's pretty ancient. Apparently the rules regarding prereleases were different in those days.

kevinbarabash:
I've migrated to eslint so this change doesn't make sense anymore.  Also, I think the correct change in this case would be to get rid of the `new`.

PR #405, opened 2015-12-01
Migrate to eslint

kevinbarabash:
Summary
We'd like contributors to use the same linter and lint rules that we use
internally.  This diff swaps out eslint for jshint and fixes all lint failures
except for the max-len failures in the test suites.

Test Plan:
- ka-lint src
- make lint
- make test

Reviewers: emily

PR #406, opened 2015-12-01
Force Firefox to start up with an empty page

gagern:
Otherwise it could happen that some Mozilla page gets shown which has a minimal size larger than the 786px we're requesting.  And the screenshot will span that entire page even if the window is smaller, resulting in a failure to adjust screenshot size.

See http://kb.mozillazine.org/Browser.startup.homepage_override.mstone and http://kb.mozillazine.org/Browser.startup.page for details.

As an alternative, we could just insert a `driver.get("about:blank")` before the `takeScreenshot`, but I prefer the clean solution of not even loading the offending page at all.

Just in case, we also include the docker image digests in the travis build log, to increase chances of reproducing what we get there.

This addresses issues discussed in https://github.com/Khan/KaTeX/pull/405#issuecomment-160857796 and https://github.com/Khan/KaTeX/pull/388#issuecomment-160967346.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,7 @@ services:
 before_script:
   - docker pull selenium/standalone-firefox:2.48.2
   - docker pull selenium/standalone-chrome:2.48.2
+  - docker images --no-trunc
 script:
   - npm test
   - dockers/Screenshotter/screenshotter.sh --verify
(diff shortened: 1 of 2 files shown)


kevinbarabash:
Thanks for investigating and finding a resolution so quickly.  LGTM!  ^_^

PR #409, opened 2015-12-05
Add support for base extended unicode characters in \text{}

kevinbarabash:
We'd like to support rendering extended latin characters in
order to support rendering text from European languages within
\text{}.

This diff integrates the font files from https://github.com/Khan/MathJax-dev/pull/1
and updates lexer.js and symbols.js so that KaTeX supports ßæœøÆŒØ within \text{}.
The plan is to support these characters with common accents to support more
of the latin-1 unicode block.

Test Plan:
- make test
- run all tests (including screenshot tests) on travis-ci


kevinbarabash:
Here's what it looks like in the test app.
screen shot 2015-12-04 at 6 18 26 pm

spicyj:
Neat.

src/Lexer.js
```diff
@@ -46,6 +46,7 @@ function Token(text, data, position) {
 var tokenRegex = new RegExp(
     "([ \r\n\t]+)|(" +                                // whitespace
     "---?" +                                          // special combinations
+    "|[\u00C0-\u00FF]" +
```
xymostech: Can you add a comment about what these match?

gagern: The `\]-\u2027` range should already include this, since `\]` is `\u005D`. If you look at `[!-\[\]-\u2027\u202A-\uD7FF\uF900-\uFFFF]` as a whole, that's almost any BMP codepoint except for the C0 control sequences `\u0000-\u001F`, the space `\u0020`, the backslash `\u005C`, the line separator `\u2028` and the paragraph separator `\u2029` which you might consider whitespace, the surrogate characters of UTF-16 `\uD800-\uDFFF` which are only matched if they form a complete surrogate pair, and the private use area `\uE000-\uF8FF`. So to make it short, this change should not be neccessary, in my opinion.

kevinbarabash:
I'm going close this because https://github.com/Khan/KaTeX/pull/508 handles these characters, plus accented characters, and more.

Issue #422, opened 2015-12-29
Add support for \kern


Pomax:
I'm using the following block of LaTeX, to make sure the maths makes visual sense to the reader, but KaTeX does not expect `\hskip` in this block, and errors out:
```
\begin{align*}
  linear &= \hskip{2.5em} 1 + 1 \\
  square &= \hskip{1.7em} 1 + 2 + 1\\
  cubic &= \hskip{0.85em} 1 + 3 + 3 + 1\\
  hypercubic &= 1 + 4 + 6 + 4 + 1
\end{align*
```
The error it generates is: 
```
Error: KaTeX parse error: Expected & or \\ or \end at position 42: gned}linear &= ̲\hskip{2.5em} 
```

kevinbarabash:
In this situation you could also use `\kern` which we also don't support yet, but is easier to implement because we don't have to worry about line splitting or stretch/squash of glue.  I'm going to rename the issue to `implement \kern` with the expectation that you should be able to use `\kern` anywhere.

Pomax:
had to look that one up (yay, TeX by Topic), but yeah that would be fine too.

WChargin:
@Pomax A few unsolicited comments about your original TeX!

First, it's either `\hskip2.5em` or `\hspace{2.5em}`; the former is a TeX primitive while the latter is a LaTeX command. Second, you'll certainly want to enclose the labels on the left in `\mathrm`s so that the spacing is correct—e.g., on the `li` pair in "linear."

Third, and most importantly, doing horizontal spacing manually like this is probably a Bad Idea™. When I change the font to mathpazo, everything will break. When you're in `\scriptsize`, everything will break. The proper way to do this is to think about the layout you want: a left-aligned column with the names, then a column with an equals sign, then a centered column with the values. This calls for an `array`. So you might try this:
```tex
\begin{equation*}
  \begin{array}{lcc}
    \mathrm{linear}     &=& 1 + 1  \\
    \mathrm{square}     &=& 1 + 2 + 1 \\
    \mathrm{cubic}      &=& 1 + 3 + 3 + 1 \\
    \mathrm{hypercubic} &=& 1 + 4 + 6 + 4 + 1
  \end{array}
\end{equation*}
```
This will almost work, but the spacing around the equals signs will be way too big, because it will be getting the normal column gutter spacing instead of the proper spacing for a relation:
[...]

PR #423, opened 2015-12-31
Add support for \kern

kevinbarabash:
Summary:
Addresses https://github.com/Khan/KaTeX/issues/422.
This only supports em and ex units and doesn't handle vertical layouts.
Negative kerning works.

Test Plan:
- make test
- make screenshots (verify that d is slightly overlapping c in the screenshots)

Reviewers: emily


spicyj:
Can you add something with a changed font size to the screenshot, like a frac? Not sure offhand if the space should scale or not.

kevinbarabash:
It looks like LaTeX is not scaling `\kern`.  Here are some screenshots.

LaTeX:
screen shot 2015-12-30 at 6 07 37 pm

KaTeX:
screen shot 2015-12-30 at 6 07 43 pm

kevinbarabash:
It's a little hard to tell from the screenshot, but the distance between `a` and `b` in both the fraction and not-in-the-fraction is 87 pixels in both cases (as measured by dev tools inspector).
screen shot 2015-12-30 at 6 15 23 pm

xymostech:
@kevinbarabash this looks good to me!

kevinbarabash:
@xymostech thanks

PR #431, opened 2016-02-01
Fix sqrt bar

chrismessiah:


chrismessiah:
Meant to make a pull req to my own fork.

PR #436, opened 2016-02-10
Bundle css and fonts in npm package.

### LINKED ISSUE #114: Include built files in the repository
This will allow folks to take advantage of all of the perks that package managers provide.

If you're strictly using KaTeX on the server, then you won't run into this issue as you won't really need the built files. But when working on the client it's convenient to not have to worry about building the project, as you can just reference the built files directly.

For small projects [the manual installation process](https://github.com/Khan/KaTeX#usage) works fine, but the benefits of just being able to run `npm update` or [`bower update`](https://github.com/Khan/KaTeX/pull/112#discussion_r17942712) to get the latest versions for things (with the dependency tree sorted) quickly becomes relevant as a project's dependencies grow, especially as the project ages and its dependencies become out-of-date.

It might seem odd to include the built files, but a large number of popular libraries do it. Some examples are [Backbone](https://github.com/jashkenas/backbone/blob/master/backbone-min.js), [Marionette](https://github.com/marionettejs/backbone.marionette/blob/master/lib/backbone.marionette.min.js),  [Angular's proxy repository](https://github.com/angular/bower-angular), Ember's (unofficial?) [proxy library](https://github.com/components/ember), and [Bootstrap](https://github.com/twbs/bootstrap/tree/master/dist).

tirsen:
Fixes #114 for npm packages.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,5 @@
 .PHONY: build dist lint setup copy serve clean metrics test zip contrib
-build: setup lint build/katex.min.js build/katex.min.css contrib zip compress
+build: lint build/katex.min.js build/katex.min.css contrib zip compress
 
 dist: build
 	rm -rf dist/
--- a/package.json
+++ b/package.json
@@ -10,7 +10,8 @@
   "files": [
     "katex.js",
     "cli.js",
-    "src/"
+    "src/",
+    "dist/"
   ],
   "license": "MIT",
   "devDependencies": {
@@ -31,7 +32,8 @@
   },
   "bin": "cli.js",
   "scripts": {
-    "test": "make lint test"
+    "test": "make lint test",
+    "prepublish": "make dist"
   },
   "dependencies": {
     "match-at": "^0.1.0"


kevinbarabash:
@tirsen thanks for the pull request.  Before merging this, could you sign our [CLA](www.khanacademy.org/r/cla)?

spicyj:
Can we include 'dist' instead of 'build'?

spicyj:
To match the bower package.

tirsen:
It now includes dist but I had to remove the call to npm install from the "dist" target because npm install automatically calls "prepublish". I was looking into having a separate "dist-no-npm" target but it all gets pretty messy.

I think maybe the best is to have travis call "npm install" and that calls "make dist" instead of calling "make dist" which calls "npm install"? Not entirely happy with that either, open to suggestions. :-)

tirsen:
And I signed the CLA. :-)

kevinbarabash:
@tirsen thanks for signing the CLA.  I'm not that knowledgable about npm package so I'll leave @spicyj to make the call on that.

tirsen:
Done.

tirsen:
It looks like Travis actually runs npm install and npm test (auto detected probably?). So I think we're all good. The only issue is if you run "make" directly assuming it will run npm install, which it will no longer. (Since "npm install" now runs "make dist".)

spicyj:
Thanks!

razor-x:
Maybe new issue, but it would be nice to get the auto-render lib in there as well, or publish it as it's own npm package.

PR #444, opened 2016-03-01
README: Use HTTPS where possible

mathiasbynens:
There’s no reason not to always use HTTPS. See https://github.com/konklone/cdns-to-https#background for more info.


mathiasbynens:
Btw, consider updating this URL to use HTTPS as well:

PR #465, opened 2016-04-27
Linebreaks

mull:
I'm sending this PR only to start a discussion and to hopefully communicate that some of us in the community want to help you guys reach this feature :dog:. I don't pretend to think that this implementation is all there is to it, of course.

The commit 735b307 has the relevant code.

This is testable by running:

1. `make dist`
2. `python -m SimpleHTTPServer` // in project root
3. `open http://localhost:8000/static/`

For me the standout problem to line breaks seemed to be that elements weren't getting their heights correctly set. I suppose this is due to something with how the web font is used (larger text without setting larger font size.) "X-ing" the line height for each size class however solves this problem.

The problem that this PR does not solve however is where line breaks happen. MathJax for example would be smart enough to split things up, whereas in this branch the line break is "breaking up a group." I suspect that this can be fixed with CSS, though I need to refresh my memory.
screen shot 2016-04-27 at 18 02 11

Thoughts? I'm sure you guys had considered something like this already.

### DIFF
--- a/static/katex.less
+++ b/static/katex.less
@@ -16,7 +16,7 @@
 
 .katex {
     font: normal 1.21em KaTeX_Main;
-    line-height: 1.2;
+    display: inline-block;
     white-space: nowrap;
 
     // Protect elements inside .katex from inheriting text-indent.
@@ -27,6 +27,7 @@
         // which is undesireable ("to $x$,"). Instead, adjust the .katex-html
         // style and put nowrap on the inline .katex element.
         display: inline-block;
+        white-space: initial;
     }
 
     .katex-mathml {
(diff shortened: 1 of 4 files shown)


mull:
PS. I did not test this in anything else than Chrome, as I just wanted a proof of concept.

static/katex.less
```diff
@@ -442,10 +443,10 @@
     }
 
     .delimsizing {
-        &.size1 { font-family: KaTeX_Size1; }
```
spicyj: Why were these changes needed?

spicyj:
I think this is generally the right idea. I was thinking that we should make the entire expression wrappable but each unit inside unbreakable. So that means a left/right atom wouldn't wrap and neither would something created with braces like `{a + 2}`. TeX actually has more complicated breaking rules – like it will generally only break after a `+`, not before. It would be really nice if we could encode those rules properly in the output. Maybe one option is to even output ordinary HTML `wbr` tags. We should at least start by looking in the TeXbook to see what those rules are.

static/katex.less
mull: Do you mean why I gave `size1` a line-height? For consistency, and because I don't even know half what I'm doing. :grin:

spicyj: For consistency with what? :) If there is literally not fixing anything, we can remove it – but I assume you added it to fix some particular problem you saw.

mull: Good point. I never tried without it. I just looked at the code and said "hey, each of these should get a line height and I think one of the problems is solved" :grin:

mull:
I might or might not be able to approach the TeXbook and get some progress done. At our company we'll go with this little change for now as a poor workaround for small devices. I've never written a line of TeX in my life, going through the book might be a daunting task! If I can assist you in any way meanwhile to get the ball rolling let me know. :-)

kevinbarabash:
@mull I'm going to close this.  Let me know if you do any investigating in the TeXbook?

Issue #471, opened 2016-05-05
support small caps


msridhar:
It'd be cool if KaTeX supported `\textsc`.  I think it's just a matter of adding the style `font-variant: small-caps` in the right place.  If someone can suggest to me the right place(s) to change in `buildHTML.js`, I could potentially take a crack at it sometime.

kevinbarabash:
If anyone is interested in the picking this up, the PR was a good start.  There were a couple of issues that still need to be resolved, see the notes in #472.

edemaine:
Recently got a feature request for this.  I think the right approach is probably to add to our font, as in https://github.com/KaTeX/katex-fonts/issues/28

hanryhu:
We would love if you could implement this!  We use KaTeX through Gradescope and would really like to mention algorithms in a nice font for our online exams.

\- Design and Analysis of Algorithms teaching team at MIT

ylemkimon:
According to http://ctan.forsale.plus/fonts/amsfonts/doc/amsfndoc.pdf cmcsc10 is used for small caps.

<img width="863" alt="screen shot 2018-02-03 at 1 08 06 am" src="https://user-images.githubusercontent.com/1044413/35763652-c8caf556-087e-11e8-8dcc-397cc942bf37.png">

_Originally posted by @kevinbarabash in https://github.com/KaTeX/katex-fonts/issues/28#issuecomment-362783677_

RexYuan:
Any update on this?

yuvaltassa:
Also hoping for this to be reviewed soon.

Small caps are important!

PR #472, opened 2016-05-06
Add support for \textsc.

msridhar:
This is a first attempt, and may still have issues.  In particular,
I am not sure what needs to be done for MathML rendering.

Fixes Khan/KaTeX#471

### DIFF
--- a/src/buildCommon.js
+++ b/src/buildCommon.js
@@ -406,6 +406,10 @@ var fontMap = {
         fontName: "Main-Regular",
     },
 
+    "textsc": {
+        variant: "normal",
+        fontName: "Main-Regular",
+    },
     // "mathit" is missing because it requires the use of two fonts: Main-Italic
     // and Math-Italic.  This is handled by a special case in makeOrd which ends
     // up calling mathit.
--- a/src/functions.js
+++ b/src/functions.js
@@ -492,7 +492,7 @@ defineFunction([
 
 defineFunction([
     // styles
-    "\\mathrm", "\\mathit", "\\mathbf",
+    "\\mathrm", "\\mathit", "\\mathbf", "\\textsc",
 
     // families
     "\\mathbb", "\\mathcal", "\\mathfrak", "\\mathscr", "\\mathsf",
--- a/static/katex.less
+++ b/static/katex.less
@@ -92,6 +92,11 @@
         font-style: italic;
     }
 
+    .textsc {
+        font-family: KaTeX_Main;
+        font-variant: small-caps;
+    }
+
     // This value is also used in fontMetrics.js, if you change it make sure the
     // values match.
     @ptperem: 10.0;


msridhar:
Happy to address comments.  I Googled around for MathML and small caps, but couldn't find anything helpful.  So, I just used the same `fontMap` entry as `"mathrm"` in `buildCommon.js`.

msridhar:
Just a ping for any feedback on this.  Happy to make changes.

kevinbarabash:
@msridhar sorry for the delay.  Can you post some screenshots?  I'm not sure how the browser actually implements this.  If it's swapping out glyphs, e.g. using unicode small caps A (U+1D00) for every `a` then the browser will end up using a different font.  It might be scaling existing glyphs in the font if the font doesn't include specific small caps glyphs.  Either way, we don't have the metrics for those.  That being said, we only need the height and depth.  It looks like we could probably use the x-height of the font as the height for the small caps and since the small caps don't have descenders, the depth should be 0.

msridhar:
screen shot 2016-05-27 at 10 20 25 am
Is this the kind of screenshot you're looking for?  Let me know if I should try some other text.

One thing you can see is that space characters seem to be getting ignored.  Typically they won't be needed, but I can try to look into why this is happening.

kevinbarabash:
@msridhar could you have a look at why we're dropping spaces?  I want to check this branch out and play around with before merging it.  I'm concerned about how it handles vertical layouts especially since the font metrics are going to be wrong for the characters now.

kevinbarabash:
This might be something we put behind a flag.

kevinbarabash:
As for MathML support, you could try the `<mi mathvariant='small-caps'>ab</mi>` and see that works in your favorite MathML viewer.

msridhar:
Thanks for the feedback! I'm on vacation now, will look in a couple weeks.

edemaine:
@msridhar Any updates on this?

msridhar:
@edemaine unfortunately I never had the time to pick up this work, and I likely won't have time soon.  The changes I made were quite simple, but I'm not sure what exactly it would take to address the remaining comments.

kevinbarabash:
@msridhar sorry for the delay in responding.  There are two issues that still need to be resolved:
- getting the correct vertical metrics for each of the glyphs (lowercase y should not have any depth/descent, but its height/ascent should be greater than the standard y)... using the metrics from the standard 'y' when rendering a small caps 'y' will result in incorrect positioning in vertical layouts such as `\frac` and `\sqrt`.
- spaces should render as spaces

Here's what the `\textsc{Lorem Ipsum Dolar}` should look like according to quicklatex.com:
screen shot 2017-09-13 at 9 50 39 pm

kevinbarabash:
I believe that the browser is doing something to the font to generate the small caps.  I found a tool called [cap-height](https://github.com/sebdesign/cap-height) that can measure the height of small caps so we could use it to figure out what the height should be.

msridhar:
Hey, so if someone else could take this one over the finish line that'd be
great. I'm not going to have time in the near future.

kevinbarabash:
@msridhar I'm going to close the PR for now.  I'll paste the details of what work remains in the issue if someone else wants to pick this up.

PR #485, opened 2016-05-28
Fix #482: <Math Atom> commands [WIP]

### LINKED ISSUE #482: Implement <math atom> commands
The `<math atom>` commands are: 
- \mathop
- \mathrel
- \mathord
- \mathinner
- etc...

I feel that this is a necessary step before implementing user-defined macros.  An example use is to allow for question marks placed over equal signs `\stackrel{?}{=}` which is defined by 

```
\stackrel#1#2 \mathrel {\mathop {#2}\limits ^{#1}}
```

Or for user-defined operators like `\def\limsup {\mathop{\overline\mathrm{lim}}\limits}`.

I started working on an implementation and I was able to for the most part except that `\mathop{=}\limits^{?}` wouldn't work primarily due to TeX logic living in BuildHTML.js.  Certainly I could hack it out, but there were other issues involved such as implementing `\mathrel` commands would essentially require a new `relgroup`, just like there is a [`mathordgroup`](https://github.com/Khan/KaTeX/blob/master/src/buildHTML.js#L232) and so on to allow recursive expansion.  This brought me to conclude that it would most likely be better to wait for a new parser/AST implementation before continuing.

cbreeden:
This is a first attempt to fix #482.  Essentially this fix would require each atomic type assume that it's nucleus is either a single object, or contains another mathlist.  This is the meaning of the `if (group.value.value)` statements found in `BuildHTML.js`.  Would this make `ordgroup` obsolete?

A current standing issue with this implementation (besides my negligence of MathML for the time being) is that inside a math atom command, spacing will not be honored.  For instance in `\mathop{a+b}` the `a+b` will all be spaced as if there were `mord` elements.  This is because the `mop` span wrapping `a+b` does not have the class `textstyle`.  A working fix for this is to remove the `>`'s from the style sheet this area:

``` less
    .textstyle {
        > .mord {
            & + .mord {}
```

But besides that it looks like it could be a good start.  I was hoping that I could obtain some feedback.  Here are some example renderings:

Input:

``` latex
\mathop{\overline\mathrm{lim}}\limits_{x\to\infty}f(x)
```

Renders:
limsup

Input:

``` latex
a\mathrel{\mathop{=}\limits^{\blue ?}}b
```

Renders:
stackrel


src/functions.js
```diff
@@ -159,6 +159,74 @@ defineFunction("\\color", {
...
+       return {
+           type: "rel",
+           value: body.value,
+       }
+   } else {
+       return {
+           type: "rel",
+           value: [body],
+       }
+   }
```
kevinbarabash: this could be simplified to 
```js
return {
   type: "rel",
   value: body.type === "ordgroup" ? body.value : [body]
};
```

kevinbarabash:
@cbreeden thanks for the pull request.  It looks like travis is failing on lint.  Could you run `make lint` and fix any errors?

In terms of feedback on the design @xymostech will be able to answer your questions much better than I would.

.gitignore
```diff
@@ -9,3 +9,4 @@ diff.png
 /test/symgroups.aux
 /test/symgroups.log
 /test/symgroups.pdf
+.vscode
```
kevinbarabash: Could you put this in your global .gitignore instead?

src/functions.js
cbreeden: Oh good point I will do that.  I spent quite a bit of time fiddling around trying to figure out why atoms are wrapped twice (parseFunctions was culprit if I remember) and went through a few iterations to figure out what to put here.

.gitignore
cbreeden: Sure, good idea

src/buildHTML.js
```diff
@@ -716,6 +728,9 @@ groupTypes.op = function(group, options, prev) {
     var supGroup;
     var subGroup;
     var hasLimits = false;
+
+    // TODO(cbreeden) investigate the possibility of putting the supsub
+    //   tages in the atoms themselves.  Allow for easier recursion?
```
kevinbarabash: Could you give an example of what you mean by this?

src/buildHTML.js
```diff
@@ -740,6 +755,10 @@ groupTypes.op = function(group, options, prev) {
         large = true;
     }
 
+    // TODO(cbreeden) Currentl mathop atoms have the assumed structure:
+    //   group<atom>.value === { body, symbol, limits, type }
+    //   this may make it difficult for ops that are not either
+    //   a single symbol or list of text.
```
kevinbarabash: It seems like those are all of the cases we handle in `groupTypes.op`... are there other types of operators that we haven't handled yet that you're concerned about?

src/buildMathML.js
```diff
@@ -325,6 +325,10 @@ groupTypes.op = function(group) {
         // This is a symbol. Just add the symbol.
         node = new mathMLTree.MathNode(
             "mo", [makeText(group.value.body, group.mode)]);
+    } else if (group.value.value) {
+        // do nothing for now
+        var inner = buildExpression(group.value.value, options);
+        node = new mathMLTree.MathNode("mo", inner);
```
kevinbarabash: I think this makes sense.  Is there something additional that you think we should be doing in the future?

src/buildHTML.js
cbreeden: That was just an idea I was toying with.  I believe the TeX places the scripts inside their respective nodes -- something like if you could imagine each ParseNode had a `superscripts:`, and `subscripts:` field.  The engine will do preliminary processing on the nodes (like changing atom types, etc.), before finally looking for a `scripts` field.  I guess it kind of makes some properties (such as atomtype) more transparent while processing (ie, you don't need to recursively look into the base to determine the atom type).  But I abandoned the idea.  It didn't seem worth it.  Something like this would probably only be useful if other parsenode were refactored to be inline: ie stylechanges.  Where for instance, instead of having a `StyleChange` node wrap the contents it imposed on, you have a single node `StyleChange` which just changes the state of the HTMLRendered.  Anyway, it didn't seem necessary to go down this path, as these recursive definitions seemed quite necessary/natural give the HTML/CSS target.

src/buildHTML.js
cbreeden: Sorry, I don't remember what my concern was here.  We could probably remove most of these `TODO`.  They were just my notes while thinking of alternatives, I believe.

src/buildMathML.js
cbreeden: I don't know.  To be honest, I wasn't happy with the approach.  Basically, if I remember correctly, I just claimed that some `ParseNode` values may be a list of parse nodes, and if they are then the current parsenode is simply an atom-type wrapper.  But I also haven't thought of any better way.

cbreeden:
closing as fixed in 6bb62b11b4ea

PR #489, opened 2016-06-01
Parsing Notes

cbreeden:
I wrote these up notes while figuring out the internal tree that KaTeX uses when parsing TeX.  I thought maybe someone else might find this useful as well, and perhaps could be a good reference for later on.

Unfortunately, I am not able to make pull requests to the wiki, so this is my wiki pull request.

Warning: I'm sure there are a few typos and things that could be cleared up.

### DIFF
--- a/ParsingNotes.md
+++ b/ParsingNotes.md
@@ -25,10 +25,10 @@ syntax tree which is simply an array of `ParseNode`s.
 The folowing is a list of currently implemented ParseNodes.  They can roughly
 categorized into four different groups.
 
-1. **Symbols**
-2. **Functions**
-3. **Math Lists**
-4. **Wrappers**
+1. [**Symbols**](#Symbols)
+2. [**Math Lists**](#Math-Lists)
+3. [**Functions**](#Functions)
+4. [**Wrappers**](#Wrappers)
 
 ### Symbols
 
@@ -147,6 +147,7 @@ You can find these examples in
  - Styling: `\displaystyle`, `\textstyle`, etc.
  - Sizing: `\Large`, `\huge`, `\tiny`, etc.
  - Font Changes: `\mathit`, `\mathbb`, `\mathrm`, etc.
+ 
  
  These are desgined to be a transparent as possible when being passed through the renderer
  and so these usually have the structure of something like this (this particular example is
--- a/ParsingNotes.md
+++ b/ParsingNotes.md
@@ -25,14 +25,14 @@ syntax tree which is simply an array of `ParseNode`s.
 The folowing is a list of currently implemented ParseNodes.  They can roughly
 categorized into four different groups.
 
-1. [**Symbols**](#Symbols)
-2. [**Math Lists**](#Math-Lists)
-3. [**Functions**](#Functions)
-4. [**Wrappers**](#Wrappers)
+1. [Symbols](#Symbols)
+2. [Math Lists](#Math-Lists)
+3. [Functions](#Functions)
+4. [Wrappers](#Wrappers)
 
 ### Symbols
 
-Syombols are the easiest of the four groups in that they have a minimal structure.
+Symbols are the easiest of the four groups in that they have a minimal structure.
 A typical example is the following:
 
 ```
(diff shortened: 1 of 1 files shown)


ParsingNotes.md
```diff
@@ -0,0 +1,167 @@
...
+categorized into four different groups.
+
+1. [**Symbols**](#Symbols)
+2. [**Math Lists**](#Math-Lists)
+3. [**Functions**](#Functions)
+4. [**Wrappers**](#Wrappers)
+
+### Symbols
+
+Syombols are the easiest of the four groups in that they have a minimal structure.
```
kevinbarabash: Symbols

ParsingNotes.md
```diff
@@ -0,0 +1,167 @@
...
+type: "bin"
+value: "+"
+mode: "math"
+```
+Any token that can be found in `symbols.js` will have this structure.  It can occur
+in either `math` mode or `text` mode, and the type can be any of the basic `atom`
+types defined in TeX--accent, bin, close, inner, mathord, op, open, punct, rel--but
+also include two special cases: spacing, textord.  The value is simply the unicode
+character corresponding to the given symbol, for example `\u222b` is the symbol
+corresponding to the integration operator `\int`; This is the replace field in the
```
kevinbarabash: put `replace` in back ticks.

ParsingNotes.md
cbreeden: yeah, i think I should probably go ahead and fix the typos in this document.  I have found quite a few.

ParsingNotes.md
```diff
@@ -0,0 +1,167 @@
...
+definition of a symbol, and is not included in the `ParseNode`. The file `symbols.js` 
+also contains information about which font the unicode symbol is placed in.  The
+general design rule for not including this information in the `ParseNode` is to not
+incorporate more infromation that is necessary for `buildMathML` to do it's job.
+
+### Math Lists
+
+Math lists are supposed to be like a new list of `ParseNodes`.  This allows for
+easy recursion while rendering in the same spirit that most reference TeX
+implementations will compile.  As of right now there is only on example, the
```
kevinbarabash: The last part of the "This allows for... " sentence doesn't make sense.  Maybe split them into separate sentences.

kevinbarabash: "..there is only one example,"

ParsingNotes.md
```diff
@@ -0,0 +1,167 @@
...
+as shown in the `genfrac` type above that is expected from the parser.  One last
+example, which is quite common, is the `subsup` node:
+
+```
+type: "supsub"
+value:
+    base: ParseNode  //either a symbol or ordgroup
+    sup:  ParseNode  //either a symbol or ordgroup
+    sub:  ParseNode  //either a symbol or ordgroup
+mode: "math"
```
kevinbarabash: I didn't realize that `supsub` didn't have second level of indirection like the other functions.

ParsingNotes.md
```diff
@@ -0,0 +1,167 @@
...
+
+You can find these examples in 
+
+ - `\text`, `\phantom`, `\color`
+ - Styling: `\displaystyle`, `\textstyle`, etc.
+ - Sizing: `\Large`, `\huge`, `\tiny`, etc.
+ - Font Changes: `\mathit`, `\mathbb`, `\mathrm`, etc.
+ 
+ 
+ These are desgined to be a transparent as possible when being passed through the renderer
```
kevinbarabash: "...as transparent as possible..."

kevinbarabash:
@cbreeden thanks for typing this up.  It would be nice to simplify the AST at some point.  I think we probably also want create a more general tree structure based on the HTML tree that could be used by multiple rendering backends.

cbreeden:
That was a goal of mine as well, at least when I first started.  I quickly started running into problems when I started to realize the diverse structure of `ParseNode` which was the motivation for me to write up a survey of the current structure and somehow provide some kind of way to conceptualize how to think about them.

Anyways, I thought that this could be a useful document to start the discussion of simplifying the AST if that route is desired.

ParsingNotes.md
cbreeden: Essentially because it's not in `functions.js`.  Everything in `functions.js` will have a second level of indirection.  So perhaps wasn't a good example.

kevinbarabash:
@cbreeden why did you close this?

Issue #470, opened 2016-05-04
Fonts dont always load during automated screenshot chrome test


mathijshenquet:
Some selenium chrome test containing special fonts (mathbb, mathfrak, etc) randomly fail on my system. Visually the characters set in the special fonts are invisible. Possibly this is related to[1](https://productforums.google.com/forum/#!topic/chrome/kBzYJZV-qMM;context-place=topicsearchin/chrome/category$3Amac|sort:relevance|spell:false)

kevinbarabash:
We could try using https://github.com/typekit/webfontloader to force load fonts before taking the screenshot.

kevinbarabash:
The screenshotter has been fairly reliable of late so I'm going to close this issue.

PR #490, opened 2016-06-01
Fix #488 Subsup alignments

### LINKED ISSUE #488: Subscripts and superscripts inside franctions are centered; should be left-aligned
Consider the term `\frac{1}{x_{666}^{9}}`

We would expect this to render as follows:

photo489820666715058905

Instead, 9 is centered wrt 666:

photo83060647871494590

This doesn't happen outside of fractions, but may also occur in some other 'modes'?

screenshot from 2016-05-31 15-29-19

cbreeden:
Ref #488. I found the selector `msupsub` in the CSS.  I suspect this line was intended to be coded but was forgotten.

Input:
```
\frac{1}{\hat{\omega}^{8}_{888}} \quad \displaystyle\sum_{\omega^{8}_{888}}
```

Before Pull Request:
scripts_broke

After Pull Request:
scripts_fix


xymostech:
You're right! Looks like in https://github.com/Khan/KaTeX/commit/29b00ee6 I accidentally deleted it!

Would you mind adding a screenshotter test for this? You'll have to add something to test/screenshotter/ss_data.yaml and then run the screenshotter on your computer and check in the new images.

src/buildHTML.js
```diff
@@ -376,6 +376,7 @@ groupTypes.supsub = function(group, options, prev) {
         supsub.children[1].style.marginRight = scriptspace;
     }
 
+    supsub.classes.push("msupsub");
```
xymostech: We generally prefer doing something like `makeSpan(["msupsub"], [supsub])` instead of mutating elements, does that still work?

cbreeden: Fixed, maybe I should've inlined this call in the final final makeSpan call.  Just let me know.

cbreeden:
So there was a strange issue when I was running `screenshotter.sh` from docker.  There was one file modification:

mathbf-chrome

Where it should've looked more like firefox:

mathbf-firefox

I got the same results when I tried on master.  I'm not sure what the issue is on my end, but I didn't add this modification to the commit.

test/screenshotter/ss_data.yaml
```diff
@@ -106,6 +106,8 @@ SqrtRoot: |
 SupSubCharacterBox: a_2f_2{f}_2{aa}_2{af}_2
 SupSubHorizSpacing: |
     x^{x^{x}}\Big|x_{x_{x_{x_{x}}}}\bigg|x^{x^{x_{x_{x_{x_{x}}}}}}\bigg|
+SupSubLeftAlignReset: |
```
xymostech: Could you add a plain `\omega^{8}_{888}` not in a fraction here too, since those environments are apparently different?

kevinbarabash:
@cbreeden it passed the CI build so it should be good.

xymostech:
I think that screenshotting bug isn't your fault, looks like other people are seeing it: https://github.com/Khan/KaTeX/issues/470

One more request, but looks good!

cbreeden:
Dagnit, I should've taken that rebase a chance to fix the typo in my comment. :)

test/screenshotter/ss_data.yaml
cbreeden: I might have misinterpreted what you said but I added one in.

xymostech: nope, looks perfect

xymostech:
Thanks!

PR #491, opened 2016-06-02
Refactor infix handling to remove depency on backtracking.

cbreeden:
Ref #250 for adding macro support.  I think it has been determined that before any macro support is added, we need to remove any dependency of using the lexer at arbitrary positions.  I have refactored the infix handling to remove such dependencies.

Currently we handle infix by scanning each list of `ParseNode`s after completion to see if there are any infixes.  This implementation just sets a parser variable once an infix is found.

I know the `defer_infix` part may be a little confusing, but this is essentially a hack to keep backward compatibility with style changes (like `\displaystyle`)--it's the same reason, I suppose, that we had to write a hack to parse infixes at the end of `parseExpression`.  We handle stylechanges a little differently than TeX (which treats style changes as inline nodes instead of wrappers) which makes it a little awkward to handle.  But I believe this is handled correctly.  Another option is to check for style changes after `var atom = this.parseAtom();` and terminate the mathlist there since style changes should _aways_ be the last node in a mathlist.

As of right now, I'm passing Jasmine and the Screenshooter.  Any ideas for adding further tests?

Other things to consider:
 - Plan for error handling and parser positions.
 - Remove parser positions from functions.  No one seems to be using it anyway.

### DIFF
--- a/src/Parser.js
+++ b/src/Parser.js
@@ -137,34 +137,75 @@ Parser.prototype.parseExpression = function(breakOnInfix, breakOnToken) {
     // we reached the end, a }, or a \right)
     while (true) {
         var lex = this.nextToken;
-        var pos = this.pos;
+        var numer;
+        var denom;
+
         if (endOfExpression.indexOf(lex.text) !== -1) {
             break;
         }
         if (breakOnToken && lex.text === breakOnToken) {
             break;
         }
+
+        if (this.infix_defer) {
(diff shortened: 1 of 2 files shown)


cbreeden:
I put a little more thought into the implementation, and was able to come up with a failing test: `1 \over {2 \over 3}`.  This needs to be fixed first.

PR #492, opened 2016-06-04
Fix #4 - Spacing interactions between atoms

### LINKED ISSUE #4: Convert Bin + Rel/Close/Punct to Ord + Rel/Close/Punct
6. If the current item is a Rel or Close or Punct atom, and if the most recent previous atom was Bin, change that previous Bin to Ord.

cbreeden:
Ref #4.  This was to fix the spacing issues of binary operators when the next element is a relative operator, a closing operator, a punctuation or at the end of the list.  I just added a `next` parameter to the `buildExpression` chain.  I added as many relevant tests I could think of.

I also fixed an issue where binaries operators weren't respecting `phantom` environments.

So now it's possible to have a >< face with dexter hair: 

```
3\times)
```

dexter

### DIFF
--- a/src/buildHTML.js
+++ b/src/buildHTML.js
@@ -22,11 +22,19 @@ var makeSpan = buildCommon.makeSpan;
  * nodes. This function handles the `prev` node correctly, and passes the
  * previous element from the list as the prev of the next element.
  */
-var buildExpression = function(expression, options, prev) {
+var buildExpression = function(expression, options, prev, next) {
     var groups = [];
+    var localnext;
     for (var i = 0; i < expression.length; i++) {
         var group = expression[i];
-        groups.push(buildGroup(group, options, prev));
+        // While in a subtree, next should be the next sibling
+        // If at the last node, next is the parent's next sibling.
+        if (i + 1 < expression.length) {
+            localnext = expression[i+1];
+        } else { 
+            localnext = next;
+        }
+        groups.push(buildGroup(group, options, prev, localnext));
         prev = group;
     }
     return groups;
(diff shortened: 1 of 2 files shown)


kevinbarabash:
@cbreeden thanks for the PR.  Not sure when I'll have time to review this, hopefully soonish.

kevinbarabash:
@cbreeden sorry for the delay with this.  There will probably be a merge conflict with https://github.com/Khan/KaTeX/pull/567, but I think we still want this.

kevinbarabash:
> 6. If the current item is a Rel or Close or Punct atom, and if the most recent previous atom was Bin, change that previous Bin to Ord.

In order to minimize changes I'd rather update `groupTypes.rel`, `groupTypes.close` and `groupTypes.punct` to find the previous atom and update it as necessary.

kohler:
FWIW, I have a commit that depends on #567 (and #567 is actually required for a full fix). I hadn't made a PR yet because #567 is in process.

kohler@6af1c96c523e4174374764291ea897b879ee9ce0

cbreeden:
> In order to minimize changes I'd rather update groupTypes.rel, groupTypes.close and groupTypes.punct to find the previous atom and update it as necessary.

Yeah, I'll admit that I just didn't like the idea of modifying the HTML nodes after they have been constructed.  But after seeing @kohler's commit, which modifies all the required atomtypes after building the expression, I think it's a clean solution in my opinion.  We might be better off taking that approach.

kevinbarabash:
In that case I'm going to close this in anticipation of a PR from @kohler with https://github.com/kohler/KaTeX/commit/6af1c96c523e4174374764291ea897b879ee9ce0.

PR #493, opened 2016-06-21
Allow macro definitions in settings

gagern:
This is the next step towards macro support in KaTeX. It doesn't do `\def` or `\newcommand` yet (as is the ultimate goal of #250), but it allows the user to declare macros as part of their settings:

```js
katex.render("\\exp\\foo", div, {macros: {"\\exp": "e^", "\\foo": "23.4x"}})
```

is equivalent to rendering `e^23.4x`, including the fact that only the `2` is the exponent here, just as you'd expect from proper TeX behavior. (You can define `\foo` as `{23.4x}` to fix this possibly unintended but compatible result.)

A [previous version of this code](https://github.com/gagern/KaTeX/compare/8496a1e40f8dfbfc8dd7d2e7b5096bd5c2b01c02...b81fcb8765bbdb77c0514c566d651a8726dde15f) has been lying around for about one year, during which it was [in active use by the CindyJS project](https://github.com/CindyJS/CindyJS/blob/f4eeb07f07ace58add34f7b4d028da81288e8d58/plugins/katex/src/js/katex-plugin.js#L107-L145). The code was discussed to some extent in https://github.com/Khan/KaTeX/issues/250#issuecomment-119358745, but according to https://github.com/Khan/KaTeX/issues/266#issuecomment-166949076 I had wanted to polish the error reporting into a pull request in its own right, and somehow more important things got in the way and I never got round to finishing that.

For this reason, the first commit of this pull request here is tightly coupled with large-scale modifications to how errors are being reported. The changes are just so tightly integrated with code that moves the parsing of special groups from the lexer to the parser that I feel that separating things again would mean a massive amount of work for little benefit. Doing the error reporting changes without the restructuring to the parser means doing error reporting for certain kinds of objects more than once. Conversely, doing the parser rewrite without a change to error reporting would mean work to carry the positions along for all places that require them. So I kindly ask you to review these changes together.

The code is fresh out of my rebase. It passes all tests. Many of the tests I modified, particularly those about error reporting, but overall (perhaps with some *very* few exceptions) I think that the quality of the error reporting has improved by this. Apart from passing tests, I'm far from convinced myself that the code is ready for merging in its current form. But I'd like to discuss my concerns, and any concerns you may have, in order to polish the code up to a mergeable form. The concerns I currently have are the following:

* Are error positions zero-based or one-based? Current code has zero-based internally and in the `position` property of the error object, but one-based in the message string. This felt most consistent with the previous situation, where many errors were underlined off by one.
* Do we still want to pass position information to functions or environments, or should it be enough that groups and tokens will carry position information?
* I left the hierarchy of test cases in `errors-spec.js` unmodified, even when I moved some of the code from the lexer to the parser. Should we adjust the file, even if doing so will make the diff harder to read? Or should we clean this up in a subsequent PR?

### DIFF
--- a/src/Lexer.js
+++ b/src/Lexer.js
@@ -21,7 +21,25 @@ function Lexer(input) {
     this.pos = 0;
 }
 
-// The resulting token returned from `lex`.
+/**
+ * The resulting token returned from `lex`.
+ *
+ * It consists of the token text plus some position information.
+ * The position information is essentially a range in an input string,
+ * but instead of referencing the bare input string, we refer to the lexer.
+ * That way it is possible to attach extra metadata to the input string,
+ * like for example a file name or similar.
+ *
+ * The position information (all three parameters) is optional,
+ * so it is OK to construct synthetic tokens if appropriate.
+ * Not providing available position information may lead to
+ * degraded error reporting, though.
+ *
+ * @param {string}  text   the text of this token
+ * @param {number=} start  the start offset, zero-based inclusive
+ * @param {number=} end    the end offset, zero-based exclusive
+ * @param {Lexer=}  lexer  the lexer which in turn holds the input string
+ */
 function Token(text, start, end, lexer) {
     this.text = text;
     this.start = start;
@@ -29,6 +47,13 @@ function Token(text, start, end, lexer) {
     this.lexer = lexer;
 }
 
+/**
+ * Given a pair of tokens (this and endToken), compute a “Token” encompassing
+ * the whole input range enclosed by these two.
+ *
+ * @param {Token}  endToken  last token of the range, inclusive
+ * @param {string} text      the text of the newly constructed token
+ */
 Token.prototype.range = function(endToken, text) {
     return new Token(text, this.start, endToken.end, this.lexer);
 };
--- a/src/Lexer.js
+++ b/src/Lexer.js
@@ -55,6 +55,9 @@ function Token(text, start, end, lexer) {
  * @param {string} text      the text of the newly constructed token
  */
 Token.prototype.range = function(endToken, text) {
+    if (endToken.lexer !== this.lexer) {
+        return new Token(text); // sorry, no position information available
+    }
     return new Token(text, this.start, endToken.end, this.lexer);
 };
 
--- a/src/Lexer.js
+++ b/src/Lexer.js
@@ -17,15 +17,21 @@ var ParseError = require("./ParseError");
 
 // The main lexer class
 function Lexer(input) {
-    this._input = input;
+    this.input = input;
+    this.pos = 0;
 }
 
 // The resulting token returned from `lex`.
-function Token(text, data, position) {
+function Token(text, start, end, lexer) {
     this.text = text;
-    this.data = data;
-    this.position = position;
+    this.start = start;
+    this.end = end;
+    this.lexer = lexer;
 }
+
+Token.prototype.range = function(endToken, text) {
+    return new Token(text, this.start, endToken.end, this.lexer);
+};
 
 /* The following tokenRegex
  * - matches typical whitespace (but not NBSP etc.) using its first group
@@ -52,111 +58,26 @@ var tokenRegex = new RegExp(
     ")"
 );
 
-var whitespaceRegex = /\s*/;
-
 /**
- * This function lexes a single normal token. It takes a position and
- * whether it should completely ignore whitespace or not.
+ * This function lexes a single token.
  */
-Lexer.prototype._innerLex = function(pos, ignoreWhitespace) {
-    var input = this._input;
+Lexer.prototype.lex = function() {
+    var input = this.input;
+    var pos = this.pos;
     if (pos === input.length) {
-        return new Token("EOF", null, pos);
+        return new Token("EOF", pos, pos, this);
     }
     var match = matchAt(tokenRegex, input, pos);
     if (match === null) {
         throw new ParseError(
             "Unexpected character: '" + input[pos] + "'",
-            this, pos);
-    } else if (match[2]) { // matched non-whitespace
-        return new Token(match[2], null, pos + match[2].length);
-    } else if (ignoreWhitespace) {
-        return this._innerLex(pos + match[1].length, true);
-    } else { // concatenate whitespace to a single space
-        return new Token(" ", null, pos + match[1].length);
+            new Token(input[pos], pos, pos + 1, this));
     }
-};
-
-// A regex to match a CSS color (like #ffffff or BlueViolet)
-var cssColor = /#[a-z0-9]+|[a-z]+/i;
-
-/**
- * This function lexes a CSS color.
- */
-Lexer.prototype._innerLexColor = function(pos) {
-    var input = this._input;
-
-    // Ignore whitespace
-    var whitespace = matchAt(whitespaceRegex, input, pos)[0];
-    pos += whitespace.length;
-
-    var match;
-    if ((match = matchAt(cssColor, input, pos))) {
-        // If we look like a color, return a color
-        return new Token(match[0], null, pos + match[0].length);
-    } else {
-        throw new ParseError("Invalid color", this, pos);
-    }
-};
-
-// A regex to match a dimension. Dimensions look like
-// "1.2em" or ".4pt" or "1 ex"
-var sizeRegex = /(-?)\s*(\d+(?:\.\d*)?|\.\d+)\s*([a-z]{2})/;
-
-/**
- * This function lexes a dimension.
- */
-Lexer.prototype._innerLexSize = function(pos) {
-    var input = this._input;
-
-    // Ignore whitespace
-    var whitespace = matchAt(whitespaceRegex, input, pos)[0];
-    pos += whitespace.length;
-
-    var match;
-    if ((match = matchAt(sizeRegex, input, pos))) {
-        var unit = match[3];
-        // We only currently handle "em" and "ex" units
-        if (unit !== "em" && unit !== "ex") {
-            throw new ParseError("Invalid unit: '" + unit + "'", this, pos);
-        }
-        return new Token(match[0], {
-            number: +(match[1] + match[2]),
-            unit: unit,
-        }, pos + match[0].length);
-    }
-
-    throw new ParseError("Invalid size", this, pos);
-};
-
-/**
- * This function lexes a string of whitespace.
- */
-Lexer.prototype._innerLexWhitespace = function(pos) {
-    var input = this._input;
-
-    var whitespace = matchAt(whitespaceRegex, input, pos)[0];
-    pos += whitespace.length;
-
-    return new Token(whitespace[0], null, pos);
-};
-
-/**
- * This function lexes a single token starting at `pos` and of the given mode.
- * Based on the mode, we defer to one of the `_innerLex` functions.
- */
-Lexer.prototype.lex = function(pos, mode) {
-    if (mode === "math") {
-        return this._innerLex(pos, true);
-    } else if (mode === "text") {
-        return this._innerLex(pos, false);
-    } else if (mode === "color") {
-        return this._innerLexColor(pos);
-    } else if (mode === "size") {
-        return this._innerLexSize(pos);
-    } else if (mode === "whitespace") {
-        return this._innerLexWhitespace(pos);
-    }
+    var text = match[2] || " ";
+    var start = this.pos;
+    this.pos += match[0].length;
+    var end = this.pos;
+    return new Token(text, start, end, this);
 };
 
 module.exports = Lexer;
--- a/src/MacroExpander.js
+++ b/src/MacroExpander.js
@@ -13,8 +13,7 @@ function MacroExpander(input, macros) {
 }
 
 /**
- * Expand first token while that is expandable,
- * return first non-expandable token that remains.
+ * Recursively expand first token, then return first non-expandable token.
  */
 MacroExpander.prototype.nextToken = function() {
     for (;;) {
@@ -58,7 +57,7 @@ MacroExpander.prototype.get = function(ignoreSpace) {
  * Undo the effect of the preceding call to the get method.
  * A call to this method MUST be immediately preceded and immediately followed
  * by a call to get.  Only used during mode switching, i.e. after one token
- * was got in the old mode but should bet got again in a new mode
+ * was got in the old mode but should get got again in a new mode
  * with possibly different whitespace handling.
  */
 MacroExpander.prototype.unget = function(token) {
--- a/src/MacroExpander.js
+++ b/src/MacroExpander.js
@@ -7,20 +7,24 @@ var Lexer = require("./Lexer");
 
 function MacroExpander(input, macros) {
     this.lexer = new Lexer(input);
-    this.macros = macros || {};
+    this.macros = macros;
     this.stack = []; // contains tokens in REVERSE order
     this.discardedWhiteSpace = [];
 }
 
+/**
+ * Expand first token while that is expandable,
+ * return first non-expandable token that remains.
+ */
 MacroExpander.prototype.nextToken = function() {
     for (;;) {
         if (this.stack.length === 0) {
             this.stack.push(this.lexer.lex());
         }
-        var top = this.stack.pop();
-        var name = top.text;
+        var topToken = this.stack.pop();
+        var name = topToken.text;
         if (!(name.charAt(0) === "\\" && this.macros.hasOwnProperty(name))) {
-            return top;
+            return topToken;
         }
         var expansion = this.macros[name];
         if (typeof expansion === "string") {
@@ -31,7 +35,7 @@ MacroExpander.prototype.nextToken = function() {
                 expansion.push(tok);
                 tok = bodyLexer.lex();
             }
-            expansion.reverse();
+            expansion.reverse(); // to fit in with stack using push and pop
             this.macros[name] = expansion;
         }
         this.stack = this.stack.concat(expansion);
@@ -50,6 +54,13 @@ MacroExpander.prototype.get = function(ignoreSpace) {
     return token;
 };
 
+/**
+ * Undo the effect of the preceding call to the get method.
+ * A call to this method MUST be immediately preceded and immediately followed
+ * by a call to get.  Only used during mode switching, i.e. after one token
+ * was got in the old mode but should bet got again in a new mode
+ * with possibly different whitespace handling.
+ */
 MacroExpander.prototype.unget = function(token) {
     this.stack.push(token);
     while (this.discardedWhiteSpace.length !== 0) {
--- a/src/MacroExpander.js
+++ b/src/MacroExpander.js
@@ -1,0 +1,60 @@
+/**
+ * This file contains the “gullet” where macros are expanded
+ * until only non-macro tokens remain.
+ */
+
+var Lexer = require("./Lexer");
+
+function MacroExpander(input, macros) {
+    this.lexer = new Lexer(input);
+    this.macros = macros || {};
+    this.stack = []; // contains tokens in REVERSE order
+    this.discardedWhiteSpace = [];
+}
+
+MacroExpander.prototype.nextToken = function() {
+    for (;;) {
+        if (this.stack.length === 0) {
+            this.stack.push(this.lexer.lex());
+        }
+        var top = this.stack.pop();
+        var name = top.text;
+        if (!(name.charAt(0) === "\\" && this.macros.hasOwnProperty(name))) {
+            return top;
+        }
+        var expansion = this.macros[name];
+        if (typeof expansion === "string") {
+            var bodyLexer = new Lexer(expansion);
+            expansion = [];
+            var tok = bodyLexer.lex();
+            while (tok.text !== "EOF") {
+                expansion.push(tok);
+                tok = bodyLexer.lex();
+            }
+            expansion.reverse();
+            this.macros[name] = expansion;
+        }
+        this.stack = this.stack.concat(expansion);
+    }
+};
+
+MacroExpander.prototype.get = function(ignoreSpace) {
+    this.discardedWhiteSpace = [];
+    var token = this.nextToken();
+    if (ignoreSpace) {
+        while (token.text === " ") {
+            this.discardedWhiteSpace.push(token);
+            token = this.nextToken();
+        }
+    }
+    return token;
+};
+
+MacroExpander.prototype.unget = function(token) {
+    this.stack.push(token);
+    while (this.discardedWhiteSpace.length !== 0) {
+        this.stack.push(this.discardedWhiteSpace.pop());
+    }
+};
+
+module.exports = MacroExpander;
--- a/src/ParseError.js
+++ b/src/ParseError.js
@@ -14,7 +14,7 @@ function ParseError(message, token) {
     var start;
     var end;
 
-    if (token && token.lexer) {
+    if (token && token.lexer && token.start <= token.end) {
         // If we have the input and a position, make the error a bit fancier
 
         // Get the input
--- a/src/ParseError.js
+++ b/src/ParseError.js
@@ -2,6 +2,12 @@
  * This is the ParseError class, which is the main error thrown by KaTeX
  * functions when something has gone wrong. This is used to distinguish internal
  * errors from errors in the expression that the user provided.
+ *
+ * If possible, a caller should provide a Token or ParseNode with information
+ * about where in the source string the problem occurred.
+ *
+ * @param {string} message  The error message
+ * @param {(Token|ParseNode)=} token  An object providing position information
  */
 function ParseError(message, token) {
     var error = "KaTeX parse error: " + message;
--- a/src/ParseError.js
+++ b/src/ParseError.js
@@ -3,25 +3,43 @@
  * functions when something has gone wrong. This is used to distinguish internal
  * errors from errors in the expression that the user provided.
  */
-function ParseError(message, lexer, position) {
+function ParseError(message, token) {
     var error = "KaTeX parse error: " + message;
+    var start;
+    var end;
 
-    if (lexer !== undefined && position !== undefined) {
+    if (token && token.lexer) {
         // If we have the input and a position, make the error a bit fancier
 
-        // Prepend some information
-        error += " at position " + position + ": ";
-
         // Get the input
-        var input = lexer._input;
-        // Insert a combining underscore at the correct position
-        input = input.slice(0, position) + "\u0332" +
-            input.slice(position);
+        var input = token.lexer.input;
+
+        // Prepend some information
+        start = token.start;
+        end = token.end;
+        if (start === input.length) {
+            error += " at end of input: ";
+        } else {
+            error += " at position " + (start + 1) + ": ";
+        }
+
+        // Underline token in question using combining underscores
+        var underlined = input.slice(start, end).replace(/[^]/g, "$&\u0332");
 
         // Extract some context from the input and add it to the error
-        var begin = Math.max(0, position - 15);
-        var end = position + 15;
-        error += input.slice(begin, end);
+        var left;
+        if (start > 15) {
+            left = "…" + input.slice(start - 15, start);
+        } else {
+            left = input.slice(0, start);
+        }
+        var right;
+        if (end + 15 < input.length) {
+            right = input.slice(end, end + 15) + "…";
+        } else {
+            right = input.slice(end);
+        }
+        error += left + underlined + right;
     }
 
     // Some hackery to make ParseError a prototype of Error
@@ -30,7 +48,7 @@ function ParseError(message, lexer, position) {
     self.name = "ParseError";
     self.__proto__ = ParseError.prototype;
 
-    self.position = position;
+    self.position = start;
     return self;
 }
 
--- a/src/Parser.js
+++ b/src/Parser.js
@@ -46,8 +46,9 @@ var ParseError = require("./ParseError");
  * Main Parser class
  */
 function Parser(input, settings) {
-    // Make a new lexer
-    this.gullet = new MacroExpander(input, settings && settings.macros);
+    // Make a new lexer (mouth) and macro expander (gullet)
+    // for this parser (stomach, in the language of TeX)
+    this.gullet = new MacroExpander(input, settings.macros);
     // Store the settings for use in parsing
     this.settings = settings;
 }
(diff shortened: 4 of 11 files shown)


gagern:
> It passes all tests.

It passes all *jasmine* tests. It fails two screenshotter tests, namely DelimiterSizing and LeftRightStyleSizing. Need to investigate further.

gagern:
> It fails two screenshotter tests[…]

False alarm: These apparently were due to incomplete page loads, and incomplete fonts in particular. With the `--verify` option to `screenshotter.sh` I could verify that all tests eventually render as expected. Which is what Travis was running, too, as I should have recalled.

cbreeden:
> Doing the error reporting changes without the restructuring to the parser means doing error reporting for certain kinds of objects more than once. Conversely, doing the parser rewrite without a change to error reporting would mean work to carry the positions along for all places that require them.

I was playing around yesterday with a parser refactor to simplify the call hierarchy.  How would you ideally handle error reporting if such a refactor did occur?

cbreeden:
Essentially the API for the parsing I had in mind, boils down to doing doing stuff like this:

- Flatten call stacks by defining `parseSymbol`, `parseFunction`, `parseGroup`, and `parseImplicitGroup` to do what they sound like they do, only returning something if they math.
- Parsing for an atom essentially becomes:
```
var atom = parseSymbol() || parseGroup() || parseImplicitGroup() || parseFunction();
```
- Parsing for function arguments would become:
```
var argument = parseSymbol() || parseGroup() || tryFunctionExpand(baseGreediness);
```
where tryFunctionExpand simply just checks to see if the function is greedier and returns it if so, and does error handling otherwise.

Hopefully I can finish this prototype soon.  Would this API help with error reporting for you do you think?

gagern:
@cbreeden: I'm not sure what the goal of that restructuring would be, what problem it would solve. To me it sounds as if it would move complexity from the single forwarding call (e.g. `parseGroup` calling `parseSymbol`) and instead move that same complexity into *all* the places that currently invoke `parseGroup`. Leading to quite a bit of code duplication, and maintainance work those `||` chains in sync in case some other construct gets added. So at the moment I'm not convinced that this is a good idea. But perhaps it's best if you formulate your idea in a separate issue, and we continue discussion there. That's what I did for my parser restructuring in #266.

Anyway, regarding error reporting I don't feel like this has any impact one way or another. Ideally I'd like for every parse node to carry information about its lexer, start and end position. I've already done this in part here, for simple groups and individual symbols, but not for some of the more exotic constructs.

src/Parser.js
```diff
@@ -600,80 +593,148 @@ Parser.prototype.parseArguments = function(func, funcData) {
...
+        false);
+};
+
+/**
+ * Parses a size specification, consisting of magnitude and unit.
+ */
+Parser.prototype.parseSizeGroup = function(optional) {
+    var res = this.parseSpecialGroup(
+        "size", optional,
+        /^ *(-?) *(?:(?:\d+(?:\.\d*)?|\.\d+) *[a-z]{0,2} *)?$/,
```
cbreeden: I am under the impression that if `parseSpecialGroup` simply handle the unwrapping of brackets, switching modes and returned a string, and kept the matching logic in `parseSizeGroup` and `parseColorGroup` we could simplify the logic here.

gagern: My goal was to have the error reporting in one place, and also to catch ill-formed special groups pretty early, thanks to that `reOnline` parameter. But I do confess that I see no compelling reason to do so. If one were to do this naively, we might get strange error messages: `\rule{2{cm}}` would complain abouut `2{cm` being an invalid length, instead of `2{…` being an invalid length. But I don't think this is a real problem, and wouldn't try to support nested groups unless there was some good reason to do so.

src/ParseError.js
```diff
@@ -3,25 +3,43 @@
...
  * errors from errors in the expression that the user provided.
  */
-function ParseError(message, lexer, position) {
+function ParseError(message, token) {
     var error = "KaTeX parse error: " + message;
+    var start;
+    var end;
 
-    if (lexer !== undefined && position !== undefined) {
+    if (token && token.lexer) {
```
kevinbarabash: is `lexer` an optional param to the `Token` constructor?

kevinbarabash:
I'll do a closer read of this this afternoon.  Lots of changes going on here.

src/ParseError.js
gagern: Yes. Unless I'm missing something, I'm not *actually* constructing tokens without a lexer reference. But I envision that there *might* be situations where it makes sense to construct artificial tokens which don't actually occur in source.

Furthermore, the parse node which [results from a `parseGroup` call](https://github.com/Khan/KaTeX/pull/493/files/e3dda607afd87a8e6b62fc4135a8f4d50cbee697#diff-a0c375bdbb580536aa70722d5d376137L676) will also behave like a token, in that [it will contain `start`, `end` and `lexer` properties](https://github.com/Khan/KaTeX/pull/493/files/e3dda607afd87a8e6b62fc4135a8f4d50cbee697#diff-1e3e2449fa5a2f1a38c89076216368ceR8). And [some](https://github.com/Khan/KaTeX/pull/493/files/e3dda607afd87a8e6b62fc4135a8f4d50cbee697#diff-af9da43267175de02306a3e75d6c1e88R107) [calls](https://github.com/Khan/KaTeX/pull/493/files/e3dda607afd87a8e6b62fc4135a8f4d50cbee697#diff-0ce252df8a9f3b1e09f8552116f0abbeR580) to the `ParseError` constructor pass in a parse node, not a raw token, so I'm not perfectly confident that every group that might end up in such a setup will actually contain the required source position information.

Do you think I should name the argument differently, since it's not only tokens? If so, would `tokenOrNode` or `pos` be a better name?

kevinbarabash: I think this is fine, it's probably worth add a comment to the Token constructor which points out that `lexer` is optional and why that's desirable.

src/Lexer.js
```diff
@@ -52,111 +58,26 @@ var tokenRegex = new RegExp(
...
-        return this._innerLexColor(pos);
-    } else if (mode === "size") {
-        return this._innerLexSize(pos);
-    } else if (mode === "whitespace") {
-        return this._innerLexWhitespace(pos);
+            new Token(input[pos], pos, pos + 1, this));
     }
+    var text = match[2] || " ";
+    var start = this.pos;
+    this.pos += match[0].length;
```
kevinbarabash: shouldn't this be `match[2].length`?

src/MacroExpander.js
```diff
@@ -0,0 +1,60 @@
...
+var Lexer = require("./Lexer");
+
+function MacroExpander(input, macros) {
+    this.lexer = new Lexer(input);
+    this.macros = macros || {};
+    this.stack = []; // contains tokens in REVERSE order
+    this.discardedWhiteSpace = [];
+}
+
+MacroExpander.prototype.nextToken = function() {
```
kevinbarabash: It be good to have a comment here about how this function expands macros until it can't and then returns the next atom.  I was thinking it might be clearer to have a separate function that just does the expansion art, thoughts?

src/MacroExpander.js
```diff
@@ -0,0 +1,60 @@
...
+        var expansion = this.macros[name];
+        if (typeof expansion === "string") {
+            var bodyLexer = new Lexer(expansion);
+            expansion = [];
+            var tok = bodyLexer.lex();
+            while (tok.text !== "EOF") {
+                expansion.push(tok);
+                tok = bodyLexer.lex();
+            }
+            expansion.reverse();
```
kevinbarabash: Why is the `expansion` reverse?  This is probably worth a comment.

src/MacroExpander.js
kevinbarabash: It seems like `nextToken` could expand everything all at once.  Is the goal to avoid having to parse things twice?

src/Lexer.js
gagern: Definitely not. If the input is a long sequence of whitespace, you want to emit a single space token, but you want to advance the input position past the whole sequence. `match[2]` would not be set at all in a match against whitespace, hence taking its length would likely throw an exception.

src/parseData.js
```diff
@@ -1,10 +1,15 @@
...
-function ParseNode(type, value, mode) {
+function ParseNode(type, value, mode, firstToken, lastToken) {
     this.type = type;
     this.value = value;
     this.mode = mode;
+    if (firstToken) {
+        this.lexer = firstToken.lexer;
+        this.start = firstToken.start;
+        this.end = (lastToken || firstToken).end;
+    }
```
kevinbarabash: I'm confused by this.  Why does a `ParseNode` need a reference to `firstToken`'s lexer?

src/MacroExpander.js
```diff
@@ -0,0 +1,60 @@
...
+    if (ignoreSpace) {
+        while (token.text === " ") {
+            this.discardedWhiteSpace.push(token);
+            token = this.nextToken();
+        }
+    }
+    return token;
+};
+
+MacroExpander.prototype.unget = function(token) {
```
kevinbarabash: If you try to do more than one `unget` in a row we might end up with messed up whitespace.  Is that something we should be concerned with?  Regardless, it's probably worth writing a comment about it.

src/Parser.js
```diff
@@ -47,7 +47,7 @@ var ParseError = require("./ParseError");
  */
 function Parser(input, settings) {
     // Make a new lexer
-    this.lexer = new Lexer(input);
+    this.gullet = new MacroExpander(input, settings && settings.macros);
```
kevinbarabash: It's probably worth putting a comment here too about why it's called the `gullet`.

src/MacroExpander.js
gagern: A comment is definitely a good idea. I think the term “atom” is already used with a different meaning inside the parser. I'm not sure what the correct TeX term for a non-expandable piece of input is, perhaps I should look that up.

I guess I'd rather not expand everything at once because one day we will get definitions. And for that to work, we need to change the set of known macros before we continue parsing. Defining macros is technically inside the gullet, I guess, so it would be possible to have some API which returns the whole expanded token stream. But implementing that is easier if we have *some* point where the tokens pass through one at a time, and definitions can take effect for subsequent tokens. And since the parser consumes tokens one at a time, there is little point in having a “convert-everything-at-once” layer in there at all.

Plus it might happen that in order to get scoping right, i.e. the distinction between `\def` and `\gdef`, we would one day welcome help from the parser to tell us when one group ends so that we can close its scope. I don't say that we should take this approach, but having it as an option feels valuable somehow.

src/MacroExpander.js
```diff
@@ -0,0 +1,60 @@
...
+    this.stack = []; // contains tokens in REVERSE order
+    this.discardedWhiteSpace = [];
+}
+
+MacroExpander.prototype.nextToken = function() {
+    for (;;) {
+        if (this.stack.length === 0) {
+            this.stack.push(this.lexer.lex());
+        }
+        var top = this.stack.pop();
```
kevinbarabash: can you rename `top` to `topToken`?

src/MacroExpander.js
gagern: Well, the whole set of current tokens is a stack, and while the `push` and `pop` methods operate on the tail of the JavaScript array, they operate on the head of the TeX token sequence. I guess that `unshift` and `shift` are more costly in most implementations, thus reversing macro bodies when they first get lexed seemed cheaper, performance-wise. Would the following be an appropriate comment to sum this up?

```js
expansion.reverse(); // to fit in with tail-modifying stack
```

src/MacroExpander.js
gagern: >  I was thinking it might be clearer to have a separate function that just does the expansion art, thoughts?

I think the function *is* just doing expansion, plus parsing of macro bodies from strings to token arrays. The latter coulc conceivably be moved to a separate function, `getMacroBody` or whatever, if you prefer. I have no strong feelings either way, as long as it's all internal API so we can change it when the need arises.

src/MacroExpander.js
gagern: `unget` is only ever called when mode-switching, so it ungets one token immediately before requesting the next token, which will either be the same token again, or the token after the whitespace if the token was a whitespace token and the new mode ignores whitespace. So no, we need not be concerned, but yes, a comment to this effect is definitely a good idea.

src/Parser.js
```diff
@@ -137,31 +142,25 @@ Parser.prototype.parseExpression = function(breakOnInfix, breakOnToken) {
...
-        var pos = this.pos;
         if (endOfExpression.indexOf(lex.text) !== -1) {
             break;
         }
         if (breakOnToken && lex.text === breakOnToken) {
             break;
         }
+        if (breakOnInfix && functions[lex.text] && functions[lex.text].infix) {
+            break;
+        }
```
kevinbarabash: The only case in which `breakOnToken` is actually used is when we we're parsing a special group.  What situation does this handle?

src/Parser.js
gagern: Either that, or call the property `macroExpander`, `me` or something like that. Preferences?

src/Parser.js
```diff
@@ -137,31 +142,25 @@ Parser.prototype.parseExpression = function(breakOnInfix, breakOnToken) {
     // we reached the end, a }, or a \right)
     while (true) {
         var lex = this.nextToken;
-        var pos = this.pos;
         if (endOfExpression.indexOf(lex.text) !== -1) {
             break;
         }
         if (breakOnToken && lex.text === breakOnToken) {
```
kevinbarabash: We should probably rename `breakOnToken` to `breakOnChar` so that the word "token" doesn't become too overloaded.

src/ParseError.js
```diff
@@ -3,25 +3,43 @@
  * functions when something has gone wrong. This is used to distinguish internal
  * errors from errors in the expression that the user provided.
  */
-function ParseError(message, lexer, position) {
+function ParseError(message, token) {
```
kevinbarabash: nice cleanup

src/parseData.js
gagern: This is all about error reporting. Ideally, we should be able to take an arbitrary piece of parse tree and highlight its corresponding source region. For this we need the start and end positions as well as the input string. And the input string is stored in the lexer, thus we need the lexer.

We might conceivably use just the input string instead of the lexer, both here and for `Token`. But using the lexer allows us to add metadata to a source string at some later point. Things like file name, or some other description of where this string came from. For example “in definition of macro …” for source regions coming from an external macro definition string would already make a lot of sense, and be fairly easy with the current approach.

src/Parser.js
```diff
@@ -600,80 +593,136 @@ Parser.prototype.parseArguments = function(func, funcData) {
...
 
 /**
- * Parses a group when the mode is changing. Takes a position, a new mode, and
- * an outer mode that is used to parse the outside.
+ * Parses a group when the mode is changing.
  *
  * @return {?ParseFuncOrArgument}
  */
-Parser.prototype.parseSpecialGroup = function(innerMode, optional) {
+Parser.prototype.parseGroupOfType = function(innerMode, optional) {
```
kevinbarabash: :+1:

src/Parser.js
gagern: But we break if the next *token* has the given text. True, currently we only use this for `]` which is a one-char token. But we couldn't break on `\` but we could break on `\\`, so token is more fitting. We could use `breakOnTokenText` to distinguish this from the token class, if you prefer.

src/Parser.js
gagern: This is the replacement for the rewininding which I deleted a few lines down. Before this change, if the next *atom* was of type infix, then we rewound the input stream to parse everything a second time. Now I'd like to concentrate on linear operations, with no jumping around in the input stream. Thus we have to detect an infix token before it's parsed to an atom. Thus the check above, and the addition of the matching metadata to the table of functions.

src/Parser.js
```diff
@@ -600,80 +593,136 @@ Parser.prototype.parseArguments = function(func, funcData) {
...
 };
 
 /**
+ * Parses a group, essentially returning the string formed by the
+ * brace-enclosed tokens plus some position information.
+ *
+ * @param modeName {string} Used to describe the mode in error messages
+ * @param optional {boolean} Whether the group is optional or required
+ */
+Parser.prototype.parseSpecialGroup = function(modeName, optional) {
```
kevinbarabash: `parseSpecialGroup` used to do something different which may be confusing for people who have worked in the code base previously... maybe `parseStringGroup`?

src/Parser.js
```diff
@@ -600,80 +593,136 @@ Parser.prototype.parseArguments = function(func, funcData) {
...
+
+/**
+ * Parses a size specification, consisting of magnitude and unit.
+ */
+Parser.prototype.parseSizeGroup = function(optional) {
+    var res = this.parseSpecialGroup("size", optional);
+    if (!res) {
+        return null;
+    }
+    var match = (/(-?) *(\d+(?:\.\d*)?|\.\d+) *([a-z]{2})/).exec(res.text);
```
kevinbarabash: I see you catch unsupported units below, nice!

src/Parser.js
```diff
@@ -600,80 +593,136 @@ Parser.prototype.parseArguments = function(func, funcData) {
...
+    if (!match) {
+        throw new ParseError("Invalid size: '" + res.text + "'", res);
+    }
+    var data = {
+        number: +(match[1] + match[2]),
```
kevinbarabash: // +(sign + number)

src/Settings.js
```diff
@@ -23,6 +23,7 @@ function Settings(options) {
     this.displayMode = get(options.displayMode, false);
     this.throwOnError = get(options.throwOnError, true);
     this.errorColor = get(options.errorColor, "#cc0000");
+    this.macros = options.macros;
```
kevinbarabash: Could you provide a default here so we don't have to check if `settings.macros` is defined when creating the `MacroExpander`?

static/main.js
```diff
@@ -10,25 +10,36 @@ function init() {
...
     }
 
     var match = (/(?:^\?|&)text=([^&]*)/).exec(window.location.search);
     if (match) {
         input.value = decodeURIComponent(match[1]);
     }
 
+    var macros = {};
+    var options = {};
+    var reMacro = /(?:^\?|&)(?:\\|%5[Cc])([A-Za-z]+)=([^&]*)/g;
```
kevinbarabash: `macroRegex` to follow the existing naming convention

test/errors-spec.js
```diff
@@ -61,16 +61,15 @@ beforeEach(function() {
 describe("Parser:", function() {
 
     describe("#handleInfixNodes", function() {
-        // TODO: The position information here is broken, should be fixed.
         it("rejects repeated infix operators", function() {
             expect("1\\over 2\\over 3").toFailWithParseError(
-                   "only one infix operator per group at position -1: " +
-                   "1\\over 2\\over ");
+                   "only one infix operator per group at position 9: " +
+                   "1\\over 2\\̲o̲v̲e̲r̲ 3");
```
kevinbarabash: I don't get how this works, but cool!

test/errors-spec.js
```diff
@@ -91,84 +90,85 @@ describe("Parser:", function() {
     });
 
     describe("#parseAtom", function() {
-        // TODO: The positions in the following error messages appear to be
-        // off by one, i.e. they should be one character later.
         it("rejects \\limits without operator", function() {
             expect("\\alpha\\limits\\omega").toFailWithParseError(
                    "Limit controls must follow a math operator" +
-                   " at position 6: \\alpha̲\\limits\\omega");
+                   " at position 7: \\alpha\\̲l̲i̲m̲i̲t̲s̲\\omega");
```
kevinbarabash: why are the position numbers different?

test/errors-spec.js
```diff
@@ -224,32 +224,42 @@ describe("Parser.expect calls:", function() {
         });
         // Can't test for the [ of an optional group since it's optional
         it("complains about missing } for color", function() {
-            expect("\\color{#ffffff {text}").toFailWithParseError(
-                   "Expected '}', got '{' at position 16:" +
-                   " color{#ffffff {̲text}");
+            expect("\\color{#ffffff{text}").toFailWithParseError(
+                   "Invalid color: '#ffffff{text' at position 8:" +
```
kevinbarabash: nice 👍

test/errors-spec.js
```diff
@@ -311,34 +320,35 @@ describe("Lexer:", function() {
...
     describe("#_innerLexSize", function() {
         it("reject size without unit", function() {
             expect("\\rule{0}{2em}").toFailWithParseError(
-                   "Invalid size at position 6: \\rule{̲0}{2em}");
+                   "Invalid size: '0' at position 7: \\rule{0̲}{2em}");
         });
         it("reject size with bogus unit", function() {
             expect("\\rule{1au}{2em}").toFailWithParseError(
-                   "Invalid unit: 'au' at position 6: \\rule{̲1au}{2em}");
+                   "Invalid unit: 'au' at position 7: \\rule{1̲a̲u̲}{2em}");
```
kevinbarabash: someday ;)

test/katex-spec.js
```diff
@@ -57,8 +57,39 @@ var getParsed = function(expr, settings) {
...
+    if (expr.lexer && typeof expr.start === "number") {
+        delete expr.lexer;
+        delete expr.start;
+        delete expr.end;
+    }
+    Object.keys(expr).forEach(function(key) {
+        stripPositions(expr[key]);
+    });
+    return expr;
+};
```
kevinbarabash: can you put a comment and a TODO to make this better... not quite sure what to do here so we can punt on this

src/Settings.js
gagern: The `MacroExpander` constructor already provides a default. The `Parser` constructor passes `settings && settings.macros`. I guess that the situation where `settings` is undefined shouldn't really be possible with out public API. So I could just pass on `settings.macros` unconditionally, and avoid the default in `MacroExpander` as well. Yes, this should work nicely.

test/katex-spec.js
```diff
@@ -1802,3 +1857,24 @@ describe("The symbol table integraty", function() {
...
+    };
+
+    it("should produce individual tokens", function() {
+        compareParseTree("e^\\foo", "e^1 23", {"\\foo": "123"});
+    });
+
+    it("should allow for multiple expansion", function() {
+        compareParseTree("1\\foo2", "1aa2", {
+            "\\foo": "\\bar\\bar",
+            "\\bar": "a",
```
kevinbarabash: nice!

src/MacroExpander.js
kevinbarabash: I don't have super strong feelings either so let's leave it as is.

src/MacroExpander.js
kevinbarabash: b/c we have to `pop` to get the next once which would be the last one w/o the reverse.  got it

test/errors-spec.js
gagern: Well, the old `ParseError` implementation did take the first `position` characters, and then add a combining underscore to that. Which means that for highlighting it assumed its positions to be one-based: you'd have to take one character if you wanted to add an underscore to the first character of the string. Probably due to lack of testing, many error messages returned the zero-based position of the problem, though, since zero-based indexing is natural in JavaScript. I left the one-based indexing in the error message for now, but tried to fix the internal positions to actually highlight where the problem is, not off by one. If you look at the deleted Parser code, you will notice that *some* of the `ParseError` constructor calls have a `+ 1` in their position arguments, while others do not.

src/parseData.js
kevinbarabash: Cool.  It's probably worth having that in a comment.

src/Parser.js
kevinbarabash: I kind of like the `gullet`.  It's something that I think is unique to the way TeX operates and including that terminology in KaTeX is a great way to pay homage.  ^_^

src/Parser.js
kevinbarabash: Cool!  Nice optimization.

test/errors-spec.js
gagern: Glad you like this change.

Personally I liked the old one better. Which is why prior to the e3dda607afd87a8e6b62fc4135a8f4d50cbee697 suggested by @cbreeden it would report the invalid color as `'#ffffff{…'`, reporting the error in the same place as the old code did. But if you like the new message, that confirms that the code can stay as it is, with the simplifications from the commit I just mentioned.

src/Parser.js
kevinbarabash: Let's go with `breakOnTokenText`.

test/errors-spec.js
gagern: The key is not just having a single position, but a range formed by a pair of positions. With that, underlining each character is a piece of cake. Although the actual rendering may show huge differences in quality depending on fonts or font rendering engines.

It *might* be possible to [install some special code](http://bit.ly/object-formatters) to pretty-print KaTeX error messages in an even nicer way for Chrome…

gagern:
@kevinbarabash Thanks for the review. I had the feeling you were mostly done, and incorporated your suggestsions in a new commit. I also have a second commit aiming at making input position ranges more robust, particularly when mixing tokens from different lexers (thanks to macro expansion). So that may need a fresh review, but should be minor.

gagern:
I'd say the Travis error reported in https://travis-ci.org/Khan/KaTeX/builds/140163515 is likely some spurious problem with Docker which I don't understand. https://travis-ci.org/gagern/KaTeX/builds/140163512 builds just fine.

gagern:
@kevinbarabash What's the status here? Waiting for more review on your part? Waiting for a second opinion from someone? Doing some larger-scale testing? Holiday season?

kevinbarabash:
Sorry... last week was really busy.  Thanks for making all of the changes.  I'll do a final once over pass sometime today.

src/MacroExpander.js
```diff
@@ -0,0 +1,71 @@
...
+        }
+    }
+    return token;
+};
+
+/**
+ * Undo the effect of the preceding call to the get method.
+ * A call to this method MUST be immediately preceded and immediately followed
+ * by a call to get.  Only used during mode switching, i.e. after one token
+ * was got in the old mode but should bet got again in a new mode
```
kevinbarabash: ...should *get* got again...

src/MacroExpander.js
```diff
@@ -0,0 +1,71 @@
...
+function MacroExpander(input, macros) {
+    this.lexer = new Lexer(input);
+    this.macros = macros;
+    this.stack = []; // contains tokens in REVERSE order
+    this.discardedWhiteSpace = [];
+}
+
+/**
+ * Expand first token while that is expandable,
+ * return first non-expandable token that remains.
```
kevinbarabash: "Expand the first token and all tokens it generates." might be a bit clearer.

src/Parser.js
```diff
@@ -46,8 +46,9 @@ var ParseError = require("./ParseError");
  * Main Parser class
  */
 function Parser(input, settings) {
-    // Make a new lexer
-    this.lexer = new Lexer(input);
+    // Make a new lexer (mouth) and macro expander (gullet)
+    // for this parser (stomach, in the language of TeX)
+    this.gullet = new MacroExpander(input, settings.macros);
```
kevinbarabash: The comment is a bit confusing b/c it's the MacroExpander that creators the lexer.

src/MacroExpander.js
gagern: But also incorrect. If you have `\def\foo{\bar\bar}\def\bar{x}` then all the tokens `\foo` generates would be `xx` but it only expands this to `x\bar` then returns the `x`without expanding the second `\bar` at this point.

kevinbarabash: The first token `\bar` only generates `x`.  Another way of stating this might be "recursively expand the next token".

gagern:
@kevinbarabash I reworded those comments. Do you want commits squashed?

gagern:
Travis is complaining [again](https://github.com/Khan/KaTeX/pull/493#issuecomment-228516475), but I'm investigating that and will report in another issue, or hopefully a pull request with a fix.

kevinbarabash:
I re-ran the travis tests and they're passing now.

static/main.js
```diff
@@ -10,25 +10,36 @@ function init() {
...
 
+    var macros = {};
+    var options = {};
+    var macroRegex = /(?:^\?|&)(?:\\|%5[Cc])([A-Za-z]+)=([^&]*)/g;
+    var macroString = "";
+    while ((match = macroRegex.exec(window.location.search)) !== null) {
+        options.macros = macros;
+        macros["\\" + match[1]] = decodeURIComponent(match[2]);
+        macroString += "&" + match[0].substr(1);
+    }
```
kevinbarabash: nice addition, I checked out your branch locally to try this out and it works as advertised

src/Settings.js
```diff
@@ -23,6 +23,7 @@ function Settings(options) {
     this.displayMode = get(options.displayMode, false);
     this.throwOnError = get(options.throwOnError, true);
     this.errorColor = get(options.errorColor, "#cc0000");
+    this.macros = options.macros || {};
```
kevinbarabash: thanks :)

kevinbarabash:
Thanks for rewording the comments.  This is good to go now.  This should allow us to start tackling a number of issues that were marked as "needs macros" by changing what the default macros are and merging custom macros in.

arogozhnikov:
Is there any support for macros in auto-render?
Is defining macro in settings currently the only option?

kevinbarabash:
Currently the only way to define a macro is through options.  The function that the auto-render code exports accepts an `options` object as the second parameter so defining `macros` on that object should work.

arogozhnikov:
Hi @kevinbarabash, thanks for pointing.

It didn't work with 0.6 (I see release was before this merge), so I as I understand current stable code doesn't have define / macro or any other way to create shortcuts, I'll have to write my own microcode.

kevinbarabash:
@arogozhnikov we haven't done a release with the macro changes yet.  Please try `master`.

edemaine:
0.7.0 has macro support (just tested).  Some documentation should be added to README...

gagern:
@edemaine: 0.7.0 isn't available on CDN, probably due to #613. So I'd wait for #615 then wait for a 0.7.1 release then update the readme to refer to the new version and describe the new functionality. I've started a formulation in [readmeMacros](https://github.com/Khan/KaTeX/compare/master...gagern:readmeMacros), though.

edemaine:
@gagern Oops, right.  I'm already using it via NPM, so I forgot.

Nice documentation.  If I might make a suggestion, it could help to mention something like "custom macros" to introduce the concept, before/while talking about the format.  (I don't seem to be able to comment on the diff until it's a PR, so I'll put this here).

gagern:
@edemaine: Good suggestion. Better now? BTW, you can comment on *commits*, and there is a list of commits associated with the diff.

edemaine:
@gagern Looks great! And thanks for the Github tip!  I'm still learning...

librilex:
I would like to add the `\sfrac` macro to KaTeX. So I tried this:
```javascript
katex.render("\\sfrac{1}{2}", div, {macros: {"\\sfrac": "^{#1}\\!/_{#2}"}})
```
But I get the following parse error: `Expected '}', got '#' at position 3: ^{#̲1}\!/_{#2}`. I also tried ## but I still get the same error. How can I define a macro with more than one input?

gagern:
@librilex #605 will add support for macro arguments.

arogozhnikov:
@gagern is support planned also in auto-render?

gagern:
@librilex #656 is working on that.

balthild:
Can we define macros with JS?
For example, `\dv{f}{x}` → `\frac{df}{dx}` and `\dv{f}{x}{y}` → `\frac{d^2f}{dx\,dy}`. We cannot do this with only `#`s.

edemaine:
@balthild It probably makes sense to discuss this in a separate issue, but while it's here: Yes, you can bind a macro to a JavaScript function.  See `src/macros.js` for some examples.

However, you cannot really have a TeX macro with two **or** three arguments like that.  Perhaps you meant `\pdv{f}{x}{x}` in that example?  That would be doable in TeX once we have `\if` conditionals (see #1003).

PR #501, opened 2016-07-08
Add align and align* environments which alias aligned

kevinbarabash:
Summary:
The align/align* environments create their own math environment
and normally aren't allowed within an existing math environment
which KaTeX starts in.  The rendering of align/align* matches
the rendering of aligned with the exception of equation numbering
which KaTeX doesn't support yet.  Allowing users to use these
environments allows copy/pasting of TeX content from other sources
such as TeX documents and math appearing on the web on sites using
MathJax.

Test Plan:
- make test
- open 0.0.0.0:7936 and try the following TeX code:
  `\begin{aligned}x&=y & X&=Y & a&=b+c\\ x'&=y' & X'&=Y' & a'&=b\\ x+x'&=y+y' & X+X'&=Y+Y' & a'b&=c'b\end{aligned}`
- swap out aligned for align and align* and make sure they render the same

### DIFF
--- a/src/environments.js
+++ b/src/environments.js
@@ -184,7 +184,7 @@ defineEnvironment("cases", {
 // except it operates within math mode.
 // Note that we assume \nomallineskiplimit to be zero,
 // so that \strut@ is the same as \strut.
-defineEnvironment("aligned", {
+defineEnvironment(["aligned", "align", "align*"], {
 }, function(context) {
     var res = {
         type: "array",
--- a/test/katex-spec.js
+++ b/test/katex-spec.js
@@ -1704,6 +1704,10 @@ describe("An aligned environment", function() {
     it("should parse its input", function() {
         expect("\\begin{aligned}a&=b&c&=d\\\\e&=f\\end{aligned}")
             .toParse();
+        expect("\\begin{align}a&=b&c&=d\\\\e&=f\\end{align}")
+            .toParse();
+        expect("\\begin{align*}a&=b&c&=d\\\\e&=f\\end{align*}")
+            .toParse();
     });
 
 });


spicyj:
Does this match MathJax behavior?

kevinbarabash:
@spicyj yes it does.

kevinbarabash:
@gagern any thoughts on adding these environment aliases?

gagern:
This PR should mention #61.

If we start shipping `align` as an alias without equation numbers, then we run a certain risk that if we add equation numbers at a later time, it will break existing content which somehow relied on the absence of equation numbers. So I have my doubts about that (and expressed them in https://github.com/Khan/KaTeX/issues/61#issuecomment-138820008).

If we could rely on people reading documentation, I'd say we could simply warn them against using it this way, but they probably wouldn't read that part anyway. Should we make this a config option, like `align_without_equation_numbering: true`, and if that is set we support it now, and keep supporting it without equation numbers in the future?

I was surprised to see that MathJax apparently doesn't number equations inside `align`, at least in their demo box. But in my understanding, KaTeX' reference has been LaTeX, not MathJax, so I'd go for that, particularly since `align*` exists and a mass replacement of `align` by `align*` shuld be quite feasible when migrating content from MathJax to KaTeX. In https://github.com/Khan/KaTeX/issues/61#issuecomment-139084224 @xymostech was unsatisfied with the MathJax approach.

I'm wondering whether we should somehow ensure that such constructs are only allowed at the top level. I.e. `\begin{aligned*}…\end{aligned*} … \begin{aligned*}…\end{aligned*}` should not be permissible input, since we likely won't do the paragraph typesetting between these right so that shouldn't end up in our lap. Perhaps we also should reject this environment in inline mode, since it only makes sense in display mode imo. I already wrote as much in https://github.com/Khan/KaTeX/issues/61#issuecomment-139135833.

It would be nice to see this integrated in the auto-renderer. So that we extract these fragments as well.

Thanks @kevinbarabash for picking this up. And sorry for me being this critical about the change.

kevinbarabash:
@gagern this PR was mainly to solve an issue that we're having at KA.  I think another option would be for us (KA) to use the new macro functionality to add our own custom aliases.

kevinbarabash:
After thinking about this a bit more, macros aren't enough to fix this.  I think I'll just swap out `begin{align}` and `end{align}` before sending the string to katex.

gagern:
That's the approach I'd suggest as well.

oddhack:
FWIW we've just run into this as well trying to port some MathJax content. AFAICT the best option at this point seems to be a mass replacement of \begin{align*} with \[ \begin{aligned} similarly for \end.

edemaine:
Maybe KaTeX could support this as a feature in autorender?  See also #604.

kevinbarabash:
Not sure where the best place to put this is. People who aren't using the auto-renderer may also want common substitutions... another one being non-breaking spaces to regular spaces.

gagern:
Trying to decide between core code, auto renderer and a new module of contributed code, can you provide details as to how they differ? I get the impression you don't want to be burdened by auto-renderer stuff, but in fact I don't think anything in terms of maintainance burden would change if that were part of the core library. If we could include the auto renderer, then anything in between could be included as well, simply to be enabled by the flick of some option.

kevinbarabash:
I'm more concerned about how large the bundle gets.  @gagern are you suggesting providing some way to create custom builds?

gagern:
@kevinbarabash: That was not my intention, although doing so would be an option. I'd first check how much it really changes the size.

PR #502, opened 2016-07-08
Add some more symbols

gagern:
This adds support for the following input sequences:

    -- --- ` ' `` '' \degree \pounds \maltese

resulting in – — ‘ ’ “ ” ° £ ✠ symbols already present in our fonts.

As part of this modification, the recognition of multiple dashes was moved from the lexer to the parser.  This is neccessary since in math mode a sequence of hyphens is just a sequence of minus signs.  Just like a pair of apostrophes in math mode is a double prime not a right double quotation mark.
To make this easier, `parseGroup` and `parseOptionalGroup` have been merged.


kevinbarabash:
Thanks for the PR.  I'll try to look at this sometime this week.

src/buildCommon.js
```diff
@@ -26,6 +26,7 @@ var greekCapitals = [
 var dotlessLetters = [
```
kevinbarabash: Can you change this to `italicLetters`?

src/Parser.js
```diff
@@ -707,53 +702,71 @@ Parser.prototype.parseSizeGroup = function(optional) {
...
 
 /**
- * Parses a group, which is either a single nucleus (like "x") or an expression
- * in braces (like "{x+y}")
+ * If the argument is false or absent, this parses an ordinary group,
+ * which is either a single nucleus (like "x") or an expression
+ * in braces (like "{x+y}").
+ * If the argument is true, it parses either a bracket-delimited expression
+ * (like "[x+y]") or returns null to indicate the absence of a
+ * bracket-enclosed group.
```
kevinbarabash: nice!

src/Lexer.js
```diff
@@ -78,9 +77,8 @@ Token.prototype.range = function(endToken, text) {
  * still reject the input.
  */
 var tokenRegex = new RegExp(
-    "([ \r\n\t]+)|(" +                                // whitespace
-    "---?" +                                          // special combinations
```
kevinbarabash: makes sense

dockers/Screenshotter/screenshotter.js
```diff
@@ -69,6 +69,9 @@ var opts = require("nomnom")
         flag: true,
         help: "Check whether screenshot matches current file content",
     })
+    .option("wait", {
+        help: "Wait this many seconds between page load and screenshot",
+    })
```
kevinbarabash: Can you put this in a separate PR?

src/buildCommon.js
gagern: I think I'll go for `mainitLetters`, since in the context of the `mathit` function, the emphasis is not on italic but on main instead of math.

kevinbarabash:
@gagern thanks for splitting this PR.

PR #503, opened 2016-07-08
Upgrade less to version 2

### LINKED ISSUE #500: Update to less 2
Running some of the build tools (like the server) on a recent version of node will print the following message:

> fs: re-evaluating native module sources is not supported. If you are using the graceful-fs module, please update it to a more recent version.

This is due to us using an outdated version of [less](https://www.npmjs.com/package/less). We are at 1.7.5 but latest is 2.7.1. That's a new major version, so there will be breaking changes. Luckily they are [well documented](http://lesscss.org/usage/#v2-upgrade-guide), so we should see whether any of that affects us, and then switch to using that version.

Other packages might be worth updating as well:

* browserify (10.2.6 to 13.0.1)
* clean-css (2.2.23 to 3.4.18)
* eslint (1.10.3 to 3.0.1)
* express (3.3.8 to 4.14.0)
* glob (5.0.15 to 7.0.5)
* pako (0.2.7 to 1.0.1)
* uglify-js (2.4.24 to 2.7.0)

The last one, uglify-js, is just a semver-minor upgrade, so that *should* not break anything if they do their versioning right. For the others, we'd have to identify breaking changes and see whether they affect us.

Do you want all these upgrades in one PR if possible, or would you prefer just upgrading less and leaving all the other versions alone as long as they don't cause any serious trouble?

gagern:
This fixes a problem where the graceful-js dependency of the old less version is incompatible with node 7, and will cause warnings on node 6. The resulting `katex.min.css` is *exactly* the same as with the previous version, so the major version upgrade does not seem to affect the outcome.

Fixes #500.

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
     "jasmine-core": "^2.3.4",
     "js-yaml": "^3.3.1",
     "jspngopt": "^0.1.0",
-    "less": "~1.7.5",
+    "less": "~2.7.1",
     "nomnom": "^1.8.1",
     "pako": "0.2.7",
     "selenium-webdriver": "^2.46.1",
--- a/server.js
+++ b/server.js
@@ -61,19 +61,18 @@ app.get("/katex.css", function(req, res, next) {
             return;
         }
 
-        var parser = new less.Parser({
+        less.render(data, {
             paths: ["./static"],
             filename: "katex.less",
-        });
-
-        parser.parse(data, function(err, tree) {
+        }, function(err, output) {
             if (err) {
+                console.error(String(err));
                 next(err);
                 return;
             }
 
             res.setHeader("Content-Type", "text/css");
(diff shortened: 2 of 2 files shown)


gagern:
Travis fails, and this time the failure appears to be related to this issue. Need to investigate further…

gagern:
Works now. Ready to merge if you want to.

kevinbarabash:
Cool.  Thanks for being proactive with this issue.

kevinbarabash:
Note: this requires a `npm install` to update deps.

gagern:
Do we want a Makefile rule to run `npm install` if the `package.json` is more recent than a stamp generated after `npm install`?

kevinbarabash:
That would be awesome.  In perseus we just always run `npm install` which is a little wasteful.

Issue #499, opened 2016-07-06
Travis screenshot tests get stuck


gagern:
There apparently is some chance that a Travis screenshot test will get stuck, until it gets killed by Travis after 10 minutes with no output.

```
Starting container for selenium/standalone-chrome:2.48.2
Container c624b8e09ffb started, creating screenshots...
(node:4721) fs: re-evaluating native module sources is not supported. If you are using the graceful-fs module, please update it to a more recent version.
Selenium driver at http://172.17.42.1:32769/wd/hub
KaTeX URL is http://172.17.42.1:46648/

No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.

The build has been terminated
```

I've failed to reproduce this locally using selenium-webdriver 2.48.2 or 2.53.3. I've added some debug output to `screenshotter.js` and found that on Travis it fails after the driver has been built in `buildDriver` but before the first screenshot has been taken to tune the window size in `setSize`. So it looks as though the problem lies somewhere in essentially the following [chain of commands](https://github.com/Khan/KaTeX/blob/b49eee4de70e787e557f4bfd0b787a728dcaae81/dockers/Screenshotter/screenshotter.js#L219-L232):

``` js
driver.manage().timeouts().setScriptTimeout(3000).then(function() {
    return driver.manage().window().setSize(1024, 768);
}).then(function() {
    return driver.takeScreenshot();
}).then(function(img) {
    // process screenshot
});
```

This fails to reach its final step. I can't see anything wrong with this code, so I guess a Selenium bug of some kind might be involved as well. Nevertheless, we should try to find it or work around it, in order to avoid having to re-trigger Travis tests on a regular basis.

kevinbarabash:
Re-triggering Travis is a bit of a pain.  Maybe there's some way that we could stop screenshot tests before they time out.

gagern:
I'll look into that approach.

gagern:
[My watchdog branch](https://github.com/gagern/KaTeX/commits/watchdog) has a workaround that should detect Travis getting stuck. It's built on #494, and it entails quite a number of code changes considering it's only a workaround. I've been trying to test this but so far I have been unable to reproduce the Travis problem, in a dozen runs or so. Apparently there were some screenshots taking more than one minute in one run, but not the huge delay while waiting to set the size. I wonder whether Travis somehow made the problem go away. So I'd suggest we wait a bit, see if the problem persists, and consider using that approach if it does and after #494 got merged.

[...]

PR #505, opened 2016-07-14
Call “npm install” automatically if package.json changed

gagern:
This adds a stamp file which is used to detect whether the `package.json` file got updated since the last `npm install`.  If so, `npm install` is run again to update all modules to the version described in `package.json`. This happens as a dependency of only those modules which actually need some npm-installed module.

Setting the corresponding make variable to the empty string disables the feature, which is used by the `make` invocation in the `prepublish` script inside `package.json` to avoid infinite loops.  It can also be used by developers working in an environment with reduced connectivity, as long as they know what they are doing.

This is in reaction to https://github.com/Khan/KaTeX/pull/503#issuecomment-232262910. @kevinbarabash, let me know if you'd prefer other names for the stamp file and/or the Makefile variable.

Can we place the stamp file inside the `build` directory, risking that a) people may copy that to their webservers (slightly unclean but little harm there I think) or might be doing `make clean` on a regular basis and would get annoyed if that caused a fresh `npm install` each time? If the stamp were in the `build` directory, I'd probably omit the leading dot.

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,7 +33,7 @@
   "bin": "cli.js",
   "scripts": {
     "test": "make lint test",
-    "prepublish": "make dist"
+    "prepublish": "make NIS= dist"
   },
   "dependencies": {
     "match-at": "^0.1.0"
--- a/Makefile
+++ b/Makefile
@@ -13,19 +13,24 @@ export UGLIFYJS = $(realpath ./node_modules/.bin/uglifyjs) \
 	--beautify \
 	ascii_only=true,beautify=false
 
-setup:
+# The prepublish script in package.json will override the following variable,
+# setting it to the empty string and thereby avoiding an infinite recursion
+NIS = .npm-install.stamp
+
+$(NIS) setup: package.json
 	npm install
+	@touch $(NIS)
 
-lint: katex.js server.js cli.js $(wildcard src/*.js) $(wildcard test/*.js) $(wildcard contrib/*/*.js) $(wildcard dockers/*/*.js)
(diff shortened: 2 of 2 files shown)


gagern:
The Travis complaint is another instance of #499.

gagern:
Rebased to avoid conflict, and Travis is happy as well this time.

package.json
```diff
@@ -33,7 +33,7 @@
   "bin": "cli.js",
   "scripts": {
     "test": "make lint test",
-    "prepublish": "make dist"
+    "prepublish": "make NIS= dist"
```
kevinbarabash: Why `NIS=` instead of `NIS`?

gagern: It is a variable, not a make goal. This argument changes the `NIS` variable from its default setting (a file name) to the empty string. So it essentially removes `NIS` as a dependency from all the rules mentioning it, thus avoiding an infinite recursion. Should I add a comment, or are comments a bad idea in a file that gets re-written by npm from time to time?

kevinbarabash: Cool.  Maybe add a comment in the Makefile.  `npm` won't be happy if there are comments in `package.json`.

gagern:
I added a comment to the Makefile.

(And as I augmented my commit, I noticed that I had placed a paragraph on that variable in the commit message, from where it got transferred to the initial message of this PR. But we should make an effort to make code understandable without reading commit messages or pull requests, so the comment is the right thing to do in any case, I think.)

kevinbarabash:
@gagern thanks!

Issue #358, opened 2015-09-22
Possible non-standard command "\cursor" for WYSIWYG editor


daniel3735928559:
I've been playing around with making [a WYSIWYG mathematics editor](https://github.com/daniel3735928559/guppy) using KaTeX for rendering, and have run into some issues with the cursor.  The ideal would be something like a \cursor command that inserts a vertical line of width 1px, of sensible height, and that doesn't affect the spacing of anything around it.  

Experimentally, I've tried modifying \rule (since I don't use it otherwise) by adding some lines in buildHTML.js right before `// Style the rule to the right size` to force the size: 

```
        rule.style.padding = "0px";
        rule.style.marginRight = "-1px";
        rule.style.borderLeftWidth = "1px";
```

With this modification, the actual TeX used for the cursor currently is `\rule[-0.5ex]{0em}{0.7em}`.  This works acceptably, but there seem to be a few more complex interactions/margin-settings that this doesn't account for.  

To see it working, go to the [demo](http://daniel3735928559.github.io/guppy/) and enter `1+2+3+4` and use the arrow keys to move the cursor around.  Nothing else shifts around as the cursor moves.  

To see it not working as well, enter `x+sin(x)` and move the arrow keys between the `+` and the `sin` and see that the `sin` shifts to the right.

I haven't been able to chase up what is responsible for this, and various permutations of these modifications as well as attempts to use \rlap and \llap don't do as well as the above, so I figured I'd ask to see if someone who knows more about the KaTeX side knows what's up and/or whether there's any more general interest in such a feature.

xymostech:
TeX has rules for putting horizontal space between elements, and that's what you're seeing. In this case, when you put the cursor between the `+` and `\sin`, you get the `bin + ord` interaction (because `+` is a bin and `\rule` is an ord) plus the `ord + op` interaction (because `\sin` is an op), whereas when there is no cursor you only get `bin + op`, which adds up to a different amount. 

There might be a way to get the alignment to work using some funkiness from TeX (I'm not exactly sure what, I'll think about it), but we don't have those things implemented in KaTeX right now.

As for the CSS you're manually applying:

```
rule.style.padding = "0px";
```

This shouldn't be needed.

```
rule.style.marginRight = "-1px";
```

This can probably be accomplished with a `\kern`.

```
rule.style.borderLeftWidth = "1px";
```

[...]

Issue #376, opened 2015-10-30
Create an intermediate representation that can be used to render to multiple targets


kevinbarabash:
Current efforts such as https://github.com/Khan/KaTeX/pull/251 use the output of `buildHTML` which contains some HTML specific data which is necessary for canvas rendering.  It would be nice if we had a more minimal intermediate representation that could be use for rendering HTML, MathML, Canvas, and SVG.

xymostech:
It probably makes sense to base this representation on what can come in TeX's math mode, which can be found on page 290 of the TeXbook. We can choose a useful subset of them, (like the ones that we currently support) and figure out all the parameters necessary to render them.

We also might want to think about how we should handle state going forwards. Right now, the only state modification we really allow are style changes (like `\displaystyle`) which lets us basically sidestep the problem entirely using the funky "implicit groups" but we might want to come up with a representation that more correctly models how TeX works for the future (in case we want to start allowing other kinds of state changes, like variables or macros).

I could write up a doc detailing such a representation, or we could work on this together some time?

kevinbarabash:
We had a chat and decided to do a couple of things:
- simplify ParseNode/ParseResult into POJOs w/ functions that return those POJOs (this is so that we maintain output that can be JSON serialized)
- audit existing parse node types to see where we deviate from commands being used in chapters 24 and 25.

spontaliku-softaria:
My vision of this issue as far as I currently understand it:
* KaTeX parser generates CST which is potentially less readable than AST. It contains more data than external user (author of an editor for example) absolutely needs to provide. 
Trivial example: `\\frac` node has `type`, `numer` and `denom` fields which are semantic and are needed, but also `hasBarLine`, `leftDelim`, `rightDelim` and `size` whose necessity may not be obvious to external user.
* Both buildHtml and buildMathML accept CST.
* An editor (#809) most likely would want to get rendered HTML from AST.
* There are use cases for bare parser. I guess, this cases would want AST instead of CST too.

So the task is to provide an AST and a way to get CST from AST while keeping changes minimal and backward-compatible. An explicit list of all AST nodes should be provided to simplify tree building from outside of KaTeX for external user.

Also I'd like to avoid changes in buildHtml and buildMathML if possible.

Thus I propose a plan:
* Make a set of classes representing AST nodes. Filed `type` of ParseNode and file `functions.js` are sources of information what classes are needed. Example for `op` (\sin, \max, \int, \mathop, etc.):
  ```javascript
[...]

PR #506, opened 2016-07-16
Cursor

daniel3735928559:
Going back to #358, this adds a `\cursor` command with an optional shift argument, and an argument for height (the same arguments as the `\rule` command, but without width).  I needed this for a WYSIWYG mathematics editor project [guppy](http://github.com/daniel3735928559/guppy) (see also [demo](http://daniel3735928559.github.io/guppy/)).

The original idea was to try to mess with `\kern` and `\rule` and similar to get the desired result of a vertical line that is 1 pixel thick and doesn't affect the appearance/spacing/etc. of the surrounding equation wherever it is inserted.  But part of the trouble seems to be that `\rule` gets a `.mord` class which may affect the spacing between elements if, for example, it is between a `.mbin` element and a `.mord` element.  So I ended up creating a separate command for this purpose.

Since this is not a command from standard LaTeX, I am not sure what the policy is on actually including this, but I wanted to at least document how it could be done (and/or find out if there is a better way), in case anyone else also wants this.  If there is appetite for including it, I am happy to go through the rest of the proper procedure (squashing commits, maybe adding tests, etc.) also.

If it is best not to have it in the actual repo, I would be interested in some mechanism by which this could be made into a plugin, (since as it is, I end up needing to constantly merge in updates to KaTeX to my repository, whereas if I could just pull in the latest version and have the separate plugin, this might be simplified).

### DIFF
--- a/src/buildHTML.js
+++ b/src/buildHTML.js
@@ -53,6 +53,7 @@ var groupToType = {
     overline: "mord",
     underline: "mord",
     rule: "mord",
+    cursor: "cursor",
     leftright: "minner",
     sqrt: "mord",
     accent: "mord",
--- a/src/buildHTML.js
+++ b/src/buildHTML.js
@@ -191,10 +191,13 @@ groupTypes.bin = function(group, options, prev) {
         var atoms = prevAtom.value.value;
         prevAtom = atoms[atoms.length - 1];
     }
+    if (prevAtom && prevAtom.type === "cursor") {
+        prevAtom = prevAtom.value.previous;
+    }
     // See TeXbook pg. 442-446, Rules 5 and 6, and the text before Rule 19.
     // Here, we determine whether the bin should turn into an ord. We
     // currently only apply Rule 5.
-    if (!prev || utils.contains(["mbin", "mopen", "mrel", "mop", "mpunct"],
+    if (!prevAtom || utils.contains(["mbin", "mopen", "mrel", "mop", "mpunct"],
             getTypeOfGroup(prevAtom))) {
         group.type = "textord";
         className = "mord";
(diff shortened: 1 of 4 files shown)


kevinbarabash:
@daniel3735928559 in general we try to avoid non-standard commands.  That being said, being able to use KaTeX for editing would be immensely useful.  I think having some sort of plugin system would be nice, but I think we'd want to work on defining a display agnostic layout data structure first.  This change would break plugins if they were using the existing system.  I'm going to leave this open for now as a reference how to do cursors and as a reminder that people will find a plugin system useful.

gagern:
When devising a plugin system, please keep the render-to-canvas from #251 in mind as well as the #376 intermediate representation it motivated (and which you probably meant by “display agnostic layout data structure”). Do you want to open a ticket to discuss requirements for a plugin system, @kevinbarabash?

kevinbarabash:
Blocked on #762.

kevinbarabash:
I think this could be done in a similar way to how #1074 implements `\class` in the contrib folder.

kevinbarabash:
@daniel3735928559 sorry for letting this sit in limbo for so long.  If you're still interested in adding this feature, it's probably best to wait until #1074 lands before reworking things.

daniel3735928559:
I am indeed still interested!  I haven't looked at that PR yet in depth; while a compile-time plugin system still requires me to retain a KaTeX source module (and could in principle lead to a bunch of out-of-date such modules in the wild, especially if a plugin is very widely used like `\class` might be), at least I wouldn't have to rebase these changes constantly so it's an improvement over the current state.  

If this is the way forward then I can update this PR when it is ready.

kevinbarabash:
I'm going to close this for now since #1074 has been closed.  I think we'll probably start looking at how to support a plugin architecture after the 1.0 release.

PR #507, opened 2016-07-16
Class

daniel3735928559:
In a similar spirit to #506 (and for the same reason--this time as part of the plan to add mouse support to the editor), adds an `\xmlClass` function that adds adds its first argument as a class to whatever its second argument is.  For example, `\xmlClass{foo}{x}+2` will render `x+2` like normal, but the span for the `x` will have class `.foo` in addition to all its other usual classes.  

This change ended up being a little bit invasive, and I am also interested in any methods of doing the same thing more simply.

### DIFF
--- a/src/Parser.js
+++ b/src/Parser.js
@@ -612,6 +612,10 @@ Parser.prototype.parseGroupOfType = function(innerMode, optional) {
     if (innerMode === "size") {
         return this.parseSizeGroup(optional);
     }
+    if (innerMode === "string") {
+        var val = this.parseStringGroup("string", optional);
+        return new ParseFuncOrArgument(val.text, false);
+    }
 
     this.switchMode(innerMode);
     if (innerMode === "text") {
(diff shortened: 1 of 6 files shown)


kevinbarabash:
@daniel3735928559 thanks for the pull request.  I won't be merging this b/c we'd like to avoid non standard commands in katex-core.  I would like to eventually have a plugin system.  This change highlights the need for plugins to have access to specific renderers.

kevinbarabash:
Blocked on #762.

rrandallcainc:
I know this is non-standard according to LaTeX standards, but I do feel like this could represent a core feature for KaTeX. MathJax has a similar feature, which can be found here: http://docs.mathjax.org/en/latest/tex.html#html

I think it could be difficult to implement without access to the KaTeX core.

kevinbarabash:
Superseded by #1074.

Issue #123, opened 2014-09-25
Alternative font support (AMS Euler, STIX, etc)


Hywan:
Hello :-),

I use this font for some projects: https://github.com/khaledhosny/euler-otf. This is an open-source port of AMS Euler, called NeoEuler. I wonder how much difficult would it be to use NeoEuler with KaTeX?

Thanks!

raichu:
Please also consider adding [esstix](https://www.ctan.org/pkg/esstix) [font](https://en.wikipedia.org/wiki/STIX_Fonts_project#ESSTIX) ([ttf](http://www.w3.org/Amaya/Distribution/esstix.tar.gz)).

kevinb7:
I had a look at the .ttf files and I couldn't seem to find any that contained plain old italics for latin characters or normal numbers.  Am I missing something?

spicyj:
@kevinb7 One of them should definitely have numbers, but there's no text font. In Concrete Mathematics (the book for which the font was made) they used a text font called Concrete:

http://tex.stackexchange.com/questions/103983/how-do-i-use-ams-euler#comment227008_103983

raichu:
It seems Elsevier didn't include the text font.
But it does contain Greek letters and mathematical symbols.

Anyone know what the text font in [this document](https://lists.w3.org/Archives/Public/www-amaya-dev/2000Apr/att-0041/Readme.pdf)? It looks pretty close.

octatoan:
Any updates on this?

Concrete Mathematics + Palatino is very nice. Both are Hermann Zapf fonts, and he's said they go well together.

kevinb7:
It's possible, but it would take some elbow grease.  One of the issues is that KaTeX isn't a full TeX engine, we don't support macros or (re)defining commands.  I believe that many LaTeX packages redefine certain basic TeX commands to override the default fonts that are used for math like cmr10, cmmi10, cmsy10, etc.  Also, KaTeX supports many high level commands but doesn't implement those low level commands that everything is built upon in TeX so swapping out fonts would require a different approach.

[...]

PR #508, opened 2016-07-20
Add support for Latin-1, Cyrillic, and CJK characters inside \text{}

kevinbarabash:
Summary:
This diff provides support for Latin-1, Cyrillic, and CJK characters
inside \text{} groups.  For Latin-1 and Cyrillic characters we use
glyph metrics from a glyph from Basic Latin that has roughly the same
bounding box.  We use the metrics for a capital 'M' to approximate the
full-width CJK characters.  Half-width characters are not supported yet.

Test Plan:
- make test

TODO: 
- [x] make screenshots


kevinbarabash:
screen shot 2016-07-20 at 4 53 40 pm

spicyj:
Those CJK characters don't look like Computer Modern. Are they supposed to?

kevinbarabash:
@spicyj good eye.  This diff is a stop-gap measure for Latin-1.  The plan is to eventually get https://github.com/Khan/KaTeX/pull/409 merged in and then create accented characters using combining marks that already exist in our fonts.  When the happens I use the correct font metrics for those characters.

I played around with it a bit and it seems to be doing the correct the thing for fractions and roots.  I'll post some more screenshots in a bit.

kevinbarabash:
The screenshot tests are failing b/c I didn't include a screenshot in this diff b/c my docker setup is broken for the moment.  I'm going to try out @gagern's `docker-machine` branch.

kevinbarabash:
@spicyj sorry you were asking about the CJK characters.  Those are also not CM.  I'm hesitant to include a specific font especially for CJK because it will take a while to download.

spicyj:
Yeah. We could have a separate font file for CJK and browsers should only download it if it's used, but I'm guessing there isn't even CM type for CJK chars? The Latin-1 chars do look like CM to me – am I crazy?

kevinbarabash:
> The Latin-1 chars do look like CM to me – am I crazy?

Times New Roman is pretty close to CM.

spicyj:
No, this is what those chars look like in Times New Roman:

image

gagern:
This is a nice feature, and pretty important in Germany (due to ÄÖÜäöüß). Personally I'd say that even imperfect rendering should be preferable to no rendering in these cases. So a stopgap like this is great!

In case this approach will be around a while longer (and I assume that to be the case for CJK at least), perhaps we should empower site authors by providing config callbacks to determine CSS classes and character metrics for given glyphs. But that's probably best discussed in a follow-up ticket once the core here is in. Or as part of #123 and its duplicates.

Regarding this whole substitute letter magic I'd suggest keeping two things in mind. Firtly, you are substituting the metrics of a *similar* letter but in a completely *different* and unknown font. So don't get your expectations too high about the accuracy of that. Secondly, even though #366 added widths, KaTeX isn't actually using them, is it? So the key point to concentrate on for substitute characters is vertical metrics. Which is tricky, since descenders in particular vary greatly between fonts and languages. In many fonts ß will have a considerable descender, but you can't know that if you don't know the font. Ignoring accents will lead to incorrect vertical metrics as well. So my main point is: don't put too much effort (and code size) into this since it's going to be wrong in many cases despite your efforts.

FYI: I just had a look at [CSS3 Writing Modes](https://www.w3.org/TR/css-writing-modes-3/#inline-alignment) and verified that the dominant baseline of mixed-script inline text is defined to be alphabetic (for horizontal writing), deferring configurability to “a future CSS module”. So if the font doesn't fit its metrics, we'd most likely see problems at the upper boundary of the box.

static/katex.less
```diff
@@ -15,7 +15,7 @@
 }
 
 .katex {
-    font: normal 1.21em KaTeX_Main;
+    font: normal 1.21em KaTeX_Main, Times New Roman;
```
gagern: We should definitely have `serif` as a last item in this list, in case a user doesn't have Times New Roman installed. I'm also thinking whether we should prefer [DejaVu](http://dejavu-fonts.org/wiki/Main_Page) Serif over Times New Roman (i.e. list both but DejaVu first). I think that DejaVu might contain a wider range of glyphs, and that it would be easier for site admins to legally ship a webfont version of it to ensure consistent rendering. To my knowledge Times New Roman is still proprietary, plus the version on Mac is different from the one on Windows.

test/test.html
```diff
@@ -6,6 +6,8 @@
         <script src="jasmine/boot.js"></script>
         <link rel="stylesheet" href="jasmine/jasmine.css">
         <script src="katex-spec.js"></script>
+        <script src="error-spec.js"></script>
+        <script src="unicode-spec.js"></script>
```
gagern: The way I see it, these files [should get concatenated by `server.js`](https://github.com/Khan/KaTeX/blob/befe1c1af7a2ce88a1c7099ee260f05d834d76c6/server.js#L52) since #389, so this should not be needed. Or did you actually encounter a situation where it *was* needed? If not, should we aim for different names for the `katex-spec.js` as contained in the repository and the combined one served by the server? Which one to rename? I think `katex-spec.js` makes a lot of sense for the combined version, so `main-spec.js` or similar for the single file might be better.

src/Parser.js
```diff
@@ -757,6 +757,9 @@ Parser.prototype.parseOptionalGroup = function() {
     }
 };
 
+var cjkRegex =
+    /[\u3040-\u309F]|[\u30A0-\u30FF]|[\u4E00-\u9FAF]|[\uAC00-\uD7AF]/;
+
```
gagern: Where did those ranges come from? Adding that information as a comment might be useful if we have to extend the ranges at some point.

src/fontMetrics.js
```diff
@@ -121,6 +121,144 @@ var metrics = {
...
+    'э': 'e',
+    'ю': 'm',
+    'я': 'r',
+};
+
+var KoreanSyllablesRegex = /[\uAC00-\uD7AF]/;
+var HiraganaRegex = /[\u3040-\u309F]/;
+var KatakanaRegex = /[\u30A0-\u30FF]/;
+var cjkRegex = /[\u4E00-\u9FAF]/;
+var regex = /[\u3040-\u309F]|[\u30A0-\u30FF]|[\u4E00-\u9FAF]|[\uAC00-\uD7AF]/;
```
gagern: It appears that only the last of these five regular expressions is actually being used. The others are useful documentation, but perhaps better written as comments. It is particularly confusing that the `cjkRegex` in Parser corresponds to `regex` here.  Perhaps that RegExp should be shared between modules somehow?

gagern:
Look at the `ï` in the first row of your screenshot! It looks as though an i with dot was combined with a combining diaresis. So I'd guess that these glyphs are indeed CM, and your browser's text rendering engine is doing its best to generate the glyphs in question from the requested font before falling back to another font.

kevinbarabash:
screen shot 2016-07-21 at 9 46 46 am

screen shot 2016-07-21 at 9 48 01 am

src/fontMetrics.js
kevinbarabash: I'll fix this up.

kevinbarabash:
@gagern really good points about the inexact science of providing substitute metrics for characters.  The ascender is the important thing.  Unfortunately, characters with accents will have the wrong ascender for now so people should avoid doing `\vec{\text{À}}` for now.

kevinbarabash:
> No, this is what those chars look like in Times New Roman:

I really wasn't paying attention to the `i`.  It has

kevinbarabash:
The KaTeX fonts have the dotless i in the right position.  I wonder why the browser isn't using it.

kevinbarabash:
The screenshot tests made me sad b/c there are not CJK fonts install on the selenium docker images which means we'll have to make some custom images.  Hopefully we can derive them from the images we're currently using.

gagern:
Instead of modifying the selenium images, I'd suggest adding some suitable fonts as webfonts, adding them to the css of the test html, and serving them as static content.

kevinbarabash:
@gagern great idea.  I'll do that as part of this PR.

In the meantime, I've update the code to use serif fonts for CJK glyphs and `i` with accents don't have the extra dot anymore.

screen shot 2016-07-22 at 11 40 19 am

kevinbarabash:
I just noticed I have Korean twice.

src/domTree.js
```diff
@@ -183,6 +183,12 @@ function symbolNode(value, height, depth, italic, skew, classes, style) {
     this.classes = classes || [];
     this.style = style || {};
     this.maxFontSize = 0;
+
+    if (cjkRegex.test(value)) {
+        // Force serif fonts for CJK glyphs because some browsers/OSs default
+        // to a sans-serif font for these glyphs.
+        this.style['font-family'] = 'serif';
+    }
```
gagern: Can you do this in a css class, instead of using the style attribute? A class can be overridden by the site admin by adding some more css, but a style can not (except perhaps by abusing the important flag). Or perhaps multiple classes, one for each script?

kevinbarabash: A class is probably better.

kevinbarabash:
Although I added the "fallback" class and was able to get that working with different fonts in the test page, I wasn't able to get those other fonts showing when doing the screenshot tests.  I'm going to leave that as a task for a future PR.

gagern:
It appears DejaVu doesn't have Korean, contrary to what I'd have expected. Maybe we should choose [something different](https://en.wikipedia.org/wiki/List_of_CJK_fonts), perhaps something [available on TeX as well](http://tex.stackexchange.com/q/53599/16923).

xymostech:
This looks fine to me, once the screenshot tests look good :)

kevinbarabash:
The screenshots work now.  It took a while finding fonts that actually worked inside of the docker container.

gagern:
What's the license of these font's you're putting into the repo? For the [Honoka Mincho](http://font.gloomy.jp/honoka-mincho-dl.html) font I'd guess [these terms](http://ipafont.ipa.go.jp/ipa_font_license_v1.html#LicenceEN) apply, which as far as I understand it requires redistribution of the license term along with the font. The [Baekmuk Batang](http://kldp.net/projects/baekmuk/) tarball contains a [`LICENSE` file](https://fedoraproject.org/wiki/Licensing:Baekmuk?rd=Licensing/Baekmuk) which also requires distribution of said notice along with the font.

gagern:
Personally I'd prefer not to put 18M of font data into the repository just for this single test, and instead have that downloaded by some script, or perhaps packaged as an npm dependency. This is not my project, so others will have to decide on this, but I'd like to make you aware of the considerable increase in repo size as a consequence of this PR as it stands now.

kevinbarabash:
@gagern really good point about the license and the size of these fonts.

kevinbarabash:
I put the fonts in a separate repo along with LICENSE files for each of them.

.travis.yml
```diff
@@ -10,4 +10,5 @@ before_script:
   - docker images --no-trunc
 script:
   - npm test
+  - git clone https://github.com/Khan/KaTeX-test-fonts test/screenshotter/unicode-fonts
```
gagern: Do you want to clone a specific commit of that repo, so that you can update the fonts without breaking existing screenshots, and then can update the commit id and the screenshots in one go?

kevinbarabash: I didn't think the fonts would be changing, but better safe than sorry.

kevinbarabash:
It looks like everyone's concerns have been address so I'm going to merge this now.

i-am-the-slime:
How does this work? Do I need to configure something special?
I just get this:

```
ParseError: KaTeX parse error: Expected '}', got 'ß' at position 9: \text{Maß̲einheit}```. Same for `ä` etc.

kevinbarabash:
@i-am-the-slime this is on `master`.  We still need to do a release.

PR #510, opened 2016-07-23
Add --wait option to screenshotter

gagern:
Sometimes a page might not be ready when the screenshot is taken, due to fonts still being loaded or some such.  In `--verify` mode this is taken care of by repeated attempts.  But when creating a screenshot for the first time, it might be desirable to wait a given number of seconds for things to settle down.  So this commit introduces a `--wait` option to the screenshotter.

This is a spin-off from https://github.com/Khan/KaTeX/pull/502#discussion_r71964858.

### DIFF
--- a/dockers/Screenshotter/screenshotter.js
+++ b/dockers/Screenshotter/screenshotter.js
@@ -69,6 +69,9 @@ var opts = require("nomnom")
         flag: true,
         help: "Check whether screenshot matches current file content",
     })
+    .option("wait", {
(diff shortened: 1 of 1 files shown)


kevinbarabash:
This is much appreciated.  When I run the screenshot I end up reverting changes to the screenshots where I didn't modify/create the test case in ss_data.yaml.  It's okay b/c `verify` runs on travis, but it's not idea.

PR #522, opened 2016-08-27
Provide link to auto-render.min.js in usage example

ViktorQvarfordt:
A full url to the built (browserified and minified) `auto-render.min.js` should be available for the same reason that the full url to `katex.min.js.` is available in the [main readme](https://github.com/Khan/KaTeX#usage).


kevinbarabash:
@ViktorQvarfordt thanks for pull request.

Issue #539, opened 2016-09-21
Invalid input type gives cryptic error


shuttle1987:
Recently I ran into a situation where the wrong type was passed to `renderToString`:

`katex.renderToString({bogusInputType: "yes"})`

Somewhat as expected this gives a `TypeError`, but it is due to something unrelated:

`TypeError: input.slice is not a function`

Tracking this down I found that this comes from within `ParseError`,  but I'm concerned that it getting this deep without being handled earlier. The error handling here is purely coincidental and will be sidestepped entirely if the input is sliceable such as:

`var result = katex.renderToString(["OK input", {bogusInputType: "yes"}], {throwOnError: false})`

Using the settings flag to not throw on ParseError still raises the same exception from within ParseError. After looking carefully through the docs/code I came to understand that "throwOnError" is only for LaTeX parsing errors, not more general error handling, so this is to be expected. However it is a bit confusing at first when using KaTeX and adding some example code that clearly shows this would help new users.

Here's this in action: https://runkit.com/shuttle1987/57e13d446eefb414003d303d

I'm more than happy to make a PR to improve this. I propose that non-string input throws immediately and with a message that makes it really clear why the input is invalid.

gagern:
I'd welcome a PR to this effect.

gagern:
Fixed per #540.

PR #540, opened 2016-09-23
Invalid input error message

shuttle1987:
Fix for issue referenced in #539

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,17 +55,6 @@ For example:
 katex.render("c = \\pm\\sqrt{a^2 + b^2}", element, { displayMode: true });
 ```
 
-Note that the `throwOnError` option only deals with how TeX related parsing errors are handled.
-Other errors such as passing in an incorrect type for an expression will still throw a TypeError even if the `throwOnError` option is `false`.
-
-For example:
-
-```js
-katex.render("\\unsupportedCommand", element, { throwOnError: true }); // throws ParseError
-katex.render("\\unsupportedCommand", element, { throwOnError: false }); // does not throw ParseError
-katex.render({invalidInputDataType: "yes this is!"}, element, { throwOnError: false }); // throws TypeError
-```
-
 #### Automatic rendering of math on a page
 
 Math on the page can be automatically rendered using the auto-render extension. See [the Auto-render README](contrib/auto-render/README.md) for more information.
--- a/README.md
+++ b/README.md
@@ -55,6 +55,17 @@ For example:
 katex.render("c = \\pm\\sqrt{a^2 + b^2}", element, { displayMode: true });
 ```
 
+Note that the `throwOnError` option only deals with how TeX related parsing errors are handled.
+Other errors such as passing in an incorrect type for an expression will still throw a TypeError even if the `throwOnError` option is `false`.
+
+For example:
+
+```js
+katex.render("\unsupportedCommand", element, { throwOnError: true }); // throws ParseError
+katex.render("\unsupportedCommand", element, { throwOnError: false }); // does not throw ParseError
+katex.render({invalidInputDataType: "yes this is!"}, element, { throwOnError: false }); // throws TypeError
+```
+
 #### Automatic rendering of math on a page
 
 Math on the page can be automatically rendered using the auto-render extension. See [the Auto-render README](contrib/auto-render/README.md) for more information.
--- a/README.md
+++ b/README.md
@@ -61,8 +61,8 @@ Other errors such as passing in an incorrect type for an expression will still t
 For example:
 
 ```js
-katex.render("\unsupportedCommand", element, { throwOnError: true }); // throws ParseError
-katex.render("\unsupportedCommand", element, { throwOnError: false }); // does not throw ParseError
+katex.render("\\unsupportedCommand", element, { throwOnError: true }); // throws ParseError
+katex.render("\\unsupportedCommand", element, { throwOnError: false }); // does not throw ParseError
 katex.render({invalidInputDataType: "yes this is!"}, element, { throwOnError: false }); // throws TypeError
 ```
 
--- a/katex.js
+++ b/katex.js
@@ -19,6 +19,9 @@ var utils = require("./src/utils");
  * given.
  */
 var render = function(expression, baseNode, options) {
+    if (!(typeof expression === 'string' || expression instanceof String)) {
+        throw new TypeError('render requires string typed expression');
+    }
     utils.clearNode(baseNode);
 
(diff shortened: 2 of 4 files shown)


README.md
```diff
@@ -55,6 +55,17 @@ For example:
 katex.render("c = \\pm\\sqrt{a^2 + b^2}", element, { displayMode: true });
 ```
 
+Note that the `throwOnError` option only deals with how TeX related parsing errors are handled.
```
gagern: That's less precise than it could be: the option only deals with unsupported commands, i.e. command names we know nothing about. Other invalid input, like e.g. mismatched braces, will cause exceptions as well no matter this setting. Actually the option description from the list of options above this part is correct in this respect.

README.md
```diff
@@ -55,6 +55,17 @@ For example:
...
+Other errors such as passing in an incorrect type for an expression will still throw a TypeError even if the `throwOnError` option is `false`.
+
+For example:
+
+```js
+katex.render("\\unsupportedCommand", element, { throwOnError: true }); // throws ParseError
+katex.render("\\unsupportedCommand", element, { throwOnError: false }); // does not throw ParseError
+katex.render({invalidInputDataType: "yes this is!"}, element, { throwOnError: false }); // throws TypeError
+```
+
```
gagern: Personally I don't think we need to add three paragraphs here just to discuss that one option. If you think the current wording is easily misread, perhaps add some text like “Other malformed inputs will always throw errors, though” to the description of the option within the bullet list. If you'd like to stress the kind of thing you're fixing here, making it “appropriate errors” in that sentence might express that, but I guess there will be other situations we haven't considered, so I'd not add that. Having useful error messages is useful to the people who encounter them, but in my opinion not worth documenting in the README, since that file should be fairly concise. The example is in my opinion an example of how to use options in general, and we should not have one example for every option.

katex.js
```diff
@@ -47,6 +50,9 @@ if (typeof document !== "undefined") {
  * Parse and build an expression, and return the markup for that.
  */
 var renderToString = function(expression, options) {
+    if (!(typeof expression === 'string' || expression instanceof String)) {
+        throw new TypeError('renderToString requires string typed expression');
+    }
     var settings = new Settings(options);
 
     var tree = parseTree(expression, settings);
```
gagern: Putting both these checks in `parseTree` would avoid duplication. That way you couldn't include an API-level function name, but usually browsers will indicate `katex.js` as the source of this exception, and for those who need more detail the stack traces will likely be more useful anyway.

test/katex-spec.js
```diff
@@ -1382,6 +1382,15 @@ describe("An HTML font tree-builder", function() {
...
     });
+
+    it("should throw TypeError when the expression is of the wrong type", function() {
+        expect(function() {
+            katex.renderToString({badInputType: "yes"});
+        }).toThrowError(TypeError);
+        expect(function() {
+            katex.renderToString(["foo", {badInputType: "yes"}]);
+        }).toThrowError(TypeError);
+    });
```
gagern: This only tests one of the two codepaths above, which will be OK if you move the checks to a single location.

What's the point of the second test? If it is the array, I'd avoid the nested object. But I don't see much need for a second object-based test. `null`, `undefined`, number or boolean would make for more interesting tests, but perhaps not interesting enough to actually do them.

Since your check you added assumes that `String` objects are valid input as well, we might add a test to verify that a `String` object as input does parse all right.

README.md
shuttle1987: I agree that the README is not the best place for this and that putting it in with the other documentation would be a lot better. Where are the docs for this project?

katex.js
shuttle1987: Implemented in https://github.com/Khan/KaTeX/pull/540/commits/73b9fa5cc69fa586198ac6f84f6772a1d237036c

README.md
shuttle1987: Would it be better to say something such as:  "Note that the `throwOnError` option only deals with how missing TeX commands are handled by the KaTeX parser. Other invalid input such as mismatched braces will always throw exceptions regardless of the `throwOnError` setting."

test/katex-spec.js
shuttle1987: Suggestion of testing for more diverse types done in https://github.com/Khan/KaTeX/pull/540/commits/95acda3824e6587160554e33ea0674aaaf63f017

gagern: I'd say allow `String` objects but don't advertise that fact in any documentation, because it will be hardly relevant to most people. The way I see it, the distinction only matters up to the tokenization step. After that it's native strings either way. So if we have one single test case using `String` objects, I'd be confident that we actually support them just like we do native strings.

README.md
gagern: Depends on whether we want to keep this paragraph at all. If so, then yes, this would be better. “missing” is still inferior to “unsupported”, because it could also mean commands missing from the input like e.g. a missing `\right` or `\end`. If we drop this paragraph, then I'd not include the “such as mismatched braces” portion in order to keep the bullet list concise.

test/katex-spec.js
shuttle1987: I added a test case simply for not throwing on correct types being supplied in https://github.com/Khan/KaTeX/pull/540/commits/3dc527683003d918f487e09f4d7dc810e0e5a894 I also think it's good to not advertise using `String` objects. I could see reusing some of the tests from `describe("An HTML font tree-builder", ` with `String` objects if the coverage tests are desirable. I can add those test cases if that is desired.

gagern: Just put a single test to [the parser test group](https://github.com/Khan/KaTeX/blob/a16ae7a5ebc97cfb610a2143254393918fd0650d/test/katex-spec.js#L199-L209). Something simple like `expect(new String("xy")).toParseLike("xy")`.

README.md
gagern: Some bits of information are in the Wiki. Other than that, I guess there hasn't been any need for more documentation so far.

test/katex-spec.js
shuttle1987: Tests implemented in https://github.com/Khan/KaTeX/pull/540/commits/cd584e0a1531f087c615577d018ededef061ed7d

README.md
shuttle1987: After the discussion here I see I don't have the required understanding to make this change in the docs so I've removed it from this PR in this commit: https://github.com/Khan/KaTeX/pull/540/commits/42e21c2232e55dc360578244feee280a00510ae4

gagern:
Thank you for your contribution!

Issue #408, opened 2015-12-02
Superscript in the denominator of a fraction partially covered by line


MKevin3:
![ios katex](https://cloud.githubusercontent.com/assets/2380650/11531637/e973c4cc-98c3-11e5-8f79-c4c2a479c0a0.png)

This occurs on both iOS and Android. Here is what it looks like in MathJax for comparison

![mathjax](https://cloud.githubusercontent.com/assets/2380650/11531650/fb549914-98c3-11e5-8bd1-8eac1d5b6a45.png)

It is not a deal breaker but it would be nice to have more vertical spacing between the line and the denominator.

kevinbarabash:
What font size are you using?  Is this with `\displaystyle`?

MKevin3:
We get the data from the server - I hope this is a pared down version that shows the issue for you

display style is inline

<!DOCTYPE html>
<html>
<head>
<meta name="viewport" content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0">
<link rel="stylesheet" href="http://cdnjs.cloudflare.com/ajax/libs/KaTeX/0.5.1/katex.min.css">
<script src="http://cdnjs.cloudflare.com/ajax/libs/KaTeX/0.5.1/katex.min.js"></script>
<script type="text/javascript">
  document.addEventListener("DOMContentLoaded", function(event) {
    var imgs=document.querySelectorAll('img[src*="formula_image"]');
    for(var i=0;i<imgs.length;i++){
      var img = imgs[i];
      var formula = img.getAttribute("title");
      if(!formula || (formula && formula.trim() == "")) {
        formula=img.getAttribute("alt");
      }
      if(formula && formula.trim() != "") {
        try {
          var span=document.createElement("span");
          katex.render(formula,span);
          span.setAttribute("class","formula_image");
          img.parentNode.insertBefore(span,img.nextSibling);
          span.style.display="block";
          imgs[i].style.display="none";
        }
        catch(err) {}
      }
    }
  });
</script>
</head>
<body>
<div class="main">
​
​
<div class="box question">
<p>
<img title="\frac{2x^2-5x-12}{x^2-11x+28}" src="http://vt-s3-files.s3.amazonaws.com/uploads/formula_image/image/252552/gif.latex" alt="" style="display: none;">
</p>
</div>
</div>
​
</body></html>

sschendel:
[...]

Issue #542, opened 2016-10-06
poor spacing in fractions


jason-s:
`$\alpha _ e = \frac{3K _ e I _ \text{max}}{2 J}$`

KaTeX: 

![image](https://cloud.githubusercontent.com/assets/738893/19167677/a01a7810-8bc1-11e6-8f10-0b56356219cc.png)

MathJax:

![image](https://cloud.githubusercontent.com/assets/738893/19167685/acc12a0a-8bc1-11e6-93df-a72f198a1e3e.png)

sthoch:
Could be a font-size related issue like #41. At 4-5pt I get a similiar rendering to yours, above that, sometime the subscript touch the fraction bar. How small are you rendering? In what context are you using KaTeX?

gagern:
Yes, this looks very much like #41. Again the vertical spacing appears correct at larger sizes, with “correct” meaning “as TeX does it”:

![Comparison](https://cloud.githubusercontent.com/assets/190759/19184781/fa6ffa84-8c7e-11e6-831c-23f758b2ffce.png)

TeX in green, KaTeX in red. You see that the vertical spacing matches even though the horizontal one does not.

I'm closing this as a duplicate of #41.

Issue #41, opened 2014-09-15
Exponents sometimes collide with fraction bars


jwmerrill:
![image](https://cloud.githubusercontent.com/assets/48409/4277932/63382afe-3d13-11e4-855e-60b30e297e10.png)

Output on right is from latex for comparison. In KaTeX, the `-6 \pi` and `-8 \pi` exponents collide with the fraction bars above them. I'm on OS 10.9.4 and I see this issue in both Chrome and Firefox.

gagern:
Running [our `texcmp` tool](https://github.com/Khan/KaTeX/tree/v0.6.0/dockers/texcmp) against this test case confirms that at large sizes the vertical spacing of KaTeX agrees with what TeX does:

![Comparison with TeX](https://cloud.githubusercontent.com/assets/190759/19184113/de9cb64c-8c7b-11e6-9cbb-d1e6ce429e13.png)

TeX rendering in green, KaTeX in red. (Horizontal spacing is decidedly different, probably because KaTeX uses the same font at all sizes while TeX will use a different font with wider glyphs for smaller styles. Although bugs like #249 may contribute to this as well.)

So if we want TeX compatibility then we have to accept the incorrect rendering by some browsers at some font sizes. If we want to ensure some minimal spacing, then we can certainly do so. I'd say somewhere in [this clearance computation](https://github.com/Khan/KaTeX/blob/v0.6.0/src/buildHTML.js#L415-L432). And if we want to give users a choice, we can introduce a config option for this. Perhaps even one where people can choose the amount of additional separation they would like. Such a setting should probably apply to matrices as well.

> Maybe we can use some fancy CSS to ensure minimum widths of the gaps between elements or something.

This would in my opinion be fancy only if it were to introduce a font-size specific element, i.e. something like `0.5em + 3px` or similar. Doing this kind of thing would break a million places and some very useful applications, so I'd rather avoid it. Anything else would be simply tweaking some shifts in JavaScript code, with no changes to the KaTeX stylesheet.

gagern:
> confirms that at large sizes the vertical spacing of KaTeX agrees with what TeX does

[...]

PR #545, opened 2016-10-08
Use style specific FONTDIM metrics

kevinbarabash:
FONTDIM metrics include metrics like sup2, sup3, etc. which are used for
position sub/super-scripts, fractions, delimiters, etc.  TeX uses three
different font styles: textfont2 (DISPLAY & TEXT), scriptfont2 (SCRIPT), and
scriptscriptfont2 (SCRIPTSCRIPT) and has different sets of metrics for each.
This diff adds style specific metrics for better TeX compliance.

Addresses https://github.com/Khan/KaTeX/issues/408.

Test Plan:
- [x] TODO: update screenshot tests


src/fontMetrics.js
```diff
@@ -18,30 +18,28 @@ var cjkRegex = require("./unicodeRegexes").cjkRegex;
...
-var sigma16 = 0.150;
-var sigma17 = 0.247;
-var sigma18 = 0.386;
-var sigma19 = 0.050;
-var sigma20 = 2.390;
-var sigma21 = 1.01;
-var sigma21Script = 0.81;
-var sigma21ScriptScript = 0.71;
-var sigma22 = 0.250;
+var sigma1 = 0.025;                  // slant
```
kevinbarabash: `tftopl` reported this as 0.25.  It's not actually used anywhere.  I was thinking I should update this.

src/fontMetrics.js
```diff
@@ -18,30 +18,28 @@ var cjkRegex = require("./unicodeRegexes").cjkRegex;
...
+var sigma12 = [0.345, 0.344, 0.532]; // denom2
+var sigma13 = [0.413, 0.503, 0.504]; // sup1
+var sigma14 = [0.363, 0.431, 0.404]; // sup2
+var sigma15 = [0.289, 0.286, 0.294]; // sup3
+var sigma16 = [0.150, 0.143, 0.200]; // sub1
+var sigma17 = [0.247, 0.286, 0.400]; // sub2
+var sigma18 = [0.386, 0.353, 0.494]; // supdrop
+var sigma19 = [0.050, 0.071, 0.100]; // subdrop
+var sigma20 = [2.390, 1.700, 1.980]; // delim1
+var sigma21 = [1.010, 1.157, 1.420]; // delim2
```
kevinbarabash: The values for `sigma21` that were here before had premultiplied the style's `fontSizeMultiplier`.  Since our inline styles use `em`s to position thing, I would assume that no scaling would be required because `em`s are relative to the font-size of the element.

src/fontMetrics.js
```diff
@@ -72,48 +70,55 @@ var ptPerEm = 10.0;
...
+        }
+        throw new Error("Unexpected style size: " + style.size);
+    };
+}
+
 /**
  * This is just a mapping from common names to real metrics
  */
 var metrics = {
     xHeight: sigma5,
```
kevinbarabash: All of the values for `sigma5` are the same.  Perhaps I should change this to an array and change this to `getXHeight` for consistency's sake.

src/environments.js
```diff
@@ -161,7 +162,8 @@ defineEnvironment("cases", {
             type: "align",
             align: "l",
             pregap: 0,
-            postgap: fontMetrics.metrics.quad,
+            // TODO(kevinb) get the current style
+            postgap: fontMetrics.metrics.getQuad(Style.TEXT),
```
kevinbarabash: The parser doesn't seem to have any idea of what the current style is, but maybe it should.

kevinbarabash:
@gagern if you have time, could you check out this branch and run the screenshotter?

gagern:
I had some problems recreating the Unicode screenshots locally. I'll have to investigate that further one day. Anyway, screenshots are here, Travis will look at them. And we can inspect them here in GitHub.

kevinbarabash:
@gagern thanks for adding the screenshots!  I was hoping that these changes would've improved the `\sqrt` rendering more.

src/environments.js
kevinbarabash: Sounds a little dicey.  I'll add a more detail comment in here, but I think we can leave this issue for a future diff.

src/fontMetrics.js
```diff
@@ -18,30 +18,28 @@ var cjkRegex = require("./unicodeRegexes").cjkRegex;
 // dimensions, one for each of textstyle, scriptstyle, and scriptscriptstyle,
 // but we only use the textstyle ones, and scale certain dimensions accordingly.
```
csilvers: Isn't this comment obsolete now?

gagern:
I can tweak my differencing code to do three-way comparisons between TeX and KaTeX before and after this change. Will take a bit of time, though…

src/fontMetrics.js
kevinbarabash: It is.  Thanks for catching this.  I'll update the comment.

kevinbarabash:
@csilvers I flipped between the new and old rendering of the example in https://github.com/Khan/KaTeX/issues/41 and there were two main differences that I could see:
- the `-2\pi` exponent on the top of the continued fraction is slightly higher
- the gap between the fraction bars and the rest of the exponents is slightly bigger 1~2px at a font size of 72px

@gagern even re-running the Tex vs. KaTeX comparison from https://github.com/Khan/KaTeX/issues/408 would be helpful in seeing the difference.

gagern:
I wonder whether it would be better to have these as properties on the `Style` objects, instead of methods in the `fontMetrics` module. So we could write `options.style.sub1` or `options.style.metrics.sub1` instead of `fontMetrics.metrics.getSub1(options.style)`. No strong opinion, just a thought.

gagern:
Here are some three-way diffs, which may be a bit hard to read. Red channel is TeX, green channel is this branch, blue channel is master. Since text is black, that means cyan text is TeX, magenta text is this branch, yellow text is master. If the colors overlap, then red and cyan regions are not affected by this change. Blue and yellow areas are good (improved match), while green and magenta areas would be bad (reduced match). I aligned TeX vs. this branch, but then put master at the same position as this branch here.

* :smiley:  
issue408  
The large digits and the fraction bar are blue and yellow, so the overall vertical alignment (starting from the common alignment of the top exponent) is better now. The cyan TeX 3 is also matched more closely by the magenta improved one than by the yellow master one.

* :neutral_face:  
issue542  
No changes here for #542. Presumably because we only see subscripts here.

* :smiley:  
leftright  
definitely better

* :smiley:  
binomtest  
slight improvement

* :smiley:  
exponents  
considerable improvement

I haven't seen any degraded performance in these comparisons. So I'd say this is definitely a step in the right direction. I'll wait on your opinion on where to put this information before merging.

kevinbarabash:
@gagern these are really cool screenshots.  I'm glad to see we're moving in the right direction.  Thanks for making them and thanks for explaining how to interpret them.

`fontMetrics.metrics.getSub1(options.style)` is quite verbose.  I'll update to use `options.style.metrics.sub1`.

kevinbarabash:
@gagern could you have another look at this?

src/fontMetrics.js
```diff
@@ -10,38 +10,53 @@ var cjkRegex = require("./unicodeRegexes").cjkRegex;
...
-var sigma19 = 0.050;
-var sigma20 = 2.390;
-var sigma21 = 1.01;
-var sigma21Script = 0.81;
-var sigma21ScriptScript = 0.71;
-var sigma22 = 0.250;
+// In TeX, there are actually three sets of dimensions, one for each of
+// textstyle, scriptstyle, and scriptscriptstyle.  These are provided in the
+// the arrays below, in that order.  Metrics with only a single value are
+// assumed to be the same for all three font styles.
```
gagern: That sentence no longer applies.

src/buildHTML.js
```diff
@@ -590,7 +594,8 @@ groupTypes.array = function(group, options, prev) {
         body[r] = outrow;
     }
 
-    var offset = totalHeight / 2 + fontMetrics.metrics.axisHeight;
+    var offset =
+        totalHeight / 2 + style.metrics.axisHeight;
```
gagern: The line break here appears unmotivated, as we are well below 80 characters.

src/buildHTML.js
```diff
@@ -524,6 +526,8 @@ groupTypes.array = function(group, options, prev) {
     var nc = 0;
     var body = new Array(nr);
 
+    var style = options.style;
```
gagern: This is called `style` here, `fstyle` in `genfrac` and `options.style` in supsub. Perhaps we should use the same name in all three of these functions.

src/buildHTML.js
```diff
@@ -347,16 +347,18 @@ groupTypes.supsub = function(group, options, prev) {
     } else {
         supShift = Math.max(
             supShift, minSupShift,
-            sup.depth + 0.25 * fontMetrics.metrics.xHeight);
-        subShift = Math.max(subShift, fontMetrics.metrics.sub2);
+            sup.depth + 0.25 * options.style.metrics.xHeight);
+        subShift = Math.max(
+            subShift,
+            options.style.metrics.sub2);
```
gagern: Why the line breaks? And why two distinct calls to `Math.max`, instead of including all of these into a single call?

src/Style.js
```diff
@@ -6,6 +6,18 @@
  * information about them.
  */
 
+var sigmas = require("./fontMetrics.js").sigmas;
+
+var metrics = [];
+
+for (var i = 0; i < 3; i++) {
+    metrics[i] = Object.keys(sigmas).reduce(function(res, key) {
```
gagern: According to MDN, [`Array.prototype.reduce`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/reduce) and [`Object.keys`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/keys) require IE 9, while KaTeX officially supports IE 8. So this should probably be reformulated using something like

```js
var metrics = [{}, {}, {}];
for (var key in sigmas)
    for (var i = 0; i < 3; i++)
        metrics[i][key] = sigmas[key][i];
for (var i = 0; i < 3; i++)
    metrics[i].emPerEx = sigmas.xHeight[i] / sigmas.quad[i];
```

I wonder whether we should add some tool to our build pipeline to either catch such issues or polyfill required functionality. Or do we already have a polyfill, and I wasn't aware of it? `package.json` doesn't list any obvious candidates.

.gitignore
```diff
@@ -1,4 +1,6 @@
+.npm-install.stamp
 build
+dist
```
gagern: These changes are functionally unrelated and should be handled by #546 I think.

src/Style.js
kevinbarabash: Eek.  I've been spoiled developing for IE10+.  Not having `Object.keys` is really sad.  It would be nice to have a polyfill.  In order to keep KaTeX small I'd probably want to have a legacy and a modern build.

gagern:
Looks fine to me now, thanks a lot! I guess I should squash-merge this, right?

kevinbarabash:
@gagern go for it.  Thanks for the review.

Issue #544, opened 2016-10-08
Add dist and .npm-install.stamp to .gitignore


kevinbarabash:
I was working on a PR and found myself having to delete `dist` and `.npm-install.stamp` whenever I wanted to push a change.

gagern:
We probably should add those two items to `.gitignore` so you won't have to delete them.

The purpose of #505 was to run `npm install` upon changed `package.json`, which in turn will build `dist`. To avoid that, we would need to find a way to update dependencies without running the `make dist` step in the `prepublish` script. Possible solutions:

1. Call `KaTeX_noinstall=1 npm install` from the `Makefile` (i.e. with an environment variable), and change the `prepublish` script to `test -z "${KaTeX_noinstall}" && make NIS= dist`. This will work on common *nix shells, but likely fail on Windows.

2. Call `KaTeX_noinstall=1 npm install` from the `Makefile` as above, and change the `Makefile` to turn `dist` into a no-op if that variable is set. May require GNU make; I'm not sure whether case distinctions at the top level are a feature of POSIX make.

3. Pass the `--ignore-scripts` flag to `npm install` in some clever way, as detailed in https://github.com/npm/npm/issues/9590. I'm not sure whether I should be slightly worried that this flag might get applied recursively, suppressing required install scripts for dependencies as well. If so we could end up in a hard-to-debug mess. Will need proper testing for this.

4. Invoke some other npm command, like `npm update`. In my experience doing so will install different versions, particularly also for nested packages, so I never managed to get this up and running.

kevinbarabash:
Updating `.gitignore` seems like the simplest solution.  The build step itself is fast enough.

gagern:
One problem with this minor modification is that the build step includes things like style changes. So you can't use `make live` to evaluate the functionality of some piece of code if that piece of code is syntactically invalid. I've seen this in relation to my render-to-canvas code, which was written against an older set of style rules. Merging that into latest master meant I couldn't start a live server before I had tweaked all the affected style choices.

gagern:
> 2. Call `KaTeX_noinstall=1 npm install` from the `Makefile` as above, and change the `Makefile` to turn `dist` into a no-op if that variable is set. May require GNU make; I'm not sure whether case distinctions at the top level are a feature of POSIX make.

We are already using more make functionality than guaranteed by POSIX, so I'd say we already depend on GNU make as it is, and using its functionality for one more application is a good thing.

kevinbarabash:
[...]

PR #546, opened 2016-10-09
Make

gagern:
This addresses the issues raised in #544.

First we let git ignore `.npm-install.stamp` and `dist`. Adding the latter to `.gitignore` should _not_ affect npm packaging, since that is based on a whitelist in `package.json` which does mention `dist`. I tested that `npm pack .` does still include `dist` after this modification.

Then we change the `Makefile` to only fetch dependencies but do not build KaTeX itself. The make conditionals used here are not part of POSIX make but a GNU extension.  But we already use functionality not mandated by POSIX (namely many of the functions like `wildcard`), so this should not make portability any worse than it already is.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,8 @@ node_modules
 npm-debug.log
 last.png
 diff.png
+/.npm-install.stamp
+/dist/
 /test/screenshotter/tex/
 /test/screenshotter/diff/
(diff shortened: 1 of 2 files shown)


kevinbarabash:
LGTM.  Thanks for improving the dev workflow.

PR #549, opened 2016-10-11
Fix texcmp issues related to bit depth and kerning

gagern:
* Ensure bit depth 8
* Print affected file if PNG failed to read (e.g. due to wrong bit depth)
* Disable running Kern test case through TeX as doing so fails


kevinbarabash:
Changes look good.  Were some of the PNGs being output with 16 bits per channel?

gagern:
Using my local TexLive 2016 I indeed saw some 16 bps files. Don't remember which ones.

PR #553, opened 2016-10-25
Create CODE_OF_CONDUCT.md

mroth:


kevinbarabash:
@mroth thanks for the PR.  I'll have a look tomorrow.

kevinbarabash:
@mrot this is really good!  What differences, if any, are there from http://contributor-covenant.org/version/1/4/?

mroth:
@kevinbarabash Apologies, looks like this all happened while I was out of the country last week.  The only change in this PR from the standard doc was adding our new external email address, but please see our KA `#opensource` Slack channel for much of the discussion around this thus far. 

(Which is bad for transparency reasons I know, but that just happened to be where it was most convenient to discuss at the time.)

Issue #533, opened 2016-09-16
\color causes incorrect spacing with some adjacent elements when `\displaystyle` is present in the coloured block


daniel3735928559:
**Steps performed**: On either of Chromium 53 or Firefox 48.0.2 (both on Linux), navigated to https://khan.github.io/KaTeX/ and entered `\color{red}{\displaystyle\int x}+1` and `\displaystyle\int x+1` into the "Type an expression" box.

**Expected behaviour**: The two expressions should look the same, but one having the integral coloured red and the other not.

**Observed behaviour**: In the coloured version, there is very little space between the x and the + compared to in the non-coloured version.

This issue also exists when testing locally by running `node server.js` on a local clone of KaTeX.  It seems to be related to the use of `\displaystyle`, since without that it behaves as expected (though then the integral is rendered smaller on the local version; for some reason in the online demo the integral is rendered at displaystyle size regardless.)

I made a brief effort to chase this up but don't have time to get into it in the immediate future, so I thought I'd document it in case someone else runs into it or knows a quick fix.

kevinbarabash:
@daniel3735928559 thanks for the report.  I was able to reproduce this.  MathJax and LaTeX handles make sure the correct space is there.  In the markup we wrap the group in a `<span>`, but this makes solving the problem complicated because we can't make the wrapper an `mord` because then spacing on the left side of the group might be wrong.  Maybe there's a way that we can output groups without the extra `<span>` wrapper if it's not needed.

kevinbarabash:
`{\displaystyle\int x}+1` renders correctly, so in that case we're not adding an extra `<span>` around `\int x`.  Also, the color is applied independently to each glyph.

Issue #541, opened 2016-09-29
Got wrong space of LaTeX "\!=\!=\!="?


bzsy:
hi all,

I have tried to render **`2Na_2O_2+2H_2O\ \!=\!=\!=\ 4NaOH+O_2\uparrow`** by KaTeX and MathJax.
It looks like KaTeX ignored the`\!` before `=`.
Hope anyone can help, thanks a lot :)

**KaTeX:**
![image](https://cloud.githubusercontent.com/assets/3096023/18949291/4e99e826-866d-11e6-84bf-00edba35e869.png)

**MathJax:**
![image](https://cloud.githubusercontent.com/assets/3096023/18949311/5ea1126c-866d-11e6-981c-a209127d1214.png)

sthoch:
Interesting: `==` results in less spacing between the equal signs than `=\!=` ...

As a workaround you can use `{=}\!{=}\!{=}`.

gagern:
I can think of two possible reasons.

1. A binary operation which is not surrounded by stuff compatible with a binary operation gets demoted to an ordinary symbol. According to #4 we don't do that yet, although #492 contains a pull request for this which needs review.

2. I'm not sure how spacing constructs factor into this Ord/Bin/… classification. It might well be that they are ignored for the sake of adjacency computation. (Someone should read this up in the TeX book but I don't have time just now.) If so, I guess our CSS doesn't reflect this fact.

bzsy:
Thank you guys. `{=}\!{=}\!{=}` works well.

kevinbarabash:
Fixed on `master`.

PR #494, opened 2016-06-21
Make screenshotter work with docker-machine

gagern:
Since [`boot2docker` has been superseded by `docker-machine`](https://docs.docker.com/engine/installation/mac/), we need this to support developers on OS X.  The changes to the bash script ensure that we clean up our containers even if taking screenshots gets hung and requires a keyboard interrupt, as happens if the IP addresses were guessed incorrectly.

PR #565, opened 2016-11-20
Fix #534: Teach getTypeOfGroup about font groups.

### LINKED ISSUE #534: \mathbf with subscripts/superscripts causes incorrect spacing
```
\[ A*B+C-D = E \]
\[ A_1 * B^2 + C_3^4 + D^5_6 = E_0 \]
\[ \mathbf{A}*\mathbf{B}+\mathbf{C}-\mathbf{D} = \mathbf{E} \]
\[ \mathbf{A}_1 * \mathbf{B}^2 + \mathbf{C}_3^4 + \mathbf{D}^5_6 = \mathbf{E}_0 \]
```

KaTeX on Chrome 52.0.2743.116 m:

image

MathJax on Chrome 52.0.2743.116 m:

image

kohler:
The type of a font group is the type of its body. This teaches KaTeX that
a span created for a font group, such as for `\mathbf{X}^2`, has the same
type as the underlying group.


kevinbarabash:
@kohler thanks for the PR.  Can you check if this fixes https://github.com/Khan/KaTeX/issues/533 as well?  Also, could you add a screenshot test for this?

kohler:
#533 is a different problem and requires a different fix. This PR has been updated with that fix.

kohler:
Sorry @kevinbarabash Trying to fix/add tests—I’m no Docker expert. On my Mac with newly-installed Docker for Mac, `dockers/Screenshotter/screenshotter.sh --verify` seems to just hang. Should I expect that?

kevinbarabash:
@kohler can you post what you're seeing the console?

kohler:
Docker for Mac doesn't support networking in the usual way, apparently. There’s some stuff on their website about it. I had to supply both `--selenium-ip` and `--katex-ip` arguments, rather than/in addition to `--container`.

The current PR is failing two screenshot tests, LimitControls and SupSubLeftAlignReset. But I think the PR’s output is better: the older output (in blue) had less space than it should; the newer output (in red) restores the space, and looks visually closer to the output of straight TeX.

If you agree, I’ll update these screenshot tests and add some new ones.

limitcontrols-diff
supsubleftalignreset-diff

gagern:
> Docker for Mac doesn't support networking in the usual way, apparently.

Yay, yet another way to install Docker, after I got docker-machine to work for me in #494. Makes me wonder whether I should try to come to terms with this, or whether I should just wait to see what approach they use next month.

> There’s some stuff on their website about it.

I guess you mean [Networking features in Docker for Mac](https://docs.docker.com/docker-for-mac/networking/).

> I had to supply both `--selenium-ip` and `--katex-ip` arguments, rather than/in addition to `--container`.

Did you add another IP address to your loopback device, as suggested by these documents? If not, what did you use for the `--katex-ip` and how did you find it?

I'm currently thinking about how to auto-detect this situation, and handle it with less need for user intervention.

kohler:
> I guess you mean Networking features in Docker for Mac.

Yep!

> Did you add another IP address to your loopback device, as suggested by these documents? If not, what did you use for the `--katex-ip` and how did you find it?

I tried the IP network alias suggested by that document, choosing the 172.17/16 network suggested by the screenshotter’s Docker config parsing. It didn’t work, so I just supplied my host’s local IP, found by `ifconfig en0`. This required some changes to screenshotter.js—included in this PR—so that `--katex-ip` is obeyed even in the presence of `--container`.

kohler:
I’m also testing a fix for #541. This fix involves making `.mspace` nodes their own class, rather than examples of `.mord`. TeX glues aren’t atoms.

The combination of that fix and this PR restores the original SupSubLeftAlignReset screenshotter test behavior.

How to move forward? Would you prefer separate fixes for all these, plus a PR for the screenshotter changes?

@kevinbarabash

gagern:
Personally I'd say the screenshotter is something semantically distinct, so that should be a separate PR. If the effects of the other two really cancel out completely, then having a single PR with two commits might be nice, to see that it really remains unmodified. In that case, it would be nice to rebase the branch so that it has two commits for the two distinct changes. But that may wait till after review. And if the image just looks mostly the same, but not pixel-exact, then I'd go for two PRs there as well.

kohler:
Closing this in favor of a new PR, #567.

PR #567, opened 2016-11-21
Fix #533, #534, #541.

### LINKED ISSUE #533: \color causes incorrect spacing with some adjacent elements when `\displaystyle` is present in the coloured block

kohler:
- #534: Implement getTypeOfGroup for font groups.
- #533, #541: Improve the ways spaces are applied to lists. Since
  CSS adjacency implements mathematical spacing, it's incorrect to
  introduce "convenience spans" for spaces and display changes into
  the generated HTML -- those spans break adjacency. Apply display
  changes directly, and shift space spans into adjacent atoms.

Requires updates to two screenshotter tests, LimitControls and
SupSubLeftAlignReset. The new results for these tests are closer
to TeX output than the old results.

Also requires updates to Jasmine tests, since those assumed output
structures that have changed.


kohler:
I have a fix for #4 that's based on this commit, and will submit it for review after hearing back.

kohler:
Hi, any chance I could talk to a contributor about the approach taken by this PR?

KaTeX's approach to TeX operator spacing uses CSS adjacency rules, as in `.mathord + .mathbin { margin-left: ... }`. This means that, for correct spacing, KaTeX must ensure that CSS adjacency corresponds to TeX operator adjacency.

But in some cases KaTeX doesn't ensure this. Style changes, size changes, and spaces introduce spurious spans into the HTML output. These spans mess up operator spacing: TeX considers two boxes adjacent (and space them accordingly), but in CSS the spans aren't adjacent. The same root cause is behind #533, #541, #136, and to some extent #4.

The approach taken here—eliminating spurious spans—seems like a good fit for current KaTeX, although it is not a trivial change. I've enjoyed working on these features and others but if there's no chance of merge I should do something else.

@kevinbarabash

kevinbarabash:
@kohler definitely interested in this PR.  I'll to do a first past this weekend.

static/katex.less
```diff
@@ -101,111 +101,120 @@
...
+        & + .mord.mtight { margin-left: @thinspace; }
+        & + .mop.mtight { margin-left: @thinspace; }
+    }
+
+    .mclose {
+        & + .mop.mtight { margin-left: @thinspace; }
+    }
+
+    .minner {
+        & + .mop.mtight { margin-left: @thinspace; }
```
kevinbarabash: I've checked these changes against https://books.google.ca/books?id=rstheMEAf2gC&pg=PA257&lpg=PA257&dq=texbook+ord+bin+rel+op+punct+spacing+rules&source=bl&ots=TU_zSmaVkr&sig=4ndj36478T4s9EI7OgrIAwwlUws&hl=en&sa=X&ved=0ahUKEwjltL3-zMfQAhUC9IMKHZhVCK0Q6AEIHTAA#v=onepage&q=texbook%20ord%20bin%20rel%20op%20punct%20spacing%20rules&f=false.  They look good.

test/screenshotter/ss_data.yaml
```diff
@@ -36,6 +37,7 @@ Cases: |
 Colors:
     tex: \blue{a}\color{#0f0}{b}\color{red}{c}
     nolatex: different syntax and different scope
+ColorSpacing: \color{black}{\displaystyle \int x} + 1
```
kevinbarabash: Can you change the color to something other than black?

test/screenshotter/ss_data.yaml
```diff
@@ -106,6 +109,7 @@ Sqrt: |
     ^{\sqrt{\sqrt{\sqrt{x}}}}}
 SqrtRoot: |
     1+\sqrt[3]{2}+\sqrt[1923^234]{2^{2^{2^{2^{2^{2^{2^{2^{2^{2^{2^2}}}}}}}}}}}
+StyleSwitching: a\cdot b\scriptstyle a\cdot b
```
kevinbarabash: Could you add some style switches at boundaries between atom types, e.g. `a\cdot\textstyle b`

src/buildCommon.js
```diff
@@ -35,7 +35,7 @@ var mainitLetters = [
  * Correctly pulls out metrics for the character, and optionally takes a list of
  * classes to be attached to the node.
  */
-var makeSymbol = function(value, style, mode, color, classes) {
+var makeSymbol = function(value, style, mode, options, classes) {
```
kevinbarabash: Can you rename `style` to `fontFamily`?

src/buildHTML.js
```diff
@@ -80,12 +118,13 @@ var getTypeOfGroup = function(group) {
...
-        return getTypeOfGroup(group.value.value);
-    } else if (group.type === "styling") {
-        return getTypeOfGroup(group.value.value);
+    } else if (group.type === "color" || group.type === "sizing"
+               || group.type === "styling") {
+        // Return type of rightmost element of group.
+        var atoms = group.value.value;
+        return getTypeOfGroup(atoms[atoms.length - 1]);
+    } else if (group.type === "font") {
+        return getTypeOfGroup(group.value.body);
```
kevinbarabash: Shouldn't we get the rightmost atom of the `group.value.body` for `fonts?

src/buildHTML.js
```diff
@@ -17,18 +17,56 @@ var utils = require("./utils");
...
+            if (lastSpace === -1) {
+                lastSpace = i;
+            }
+        } else if (lastSpace >= 0) {
+            var node = groups[i];
+            if (node instanceof domTree.symbolNode) {
+                node = groups[i] = makeSpan(node.classes, [node]);
+            }
+            buildCommon.prependChildren(node,
+                groups.splice(lastSpace, i - lastSpace));
```
kevinbarabash: I found this kind of hard to follow.  I think maybe doing two passes would make this more understandable, in the first pass we could add divide `groups` up into set of arrays and the second pass could create the spans.  This should avoid the need for `prependChildren`.

src/buildHTML.js
```diff
@@ -392,12 +439,17 @@ groupTypes.genfrac = function(group, options, prev) {
...
     var nstyle = style.fracNum();
     var dstyle = style.fracDen();
+    var newOptions;
 
-    var numer = buildGroup(group.value.numer, options.withStyle(nstyle));
-    var numerreset = makeSpan([style.reset(), nstyle.cls()], [numer]);
+    newOptions = options.withStyle(nstyle);
+    var numer = buildGroup(group.value.numer, newOptions);
+    var numerreset = makeSpan([style.reset(), nstyle.cls()],
+        [numer], newOptions);
```
kevinbarabash: Is `newOptions` necessary for the `numerreset`?  The `.reset-X` rules don't seem to interact with `.tight`.  Am I missing something?

src/buildHTML.js
```diff
@@ -684,14 +736,14 @@ groupTypes.spacing = function(group, options, prev) {
         // things has an entry in the symbols table, so these will be turned
         // into appropriate outputs.
         return makeSpan(
-            ["mord", "mspace"],
+            ["mspace"],
```
kevinbarabash: Why did `"mord"` disappear?

src/buildHTML.js
```diff
@@ -684,14 +736,14 @@ groupTypes.spacing = function(group, options, prev) {
...
             [buildCommon.mathsym(group.value, group.mode)]
         );
     } else {
         // Other kinds of spaces are of arbitrary width. We use CSS to
         // generate these.
         return makeSpan(
-            ["mord", "mspace",
+            ["mspace",
```
kevinbarabash: Same.

src/buildHTML.js
```diff
@@ -750,8 +802,8 @@ groupTypes.op = function(group, options, prev) {
         // If this is a symbol, create the symbol.
         var fontName = large ? "Size2-Regular" : "Size1-Regular";
         base = buildCommon.makeSymbol(
-            group.value.body, fontName, "math", options.getColor(),
-            ["op-symbol", large ? "large-op" : "small-op", "mop"]);
+            group.value.body, fontName, "math", options,
+            ["mop", "op-symbol", large ? "large-op" : "small-op"]);
```
kevinbarabash: Thanks... this order for the classes is better.  It would be nice if the param order between `makeSpan` and `makeSymbol` was more similar.  Can you add a TODO to clean that up later (there's already enough changes in this PR)?

src/buildHTML.js
```diff
@@ -1058,17 +1116,25 @@ groupTypes.sizing = function(group, options, prev) {
...
-                   style.cls()],
-                  inner)]);
-
-    // Calculate the correct maxFontSize manually
     var fontSize = buildCommon.sizingMultiplier[group.value.size];
-    span.maxFontSize = fontSize * style.sizeMultiplier;
+    fontSize = fontSize * style.sizeMultiplier;
+
+    // Add size-resetting classes to the inner list and set maxFontSize
+    // manually.
```
kevinbarabash: Is this to handle a size change inside of a color group?  Can you update this comment to include why we're doing this?

src/buildHTML.js
```diff
@@ -1058,17 +1116,25 @@ groupTypes.sizing = function(group, options, prev) {
...
+    for (var i = 0; i < inner.length; i++) {
+        var pos = utils.indexOf(inner[i].classes, "sizing");
+        if (pos < 0) {
+            inner[i].classes.push("sizing", "reset-" + options.size,
+                                  group.value.size, style.cls());
+            inner[i].maxFontSize = fontSize;
+        } else if (inner[i].classes[pos + 1] === "reset-" + group.value.size) {
+            inner[i].classes[pos + 1] = "reset-" + options.size;
+        }
+    }
```
kevinbarabash: It seems like this is going to result in a lot of `reset-` styles being added.  Is that the case?  It seems like this is unavoidable.

test/screenshotter/ss_data.yaml
kohler: Sure!

test/screenshotter/ss_data.yaml
kohler: Yep.

src/buildCommon.js
kohler: OK

src/buildHTML.js
kohler: No. Font bodies are not arrays—unlike for style, etc. You can see this both in the creation code (in `functions.js`) and in `groupTypes.font`, which calls `buildGroup()` on the body rather than `buildExpression` on the body array.

src/buildHTML.js
kohler: I've tried a different one-pass arrangement.

src/buildHTML.js
kohler: I doubt the newOptions are necessary, but neither are they harmful.

src/buildHTML.js
kohler: Because “mspace” nodes are no longer considered atoms. They are not atoms and TeX doesn't treat them as such. So they shouldn't affect operator spacing. If they were `mord` they'd affect operator spacing (e.g. if followed by an `mbin`)

src/buildHTML.js
kohler: OK

src/buildHTML.js
kohler: Good catch

src/buildHTML.js
```diff
@@ -1083,12 +1149,18 @@ groupTypes.styling = function(group, options, prev) {
...
 
     var newStyle = styleMap[group.value.style];
+    var newOptions = options.withStyle(newStyle);
 
     // Build the inner expression in the new style.
     var inner = buildExpression(
-        group.value.value, options.withStyle(newStyle), prev);
+        group.value.value, newOptions, prev);
+
+    // Add size-resetting classes to the inner list.
```
kohler: Good catch

src/buildHTML.js
kohler: The old code set maxFontSize manually as well. In case you meant the `pos` stuff, I've updated the comment.

kohler:
OK @kevinbarabash I’ve addressed your comments! Let me know if you'd like anything else.

src/buildHTML.js
kevinbarabash: Right b/c styles apply to everything which follows them whereas the font contains a single group.  It's been a while since I've looked at the code – can you tell?

src/buildHTML.js
kevinbarabash: In that case it's probably best to either always pass `options`.  There are a few places where it's not being passed, `groupTypes.spacing`, `groupTypes.sqrt`, `groupTypes.array`, and `buildExpression`.  Can you add it to those call sites.

src/buildHTML.js
kevinbarabash: That makes sense.

src/buildHTML.js
kevinbarabash: 👍

src/buildHTML.js
```diff
@@ -17,18 +17,60 @@ var utils = require("./utils");
...
+            if (j < groups.length) {
+                group = groups[j];
+                if (group instanceof domTree.symbolNode) {
+                    group = groups[j] = makeSpan(group.classes, [group]);
+                }
+                buildCommon.prependChildren(group, groups.splice(i, j - i));
+            } else {
+                i = j;
+            }
+        }
```
kevinbarabash: This is still really hard to follow.  I was thinking something more along the lines of the following (if I understand the original code correctly):
```
var spans = [];
var children = [];

for (i = 0; i < groups.length; i++) {
    var child = groups[i];
    if (isSpace(child && children.length > 0) {
        spans.push(makeSpan(children[0].classes, children));
        children = [];
    } else {
        children.push(child);
    }
}

if (children.length > 0) {
    spans.push(makeSpan(children[0].classes, children));
}

return spans;
```

src/buildHTML.js
kohler: Yeah no look let me push back on this. There are `makeSpan` calls in `delimiter.js`, etc., that as of this commit have no `options` to pass. How about I put a TODO to improve API consistency.

src/buildHTML.js
kohler: This doesn't do the same thing. Aside from the obvious bugs (the `isSpace(child)` case doesn’t actually put the current child anywhere), by introducing spurious spans, it would screw up class adjacency.

I think the code is pretty simple and you should take it as is.

kohler: The following code should work but is longer. If you really won't take the `splice` solution i'll do this.

```
    var spaces = [];
    var nextGroups = [];
    for (i = 0; i < groups.length; i++) {
        group = groups[i];
        if (isSpace(group)) {
            spaces.push(group);
        } else {
            if (spaces.length > 0) {
                if (group instanceof domTree.symbolNode) {
                    group = makeSpan(group.classes, [group]);
                }
                buildCommon.prependChildren(group, spaces);
                spaces = [];
            }
            nextGroups.push(group);
        }
    }
    if (spaces.length > 0) {
        Array.prototype.push.apply(nextGroups, spaces);
    }
```

kevinbarabash: Thanks for catching my mistake with `isSpace(child`.  

This is better.  I was confused as to what was being prepended.  `buildCommon.prependChildren(group, spaces);` makes that really clear.  Question about the `instanceof domTree.symbolNode` check: if it fails is it possible that the `group` being passed into `prependChildren` is not a span?

kohler: OK. I still don't find the original confusing. Would 

```
var spaces = groups.splice(i, j - i);
```

make it clearer. This isn't a huge deal but creating additional arrays isn't free; in the normal case where there are no spaces my code doesn't do any array allocation

kohler: As for `instanceof domTree.symbolNode`, as the immediately previous comment says,

```
    // At this point `groups` consists entirely of `symbolNode`s and `span`s.
```

because the documentFragments have been spliced in

src/buildHTML.js
kevinbarabash: A TODO sounds good.

kohler:
OK, I've added the TODO (in commit 0911e5f) and updated the problematic loop (in d3ecb33). I hope we are now ready to merge.

kevinbarabash:
@kohler thanks for updating the loop.  I left a minor nit in the inline comments.  I understand the desire to keep performance in check.  TBH, performance isn't something I've put a whole lot of thought into :(, but it's one of the selling points of KaTeX so I wouldn't want to do anything that adversely affects it.  We should probably have some sort of performance tests that we can run on each build to avoid perf regressions.

kevinbarabash:
> This slightly changes the vertical position of the Sizing test. Not
sure the vertical position matters, so change the test.

I noticed that too.  Can you run texcmp.sh to see how it compares against LaTeX's output?  See https://github.com/Khan/KaTeX/blob/e6de31d2d622f01aae6d465cf332a268332942f4/dockers/texcmp/README.md for details.

kohler:
It doesn't work, either on local or docker.

At least this diff is needed then I could get local to work

```
diff --git a/dockers/texcmp/texcmp.js b/dockers/texcmp/texcmp.js
index 3c16dc5..c164e5d 100644
--- a/dockers/texcmp/texcmp.js
+++ b/dockers/texcmp/texcmp.js
@@ -7,7 +7,7 @@ var path = require("path");
 var Q = require("q"); // To debug, pass Q_DEBUG=1 in the environment
 var pngparse = require("pngparse");
 var fft = require("ndarray-fft");
-var ndarray = require("ndarray-fft/node_modules/ndarray");
+var ndarray = require("ndarray");
 
 var data = require("../../test/screenshotter/ss_data");
 
```

Docker looks like some latex packages are missing from the docker

The diff is attached; there are font differences clearly; the baselines are equal, meaning the output of this PR is better than the previous

sizing

kevinbarabash:
@kohler thanks for checking the output for the Sizing test.  Hmm... I'm thinking the issue with `ndarray` is probably a different version of `npm`... it's using a flat directory structure now.  We should probably add `ndarray` as a direct dependency.

PR #568, opened 2016-11-22
Improved Docker for Mac support

gagern:
This should allow using a Docker for Mac install without the need for user intervention.

Docker for Mac is detected by looking at the operating system and the `DOCKER_HOST` environment variable. If users configure their docker host using a docker config file, we might not notice that fact. If the OS is Darwin and the environment variable is not set, we assume Docker for Mac is being used, connecting to a local Unix domain socket by default. In that case, we set the IP to connect to the Selenium server to `"localhost"` as we can't directly reach the bridge device inside the emulated Linux. In order to find an IP address which Selenium can use to connect back to KaTeX, we try connecting to all available network addresses, taking the first one where that connection succeeds.

There are two drive-by fixes in separate commits. One improves operations if the user names an incorrect test case via the `--include` command line switch. The other makes the server work if called from a directory other than the base directory of the KaTeX development tree.

### DIFF
--- a/dockers/Screenshotter/screenshotter.js
+++ b/dockers/Screenshotter/screenshotter.js
@@ -288,6 +288,11 @@ function takeScreenshot(key) {
     var itm = data[key];
     if (!itm) {
         console.error("Test case " + key + " not known!");
+        listOfFailed.push(key);
+        if (exitStatus === 0) {
+            exitStatus = 1;
+        }
+        oneDone();
         return;
     }
 
@@ -303,15 +308,7 @@ function takeScreenshot(key) {
     if (opts.wait) {
         browserSideWait(1000 * opts.wait);
     }
-    driver.takeScreenshot().then(haveScreenshot).then(function() {
-        if (--countdown === 0) {
-            if (listOfFailed.length) {
-                console.error("Failed: " + listOfFailed.join(" "));
-            }
-            // devServer.close(cb) will take too long.
-            process.exit(exitStatus);
-        }
-    }, check);
+    driver.takeScreenshot().then(haveScreenshot).then(oneDone, check);
 
     function haveScreenshot(img) {
         img = imageDimensions(img);
@@ -358,6 +355,16 @@ function takeScreenshot(key) {
             return promisify(fs.writeFile, file, buf).then(function() {
                 console.log(key);
             });
+        }
+    }
+
+    function oneDone() {
+        if (--countdown === 0) {
+            if (listOfFailed.length) {
+                console.error("Failed: " + listOfFailed.join(" "));
+            }
+            // devServer.close(cb) will take too long.
+            process.exit(exitStatus);
         }
     }
 }
--- a/dockers/Screenshotter/screenshotter.js
+++ b/dockers/Screenshotter/screenshotter.js
@@ -6,6 +6,7 @@ var fs = require("fs");
 var http = require("http");
 var jspngopt = require("jspngopt");
 var net = require("net");
+var os = require("os");
 var pako = require("pako");
 var path = require("path");
 var selenium = require("selenium-webdriver");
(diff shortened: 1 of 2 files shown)


kevinbarabash:
@gagern thanks for improving docker support.  I'll give it a try tomorrow.  It looks like there's some lint causing travis to fail.

kevinbarabash:
I gave this a try a got the following error:
```
docker: Error response from daemon: client is newer than server (client API version: 1.24, server API version: 1.21).
```
Then I ran: `docker-machine restart default` and waited for it complete, ran `docker-machine env` and re-exported the correct environment variables, and finally `make screenshots` and it worked like a charm.

There were some differences in `MathDefaultFonts-chrome.png` and `SupSubCharacterBox-chrome.png`.  I thought we retried screenshots that were different but I didn't see any indication of retries in the logs.

gagern:
Anyway, `docker-machine env` is the [Docker Toolkit](https://docs.docker.com/toolbox/overview/) approach, and while it is good to know that this still works as it used to, the new feature here is that it will work using [Docker for Mac](https://docs.docker.com/docker-for-mac/), too. That's a different setup, with [different tradeoffs](https://docs.docker.com/docker-for-mac/docker-toolbox/).

If the version mismatch is due to an upgrade in Docker Tools, that would be that. If it's because you installed Docker for Mac, too, then notice that https://github.com/docker/for-mac/issues/455 makes it difficult to have both coexist at different docker API versions.

The screenshot retries only affect the `--verify` mode. In that mode, the screenshotter will *never* write files to disk, but it will compare obtained screenshots against those on disk, and will re-try a number of times if there is a difference. In non-verify write-to-disk mode, the first obtained screenshot gets written, although one may use a timeout to wait for things to settle down before taking the screenshot.

kevinbarabash:
After some debugging, I discovered that `boot2docker` was still installed.  After deleting that it works as expected.

dockers/Screenshotter/screenshotter.js
```diff
@@ -273,6 +276,64 @@ function imageDimensions(img) {
...
+    }
+    console.log("Looking for host IP among " + ips.join(", "));
+
+    // Load a data: URI document which attempts to contact each of these IPs
+    var html = "<!doctype html>\n<html><body>\n";
+    html += ips.map(function(ip) {
+        return '<script src="http://' + ip + ':' + katexPort +
+            '/ss-connect.js?ip=' + encodeURIComponent(ip) +
+            '" defer></script>';
+    }).join("\n");
```
kevinbarabash: wow... I'm surprised there isn't an easier way/some library that already does this.

kevinbarabash:
@gagern as an aside, I was thinking we might set the `"engine"` field in `package.json` to node 6.5+ so that we have access to ES6 constructs in the docker js scripts.  It's part of a larger theme of modernizing the code base, see https://github.com/Khan/KaTeX/projects/5.

gagern:
@kevinbarabash, I noticed you squash-merged my commits, even though they addressed different issues. Out of curiosity, do you prefer having such changes lumped together, or should I continue generating distinct commits for distinct issues in the future?

I'm all for using ES6 in all our development scripts. Do you want to start a branch and collect related changes, or do you want to just bump the engine requirement, and take subsequent improvements as separate pull requests?

How about the main code base? We could [babelify](https://github.com/babel/babelify) or [es6ify](https://github.com/thlorenz/es6ify) the code there, to end up with ES5 runtime code despite using ES6 features for development. Some parts of the codebase might benefit considerably from this.

kevinbarabash:
@gagern my mistake squash-merging.  :(

I think bumping the engine requirement and then updating the code afterwards is probably the most straight forward.

For the main codebase, let's go with babelify (it's more up to date).  There are certain things I think we should take care to do in order to keep size down.  I'll open a separate issue with details.

PR #570, opened 2016-11-23
Check out test fonts in screenshotter.sh

gagern:
This ensures that running screenshots as indicated by the screenshotter readme will correctly render the Unicode examples, even if not run on Travis.  It also fixes the commit ID of the unicode fonts.

One alternative would have been using a git submodule.  But as many tools will initialize submodules automatically, that would mean big downloads for anyone using such a tool to clone KaTeX.

Another alternative would be tweaking the server to fetch the files on demand if the local copy is unavailable.  But that would cause additional network overhead, so grabbing the files just once should be simpler.

### DIFF
--- a/dockers/Screenshotter/screenshotter.sh
+++ b/dockers/Screenshotter/screenshotter.sh
@@ -14,6 +14,18 @@ cleanup() {
     container=
 }
 
+SS_DIR=$(dirname "$0")
+TOP_DIR=${SS_DIR}/../..
+FONTS_DIR=${TOP_DIR}/test/screenshotter/unicode-fonts
+if [[ ! -d "${FONTS_DIR}" ]]; then
+    echo "Cloning test fonts repository"
+    git clone https://github.com/Khan/KaTeX-test-fonts "${FONTS_DIR}" \
+        || exit 2
+fi
+pushd "${FONTS_DIR}" || exit 2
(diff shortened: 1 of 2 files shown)


dockers/Screenshotter/screenshotter.sh
```diff
@@ -14,6 +14,18 @@ cleanup() {
...
+TOP_DIR=${SS_DIR}/../..
+FONTS_DIR=${TOP_DIR}/test/screenshotter/unicode-fonts
+if [[ ! -d "${FONTS_DIR}" ]]; then
+    echo "Cloning test fonts repository"
+    git clone https://github.com/Khan/KaTeX-test-fonts "${FONTS_DIR}" \
+        || exit 2
+fi
+pushd "${FONTS_DIR}" || exit 2
+git checkout --detach 99fa66a2da643218754c8236b9f9151cac71ba7c || exit 2
+popd || exit 2
```
kevinbarabash: `pushd`/`popd` look like they're super helpful.  What happens if you pushd and then the checkout fails?

kevinbarabash: I was thinking it might leave the user in FONTS_DIR, but after a quick test I see that's not how shell scripts work.  I don't do much shell program.

Issue #354, opened 2015-09-11
incorrect spacing around dotless i (and j)


kevinbarabash:
See https://github.com/Khan/KaTeX/pull/349 for details.

kevinb7:
Reported by @gagern.

gagern:
A texcmp comparison, with TeX in green and KaTeX in red:
![mathdefaultfonts](https://cloud.githubusercontent.com/assets/190759/9790459/718cd134-57d6-11e5-8766-54b5cff155f6.png)

Right now I can think of three possibilities: either the glyph is positioned differently in the font, or the typesetting mechanics do some italic correction or some such differently, or TeX adds some explicit spacing in a different way.

kevinbarabash:
For some reason we add extra `margin-right` on the preceding character.

kevinbarabash:
Moving this to the backlog.

Issue #563, opened 2016-11-15
Add support for \atop


NexiePrado:
I just trying to render formule like
```
$\\mathrm{\\normalsize{~+2~~~~+3~~~+4~-2~~~~}\\atop\\Large{Mg_xAl_2(SiO_4)_3}}$
```
But it rendered nothing. How I can render nested text expressions?

kevinbarabash:
We don't support `\atop` yet.

kevinbarabash:
Support has been added to `master`.

Issue #530, opened 2016-09-14
Support '\buildrel'


daniloascione:
Please add support for plain Tex \buildrel macro.
Ex: R^t\buildrel\gamma\over\longrightarrow p

kevinbarabash:
@daniloascione we're keen to support `\stackrel` because we have a fair amount of content that uses that.  It seems like `\stackrel{\gamma}{\longrightarrow} should do the same thing once `\stackrel` is implemented.

kevinbarabash:
I'm going to close this as support for `\stackrel` has been added to `master`.

Issue #498, opened 2016-07-05
add support for /textrm


szhuge:
Just discovered that \textrm doesn't work but \mathrm does.  We use \textrm, would be awesome to have that supported as well!

spicyj:
We do have \text.

deltakam:
Then, it wouldn't it be great to have \textrm as an alias to \text? Equations are usually not written just to work with KaTex.

I would vote for aliasing \mbox, too.

spicyj:
mbox and text differ at least in their behavior when used in a subscript or superscript and you should generally prefer text in LaTeX too. I'm not sure on text vs textrm.

spicyj:
To be clear: not saying we shouldn't support either of those, but we should make sure to do it properly.

deltakam:
I agree that we should do it properly, but if it will take forever, it would be better to have it as an alias -- it will have different effect so it might be ugly but for most cases not too serious. I would rather have it ugly than having the whole equation not drawn properly.

PR #573, opened 2016-11-28
Fix atom-spacing, add text-mode fonts, and miscellany

kohler:
* Complete fix for spacing between groups.
    * Complete implementation of TeX rules for devolving `bin` atoms to `ord`.
    * Fix class of `\llap` and `\rlap`.
    * Mark delimiters as `open` and `close` appropriately.
    * Internals cleanup.
* Add `\mathop`, `\mathrel`, etc. class-setting commands.
* Add `\stackrel`.
* Support `\atop`.
* Text-mode improvements.
    * Do not apply italic correction to characters in text mode; TeX does not.
    * Combine adjacent spans when possible, improving browser font rendering.
    * Support `\text{rm,it,bf,tt,sf,sc,normal}`.
* Symbol improvements.
    * Add text-mode support for several symbols and punctuations.


kohler:
This is one PR rather than several because I've already spent more time on this than I can afford. Thanks for KaTeX.

kevinbarabash:
@kohler wow, this is an epic PR... so many new features and bug fixes!  Each commit look pretty self contained so I'll just merge this without squashing.

gagern:
Great set of changes! I've looked through them, and wrote down some comments and questions.

Re 9afac38da71a99dc31c5d77f4100140caa1b56e3: The `\atopfrac` is artificial, not a real TeX command, right? Is there an AMS command which has the same effect but follows the `\‹name›{‹group1›}{‹group2›}` convention? If not, perhaps we should use some artificial name which can't be generated as a token by the lexer. Something like `\atop-frac` or even `atop-frac` should work, I think, since at that point the leading `\` is irrelevant.

Re 8882bb538540a8d363cd525e15ff8ddd10401622: See https://github.com/Khan/KaTeX/pull/261#issuecomment-114797510 and following comments for prior discussion. To make it short, as far as I understand it the consensus regarding unicode mappings was to either verify that some Unicode-supporting TeX implementation does handle the symbols like this, or to put them behind some opt-in guard. Otherwise we risk incompatibility with Unicode-aware TeX versions. Unlikely in this case, I think, but it's hard to draw the line. So I'll leave it to others to decide whether a conflict of this kind is sufficiently unlikely in this case.

Re eccf74e2a2cee4b233c19ced199ff95fd5bccb45: As you only use `toplist[0]`, the rest of that argument apparently gets discarded. Is that what TeX does, too?

Re e0b5d0a8711151ad0f20f1dc27d4a08ccc88885b: Out of curiosity (and for future reference), have you found this in a book, in the TeX sources, or by dumping TeX data structures?

Re 161337215b981d94e2c8a9b6432fc3e575cda6df: I'd avoid modifying the `classes` argument: that may lead to surprises as it's kind of unexpected from the perspective of the caller.

Re 1448c371ccaaae07316a0b44499f68f8c42e6005: Can you point out the TeX book chapter or similar where the rules for italic correction are spelled out?

Re c05e323931bcde92c8ec97e7254cf30435ca7921: The changes in the screenshots are pretty small. Perhaps I should try to create a visual diff of these, if I find the time.

Re 6a1aa76e4c2cf90be5c0dc8573360f13d87e7064: Can we rely on the small-caps variant getting rendered in a sufficiently compatible fashion on all platforms, or could some browsers decide to use an installed font which has a built-in small-caps variant instead of generating small caps from the named font? How come we don't need metric information for the small caps variant here?

kohler:
Great! Comments all addressed in latest commits.

Re 9afac38: Yes, `\atopfrac` is artificial. I couldn’t find an AMS equivalent; people suggest `\genfrac`. Using a non-typeable name seems reasonable.

> Re 8882bb5: See #261 (comment) and following comments for prior discussion. …

I can only partially understand that comment chain; but to your question, yes, pdfeTeX supports these symbols (curly quotes, en dash, em dash) with the meanings I assign.

```tex
\documentclass{article}
\RequirePackage[utf8]{inputenc}
\begin{document}
“hello”
‘hello’
—hello
–hello
\end{document}
```

> Re e0b5d0a: Out of curiosity (and for future reference), have you found this in a book, in the TeX sources, or by dumping TeX data structures?

Dumping TeX data structures and looking at sources.

> Re eccf74e: As you only use toplist[0], the rest of that argument apparently gets discarded. Is that what TeX does, too?

Very nice catch!! This was a bug. The correct thing to pass is just `top` since supsubs are special. Updated the test.

> Re 1613372: I'd avoid modifying the classes argument: that may lead to surprises as it's kind of unexpected from the perspective of the caller.

OK

> Re 1448c37: Can you point out the TeX book chapter or similar where the rules for italic correction are spelled out?

There is an index reference to “italic corrections” in the texbook. p441 is where it is said that italic corrections are added to characters encountered in math mode. Elsewhere is where it is said that italic corrections are not added in horizontal mode.

> Re 6a1aa76: Can we rely on the small-caps variant getting rendered in a sufficiently compatible fashion on all platforms, or could some browsers decide to use an installed font which has a built-in small-caps variant instead of generating small caps from the named font? How come we don't need metric information for the small caps variant here?

*Shrug.* The PR’s textsc is better than nothing but the answers to your questions are not the ones you want. If you don't want a halfway-decent-but-not-perfect textsc, then I will remove textsc from the PR.

I also added another commit which supports, in text mode, TeX’s ligatures ?\` and !\`, which compile to ¿ and ¡. (TeXbook page 51)

kohler:
Actually you probably don’t want that last commit (2434621) b/c the katex fonts lack the ¿ and ¡ characters, so the browser must fall back to its default fonts.

src/buildHTML.js
```diff
@@ -21,27 +21,48 @@ var isSpace = function(node) {
...
+
 /**
  * Take a list of nodes, build them in order, and return a list of the built
- * nodes. This function handles the `prev` node correctly, and passes the
- * previous element from the list as the prev of the next element, ignoring
- * spaces. documentFragments are flattened into their contents, so the
- * returned list contains no fragments.
+ * nodes. documentFragments are flattened into their contents, so the
+ * returned list contains no fragments. `isComplete` is true if `expression`
+ * is complete (no atoms will be added on either side).
```
kevinbarabash: Maybe `isRealGroup` would be a better name than `isComplete`.  The `groupType` handlers for things like `color`, `styling`, etc. where `isComplete` is `false` are "groups" that disappear after this processing.

src/buildHTML.js
```diff
@@ -125,6 +125,8 @@ var groupToType = {
     leftright: "minner",
     sqrt: "mord",
     accent: "mord",
+    llap: "mord",
+    rlap: "mord",
```
kevinbarabash: How did you determine that TeX always generates `mord` atoms?

src/delimiter.js
```diff
@@ -494,19 +497,22 @@ var makeCustomSizedDelim = function(delim, height, center, options, mode) {
...
     }
 };
 
 /**
  * Make a delimiter for use with `\left` and `\right`, given a height and depth
  * of an expression that the delimiters surround.
  */
-var makeLeftRightDelim = function(delim, height, depth, options, mode) {
+var makeLeftRightDelim = function(delim, height, depth, options, mode,
+                                  classes) {
```
kevinbarabash: There aren't any updates to any of the other screenshots.  Did these changes have an effect on any of the delimiters?

src/functions.js
```diff
@@ -378,6 +371,19 @@ defineFunction([
...
 
+// \mathop class command
+defineFunction("\\mathop", {
+    numArgs: 1,
+}, function(context, args) {
+    var body = args[0];
+    return {
+        type: "op",
+        limits: false,
+        symbol: false,
```
kevinbarabash: Maybe this should be `symbol: true,`, from the TeXbook (page 178):

> ‘\mathop{\phantom\sum}’ will leave exactly the right amount of blank space. (The ‘\mathop’ here makes this phantom behave like \sum, i.e., as a large operator.)

src/buildHTML.js
```diff
@@ -101,66 +101,20 @@ var buildExpression = function(expression, options, isComplete) {
...
+    if (node instanceof domTree.documentFragment) {
+        if (node.children.length) {
+            return getTypeOfDomTree(
+                node.children[node.children.length - 1]);
+        }
     } else {
-        return groupToType[group.type];
+        if (utils.contains(["mord", "mop", "mbin", "mrel", "mopen", "mclose",
+                            "mpunct", "minner"], node.classes[0])) {
+            return node.classes[0];
```
kevinbarabash: I worry about this.  Even though the class always comes first, it's not obvious that it has to come first.

src/functions.js
```diff
@@ -583,6 +587,9 @@ defineFunction(["\\over", "\\choose"], {
         case "\\choose":
             replaceWith = "\\binom";
             break;
+        case "\\atop":
+            replaceWith = "\\\\atopfrac";
+            break;
```
kevinbarabash: Update one of the screenshot tests to include this?

src/symbols.js
```diff
@@ -652,3 +652,13 @@ for (i = 0x0410; i <= 0x044F; i++) {
...
+
+// Unicode versions of existing characters
+defineSymbol(text, main, textord, "\u2013", "–");
+defineSymbol(text, main, textord, "\u2014", "—");
+defineSymbol(text, main, textord, "\u2018", "‘");
+defineSymbol(text, main, textord, "\u2019", "’");
+defineSymbol(text, main, textord, "\u201c", "“");
+defineSymbol(text, main, textord, "\u201d", "”");
+defineSymbol(text, main, textord, "\u00a1", "¡");
+defineSymbol(text, main, textord, "\u00bf", "¿");
```
kevinbarabash: The fonts don't include these, but we could render them and use a CSS transform to flip them.  You should also add metrics for these in fontMetrics.js.  I think the height/depth for `y` should be pretty close.

src/buildCommon.js
```diff
@@ -47,8 +47,12 @@ var makeSymbol = function(value, fontFamily, mode, options, classes) {
 
     var symbolNode;
     if (metrics) {
+        var italic = metrics.italic;
+        if (mode === "text") {
+            italic = 0;
+        }
```
kevinbarabash: I was hoping this would fix https://github.com/Khan/KaTeX/issues/354.

src/domTree.js
```diff
@@ -220,6 +224,34 @@ function symbolNode(value, height, depth, italic, skew, classes, style) {
     }
 }
 
+symbolNode.prototype.tryCombine = function(sibling) {
+    if (!sibling
+        || !(sibling instanceof symbolNode)
+        || this.italic > 0
+        || createClass(this.classes) !== createClass(sibling.classes)
```
kevinbarabash: I wonder if `createClass` should sort it's the classnames?  Maybe add a TODO by the `createClass` definition.

src/domTree.js
```diff
@@ -220,6 +224,34 @@ function symbolNode(value, height, depth, italic, skew, classes, style) {
...
+            && this.style[style] !== sibling.style[style]) {
+            return false;
+        }
+    }
+    for (style in sibling.style) {
+        if (sibling.style.hasOwnProperty(style)
+            && this.style[style] !== sibling.style[style]) {
+            return false;
+        }
+    }
```
kevinbarabash: Could you add a `.equals` operator to `Style` or a `equals` function to utils.js.

src/functions.js
```diff
@@ -126,16 +126,27 @@ defineFunction("\\sqrt", {
     };
 });
 
-// Some non-mathy text
-defineFunction("\\text", {
+// Non-mathy text, possibly in a font
+var textFunctionStyles = {
+    "\\text": undefined, "\\textrm": "mathrm", "\\textsf": "mathsf",
+    "\\texttt": "mathtt", "\\textnormal": "mathrm", "\\textbf": "mathbf",
+    "\\textit": "textit", "\\textsc": "textsc",
```
kevinbarabash: Can you add a `mathit` as well?

static/katex.less
```diff
@@ -49,6 +49,18 @@
...
+        font-style: normal;
+    }
+
+    .textit {
+        font-style: italic;
+    }
+
+    .textsc {
+        font-variant: small-caps;
+    }
```
kevinbarabash: Since `small-caps` doesn't have super wide support yet we should probably leave this out or make it optional for now.  A more robust way of implementing this would be to use CSS transforms to scale normal caps.  Unfortunately, transforms don't affect the flow and so then have to adjust for the gaps left.  :(

src/delimiter.js
kohler: Not according to the screenshots or a couple other random tests

src/domTree.js
kohler: Mneh. That would never be useful in current KaTeX and would have nonzero cost.

src/domTree.js
kohler: Why?

src/symbols.js
kohler: Maybe the best thing is to leave ¡ and ¿ out of this megaPR and open a separate issue for them

src/domTree.js
kohler: To be clearer, the consequence of not adding `.equals` is that the following:

`\text{\color{black}{X}\color{black}{Y}}`

will come out as two black spans rather than a single black span, whereas

`\text{\color{black}{XY}}`

will be a single combined span. Who cares.

src/functions.js
kohler: I added a Jasmine test, not everything needs a screenshot test. But ok fine whatevs

src/buildHTML.js
kohler: I share this concern, but didn't want in this commit to change the makeSpan interface to force an explicit mclass argument. I have added a TODO to make that happen.

If the class is in the wrong position the consequence will be mis-spacing in complex situations, something that KaTeX already suffered from and that this commit improves. Not great but not a disaster.

If KaTeX had a debug mode one could add debug-mode assertions about math classes or something.

src/functions.js
kohler: \mathit is already defined darling

static/katex.less
kohler: Removing \textsc from the PR anyway

src/functions.js
kevinbarabash: Right you are.  My question should've been why are the other `\textXX` defined as aliases to their `\mathXX` counter parts, but `textit` is not?

static/katex.less
kevinbarabash: I was looking at http://caniuse.com/#feat=font-variant-alternates which appears to be some extension to `font-variant`.

src/domTree.js
kevinbarabash: Sorry for not providing more detail.  It looked some of the code here was being used to compare two styles.  This suggestion was more of a nice-to-have kind of thing and can be done in the future if we actually need to compare styles elsewhere.

src/symbols.js
kevinbarabash: Sounds good.

src/functions.js
kohler: `symbol` is linked with the argument type of `value`. if `symbol` is true, then `value` must be suitable for passing to `makeSymbol`, which isn't true here.

src/functions.js
kevinbarabash: Historically we've asked that PRs for new commands come with screenshot tests.  I'm sure there's lots of functionality that we don't have screenshot tests for.  I feel like we should have more screenshot tests so that we can have more comprehensive coverage.  I'd love to hear your thoughts on testing.

kohler:
I've updated the PR according to the comments above, and removed ¡ ¿ \textsc. Hopefully we're good.

src/buildHTML.js
kevinbarabash: Fair enough.  Having a debug mode would be interesting, but I'd want their to be some way to strip out that code.

src/functions.js
kohler: Testing is great, creating tests is nonzero work, updating tests is nonzero work, running tests takes time (Jasmine tests are much faster). KaTeX’s behavior is not always “right”, I've had to update a lot of screenshotter images because previously-encoded behavior was wrong, and I'm sure as time goes on that will be repeated. But the occasional screenshotter test was valuable to catch a problem (StyleSwitching in particular).

I find tests for edge cases more important than tests for typical cases, so just adding a screenshotter test for, say, “{1 \atop 2}” (which is what I did) is low-value. When a bug is fixed, particularly a tricky one, it's usually important to add a test.

src/functions.js
kevinbarabash: Nevermind... I see why.  `.mathit` sets `font-family: KaTeX_Math;`.  Sorry for all of the silly questions about my own code.  :(

kevinbarabash:
@kohler LGTM.  Thanks again for all of these awesome changes.

kohler:
Thanks! FWIW I think the following issues are closed by this & the previous PR: #563; #541; #534; #530 (sort of: we have \stackrel, not \buildrel); #498; #474; #468; #432; #424; #404; #30.

src/buildCommon.js
kohler: I don't think it will. In the MathDefaultFonts test (that image), italic correction SHOULD be applied for every character.

BTW I have some explanations for questions in #349. KaTeX's \mathit command incorrectly selects the math italic font. The LaTeX \mathit command actually selects the TEXT italic font, which has somewhat narrower letters than true math italic. \mathit means “use text italic in math mode”, it doesn't mean “use math italic.“ Similarly for \mathrm, \mathtt, etc. From fontmath.ltx:

```
\DeclareMathAlphabet      {\mathbf}{OT1}{cmr}{bx}{n}
\DeclareMathAlphabet      {\mathsf}{OT1}{cmss}{m}{n}
\DeclareMathAlphabet      {\mathit}{OT1}{cmr}{m}{it}
\DeclareMathAlphabet      {\mathtt}{OT1}{cmtt}{m}{n}
```

kevinbarabash:
@kohler thanks for listing out all of those issues.  I'll close them tomorrow.

astoilkov:
This is really awesome PR. Do you guys know when we could expect release including this changes.

kevinbarabash:
@astoilkov we're going to try to do one either before the end of the year or earlier in the new year.

astoilkov:
Perfect. Thanks so much for the information @kevinbarabash

PR #578, opened 2016-12-08
Handle \middle.

### LINKED ISSUE #527: Support `\middle`
Initially found in https://github.com/Khan/KaTeX/issues/526

It's pretty clear what this function is supposed to do (the same thing as `\left` and `\right`, but inside of that expression). It doesn't exist in TeX though, and I can't figure out where it's defined in LaTeX, which is terribly confusing.

kohler:
Fixes #527.


xymostech:
[clabot:check]

src/buildHTML.js
```diff
@@ -1231,6 +1247,19 @@ groupTypes.leftright = function(group, options) {
...
 };
 
+groupTypes.middle = function(group, options) {
+    var middleDelim;
+    if (group.value.value === ".") {
+        middleDelim = makeNullDelimiter(options, []);
+    } else {
+        middleDelim = delimiter.sizedDelim(
+            group.value.value, 1, options,
+            group.mode, []);
```
xymostech: You don't actually need to make this delimiter here, right? Shouldn't it always be overridden by the \left\right handler?

test/katex-spec.js
```diff
@@ -1042,6 +1042,16 @@ describe("A left/right parser", function() {
...
         expect(normalEmpty).toParse();
     });
+
+    it("should handle \\middle", function() {
+        var normalMiddle = "\\left( \\dfrac{x}{y} \\middle| \\dfrac{y}{z} \\right)";
+        expect(normalMiddle).toParse();
+    });
+
+    it("should error when \\middle is not in \\left...\\right", function() {
+        var unmatchedMiddle = "(\\middle|\\dfrac{x}{y})";
```
xymostech: Could you test that `\left( { \middle| } \right)` doesn't work?

test/katex-spec.js
```diff
@@ -1042,6 +1042,16 @@ describe("A left/right parser", function() {
         var normalEmpty = "\\Bigl .";
         expect(normalEmpty).toParse();
     });
+
+    it("should handle \\middle", function() {
+        var normalMiddle = "\\left( \\dfrac{x}{y} \\middle| \\dfrac{y}{z} \\right)";
```
xymostech: Could you try putting more than one `\middle` in here? Also, maybe test `\middle` with nested `\left\right`s?

src/buildHTML.js
```diff
@@ -1215,6 +1220,17 @@ groupTypes.leftright = function(group, options) {
     // Add it to the beginning of the expression
     inner.unshift(leftDelim);
 
+    // Handle middle delimiters
+    if (hadMiddle) {
+        for (i = 1; i < inner.length; i++) {
```
xymostech: `i` should start at `0`, no?

kevinbarabash:
@xymostech I'll have a look this evening.

src/buildHTML.js
kohler: Yes, but I'm trying to make it fail more cleanly in case a stray \middle gets in.

test/katex-spec.js
kohler: Ah but it *does* work. I know it shouldn't, but catching this error case would be harder, and I didn't see much point.

test/katex-spec.js
kohler: no problem; i'll leave the screenshotter as is though

test/katex-spec.js
xymostech: Well, the `\middle` won't be rewritten with the correct delimiter if this happens, because it won't be a direct child of the `\left\right` thing, so it probably shouldn't work? (That's actually just a guess, you should test it)

src/buildHTML.js
```diff
@@ -1215,6 +1220,18 @@ groupTypes.leftright = function(group, options) {
...
+            if (inner[i].isMiddle) {
+                // Apply the options that were active when \middle was called
+                inner[i] = delimiter.leftRightDelim(
```
kevinbarabash: The documentation for `leftRightDelim` should be updated, but let's leave this a TODO for now.

src/buildHTML.js
kevinbarabash: Shouldn't a stray `\middle` be caught by the parser?

kohler: There's another comment on this. If LaTeX compatibility were an absolute requirement, then yes, the parser would have to catch every stray `\middle`. But that didn't seem super easy in katex.

kevinbarabash: Are there cases of stray `\middle`s that `leftrightDepth` counting won't catch?

PR #579, opened 2016-12-08
Allow unbraced kerns, such as \kern1em.

kohler:
This is actually the *only* syntax TeX allows; braced kern units
are invalid.

### DIFF
--- a/src/Parser.js
+++ b/src/Parser.js
@@ -662,6 +662,35 @@ Parser.prototype.parseStringGroup = function(modeName, optional) {
 };
 
 /**
+ * Parses a regex-delimited group: the largest sequence of tokens
+ * whose concatenated strings match `regex`. Returns the string
+ * formed by the tokens plus some position information.
+ *
+ * @param {RegExp} regex
+ * @param {string} modeName  Used to describe the mode in error messages
+ */
+Parser.prototype.parseRegexGroup = function(regex, modeName) {
+    var outerMode = this.mode;
+    this.mode = "text";
+    var firstToken = this.nextToken;
+    var lastToken = firstToken;
+    var str = "";
+    while (this.nextToken.text !== "EOF"
+           && regex.test(str + this.nextToken.text)) {
+        lastToken = this.nextToken;
+        str += lastToken.text;
+        this.consume();
+    }
+    if (str === "") {
+        throw new ParseError(
+            "Invalid " + modeName + ": '" + firstToken.text + "'",
+            firstToken);
+    }
+    this.mode = outerMode;
+    return firstToken.range(lastToken, str);
+};
+
+/**
  * Parses a color description.
  */
 Parser.prototype.parseColorGroup = function(optional) {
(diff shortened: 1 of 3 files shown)


src/Parser.js
```diff
@@ -682,11 +711,17 @@ Parser.prototype.parseColorGroup = function(optional) {
  * Parses a size specification, consisting of magnitude and unit.
  */
 Parser.prototype.parseSizeGroup = function(optional) {
-    var res = this.parseStringGroup("size", optional);
+    var res;
+    if (!optional && this.nextToken.text !== "{") {
+        res = this.parseRegexGroup(
+            /^[-+]? *(?:$|\d+|\d+\.\d*|\.\d*) *[a-z]{0,2}$/, "size");
```
csilvers: Does TeX allow tabs (or other whitespace) here?  (I admit it's unlikely to matter in practice.)

csilvers: Also,** `.` by itself is a valid size?  (`\\kern.px`)  If so, may be interesting to add a test-case for that.

src/Parser.js
```diff
@@ -662,6 +662,35 @@ Parser.prototype.parseStringGroup = function(modeName, optional) {
...
+        lastToken = this.nextToken;
+        str += lastToken.text;
+        this.consume();
+    }
+    if (str === "") {
+        throw new ParseError(
+            "Invalid " + modeName + ": '" + firstToken.text + "'",
+            firstToken);
+    }
+    this.mode = outerMode;
```
csilvers: I know it's unlikely anyone is going to try to look at `this` again after you throw the `ParseError`, but seems like it's good hygenie to reset `this.mode` before you do the `str === ""` test (since I don't think you need the mocked mode anymore after the while loop).

kohler: `parseStringGroup` and many other preexisting katex functions dont do this

src/Parser.js
kohler: no it's not, this regex must match any PREFIX of a valid size, the next regex down limits it to valid sizes only

kohler: just matching the earlier katex code's space treatment here

kevinbarabash: I opened https://github.com/Khan/KaTeX/issues/583 to look at all of regexes.

src/Parser.js
kevinbarabash: We don't catch any of the errors we throw.  @csilvers this only matters if we were catching errors and attempting to recover in someway?  In that cause it would also be acceptable to reset the `this.mode` inside the `if` block, right?

test/katex-spec.js
```diff
@@ -999,6 +999,31 @@ describe("A kern parser", function() {
...
+
+    it("should not parse invalid units", function() {
+        expect(badUnitRule).toNotParse();
+        expect(noNumberRule).toNotParse();
+    });
+
+    it("should parse negative sizes", function() {
+        var parse = getParsed("\\kern-1em")[0];
+        expect(parse.value.dimension.number).toBeCloseTo(-1);
+    });
```
kevinbarabash: Can you add a test case for parsing `"\\kern+1em"`?

kevinbarabash:
@kohler thanks for adding support for this syntax.  I think it'll come it really handle once we get better macro support.

src/Parser.js
gagern: In my mind that the whole operation of the Parser is kind of atomic. Either the parsing works and returns a result, or it does not and throws an exception. In neither of these cases is the final state of the Parser object of any relevance, and I'd keep it that way.

PR #580, opened 2016-12-08
Support "mu"-denominated sizes and \mkern

kohler:


src/functions.js
```diff
@@ -205,7 +205,7 @@ defineFunction("\\rule", {
     };
 });
 
-defineFunction("\\kern", {
+defineFunction(["\\kern", "\\mkern"], {
```
kevinbarabash: According to TeXbook on pg. 168, `\mkern` only accepts measurements given in `mu` units and `\kern` cannot accept `mu` units.  Can you add a comment describing this deviation from spec?  We'll probably want to enforce these constraints at some point, but it doesn't seem critical at the moment.

src/buildHTML.js
```diff
@@ -1239,21 +1238,11 @@ groupTypes.rule = function(group, options) {
...
-    if (group.value.width.unit === "ex") {
-        width *= style.metrics.xHeight;
-    }
-
-    var height = group.value.height.number;
-    if (group.value.height.unit === "ex") {
-        height *= style.metrics.xHeight;
-    }
+    var width = calculateSize(group.value.width, style);
+    var height = calculateSize(group.value.height, style);
```
kevinbarabash: Love the cleanup.

kohler:
Comments addressed

kevinbarabash:
@kohler thanks for the PR.

PR #581, opened 2016-12-08
BUG FIX: Correct spacing next to symbols when styles switch.

kohler:
Two nodes were sharing a list of classes, which led to the style
change applying multiple times. Fix that.

(This is my bug, from the original spacing PR.)


kohler:
Cc @kevinbarabash

src/buildHTML.js
```diff
@@ -83,7 +83,7 @@ var buildExpression = function(expression, options, isRealGroup) {
             i--;
         } else if (spaces) {
             if (groups[i] instanceof domTree.symbolNode) {
-                groups[i] = makeSpan(groups[i].classes, [groups[i]]);
+                groups[i] = makeSpan([].concat(groups[i].classes), [groups[i]]);
```
kevinbarabash: How was `groups[i].classes` getting mutated?

kohler: Should this be written as `groups[i].classes.slice()` instead?

kevinbarabash:
@kohler thanks for noticing the issue and fixing it so promptly.

Issue #393, opened 2015-11-19
Support for \mod, \bmod, \pmod, and \pod


jschaf:
Got a parse error using `\bmod`.

```
ParseError: KaTeX parse error: Expected 'EOF', got '\bmod' at position 7: k \bmod̲ 3
```

Here's some detail on the different modulus operators in LaTeX.  http://www.matthewflickinger.com/blog/archives/2005/02/20/latex_mod_spacing.asp

kevinbarabash:
It shouldn't be too hard to implement these.  I'm guessing `\mod` and `\bmod` are more commonly used.

maryfries:
:+1:

chmln:
Hate to +1, but lack of modulus is a significant drawback

gagern:
These macros are defined in amsmath.sty as follows:

```tex
\renewcommand{\bmod}{\nonscript\mskip-\medmuskip\mkern5mu\mathbin
  {\operator@font mod}\penalty900
  \mkern5mu\nonscript\mskip-\medmuskip}
\newcommand{\pod}[1]{\allowbreak
  \if@display\mkern18mu\else\mkern8mu\fi(#1)}
\renewcommand{\pmod}[1]{\pod{{\operator@font mod}\mkern6mu#1}}
\newcommand{\mod}[1]{\allowbreak\if@display\mkern18mu
  \else\mkern12mu\fi{\operator@font mod}\,\,#1}
```

So a `\bmod` always has 5/18em before and after the operator name, without the extra 4/18em you'd [usually get](https://github.com/Khan/KaTeX/blob/v0.6.0/static/katex.less#L108) from a binary operation in non-script style. I guess that if the `\bmod` gets turned into an atom other than Bin, the negative skip might actually reduce the surrounding space. The `\penalty` is irrelevant as we don't wrap, as is the `\allowbreak`. A `\pod` has 1em leading space in display style, or 8/18em otherwise. A `\mod` has 1em resp. 12/18 em before and 6/18em after.

I guess we could turn all those offsets into `\kern` or the corresponding internal representation.

kevinbarabash:
It would be nice if we could parse `mu` units and support `mkern` and all of the `muskips`.  If we hardcode values for these it means more work later if we want to support redefining `medmuskip` etc.

gagern:
[...]

PR #582, opened 2016-12-08
Support \bmod, \pmod, \pod, \mod.

kohler:
Not super clean but it works. The spaces, etc. were translated from the LaTeX code in #393.

This PR includes the commit in PR #581, because some of the screenshotter images depend on that fix. OH WELL


src/buildHTML.js
```diff
@@ -916,6 +916,57 @@ groupTypes.op = function(group, options) {
...
+            buildCommon.mathsym("m", group.mode),
+            buildCommon.mathsym("o", group.mode),
+            buildCommon.mathsym("d", group.mode)];
+        if (group.value.modType === "bmod") {
+            inner.push(makeSpan(["mbin"], modInner, options));
+            inner.push(makeSpan(["mspace", "thickspace"], [], options));
+            if (!options.style.isTight()) {
+                inner.push(makeSpan(
+                    ["mspace", "negativemediumspace"], [], options));
+            }
```
kevinbarabash: It took me a while to realize that this `if` statement corresponds to `\nonscript\mskip-\medmuskip`.

src/buildHTML.js
```diff
@@ -916,6 +916,57 @@ groupTypes.op = function(group, options) {
...
+
+    if (group.value.modType === "pod" || group.value.modType === "pmod") {
+        inner.push(buildCommon.mathsym("(", group.mode));
+    }
+
+    if (group.value.modType !== "pod") {
+        var modInner = [
+            buildCommon.mathsym("m", group.mode),
+            buildCommon.mathsym("o", group.mode),
+            buildCommon.mathsym("d", group.mode)];
```
kevinbarabash: It would be nice to have a function that builds these kinds of strings.  Not necessary for this PR.

src/buildHTML.js
```diff
@@ -916,6 +916,57 @@ groupTypes.op = function(group, options) {
...
+                ["mspace", "negativemediumspace"], [], options));
+        }
+        inner.push(makeSpan(["mspace", "thickspace"], [], options));
+    } else if (options.style.size === Style.DISPLAY.size) {
+        inner.push(makeSpan(["mspace", "quad"], [], options));
+    } else if (group.value.modType === "mod") {
+        inner.push(makeSpan(["mspace", "twothirdsemspace"], [], options));
+    } else {
+        inner.push(makeSpan(["mspace", "eightmuspace"], [], options));
+    }
```
kevinbarabash: Logic looks good.  It'll be nice when we can write these as macros, but this is good for now.

src/buildHTML.js
```diff
@@ -916,6 +916,57 @@ groupTypes.op = function(group, options) {
...
+            buildCommon.mathsym("d", group.mode)];
+        if (group.value.modType === "bmod") {
+            inner.push(makeSpan(["mbin"], modInner, options));
+            inner.push(makeSpan(["mspace", "thickspace"], [], options));
+            if (!options.style.isTight()) {
+                inner.push(makeSpan(
+                    ["mspace", "negativemediumspace"], [], options));
+            }
+        } else {
+            Array.prototype.push.apply(inner, modInner);
```
kevinbarabash: We're going to be enabling support for some ES6 constructs including the spread operator after the next release so things like this will be easier.

src/buildMathML.js
```diff
@@ -341,6 +341,31 @@ groupTypes.op = function(group, options) {
...
+groupTypes.mod = function(group, options) {
+    var inner = [];
+
+    if (group.value.modType === "pod" || group.value.modType === "pmod") {
+        inner.push(new mathMLTree.MathNode(
+            "mo", [makeText("(", group.mode)]));
+    }
+    if (group.value.modType !== "pod") {
+        inner.push(new mathMLTree.MathNode(
+            "mo", [makeText("mod", group.mode)]));
```
kevinbarabash: Where does the space before the `mod` get added?

static/katex.less
```diff
@@ -308,10 +312,22 @@
             width: @thickspace;
         }
 
+        &.onethirdemspace {
+            width: 0.333333em;
+        }
```
kevinbarabash: Maybe `.sixmuspace` would be better since amsmath.sty uses `6mu`.

static/katex.less
```diff
@@ -308,10 +312,22 @@
...
+            width: 0.444444em;
+        }
+
         &.enspace {
             width: 0.5em;
         }
 
+        &.twothirdsemspace {
+            width: 0.666667em;
+        }
```
kevinbarabash: Maybe `.twelvemu` for similar reasons.

kevinbarabash:
> This PR includes the commit in PR #581, because some of the screenshotter images depend on that fix.

"merge and squash" should clean that up nicely.

src/buildHTML.js
kohler: Will add comment

src/buildMathML.js
kohler: If it's important that mathml have the same space widths as html I suppose I should repeat the logic from buildhtml here. I don't know, though, what the mathml is really for.

kevinbarabash: After looking at http://caniuse.com/#feat=mathml it's probably safe to say that it's not worth supporting due to spotty browser support.  Originally we wanted to use the MathML rendering for a11y reasons so the browser could read the math out loud.  That didn't work that well so we ended up converting the KaTeX parse tree to a string which the browser could read.

kohler: OK, then I’ll leave this alone for now. The revised commit has all the other requested changes.

PR #586, opened 2016-12-16
Misspelled variable "candiate"

tabatkins:
All uses have the same misspelling, so this wasn't a functionality bug, but still better to have things spelled correctly in the source. ^_^


tabatkins:
[clabot:check]

kevinbarabash:
@tabatkins thanks!

PR #590, opened 2016-12-25
README.md - added integrity

mateddy:
Integrity generated from https://www.srihash.org/
Reference: https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity


mateddy:
[clabot:check]

PR #598, opened 2017-01-03
Load full-window page before adjusting browser window size

gagern:
Experimenting with selenium-webdriver@3.0.1 and Firefox 50.1.0 I observed screenshots having a height of merely 8 pixels.  Presumably the margin or padding of an otherwise empty document.  So in order to get the actual size of the document area, the screenshotter now loads a document which fills the entire viewport.

This might be useful in the future, and perhaps might help with the hung processes described in #494 which were never fully understood.


kevinbarabash:
LGTM.  Thanks!

Issue #601, opened 2017-01-05
Warn when using an XHTML DOCTYPE


oddhack:
I see several other open issues about fractions, but I think this is different, and so badly rendered that I'm wondering if I'm doing something obviously bogus (first-time KaTeX user here).

There's a test case with a couple of fractions at http://alumnus.alumni.caltech.edu/~jon/frac.html . This uses KaTeX 0.6.0 from the CDN, which I believe to be current.

And I've attached an image showing rendering in Chromium 55 / Debian 8 (same results with current Firefox). I've tried scaling up and down between about 50% - 200% and it looks just as bad in all cases.
![frac](https://cloud.githubusercontent.com/assets/4693344/21681105/87e8aa54-d302-11e6-9faa-18a854108487.gif)

oddhack:
N.b. here's the math I'm trying to render, extracted from the HTML for future reference: 

\[ half = \frac{1}{2} \]
\[ f = \frac{c}{{2^b - 1}} \]
\[ max_{clamped} \over {2^{(exp'-B-N)}} + \frac{1}{2} \]

kevinbarabash:
@oddhack I noticed you're using an XHTML doctype, try `<!DOCTYPE html>` and see if that fixes it.

edemaine:
@kevinbarabash In local testing, that was indeed the problem.  I see that KaTeX has a warning for quirks mode; maybe it can also have a warning for XHTML mode?  (And maybe someday XHTML support... if that is important.)

kevinbarabash:
@edemaine thanks for testing that out.

xymostech:
@oddhack I did a bunch of digging into what is causing this problem. It like there's a difference in rendering between XHTML and HTML 5, maybe due to the limited-quirks mode (which is like quirks but with fewer quirks, and which is turned on in XHTML 1.0): https://quirks.spec.whatwg.org/#the-line-height-calculation-quirk

I made a minimal (I think) repro case for what's going on here: http://jsbin.com/fihiboqopu/1/edit?html,css,output If you replace the doctype with the XHTML one, you'll see the "2" inside of the box shift vertically a little bit. This is a reproduction of the HTML that KaTeX uses to vertically position things. Because of that difference, fractions render poorly when using XHTML. This same thing happens in quirks mode, (i.e. if you remove the doctype completely the same thing happens), which leads me to believe it has something to do with limited-quirks mode.

[...]

PR #602, opened 2017-01-05
Warn and disable rendering when in XHTML doctype.

edemaine:
Fix for #601, assuming we also want to disable rendering in XHTML mode (just like quirks mode).  Tested on http://alumnus.alumni.caltech.edu/~jon/frac.html and on normal `<!DOCTYPE html>`.

### DIFF
--- a/katex.js
+++ b/katex.js
@@ -29,8 +29,8 @@ var render = function(expression, baseNode, options) {
     baseNode.appendChild(node);
 };
 
-// KaTeX's styles don't work properly in quirks mode. Print out an error, and
-// disable rendering.
+// KaTeX's styles don't work properly in quirks mode or in XHTML mode.
+// Print out an error, and disable rendering.
 if (typeof document !== "undefined") {
     if (document.compatMode !== "CSS1Compat") {
         typeof console !== "undefined" && console.warn(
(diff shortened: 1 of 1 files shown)


oddhack:
I don't have any understanding of the technical issues behind not supporting rendering in XHTML mode, but will note that our use case is an asciidoc->Docbook->XHTML path, so basically this change would prevent use of katex in our specific situation until we could revamp the pipeline to target the html5 backend, which has been held up for other reasons (we are kinda hoping to migrate from mathjax and are evaluating the needed changes).

gagern:
As discussed in https://github.com/Khan/KaTeX/issues/601#issuecomment-270886445, XHTML isn't the problem, but limited-quirks mode is. According to [the WHATWG HTML LS](https://html.spec.whatwg.org/multipage/syntax.html#the-initial-insertion-mode), limited-quirks mode is selected if

> * The public identifier starts with: "`-//W3C//DTD XHTML 1.0 Frameset//`"
> * The public identifier starts with: "`-//W3C//DTD XHTML 1.0 Transitional//`"
> * The system identifier is not missing and the public identifier starts with: "`-//W3C//DTD HTML 4.01 Frameset//`"
> * The system identifier is not missing and the public identifier starts with: "`-//W3C//DTD HTML 4.01 Transitional//`"

So we might want to reproduce these checks, unless someone has a better idea of how to detect limited-quirks mode. But as I see some real chance that we'll be able to fix the errors in limited-quirks mode pretty soon, you might want to wait and see whether that is indeed the case.

edemaine:
Indeed, fixing the root issue seems preferable!  Let me know if you'd like this PR revised anyway.

gagern:
I'd say this can be closed if #607 gets accepted.

kevinbarabash:
https://github.com/Khan/KaTeX/pull/608 handles quirks mode as well as limited quirks mode which I believe makes this PR moot.  Please re-open if that is actually not the case.

Issue #249, opened 2015-06-17
spacing around fractions is incorrect


kevinbarabash:
The image diff below shows KaTeX rendering in black and LaTeX rendering in white.  Notice that spacing around the fractions and solidus don't match up.
![b6ed6322-14fd-11e5-843f-cc2814e0d921](https://cloud.githubusercontent.com/assets/1044413/8215794/bf07b372-14ef-11e5-8bc4-0b54c788694f.png)

I'll post a minimal repro case later.

kevinb7:
Credit goes to @gagern for finding this issue.  I really like idea of these images diffs to compare KaTeX output to LaTeX.

gagern:
Using the image diff format from #268, the diff now looks like this (green is LaTeX, red is KaTeX):

![arrays-diff](https://cloud.githubusercontent.com/assets/190759/8351742/d7bec252-1b2f-11e5-8f3d-5648b90960c8.png)

The difference in the `NestedFractions` test is probably due to the same problem:

![nestedfractions-diff](https://cloud.githubusercontent.com/assets/190759/8351764/f92d9472-1b2f-11e5-9382-24381a409e25.png)

kevinb7:
I noticed that the our fraction bars aren't wide enough in some cases when I was working on ```\cfrac```, see https://github.com/Khan/KaTeX/pull/135.

gagern:
OK, I did some research on this. TeX puts 1.2pt of space before and after every fraction. This is the value of `\nulldelimiterspace`. So the branches of the `genfrac` handler where it adds delimiters should add such null delimiters if no actual delimiters are given. The `leftright` handler already adds null delimiters, so we can duplicate functionality from there.

I also noticed that when I apply `\showlists` to some fraction, each fraction is contained in a `\mathord`, and is not simply of type Inner. This is probably due to the fact that the definition of `\frac` reads

```tex
\DeclareRobustCommand{\frac}[2]{{\begingroup#1\endgroup\@@over#2}}
```

which has two sets of braces, so it forms a brace-delimited group which gets translated into a `\mathord`.

Both the `\nulldelimiterspace` and the mechanism turning the Inner into an Ord are discussed in https://github.com/mathjax/MathJax/issues/969, with valuable explanations from David Carlisle.

ylemkimon:
Closing as #295 is merged.

PR #606, opened 2017-01-06
Change cell style to 'text' in {array}, {matrix}, {cases}. Add {darray} and {dcases}.

### LINKED ISSUE #603: Different lines in {cases} environment collide
There's an apparent rendering bug in this code:

\\[ \begin{aligned}
d =
  \begin{cases}
    level_{base},     & \lambda \leq \frac{1}{2} \\
    nearest(\lambda), & \lambda > \frac{1}{2},
                        level_{base} + \lambda \leq
                        q + \frac{1}{2} \\
    q,                & \lambda > \frac{1}{2}, level_{base} + \lambda > q + \frac{1}{2}
  \end{cases}
\end{aligned} \\]

where the 1/2 fractions bump into the content of the cases above and below them. This doesn't happen when rendered by LaTeX. I did poke through the other issues first and it's not obvious this is a duplicate, though it may be. Seeing this in both Firefox and Chromium.

eq41

edemaine:
Fixes #603.  LaTeX's `{array}` environment switches to text mode and wraps all cells within `$...$`, effectively switching to text mode.  AMSMath's `{matrix}` and `{cases}` build on this, so inherit the same behavior.  I added `{darray}` and `{dcases}` which avoid this behavio (setting cells in display mode).  There doesn't seem to be a `{dmatrix}` environment in any LaTeX package, though it would be easy to add one.

I'm a little unclear on whether all cells should be wrapped in an `ordgroup` at the top level, or if this was just a way to get multiple parsed objects into a single `ParseNode`.  I've assumed the latter, but it's easy to tweak to wrap the `styling` node in an `ordgroup`, or vice versa.


gagern:
Thanks a lot for the research and pull request! Code looks good to me. Travis is unhappy about the screenshots. Can you recreate those, or do you want me to do that?

edemaine:
I'd be happy to do it, but you'd have to teach me how. :-)

gagern:
Start by reading https://github.com/Khan/KaTeX/tree/master/dockers/Screenshotter and let me know if you get stuck.

edemaine:
Thanks @gagern!  The testing was quite revealing indeed: it seems that it really is important to have an `ordgroup` containing the cell, and then wrap a styling around that; otherwise, $+1$ renders like ${}+1$ (with too much space) inside an array cell.  Not quite sure why exactly, but the fix was easy enough, and the result is that all screenshots are exactly as they were.  (Relatively few tests run in display mode, so none of them were encountering the wrong behavior of displaystyle within an array.)

kevinbarabash:
@edemaine could you add screenshot tests where your changes affect the output?

edemaine:
@kevinbarabash Added.  I also tested against the tex renderer, but there is still some spacing discrepancy (green is LaTeX).  I don't think this is related to my patch, but should test more...

arraymode

edemaine:
Confirmed: the tight formatting is consistent with master branch.  The test doesn't lie within 1024x768 in master, because of the bug being fixed (`\frac`s render in display mode), but the screenshot still drives the point home:

arraymode-chrome1

kevinbarabash:
@edemaine it's hard for me to tell the difference between master and this.  It looks like there's a bit more space.  How does the layout from https://github.com/Khan/KaTeX/issues/603 look?

edemaine:
Here is the example from #603, first rendered by this PR:

cases2-chrome

For comparison, here is the rendering of the same example with display mode turned off (from YAML):

cases1-chrome

These look identical to me except for a top vertical space, presumably caused by the difference between inline and display math.  I confirmed this by overlaying the images in Photoshop.

gagern:
Very interesting test case, since it has so many differences from LaTeX! If we can get all of them right we should be doing a good lot better in some other situations, too. Could it be our `\vdots` has incorrect vertical dimensions? The horizontal spacing might be in part related to #249, but we also have wider spacing between the `x_1` and the next `\partial`, so there might be a wrong class somewhere there, or the CSS styles don't behave as we expect them to. Lots of work…

edemaine:
@kevinbarabash Let me know if you'd like any more changes to this branch.  I'm pretty sure nothing has changed from master other than the intended display -> text style change in cells.  In particular, all previous screenshot texts (which include some text-style {array} and {cases}) are rendered exactly the same as master, even though I'm still putting in a styling node that converts every cell to text style.  (Though now that I mention this, perhaps I should avoid the extra styling nodes in this case?  LaTeX "has" them, but needlessly inefficient here I guess.)

edemaine:
FYI, it seems the `context` doesn't keep track of whether we're in display style or text style (only math vs. text mode), so the array environments can't detect whether it's necessary to explicitly switch each cell to text or display style.  This is probably fine (as I said, I think it's effectively the same as what LaTeX does), but if you want more efficiency in the parse tree, I could look into adding the current style to `context`...

I'd like to add `*{10}` from #269, but I should wait for this to be merged into master.  FYI, I just resolved conflicts by updating to ES2016.

gagern:
Looks good to me. Merged. Thanks a lot, @edemaine, for your initiative in fixing the problem, and for your patience regarding how much tweaking it took to get this right.

kevinbarabash:
@edemaine sorry for not replying sooner.  I tried to reproduce the behavior seen in https://github.com/Khan/KaTeX/issues/603 by checking out the previous commit I wasn't able to repro the issue using `\cases`.  Rendering the example from 603 using `\dcases` however shows the same collision between lines.

PR #607, opened 2017-01-07
Fix positioning in limited quirks mode

### LINKED ISSUE #601: Warn when using an XHTML DOCTYPE

gagern:
This fixes #601 by adding a zero-width space to every span that contains child nodes.  The symbol avoids the “[blocks ignore line-height quirk](https://quirks.spec.whatwg.org/#the-blocks-ignore-line-height-quirk)”.

All screenshot tests render the same before and after this commit, and they also render the same if the doctype is set to HTML 4.01 transitional in order to [trigger](https://html.spec.whatwg.org/multipage/syntax.html#the-initial-insertion-mode) [limited quirks mode](https://dom.spec.whatwg.org/#concept-document-limited-quirks).

### DIFF
--- a/src/domTree.js
+++ b/src/domTree.js
@@ -88,6 +88,9 @@ span.prototype.toNode = function() {
     for (var i = 0; i < this.children.length; i++) {
         span.appendChild(this.children[i].toNode());
     }
+    if (this.children.length) {
+        span.appendChild(document.createTextNode("\u200b"));
+    }
 
     return span;
 };
@@ -132,6 +135,9 @@ span.prototype.toMarkup = function() {
     // Add the markup of the children, also as markup
     for (var i = 0; i < this.children.length; i++) {
         markup += this.children[i].toMarkup();
+    }
+    if (this.children.length) {
+        markup += "&#x200b;";
     }
 
     markup += "</span>";


xymostech:
Ack! I was making a commit for this, you beat me to it :) Overall though, this seems very heavy handed. Adding zero-width spaces to almost every single span in the tree is going to be a lot of zero-width spaces. That might not be a terribly big deal when rendering normally, but I imagine that would inflate the server-rendered size a bunch.

I have a PR I'm about to send that should fix it in just the places that need it. I pushed it here, if you wanna take a look: https://github.com/Khan/KaTeX/commit/f7d9569547bae02f625ddf50e54b20241592d672

gagern:
Yes, I'm kind of worried, too. One thing worth considering is that we adapt #602 to detect limited quirks mode from the doctype and only add the symbols if required. Of course that won't work for server-side rendering. There we could use a flag, hoping that people manage to set the flag if they don't manage to choose an up-to-date doctype.

xymostech:
(I made the PR at #608)

edemaine:
I like the idea of a `limitedQuirksMode` option, whose default tries to detect current browser behavior (and `false` if server-side).

kevinbarabash:
Closing because I believe this PR is superseded by https://github.com/Khan/KaTeX/pull/608.  Please re-open if that is not actually the case.

PR #608, opened 2017-01-07
Make KaTeX work in Quirks mode

xymostech:
Summary: In issue #601, it was noticed that the KaTeX bug with the fraction
bars overlapping the text was occuring with an XHTML doctype. This indicated
that the bug we were seeing was caused by both quirks mode and limited-quirks
mode (which is a version of quirks mode with fewer quirks and is enabled for
various doctypes including some XHTML ones).

Based on the [quirks spec](https://quirks.spec.whatwg.org/), it appears that
there are only two quirks in limited-quirks mode, both having to do with a
line-height calculation. @gagern figured out that if we added some zero-width
spaces in our elements, we would stop triggering the quirk, which would make
our fractions render correctly in limited-quirks mode.

I implemented that change, and ran the screenshotter in limited-quirks mode.
There were several other places that suffered from the same quirk, but were
also easily fixed via adding zero-width spaces. I then ran the screenshotter in
quirks mode, and discovered that (once an appropriate meta charset was added),
everything looked correct still.

So, this diff fixes all of the places that the limited-quirks mode quirks
affect our rendering, and removes the warning about rendering in quirks mode.

I also added support to our screenshotter to render things in both no-quirks
and quirks mode, to ensure that things don't break in the future. I copied the
non-quirks images to the quirks images, and ran the screenshotter with
`--verify` to make sure that they look the same.

I have some thoughts that I'd like to hear opinions about:
 - I'm not super happy with how the screenshot tests work. Ideally we'd test
   both quirks mode and non-quirks mode against the same images, since we'd
   like them to be the same. I'm not sure how to make that work well, though,
   since then people wouldn't be able to tell if it's a quirks-mode problem or
   not.
 - I removed the doctype in the testing page file, so all testing would now be
   done in quirks-mode. Not sure if we really want that.
 - I need to test this in IE, but it looks like the trailing commas change we
   made with eslinting is causing problems (cause IE doesn't like trailing
   commas).

Test Plan:
 - `./dockers/Screenshotter/screenshotter.sh --verify`


gagern:
I was thinking about a screenshotter flag to switch doctype which would take a list of types for `--verify` mode, running each of them against the stored files. That way we could check all modes or generate new screenshots in any mode. Will look into this and push to your branch, if you don't mind.

xymostech:
Okay, I did a bit of testing on IE 9 and found that quirks mode is awful. So I'm gonna modify this to go back to warning on quirks mode and to just testing limited-quirks mode everywhere.

I'll do that later, though. I'm gonna go make the release...

gagern:
@kevinbarabash, since this PR here now contains a commit by Emily and one by me, perhaps you should be the one to review it.

dockers/Screenshotter/README.md
```diff
@@ -21,7 +21,7 @@ docker containers with the selenium setups running.  Essentially you
 are encouraged to reproduce the steps from `screenshotter.sh`
 manually.  Example run for Firefox:
 
-    container=$(docker run -d -P selenium/standalone-firefox:2.46.0)
+    container=$(docker run -d -P selenium/standalone-firefox:2.48.2)
```
kevinbarabash: Why the version bump?

dockers/Screenshotter/screenshotter.js
```diff
@@ -72,6 +72,14 @@ var opts = require("nomnom")
     .option("wait", {
         help: "Wait this many seconds between page load and screenshot",
     })
+    .option("mode", {
+        choices: ["no-quirks", "limited-quirks", "quirks", "all"],
+        list: true,
+        help: "Render mode, determined by doctype, " +
+            "may be given multiple times, " +
+            "values [no-quirks|limited-quirks|quirks|all], " +
+            "defaults to all for --verify or no-quirks otherwise",
```
kevinbarabash: That sounds reasonable.

dockers/Screenshotter/screenshotter.js
```diff
@@ -393,7 +440,8 @@ function takeScreenshot(key) {
              * output file name for one of these cases, we accept both.
              */
             key += "_alt";
-            file = path.join(dstDir, key + "-" + opts.browser + ".png");
+            basename = key + basename.slice(key.length - 4);
```
kevinbarabash: I don't understand why we're slicing basename in this way.

dockers/Screenshotter/screenshotter.js
```diff
@@ -343,16 +362,39 @@ function findHostIP() {
...
+    "limited-quirks": '<!DOCTYPE HTML PUBLIC ' +
+        '"-//W3C//DTD HTML 4.01 Transitional//EN" ' +
+        '"http://www.w3.org/TR/html4/loose.dtd">',
+    "quirks": '<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 3.2 Draft//">',
+};
+
+var qabbr = {
+    "no-quirks": "nq",
+    "limited-quirks": "lq",
+    "quirks": " q",
```
kevinbarabash: The extra whitespace looks like it shouldn't be there, maybe `"qm"` or `"qu"` and add a comment that these abbreviations need to be two characters.

server.js
```diff
@@ -50,6 +50,23 @@ app.use("/test/jasmine",
     )
 );
 app.get("/test/katex-spec.js", serveBrowserified("test/*[Ss]pec.js"));
+// Serve the test html file but with the doctype removed to trigger quirks
+// mode.
+app.get("/test/screenshotter/test-quirks.html", function(req, res, next) {
```
kevinbarabash: This URL doesn't seem to be used by screenshotter.js, is that intentional?

server.js
```diff
@@ -50,6 +50,23 @@ app.use("/test/jasmine",
...
+    fs.readFile(testFile, {encoding: "utf8"}, function(err, data) {
+        if (err) {
+            next(err);
+            return;
+        }
+
+        res.setHeader("Content-Type", "text/html");
+        const lines = data.split('\n');
+        // We remove the first line of the file, which contains the doctype.
+        lines.splice(0, 1);
```
kevinbarabash: The doctype in test/screenshotter/test.html was removed by this diff, is this still necessary?

dockers/Screenshotter/README.md
gagern: I wondered about that at first, too. But it turns out that's what the `screenshotter.sh` script is using. So this updates the README to match the automated generation. More consistent that way.

dockers/Screenshotter/screenshotter.js
gagern: We turn `"Lap-firefox-nq"` into `"Lap_alt-firefox-nq"`. We know that the string starts with `"Lap-"` at this point, and that's what we remove. Would you prefer I change this to `basename = basename.replace("Lap", "Lap_alt")` for better maintainability?

dockers/Screenshotter/screenshotter.js
gagern: I (a) wanted them all the same width so that the test names align on output and (b) have the abbreviation letters match the words of the option. Can add a comment to that effect.

server.js
gagern: Indeed, that's something left over from how @xymostech approached this. Should be removed I think.

dockers/Screenshotter/screenshotter.js
gagern: Actually I think I my thought was that we might have to handle other screenshots in a similar way one day. So `key.length - 4` is the length of the key before I added the `"_alt"` to it. The slice is everything after that, i.e. all the things that got appended to the original key. I'll try an alternate formulation in my next commit.

gagern:
I'm not sure how a GitHub squash-merge would attribute the different commits. Should we prefer merges in such situations? It's not that I consider my code changes in this specific case to be so important they have to be attributed to me. More of a general question.

kevinbarabash:
@xymostech it looks like the unicode tests are failing reliably.  Do those tests fail locally for you?

edemaine:
Hmm, seems we lost track of this.  Needs some rebasing/conflict resolution and possibly fixing of Unicode screenshot tests?

gagern:
I see no conflicts, merges cleanly with master. The Unicode test is a different matter. I can reproduce this locally, and it is a change for the worse. Chrome doesn't render at all, and Firefox uses placeholder boxes for the second half. Needs investigation as to why the fonts aren't available to the tests any more.

gagern:
OK, Unicode fonts made accessible again. Travis tests against a merge into head, so that works, too. I'd say this should be good to go now.

kevinbarabash:
@gagern @xymostech sorry for not merging this sooner.  I'm going to wait until https://github.com/Khan/KaTeX/pull/679 is merged since some rebasing is required anyway.

kevinbarabash:
#679 has been merge so this is no longer blocked.  There are some conflicts that need resolving before this can be landed though.

kevinbarabash:
> I'm not sure how a GitHub squash-merge would attribute the different commits. Should we prefer merges in such situations? It's not that I consider my code changes in this specific case to be so important they have to be attributed to me. More of a general question.

@gagern I'm not sure how the squash merge will work.  I'm guessing that github will say xymostech committed with gagers and kevinbarabash (if I end up doing the merge).

dockers/Screenshotter/screenshotter.js
```diff
@@ -379,22 +424,28 @@ function takeScreenshot(key) {
...
     driver.call(loadMath);
 
     function loadMath() {
         if (!opts.reload && driverReady) {
             driver.executeAsyncScript(
                     "var callback = arguments[arguments.length - 1]; " +
                     "handle_search_string(" +
-                    JSON.stringify("?" + itm.query) + ", callback);")
+                    JSON.stringify(
+                        "?mode" + mode + "&" + itm.query) + ", callback);")
```
kevinbarabash: @kohler I assume we want both the `mode` and `itm.query` in the query string here, is that right?

dockers/Screenshotter/screenshotter.js
```diff
@@ -358,16 +377,42 @@ function findHostIP() {
...
+
 function takeScreenshots() {
-    listOfCases.forEach(takeScreenshot);
+    const html = fs.readFileSync(require.resolve(
+        "../../test/screenshotter/test.html"));
+    function handler(req, res, next) {
+        res.send(doctypes[req.query.mode] + "\n" + html);
+    }
+    app.get("/ss-render.html", handler);
+    app.get("/babel/ss-render.html", handler);
```
kevinbarabash: We probably shouldn't be calling this anymore since server.js isn't serving `babel` versions of URLs anymore.

katex.js
```diff
@@ -29,20 +29,6 @@ let render = function(expression, baseNode, options) {
     baseNode.appendChild(node);
 };
 
-// KaTeX's styles don't work properly in quirks mode. Print out an error, and
-// disable rendering.
-if (typeof document !== "undefined") {
-    if (document.compatMode !== "CSS1Compat") {
-        typeof console !== "undefined" && console.warn(
-            "Warning: KaTeX doesn't work in quirks mode. Make sure your " +
-                "website has a suitable doctype.");
```
kevinbarabash: I'm really stoked about this going away b/c even though we had this nice warning a fair number of people weren't seeing it.

dockers/Screenshotter/screenshotter.js
kevinbarabash: I tried removing the second `get` but the screenshotter hung after starting the chrome tests after completing the firefox tests.  Not sure why since server.js doesn't serve this URL.

kevinbarabash: I switched the order in which browsers run the tests putting Chrome first and that fixed the hanging issue I was seeing.  Maybe firefox wasn't cleaning up its web driver stuff quickly enough, idk.

dockers/Screenshotter/screenshotter.js
kevinbarabash: After reading through the code in `ss-render.html` I see that it doesn't make use of `mode` and actually the only thing that does anything with `mode` is `takeScreenshots` which determines which doctype string to prepend the `ss-render.html` with.  It probably doesn't need to be included in the `url` either by I'm going to leave it there as a cache busting mechanism.

kevinbarabash:
@gagern so when squashing and merging, github only gives credit for the person who opened the PR and the person who merged it.  Next time I'll rebase when there's multiple contributors.

kohler:
@kevinbarabash So up above @xymostech said “Okay, I did a bit of testing on IE 9 and found that quirks mode is awful. So I'm gonna modify this to go back to warning on quirks mode and to just testing limited-quirks mode everywhere.” Would it be OK to not test full-quirks mode?

kevinbarabash:
@kohler thanks.  I was a little too eager to merge this.  I'm going to revert this and will refactor following that guidance.

PR #612, opened 2017-01-09
Update dependencies to more recent version

gagern:
I thought the recent release of 0.7.0 and the planned switch to ES6 syntax might be a good opportunity to modernize some of our dependencies as well.

All these version ranges include the latest version at the time of this commit, except for the selenium webdriver.  There version 3 is incompatible with version 2, and switching to a version 3 library appears to require switching to version 3 docker images as well, which would entail using different browser versions which in turn would lead to differences for a large number of screenshots.  That doesn't appear warranted at this time.


kevinbarabash:
Thanks for upgrading deps.  I'm especially glad to see eslint 3.  Most (all?) of the third party plugins for eslint require 2+.

Issue #588, opened 2016-12-20
Support for `\verb`


allisonshaw:
Are there plans to support `\verb` in the near future? If not, is there an alternative to manually escaping every character KaTeX treats as special? 
There are times where putting some verbatim material without all the escaping is certainly useful.

kevinbarabash:
No plans to implement this per se, but a PR would welcome.  It seems a little tricky because it treats characters differently from the default.

gagern:
I think I'd implement `\verb` at the lexer level. As we don't have catcodes and all that, doing it any other way would be *very* tricky. On the other hand, it would be fairly easy to match `/\\verb(.)(.*?)\1/` or similar (perhaps `/\\verb\s*(\S)(.*?)\1/` if TeX allows whitespace between the command and the delimiter), and return that as a special kind of token, to receive appropriate processing down the line.

edemaine:
Implemented in #614, following @gagern's suggestion.

edemaine:
#614 was merged.

kevinbarabash:
@edemaine thanks for closing.

Issue #574, opened 2016-11-29
KaTeX fonts have too-thin space


kohler:
The space character in KaTeX fonts is a bit too thin. In KaTeX_Main-Regular.ttf for example, space is 250 units wide. Since cmr10 has `(FONTDIMEN ... (SPACE R 0.333))`, the space should be 333 units wide. For confirmation, 333 is the width given to the `space` and `nbspace` characters in the Type 1 CMR10 font distributed by AMS.

Is this a big deal? Not really. But it explains why (for example) the Cases screenshotter test output looks more cramped on KaTeX than LaTeX.

kevinbarabash:
@kohler nice find.  I ran `tftopl cmr10` and realized that all of the sigma values in https://github.com/Khan/KaTeX/blob/master/src/fontMetrics.js#L37-L42 that `0.0` should be non-zero numbers.  We aren't using those particular sigmas right now, but we might at some point.

Where is the space character actually used though?

kohler:
In text mode, and for `\ `. But text mode mainly.

kohler:
@kevinbarabash Can you provide a bit of guidance on how this should be fixed? I'd like to take a crack but not sure where to begin.

kevinbarabash:
Hmm... I wonder if it makes sense to render spaces in text mode and `\ ` as a kern of a particular size with the size depending on whether we're display/text style, script style, or scriptscript style.  The reason why I'm not suggesting changing the font is that we only have a single font which would be script and scriptscript styles would be wrong.  Another option would be to add styles to .less for the spaces.  The three values that should appear for space are `0.333, 0.385, 0.472` corresponding to cmr10, cmr7, and cmr5.

kevinbarabash:
I thought we added these values to `sigmasAndXis` in fontMetrics.js, but they're all `0.000`.  We should start by adding those values and then update how we handle spaces to use those values.  I'm a little foggy thought about what to do for different fonts.  It seems like we'd need to extract and store these values for other fonts as well like `cmbx`, `ctti`, etc.

edemaine:
I fixed the width of space in the typewriter font, I believe, but I guess the others could use adjustment. I don't think LaTeX has these characters in its fonts (it uses glue), so it makes sense that they need manual adjustment.

PR #614, opened 2017-01-09
Implement \verb

edemaine:
Here's an implementation of `\verb`, following #588.  As @gagern suggested, the main parsing happens in the lexer, followed by code in `parseSymbol` to handle commands *starting* with `\verb`.

Playing around with LaTeX, I found (and implemented) the following behavior:
* `\verb` will treat a following newline character as the start/end character.
* Otherwise, the `\verb` argument cannot have a newline in it.
* `\verb` will treat a following space as the start/end character.

I'm pretty new to generating output in MathML and HTML, and struggled to get the `mathtt` font, so please check that code in particular...


src/Lexer.js
```diff
@@ -80,6 +80,8 @@ var tokenRegex = new RegExp(
     "([ \r\n\t]+)|" +                                 // whitespace
     "([!-\\[\\]-\u2027\u202A-\uD7FF\uF900-\uFFFF]" +  // single codepoint
     "|[\uD800-\uDBFF][\uDC00-\uDFFF]" +               // surrogate pair
+    "|\\\\verb\\*([^\0]).*?(?:\\3|\n|$)" +            // \verb*
```
gagern: Why `[^\0]`? If you want any character, including newline, you may write that as `[^]`. If there is some special semantics for `\0` it should be explained in a comment or the PR discussion.

src/Lexer.js
```diff
@@ -80,6 +80,8 @@ var tokenRegex = new RegExp(
...
+    "|\\\\verb([^*a-zA-Z]).*?(?:\\4|\n|$)" +          // \verb unstarred
```
gagern: How about just writing `.*?\\4` (and likewise for `\\3`)? The use of `.` already excludes `\n`, as well as `\r`, `\u2028` and `\u2029`. I think that is a reasonable fit for how TeX does this, but `[\n]*?\\4` would be conceivable as well. If the terminating delimiter is *not* found before the end of line, then this would simply fall through to the function name getting matched against `\verb`, so we could define a function of that name to report the error.

src/Parser.js
```diff
@@ -839,6 +839,26 @@ Parser.prototype.parseSymbol = function() {
         return new ParseFuncOrArgument(
             new ParseNode("textord", nucleus.text, this.mode, nucleus),
             false, nucleus);
+    } else if (nucleus.text.slice(0, 5) === "\\verb" &&
+               !nucleus.text.charAt(5).match(/[a-zA-Z]/)) {
```
gagern: How about `/^\\verb[^a-zA-Z]/.test(nucleus.text)`?

src/buildCommon.js
```diff
@@ -367,6 +367,16 @@ var makeVList = function(children, positionType, positionData, options) {
     return vlist;
 };
 
+// Converts verb group into body string, dealing with \verb* form
+var makeVerb = function(group, options) {
+    var text = group.value.body;
+    if (group.value.star) {
+        text = text.replace(/ /g, '\u2423');  // Open box
```
gagern: Is that symbol included in the tt font? Come to think of it, do we want to complain if the input contains any symbol not present in our tt font, as we can't guarantee consistent rendering or correct metrics for these?

src/Lexer.js
edemaine: I confirmed that `\0` didn't work in LaTeX, with error message `Text line contains an invalid character.`  But that's probably just TeX being unhappy with the character in general.  I think it's better to allow `\0`, but I didn't know how.  Thanks for the tip!

src/Parser.js
edemaine: Much better.

src/Lexer.js
edemaine: Detecting unmatched delimiters was my goal, but that approach works too.  Changed.

src/buildCommon.js
edemaine: It seems that the symbol is the wrong size, so probably not in the font...  I thought that the usual "no metrics" warning would trigger in this case, but it seems not to, because I'm passing an entire string instead of individual characters.  I changed the main code to look up characters one at a time, then try to combine them, as in `text`.

edemaine:
I implemented your comments.  (Thanks @gagern!)  The last issue seems to be the space character in `\verb*`, which isn't in the `mathtt` font.  Any ideas on how to fix this?  Can we add it to the font, with proper spacing, somehow?

src/Parser.js
```diff
@@ -839,6 +839,24 @@ Parser.prototype.parseSymbol = function() {
...
+        this.consume();
+        var arg = nucleus.text.slice(5);
+        var star = (arg.charAt(0) === "*");
+        if (star) {
+            arg = arg.slice(1);
+        }
+        // Lexer's tokenRegex is constructed to always have matching
+        // first/last characters.
+        if (arg.length < 2 || arg.charAt(0) !== arg.slice(-1)) {
+            throw new ParseError("\\verb failure to parse");
```
gagern: What input would trigger this? Actually it might make sense to document error conditions in [`test/errors-spec.js`](https://github.com/Khan/KaTeX/blob/master/test/errors-spec.js).

edemaine: No inputs would trigger this.  I was treating it as an assertion before the next line which strips those characters.  But I can also remove it if you prefer.  I'll add an error test; thanks for pointing me to that.

gagern:
While I have some rough ideas of how the font generation works behind the scenes, I haven't actually done that. What I understand is roughly this: The TeX fonts were taken, thickened, got unicode mappings and were saved to various font formats. The metrics were derived directly from the texmf metrics descriptions, but with some replacements. That part is included in [our `metrics` directory](https://github.com/Khan/KaTeX/tree/master/metrics). Some of the font generation was inherited from MathJax, but development forked off at some point, as far as I understand. Maybe https://github.com/Khan/MathJax-dev/tree/master/fonts is where all the magic happens, but I haven't looked closely at that yet. @kevinbarabash and @xymostech will know more.

By the way, your branch as it stands is difficult to review due to the merge from master. Can you rebase that on current master? `git fetch origin && git rebase -i origin/master` should do the trick, except you'll likely have some conflicts along the way and should resolve these appropriately.

edemaine:
Rebased.  Personally, I think we could merge this PR into master now, and turn the missing character into a separate issue.  But I understand if you'd rather wait, or remove `\verb*` support until then.

gagern:
I would be fine with this approach too, if we think we can get the character before the next release. I wouldn't want an incomplete feature in a release, since people may depend on behavior which won't be forward compatible. @kevinbarabash, can you judge how difficult adding that character will be? Should we disable the `\verb*` command for now? Apart from that, I'd like another look at the code now that it's rebased, but it's too late in the day around here.

kevinbarabash:
It isn't too difficult, here are the steps:
- modify https://github.com/Khan/MathJax-dev/blob/master/fonts/OTF/TeX/makeFF#L1500-L1529 to include the new character
- follow the instructions in dockers/MathJaxFonts

It's been a while since I've done it so hopefully it all still just works.  I checked my system and `cmtt` does indeed have metrics for character 32 which should be the space.

kevinbarabash:
@edemaine give it a try.  Once you check out https://github.com/Khan/MathJax-dev you should be able to run the command that the docker runs locally to test your changes out which should be:
```
cp MathJax-dev/default.cfg MathJax-dev/custom.cfg
make -C MathJax-dev custom.cfg.pl
```

gagern:
@kevinbarabash I think we should have these steps documented somewhere. What do you suggest, Wiki or a README file in our repository or something else?

edemaine:
@kevinbarabash I get the following error:

```
make ttf eot woff woff2
perl makeBlacker 15 # values between 10 and 30 seem best
Editing mftrace
Can't read '': No such file or directory
make: *** [lib/mftrace-modified] Error 2
```

caused by the following code in `makeBlacker`:

```
sub editMftrace {
  my $oldMFTRACE = $MFTRACE_PATH;
  $MFTRACE = "./lib/mftrace-modified";
  print "Editing mftrace\n";
  open(MFT,$oldMFTRACE) || die "Can't read '$oldMFTRACE': $!\n";
```

I can confirm that there is no `lib/mftrace-modified`.  How does this get generated?

gagern:
I see the Dockerfile and the corresponding README *are* in the KaTeX repo. Makes my request to have this documented boil down to adding a hint on how to add symbols.

I see that the Dockerfile does reference Google Code in a number of places. I'm currently trying to modernize that.

kevinbarabash:
@edemaine not sure what's going on.  I'll give it a try later today.

@gagern I'll post a set of working instructions, once I figure them out... probably on the wiki.

gagern:
Check out #624 if you want to tweak the fonts. There you can use your own fork of the MathJax-dev repository, instead of the official one. It should even be possible to include `git clone` flags like `--branch`, but it should be in the same argument as the URL. Report on #624 if anything doesn't work as intended. Also report if it does, so that additional features you might use can get documented.

gagern:
I've included a font that includes the space symbol, based on https://github.com/Khan/MathJax-dev/pull/2. The `\verb*` looks good now, I'd say. I also noticed that the when using the unstarred version, the spaces are smaller than all the other glyphs. In the long run we should try to make the Teletype space equally wide. @kevinbarabash do you know where to tweak that?

src/buildHTML.js
```diff
@@ -1232,6 +1227,18 @@ groupTypes.delimsizing = function(group, options) {
             [group.value.mclass]);
 };
 
+groupTypes.verb = function(group, options) {
+    var text = buildCommon.makeVerb(group, options);
+    var body = [];
+    var i;
+    for (i = 0; i < text.length; i++) {
+        body.push(buildCommon.makeSymbol(
+            text[i], "Main-Regular", group.mode, options, ["mathtt"]));
```
gagern: It feels a bit weird that `\verb` can handle input symbols which other commands can not, e.g. `\mathtt{γ}` doesn't work yet, but `\verb|γ|` does, with a warning to console, and `\verb|“|` doesn't even warn. Some might argue that incorrectly displaying a glyph is better than not displaying it at all, while others might worry about breaking existing content if we start supporting these symbols in the future.

So far (and [in the context of unicode input](https://github.com/Khan/KaTeX/pull/261#issuecomment-115169955)) KaTeX chose the latter approach: don't support input unless you can do so in a compatible fashion. So perhaps we should filter for those symbols we actually have Typewriter metrics for?

If we take the other approach, and deliberately allow all symbols, even those we don't have fonts or metrics for, then it might be a good idea to consider surrogate pairs in this loop here. If the input contains a surrogate pair, then we want to keep the pair intact while building a symbol node from it.

test/screenshotter/ss_data.yaml
```diff
@@ -148,6 +148,8 @@ UnsupportedCmds:
     noThrow: 1
     errorColor: "#dd4c4c"
     nolatex: deliberately does not compile
+Verb:
+    \verb \verb , \verb|\verb|, \verb!<x> & </y>!
```
gagern: Now that we have ␣, there should be a `\verb*` in there as well. Perhaps use an array environment to provide multiple lines of content in a single screenshot.

src/buildHTML.js
```diff
@@ -1232,6 +1227,18 @@ groupTypes.delimsizing = function(group, options) {
...
+    }
+    buildCommon.tryCombineChars(body);
+    return makeSpan(["mord", "verb"], body, options);
```
gagern: Why the `"verb"` class? I don't see any specific styling for this. And the more classes we use in KaTeX, the more classes may not be usable by the surrounding page for their own purpose. Personally I'd even suggest prefixing all our classes with `KaTeX-` or some such, in order to avoid conflicts. Unless we do that, I'd be extra careful to only introduce new classes when we have to.

edemaine: Both good points.  I was imagining a user might want to style `\verb`s in some special way (like Github here does with `backquotes`), but it's a pretty hypothetical situation, and maybe not appropriate for LaTeX formatting, so probably not worth keeping this class.  I'll remove it.

edemaine:
@gagern Thanks for updating the font!  The `fontMetricsData.js` you committed has trailing commas which eslint doesn't like.  Should I be rebasing to es6 branch or new master, or should that be fixed upstream?  (Presumably shouldn't fix it manually...)

I agree that the space in `\verb| |` is too short.

gagern:
We've taught eslint to like (and in fact enforce) trailing commas again, in #622. Unfortunately the file is now in conflict, presumably with this. I feel all those comma changes were a pretty annoying detour, and will be happy once we no longer have any active branches without trailing commas. Until then we'll need to rebase – again.

edemaine:
Indeed, looks like characters 2018 and 2019 were added to the typewriter font by someone else in the meanwhile.  I assume those should be preserved.  (Have you all been keeping your `MathJax-dev` repository in sync?  It's a little disconcerting that I'm manually merging an autogenerated file...)

For reference, here's the diff:

```
<<<<<<< HEAD
        "2018": [0, 0.61111, 0, 0],
        "2019": [0, 0.61111, 0, 0],
        "8242": [0, 0.61111, 0, 0],
=======
        "8216": [0, 0.61111, 0, 0],
        "8217": [0, 0.61111, 0, 0],
        "8242": [0, 0.61111, 0, 0],
        "9251": [0.11111, 0.21944, 0, 0],
>>>>>>> ff241b1... Include space symbol in typewriter font, and fix single quotes
```

which I'm merging to the union

```
        "2018": [0, 0.61111, 0, 0],
        "2019": [0, 0.61111, 0, 0],
        "8216": [0, 0.61111, 0, 0],
        "8217": [0, 0.61111, 0, 0],
        "8242": [0, 0.61111, 0, 0],
        "9251": [0.11111, 0.21944, 0, 0],
```

src/buildHTML.js
edemaine: I don't have strong feelings either way.  Maybe I'm used to the warnings for unsupported symbols because of my experience with #618. 😄 I also feel like it's not crazy to pass through text of `\verb` mostly verbatim, but warning when that might do bad things. (If you like, that warning could be modified to warn of possible incompatibilities with future versions, i.e., don't rely on this behavior.)  Anyway, let me know what you think is best and we can work on a tweak either way.  (I'm afraid I don't actually know how to do the surrogate pair detection, or to fail when characters are not in the font, but I'm guessing that is done elsewhere in the code that I can mimic/re-use.)

test/screenshotter/ss_data.yaml
edemaine: Good ideas; done.

edemaine:
I was still getting "No character metrics" warning for `␣` until I realized `\verb` was using `Main-Regular` but needed to use `Typewriter-Regular`.  Fixed.  Now I properly get a warning that space (` `) has no character metrics defined, so maybe the fix is to define character metrics for it?

Somewhat related, I also get that `*` is not in the Typewriter font, which seems like an odd omission.  Maybe worth reviewing which ASCII symbols are missing and adding them appropriately?

edemaine:
Oops, `*` was failing only in `\verb`, not `\texttt`.  I needed to switch `\verb` to text mode instead of math mode.

Typewriter font needs character metrics defined for space (32) and for no-break space (160).  (I realized that `\verb` maps spaces to no-break spaces.)

kevinbarabash:
I did some digging to see what LaTeX is doing for spaces in `\texttt` and `\texttt{a b}` results in the following:
```
\mathord
.\mathchoice
.D\mathord
.D.\hbox(6.11111+0.0)x15.74986
.D..\OT1/cmtt/m/n/10 a
.D..\glue 5.24995
.D..\OT1/cmtt/m/n/10 b
.T\mathord
.T.\hbox(6.11111+0.0)x15.74986
.T..\OT1/cmtt/m/n/10 a
.T..\glue 5.24995
.T..\OT1/cmtt/m/n/10 b
.S\mathord
.S.\hbox(4.27777+0.0)x11.15639
.S..\OT1/cmtt/m/n/7 a
.S..\glue 3.7188
.S..\OT1/cmtt/m/n/7 b
.s\mathord
.s.\hbox(3.05554+0.0)x7.96884
.s..\OT1/cmtt/m/n/5 a
.s..\glue 2.65628
.s..\OT1/cmtt/m/n/5 b
```
The glue for `D` and `T` styles matches the width of the the teletype characters which are 0.524995 em wide.  Our `ptPerEm` variable corrects for the factor of 10 difference.  The glue for `S` and `s` is different.

kevinbarabash:
I also tried `\verb!a b!\scriptstyle\verb!a b!\scriptscriptstyle!a b!` and got the following:
```
\mathord
.\hbox(6.11111+0.0)x15.74986
..\OT1/cmtt/m/n/10 a
..\penalty 10000
..\glue 5.24995
..\OT1/cmtt/m/n/10 b
\scriptstyle
\mathord
.\hbox(6.11111+0.0)x15.74986
..\OT1/cmtt/m/n/10 a
..\penalty 10000
..\glue 5.24995
..\OT1/cmtt/m/n/10 b
\scriptscriptstyle
\mathord
.\hbox(6.11111+0.0)x15.74986
..\OT1/cmtt/m/n/10 a
..\penalty 10000
..\glue 5.24995
..\OT1/cmtt/m/n/10 b
```
which seems to indicate that spaces in `\verb` behaves different from in `\texttt`.

edemaine:
@kevinbarabash `\verb` seems to ignore `\scriptstyle` and `\scriptscriptstyle` in LaTeX: they all appear at full size.  `$\verb|a b|\scriptsize\verb|a b|$` matches the glues from `\texttt`:

```
\mathord
.\hbox(6.11111+0.0)x15.74986
..\OT1/cmtt/m/n/10 a
..\penalty 10000
..\glue 5.24995
..\OT1/cmtt/m/n/10 b
\mathord
.\hbox(4.27777+0.0)x11.15639
..\OT1/cmtt/m/n/7 a
..\penalty 10000
..\glue 3.7188
..\OT1/cmtt/m/n/7 b
```

kevinbarabash:
I didn't realize there was a `\scriptsize` in addition to `\scriptstyle`.

gagern:
Can we generate a space character in the font by using an empty glyph box of the appropriate size?

edemaine:
@kevinbarabash `\displaystyle`, `\textstyle`, `\scriptstyle`, `\scriptscriptstyle` are for math mode.  `\Huge`, `\huge`, `\LARGE`, `\Large`, `\large`, `\normalsize`, `\small`, `\footnotesize`, `\scriptsize`, `\tiny` are (I think primarily) for text mode.  `\verb` is fundamentally a text-mode command, I guess, despite working in both modes.  All a little confusing -- while `\scriptstyle` and `\scriptsize` sound like they correspond, I have no idea which size command corresponds to `\scriptscriptstyle`...  I can see where the LaTeX size commands are defined (`size##.clo`) but I'm not sure how TeX defines the font sizes for style commands.

gagern:
> … I have no idea which size command corresponds to `\scriptscriptstyle` …

According to [the Tex Book](http://www.ctex.org/documents/shredder/src/texbook.pdf) Appendix B, plain TeX contains these settings:

```tex
\textfont0=\tenrm \scriptfont0=\sevenrm \scriptscriptfont0=\fiverm
\def\rm{\fam0 \tenrm}
\textfont1=\teni \scriptfont1=\seveni \scriptscriptfont1=\fivei
\def\mit{\fam1 } \def\oldstyle{\fam1 \teni}
\textfont2=\tensy \scriptfont2=\sevensy \scriptscriptfont2=\fivesy
\def\cal{\fam2 }
\textfont3=\tenex \scriptfont3=\tenex \scriptscriptfont3=\tenex
```

Looking for corresponding settings in LaTeX I notice commands like `\ssf@size` and `\DeclareMathSizes`. Looking for the latter in [`source2e.pdf`](http://mirrors.ctan.org/macros/latex/base/source2e.pdf) I see a bunch of calls from [`fontdef.dtx`](http://mirrors.ctan.org/macros/latex/base/fontdef.dtx):

```tex
 \DeclareMathSizes{5}{5}{5}{5}
 \DeclareMathSizes{6}{6}{5}{5}
 \DeclareMathSizes{7}{7}{5}{5}
 \DeclareMathSizes{8}{8}{6}{5}
 \DeclareMathSizes{9}{9}{6}{5}
 \DeclareMathSizes{\@xpt}{\@xpt}{7}{5}
 \DeclareMathSizes{\@xipt}{\@xipt}{8}{6}
 \DeclareMathSizes{\@xiipt}{\@xiipt}{8}{6}
 \DeclareMathSizes{\@xivpt}{\@xivpt}{\@xpt}{7}
 \DeclareMathSizes{\@xviipt}{\@xviipt}{\@xiipt}{\@xpt}
 \DeclareMathSizes{\@xxpt}{\@xxpt}{\@xivpt}{\@xiipt}
 \DeclareMathSizes{\@xxvpt}{\@xxvpt}{\@xxpt}{\@xviipt}
```

I'd read this to mean that at a text size of 10pt as used by KaTeX, the full size math font will have 10pt, the script style will use 7pt and the scriptscript style will use 5pt. If the text font size is not this but some other integer from the list, another preloaded setting will be used. If no such setting is found, e.g. in the case of fractional font sizes, `\calculate@math@sizes` from [`ltfssbas.dtx`](http://mirrors.ctan.org/macros/latex/base/ltfssbas.dtx) will compute fonts based on `\defaultscriptratio` and `\defaultscriptscriptratio` which default to `.7` and `.5` respectively. That's what KaTeX is using, too, in [its stylesheet](https://github.com/Khan/KaTeX/blob/v0.7.0/static/katex.less#L228-L238).

I *think* that the original Idea of TeX was to set the text font outside the math formula using the text sizing commands, then adjust the math size using the styles instead. Using text sizing commands in math mode is probably poorly supported in many situations. On the other hand, it can be quite useful at times.

edemaine:
I see, so `\textstyle`, `\scriptstyle`, and `\scriptscriptstyle` are all relative to the current text size.  E.g., in 10pt mode, `\large` switches to 12pt, which increases `\scriptstyle` and `\scriptscriptstyle` by 1pt each.  Neat...

edemaine:
FWIW, it looks like KaTeX (in `Styles.js`'s `styles`) follows the `.7` and `.5` ratios in all cases, instead of referring to this table for special cases.  Cleaner this way, but for full LaTeX compatibility this could change...  (I think the reason for it might be archaic, namely, fonts looking better at integer sizes instead of random sizes.)

gagern:
I thought the same. One thing has me wondering, though: there appears to be a difference between release and non-release builds of LaTeX. `\latexrelease.sty` has this table, `\DeclareMathSizes` does not. So perhaps even in the LaTeX world there are situations where the ratios are used even where they lead to fractional font sizes. And I honestly think it's more of a bug than a feature if changing the text font size from 12pt to 12.001pt has a noticable difference.

edemaine:
@gagern With all your recent experience with fonts, do you know how to fix the space character in typewriter mode?  Although, I'm not sure whether that's an issue with this `\verb` patch, or an issue with `tt` in general...

kevinbarabash:
We could edit the .ttf font, copy any glyph into the `0x20` position, and then delete the path for the newly created `0x20`.  We could also special case things where if we find a space character that's supposed to be rendered using the `tt` font, we could render an empty span with the correct width.

gagern:
I still haven't understood (nor really investigated) exactly how the font creation process works. Adding a glyph that is not present in the TeX font would have to be done by editing that [`Makefile`](https://github.com/Khan/MathJax-dev/blob/master/fonts/OTF/TeX/Makefile), so that it fits in with out dockerized font creation process, right?

Looking a bit closer at the font creation, am I to understand that our TTF fonts are actually generated from *bitmap* versions of computer modern? Why the hell is that? I would expect that we should obtain better quality at smaller file sizes if we could create fonts more directly from metafont sources of Type 1 vector fonts. I am kind of surprised to see that [cm-unicode](http://cm-unicode.sourceforge.net/) is using [mftrace](http://lilypond.org/mftrace/), too. Strange!

gagern:
http://stackoverflow.com/questions/20384088/script-for-metafont-to-ttf-conversion suggested `md2pt1` as an alternative to `mftrace`. Sounds reasonable to me. The way I see it, both are perl scripts that call `mf` resp. `mpost` at some point, so the same patch to make the font a bit blacker should be applicable there as well. I see lillypond [switched from `mftrace` to `mf2pt1`](https://github.com/lilypond/lilypond/commit/37dd6278153b188ba59c4c58ff90540c4d7d69ad) in 2008, perhaps we can learn something from that.

@kevinbarabash can you edit the settings of https://github.com/Khan/MathJax-dev to enable issues on that project? I guess it might be more appropriate to discuss this there, instead of further distracting from the issue at hand here.

kevinbarabash:
I was thinking of editing the .ttf file after it's generated maybe using python fonttools or opentype.js.  Having that process fit in with the Makefile makes sense.

I didn't realize that we were tracing bitmaps.  [mf2pt1](http://www.ctan.org/pkg/mf2pt1) sounds like a superior approach.  I've enabled issues in https://github.com/Khan/MathJax-dev so that we can discuss this further there.

edemaine:
Revitalizing this PR:
* Rebased
* Implemented space support via manual spacing (like LaTeX does, but unlike how we deal with spaces elsewhere in KaTeX).
* Fixed the issue I brought up earlier that `\verb` ignores the style and only handles the size, and updated the screenshotter test to confirm this.

Here's a texcmp:

verb

I did not check whether characters are in the font, and rather rely on the warning that KaTeX produces about missing characters (see @gagern's old review, first comment).  I'm not sure this is best.  If someone agrees, I'll work on it.  (I think I'd need to make a very simple custom lexer for the `\verb` string, and then mimic a small subset of `parseSymbol`.)

kevinbarabash:
I was really confused as to why this was working b/c the visible space (open box) character doesn't show up in Apple's Font Book font viewing utility even though KaTeX appears to be rendering the glyph just fine.  I tried opening KaTeX_Typewriter-regular in https://opentype.js.org/glyph-inspector.html and it shows up correctly.  Not sure why Font Book wasn't displaying it.

screen shot 2017-08-22 at 10 27 11 pm

kevinbarabash:
In the font file the glyph is at position 0x20 which is normally the space.  I looked at some other fonts and Font Book doesn't render the space in any of them which I guess makes sense b/c in other fonts it's blank so there's nothing to look at.

src/buildCommon.js
```diff
@@ -169,6 +169,20 @@ const makeOrd = function(group, options, type) {
 };
 
 /**
+ * Combine as many characters as possible in the given array of characters
+ * via their tryCombine method.
+ */
+const tryCombineChars = function(chars) {
```
kevinbarabash: Maybe this should be:
```
const tryCombineTextNodes = function(nodes) {
```
It's possible that we might also get passed a `documentFragment` so we should probably add a `tryCombine` method on that class as well just in case.

test/errors-spec.js
```diff
@@ -178,6 +178,17 @@ describe("Parser:", function() {
...
 
+    describe("#verb", function() {
+        it("complains about mismatched \\verb with end of string", function() {
+            expect("\\verb|hello").toFailWithParseError(
+                "\\verb ended by end of line instead of matching delimiter");
+        });
+        it("complains about mismatched \\verb with end of line", function() {
+            expect("\\verb|hello\nworld|").toFailWithParseError(
+                "\\verb ended by end of line instead of matching delimiter");
+        });
```
kevinbarabash: Is there a way to test the exception being throw in Parser.js?

kevinbarabash:
@edemaine friendly ping.  I left some comments and there are a few merge conflicts, but I think this PR is pretty close to being done.

src/buildHTML.js
```diff
@@ -1203,6 +1198,30 @@ groupTypes.delimsizing = function(group, options) {
...
+    const text = buildCommon.makeVerb(group, options);
+    const body = [];
+    // \verb enters text mode and therefore is sized like \textstyle
+    const newOptions = options.havingStyle(options.style.text());
+    for (let i = 0; i < text.length; i++) {
+        if (text[i] === ' ') {  // Space appear as nonbreaking space
+            // The space character isn't in the Typewriter-Regular font,
+            // so we implement it as a kern of the same size as a character.
+            // 0.524995 is the width of a texttt character in LaTeX.
+            // It automatically gets scaled by the font size.
```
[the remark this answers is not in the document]
edemaine: You suggested the reverse in #574. 😄 FWIW, TeX implements spaces as skips instead of characters, but I agree it'd be more symmetric (and simpler) to use a space character. I'm not sure where to collect together font issues -- perhaps on new font repo?   Added https://github.com/KaTeX/katex-fonts/issues/1

src/buildHTML.js
```diff
@@ -1203,6 +1198,30 @@ groupTypes.delimsizing = function(group, options) {
...
+            const rule = makeSpan(["mord", "rule"], [], newOptions);
+            rule.style.marginLeft = "0.524995em";
```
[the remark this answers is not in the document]
edemaine: Changed.

src/buildCommon.js
```diff
@@ -394,6 +408,17 @@ const makeVList = function(children, positionType, positionData, options) {
     return vtable;
 };
 
+// Converts verb group into body string, dealing with \verb* form
+const makeVerb = function(group, options) {
+    let text = group.value.body;
+    if (group.value.star) {
+        text = text.replace(/ /g, '\u2423');  // Open Box
+    } else {
+        text = text.replace(/ /g, '\xA0');    // No-Break Space
```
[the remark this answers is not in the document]
edemaine: `\verb|hello                                      world|` will not break in LaTeX, and furthermore renders as multiple spaces, whereas the usual behavior for invisible spaces is that they combine into one.

I'm less clear on why I write it as `\xA0` here and (formerly) ` ` in `buildHTML.js`.  Probably the former is clearer, so I've changed the latter.  Not sure if we have a preferred style on this matter.

src/Parser.js
```diff
@@ -908,6 +908,24 @@ class Parser {
...
+            this.consume();
+            let arg = nucleus.text.slice(5);
+            const star = (arg.charAt(0) === "*");
+            if (star) {
+                arg = arg.slice(1);
+            }
+            // Lexer's tokenRegex is constructed to always have matching
+            // first/last characters.
+            if (arg.length < 2 || arg.charAt(0) !== arg.slice(-1)) {
+                throw new ParseError("\\verb failure to parse");
```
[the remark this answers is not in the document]
edemaine: This is actually more of an assertion: the regex *should* guarantee that that error is *never* thrown.   The actual error that will get thrown is `\\verb ended by end of line instead of matching delimiter` which is more descriptive.  (It doesn't mention the delimiter character, but that is not trivial to add... Let me know if I should work on it.)

src/Parser.js
```diff
@@ -908,6 +908,24 @@ class Parser {
...
+            }
+            arg = arg.slice(1, -1);
```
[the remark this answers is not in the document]
edemaine: Added a comment.

test/errors-spec.js
edemaine: I'm confused -- isn't that what this is doing?  Or do you mean confirming that the error is actually of type `ParseError`?  We could probably rewrite `toFailWithParseError` to do this.

edemaine:
@kevinbarabash Sorry for the delay (been tied up lately), and thanks for the reminder.  Just merged with master, and added one commit addressing your comments.  Hopefully good to go now.

test/errors-spec.js
kevinbarabash: This is the error that's being thrown in `functions.js`.  The error thrown in `Parser.js` has the text `"\\verb failure to parse"`.

kevinbarabash: From your comment above it sounds like it should never happen though, so I guess we can't test for it.

edemaine: Oh, I see.  As described above, that error can never be thrown.

src/Parser.js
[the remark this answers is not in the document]
kevinbarabash: I thought the first and last characters of the `arg` were the delimiter so it seemed like a trivial change to add those characters to the error message.  If this case should never happen it's probably more important to say in the error that this is a bug and that it should be reported as such.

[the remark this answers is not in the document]
edemaine: Rewrote the error message.

src/buildHTML.js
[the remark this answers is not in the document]
kevinbarabash: I forgot about that comment.  I think the difference here is that `\verb` is kind of like `\text` and there we use real spaces.  I think what you have here is fine.

kevinbarabash:
The screenshots need to be regenerated... probably b/c of that change from `0.524995` to `0.525`.

edemaine:
Thanks! Sorry for the screenshot oversight.

Issue #613, opened 2017-01-09
0.7.0 no dist folder


ymakhloufi:
When pulling in katex 0.7.0 via bower, I don't get a dist-folder but only a bower.json file in the bower_components/katex/ folder. Even when I remove it and pull it in as a sub dependency of angular-katex it is the same result. Why could that be?

gagern:
OK, this has been reported before, in #240 which got closed as a duplicate of #114. That one was aimed to address both bower and npm, and got fixed with #436 closing the npm half of the issue. The bower problem apparently remains open. @spicyj do you agree with this analysis?

Bower 1.8.0 reports this for me:

```
bower install katex
bower not-cached    https://github.com/Khan/KaTeX.git#*
bower resolve       https://github.com/Khan/KaTeX.git#*
bower download      https://github.com/Khan/KaTeX/archive/v0.7.0.tar.gz
bower extract       katex#* archive.tar.gz
bower invalid-meta  for:…/bower.json
bower invalid-meta  The "name" is recommended to be lowercase, can contain digits, dots, dashes
bower invalid-meta  The "main" field cannot contain minified files
bower invalid-meta  The "main" field cannot contain minified files
bower resolved      https://github.com/Khan/KaTeX.git#0.7.0
bower install       KaTeX#0.7.0

KaTeX#0.7.0 bower_components/KaTeX
```

That tarball is a *source* archive, so it doesn't contain `dist`, nor should it. The error messages are something else but shouldn't be too severe.

ymakhloufi:
Well it appears as if pulling in v0.6.0 with bower works, only from v0.7.0 ist is a problem, but since we need the support for umlaut characters, that is not an option...

zeitgeist87:
I have the exact same issue. Bower 1.8.0 reports this:

```bash
node_modules/bower/bin/bower install -V
bower katex#*               not-cached https://github.com/Khan/KaTeX.git#*
bower katex#*                  resolve https://github.com/Khan/KaTeX.git#*
bower katex#*                 download https://github.com/Khan/KaTeX/archive/v0.7.0.tar.gz
bower katex#*                  extract archive.tar.gz
bower katex#*             invalid-meta for:/tmp/andreas/bower/f1a02a1ae7fa47b237f875a1e98ba21e-7732-wZegpf/bower.json
bower katex#*             invalid-meta The "name" is recommended to be lowercase, can contain digits, dots, dashes
[...]

PR #615, opened 2017-01-09
Improve release script and bower support

gagern:
I've had a long look at the release script, particularly with the aim of fixing #613. Here is what I changed:

* unignore the dist directory for the tag commit
* avoid some bower warnings by changing name to lower case and referencing unminified files
* drop [deprecated `version` field](https://github.com/bower/spec/blob/master/json.md#version) from bower.json file and the commands to manage it
* allow creating releases off branches other than master (e.g. for 0.7.1 from 0.7 branch with bower support?)
* clean build and dist directories to avoid stale files lying around
* create annotated tags for releases (see [Why should I care about lightweight vs. annotated tags?](http://stackoverflow.com/q/4971746/1468366) on SO for some discussion)

As these are distinct commits, you might also cherry-pick those you like and reject the others.

I started this off an older master, and had some solution for portable sed. Only then did I notice that 4327e85796e343961fa105673a813db091813f8c already took care of that. I was using this code instead:

```sh
# Portable sed -i, see http://unix.stackexchange.com/q/92895/20807
case $(sed --help 2>&1) in
    *GNU*) sedi() { sed -i "$@"; };;
    *) sedi() { sed -i "" "$@"; };;
esac
```

and then using `sedi` as a command. Just in case you prefer that over creating then removing a backup file, let me know and I'll create another commit.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -1,6 +1,5 @@
 {
   "name": "katex",
-  "version": "0.8.0-pre",
   "main": [
     "dist/katex.js",
     "dist/katex.css"
--- a/bower.json
+++ b/bower.json
@@ -1,9 +1,9 @@
 {
-  "name": "KaTeX",
+  "name": "katex",
   "version": "0.8.0-pre",
   "main": [
-    "dist/katex.min.js",
-    "dist/katex.min.css"
+    "dist/katex.js",
+    "dist/katex.css"
   ],
   "homepage": "http://khan.github.io/KaTeX/",
   "description": "Fast math typesetting for the web.",
--- a/release.sh
+++ b/release.sh
@@ -45,8 +45,7 @@ git add .gitignore dist/
 # http://stackoverflow.com/a/22084103 for why we need the .bak file to make
 # this mac & linux compatible)
 sed -i.bak -E 's|"version": "[^"]+",|"version": "'$VERSION'",|' package.json
-sed -i.bak -E 's|"version": "[^"]+",|"version": "'$VERSION'",|' bower.json
-rm -f package.json.bak bower.json.bak
+rm -f package.json.bak
 
 # Make the commit and tag, and push them.
 git add package.json bower.json
@@ -63,8 +62,7 @@ if [ ! -z "$NEXT_VERSION" ]; then
 
     # Edit package.json and bower.json to the right version
     sed -i.bak -E 's|"version": "[^"]+",|"version": "'$NEXT_VERSION'-pre",|' package.json
-    sed -i.bak -E 's|"version": "[^"]+",|"version": "'$NEXT_VERSION'-pre",|' bower.json
-    rm -f package.json.bak bower.json.bak
+    rm -f package.json.bak
 
     git add package.json bower.json
     git commit -n -m "Bump master to v$NEXT_VERSION-pre"
--- a/release.sh
+++ b/release.sh
@@ -16,20 +16,21 @@ fi
 
 VERSION=$1
 NEXT_VERSION=$2
+BRANCH=$(git rev-parse --abbrev-ref HEAD)
 
 if [ -z "$NEXT_VERSION" ]; then
-    PROMPT="About to release $VERSION. Look good? [y/n] "
+    echo "About to release $VERSION from $BRANCH. "
 else
-    PROMPT="About to release $VERSION and bump master to $NEXT_VERSION-pre. Look good? [y/n] "
+    echo "About to release $VERSION from $BRANCH and bump to $NEXT_VERSION-pre."
 fi
 
-read -r -p "$PROMPT" CONFIRM
+read -r -p "Look good? [y/n] " CONFIRM
 if [ "$CONFIRM" != "y" ]; then
-    exit
+    exit 1
 fi
 
 # Make a new detached HEAD
-git checkout master
+git checkout "$BRANCH"
 git pull
 git checkout --detach
 
@@ -58,7 +59,7 @@ npm publish
 
 if [ ! -z "$NEXT_VERSION" ]; then
     # Go back to master to bump
-    git checkout master
+    git checkout "$BRANCH"
 
     # Edit package.json and bower.json to the right version
     sed -i.bak -E 's|"version": "[^"]+",|"version": "'$NEXT_VERSION'-pre",|' package.json
@@ -67,7 +68,7 @@ if [ ! -z "$NEXT_VERSION" ]; then
 
     git add package.json bower.json
     git commit -n -m "Bump master to v$NEXT_VERSION-pre"
-    git push origin master
+    git push origin "$BRANCH"
 
     # Go back to the tag which has build/katex.tar.gz and build/katex.zip
     git checkout "v$VERSION"
--- a/release.sh
+++ b/release.sh
@@ -1,29 +1,94 @@
 #!/usr/bin/env bash
 
 set -e -o pipefail
+shopt -s extglob
 
-if [ $# -lt 1 ]; then
+VERSION=
+NEXT_VERSION=
+BRANCH=$(git rev-parse --abbrev-ref HEAD)
+NARGS=0
+DRY_RUN=
+INSANE="exit 1"
+
+# usage [ERROR-MESSAGES...] EXIT-STATUS
+usage() {
+    while [[ $# -gt 1 ]]; do
+        echo "$1" >&2
+        shift
+    done
     echo "Usage:"
-    echo "./release.sh <VERSION_TO_RELEASE> [NEXT_VERSION]"
+    echo "./release.sh [OPTIONS] <VERSION_TO_RELEASE> [NEXT_VERSION]"
+    echo ""
+    echo "Options:"
+    echo " --dry-run|-n: only print commands, do not execute them."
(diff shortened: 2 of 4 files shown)


bower.json
```diff
@@ -1,9 +1,8 @@
 {
-  "name": "KaTeX",
-  "version": "0.8.0-pre",
+  "name": "katex",
   "main": [
-    "dist/katex.min.js",
-    "dist/katex.min.css"
+    "dist/katex.js",
+    "dist/katex.css"
```
xymostech: Hrm. It's annoying that bower doesn't let us specify several different things with the same filetype. I think people would probably prefer the minified version to be the default? Not sure though.

gagern:
> There haven't been any changes in master since the 0.7.0 release, right?

Yes [there were](https://github.com/Khan/KaTeX/compare/v0.7.0...master): the version bump, and the commit and merge from #612 which theoretically might break things. I deliberately wanted that one for after 0.7.0. I guess it depends on what we want 0.7.x to be. If it's bug fixes only, we should branch a maintainance branch called v0.7 off from 0.7.0^, and release from that. If we use 0.7.x for new features too, the way we'd use a 7.x.0 once we get past the leading zero, then we shouldn't call the development version 0.8.0-pre. Releasing off a non-master branch has the added benefit of actually trying out that script, now that these changes are fresh.

> In general I feel like it's bad form to allow releases from non-master.

Why? Patch level releases x.y.z should be from a x.y maintainance branch. That branch itself either forked from master around the x.y.0 release, or from the version x branch if latest development had been at a newer major version by the time x.y.0 got released. At least that's the way I see it.

> The only other thing I think I would like our release script to do is have a `--dry-run` option so that we can make sure the run would finish successfully, without actually pushing/making tags/etc.

Good idea. During development I had created shell functions which just printed the results from the calls to `git` and `npm`. Conditionally doing that might be easy. I'll look into it.

bower.json
gagern: I couldn't find any documentation about *why* that requirement exists. bower/bower@1c62adcdb6f09875434de2d886c23731a0bae53c just says it's following the spec, and https://github.com/bower/spec/blob/master/json.md had that line since the content got copied from Google docs.

I can only guess that they want subsequent steps to minify the sources. The way I understand it, they might even handle modules, so specifying our main unbrowserified entry point in `bower.json` *might* work, too.

xymostech:
> Why? Patch level releases x.y.z should be from a x.y maintainance branch. That branch itself either forked from master around the x.y.0 release, or from the version x branch if latest development had been at a newer major version by the time x.y.0 got released. At least that's the way I see it.

Okay yes, that makes sense to me. Maybe my main concern is that we're allowing for running the release script on *any* branch. I wish we could codify what you just said in code, since I know I've accidentally run the release script from the wrong place before. But it does get printed out now. So I'm fine with this.

> Good idea. During development I had created shell functions which just printed the results from the calls to git and npm. Conditionally doing that might be easy. I'll look into it.

Another thing that happened to me was I wasn't logged in to `npm`, so it got to the place where it made a new tag and then failed to actually do the upload, which required me to manually delete the tag. Maybe we could check that you're logged in as a user with the correct permissions with something like `npm owner ls katex | grep $(npm whoami)`. Just throwing out ideas here. :)

gagern:
Now I've got some sanity checks: a check for uncommited changes, one for the name of the current branch, and one for the npm login. And I've added an option to not exit after a failed sanity check. Do you want it this way? Or should the sanity checks just report the issue, and the user can decide at the confirmation whether or not to ignore them? Should the dry run mode imply not exiting on failed sanity checks? Should the dry run mode indeed execute make, or skip that, too? Should the dry run mode perform all local git operations, then skip pushing them and instead remove the local tag again?

bower.json
xymostech: Yeah, I'm not really sure what `main` means I guess. It does say that you should use unminified things though, so at least this change is correct.

xymostech:
Ahh! Beautiful! I love how `--insane` and `--dry-run` work.

> Or should the sanity checks just report the issue, and the user can decide at the confirmation whether or not to ignore them?

That might be better if someone just wants to skip one of the checks, but still wants to be warned of other things. That shouldn't be too hard, just editing what `$INSANE` is set to, right?

> Should the dry run mode imply not exiting on failed sanity checks? Should the dry run mode indeed execute make, or skip that, too? Should the dry run mode perform all local git operations, then skip pushing them and instead remove the local tag again?

I like that it executes `make`, cause that might actually catch bugs. I'm not sure what bugs would be caught by running `git` for the local operations that aren't already caught by the sanity checks, so I'm happy how this is.

gagern:
> That shouldn't be too hard, just editing what `$INSANE` is set to, right?

I guess in that case I'd make `INSANE` and `--insane` go away altogether. Simply `echo … >&2` for every failed sanity check, and rely on the fact that a sane user will say `no` to the subsequent `Look good?` prompt. Makes the code simpler, too.

> I'm not sure what bugs would be caught by running `git` for the local operations

Operations might fail because a merge is in progress, or a rebase, or because the remote isn't called origin, or because there are some git hooks that do fancy stuff, or because the git doesn't have a user id configured, or a million other reasons I can't think of just now.

gagern:
As I've just accepted #590 and #591, I've come up with a way to make updating [Subresource Integrity](https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity) hashes part of the release process. I've also dropped the `--insane` flag, as discussed, relying on a modified prompt instead.

The new script I created for this purpose is making heavy use of ES6 already. I've not subjected it to eslint yet, but intend to do so once #617 has been accepted.

gagern:
How do we proceed here and with 0.7.1? I think this one here should go into master, once you're happy with it. When that happens, we should also merge something like this into [the v0.7 branch](https://github.com/Khan/KaTeX/commits/v0.7), which I created by forking off [v0.7.0^](https://github.com/Khan/KaTeX/commits/v0.7.0%5E).

We should decide whether we want to maintain the 0.7 line of development via cherry-picks of bug fixes (whitelist) or via merges from mainline excluding breaking changes which need to be reverted (blacklist). I'd say that after 1.0.0 we should use cherry-picks for patch-level releases and merges for minor-level releases, but prior to 1.0.0 we are more flexible. With the planned switch to ES6 notation, merging large amounts of changes would probably become a pain, though. So I'd suggest we cherry-pick selected commits to v0.7. If you agree I'd do so once this here gets merged into master.

If you want me to, I can try rolling a 0.7.1 release using this updated script. To do so I'll need access to the katex package on npm, though.

gagern:
Thanks for merging, @kevinbarabash! I'll backport this to the v0.7 branch.

kevinbarabash:
@gagern sorry for the delay.  My vote would be for cherry-picks especially since we've already merged the ES6 notation changes.

gagern:
Cherry-picked to v0.7 in ad571b945f3c597474dc64a149a0106c0d36961a. @kevinbarabash, do you want to build a 0.7.1 release using this? Checking out the v0.7 branch and running the script with `0.7.1` as the only argument *should* do the trick. And if not, we want to know about it now.

gagern:
Oh, I see I've been granted npm permissions for this. Do you want me to run the release script, to better inspect whether my changes work as intended and so I get some experience doing releases, too?

release.sh
```diff
@@ -1,74 +1,161 @@
...
 git commit -n -m "v$VERSION"
-git tag "v$VERSION"
+git tag -a "v$VERSION" -m "v$VERSION"
 git push origin "v$VERSION"
 
 # Update npm (bower and cdnjs update automatically)
 npm publish
 
 if [ ! -z "$NEXT_VERSION" ]; then
     # Go back to master to bump
```
kevinbarabash: This comment should be updated.

gagern: Next time do the review before the merge… But yes, we should fix this on master, and it doesn't really matter for the 0.7 branch. I guess I'll wait till 0.7.1 is released, just in case there is anything else that needs to be changed.

release.sh
```diff
@@ -1,74 +1,161 @@
...
+
+# usage [ERROR-MESSAGES...] EXIT-STATUS
+usage() {
+    while [[ $# -gt 1 ]]; do
+        echo "$1" >&2
+        shift
+    done
     echo "Usage:"
-    echo "./release.sh <VERSION_TO_RELEASE> [NEXT_VERSION]"
+    echo "./release.sh [OPTIONS] <VERSION_TO_RELEASE> [NEXT_VERSION]"
```
kevinbarabash: For bug fix releases, would a person run `./release.sh 0.7.1 0.7.2`?

gagern: I've been asking myself this very question just now. Benefit of doing so is that we get more up-to-date information on the release branch: updated URLs, updated sri hashes, updated version number. The downside is that we'd have one more commit on the branch, but actually that might be seen as a benefit, too, since otherwise it's hard to place the release in the line of development. Do we want to enforce using a second argument?

kevinbarabash: I could go either way.  I'm not too concerned about an extra commit on a release branch.

release.sh
kevinbarabash: I don't know the release process that well I figured that you actually running the release script would suss out issues better than I'd be able to reviewing.  :(

PR #617, opened 2017-01-10
Switch to ES 6

gagern:
With 0.7.0 released, it might be a good time to do the planned switch to ES6 now. And this PR here is the first step in that direction.

1. It adds babelify to the build process, translating ES6 to ES5. The eslint configuration is adapted to cover es6, too.
2. It changes occurrences of `var` to `const` or `let` as appropriate. `eslint` only ever allows one of these two alternatives. The main purpose of this step is demonstrating that the transformations actually get applied as intended.
3. It changes the live server to serve either babelified or unbabelified files. The former is required for browsers with lacking ES6 support, which at the moment [includes Firefox](https://bugzilla.mozilla.org/show_bug.cgi?id=1094995). The latter is faster so in combination with a compatible browser it avoids slowing down development.

### DIFF
--- a/check-node-version.js
+++ b/check-node-version.js
@@ -1,0 +1,16 @@
+"use strict";
+
+var v = process.version;
+v = v.replace(/^v/,"");
+v = v.split(".");
+v = v.map(function(s){
+    return parseInt(s);
+});
+var a = v[0], b = v[1], c = v[2];
+if (a < 6 || (a == 6 && b < 5)) {
+    console.error("Node 6.5 or later required for development. " +
+                  "Version " + process.version + " found");
+    process.exit(1);
+} else {
+    console.log("OK");
+}
--- a/dockers/Screenshotter/screenshotter.js
+++ b/dockers/Screenshotter/screenshotter.js
@@ -1,27 +1,27 @@
 /* eslint no-console:0, prefer-spread:0 */
 "use strict";
 
-var childProcess = require("child_process");
-var fs = require("fs");
-var http = require("http");
-var jspngopt = require("jspngopt");
-var net = require("net");
-var os = require("os");
-var pako = require("pako");
-var path = require("path");
-var selenium = require("selenium-webdriver");
-var firefox = require("selenium-webdriver/firefox");
+const childProcess = require("child_process");
+const fs = require("fs");
+const http = require("http");
+const jspngopt = require("jspngopt");
+const net = require("net");
+const os = require("os");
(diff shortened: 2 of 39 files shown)


gagern:
Rebased onto the old and new trailing comma convention. Having this in would help testing on IE9 thanks to the browserify step.

server.js
```diff
@@ -42,22 +43,32 @@ var serveBrowserified = function(file, standaloneName) {
...
-app.get("/contrib/auto-render/auto-render.js",
-        serveBrowserified("contrib/auto-render/auto-render",
-                          "renderMathInElement"));
-
-app.get("/katex.css", function(req, res, next) {
+}
+
+function twoBrowserified(url, file, standaloneName) {
+    app.get(url, serveBrowserified(file, standaloneName, false));
+    app.get("/babel" + url, serveBrowserified(file, standaloneName, true));
```
kevinbarabash: I tried this out and there's a noticeable lag... thanks for setting this up a separate route.

dockers/Screenshotter/screenshotter.js
```diff
@@ -303,7 +303,7 @@ function findHostIP() {
     app.get("/ss-connect.js", function(req, res, next) {
         if (!katexURL) {
             katexIP = req.query.ip;
-            katexURL = "http://" + katexIP + ":" + katexPort + "/";
+            katexURL = "http://" + katexIP + ":" + katexPort + "/babel/";
```
kevinbarabash: 👍

check-node-version.js
```diff
@@ -0,0 +1,16 @@
...
+    return parseInt(s);
+});
+var a = v[0], b = v[1], c = v[2];
+if (a < 6 || (a == 6 && b < 5)) {
+    console.error("Node 6.5 or later required for development. " +
+                  "Version " + process.version + " found");
+    process.exit(1);
+} else {
+    console.log("OK");
+}
```
kevinbarabash: This is the fix for the node version check?

gagern: Yes, this version should work on all reasonable versions of GNU make, contrary to a version I had before some fixup.

kevinbarabash:
This is the future step in modernizing the codebase.  I'm excited.  Thanks @gagern.

gagern:
I realized that this does slow down the screenshotter immensely. I guess we should try to come up with a solution to do screenshot tests with a cached version. Or perhaps even with the build version, as present on the file system. Something like that.

Issue #600, opened 2017-01-04
\Box symbol gives warning


Joe312341:
When trying to use \Box I get the following warning in Chrome:

`No character metrics for '□' in style 'Main-Regular'`

It does display correctly though.

edemaine:
I'm new here, so I don't know the exact problem, but I observe that character 9633 (□) is defined in style AMS-Regular but not in style Main-Regular in `src/fontMetricsData.js`.  Confused why this isn't automatic via `metrics/extract_ttfs.py`.  (Also, can you set a style somehow?)

edemaine:
I'd really like some clarification on this... When testing locally with `make serve`, both `\cdots` and `\ldots` complain with a similar error (but different style):

```
No character metrics for '⋯' in style 'Size1-Regular'
```

As a result, they do not display correctly (their baseline is elevated) on Chrome.

Joe312341:
I am also getting this error in Firefox and Safari

kevinbarabash:
@edemaine it sounds like that's the bug.  That should be pretty easy to fix.

edemaine:
Finally tracked this down.  See #609.

kevinbarabash:
Closed by https://github.com/Khan/KaTeX/pull/618.

Issue #609, opened 2017-01-09
Sqrt is rendered broken for some symbols


spontaliku-softaria:
Enter `\sqrt \Box` on demo page, square root is being rendered without horizontal line and inner element.
Looks like this happens with any symbol which generates "No character metrics for..." warning - it has no metrics and heuristic in sqtr builder makes empty span.

edemaine:
This should be the same issue as #600.

kevinbarabash:
Closed by https://github.com/Khan/KaTeX/pull/618.

Issue #512, opened 2016-07-27
\varnothing should be more rounded than \emptyset


hudakharrufa:
I get that they refer to the same thing but \varnothing should look more rounded, yet they both look exactly the same when rendered. I don't like how they currently look :( Any hope for a fix?

spicyj:
Seems like a bug. Looks like varnothing is correctly defined to use the AMS font in our symbols file, but that isn't making its way to the actual DOM output we generate.

kevinbarabash:
`\varnothing \emptyset` should produce the following output:
![screen shot 2016-07-29 at 2 48 50 pm](https://cloud.githubusercontent.com/assets/1044413/17264408/98cc75ce-559b-11e6-8b60-95bfbfa1abd4.png)

kevinbarabash:
Closed by https://github.com/Khan/KaTeX/pull/618.

Issue #374, opened 2015-10-25
No character metrics for '■' in style 'Main-Regular'


miestasmia:
When using `\blacksquare` Chrome 46 spits out a warning `No character metrics for '■' in style 'Main-Regular'`.

Locercus:
I think this might be part of a larger Unicode support issue that'd need to be resolved. If need be, I can try test it in other browsers as well.

kevinbarabash:
@Locercus I'm not sure what's causing this.  We do have metrics for `\blacksquare` which is `\u25a0` (or 9632 in decimal), see https://github.com/Khan/KaTeX/blob/a81c4fe78df1b19401883cff4f8af2b1c0f5b033/src/fontMetricsData.js#L206.  If could test it in another browser that would be great.  I have a feeling though that this isn't a browser problem.

Locercus:
Can confirm it happens in Firefox 41 as well as Opera 31 too. I'm on GNU+Linux so I can't test IE or Safari.

kevinbarabash:
This is a Linux only bug but it is reproducible.

t-k-:
I also spot this warning, the square box displays correctly. No idea how to suppress this warning.

(my browser: Chromium 52.0.2743.85)

Menci:
`\notin` also trigger the bug.

But it displays correctly.

kevinbarabash:
For `\blacksquare` we're passing the wrong style to `fontMetrics.getCharacterMetrics(value, style);` in `makeSymbol`.  It should be `"AMS-Regular"` but we're passing `"Math-Regular"`... or the metrics should for `\blacksquare` should be in the `"Math-Regular"` table.  I'd need to investigate further to determine which it is.

kevinbarabash:
`\notin` legitimately does not have metrics.  It's in our KaTeX_Math-Regular font files, but it doesn't exist in `mapping.pl`.  A person would need to have a look at https://github.com/Khan/MathJax-dev/blob/master/fonts/OTF/TeX/makeFF (which is used to generate our font files) to see how `\notin` (decimal unicode character 2209) makes its way into the font.

kevinbarabash:
Closed by https://github.com/Khan/KaTeX/pull/618.

upupming:
If I write Chinese in Math code, I will get this warning, too:

![image](https://user-images.githubusercontent.com/24741764/50563249-27f18900-0d56-11e9-8b08-c236f8d37237.png)

[...]

PR #618, opened 2017-01-10
Fix all AMS mathord symbols

edemaine:
Here is a fix for #600 (and #609 and #512 and #374).  In fact, all AMS textord symbols were similarly broken.  Here is a complete list of symbols fixed by this PR:

`\hslash,
\triangledown,
\lozenge,
\circledS,
\circledR,
\measuredangle,
\nexists,
\mho,
\Finv,
\Game,
\Bbbk,
\backprime,
\blacktriangle,
\blacktriangledown,
\blacksquare,
\blacklozenge,
\bigstar,
\sphericalangle,
\complement,
\eth,
\diagup,
\diagdown,
\square,
\Box,
\Diamond,
\yen,
\checkmark,
\beth,
\daleth,
\gimel,
\digamma,
\varkappa,
\varnothing,
\maltese`

Basically, in `buildCommon.js`, `mathsym` had logic for choosing between font families `Main-Regular` and `AMS-Regular`, but this applies to types `rel, bin, open, close, inner, punct` only.  `makeOrd` is used specifically for `mathord` and `textord`, and needed similar logic.  To complicate things, `makeOrd` was also losing track of the original macro name and thus AMS vs. Main, replacing it with the unicode symbol.  So I had to refactor code a bit so that `mathDefault` could figure it out.


kevinbarabash:
@edemaine thanks for the PR.  How did you determine the list of broken chars?  Is it just all mathords?

edemaine:
It is all symbols in symbols.js of font "ams" and type "textord".

src/buildCommon.js
```diff
@@ -116,13 +139,19 @@ var mathit = function(value, mode, options, classes) {
...
             // glyphs for \imath and \jmath do not exist in Math-Italic so we
             // need to use Main-Italic instead
-            utils.contains(mainitLetters, value) ||
-            utils.contains(greekCapitals, value)) {
-        return makeSymbol(
-            value, "Main-Italic", mode, options, classes.concat(["mainit"]));
+            utils.contains(mainitLetters, value)) {
+            // For greek capitals, we have a choice, but Math-Italic
+            // seems closer to LaTeX.
+            //|| utils.contains(greekCapitals, value)) {
```
kevinbarabash: remove commented out code

src/buildCommon.js
```diff
@@ -116,13 +139,19 @@ var mathit = function(value, mode, options, classes) {
...
+        };
     } else {
-        return makeSymbol(
-            value, "Math-Italic", mode, options, classes.concat(["mathit"]));
+        return {
+            fontName: "Math-Italic",
+            fontClass: "mathit"
+        };
     }
 };
```
kevinbarabash: Can you update the doc string for `mathit` now that it no longer creates a symbol?

kevinbarabash:
@edemaine a few minor nits and this'll be ready to merge.  Thanks for fixing all of these symbols.

src/buildCommon.js
edemaine: I meant to ask you about this.  Yes, it seems so:

Greek capital letters are in both fonts, `Main-Italic` and `Math-Italic`.  So which should we use? The [master code](https://github.com/Khan/KaTeX/blob/master/src/buildCommon.js#L120) claims to use `Main-Italic` for Greek capital letters.  But if you try the [demo](https://khan.github.io/KaTeX/) with `\mathit{\Omega\imath}`, and inspect, you'll see that the `\Omega` gets class `mathit` while `\imath` gets `mainit`:

```html
<span class="mord mathit" style="margin-right: 0.05017em;">Ω</span>
<span class="mord mainit">ı</span>
```

In fact, the master code ends up calling `mathit()` with a single-unicode-character string (`Ω`), while `greekCapitals` is defined in terms of the macro name (`"\Omega"`).  When my code fixed this, I ended up changing Greek capital letters to use `Main-Italic` font, which caused some tests to fail.  I confirmed with `texcmp` that `Math-Italic` seems slightly closer, so kept it there.

Let me know if you'd like this code changed at all...

src/buildCommon.js
edemaine: Fixed, thanks!

edemaine:
Cool!  I just merged with master and moved to ES2016 style.

src/buildCommon.js
kevinbarabash: In that case, can you remove the commented out code?  TBH, I'm not sure why we have the same glyphs duplicated in different fonts.  It would be nice if we could avoid this and save a few bytes.  I'm curious to find out why the rendering looks different even though the glyphs are the same.  Maybe the advance or skew metrics for the Greek capitals are differ between fonts.  I think it's safe to investigate this separate from this pull request.

edemaine: Should I remove the list of greekCapitals as well?  It's currently commented out (to prevent lint from complaining).

kevinbarabash: Yes please.

edemaine: Done.

kevinbarabash:
@edemaine thanks for cleaning up the commented out code.  ^_^

edemaine:
No problem!  Now you should be able to close #600, #609, #512, and #374!

PR #619, opened 2017-01-10
Implicit \color, explicitly grouped \textcolor

edemaine:
I've implemented `\color` and `\textcolor` to behave as they do in LaTeX (specifically, the `color` package):

```latex
{\color{red}This is red.}
\textcolor{red}{This is also red}
```

Sadly, this differs from current KaTeX and MathJax which would only make the `T` red in the first example (as if `\color` were actually `\textcolor`).  So this change may break some existing code.  But I think it's for the better, as LaTeX should be the master reference here.


kevinbarabash:
I didn't realize there was a `color` package.  It's too bad about the difference in behavior.  I think we want this, we just need to communicate the breaking change along with suggested workarounds.

kevinbarabash:
`color` package reference: http://ctan.mirror.rafal.ca/macros/latex/required/graphics/color.pdf

edemaine:
As I understand it, this is the *only* way to change color in LaTeX.  There are other packages, e.g. xcolor, but they build on the same basic interface.

xymostech:
This is admittedly selfish, but KA has a *ton* of math expressions that would break because of this change. Also, our previous `\color` function was compatible with MathJax, which was the reason we implemented it the way we did.

edemaine:
FWIW, MathJax defines a [color extension](http://docs.mathjax.org/en/latest/tex.html#color) that changes the behavior to match LaTeX.  I suppose KaTeX could do the same... though personally I think it makes more sense to match LaTeX and MathJax+color.

kevinbarabash:
@xymostech we're already preprocessing our math so it shouldn't be hard to replace `\color` with `\textcolor`.

edemaine:
@xymostech Another thing we could do is add an `oldColor` or `colorIsTextColor` option to the options set, to enable backwards compatibility mode...

gagern:
I guess you could also declare a macro `\color` which expands to `\textcolor`. That way you don't have to go poking around in some string literal which might one day contain things like `\verb` and `\def` and whatnot, but could still override `\color` to behave like `\textcolor` on a per-size basis.

Nevertheless, this change should come with a bump to 0.8.0, not in the 0.7 line of releases.

test/screenshotter/ss_data.yaml
```diff
@@ -49,9 +49,10 @@ Cases: |
         a-1&\text{otherwise}
     \end{cases}
 Colors:
-    tex: \blue{a}\color{#0f0}{b}\color{red}{c}
+    tex: \blue{a}\textcolor{#0f0}{b}\textcolor{red}{c}
     nolatex: different syntax and different scope
-ColorSpacing: \color{red}{\displaystyle \int x} + 1
+ColorSpacing: \textcolor{red}{\displaystyle \int x} + 1
+ColorImplicit: \color{red}red\color{blue}blue
```
gagern: I think this test should ensure that a) ending a group restores previous color and b) ending a `\textcolor` does return to the previous implicit color. Something like

```tex
bl{ack\color{red}red\textcolor{green}{green}red\color{blue}blue}black
```

edemaine: Good point, and nice test!  Changed.

edemaine:
@gagern Good idea.  We could suggest defining a macro `'\color': '\textcolor'` in release notes to 0.8.0, assuming it gets approved at all.  I'll just reiterate that I think it's good to use LaTeX as the reference model.  Even if MathJax is the reference model, it offers both options.

gagern:
Code looks good to me. I'd like a go-ahead from @xymostech, though, to make sure the workaround for the incompatibility is feasible for KA.

edemaine:
@xymostech or @kevinbarabash, could we make a decision here?  I see two reasonable options:

1. Leave `\color` as is (two-argument function) because it's compatible with MathJax without any flags.  But provide an option that switches `\color` to behave as in LaTeX, just like MathJax does.
2. Change `\color` to make it compatible with LaTeX (one-argument function changing current group).  But provide an option that switches `\color` to behave as in MathJax (i.e. old behavior) for (somewhat manual) backwards compatibility.

Option 1 has the advantage of backwards compatibility without option changing.  Option 2 is preferable to me because KaTeX's mantra is/should be "renders just like LaTeX".  But it would probably require a significant version increment.

In both cases, we can provide `\textcolor` which behaves just like LaTeX, because that should unambiguously be a two-argument function.

kevinbarabash:
@edemaine let's go with option 2.  Will want to bump the version number to 0.8.x when we do a publish that includes this change.

edemaine:
Cool, thanks, @kevinbarabash !  That's almost what this PR does.  Preference on calling the option `oldColor` or `colorIsTextColor` or something else?

kevinbarabash:
`colorIsTextColor` is more descriptive.

edemaine:
4dd21a7 adds a `colorIsTextColor`, documentation in README, and `parsesLike` tests that confirm the exact behavior of `\color` in the three cases (no specification, and `colorIsTextColor` set to `false` or `true`).

src/functions.js
```diff
@@ -166,6 +166,14 @@ defineFunction("\\color", {
     };
 });
 
+// \color is handled in Parser.js's parseImplicitGroup
```
kevinbarabash: Thanks for commenting this.  It would be nice to have a way to mark functions in this file as to whether or not the parser needs to the parse an implicit group for them.

test/errors-spec.js
```diff
@@ -224,19 +224,19 @@ describe("Parser.expect calls:", function() {
...
                    "Invalid size: '[' at position 11: \\rule{1em}[̲2em]");
         });
         // Can't test for the [ of an optional group since it's optional
         it("complains about missing } for color", function() {
-            expect("\\color{#ffffff{text}").toFailWithParseError(
-                   "Invalid color: '#ffffff{text' at position 8:" +
-                   " \\color{#̲f̲f̲f̲f̲f̲f̲{̲t̲e̲x̲t̲}");
+            expect("\\textcolor{#ffffff{text}").toFailWithParseError(
+                   "Invalid color: '#ffffff{text' at position 12:" +
+                   " \\textcolor{#̲f̲f̲f̲f̲f̲f̲{̲t̲e̲x̲t̲}");
```
kevinbarabash: I don't even know what that character is, lol.

test/katex-spec.js
```diff
@@ -833,10 +838,26 @@ describe("A color parser", function() {
...
+    it("should use one-argument \\color if requested", function() {
+        expect(oldColorExpression).toParseLike("\\textcolor{#fA6}{xy}", {
+            colorIsTextColor: false,
+        });
+    });
+
+    it("should use two-argument \\color if requested", function() {
+        expect(oldColorExpression).toParseLike("\\textcolor{#fA6}{x}y", {
+            colorIsTextColor: true,
+        });
```
kevinbarabash: Nice tests!

test/screenshotter/ss_data.yaml
```diff
@@ -49,9 +49,10 @@ Cases: |
         a-1&\text{otherwise}
     \end{cases}
 Colors:
-    tex: \blue{a}\color{#0f0}{b}\color{red}{c}
+    tex: \blue{a}\textcolor{#0f0}{b}\textcolor{red}{c}
     nolatex: different syntax and different scope
-ColorSpacing: \color{red}{\displaystyle \int x} + 1
+ColorSpacing: \textcolor{red}{\displaystyle \int x} + 1
+ColorImplicit: bl{ack\color{red}red\textcolor{green}{green}red\color{blue}blue}black
```
kevinbarabash: nice

kevinbarabash:
@edemaine thanks for the PR.  Sorry for taking so long to make a decision regarding `\color`'s behavior.

src/functions.js
edemaine: Yes.  @gagern and I had discussed this at some point, and it felt like *all* the implicit group parsing could be moved to `functions.js`.  Worth considering in the future...

edemaine:
No worries -- big decisions are hard!  Thanks for reviewing + merging.

PR #622, opened 2017-01-11
Revert "Remove trailing commas for IE 9 compatibility"

gagern:
This reverts commit 4d2e46e7f6b57a5ea99ae7802809e88065876f9c.

Having trailing commans makes diffs easier to read as it avoids modifying a line just to add a trailing comma if there is another item to add at the end of a list.  There are plans to switch to ES6 notation and to translate that to ES5 as part of the build process (#617).  Since that translation would remove trailing commas, the IE9 problems that originally motivated the commit should vanish soon.


gagern:
@xymostech are you waiting for some other input before merging? Do you want me to merge this myself?

xymostech:
Nope, I was just waiting for you to merge it! (same goes for #623 and #615) I think in general I'll mention if I think somebody else should be looking. :)

Issue #621, opened 2017-01-11
Injecting auto-render code at end of DOM from <head> script


oddhack:
This is a -users question, not a -developers question, but it's not clear to me where to post such questions. Apologies if this is inappropriate, and if you have pointers to a good place to ask, I'd appreciate that.

I'm looking for an example of how to create a script from the document \<head\> that will inject the call to setup and execute the auto-render extension at the end of the document. This would be nice because it's tedious to inject this code when rendering multiple asciidoc source documents to HTML, and it would be simpler to just have a single block of scripts, both loading from the CDN and then setting up the auto-render call, added to the document head, for which there's a good existing asciidoc mechanism. If such an example script exists it would be great to add to the auto-render dir.

gagern:
I don't think it makes sense to auto-insert this snippet at the end. You can only do so after the document has been parsed, and in that case, you might as well run the code directly. So add something like this to your `<head>`:

```html
<script>
    document.addEventListener("DOMContentLoaded", function() {
        renderMathInElement(document.body);
    });
</script>
```

You could insert the script in that handler instead, but as I said, it would be more work for no gain.

oddhack:
@gagern that makes sense, thanks. I am very uneducated in JS and the DOM.

xymostech:
@gagern That looks useful to more people. We should add that to the auto-render docs.

PR #623, opened 2017-01-11
Describe using auto renderer in html head

gagern:
Motivated by https://github.com/Khan/KaTeX/issues/621#issuecomment-271972288.

### DIFF
--- a/contrib/auto-render/README.md
+++ b/contrib/auto-render/README.md
@@ -27,9 +27,9 @@ before the close body tag:
 
 See [index.html](index.html) for an example.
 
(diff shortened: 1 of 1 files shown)


contrib/auto-render/README.md
```diff
@@ -27,6 +27,21 @@ before the close body tag:
 
 See [index.html](index.html) for an example.
 
+If you prefer to have all your setup inside the header,
+you can use the following script there
+(instead of the one above at the end of the document):
```
xymostech: "at the end of `<body>`" maybe?

PR #624, opened 2017-01-11
Modernize font creation

gagern:
* Automate the build process, instead of writing a complicated manual
* Update sources used by the docker image away from google code
* Exclude MathJax-dev checkout from docker image, to allow image reuse
* Try to keep the individual image layers small by combining RUN commands and removing temporary files used only during each individual step

Motivated by https://github.com/Khan/KaTeX/pull/614#issuecomment-271768990.


kevinbarabash:
@gagern cool.  Is there a way to supply a local path for MathJax-dev?  If not, pushing to a remote branch is fine since people will have to do that anyways to create a PR.

gagern:
I was wondering about the best approach for local repos, too. Theoretically it should be possible to run a git server locally, and have the docker connect to that. But seeing all the trouble the screenshotter had to go to in order to identify an ip address to be used for this, I'd rather avoid that. Perhaps we should drop the git checkout altogether, and instead download the tarball, or offer a way to copy a local tarball into the container instead. Then we can have the outer script create such a tarball in case the argument is a local directory, or use it as is if it is a file, or download it if it is a url.

kevinbarabash:
Another option might be to install whatever dependencies MathJax-dev needs to run the `make` scripts correctly during development.

gagern:
I'll leave that to you. Personally I think having both the docker approach and a non-dockerized version would be nice. So I'd say merge this and if one of us finds the time we can write a Makefile to build all dependencies and use that.

dockers/MathJaxFonts/Dockerfile
```diff
@@ -23,39 +26,41 @@ RUN apt-get -qq update && apt-get -qqy install \
...
+ADD https://github.com/yui/yuicompressor/releases/download/v2.4.8/yuicompressor-2.4.8.jar /usr/share/yui-compressor/yui-compressor.jar
 
 # Download batik-ttf2svg.jar
-RUN wget "https://archive.apache.org/dist/xmlgraphics/batik/batik-1.7.zip"
-RUN unzip -qq batik-1.7.zip
-RUN mv batik-1.7/batik-ttf2svg.jar /usr/share/java/
+RUN wget "https://archive.apache.org/dist/xmlgraphics/batik/batik-1.7.zip" \
+ && unzip batik-*.zip batik-*/batik-ttf2svg.jar \
+ && mv batik-*/batik-ttf2svg.jar /usr/share/java/ \
+ && rm -r batik-*
```
kevinbarabash: I'm curious about the benefit of combining multiple commands vs. running them separate?

dockers/MathJaxFonts/Dockerfile
```diff
@@ -23,39 +26,41 @@ RUN apt-get -qq update && apt-get -qqy install \
...
-RUN tar -xzf ttf2eot-0.0.2-2.tar.gz
-RUN sed -i "1s/^/#include <cstddef>/" ttf2eot-0.0.2-2/OpenTypeUtilities.h
-RUN make -C ttf2eot-0.0.2-2/
-RUN mv ttf2eot-0.0.2-2/ttf2eot /usr/bin/
+RUN wget "https://github.com/wget/ttf2eot/archive/v0.0.2-2.tar.gz" -O ttf2eot.tar.gz\
+ && tar -xzf ttf2eot.tar.gz \
+ && sed -i "1s/^/#include <cstddef>/" ttf2eot-*/OpenTypeUtilities.h \
+ && make -C ttf2eot-*/ \
+ && mv ttf2eot-*/ttf2eot /usr/bin/ \
+ && rm -r ttf2eot*
```
kevinbarabash: The `rm` commands are necessary are they?

dockers/MathJaxFonts/Dockerfile
```diff
@@ -23,39 +26,41 @@ RUN apt-get -qq update && apt-get -qqy install \
...
-RUN git clone "https://code.google.com/p/font-compression-reference/" woff2_compress
-RUN make -C woff2_compress/woff2/
-RUN mv woff2_compress/woff2/woff2_compress /usr/bin/
-
-# Download and setup MathJax-dev
-RUN git clone "https://github.com/khan/MathJax-dev.git"
-RUN cp MathJax-dev/default.cfg MathJax-dev/custom.cfg
-RUN make -C MathJax-dev custom.cfg.pl
-
+RUN wget "https://github.com/google/woff2/archive/d9a74803fa884559879e3205cfe6f257a2d85519.tar.gz" -O woff2.tar.gz \
```
kevinbarabash: 👍

dockers/MathJaxFonts/buildFonts.sh
```diff
@@ -0,0 +1,147 @@
...
+                        pushd "$1"
+                        if [[ ! -f fonts/OTF/TeX/Makefile ]]; then
+                            echo "$1 does not look like MathJax-dev" >&2
+                            exit 1
+                        fi
+                        tar cf "$FILE" Makefile default.cfg fonts/OTF/TeX
+                        popd
+                    elif [[ -f "$1" ]]; then
+                        FILE="$1"
+                    elif [[ "$1" = http?(s)://* ]]; then
```
kevinbarabash: I tried running `./buildFonts.sh /local/path/to/MathJax-dev` and got the following error:
```
./buildFonts.sh: line 97: syntax error in conditional expression: unexpected token `('
```

dockers/MathJaxFonts/buildFonts.sh
```diff
@@ -0,0 +1,147 @@
...
+mk=\$(tar tf MathJax-dev.tar.gz | grep 'fonts/OTF/TeX/Makefile\$')
+tar xf MathJax-dev.tar.gz
+cd \"\${mk%fonts/*}\"
+cp default.cfg custom.cfg
+make custom.cfg.pl
+make -C fonts/OTF/TeX ${filetypes[*]}
+tar cf /fonts.tar ${filetypes[*]/#/fonts/OTF/TeX/}"
+
+echo "Creating and starting docker container from image $IMAGE"
+CONTAINER=$(docker create "$IMAGE" /bin/sh -c "${CMDS}")
```
kevinbarabash: So the docker image created by the DockerFile is a base image that you then modify to run the commands to actually create the fonts, is that correct?

kevinbarabash:
Just looking at all the deps, it's going to be a pain to create a cross platform script to install all the deps.  I think I installed all the deps manually the last time I edited a font and it was a pain.  This will be a huge improvement.  Thanks for making it work with local directories.

dockers/MathJaxFonts/Dockerfile
gagern: Each `RUN` command creates a separate layer of docker file system, while commands combined into one `RUN` will just contribute one layer that contains their combined effect. In this case, the resulting layer should *only* contain `/usr/share/java/batik-ttf2svg.jar`, as everything else is created and removed in the same `RUN`. This keeps the number of layers down and the total size of the image small, in order to conserve resources.

dockers/MathJaxFonts/Dockerfile
gagern: Not necessary for the image to perform as intended, no. The goal is again to clean up all the build-time resources once the desired command has been installed, to keep the size of the layer down.

dockers/MathJaxFonts/buildFonts.sh
gagern: Can you please try whether a `shopt -s extglob` at the beginning of the script fixes this?

dockers/MathJaxFonts/buildFonts.sh
gagern: Not exactly. Docker uses an *image* as a template to create *containers*. So in our case the image contains all the tools we need to create a font, while the container then runs the named commands to actually perform the font creation. You can use the same image to build several fonts, instantiating one container each time you do so. The commands are executed once inside the container constructed here, but the file system that results from running these commands is not preserved, contrary to how images work.

dockers/MathJaxFonts/buildFonts.sh
kevinbarabash: I gave that a try, but that gives the same output. :(

gagern: Well, we can make that just cover http for now, or use `[[ "$1" = http://* ]] || [[ "$1" == https://* ]]` if we can't get this glob syntax to work as intended.

kevinbarabash: Sorry... I pasted it in the wrong .sh file.  It's running now.  I'll post back once it completes.

dockers/MathJaxFonts/buildFonts.sh
kevinbarabash: A "container" seems like an "instance" from the description you gave.  This is probably one of the reasons why I've had trouble grokking docker.

dockers/MathJaxFonts/buildFonts.sh
kevinbarabash: It worked!  All of the font files were modified.  Some of those changes might be due to version changes in the deps and/or running in the docker vs. OS/X itself.

gagern: I'll check whether running the tool on my system will modify them again or not.

dockers/MathJaxFonts/buildFonts.sh
gagern: A container is an instance, yes. Container is to image like process is to executable file.

dockers/MathJaxFonts/buildFonts.sh
kevinbarabash: I was running it against `HEAD` on `master` of `khan/MathJax-dev` which has a number commits since the last time the font files were generated.

kevinbarabash: I forget whether I ran the screenshot tests after all the font files were changed.  I want to get this merged so I'm going to run the script again and then run the screenshot tests and if everything passes I'll merge this.

kevinbarabash:
I added `shopt -s extglob` at the start of `buildFonts.sh` as suggested in the comments above.

To test this I checked out https://github.com/Khan/MathJax-dev/commit/677fd6a6167e807bc3dbd635f201e153c108c3d2 and ran 
```
./dockers/MathJaxFonts/buildFonts.sh /absolute/path/toMathJax-dev
```

This resulted in all font files being updated.  To verify that it was only the timestamps being updated I made sure I had a locally copy of https://github.com/Khan/KaTeX/blob/master/.gitattributes and followed its instructions modifying ~/.gitconfig appropriately.

Then I ran `git diff static/fonts/*.ttf` and `git diff static/fonts/*.woff` and saw that only the timestamps had been modified in the TTF and WOFF files.  The output looked like:

```
diff --git a/static/fonts/KaTeX_Size4-Regular.woff b/static/fonts/KaTeX_Size4-Regular.woff
index 7826c6c..eeab3da 100644
--- a/static/fonts/KaTeX_Size4-Regular.woff
+++ b/static/fonts/KaTeX_Size4-Regular.woff
@@ -63,12 +63,12 @@
     <!-- Most of this table will be recalculated by the compiler -->
     <tableVersion value="1.0"/>
     <fontRevision value="1.0"/>
-    <checkSumAdjustment value="0xee9e05b4"/>
+    <checkSumAdjustment value="0xd67232c6"/>
     <magicNumber value="0x5f0f3cf5"/>
     <flags value="00000000 00011111"/>
     <unitsPerEm value="1000"/>
-    <created value="Mon Jun 29 22:52:19 2015"/>
-    <modified value="Mon Jun 29 22:52:19 2015"/>
+    <created value="Sun Aug 20 05:11:16 2017"/>
+    <modified value="Sun Aug 20 05:11:16 2017"/>
     <xMin value="-1903"/>
     <yMin value="-1250"/>
     <xMax value="1902"/>
@@ -2596,7 +2596,7 @@ http://scripts.sil.org/OFL
     <version value="1"/>
     <FFTimeStamp value="Tue Jul 31 14:57:19 2012"/>
     <sourceCreated value="Mon Oct 12 19:21:51 2009"/>
-    <sourceModified value="Mon Jun 29 22:52:00 2015"/>
+    <sourceModified value="Sun Aug 20 05:11:01 2017"/>
   </FFTM>
 
 </ttFont>
```

PR #628, opened 2017-01-15
Webpage: update katex to 0.7.0, show error messages

gagern:
This updates katex to version 0.7.0. As #613 is causing problems when using this in bower, I manually copied the files from the npm package instead.

I also included a bit of script to display the error message to the output area if rendering threw an exception. That should provide easier feedback as to what's wrong with a given input.


gagern:
I just realized that this change would expose the way the displaystyle was done until now, by including a `\displaystyle` in the input string. I've now changed that to a proper `displayMode: true` setting in the invocation, both to avoid confusing error messages and to demonstrate how this is meant to be handled. As a side effect, the exterior display style styling now applies to the demo output. Personally I consider this a good thing, but if you want me to I can get it back to left-aligned and without extra space around.

main.js
```diff
@@ -8,7 +8,20 @@ window.startup = function() {
...
+            });
+        } catch(err) {
+            while(demoOutput.lastChild) {
+                demoOutput.removeChild(demoOutput.lastChild);
+            }
+            var msg = document.createTextNode(err.message);
+            var span = document.createElement("span");
+            span.appendChild(msg);
+            demoOutput.appendChild(span);
+            span.setAttribute("class", "errorMessage");
```
kevinbarabash: Cool!

kevinbarabash:
> Personally I consider this a good thing, but if you want me to I can get it back to left-aligned and without extra space around.

Sounds good.  Thanks for noticing this difference and for updating `gh-pages`.

Issue #587, opened 2016-12-17
Discussion: horizontal metrics


kevinbarabash:
There are a number of features that we'd like to support that require knowing how wide a layout is:
- `\overbrace`, `\underbrace`
- `\cancel`
- `\overrightarrow`, `\overleftarrow`
- `\widehat`, `\widetilde`, etc.

There are a couple of ways that we might figure out how wide a layout might be:
- render the layout once with placeholder phantoms and `visibility: hidden;`, measure DOM nodes, update the layout and set `visibility: visible;`
- require horizontal (advance) metrics for all glyphs to be rendered

My preference is the second option.

The main issue with either of these options `\text` containing glyphs outside of fonts we provide.  There are some options for dealing with this issue:
- provide fonts covering all languages (non starter)
- measure the glyphs for the text in the browser before rendering the math that uses them
- determine metrics of safe fonts for each browser beforehand use the right set of metrics based on the user agent or detection of available fonts

I'd love to hear other people's thoughts on how to implement these features.

gagern:
Computing the width of a complete box, instead of having metrics for individual glyphs and then summing them up, can be useful for situations where the total width of some text is not necessarily the sum of the advances of the individual glyphs. Issues that come to my mind here include kerning pairs, glyph selection based on context (as position in word in [Arabic](https://en.wikipedia.org/wiki/Arabic_script_in_Unicode) or ligatures in western languages), but also hinting and rounding to pixel positions.

On the other hand, using width information instead of computed box size can be of benefit for offline processing, e.g. server-side or during page creation. In that case we might not even know all the details about how the browser will lay out the glyphs. We could try to measure runs of text using additional dependencies like freetype bindings for node, but for common TeX fonts using the built-in metrics is best, and re-using that code path for more exotic fonts reduces the amount of code we need.

Measuring glyphs in browser perhaps should be split into two approaches: either measure sizes of DOM boxes, or use a canvas to measure advance of a string. I guess that particularly for slanted fonts, the box might be wider than the total advance of its content. I'm not sure which of these we actually want, which one is closer to what TeX does. Measuring either of these in browser means waiting till the required font for that glyph is available. Right now I don't think we'll be able to check when that is the case, unless the configuration tells us the names of the fonts that are going to be used, and which glyphs they cover. Then we can use the web font loader or similar to wait for the fonts to load. In the absence of such a configuration, we could create an invisible box and poll its dimensions from some timer, but that feels very ugly.

[...]

PR #632, opened 2017-01-17
Update fonts

kevinbarabash:
The updated fonts include the last few commits to https://github.com/Khan/MathJax-dev/ as well as changes in https://github.com/Khan/MathJax-dev/pull/3.

Left/right single quotes work with the typewriter font.

The fonts also support latin ligatures, e.g. æ, œ, Æ, and Œ.


kevinbarabash:
screen shot 2017-01-16 at 11 17 10 pm

screen shot 2017-01-16 at 11 17 04 pm

gagern:
I ran this again, and couldn't reproduce the font files byte for byte. Looking at the output from `ttfdump` or `ttx`, It seems as though the only relevant change is the modification of the `created` and `modified` entries in the [`head` table](https://www.microsoft.com/typography/otspec/head.htm). Pretty annoying, since this means there is no simple way to see whether re-running the tool introduces any relevant changes, or merely updates the timestamps with no other modification. So what do we want to do?

1. Accept this as a fact of life, trust that all changes will be irrelevant or beneficial, and move on.
2. Manually inspect all differences using `ttfdump` or `ttx` and `diff`.
3. Write some script to `ttx` the files, check whether the timestamps and checksums are the only things that changed, and revert to the base version against which the comparison was performed (e.g. `HEAD`) if there are no other changes.
4. Store the `ttx` files in our repository, and only generate the other formats locally. This means developers will need more tools, and more time. But it also means we see each and every modification in the corresponding source file using the regular GitHub diffs. We could even do merge commits with conflict resolution on these files.
5. Do both: have `ttx` files for readable versions, and other file formats for use by the browsers. Use a Travis check to ensure that these stay in sync: whoever modifies one file must modify the other as well, in order to keep Travis happy.
6. Set up [`textconv` settings in `.gitattributes`](https://www.git-scm.com/docs/gitattributes#_performing_text_diffs_of_binary_files) in such a way that `git diff` shows the diff on the `ttx` file. This will be very helpful for local reviews, but not much use with the GitHub GUI. On the other hand, we want to test each PR locally before accepting, right?

To help us decide, I actually created a branch that contains TTX files. It's not using them yet, and it does have the TTF, too. But it does show the effect of your pull request here in a human readable fasion: https://github.com/gagern/KaTeX/commit/feefc7c0e1b761a9d10740648de4d71747a2001a

That diff shows that pretty much all of your font file modifications are spurious. The only relevant ones are

* [KaTeX_Main-Regular](https://github.com/gagern/KaTeX/commit/feefc7c0e1b761a9d10740648de4d71747a2001a#diff-101803650a2923ccac502aca55f1cbf8)
* [KaTeX_Main-Italic](https://github.com/gagern/KaTeX/commit/feefc7c0e1b761a9d10740648de4d71747a2001a#diff-3cd653f7ad7fde81bc6d766c1774a046)
* [KaTeX_Main-Bold](https://github.com/gagern/KaTeX/commit/feefc7c0e1b761a9d10740648de4d71747a2001a#diff-40ac7709949b07d27dc487c2b6f8cd1d)
* [KaTeX_Typewriter-Regular](https://github.com/gagern/KaTeX/commit/feefc7c0e1b761a9d10740648de4d71747a2001a#diff-e0dbdd4fca4efd07206f2cbbe53bf909)

These are changes that should actually be reviewed, but all the other modifications should in my opinion not be part of this pull request here. Can you drop them please?

I'm also a bit surprised tha [Katex_Main-BoldItalic](https://github.com/gagern/KaTeX/commit/feefc7c0e1b761a9d10740648de4d71747a2001a#diff-644db06d89ad52c511d36ceb11277a39) is *not* affected by this PR here. Is there a reason why the newly added glyphs are available only in three of our four styles?

gagern:
I've had a stab at approach 6 in #635. So far it looks as though we can't fully automate the process, though: users still have to enter some settings in one of their config files. Which allows them to choose different tools, or tools installed in locations outside their `$PATH`, or whatever. Still somewhat inconvenient. Currently it can only handle `ttf` and `woff`. `woff2` requires additional dependencies, and `eot` doesn't seem to be supported at all.

kevinbarabash:
I think 6 is definitely better than what we were doing before which was 1.  3 sounds nice, but it's also more work, but it might be worth doing that work if we think we're going to change the fonts more often which seems entirely likely considering issues around missing spaces, accent characters, Euro symbol, etc.

gagern:
Do we want to hold this merge here till we can get the things from #638 in as well? That would help keep the repo size down, but would delay #614. I guess we shouldn't delay too much if it can be avoided.

kevinbarabash:
@gagern are you concerned that too many updates to the font files will balloon the repo size because the binary diffs are big?

gagern:
Yes. The WOFF fonts in particular are compressed, so a slight change in one part of the font file will likely lead to large changes throughout the file, and therefore bad delta compression of repository data. At 2.2 M the fonts directory is the largest portion of our repository. We should not lightly add 2.2 M a couple of times if we can avoid it.

In the long run, it might make sense to only deploy the TTF files, and generate all the others locally or when rolling a release. Since the TTFs are afaik less compressed, they should delta-compress much better. If I find the time I could investigate whether we can get the machinery for WOFF, WOFF2 and EOT compiled to JavaScript using Emscripten, but that won't happen in the next couple of weeks.

edemaine:
Any way the font metrics data could be updated to include horizontal metrics?  Relevant to #587.

kevinbarabash:
@edemaine we just have to run `make metrics_extended` and commit the output, see https://github.com/Khan/KaTeX/blob/master/Makefile#L112-L113.  I added a `--width` option to `format_json.py` a while back that includes the glyph widths in the output.

edemaine:
Ah, cool.  So that is already part of this PR.  I'm not sure what the optimal timing is, but it sounds like #587 could already use this info.  If we want to continue to hold on this PR, I guess we could make and commit new metrics for the old fonts, before switching to new fonts.

edemaine:
I do think it'd be good to get horizontal metrics in, so that we can start playing with box sizes (though I'm not really sure when I can start on that).  Would it make sense to use a separate repo for fonts, so that we don't worry as much about increasing the size of this one?  Or switch to fewer font formats as suggested by @gagern?

kevinbarabash:
If we had the fonts in the a separate repo, we could set things up to do a shallow checkout of that repo... then it wouldn't matter if the size that repo ballooned.

edemaine:
FWIW, I just read [this Atlassian article](https://www.atlassian.com/blog/git/handle-big-repositories-git) about how to deal with large binary files in a Git repo.  I learned about lots of options, though I'm not sure I have a specific proposal.  Submodules could be a way to do multiple repos, but also increase use complexity...

kevinbarabash:
> If I find the time I could investigate whether we can get the machinery for WOFF, WOFF2 and EOT compiled to JavaScript using Emscripten, but that won't happen in the next couple of weeks.

Another option might be to mount `static/fonts` within the docker at the same place where the `ttf` folder would exist.  We should be able to run `make -C fonts/OTF/TeX eot woff woff2` to build the other font files.

edemaine:
Are you suggesting using a docker to build eot, woff, woff2 files?  I think that's fine as a way for people to update the fonts when the fonts change, but I don't think it should be required for users not playing with the fonts.  I.e. doesn't fix the "font changes influencing repo size" issue.  But maybe that wasn't your point.

Back to the repo size issue, I feel like the shallow checkout of another repo dedicated to fonts seems reasonable.  Will this affect npm, bower, CDN?

kevinbarabash:
I'm putting this on hold until we can move our fonts to a separate repo.

kevinbarabash:
Blocked on #816.

kevinbarabash:
Now that #891 has landed, this is unblocked.

kevinbarabash:
These changes will need to be made to https://github.com/KaTeX/katex-fonts.

kevinbarabash:
Now that the fonts are in a separate repo, all we need are the changes in symbols.js and mappings.pl.  I'm going to open a new PR b/c I think rebasing this will be more trouble than it's worth.  I'll also add commands for the ligatures, e.g. `\ae`, `\oe`, etc.

PR #635, opened 2017-01-19
Set up a diff attribute and textconv instructions for font files

gagern:
This may help us generate human-readable diffs of font files.

Motivated by https://github.com/Khan/KaTeX/pull/632#issuecomment-273822102.


kevinbarabash:
I gave this a try.  It works like a charm.  It's too bad that github doesn't have a way to run custom diff hooks.

Issue #634, opened 2017-01-19
x'^2 throws exception


gagern:
KaTeX refuses to render `x'^2` even though TeX does render this. It says

> KaTeX parse error: Double superscript at position 3: x'^̲2

[Here](https://github.com/Khan/KaTeX/blob/v0.7.0/src/Parser.js#L357-L358) is the code which treats pimes as superscripts. So apparently the font has something to do with is, since the symbol `\prime` seems to be styled such that it has to be in superscript style and position in order to work correctly.

LaTeX has the following lists for `$x'^2+x^\prime$`:

```tex
\mathord
.\fam1 x
^\mathord
^.\fam2 0
^\mathord
^.\fam0 2
\mathbin
.\fam0 +
\mathord
.\fam1 x
^\fam2 0
```

So I'd say handling this as a double superscript appears to be the correct way indeed, and we only have to get rid of the error message and do something reasonable for this input.

edemaine:
Agreed.  It's curious because TeX *doesn't* support double superscripts (`$x^a^a$` is invalid), presumably so that ambiguities like `$x^a_a_b^b^c^d_d$` don't need to be resolved.  Also, `$x^a'$` produces a double-superscript error.

Ha, check out the definition of `'`: it checks via `\futurelet` whether the next symbol is `^`, and merges the superscripts in that case.  Also checks for following `'`s which KaTeX already does.

```latex
\def\active@math@prime{^\bgroup\prim@s}
{\catcode`\'=\active \global\let'\active@math@prime}
\def\prim@s{%
  \prime\futurelet\@let@token\pr@m@s}
\def\pr@m@s{%
  \ifx'\@let@token
    \expandafter\pr@@@s
  \else
    \ifx^\@let@token
      \expandafter\expandafter\expandafter\pr@@@t
    \else
      \egroup
    \fi
  \fi}
\def\pr@@@s#1{\prim@s}
\def\pr@@@t#1#2{#2\egroup}
```

So we just need to modify the prime handling in `Parser.js` to check for `^`s in `nextToken` as well.  Once we hit `^`, can call `handleSupSubscript`, and then prepend the `\prime`s to the superscript.

edemaine:
Fixed in #636.

gagern:
Fixed by #636. Thanks for the investigation!

apowers313:
I downloaded the 0.7.1 release, which seems to be from two days after this issue was closed, but I'm still getting the problem reported here:

[...]

PR #636, opened 2017-01-19
Fix x'^2

edemaine:
Fix for #634.  Basically, `'` should be treated like a superscript constructor, where any number of `'`s become `\prime`s in the superscript, and an optional following `^...` can define the rest of the superscript.  As a result, `x^2'` does *not* work (double superscript), and nor does `x'_2'` (double superscript).  All error behaviors now correctly match LaTeX.  (`x'_2'` was accepted by KaTeX previously, but that was incorrect.)


gagern:
Great! Very quick response, nice code, and it even does the same thing as LaTeX for input like `$$\sum'^x$$` or `$x'^{x'^{x'^{x'}}}$`. The failing `1'_2^3` was not something I would have expected before your explanation, but it does make sense now and it is consistent with LaTeX, too. I'll accept this PR.

PR #637, opened 2017-01-19
Implement $...$ via styling node

edemaine:
This is a fix for #5: supporting `$...$` within `\text`.  For example, `x+\text{some $y+\text{some $z$}$}` works.  

I tested in LaTeX, and it turns out that `$$` doesn't mean anything special within `\text`, i.e., it starts and immediately ends math mode; so `\text{$$foo$$}` is equivalent to `\text{foo}` or `\text{foo$$}`.  So I didn't look for `$$`.  Perhaps I should look for `\( ... \)` and/or `\[ ... \]`, but this is also a good start.

Basically, I implemented `$` as an implicit group, and used the end-token argument for `parseExpression` to stop when hitting the closing `$`.  I returned a `styling`  (equivalent of `\textstyle`) node, and changed `mode` to `math`, which seems to force a return to math fonts.  Does this seem like the best approach?  We could instead return a `math` node, but I think it would be treated identically, at least for HTML.  In MathML, I couldn't figure out whether anything like `mrow` or `mstyle` can be nested within `mtext`... maybe it's fine?


gagern:
I wonder whether this can be made a proper function in `functions.js`. The function handler does have access to the parser, so it should be able to do essentially the same things you are doing here, without need for a special case inside the Parser.

You shouldn't assign to `mode` directly, but instead use [`switchMode`](https://github.com/Khan/KaTeX/blob/v0.7.0/src/Parser.js#L99-L103). It takes care of the fact that math mode discards whitespace while text mode retains them. As it is, `\text{a$ b$c}` will throw an exception since space is a valid symbol only in text mode. `\text{a$b$ c}` works as intended because the closing `$` gets consumed after the mode was changed back. You might move the initial `consume()` after the mode switch, to avoid using `switchMode`. But that wouldn't work if you make this a function as suggested above.

> In MathML, I couldn't figure out whether anything like `mrow` or `mstyle` can be nested within `mtext`...

Reading [the grammar](https://www.w3.org/TR/MathML3/appendixa.html#parsing_mtext) I'd say no. It can contain text, and it can contain two kinds of elements but neither of these elements can have any nested content, just attributes. So perhaps the correct approach would be to analyze the children of a run of text, and if some of it is not literally text, then we don't build a single `mtext` but instead build an `mrow` with `mtext` and other items as children.

edemaine:
I agree that `$` could be put into functions... but the same could be said of `sizeFuncs`, `styleFuncs`, and presumably even `\begin` and `\left` (all currently in `parseImplicitGroup`).  I think the idea is that the `numArgs` model of arguments in `functions.js` doesn't "fit" these implicit groups so, even though `numArgs: 0` ought to do the (zero) preparation we need for these functions to do the appropriate magic to the parser... it seemed like it should belong there?  Perhaps @xymostech could comment on preferred style?  (It looks like she wrote `parseImplicitGroup` initially.)  Is there a reason not to add an `implicit` flag or something to `defineFunction` support?...

Thanks for pointing out `switchMode` -- I missed that, as I was modeling after https://github.com/Khan/KaTeX/blob/master/src/Parser.js#L654-L670 (for no particular reason).  I wonder whether that manual assignment to `this.mode` is correct?  Anyway, I've switched to `switchMode` in `$` handling, and now various spacing around the `$` works.  (I had caught `\text{a$b$ c}` but missed `\text{a$ b$c}`!)

I haven't touched MathML support yet, but your plan sounds like a good one.

gagern:
You are right, most of the things in `parseImplictGroup` would fit in `functions.js` just as well as `$` if not even better, since the start with a `\`. At the time @xymostech wrote this code, having functions do this work was infeasible because a function execution couldn't modify the current position in the input string, it couldn't consume input. That became possible only with #386. That's not to say she would have done anything different, though.

I think there are two ways to look at it. One way is to say that a function (in the sense of `functions.js`) has all its commands parsed before any of them are executed. Then you're right, and `$` should stay in the parser with the others you mentioned. Another way would be putting all the special handling in response to a given token into `functions.js` in order to keep the Parser lean. In that case all the things you mentioned should go there. It's probably best to lease `$` where it is for now, and if we decide to move things, then we can move all of them in a separate PR.

> Is there a reason not to add an `implicit` flag or something to `defineFunction` support?

What would that flag accomplish? I don't see a need for a flag at all.

> I wonder whether that manual assignment to `this.mode` is correct?

It does the mode switch before consuming the brace both on opening and closing.

I just realized that the interaction between `switchMode` and token consumation is nontrivial as well. If I were to replace the manual assignments in there with `switchMode` calls, I'd have to do so *after* consuming the respective token. Because if I check that the next token is `{` and then switch to text mode, I might end up reading a space instead! Perhaps I should add a comment somewhere, indicating proper use of that function…

Your code as of your latest commit should be OK, as it does switch modes after consuming the token that triggered the mode switch.

kevinbarabash:
From the spec:
> An mtext element is used to represent arbitrary text that should be rendered as itself. In general, the mtext element is intended to denote commentary text.

The only thing other than text that's shown inside a `mtext` are comments.  This leads me to believe that the best approach would be to render `x+\text{some $y+\text{some $z$}$}` as the following MathML:
```
<mrow>
    <mi>x</mi>
    <mo>+</mo>
    <mtext>some </mtext>
    <mrow>
        <mi>y</mi>
        <mo>+</mo>
        <mtext>some </mtext>
        <mrow>
            <mi>z</mi>
        </mrow>
    </mrow>
</mrow>
```

gagern:
Actually I'd render `x+\text{some $y+\text{some $z$}$}` as

```xml
<mi>x</mi>
<mo>+</mo>
<mrow>
    <mtext>some </mtext>
    <mi>y</mi>
    <mo>+</mo>
    <mrow>
        <mtext>some </mtext>
        <mi>z</mi>
    </mrow>
</mrow>
```

That way, each `\text` command results in either a naked `<mtext>` or it results in an `<mrow>` containing `<mtext>` and other things.

edemaine:
I finally got around to revising this.  Rebased to master, and reimplemented the MathML outputter so that `x+\text{some $y+\text{some $z$}$}` now renders as:

```xml
<mrow>
  <mi>x</mi>
  <mo>+</mo>
  <mrow>
    <mtext>some </mtext>
    <mstyle scriptlevel="0" displaystyle="false">
      <mi>y</mi>
      <mo>+</mo>
      <mrow>
        <mtext>some </mtext>
        <mstyle scriptlevel="0" displaystyle="false">
          <mi>z</mi>
        </mstyle>
      </mrow>
    </mstyle>
  </mrow>
</mrow>
```

while `x+\text{yes}` renders as:

```xml
<mrow>
  <mi>x</mi>
  <mo>+</mo>
  <mtext>yes</mtext>
</mrow>
```

Code is relatively clean.  Time for another review!

edemaine:
P.S. Perhaps it would make sense for `styling` nodes in MathML to check whether the style has actually changed -- and use `<mstyle>` if so, and `<mrow>` otherwise?

kevinbarabash:
I was thinking it'd be nice to have to snapshot tests for our MathML output.  The existing tests are fragile, tedious to write, and hard to make sense of.

edemaine:
@kevinbarabash The MathML tests do seem to have revealed a bug in the new code.  I believe indeed that `<mi>Ω</mi>` should be `<mi mathvariant="normal">Ω</mi>`.  Now that I see that "[normal](https://developer.mozilla.org/en-US/docs/Web/MathML/Element/mi)" means "roman not italic" (for single characters), I understand that part of the old code better.  Will fix...

Error we're looking at: `Expected '<span class="katex-mathml"><math><semantics><mrow><mi>A</mi><mi>x</mi><mn>2</mn><mi>k</mi><mi>ω</mi><mi>Ω</mi><mi>ı</mi><mo>+</mo></mrow><annotation encoding="application/x-tex">Ax2k\omega\Omega\imath+</annotation></semantics></math></span>' to contain '<mi mathvariant="normal">Ω</mi>'

edemaine:
@kevinbarabash But now I see they're not perfect.  Here's an example:

```
1) A MathML font tree-builder should render \mathbb{Ax2k\omega\Omega\imath+} with the correct mathvariants
  Message:
    Expected '<span class="katex-mathml"><math><semantics><mrow><mrow><mi mathvariant="double-struck">A</mi><mi>x</mi><mn>2</mn><mi mathvariant="double-struck">k</mi><mi>ω</mi><mi mathvariant="normal">Ω</mi><mi>ı</mi><mo>+</mo></mrow></mrow><annotation encoding="application/x-tex">\mathbb{Ax2k\omega\Omega\imath+}</annotation></semantics></math></span>' to contain '<mn mathvariant="normal">2</mn>'.
```

But that MathML looks good to me -- `<mn>` has a default `math-variant` of `normal` so it doesn't need to be set in this case.  Easy enough to fix, though -- they key is to check that it *doesn't* contain `<mn mathvariant="double-struck">`.

edemaine:
I changed the `mathvariant` generation code to only emit when it doesn't result in the default output.  I had to change the tests accordingly, but this generally seems preferable.  Before `\mathit{A}` and `A` generated different output; now they both generate `<mrow><mi>A</mi></mrow>`.  (But `\mathit{\Omega}` and `\Omega` still generate different output.)

src/buildMathML.js
```diff
@@ -63,13 +63,19 @@ const getVariant = function(group, options) {
  */
 const groupTypes = {};
 
+const defaultVariant = {
+  "mi": "italic",
+  "mn": "normal",
+  "mtext": "normal",
+};
```
kevinbarabash: This should be 4 space indent, this is why travis-ci is failing at the moment.

edemaine: Oops, thanks.

src/buildMathML.js
```diff
@@ -149,11 +150,25 @@ groupTypes.ordgroup = function(group, options) {
...
+        const group = buildGroup(body[i], options);
+        if (group.type === 'mtext' && currentText != null) {
+            Array.prototype.push.apply(currentText.children, group.children);
+        } else {
+            inner.push(group);
+            if (group.type === 'mtext') {
+                currentText = group;
+            }
+        }
+    }
```
edemaine: This loop combines multiple adjacent `<mtext>` tags into a single `<mtext>` tag.  I'll add a comment to this effect.

test/screenshotter/ss_data.yaml
```diff
@@ -170,6 +170,7 @@ Symbols1: |
     \maltese\degree\pounds\$
     \text{\maltese\degree}
 Text: \frac{a}{b}\text{c~ {ab} \ e}+fg
+TextWithMath: \text{for $a < b$ and $ c < d $}.
```
kevinbarabash: Since you've added a screenshot test you'll also have to generate the image files using `make screenshots` and commit those as well.

edemaine: Sorry, yeah -- I can only run the screenshotter docker on a different machine, and pushing was the easiest way.  I'm rebasing now to merge in the images.

src/Parser.js
```diff
@@ -507,6 +507,20 @@ Parser.prototype.parseImplicitGroup = function() {
...
+        this.consume();
+        const outerMode = this.mode;
+        this.switchMode("math");
+        const body = this.parseExpression(false, "$");
+        this.expect("$", true);
+        this.switchMode(outerMode);
+        return new ParseNode("styling", {
+            style: "text",
+            value: body,
+        }, "math");
```
edemaine: Added a couple of unit tests, including on the MathML generation side, as well as a screenshot test (which also tests spacing).

edemaine:
I just looked this over and all comments were addressed.  As I recall, there are no outstanding issues.  Shall we merge?

kevinbarabash:
@edemaine thanks for pinging me about this and thanks for the PR.

edemaine:
@kevinbarabash Thanks for the review!

yanana:
When is this feature released?

kevinbarabash:
Soonish.  I've been putting off doing a release, but I really need to do one b/c there's lots of great stuff sitting on `master`.  I'll try to do one this Friday.

PR #639, opened 2017-01-20
Show compressed sizes correctly

gagern:
On my macOS 10.12 all the sizes used to show as zero.  Having them in a single shell invocation should avoid that problem in a portable fashion.


kevinbarabash:
I checked it and it works as advertised.  Running `make compress` on master also produces all zeros and I'm on 10.11.  Thanks for fixing this.

Issue #420, opened 2015-12-22
Speed test doesn't work


astoilkov:
It fails with:

```
Uncaught ReferenceError: MathJax is not defined.

message: Uncaught ReferenceError: MathJax is not defined
fileName: https://jsperf.com/katex-vs-mathjax
lineNumber: 129
```

edemaine:
Just bumped into this as well. The problem is this code in the speed test:

```html
<link href="http://xymostech.com/katex/katex.min.css" rel="stylesheet" type="text/css">
<link href="http://xymostech.com/katex/fonts/fonts.css" rel="stylesheet" type="text/css">
<script src="http://xymostech.com/katex/katex.min.js"></script>
```

It would work if it could be changed to:

```html
<link href="https://cdnjs.cloudflare.com/ajax/libs/KaTeX/0.6.0/katex.min.css" rel="stylesheet" type="text/css">
<link href="https://cdnjs.cloudflare.com/ajax/libs/KaTeX/0.6.0/fonts.css" rel="stylesheet" type="text/css">
<script src="https://cdnjs.cloudflare.com/ajax/libs/KaTeX/0.6.0/katex.min.js"></script>
```

kevinbarabash:
Looks like someone made the proposed changes in https://jsperf.com/katex-vs-mathjax/47.  It's still not working, but it looks like an unrelated issue in the setup code.

oddhack:
Perhaps this link could be removed from the README until it's working - I just ran into this as well (also giving a "ReferenceError: katex is not defined" as well as the MathJax error).

mbourne:
&lt;shamelessplug&gt; 
  In the meantime, here's my KaTeX-MathJax speed comparison: http://www.intmath.com/cg5/katex-mathjax-comparison.php 
&lt;/shamelessplug&gt;

kevinbarabash:
@mbourne would you be okay with us pointing to your speed comparison in the README?

edemaine:
[...]

PR #640, opened 2017-01-20
Switch speed test in README

edemaine:
This switches the linked performance test from http://jsperf.com/katex-vs-mathjax/ (which is broken) to http://www.intmath.com/cg5/katex-mathjax-comparison.php as offered in #420.  This seems preferable at least until we can get a working jsperf.com test again, maybe ideally one maintained by a maintainer of KaTeX.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
 
 KaTeX is a fast, easy-to-use JavaScript library for TeX math rendering on the web.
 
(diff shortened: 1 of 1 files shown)


gagern:
I think we should have a comparison on the https://khan.github.io/KaTeX/ website, as something more visual than jsperf would be of benefit.

edemaine:
Agreed.  A nice plot of actual jsperf data, say, would be good.  This PR is merely "best we can do with least effort".

PR #642, opened 2017-01-21
Fix release script after experience from 0.7.1 release

gagern:
This is a follow-up to #615. The 0.7.1 release had some quirks, and failed after the `npm publish` when checking out the original branch. I'm glad I did use the two-argument invocation, otherwise I might not have noticed some of these issues. I manually performed the required steps after that, and adapted the script here to avoid these issues.

* Also edit `dist/README.md`. Otherwise `npm publish` will overwrite that *after* the dist directory has been added, causing the git checkout to fail.  And it's the right thing to do anyway, having *all* the READMEs edited.
* Add all the modified READMEs so they get committed correctly.
* Improve two references to `master` which are no longer accurate.
* Check for uncommitted changes just before creating the tag.
* Encourage always specifying the next version, as discussed in https://github.com/Khan/KaTeX/pull/615#discussion_r97208770

### DIFF
--- a/release.sh
+++ b/release.sh
@@ -139,9 +139,9 @@ if [ ! -z "$NEXT_VERSION" ]; then
 
     # Refer to the just-released version in the documentation of the
     # development branch, too.  Most people will read docs on master.
-    node update-sri.js "${VERSION}" README.md contrib/*/README.md
+    git checkout "v${VERSION}" -- README.md contrib/*/README.md
 
-    git add package.json bower.json README.md contrib/*/README.md
+    git add package.json bower.json
     git commit -n -m "Bump $BRANCH to v$NEXT_VERSION-pre"
     git push origin "$BRANCH"
 
--- a/release.sh
+++ b/release.sh
@@ -24,9 +24,12 @@ usage() {
     echo ""
     echo "Examples:"
     echo " When releasing a new point release:"
-    echo "   ./release.sh 0.6.3"
+    echo "   ./release.sh 0.6.3 0.6.4"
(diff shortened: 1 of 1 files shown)


kevinbarabash:
@gagern thanks for doing the first bug fix deploy and thanks for updating the script.  I'll review it tomorrow...this time before merging ^_^.

release.sh
```diff
@@ -114,19 +117,20 @@ rm -f package.json.bak
...
 # Update the version number in CDN URLs included in the README files,
 # and regenerate the Subresource Integrity hash for these files.
-node update-sri.js "${VERSION}" README.md contrib/*/README.md
+node update-sri.js "${VERSION}" README.md contrib/*/README.md dist/README.md
 
 # Make the commit and tag, and push them.
-git add package.json bower.json
+git add package.json bower.json README.md contrib/*/README.md dist/README.md
 git commit -n -m "v$VERSION"
+git diff --stat --exit-status # check for uncommitted changes
```
kevinbarabash: Does the script exit prematurely if `git diff --stat --exit-status` is non-zero?

gagern: That's the idea, yes. We have an `set -e -o pipefail` in line 3 to achieve that. One benefit is that since this is a `git` command invocation, it will be printed instead of executed in dry run mode, so we can continue there even if the check for uncommitted changes up front (the one with `command git`) fails.

kevinbarabash: And that doesn't happen in the dry run because we're overriding `git` and `npm` to echo.  Cool!

PR #649, opened 2017-01-26
Use utils.deflt for Settings

tabatkins:
Rather than duplicate the definition of the undefined-defaulter, use the version already present in utils.

Not sure why the `macros` line uses a different pattern, but I've left it alone. Happy to amend to make it match if desired.

### DIFF
--- a/src/Settings.js
+++ b/src/Settings.js
@@ -3,12 +3,7 @@
  * default settings.
  */
 
-/**
- * Helper function for getting a default value if the value is undefined
- */
-function get(option, defaultValue) {
-    return option === undefined ? defaultValue : option;
-}
+var utils = require("./utils");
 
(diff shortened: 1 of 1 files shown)


tabatkins:
I copied the import statement from one of the other files, but the linter complained about the use of `var`, so I switched it to `const`. Happy to switch back to stay consistent if someone else can fix the linter, or :+1: to me ignoring it.

gagern:
I guess you copied the line using `var` from an older version of the code base. We switched to `const` and `let` in bd9db332d2887d884ec7ab3ca68488314b8618da from #617.

tabatkins:
Ah, yes, I forked on Dec 2nd and haven't re-upped yet. Thanks for the merge!

PR #656, opened 2017-02-10
Support macros in auto-render

alexpearce:
This PR adds support for [the `macros` option](https://github.com/Khan/KaTeX/blob/f3df1ccbac4c61efd55bec423bc8509f8d173fb2/README.md#rendering-options) to the [auto-renderer](https://github.com/Khan/KaTeX/tree/f3df1ccbac4c61efd55bec423bc8509f8d173fb2/contrib/auto-render). There is at least [some demand](http://stackoverflow.com/questions/41980446) for this feature.

The implementation propagates a `macros` object member to the eventual call to `katex.render`. Example usage:

```javascript
var renderedContainer = document.getElementById('foobar');
const katexOpts = {
    macros: {
        "\\R": "\\mathbb{R}",
    },
};
window.renderMathInElement(renderedContainer, katexOpts);
```

I've not added a test for this. The [existing tests](https://github.com/Khan/KaTeX/blob/f3df1ccbac4c61efd55bec423bc8509f8d173fb2/contrib/auto-render/auto-render-spec.js) for auto-render don't deal with the arguments to `renderMathInElement`, so I think I'd need to add a lot to have a test for this PR. Still, I can try and give it a go if someone is willing to give me some pointers 😄

### DIFF
--- a/contrib/auto-render/README.md
+++ b/contrib/auto-render/README.md
@@ -55,7 +55,9 @@ function renderMathInElement(elem, options)
 `elem` is an HTML DOM element. The function will recursively search for text
 nodes inside this element and render the math in them.
 
-`options` is an optional object argument with the following keys:
+`options` is an optional object argument that can have the same keys as [the 
+object passed to `katex.render`](https://github.com/Khan/KaTeX/#rendering-options),
+in addition to two auto-render-specific keys:
 
 - `delimiters`: This is a list of delimiters to look for math. Each delimiter
   has three properties:
@@ -78,3 +80,7 @@ nodes inside this element and render the math in them.
 - `ignoredTags`: This is a list of DOM node types to ignore when recursing
   through. The default value is
   `["script", "noscript", "style", "textarea", "pre", "code"]`.
+
+Note that the `displayMode` property of the options object is ignored, and is 
+instead taken from the `display` key of the corresponding entry in the 
+`delimiters` key.
--- a/contrib/auto-render/auto-render.js
+++ b/contrib/auto-render/auto-render.js
@@ -14,8 +14,11 @@ const splitWithDelimiters = function(text, delimiters) {
     return data;
 };
 
-const renderMathInText = function(text, delimiters) {
-    const data = splitWithDelimiters(text, delimiters);
+/* Note: optionsCopy is mutated by this method. If it is ever exposed in the
+ * API, we should copy it before mutating.
+ */
+const renderMathInText = function(text, optionsCopy) {
+    const data = splitWithDelimiters(text, optionsCopy.delimiters);
 
     const fragment = document.createDocumentFragment();
 
@@ -25,10 +28,11 @@ const renderMathInText = function(text, delimiters) {
         } else {
             const span = document.createElement("span");
             const math = data[i].data;
+            // Override any display mode defined in the settings with that
+            // defined by the text itself
+            optionsCopy.displayMode = data[i].display;
             try {
-                katex.render(math, span, {
-                    displayMode: data[i].display,
-                });
+                katex.render(math, span, optionsCopy);
             } catch (e) {
                 if (!(e instanceof katex.ParseError)) {
                     throw e;
@@ -48,28 +52,28 @@ const renderMathInText = function(text, delimiters) {
     return fragment;
 };
 
-const renderElem = function(elem, delimiters, ignoredTags) {
+const renderElem = function(elem, optionsCopy) {
     for (let i = 0; i < elem.childNodes.length; i++) {
         const childNode = elem.childNodes[i];
         if (childNode.nodeType === 3) {
             // Text node
-            const frag = renderMathInText(childNode.textContent, delimiters);
+            const frag = renderMathInText(childNode.textContent, optionsCopy);
             i += frag.childNodes.length - 1;
             elem.replaceChild(frag, childNode);
         } else if (childNode.nodeType === 1) {
             // Element node
-            const shouldRender = ignoredTags.indexOf(
+            const shouldRender = optionsCopy.ignoredTags.indexOf(
                 childNode.nodeName.toLowerCase()) === -1;
 
             if (shouldRender) {
-                renderElem(childNode, delimiters, ignoredTags);
+                renderElem(childNode, optionsCopy);
             }
         }
         // Otherwise, it's something else, and ignore it.
     }
 };
 
-const defaultOptions = {
+const defaultAutoRenderOptions = {
     delimiters: [
         {left: "$$", right: "$$", display: true},
         {left: "\\[", right: "\\]", display: true},
@@ -83,30 +87,14 @@ const defaultOptions = {
     ],
 };
 
-const extend = function(obj) {
-    // Adapted from underscore.js' `_.extend`. See LICENSE.txt for license.
-    let source;
-    let prop;
-    const length = arguments.length;
-    for (let i = 1; i < length; i++) {
-        source = arguments[i];
-        for (prop in source) {
-            if (Object.prototype.hasOwnProperty.call(source, prop)) {
-                obj[prop] = source[prop];
-            }
-        }
-    }
-    return obj;
-};
-
 const renderMathInElement = function(elem, options) {
     if (!elem) {
         throw new Error("No element provided to render");
     }
 
-    options = extend({}, defaultOptions, options);
+    const optionsCopy = Object.assign({}, defaultAutoRenderOptions, options);
 
-    renderElem(elem, options.delimiters, options.ignoredTags);
+    renderElem(elem, optionsCopy);
 };
 
 module.exports = renderMathInElement;


alexpearce:
[clabot:check]

gagern:
I'm wondering whether it would be better to hand through arbitrary options, like `throwOnError` or `errorColor`. That would make new options usable in the future without any additional work. `displayMode` would be the only option set internally, either on the given `options` object, or on a copy thereof. What do you think?

alexpearce:
I like the idea. It was my first thought after seeing that `renderMathInElement` only passes _two_ properties of the `options` argument on to other methods. It's makes more sense to me to just pass the whole `options` object along the method chain (that is: `renderMathInElement` → `renderElem` → `renderMathInText`).

The downside, I suppose, it that each method then needs to assume more about the structure of `options`. But we could declare that `options` should have the same structure as the [`Settings` object](https://github.com/Khan/KaTeX/blob/f3df1ccbac4c61efd55bec423bc8509f8d173fb2/src/Settings.js#L15-L22), plus the auto-render-specific `delimiters` and `ignoredTags` properties, to keep the semantics similar to [`katex.render`](https://github.com/Khan/KaTeX/blob/f3df1ccbac4c61efd55bec423bc8509f8d173fb2/katex.js#L21-L30).

kevinbarabash:
There could also be two separate options objects, one that's passed directly to `katex.render` and the other that affects the behavior of the auto-renderer itself.

alexpearce:
> There could also be two separate options objects…

This is also a nice idea!

There are two possible implementations then:

1. Extend the `Settings` class as `AutoRenderSettings`, adding the `delimiters` and `ignoredTags` properties and inheriting everything else. This `AutoRenderSettings` will then be passed along the chain, and then a version stripped of the auto-render-specific properties is passed to `katex.render`.
2. Create a standalone `AutoRenderSettings` to hold the `delimiters` and `ignoredTags` properties (constructed just like `Settings`), and then pass both an `AutoRenderSettings` instance and a `Settings` instance along the method chain. The `Settings` instance will be passed to `katex.render`.

Option 2 does sounds cleaner to me, so I propose going with that one. The API of `renderMathInElement` will change to:

```js
const renderMathInElement = function(elem, autoRenderOptions, renderOptions) { … }
```

I think this is backwards compatible, as callers only supplying an `autoRenderOptions` (matching today's API) will just mean `renderOptions` is `undefined`, and so the default settings will be used when the `Settings` object is initialised.

contrib/auto-render/auto-render.js
```diff
@@ -2,6 +2,8 @@
 /* global katex */
 
 const splitAtDelimiters = require("./splitAtDelimiters");
+const Settings = require("./../../src/Settings");
+const utils = require("./../../src/utils");
```
alexpearce: I'm not very familiar with ES2015. Is this the correct way to import these things from `katex`? Perhaps there's something like `require('katex.utils')` instead?

alexpearce:
I've implemented @kevinbarabash's suggestion.

I guess I should updated the [auto-render README](https://github.com/Khan/KaTeX/blob/f3df1ccbac4c61efd55bec423bc8509f8d173fb2/contrib/auto-render/README.md) to include the new argument?

contrib/auto-render/auto-render.js
gagern: Hmmm… I'd consider these files to be internal to KaTeX' codebase. Now we have a dependency here, which means that the contributed auto renderer may need to be adjusted when and if the files you reference get modified. It also means duplicating some code, as browserify would include those files into both its outputs. So I'd check very carefully whether that's worth the cost.

You're including `utils` for [`utils.dflt`](https://github.com/alexpearce/KaTeX/blob/43ac3d3987a7c8c73baa059303cd5a83655099a6/src/utils.js#L34-L39). The main benefit of that it that it falls back to the default whenever the argument is `undefined`. But the values expected here are arrays. Which means that a simple `||` might work just as well. If you consider `util.dflt` I would rather copy&paste that one function than having the inter-dependency between code bases.

I'd advise against a [`Settings`](https://github.com/alexpearce/KaTeX/blob/43ac3d3987a7c8c73baa059303cd5a83655099a6/src/Settings.js) object. That class is used internally, but the external interface is a simple object (oten a literal) called `options`. I see no benefit in using a `Settings` object in the public interface, and a ton of problems if we ever decide to add stuff to `Settings`.

Just my 2¢.

kevinbarabash:
@alexpearce requesting two changes:
- remove `Settings` and `util` dependencies
- update auto-render README

alexpearce:
Sorry for the very long delay 😅 

I've tried to implement the requested changes.

Gungrave223:
Can I make a suggestion to also support throwOnError and errorColor as well since you're already adding support for macros 😅

alexpearce:
Although the PR started as adding support only for macros, now anything accepted by`katex.render` is accepted by the auto-renderer. If you could test that the options you want work in this PR @Gungrave223, that'd be great!

contrib/auto-render/auto-render.js
```diff
@@ -25,10 +25,11 @@ const renderMathInText = function(text, delimiters) {
         } else {
             const span = document.createElement("span");
             const math = data[i].data;
+            // Override any display mode defined in the settings with that
+            // defined by the text itself
+            renderOptions.displayMode = data[i].display;
```
gagern: The fact that you are modifying the agument passed to this function might lead to ugly surprises in some rare situations. Using `extend` to create a clone with one modification would have been a nice alternative, so I wonder whether it might be better to keep that function around.

contrib/auto-render/auto-render.js
```diff
@@ -48,12 +49,14 @@ const renderMathInText = function(text, delimiters) {
...
 
-const renderElem = function(elem, delimiters, ignoredTags) {
+const renderElem = function(elem, delimiters, ignoredTags, renderOptions) {
     for (let i = 0; i < elem.childNodes.length; i++) {
         const childNode = elem.childNodes[i];
         if (childNode.nodeType === 3) {
             // Text node
-            const frag = renderMathInText(childNode.textContent, delimiters);
+            const frag = renderMathInText(
+              childNode.textContent, delimiters, renderOptions
```
gagern: Nit prick: we have 4 spaces of indentation everywhere else, so please conform to that. In some places we have the closing `)` on the same line as the last argument, but that is already inconsistent.

contrib/auto-render/auto-render.js
edemaine: As someone who has been bitten by this in the past (called function modifying callee's object), I agree.

alexpearce:
Thank for the feedback @gagern! I pushed a couple of commits to address your comments.

Gungrave223:
@alexpearce sorry just getting back to this...your solution does in fact work without an issue...my only question would be ... is there a reason for 2 separate objects? would it not be simplier to simply combine autoRenderOptions and renderOptions?

alexpearce:
We could have only one object, assuming the `delimiters` and `ignoredTags` keys would never conflict with the renderer settings keys.

I don't have a preference whether to require one or two objects in the API, so I'll let the KaTeX developers decide 😊

edemaine:
I would lean toward one object -- much easier to use as a caller.  I'm pretty sure we can avoid naming conflicts, and I could imagine future scenarios where it'd actually be useful to have the same option configuring both the auto-rendered and the main code (e.g., options to toggle/reroute warnings or errors).

alexpearce:
Thanks for the feedback @edemaine. I've changed the API to accept a single configuration object.

edemaine:
@alexpearce It looks like you're still modifying the given `renderOptions` argument before duplicating it (see @gagern's comment).

A new comment: Do you think `ignoredTags` should be part of the `renderOptions` argument?  It seems odd to have both arguments now.  To keep backwards compatibility, we could detect when `renderOptions` is an `Array` and treat it as a spec for `ignoredTags`.

alexpearce:
@kevinbarabash @edemaine about modifying `renderOptions`: I actually thought the comment from @gagern was referring to the object passed to `renderMathInElement`, which is [now cloned](https://github.com/Khan/KaTeX/pull/656/files#diff-b34825135eb977dc3f05a6e4b5cf6944R101). From then on everything is done in internal functions (only `renderMathInElement` is exposed), so it should be safe to modify as we 'own' it from then on.

If you still think it's worth doing then I can add it. (I would probably use `Object.assign` instead of re-adding `extend`.)

> A new comment: Do you think ignoredTags should be part of the renderOptions argument? It seems odd to have both arguments now. To keep backwards compatibility, we could detect when renderOptions is an Array and treat it as a spec for ignoredTags

I'm sorry but I'm not sure what you mean 😊 What are you suggesting, to modifying the `renderElem` signature?

contrib/auto-render/auto-render.js
```diff
@@ -83,30 +85,26 @@ const defaultOptions = {
...
+    this.ignoredTags = options.ignoredTags || defaultOptions.ignoredTags;
+}
 
 const renderMathInElement = function(elem, options) {
     if (!elem) {
         throw new Error("No element provided to render");
     }
 
-    options = extend({}, defaultOptions, options);
+    const autoRenderOptions = new AutoRenderSettings(options);
```
edemaine: This line will add keys to the passed-in `options` object.  It needs to go *after* the next line that makes the copy.

contrib/auto-render/auto-render.js
edemaine: Ah, I see now -- the copy is done by the caller.  Maybe rename `renderOptions` to `optionsCopy` and add a comment to note that it's modified within the function, in case this API is ever exposed.

contrib/auto-render/auto-render.js
```diff
@@ -83,30 +85,26 @@ const defaultOptions = {
...
+    const renderOptionsCopy = Object.assign({}, options);
 
-    renderElem(elem, options.delimiters, options.ignoredTags);
+    renderElem(
+        elem,
+        autoRenderOptions.delimiters,
+        autoRenderOptions.ignoredTags,
```
edemaine: I don't see why these two values within the options object need to be passed in separately.  Why not just pass in `renderOptionsCopy`?

contrib/auto-render/auto-render.js
```diff
@@ -62,7 +64,7 @@ const renderElem = function(elem, delimiters, ignoredTags) {
                 childNode.nodeName.toLowerCase()) === -1;
 
             if (shouldRender) {
-                renderElem(childNode, delimiters, ignoredTags);
+                renderElem(childNode, delimiters, ignoredTags, renderOptions);
```
edemaine: Same here: we can just pass in the options argument (which would probably be better named `options` or `optionsCopy`?) instead of explicitly calling out `delimeters` and `ignoredTags`.

contrib/auto-render/auto-render.js
edemaine: Also, perhaps there should be an `if (data[i].display !== null)` enclosing this code?  That way, an unspecified `display` value lets you use the caller's value.  Ah, but then the options copying would have to be done here, so maybe not worth it...

contrib/auto-render/README.md
```diff
@@ -55,7 +55,9 @@ function renderMathInElement(elem, options)
 `elem` is an HTML DOM element. The function will recursively search for text
 nodes inside this element and render the math in them.
 
-`options` is an optional object argument with the following keys:
+`options` is an optional object argument that can have the same keys as [the object 
+passed to `katex.render`](../../README.md##automatic-rendering-of-math-on-a-page),
```
edemaine: Maybe should mention, though, that `displayMode` isn't meaningful in this case, because it's value instead gets defined by `delimiters`.

edemaine:
@alexpearce Sorry, on the last read through, I missed that you were calling `Object.assign` at the outer level.  That's probably enough, unless we want to allow *not* specifying a `display` value for one of the delimiters.  (See review.)  Unclear that would be useful, though.

Yes, I am suggesting changing the function signatures on the child functions to just one options, instead of manually passing around the delimiters and ignoredTags, which are already in the options object.

edemaine:
@alexpearce Do you have some time to implement the comments?  Particularly important that the `renderOptions` get copied before being passed into `AutoRenderSettings`.  I could do the refactoring of `delimiters` and `ignoreTags` as a separate PR if it's not clear what I'm asking for.

contrib/auto-render/README.md
alexpearce: :+1: [added](https://github.com/Khan/KaTeX/pull/656/files#diff-df21e6d07c94a780602626ac9c3f604dR84)

contrib/auto-render/auto-render.js
alexpearce: I don't think that's true because `new AutoRenderSettings` will return a new object.

alexpearce:
Thanks for the reminder @edemaine. I've done the refactoring, rebased, and squashed the commits.

contrib/auto-render/auto-render.js
edemaine: Ah, you're right.  I'm a little confused by the logic now, though, after the refactor.  `autoRenderOptions` gets up to two of the options in `options`, then gets remerged back into `options` to make `optionsCopy`.  So probably this step isn't necessary, and we don't really need `AutoRenderSettings` anymore?  Unless I'm missing something...

contrib/auto-render/auto-render.js
```diff
@@ -14,8 +14,11 @@ const splitWithDelimiters = function(text, delimiters) {
     return data;
 };
 
-const renderMathInText = function(text, delimiters) {
-    const data = splitWithDelimiters(text, delimiters);
+/* Note: optionsCopy is mutated by this method. If it is ever exposed in the
+ * API, we should copy it before mutating.
```
edemaine: Nice comment!

contrib/auto-render/README.md
edemaine: Nice!

edemaine: This README link is broken (two `##`s, and I'm not sure you can really link to a `.md` file like this). I also think it's be more appropriate to link to `https://github.com/Khan/KaTeX/#rendering-options`

alexpearce: Relative links _should_ work on GitHub, but the absolute URL is clearer, so I'll switch to that.

contrib/auto-render/auto-render.js
alexpearce: I think this logic might work instead:

```js
const optionsCopy = Object.assign({}, defaultOptions, options);
```

[`defaultOptions`](https://github.com/alexpearce/KaTeX/blob/8a11755ac12899816c30f73e13d952f740d74184/contrib/auto-render/auto-render.js#L76-L88) is merged first (probably should rename this to `defaultAutoRenderOptions`) and then the user `options` is merged. So if `options` contains the same keys as `defaultOptions`, the former will override the latter. Then indeed there's no need for `AutoRenderSettings`.

What do you think?

edemaine: Sounds good to me!

contrib/auto-render/README.md
edemaine: In hindsight, I think you're right. The hashtag part needs to change, but fine to go with absolute or relative.

alexpearce:
I've implemented the remaining comments. Now this PR removes more lines than it adds 😄

edemaine:
Looks great!  I assume you've tested this latest code in active use?  (It looks like we have automated tests of `splitATDelimiters` but not `auto-render` in general.)

edemaine:
I noticed there's a sample `index.html` file (linked to in the documentation), and ran that on this branch, and all works well.

Incidentally, I had to make the following tweak to make this code run in-place.  In particular, because `auto-render.js` uses `require`, it can't run directly in a browser -- you need to run the compiled version.  Should we add this tweak too?  I can add a PR.

```patch
diff --git a/contrib/auto-render/index.html b/contrib/auto-render/index.html
index 566c147..e0291c4 100644
--- a/contrib/auto-render/index.html
+++ b/contrib/auto-render/index.html
@@ -3,9 +3,9 @@
   <head>
     <meta charset="UTF-8">
-    <script src="/katex.js" type="text/javascript"></script>
-    <link href="/katex.css" rel="stylesheet" type="text/css">
-    <script src="./auto-render.js" type="text/javascript"></script>
+    <script src="../../dist/katex.js" type="text/javascript"></script>
+    <link href="../../dist/katex.css" rel="stylesheet" type="text/css">
+    <script src="../../dist/contrib/auto-render.min.js" type="text/javascript"></script>
     <style type="text/css">
       body {
           margin: 0px;
```

alexpearce:
Good catch! Thanks a lot for the extra PR and, of course, for the review.

Issue #629, opened 2017-01-15
\vec problem in Chrome


ronkok:
The accent `\vec` will not appear in Chrome on a page whose CSS includes: `body {text-rendering: optimizeLegibility;}`

All the other accents work fine and `\vec` works fine in all the browsers that I've tested, other than Chrome. Odd.

kevinbarabash:
Verified.  Thanks @ronkok.

ronkok:
This issue was resolved by PR #670.

Issue #529, opened 2016-09-13
Arrow on vector doesn't display


jbg77:
How can I fix it, I using the 6.0.

Thank in advance!

kevinbarabash:
@jbg77 can you post the TeX code you're seeing the issue with?

jbg77:
It's `\vec{f}`

I'm using phantomJs, with the doctype: [https://github.com/Khan/KaTeX/issues/514](#514)

kevinbarabash:
Can you add some debugging to your phantomjs script to printout which resources it's downloading.  See http://phantomjs.org/troubleshooting.html.

edemaine:
Should be fixed by #670. Please reopen if you're still having trouble.

PR #670, opened 2017-03-28
Support stretchy wide elements.

ronkok:
This will fix issue #629 and possibly #529.  Currently, Chrome will fail to render the arrow in `\vec` when CSS contains `text-rendering:
optimizeLegibility;`.


ronkok:
I have resolved all the lint errors. The error messages I'm getting now are, to me, cryptic. Any help would be appreciated.

sthoch:
Cool. I feel like the arrows should be a tad thicker, but other than my gut, I have no reference atm.

The errors from here: https://travis-ci.org/Khan/KaTeX/builds/218513949 don't seem cryptic to me. What errors are you refering to?

ronkok:
Yes, I've looked the link you have provided. Yesterday, I cleaned up all the lint errors. The errors that remain start at line 858 of https://travis-ci.org/Khan/KaTeX/builds/218513949#L858. The typical remaining error says something such as:

npm ERR! Linux 4.4.0-51-generic

And though some may find that statement to be very informative, I confess that I find it cryptic.

edemaine:
Can you push your fixes to the lint errors, which will trigger a rerun of travis?  Those `npm ERR` messages are just a summary saying that lint failed, and should disappear once the lint errors are gone.

ronkok:
Well, @edemaine, I've made another commit, titled  "Fix lint errors". I hope that qualifies as the "push" that you alluded to. If not, let me know. And I appreciate the help.

ronkok:
Well, that's progress. The lint corrections have passed. Travis now fails on Sreenshotter.

I have had no success running Screenshotter locally. If anyone else would like to try, that would be fine with me.

edemaine:
Hmm.  I can run screenshotter, but I couldn't reproduce the error that Travis is getting on your branch.  My experience is that screenshotter occasionally fails randomly (on my second try, the Colors test failed), so perhaps if Travis tried again, it would work.  (But I don't know how to get Travis to try again, other than another random commit.)  Anyway, probably don't need to worry about this automatic check, and instead an actual maintainer should review this PR.

src/stretchy.js
```diff
@@ -0,0 +1,274 @@
...
+    } else if (!options.color) {
+        node.style.backgroundImage = "url(" + katexImgPath() +
+            fileName + ".svg)";
+        // By writing that image path inline, we eliminate ~40 CSS classes.
+        if (height + depth !== 0.34) {
+            node.style.height = (height + depth) + "em";
+        }
+    } else {
+        // Add color. Use CSS mask, if possible
+        if (maskIsSupported()) {
```
glebm: Currently this will not emit the mask when rendering server-side. It'd be nice to emit the mask, or both, when rendering server-side.

ronkok: When I run server.js and view it in http://localhost:7936/, I get the mask when I view it in Chrome, the background-image when I view it in Firefox, and the inline background-image when I view it in Edge. That's the same behavior I get when I view the [web page](http://copyasmarkup.com/katex/wideelements.html).

I completely agree that we want a server side rendering to emit the mask, but my experience is that we already get that. Am I doing something wrong?

edemaine: The examples you describe both involve a web browser, a.k.a. a client.  The point is that, if you run KaTeX from Node, it won't "support" mask, and so won't be omitted.  You simply can't detect support for anything in a server-only environment.  Can't you emit the mask in all cases, and it will just work when it's supported?

glebm: You could do something like:

```js
function maskIsSupported() {
  // If rendering server-side, assume the client supports CSS masks,
  // as most clients do:
  if (typeof document === "undefined") return true;
  ...
}
```

ronkok: @edemaine , I can omit the mask in all cases, but the non-mask solution relies on a XMLHttpRequest, which is slow.

@glebm, I would agree with your suggestion. If I interpret [caniuse](http://caniuse.com/#search=mask) correctly, the upcoming version of Firefox will change from partial `mask` support to full support. That would leave behind only IE and Edge. However, I'm not sure that the KaTeX owners would agree with you and me. If Edge does look at the output from such a server-side installation, it will render a span-wide blob of color instead of the subject matter.

So I thank both of you for the education on server-side behavior. This affects not just `\color` but also `\cancel`, another case where I have different code for IE and Edge.

Regarding `\cancel`, I'm going to use background-image SVGs for server-side rendering, just as I use it now for IE and Edge. The only drawback of the SVGs in most browsers is that the line does not get thicker when giant fonts sizes are employed. A 1 px line looks good in most common font sizes, so that is a good fall-back.

On the subject of `\color`, I need some guidance from the KaTeX owners as to what they will accept. The options are:
1. Do not support `\color`.
2. Support `\color` by a (slow) XMLHttpRequest in all cases.
3.  Use `mask` when it is supported. Use `mask` for server-side installations even
though Edge will not render it properly.
4.  Use `mask` when I can use feature detection to verify that it is supported. Use a XMLHttpRequest  for all server-side installations.

edemaine: If there isn't a good one-size-fits-all solution, it seems like the best thing to do would be to add an option that lets you decide between "work on Edge" vs. "work fast", when you can't autodetect it.

I'm a little confused why you need an XMLHttpRequest, though.  Is it not worth just inlining the SVG into the JavaScript code (const mySVG = "...") and using that as needed?  I remember we discussed this before, but why use external files at all, when you can embed it all in the JavaScript source?

ronkok: That's a very good point. Since our conversation, I've added some SVG files. The tactic of making composite images via CSS was giving some poor rendering issues. So the SVG code is now 32 KB. Say about 40 KB once I add the URL escape codes. By inlining that into the JavaScript, we could eliminate all XMLHttpRequest's. So now we can add option:

5. Add 40 KB of SVG code inline in the JavaScript.  No XMLHttpRequest, ever.

I still need some guidance from KaTeX owners. Which option would you prefer?

static/images/doubleleftarrow.svg
```diff
@@ -0,0 +1,43 @@
+<svg xmlns="http://www.w3.org/2000/svg">
```
glebm: These SVGs should be minified, ideally at build time. I recommend svgo for this, can easily shave off 30%.

ronkok: Agreed.

ronkok:
Here's a thought: If I inline the SVG code in the JavaScript (to enable server-side `\color`), that opens up another possibility.  I could then write all the SVGs inline, always. That would eliminate the entire /images/ folder and all its SVG files.  And KaTeX would never make a HttpRequest for an image file. So it would be faster. The downside is a lot more bloat in the HTML.

Comments?

src/stretchy.js
glebm: @ronkok 
> If Edge does look at the output from such a server-side installation, it will render a span-wide blob of color instead of the subject matter.

Note that it is possible to always render black for browsers that do not support it with something like:

```css
@supports not (mask-image: #000) {
  .stretchy {
    background-color: transparent !important;
  }
  .x-arrow {
    background-image: url(...) !important;
  }
  ...
}
```

Unfortunately, not supported on IE.

glebm:
> So it would be faster.
> So it would be faster

Not necessarily, as the browser would need to parse and render a separate SVG image for every arrow from scratch. The giant HTML payload would also make server-side rendered downloads much slower.

I'd much prefer to always use CSS masks and render black arrows for browsers that do not support them (via `@supports not` + IE hacks). The SVGs can be inlined into the CSS at build time to avoid XHR requests.

ronkok:
@edemaine,  I've created a local version with all the SVG code inline. It increases the JavaScript size by 20K, which is less than I expected. I took a pretty extreme approach to eliminating code duplication. For instance, I wrote the typical opening SVG tag only once and I wrote the path for a rightarrow only once. Then, for any specific image, I append these strings together on the fly to create the final complete SVG code. It's pretty compact code, but I suspect it would be hard to maintain and extend.

@glebm, If were to summarize your thoughts in my own words, it would go like this:
1.  Move feature detection and/or browser detection from the JavaScript to the CSS. That enables detection in server-side rendering.
2.  Use CSS `mask` when supported and render black images when `mask` is not supported. That makes inline SVG code unnecessary.

That would complicate the CSS pretty substantially, but it is in every other way very attractive. I'll work up  a prototype. I'm leaning towards this approach myself.

ronkok:
Well, I have a problem with the `@supports` approach.  Edge lies and acts as though it supports `mask` when it actually does not support it. I still like the idea of CSS feature detection. I'll see if I can find some hack.

ronkok:
@glebm, I now have `@supports` partly working. It properly controls the background-image and the mask-image in  Chrome, Edge, and Firefox. Now I need to write an inline style to set the `background-color`. The JS is `node.style.backgroundColor = options.color;`  It has to be inline because I don't know at build time what color the user will want.

How does one use `@supports` to control whether or not an *inline* style gets rendered?

glebm:
@ronkok, I believe the `!important` qualifier in the property values inside `@supports not` would override the inline styles.

static/images/doubleleftarrow.svg
```diff
@@ -0,0 +1,43 @@
...
+181.3 323 142.3 300 91 274
+93.7 272 97 270.3 101 269
+137.7 253 173.3 231.7 208 205
+208 205 218 197 218 197
+L400000 197
+400000 157 Z"/>
+<path fill="black" d="M270 159 h399460 L399460 196  L270 196 z"/>
+<path fill="black" d="M270 351 h399460 L399460 392 L270 392 z"/>
+</svg>
+</svg>
```
kevinbarabash: extra `</svg>`?

static/images/doubleleftarrow.svg
```diff
@@ -0,0 +1,43 @@
+<svg xmlns="http://www.w3.org/2000/svg">
+<svg viewBox="0 0 400000 549" preserveAspectRatio="xMinYMin slice">
```
kevinbarabash: Why the extra `<svg>`?  Couldn't `viewBox` and `preserveAspectRatio` be set on the outer `<svg>`?

kevinbarabash:
@ronkok awesome PR.  I haven't done a close reading of the changes yet, but we'll definitely want some screenshot tests for the new functionality.

static/images/doubleleftarrow.svg
ronkok: Some of the SVGs are nested, so that the inner SVG can keep an undisturbed aspect ratio. For instance, we want an arrow head to be un-distorted when the arrow gets longer or shorter. In these files, each arrow SVG contains an arrow that is 400 ems long in the inner SVG. The outer SVG is set to match the dimensions defined for it by the CSS. The outer SVG clips the inner SVG; it does not stretch the inner SVG.

Of course, we do want some of the SVGs, like `\cancel`, to vary their aspect ratio. Those files do not get the nested SVG treatment.

ronkok:
@kevinbarabash, Regarding screenshots, I agree. I also have some work underway that improves the sever-side rendering, as suggested by @glebm. I'm busy on other things for a couple of days, but sometime next week I'll get you some screenshots and a squashed PR.

ronkok:
@kevinbarabash, I've updated PR #670 with my latest work, improving server-side rendering and minfying the SVG files. I've also squashed the PR. It's ready for your review whenever you find it convenient. You'll find that the JavaScript and the HTML are simpler, but the CSS is more complex.

I've also added some Jasmine tests. I tried to add Screenshotter tests, but failed. I don't plan to try again. On the other hand, there is more than one way to contribute. If you want to assign some Screenshotter work to someone at KA, I would be happy to make a financial contribution that leaves KA whole. The Screenshotter code that I tried to add is:

```
Boxed: \boxed{F=ma} \quad \boxed{ac}\color{magenta}{\boxed{F}}\boxed{F=mg}

ExtensibleArrows: |
    \begin{array}{l}
        \xrightarrow[ab]{ABC} + \xRightarrow{ABC} + \xrightleftharpoons[ab]{ABC} + \xhookrightarrow[ab]{ABC} \\
        \xtwoheadrightarrow{ABC} + \frac{\xrightarrow[ab]{ABC}}{\xrightarrow[ab]{ABC}} + \left\lvert\xrightarrow[ab]{ABC}\right\rvert
    \end{array}

HorizontalBraces: \overbrace{\displaystyle{\oint_S{\vec E\cdot\hat n\,\mathrm d a}}}^\text{emf} = \underbrace{\frac{q_{\text{enc}}}{\varepsilon_0}}_{\text{charge}}
HorizontalBrackets: |
    \begin{array}{l}
        \overbracket{(2x+3)}^\text{factor 1}\,\underbracket{(4x+5)}_\text{factor 2} \\
        \overbracket[0.15em][1.2em]{10\rightarrow 9}^{\text{high}} \rightarrow \overbracket[0.05em]{2\rightarrow 1}^{\text{low}}
    \end{array}

LowerAccent: |
    \begin{matrix}
        \underleftarrow{AB} \quad \underrightarrow{AB} \quad \underleftrightarrow{AB} \quad \undergroup{AB} \\
        \underlinesegment{AB} \quad \undertilde{AB}  \quad \color{green}{\underrightarrow{AB}} \\
        \underrightarrow{F} + \underrightarrow{AB} + \underrightarrow{AB}^2 + \underrightarrow{AB}_2 \\
        \frac{\underrightarrow{AB}}{\underrightarrow{AB}} + \sqrt{\underrightarrow{AB}} + \left\lvert\underrightarrow{AB}\right\rvert
    \end{matrix}

StretchyAccent: |
    \begin{array}{l}
        \overrightarrow{AB} \quad \overleftarrow{AB} \quad \Overrightarrow{AB} \quad \overleftrightarrow{AB} \\
        \overgroup{AB} \quad \overlinesegment{AB} \quad \overleftharpoon{AB} \quad \overrightharpoon{AB} \quad \color{red}{\overrightarrow{AB}} \\
        \widehat{\theta} \quad \widehat{AB} \quad \widehat{ABC} \quad \widetilde{A} \quad \widetilde{AB} \quad \widetilde{ABC} \\
        \overrightarrow{F} + \overrightarrow{AB} + \overrightarrow{F}^2 + \overrightarrow{F}_2 + \overrightarrow{F}_1^2 \\
        \overrightarrow{AB}^2+\frac{\overrightarrow{AB}}{\overrightarrow{AB}} + \sqrt{\overrightarrow{AB}} + \left\lvert\overrightarrow{AB}\right\rvert + \left(\overrightarrow{AB}\right)
    \end{array}
StrikeThrough: |
    \begin{array}{l}
        \cancel x \quad \cancel{2B} + \bcancel 5 +\bcancel{5ay} \\
        \sout{5ab} + \sout{5ABC} + \xcancel{\oint_S{\vec E\cdot\hat n\,\mathrm d a}} \\
        \frac{x+\cancel B}{x+\cancel x} + \frac{x+\cancel y}{x} + \cancel{B}_1^2 + \cancel{B^2} + \left\lvert\cancel{ac}\right\rvert
    \end{array}
```

kevinbarabash:
@ronkok I'm sorry that `make screenshots` didn't work for you.  I'll check out your branch later this week when I have some time.

ronkok:
As before, I have a [web page](http://copyasmarkup.com/katex/wideelements.html) up that describes the results of PR #670. I’ve spent some time with BrowserStack, and the page now includes accurate [browser support](http://copyasmarkup.com/katex/wideelements.html#browser-support) data.

edemaine:
@ronkok It looks like you did a merge with master?  There's now a lot of irrelevant commits in this PR.  You might try `git rebase -i b9e7b6898fd68a575d953000f4741ce2a24012f0` (or some other way to specify the master commit) and follow the instructions (carefully) for interactive rebasing.

ronkok:
Yes, I needed to fix a bug -- extensible arrows should be REL, not ORD. So my attempt was to make the change, then squash all the work so that KaTeX maintainers could review it. As a new GitHub user, I have two comments:

1. GitHub’s ability to merge, diff, etc., is impressive.
2. It’s user interface is vile.

Obviously, things went wrong. I’ll try to follow your suggestion.

ronkok:
I believe that the latest commit has fixed my earlier error. @edemaine, thank you for the help.

edemaine:
@ronkok Yep, looks good now!

ronkok:
@kevinbarabash,  Two items:

1. @glebm has found a `\color` [method](http://stackoverflow.com/questions/42966641/how-to-transform-black-000-into-a-given-color-using-only-css-filters/43959853#43959853) that may render these background SVGs with the proper color in Edge. I’ll work it up tomorrow.
   
2. I was thinking about future support for the `mhchem` package and I noticed that it
   uses 7 different extensible arrows as reaction arrows. Four of them have an exact
   match in this pull request (although `mhchem` calls each by a different name). I am
   thinking that I might create SVGs for the other three and include them with names
   similar to the `mhchem` names. So that would add `\xyieldsLeftRight`,
    `\xequilibriumLeft`, and `\xequilibriumRight` to the list of extensible arrows.

mhchem-arrows
If you think that arrow idea is a bad one, let me know.

ronkok:
Update:

1.  Unfortunately, the `\color` method alluded to just above will not work in Edge. I’ll
    stick with `\color` via CSS `mask`, which is supported by all browsers except
    IE/Edge. There’s a fallback in black.
2.  It turns out that `mhchem` applies a min-width of 2em to its reaction arrows, which
   is longer than an AMS-LaTeX or `mathtools` extensible arrow (). So, to
   support a future `mhchem` extension, I propose to add seven new extensible arrows.
   Each one will correspond to a `mhchem` reaction arrow: `\xyields`, `\xyieldsLeft`,
   `\xyieldsLeftRight`, `\xmesomerism`, `\xequilibrium`, `\xequilibriumLeft`, and
   `\xequilibriumRight`. What do people think of this idea? Should we omit these functions
    from the documentation and reserve them for `mhchem`?

kevinbarabash:
@ronkok for (2) let's leave them for a second PR as this PR is already pretty big.  I'm going to try to get it review this weekend.

src/buildMathML.js
```diff
@@ -471,6 +493,65 @@ groupTypes.underline = function(group, options) {
     return node;
 };
 
+groupTypes.accentunder = function(group, options) {
```
kevinbarabash: `accentUnder`

src/buildMathML.js
```diff
@@ -471,6 +493,65 @@ groupTypes.underline = function(group, options) {
...
+    const accentNode = stretchy.mathMLnode(group.value.label);
+    const node = new mathMLTree.MathNode(
+        "munder",
+        [buildGroup(group.value.body, options), accentNode]
+    );
+    node.setAttribute("accentunder", "true");
```
kevinbarabash: TIL :)

src/buildMathML.js
```diff
@@ -471,6 +493,65 @@ groupTypes.underline = function(group, options) {
...
+    return node;
+};
+
+groupTypes.strikeThru = function(group, options) {
+    const node = new mathMLTree.MathNode(
+        "memclose", [buildGroup(group.value.body, options)]);
```
kevinbarabash: `"menclose"`

src/buildMathML.js
```diff
@@ -471,6 +493,65 @@ groupTypes.underline = function(group, options) {
...
+        case "\\bcancel":
+            notation = "downdiagonalstrike";
+            break;
+        case "\\sout":
+            notation = "horizontalstrike";
+            break;
+        default:
+            notation = "updiagonalstrike";
+    }
+    node.setAttribute("notation", notation);
```
kevinbarabash: Cool!  https://developer.mozilla.org/en-US/docs/Web/MathML/Element/menclose

src/buildMathML.js
```diff
@@ -471,6 +493,65 @@ groupTypes.underline = function(group, options) {
...
+groupTypes.xArrow = function(group, options) {
+    const arrowNode = stretchy.mathMLnode(group.value.label);
+    const upperNode = buildGroup(group.value.body, options);
+    let node;
+
+    if (group.value.below) {
+        const lowerNode = buildGroup(group.value.below, options);
+        node = new mathMLTree.MathNode(
+            "munderover", [arrowNode, lowerNode, upperNode]
+        );
```
kevinbarabash: Give `\xleftarrow[123]{}` would `upperNode` be empty?  In that case maybe we should use `munder`?

src/functions.js
```diff
@@ -626,19 +626,117 @@ defineFunction([
 defineFunction([
     "\\acute", "\\grave", "\\ddot", "\\tilde", "\\bar", "\\breve",
     "\\check", "\\hat", "\\vec", "\\dot",
-    // We don't support expanding accents yet
-    // "\\widetilde", "\\widehat"
+    "\\widehat", "\\widetilde", "\\overrightarrow", "\\overleftarrow",
+    "\\Overrightarrow", "\\overleftrightarrow", "\\overgroup",
+    "\\overlinesegment", "\\overleftharpoon", "\\overrightharpoon",
```
kevinbarabash: Should `\Overrightarrow` be `\overrightarrow`?

src/functions.js
```diff
@@ -626,19 +626,117 @@ defineFunction([
...
     const base = args[0];
+
+    const isStretchy = !utils.contains([
+        "\\acute", "\\grave", "\\ddot", "\\tilde", "\\bar", "\\breve",
+        "\\check", "\\hat", "\\vec", "\\dot",
+    ], context.funcName);
+
+    const isShifty = !isStretchy || utils.contains([
+        "\\widehat", "\\widetilde", "\\overrightarrow", "\\overrightharpoon",
+    ], context.funcName);
```
kevinbarabash: Shouldn't all of the `\\over...` commands be stretchy?

src/functions.js
```diff
@@ -626,19 +626,117 @@ defineFunction([
...
+defineFunction([
+    "\\overbrace", "\\overbracket", "\\underbrace", "\\underbracket",
+], {
+    numArgs: 1,
+    numOptionalArgs: 2,
+    argTypes: ["size", "size", "original"],
+}, function(context, args) {
+    const thickness = args[0];
+    const height = args[1];
+    const base = args[2];
```
kevinbarabash: I see we're being inconsistent with `base` and `body`.  We should clean that up at some point.  Don't worry about it in this PR since it's big enough as is.

src/functions.js
```diff
@@ -626,19 +626,117 @@ defineFunction([
...
+    return {
+        type: "horizBrace",
+        label: context.funcName,
+        isOver: (context.funcName.substr(1, 2) === "ov"),
```
kevinbarabash: `/^\\over/.test(context.funcName)` might be a bit clearer, feel free to ignore.

src/functions.js
```diff
@@ -626,19 +626,117 @@ defineFunction([
...
+// Stretchy accents under the body
+defineFunction([
+    "\\underleftarrow", "\\underrightarrow", "\\underleftrightarrow",
+    "\\undergroup", "\\underlinesegment", "\\undertilde",
+], {
+    numArgs: 1,
+}, function(context, args) {
+    const body = args[0];
+    return {
+        type: "accentunder",
```
kevinbarabash: camel case

src/stretchy.js
```diff
@@ -0,0 +1,179 @@
...
+    xleftrightharpoons : "\u21cb",
+    xtwoheadleftarrow : "\u219e",
+    xtwoheadrightarrow : "\u21a0",
+    xLongequal : "=",
+    xtofrom : "\u21c4",
+};
+
+const mathMLnode = function(label) {
+    const node = new mathMLTree.MathNode(
+        "mo", [new mathMLTree.TextNode(stretchyCodePoint[label.substr(1)])]);
```
kevinbarabash: The mix of strings starting with and without `\\` is confusing.  I kind of wish we removed the `\\` after parsing so we didn't have to deal with this here and there.  No action action required.  Just thinking about how we can make the code base easier to work with in the future.

static/images/Image-Licensing-and-Technical-Notes.txt
```diff
@@ -0,0 +1,77 @@
...
+    The outer SVG acts like a window that exposes only part of the inner SVG. The
+    outer SVG will grow or shrink to match the dimensions set for it by CSS.
+
+    The inner SVG always has a longer, thinner aspect ratio than the outer SVG.
+    After the inner SVG fills 100% of the height of the outer SVG, there is a long
+    arrow shaft left over. That left-over shaft is not shown. Instead, it is
+    sliced off because the inner SVG is set to "preserveAspectRatio='... slice'".
+
+    Thus, the reader sees an arrow that matches the subject matter width without
+    distortion.
```
kevinbarabash: Super cool!

static/images/Image-Licensing-and-Technical-Notes.txt
```diff
@@ -0,0 +1,77 @@
...
+    So browsers apply anti-aliasing as described at http://www.rastertragedy.com/.
+    A long straight arrow shaft therefore will sometimes appear as if it has a
+    blurred edge.
+
+    To mitigate this, these SVG files contain a second "brush-stroke" on the
+    arrow shafts. That is, a second long thin rectangular SVG path has been
+    written directly on top of each arrow shaft. This reinforcement causes some
+    of the screen pixels to display as black instead of the anti-aliased gray
+    pixel that a  single path would generate. So we get arrow shafts whose
+    edges appear to be sharper.
```
kevinbarabash: Nice :)

test/katex-spec.js
```diff
@@ -1777,6 +1777,211 @@ describe("An accent builder", function() {
...
+        expect(getBuilt("\\xrightarrow x")[0].classes).toContain("mrel");
+        expect(getBuilt("\\xrightarrow [under]{over}")[0].classes).toContain("mrel");
+        expect(getBuilt("\\xrightarrow +")[0].classes).toContain("mrel");
+        expect(getBuilt("\\xrightarrow +")[0].classes).not.toContain("mbin");
+        expect(getBuilt("\\xrightarrow )^2")[0].classes).toContain("mrel");
+        expect(getBuilt("\\xrightarrow )^2")[0].classes).not.toContain("mclose");
+    });
+});
+
+describe("A horizontal brace parser", function() {
```
kevinbarabash: Can you add some tests with optional arguments for `\\overbrace`?

test/katex-spec.js
```diff
@@ -1755,8 +1755,8 @@ describe("An accent parser", function() {
         expect(parse.type).toEqual("supsub");
     });
 
-    it("should not parse expanding accents", function() {
-        expect("\\widehat{x}").toNotParse();
+    it("should parse expanding accents", function() {
+        expect("\\widehat{x}").toParse();
```
kevinbarabash: Could you add tests for shifty accents?

src/buildHTML.js
```diff
@@ -1427,31 +1434,52 @@ groupTypes.accent = function(group, options) {
...
-    // Shift the accent over by the skew. Note we shift by twice the skew
-    // because we are centering the accent, so by adding 2*skew to the left,
-    // we shift it to the right by 1*skew.
-    accentBody.children[1].style.marginLeft = 2 * skew + "em";
+        accentBody = stretchy.svgSpan(group, options);
+
+        if (skew > 0) {
+            // Nudge the accent to the right.
+            const adjSize = "calc(100% - " + 2 * skew + "em) 100%";
+            accentBody.style.backgroundSize = adjSize;
```
ronkok: `2 * skew` returns a positive number. Subtracting that number from 100% makes the result less than 100%. So we're making the width shorter. The second 100%, the height, remains unchanged.

src/buildHTML.js
```diff
@@ -1473,6 +1501,230 @@ groupTypes.accent = function(group, options) {
...
+            {type: "elem", elem: line, shift: lineShift},
+        ], "individualShift", null, options);
+        return makeSpan(["mord"], [vlist], options);
+
+    } else {
+        //  \cancel, \bcancel, \xcancel
+        let node = makeSpan([], [innerGroup], options);
+
+        // buildCommon.makeVList does not work well with a tall image span.
+        // So work out vertical placement here.
```
ronkok: When I pass a height much greater than 1.0 em to `buildCommon.makeVList`, it would respond to `maxFontSize` in a curious way. When given a low value of `maxFontSize`, `makeVList` would create a span that acted as if it had a large value of negative margin. If I passed a large value of `maxFontSize`, `makeVList` would create a span with too much height. 

I never found a way to pass a value that was just right. I'm not sure it is possible. In the end, I found it more reliable to write code that I could understand.

I haven't looked at how `llap` works. I'll do that. But keep in mind that `\cancel` has to make a *background* image appear in front of its subject matter. That complicates things.

src/buildHTML.js
```diff
@@ -1473,6 +1501,230 @@ groupTypes.accent = function(group, options) {
...
+groupTypes.boxed = function(group, options) {
+    const innerGroup = buildGroup(group.value.body, options);
+    let node = makeSpan([], [innerGroup], options);
+
+    // Create a strut with which to enforce a span height.
+    const strut = makeSpan([options.style.reset(), "strut"], [], options);
+    strut.style.height = (innerGroup.depth + innerGroup.height + 0.528) + "em";
+    strut.style.verticalAlign = -(innerGroup.depth + 0.25) + "em";
+
+    node = makeSpan(["mord", "boxed"], [node, strut], options);
```
ronkok:
```
    .boxed {
        display: inline-block;
        padding: 0 0.278em 0 0.25em;
        border: 0.04em solid black;
    }
```

src/buildHTML.js
```diff
@@ -1473,6 +1501,230 @@ groupTypes.accent = function(group, options) {
...
+        // Create a span, strikeThru, with the background-image.
+        const strikeThru = stretchy.strikeSpan(group.value.label.substr(1),
+            options);
+        strikeThru.style.height = totalHeight + "em";
+        if (innerGroup.height <= style.metrics.xHeight) {
+            strikeThru.style.top = 0.2 + "em";
+        }
+
+        // Wrap the position:absolute span with a position:relative span.
+        node = makeSpan(["outer-wrap"], [node, strikeThru], options);
```
ronkok:
```
    // \cancel, \bcancel, and \xcancel
    .outer-wrap {
        display: inline-block;
        position: relative;

        > .inner-wrap {
            display: inline-block;
            padding: 0;
        }

        > .strike {
            width: 100%;
            display: block;
            position: absolute;
```
etc...

Putting a `position-absolute` span inside of a `position-relative` span, as suggested [here](https://css-tricks.com/absolute-positioning-inside-relative-positioning/), was my way of dealing with positioning when I could not get `makeVList` to work.

src/stretchy.js
```diff
@@ -0,0 +1,179 @@
...
+    node.setAttribute("stretchy", "true");
+    return node;
+};
+
+const katexImagesData = {
+                 // height, depth, fileName
+    overleftarrow : [0.334, 0, "leftarrow"],
+    underleftarrow : [0.334, 0, "leftarrow"],
+    xleftarrow : [0.261, 0.261, "xleftarrow"],
```
ronkok: I used http://glyphrstudio.com/online/ to convert the KaTeX Main and KaTeX AMS ttf font files to SVG format. I then used the `path d= ... ` geometry to create the SVG files used in this PR. The fonts conveniently extracted into SVG paths with a scale in which 1000 inside the SVG viewBox equaled 1 em in the document. 

0.334 is because the arrowhead (from KaTeX AMS) is 334 units tall inside the font. Similarly, the KaTeX Main arrowhead, used for extensible arrows, is 522 units tall inside the font, and I split that distance around the math axis.

src/stretchy.js
```diff
@@ -0,0 +1,179 @@
...
+    // and/or the mask-image.
+    const label = group.value.label.substr(1);
+    let height = 0;
+    let depth = 0;
+    const classArray = ["stretchy"];
+    let fileName = "";
+
+    if (utils.contains(["widehat", "widetilde", "undertilde"], label)) {
+        // There are four SVG images available for each function.
+        // Choose a taller image when there are more characters.
```
ronkok: Per *The TeXbook*, page 443, rule 12:

TeX has a series of glyphs available for `\widetilde` and `\widehat`. It calculates the length of the subject matter and uses the shortest glyph that is longer than the subject matter. When the subject matter is longer than any of the glyphs, TeX puts in the longest one it has. So there can be a mis-match in length, like this: .

I believe the MnSymbols package has more and longer glyphs to pick from, but it also eventually runs out of glyphs.

My basic scheme doesn't know when to stop. It will match the length of any subject matter.  And these images start to look flat when the aspect ratio gets too small. So for long strings, I use a taller image.

src/stretchy.js
```diff
@@ -0,0 +1,179 @@
...
+        const numChars = group.value.value.length;
+        if (numChars > 5) {
+            height = 0.34;
```
ronkok: I do not, not, not want to stretch these images vertically. Then the line thickness would scale with the height and they would look terrible. So I manage height by choosing a different SVG.  Both `\widehat` and `\widetilde` have four different SVG image files available. 

`\widehat` and `\widetilde` have different height break points; it was just different managing a smooth curve versus a sharp hat. But in both cases, the tallest SVG is 0.34ems tall. So that is where I stop making image choices and set a 0.34em height.

src/stretchy.js
```diff
@@ -0,0 +1,179 @@
...
+    }
+
+    let node;
+    if (utils.contains(["overbracket", "underbracket"], label)) {
+        node = buildCommon.makeSpan(["stretchy", label], [], options);
+        // Deal with the optional arguments.
+        if (group.value.thickness) {
+            const lineWt = group.value.thickness.value + "em";
+            node.style.borderWidth = (group.value.isOver ? lineWt : 0) + " " +
+                lineWt + " " + (group.value.isOver ? 0 : lineWt) + " " + lineWt;
```
ronkok: `.isOver`, when true, indicates an **over**brace or **over**bracket, as opposed to an **under**brace or **under**bracket.

src/stretchy.js
```diff
@@ -0,0 +1,179 @@
...
+        }
+        if (group.value.height) {
+            height = group.value.height;
+            node.style.height = height + "em";
+        }
```
ronkok: This particular section of code applies only to `\overbracket` and `\underbracket`, not any of the other stretchy wide elements. These two functions were created by the `mathtools` [package](http://mirror.unl.edu/ctan/macros/latex/contrib/mathtools/mathtools.pdf), which gave them two optional arguments. One of the optional arguments lets the author choose a custom height (pp 12–13). So I could not hard-code a height into the CSS. It has to be done inline.

src/stretchy.js
```diff
@@ -0,0 +1,179 @@
...
+
+    } else if (options.color) {
+        classArray.push(fileName);         // Set span height and IE image.
```
ronkok: Very much a live comment.

This class sets a `background-image` and sets the span height. In the next two line of code, I apply two more classes, which CSS will apply only to browsers that support CSS `mask`.  

Because IE and Edge don't support `mask`, they will render the fall back solution defined in the first class, a black `background-image`.  And the CSS for span height, defined by the first class, remains valid for all browsers.

src/buildMathML.js
```diff
@@ -167,6 +168,20 @@ groupTypes.color = function(group, options) {
 };
 
 groupTypes.supsub = function(group, options) {
+    // Is the inner group a relevant horizonal brace?
+    let isBrace = false;
+    let isOver;
+    let isSup;
+    if (group.value.base) {
+        if (group.value.base.value.type === "horizBrace") {
+            isSup = (group.value.sup ? true : false);
```
ronkok: Yes. Say that an author writes `\overbrace{...}^{note}`.  Then both `group.value.sup` and `group.value.base.value.isOver`  will be true, `.isSup` will be true, and control over `group.value.sup` will pass to `groupTypes.horizBrace`. 

If the author writes `\overbrace{...}_{note}` (don't ask me why), then `group.value.sup` will be false and control over `group.value.sub` will pass to `groupTypes.supsub` to be written as a normal subscript.

src/buildMathML.js
```diff
@@ -167,6 +168,20 @@ groupTypes.color = function(group, options) {
...
+            if (isSup === group.value.base.value.isOver) {
+                isBrace = true;
```
ronkok: Well, I used an attribute name that is confusing. I was worried about that. 

Back in `functions.js`, where I handle the functions `\overbrace`, `\underbrace`, `\overbracket`, and `\underbracket`, I  give the type attribute `horizBrace` to all four of them. I should have named that type attribute `horizBraceOrBracket`. Unless you can think of something less verbose.

src/buildMathML.js
ronkok: Yep.

You know, in MathML, they actually have an `accentunder` tag. I was following that and not the camelCase standard . My bad. I'll change it.

src/buildMathML.js
ronkok: Maybe I should have read this comment before I got pedantic in the previous one.

src/buildMathML.js
ronkok: <...sound of palm hitting forehead...>

src/buildMathML.js
ronkok: I'll test that. I've been using https://www.mathtowebonline.com/ and Microsoft 
Word to see how others do this MathML thing. Your comment makes perfect sense to me, but I'll check before I rush into a code change.

src/functions.js
ronkok: Well, they both exist: 

The `\Overrightarrow` accent is pretty obscure. It comes from a package called `overrightarrow`. But the SVG file is already necessary for an extensible arrow. So this accent only costs a few lines of CSS.

src/functions.js
ronkok: Yes, they are all stretchy. But not all of them are shifty.

The TeXbook, p 443, rule 12, describes that when an accent is placed over a single character, the accent is shifted to the right by the `\skewchar` of its font. And the previously existing accents (i.e. `!isStretchy`) all do that shift.  TeX also shifts `\widehat` and `\widetilde`.

LaTex does not shift `\overrightarrow` or `\overrightharpoon`. I did that on my own because I think they look dopey unshifted. If you want more TeX purity, let me know and I take the shift off those two.

src/functions.js
ronkok: Thank you.

src/functions.js
ronkok: Yep.

static/images/Image-Licensing-and-Technical-Notes.txt
ronkok: The SVG spec also has a clever way to use nested SVGs that stretch to arbitrary length. In my research, I found bug reports to Microsoft (in 2015!) that complained that the stretchers were not working. So I started long and used a slice. Per my BrowserStack testing, this method works everywhere including IE 9+

test/katex-spec.js
ronkok: I'm not sure how Jasmine would know whether or not a shift occurred. That sounds more like a Screenshotter thing. Unless I'm missing something?

test/katex-spec.js
ronkok: Will do.

src/buildHTML.js
```diff
@@ -1473,6 +1501,230 @@ groupTypes.accent = function(group, options) {
...
+    const arrowShift = -style.metrics.axisHeight + arrowBody.depth;
+    const upperShift = -style.metrics.axisHeight - arrowBody.height -
+        2 * ruleWidth;
+
+    // Generate the vlist
+    let vlist;
+    if (group.value.below) {
+        const lowerShift = -style.metrics.axisHeight
+            + lowerGroupWrap.height + arrowBody.height
+            + 3 * ruleWidth;
```
ronkok: I confess that this came from trial and error. I wrote this bit code before I had learned anything about how to read TeX macros. I was at the time sincerely hoping that someone would come along and put me right.

I now know more than I did and I'll take a look. I may yet need some help

src/stretchy.js
ronkok: I completely agree.

src/buildHTML.js
```diff
@@ -140,6 +141,8 @@ const shouldHandleSupSub = function(group, options) {
             group.value.alwaysHandleSupSub);
     } else if (group.type === "accent") {
         return isCharacterBox(group.value.base);
+    } else if (group.type === "horizBrace") {
+        return (isSup === group.value.isOver);
```
ronkok: I also like the idea of keeping the logic together. Before I started, the arguments passed to this function were: `shouldHandleSupSub(group.value.base, options)`. So inside the function, `group.value.sub` is unavailable. 

I could pass `group` to the function instead of `group.value.base`. Then all of this bit of logic can be done in one location. Of course, other parts of the function would also need to be tweaked.

src/buildHTML.js
```diff
@@ -1473,6 +1501,230 @@ groupTypes.accent = function(group, options) {
...
+            {type: "elem", elem: vSpan, shift: 0},
+            {type: "elem", elem: noteReset, shift: noteShift},
+        ], "individualShift", null, options);
+    }
+
+    return makeSpan(["mord", (group.value.isOver ? "mover" : "munder")],
+        [vlist], options);
+};
+
+groupTypes.accentunder = function(group, options) {
```
ronkok: Yep.

src/stretchy.js
```diff
@@ -0,0 +1,179 @@
...
+        height = imgData[0];
+        depth = imgData[1];
+        fileName = imgData[2];
+        if (label.substr(0, 1) === "x") {
+            classArray.push("x-arrow");     // Lengthen the arrow via padding.
+        }
+    }
+
+    let node;
+    if (utils.contains(["overbracket", "underbracket"], label)) {
```
ronkok: The preceding `if/else-if` block will select a class and set a default height. So it is necessary. 

The code within the `if (utils.contains(["overbracket", "underbracket"], label)) {` block is there for cases where the author has used optional arguments to set custom height and line weight.

src/buildHTML.js
```diff
@@ -1473,6 +1501,230 @@ groupTypes.accent = function(group, options) {
...
+        options.style.sizeMultiplier;
+
+    const accentBody = stretchy.svgSpan(group, options);
+    const accentShift = innerGroup.depth + 2 * ruleWidth + accentBody.height;
+
+    // Generate the vlist, with the appropriate kerns
+    const vlist = buildCommon.makeVList([
+        {type: "elem", elem: innerGroup, shift: 0},
+        {type: "elem", elem: accentBody, shift: accentShift},
+    ], "individualShift", innerGroup.height, options);
```
ronkok: I'll make the change.

src/buildHTML.js
```diff
@@ -1473,6 +1501,230 @@ groupTypes.accent = function(group, options) {
...
+            note = buildGroup(group.value.sup, newOptions);
+            noteReset =  makeSpan([style.reset(), style.sup().cls()],
+                [note], newOptions);
+        } else {
+            newOptions = options.withStyle(style.sub());
+            note = buildGroup(group.value.sub, newOptions);
+            noteReset =  makeSpan([style.reset(), style.sub().cls()],
+                [note], newOptions);
+        }
+        group = group.value.base;
```
ronkok: Say that a author writes `\overbrace{long equation}^{short note}`. Then `group.type` will be "supsub" and `group.value.base` corresponds to `long equation`.

src/buildHTML.js
```diff
@@ -1473,6 +1501,230 @@ groupTypes.accent = function(group, options) {
...
+    const braceBody = stretchy.svgSpan(group, options);
+
+    // Generate the vlist, with the appropriate kerns
+    const braceShift = (group.value.isOver ? -body.height - 3 * ruleWidth :
+                        body.depth + braceBody.height + 3 * ruleWidth);
+
+    let vlist = buildCommon.makeVList([
+        {type: "elem", elem: body, shift: 0},
+        {type: "elem", elem: braceBody, shift: braceShift},
+    ], "individualShift", null, options);
```
ronkok: No particular reason for `shift`. For the extensible arrows, I had to use `makeVList` in `individualShift` mode and I got accustomed to it. As far as I know, the behavior is the same. But your point is well taken. For braces, it would be more clear if I used `makeVList` in `firstBaseLine` mode. So I'll change it.

Well, I could move that block of course, but that would mean writing it twice.  The vlist in question consists of the subject matter and the brace: 

If there is a note needed above the brace, then I need to wrap the first vlist with another one to show the note: 

And I can't put all three items in the same vlist. If I did, then a long note might control width of the brace. I want the brace width to be controlled only by the subject matter:

src/buildHTML.js
```diff
@@ -1473,6 +1501,230 @@ groupTypes.accent = function(group, options) {
...
+        const strut = makeSpan([style.reset(), "strut"], [], options);
+        const totalHeight = innerGroup.depth + innerGroup.height + 0.4;
+        strut.style.height = totalHeight + "em";
+        strut.style.verticalAlign = -(innerGroup.depth + 0.2) + "em";
+
+        node = makeSpan(["inner-wrap"], [node, strut], options);
+
+        // Create a span, strikeThru, with the background-image.
+        const strikeThru = stretchy.strikeSpan(group.value.label.substr(1),
+            options);
```
ronkok: I'm not removing the first letter. I'm removing the opening `\` character. As you note elsewhere, it may have been a good idea to strip off that `\` in `functions.js`.

src/functions.js
ronkok: I've made the change.

src/buildHTML.js
kevinbarabash: That seems reasonable since we're always passing `group.value.base` at all of the call sites.

src/buildHTML.js
kevinbarabash: Got it... I saw the `+ 2` and missed the `-`.  We support template strings so you may want to change this to `calc(100% - ${2 * skew}em) 100%` for improved readability.

src/buildHTML.js
kevinbarabash: Thanks for the awesome explanation.

I didn't see that `vlist` gets used by `vspan` when `haveNote` is `true`.  This is fine as is.  

Using `firstBaseline` vs `individualShift` isn't a big deal.  I was curious more than anything else.

src/buildHTML.js
```diff
@@ -1473,6 +1501,230 @@ groupTypes.accent = function(group, options) {
...
+        group.value.height = calculateSize(group.value.height.value,
+                options.style);
+    }
+
+    // Create the stretchy element
+    const braceBody = stretchy.svgSpan(group, options);
+
+    // Generate the vlist, with the appropriate kerns
+    const braceShift = (group.value.isOver ? -body.height - 3 * ruleWidth :
+                        body.depth + braceBody.height + 3 * ruleWidth);
```
kevinbarabash: I see it's coming from `groupTypes.overline`.

src/buildHTML.js
kevinbarabash: I'm not familiar with the inner workings of makeVlist... I was unware of `makeFontSizer`.  :(  After hearing the issues you experience I don't think `llap`'s approach would help, because it's not creating an atom that has a height much greater than 1.0em.

src/buildHTML.js
kevinbarabash: I figured it out much later in the review and forgot to remove this comment.

src/buildHTML.js
kevinbarabash: Sorry... github collapsed this file and I completely missed it.  :(

src/buildMathML.js
kevinbarabash: No worries. :smile:

src/functions.js
kevinbarabash: I missed that `\\overrightarrow` is on the previous line.  Sounds good.

src/functions.js
kevinbarabash: d'oh... I missed the `!` in `const isStretchy = !utils.contains(...)`.  

Thanks for calling out the difference in behavior for `\overrightarrow` and `\overrightharpoon`.  One of our guiding principles is to match LaTeX's behavior so these should be left out.

src/stretchy.js
kevinbarabash: Cool.  Can you add this as a comment in the code?

src/stretchy.js
kevinbarabash: Cool.  I just found it curious that the tallest SVGs had the same height even though the heights of the smaller SVGs didn't seem to line up.

src/stretchy.js
kevinbarabash: It makes sense after looking the .less file.  :)

test/katex-spec.js
kevinbarabash: A `.toBuild()` test should exercise different code paths for shifted accents.  It's less important once we have the screenshot tests.

src/buildHTML.js
ronkok: Yes. I would also like to point out that the new accents include harpoons. They are like overline in that they need a healthy kern distance. Otherwise, a long flat line looks like it is pasted to the top of the subject matter

src/buildHTML.js
ronkok: Agreed.

src/stretchy.js
ronkok: Will do.

src/functions.js
ronkok: I'll take the shift off the two arrows.

If we're going to be that fastidious, then we will need to talk about the arrowheads on accents. LaTeX arrow accents are different from AMS-LaTeX arrow accents which are different from MnSymbol arrow accents. Given that discrepancy, I went my own way and used an arrowhead that is slightly different that any of them. I'll work up a comparison for use in the conversation.

test/katex-spec.js
ronkok: Well, `\widehat`, which Jasmine already tests, happens to be both stretchy and shifty. I will write a test for one that is not shifty, say `\overrightarrow`. And I'll edit the output strings to make it more clear which is which.

src/buildHTML.js
```diff
@@ -1473,6 +1501,230 @@ groupTypes.accent = function(group, options) {
...
 
+groupTypes.horizBrace = function(group, options) {
+    const style = options.style;
+    const ruleWidth = fontMetrics.metrics.defaultRuleThickness /
+        style.sizeMultiplier;
+
+    const haveNote = (group.type === "supsub");
+    let note;
+    let newOptions;
+    let noteReset;
```
ronkok: Done.

src/buildHTML.js
ronkok: Done.

src/buildHTML.js
```diff
@@ -1427,31 +1434,52 @@ groupTypes.accent = function(group, options) {
...
+
+        if (skew > 0) {
+            // Nudge the accent to the right.
+            const adjSize = "calc(100% - " + 2 * skew + "em) 100%";
+            accentBody.style.backgroundSize = adjSize;
+        }
+
+        accentBody = buildCommon.makeVList([
+            {type: "elem", elem: body},
+            {type: "kern", size: 2 * ruleWidth},
```
ronkok: The accent clearance that I used was a typo. I was trying to match `\overline` and use clearance = 3 * ruleWidth. The harpoon accent, like `\overline`, needs a good size clearance.  I'll change the existing 2 to a 3.

src/functions.js
ronkok: It's time to talk about arrow accents. Below are four images of the exact same function.
four-examples
That's from LaTeX, AMS-LaTeX,  the mathabax package, and the MnSymbol package. Why are they all so different? Well, I have a guess. The following is from AMS-LaTeX:
four-accents
One of those accents is not like the others. That is, the `\overrightarrow` is much taller than the others. AMS fixed the collision that was in plain LaTeX.  But their glyph is still much taller than other accents. I think mathabax and MnSymbol are sized more like accents.

So I used an arrowhead from KaTeX-AMS font. It's 0.334 em high instead of 0.522 em high. For comparison, mine is in the middle of this list:
five-examples
So now we have a choice to make. Naturally, I think my arrow is the best choice. MathJax chose to follow the AMS-LaTeX accent. I can support either my arrow or the AMS-LaTeX one. I would like to avoid the other three.

Let me know what you decide.

edemaine: [Here's the discussion](https://github.com/Khan/KaTeX/issues/587#issuecomment-286767664) we had about this topic before.  I agree that your arrow looks "nicest", in that I'd want to use it as a long form of `\vec`.  As linked in the discussion, lots of people want this.  I guess what would be nice if there was a way to reproduce your arrow (approximately) with LaTeX.  Is there some `\usepackage` that would do it?  `\usepackage[a]{esvect}` is probably moderately close, though I think the aspect ratio of the arrow is different.  It's weird that `\overrightarrow` doesn't use this existing glyph used by `\vec`...  I guess we could define our own `katex.sty` package that implemented it...

ronkok: The `\vec` glyph is odd. I've now looked at the geometry of a lot of the arrow glyphs in these fonts and they all have a arrow shaft thickness of exactly 0.04 em. Except `\vec`. It's thicker.

src/buildHTML.js
```diff
@@ -1473,6 +1501,230 @@ groupTypes.accent = function(group, options) {
...
+
+    // Create a strut with which to enforce a span height.
+    const strut = makeSpan([options.style.reset(), "strut"], [], options);
+    strut.style.height = (innerGroup.depth + innerGroup.height + 0.528) + "em";
+    strut.style.verticalAlign = -(innerGroup.depth + 0.25) + "em";
+
+    node = makeSpan(["mord", "boxed"], [node, strut], options);
+    node.height = (innerGroup.height + 0.425);
+    node.maxFontSize = innerGroup.maxFontSize;
+    node.depth = (innerGroup.depth + 0.325);
```
ronkok: Thank you for the nudge to get this right. The relevant macro is from ams:

```
\newcommand{\boxed}[1]{\fbox{\m@th$\displaystyle#1$}}
```

And LaTeX source2e defines parameters for `\fbox` to be: \fboxsep = 3pt;  \fboxrule = .4pt

So I need to adjust the constants you alluded to. Also the CSS. Not only that, but I also need to set the innerGroup in `Style.DISPLAY`.  I'll do all of those things.

edemaine: It'd also be good to add comments to the code that mention those definitions from ams and source2e, so that we can trace 'em back.  Thanks!!

ronkok: Yep.

kevinbarabash:
@ronkok how are the changes coming?

ronkok:
I should finish tomorrow.

src/functions.js
ronkok: I never heard back with a definite decision. So I've changed the accent arrows match AMS-LaTeX.

src/buildHTML.js
ronkok: That line of code did not shorten the image. So I changed it. The current line includes an extra pair of parens to emphasize the correct  operator precedence: `"calc(100% - " + (2 * skew) + "em) 100%"`

ronkok:
@kevinbarabash, I have added one line of code to macros.js.  Now GitHub shows a conflict at that line, and my attempt to resolve the conflict failed. I would appreciate any help that you might offer.

edemaine:
@ronkok I forget, have you rebased before? (`git rebase -i upstream/master` if you have upstream defined) This lets you deal with the conflict by tweaking the offending commit.

ronkok:
@kevinbarabash, Here's the current status of PR #670: I've picked up the review comments. The substance of the work is contained in commit ee71b1b. The next two commits resolved a small merge conflict and resolved some lint errors. At that point, the commit passed all Travis tests.

In the latest commit, I've added some Screenshotter tests to ss_data.yaml, per your request. These tests did not pass, but then I didn't expect them to pass.

The work is once again ready for your review. Let me know if you need anything more from me.

kevinbarabash:
@ronkok cool!  I'll checkout the changes and run the screenshotter to generate images for the new tests.

kevinbarabash:
@ronkok I've run the screenshot tests and opened a PR against your PR.

In case anyone's interested you can see the screen shots at https://github.com/ronkok/KaTeX/pull/1/files.

ronkok:
Thank you very much for the help with the screenshots. I'll pull them in Now, comments:

1.  Firefox should be upgraded to the current version. The current version has full, not partial, support for CSS `mask`, so it will render the proper color for stretchy accents.

2. I see a mis-alignment of the `\xcancel` over the integral group. I'll look into it.

ronkok:
@kevinbarabash, I'm afraid that I must ask for help once again. The first screenshotter run revealed an error in the placement of `\xcancel` on a tall group. I have fixed the bug, but now the screenshotter of course does not match the file that you created. Could you create a new version of that file?

kevinbarabash:
@ronkok will do.

kevinbarabash:
> Firefox should be upgraded to the current version. The current version has full, not partial, support for CSS mask, so it will render the proper color for stretchy accents.

I see what you mean.  We can do that in a separate PR.

kevinbarabash:
Things are looking good, but I'd like to do another pass before merging and it's getting late.  Hopefully I'll be able to merge this tomorrow.

ronkok:
That sounds great. I will have no computer access for much of the day tomorrow, but if anything comes up, I'll get to it in the evening.

kevinbarabash:
I tried running texcmp and was able to get it to work for some of the new screenshot tests by adding various `\usepacakge` statements to `test.tex`.  Here are what those comparisons look like (KaTeX in red and LaTeX in green):

Boxed:
boxed

ExtensibleArrows:
```
! Undefined control sequence.
<argument> \lhook 
                  \joinrel \relbar 
l.50 ...poons[ab]{ABC} + \xhookrightarrow[ab]{ABC}
                                                   \\
```

HorizontalBraces:
horizontalbraces

LowerAccent:
loweraccent
I had to disable `\undertilde` because I don't have the `undertilde` package installed.

StretchyAccent:
stretchyaccent
The first line is shifted over b/c I had to remove `\Overrightarrow` b/c I couldn't figure out what package the command is from.

StrikeThrough:
strikethrough
Not sure what's up with the LaTeX `\sout` commands rendering the strikethrough lines as underlines.

Here are the packages that are being used in test.tex:
```
\documentclass[10pt]{article}

\usepackage{amsmath,amssymb,textcomp,gensymb}
\usepackage[mathscr]{eucal}
\usepackage{eufrak}
\usepackage[papersize={133pt,100pt},margin=0.5pt]{geometry}
\usepackage{color}
\usepackage{etoolbox}
%% new packages added below
\usepackage{mhsetup}
\usepackage{mathtools}
\usepackage{empheq}
\usepackage{extpfeil}
\usepackage{fdsymbol}
\usepackage{cancel}
\usepackage[normalem]{ulem}
\usepackage{MnSymbol}
...
```

@ronkok do you know which packages I should be adding to fix the `Undefined control sequence` errors?

ronkok:
@kevinbarabash , regarding the undefined control sequence, I expect that we have an conflict between `mathtools` and `MnSymbol`.  ([Reference](https://tex.stackexchange.com/questions/28656/conflict-between-mathtools-and-mnsymbol#28662))

I'm not sure how to test elements from both packages at the same time. Maybe we need to break the tests into multiple test cases?

edemaine:
Given that `MnSymbol` seems to just undefined `\lhook`, could we use the first workaround in the Reference?

```latex
\let\originallhook\lhook
\usepackage{MnSymbol}
\let\lhook\originallhook
```

(I've mentioned before: it'd be nice to have a clear list of packages needed to get the equivalent of all KaTeX features, or even a `katex.sty` package that does so for you.  But this is work for another issue.)

kevinbarabash:
After hammering on it some more, I discovered that `fdsymbol` also defined `lhook` so I commented out `fdsymbol` and the result is:

extensiblearrows

I'm not sure why `\xtwoheadrightarrow` is so long in the LaTeX rendering.  The spacing in between the numerator/denominator and the fraction bar look a little tight.

There's a sorts of things that could be improved to bring KaTeX's rendering of the arrows and the over/under braces closer in line with LaTeX's, but I think that's something that's better handled as follow up PRs. 

I would like to eventually get some MathML tests for the new commands, but we don't have an easy to verify the structure of MathML output and I'd like to setup Jest so that we can use its snapshot testing feature before we bother adding tests to check the structure.

kevinbarabash:
@ronkok looks like I have to run the screenshot again.  I'll send you a PR and then you'll have to fix the conflicts in `katex.less`.

static/katex.less
```diff
@@ -22,6 +22,12 @@
     // Protect elements inside .katex from inheriting text-indent.
     text-indent: 0;
 
+    // Prevent a rendering bug that misplaces \vec in Chrome.
+    text-rendering: auto;
+  
+    // Prevent background resetting in Windows's high-contrast mode.
+    -ms-high-contrast-adjust: none !important;
```
kevinbarabash: This was replaced by `* { -ms-high-contrast-adjust: none !important; }` a couple of lines below in https://github.com/Khan/KaTeX/pull/733 and should be removed.

kevinbarabash:
Not sure what happened with a couple of those screenshots.  I've pulled the most recent changes to `ronkok/master` and am running `make screenshots` again.

ronkok:
I've already done the merge. I can do more. No worries. 

Sometimes the only virtue that matters is persistence.

kevinbarabash:
@ronkok thank you for this awesome PR.  I know there's lots of people waiting for all of these new commands, including Khan Academy.  We'll have to do publish soon.

ronkok:
@kevinbarabash , I realize that this PR has been more trouble for you than most. I appreciate your help and your patience. I'm glad that the value added from the PR is worth the trouble. If you need any adjustments to stretchy wide elements, I'll be glad to help.

@edemaine , @glebm, thank you as well. Between the three of you, you pulled across the finish line someone who started this work knowing very little about JavaScript or SVG and nothing at all about Git.

edemaine:
Congrats @ronkok and @kevinbarabash on this 1600-line commit!  Excited to try out the new features.

PR #678, opened 2017-04-07
Fix font typo math -> main

### LINKED ISSUE #677: Some symbols are associated with a non-existing "math" font.
Symbols \coprod, \bigvee, \bigwedge...\smallint are assigned the fontFamily "math". It should probably be "main". See symbol.js:564-581.

edemaine:
Fix #677


kevinbarabash:
We end up ignore the `font` because we the symbol fonts Size1-Size4 fonts for these.  May as well fix this so that people don't think that `math` is a font.

PR #680, opened 2017-04-20
Added MathJax compatibility script

wilbowma:
This script helps a user easily switch from MathJax to KaTeX.


contrib/mathjax-compat/README.md
```diff
@@ -0,0 +1,15 @@
+# MathJax Compatibility Extension
```
xymostech: It might make sense to mention that MathJax supports the same approach, but I agree that it should be called something more related to what it actually does.

wilbowma:
@gagern, thanks a lot for the feedback! I'm pretty new to both MathJax and KaTeX, but wanted to try to contribute some of my personal scripts for wider use.

contrib/mathjax-compat/mathjax-compat.js
```diff
@@ -0,0 +1,28 @@
...
+}
+
+{
+    const l = document.scripts.length;
+    for (let i = 0; i < l; i++) {
+        const s = document.scripts[i];
+        if (s.type === "math/tex") {
+            replaceMathJaxInline(s);
+        }
+        if (s.type === "math/tex; display") {
```
wilbowma: You're right, there are definitely some bugs in this code. There at least ought to be a "mode=" before the display, and probably ought to compare case-insensitively. Unfortunately, I'm having trouble finding how MathJax does this. I'll keep looking.

contrib/mathjax-compat/README.md
wilbowma: That's a good point. I don't actually use MathJax directly, but through a Markdown parser that generates output in using the custom script types that MathJax can understand. I figured that was the normal interface, but reading the docs I see I'm wrong.

contrib/mathjax-compat/mathjax-compat.js
xymostech: Looks like MathJax does two checks. [The first check](https://github.com/mathjax/MathJax/blob/6273842a9746731b9ecca0de18ec9fd50a36df99/unpacked/MathJax.js#L2022) removes `/ *;.*/` (i.e. whitespace and then a semicolon and everything after the semicolon) and then checks for "math/tex".

Then, [the second check](https://github.com/mathjax/MathJax/blob/master/unpacked/jax/input/TeX/jax.js#L2168) looks for an instance of `/(;|\s|\n)mode\s*=\s*display(;|\s|\n|$)/` to decide if it should be in display mode. Those could probably be combined into a single check, I think it's just done in two parts because of how MathJax is structured.

gagern:
https://github.com/mathjax/MathJax/blob/2.7.0/unpacked/jax/input/TeX/jax.js#L2168 does the mode detection. I haven't found the mime type matching yet.

gagern:
The Selenium run by Travis is having problems setting up a fixed browser window size for the screenshots. Reason unknown but surely unrelated to your changes. I triggered a rerun of the tests. If this happens more often we may have to investigate in more detail.

edemaine:
FWIW, I've seen this now and then locally and in other pull requests (specifically #670).  Maybe a time delay needs to be increased...

gagern:
OK, Travis is happy now.

A bit of background on the screenshot sizes: with selenium we can configure the window size and then observe the size of the screenshot taken from the document area of that window. We don't want to hard-code the amount of space taken up by toolbars and so on. So [`setSize`](https://github.com/Khan/KaTeX/blob/v0.7.1/dockers/Screenshotter/screenshotter.js#L254-L274) will iteratively adjust the window size until it gets a screenshot of the desired size. This worked even in the failing run here. Then it goes for the actual screenshots for test inputs, and [there it verifies](https://github.com/Khan/KaTeX/blob/v0.7.1/dockers/Screenshotter/screenshotter.js#L382-L386) that the size of those screenshots still matches the one it obtained before. It's this sanity check that failed in the previous run here.

If we can reproduce this issue with some reasonable probability, it would be interesting to run this using one of the debug docker images (for [Chrome](https://hub.docker.com/r/selenium/standalone-chrome-debug/) resp. [Firefox](https://hub.docker.com/r/selenium/standalone-firefox-debug/)) to see whether we can work out what's going on with that window. Will see whether I can reproduce locally…

contrib/mathtex-script-type/README.md
```diff
@@ -0,0 +1,16 @@
...
+
+### Usage
+
+This extension isn't part of KaTeX proper, so the script should be separately
+included in the page.
+This include should be loaded after all `script` blocks you want to render.
+
+```html
+<script src="https://cdnjs.cloudflare.com/ajax/libs/KaTeX/0.7.1/contrib/mathtex-script-type.min.js" integrity="sha384-o+v+EkJWQmZj7XwHBxehTGJKE18182WyyN2glZMTPw9g5XxjN1uwrquNuMX/NJiF"></script>
+```
```
kevinbarabash: Could you expand on this script a bit more?  I assume that I need to load KaTeX somewhere too.

Maybe something along these lines:
```html
<html>
   <head>
       <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/KaTeX/0.7.1/katex.min.css" integrity="sha384-wITovz90syo1dJWVh32uuETPVEtGigN07tkttEqPv+uR2SE/mbQcG7ATL28aI9H0" crossorigin="anonymous">
       <script src="https://cdnjs.cloudflare.com/ajax/libs/KaTeX/0.7.1/katex.min.js" integrity="sha384-/y1Nn9+QQAipbNQWU65krzJralCnuOasHncUFXGkdwntGeSvQicrYkiUBwsgUqc1" crossorigin="anonymous"></script>
   </head>
   <body>
      <script type="math/tex">x+\sqrt{1-x^2}</script>
      <script src="https://cdnjs.cloudflare.com/ajax/libs/KaTeX/0.7.1/contrib/mathtex-script-type.min.js" integrity="sha384-o+v+EkJWQmZj7XwHBxehTGJKE18182WyyN2glZMTPw9g5XxjN1uwrquNuMX/NJiF"></script>
   </body>
</html>
```

contrib/mathtex-script-type/README.md
```diff
@@ -0,0 +1,16 @@
...
+This script type is commonly used by MathJax, so this can be used to support compatibility with MathJax.
+
+### Usage
+
+This extension isn't part of KaTeX proper, so the script should be separately
+included in the page.
+This include should be loaded after all `script` blocks you want to render.
+
+```html
+<script src="https://cdnjs.cloudflare.com/ajax/libs/KaTeX/0.7.1/contrib/mathtex-script-type.min.js" integrity="sha384-o+v+EkJWQmZj7XwHBxehTGJKE18182WyyN2glZMTPw9g5XxjN1uwrquNuMX/NJiF"></script>
```
kevinbarabash: Will this actually work at the moment?  Wouldn't we have to publish a `0.7.2` version of the KaTeX in order for `mathex-script-type.min.js` to appear in the published package?

kevinbarabash: I wonder if we should just encourage people to copy/paste the script manually before we do a publish and then update this to point to the published script afterwards.

kevinbarabash: Looking at the script though, we'd probably want to run `make build/contrib` and use the built version of the script.

edemaine: Are you suggesting that the Makefile automatically tweak the README to include the latest version number?  This seems like it'd be cool, though I don't quite know how it'd work...

Perhaps a simpler solution would be to write `<script src=".../KaTeX/contrib/mathtex-script-type.min.js"></script>` and let users fill in the right path, either local to their webserver (as many are probably using it) or on CDN?  We could also give a template for CDN including, but with `...` or something else in place of the actual version number.

contrib/mathtex-script-type/README.md
edemaine: @wilbowma Could you take a pass at this?

contrib/mathtex-script-type/README.md
kevinbarabash: I worded that poorly.  I think `<script src=".../KaTeX/contrib/mathtex-script-type.min.js"></script>` makes sense.  The README could be updated in a separate diff to point to the CDN version, but as this stands right now this code won't work.

contrib/mathtex-script-type/README.md
wilbowma: @edemaine, sorry about that. Got busy with other things and forgot about this. I'll take a look today.

contrib/mathtex-script-type/README.md
wilbowma: I used this URL based on `user-contrib/auto-render/README.md`. You're right, of course, that this path won't work just now, but I assumed that this script would be handled similarly to `auto-render`.

edemaine: The trouble is that once we merge your pull request, the README immediately becomes publicly visible.  So we should probably use `...` for now, and update after a release (or change them all to `...`).

wilbowma: Also, don't the build scripts automatically update the integreity hashes in the contrib README? (see above comment on earlier version of my Makefile.) Surely they can be modified to update the version in the CDN string.

wilbowma: Ah of course. Okay.

edemaine: Oops, indeed, `update-sri.js` already updates the version number.  It looks like it will replace any URL with any version number, including `...`, so we could still use that until a release happens.  Then the release will fix it to a working version number.

edemaine:
Sorry, now that I understand what the release script does, it makes sense to keep CDN URLs, but just change the version number part to `...` where it doesn't yet exist (your new file).

edemaine:
Thanks @wilbowma!  This looks good to me.  @kevinbarabash did you want to confirm the revisions address your review?

kevinbarabash:
@wilbowma sorry about the delay.  Thank you for the PR.  It looks like release.sh will take care of updating the integrity SHAs.  I'd like to the merge right before I publish but I probably won't get to the publish until tomorrow evening.

kevinbarabash:
I forgot to include this in v0.8.3 release so I'm going to merge this now so that I goes out in the v0.8.4 release for sure.  I'll post a separate PR to update the integrity SHAs.

kevinbarabash:
I just realized that posting a separate PR to update the integrity SHAs doesn't make sense b/c the publish has to happen for the new files to be uploaded to CDNs.

Issue #662, opened 2017-02-24
\texttt{>} causes an error


jsmaniac:
In the live preview at https://khan.github.io/KaTeX/, entering any of the following causes an error:

```
\texttt{>}
\textit{>}
\textrm{>}
\textbf{>}
\text{>}

\texttt{->} and so on.
```

The error is:
```
KaTeX parse error: Expected '}', got '>' at position 9: \texttt{>̲}
```

xymostech:
What do you expect to happen in these cases? Putting `\text{>}` into LaTeX gives me an upside down question mark.

![image](https://cloud.githubusercontent.com/assets/217900/23285109/2142ffbc-f9e4-11e6-8f03-7bd8535caa27.png)

jsmaniac:
@xymostech Good point, I had forgotten TeX's bogus behaviour for these commands. The `\texttt{>}` does print fine with LaTeX though (and that's what I was using). Incidentally, the “correct” form `\textrm{\textgreater}` also fails with KaTeX.

jsmaniac:
Also, adding `\usepackage[T1]{fontenc}` to the preamble of a LaTeX document makes the other commands produce the expected `>` symbol, so i'd argue that should be KaTeX's behaviour as it is very common to use `T1`, and the behaviour seems more reasonable than the upside-down question mark.

xymostech:
Oh, I guess I didn't actually check `\texttt{>}`, you're right, that does work correctly in LaTeX! Thanks for reporting this! :)

edemaine:
Fix forthcoming in #684.

PR #684, opened 2017-04-27
Support <, >, | and many \text... commands in text mode

edemaine:
This started as a fix to #662, which reported lack of `<` and `>` support in text mode.  As suggested, I mapped them to regular `<` and `>` characters as if `\usepackage[T1]{fontenc}` is in effect.  This seems like the preferred behavior (unless you're typing in Spanish or another language that uses upside-down symbols... but then perhaps better to use Unicode symbols).

I also noticed that `|` wasn't working in text mode, but was supposed to, so I added that.  These are one-line changes.

I then implemented `\textless`, `\textgreater`, `\textbar` as aliases to these symbols.

I then started going through [this symbol list](http://mirror.utexas.edu/ctan/info/symbols/comprehensive/symbols-a4.pdf#page=14) to see what other characters were missing.  I added support for `\textdollar`, `\textunderscore`, `\textbraceleft`, `\textbraceright`, `\textbardbl`, `\textendash`, `\textemdash`, `\textquote(dbl)(left/right)` as aliases to existing symbols.  I didn't attempt to add the other text symbols, though that might be good for a future project...


kevinbarabash:
@edemaine I'm confused about `\textdollar` and the other `\text___` aliases.  Won't `\text{\textdollar}` output `\textdollar` instead of `$`?

edemaine:
`defineSymbol(text, main, textord, "$", "\\textdollar");` defines `\textdollar` to output the symbol `$` in the `main` font with type `textord`.  To confirm, here's a screenshot of the local server tester:

textdollar

src/symbols.js
```diff
@@ -109,7 +109,9 @@ defineSymbol(math, main, textord, "\u2660", "\\spadesuit");
 
 // Math and Text
 defineSymbol(math, main, textord, "\u2020", "\\dag");
+defineSymbol(text, main, textord, "\u2020", "\\dag");
 defineSymbol(math, main, textord, "\u2021", "\\ddag");
+defineSymbol(text, main, textord, "\u2021", "\\ddag");
```
kevinbarabash: may as well add `\textdagger` and `\textdaggerdbl` as well.

edemaine:
Added.   Also noticed that `\circledR`, `\checkmark`, `\pounds` work in text mode too, so added those.  And I'm fairly certain `\pounds` is a `textord` not `mathord`.  `\pounds\text{\textsterling\pounds}` now produces 3 identical symbols.

kevinbarabash:
@edemaine it looks like the change to `\pounds` change the symbols screenshot test.  Can you verify that `\pounds` is indeed a textord in math mode?  If it, please update the screenshot for the symbols test.

edemaine:
Indeed, `$\pounds$` generates a mathord, same as `$\mathsterling$` (another alias I added).  `$\text{\pounds}$` generates a textord (I believe -- at least, the LaTeX output of the symbol is different from the math version), but our font doesn't have that symbol, so I've left it generating a mathord.

src/symbols.js
```diff
@@ -600,14 +614,23 @@ defineSymbol(math, main, mathord, "\u0131", "\\imath");
...
 defineSymbol(text, main, textord, "\u201c", "``");
+defineSymbol(text, main, textord, "\u201c", "\\textquotedblleft");
 defineSymbol(text, main, textord, "\u201d", "''");
+defineSymbol(text, main, textord, "\u201d", "\\textquotedblright");
 defineSymbol(math, main, textord, "\u00b0", "\\degree");
 defineSymbol(text, main, textord, "\u00b0", "\\degree");
 defineSymbol(math, main, mathord, "\u00a3", "\\pounds");
+defineSymbol(math, main, mathord, "\u00a3", "\\mathsterling");
+defineSymbol(text, main, mathord, "\u00a3", "\\pounds");
+defineSymbol(text, main, mathord, "\u00a3", "\\textsterling");
```
kevinbarabash: You mentioned that `$\text{\pounds}$` generates a textord, should `\text{\textsterling}` do the same?  Can you add a `TODO` so that people aren't confused by the use of `mathord`s in text mode later on?

edemaine: Added.  Although I found [this discussion](https://gitter.im/Khan/KaTeX/archives/2014/09/24) where @xymostech suggests that maybe we should move away from mathord vs. textord, and rather just specify which font to use.  It seems LaTeX has a notion of "mathord" but no notion of "textord" (at least I've never seen one in debugging output)?

edemaine:
Should be good to go now.  Someday, a more thorough symbol review wouldn't hurt, but this should help!

kevinbarabash:
LGTM.  Thanks!

Issue #657, opened 2017-02-12
Feature request: support \colonequals or similar


katef:
A couple of TeX packages provide macros for this; `:=` has special spacing, where the dots are made symmetrical.

There's also the unicode codepoint, U+2254.

Thank you!

kevinbarabash:
I think it'll take a while to get the right glyph.  Would it be useful to have the `\colonequals` command output `:=` for now and then when we have the correct glyph it will start outputting the correct glyph?

gagern:
I think I've seen `\coloneqq` for this, too. Yes, I think having the command first and the glyph when we can get it would make sense. I also wonder whether we could hack something like a `\raisebox` and use that to move the colon to the symmetrical position, at least approximately. The proper type for the whole thing, I guess mathrel, would seem important, too.

katef:
I agree about getting the markup supported now (a `\raisebox` or similar sounds fine!) and drawing a proper glyph later.

SO lists a few ways to mark that up for TeX: http://tex.stackexchange.com/questions/4216/how-to-typeset-correctly

edemaine:
FWIW, [oberdiek's colonequals](http://ctan.mackichan.com/macros/latex2e/contrib/oberdiek/colonequals.pdf) implements it via `\vcenter`.  I imagine implementing `\vcenter` in KaTeX would be on the tricky side, though #637 is related.

edemaine:
The `\raisebox` solutions seem pretty good, actually -- because they use `ex` measurements, they should scale reasonably to other font sizes.  However, KaTeX doesn't actually support `\raisebox`.  `\raisebox` would be useful for #224 (`\LaTeX`) too.

edemaine:
I looked a little closer at `mathtools` which implements a centered colon very simply as
```latex
\def\vcentcolon{\mathrel{\mathop\ordinarycolon}}
\providecommand\ordinarycolon{:}
```
i.e. just `\mathrel{\mathop:}`.  In LaTeX, this centers the colon:

![colon_eq](https://cloud.githubusercontent.com/assets/2218736/25505981/461d901c-2b72-11e7-9510-58a3f077a00a.png)

In KaTeX, `\mathrel{\mathop:}=` and `:=` seem to render exactly the same (with the colon not centered, like on the right).  Is this a known issue?

edemaine:
[...]

Issue #224, opened 2015-04-29
\LaTeX not rendered


rgbkrk:
``` latex
\KaTeX
```

renders just fine but

``` latex
\LaTeX
```

doesn't. I'm all for letting `\KaTeX` be rendered, but there are plenty of times people want to render `\LaTeX`.

xymostech:
There was a pull request to add this: #131. I'd like for `\LaTeX` to be rendered using the true offsets/dimensions as in actual LaTeX, which is why it wasn't merged in. This is something that would probably be nicer if this was just implemented as a macro whenever we figure that out, but if you (or anyone else) wants to try to incorporate the true metrics.

gagern:
A non-macro definition on the other hand should be fairly easy. I've written some details about that in the pull request. Read that first to better understand the macro in question.

Doing this as a macro following the official definition would be really tricky, since we'd need kerning (which might be a good thing to have in any case, and which shouldn't be too hard except for the strange lexing rules for lengths), and we'd need box registers and ways to measure the content of boxes at least vertically (I believe horizontally is impossible), and we'd need raw box construction commands like `\vbox to …`, and we'd need flexible glue for `\vss`. I don't think this is realistic in the near future, even though I've recently made some progress regarding macro definitions.

The logo could probably also be described as a macro using simpler means, like e.g. `\raisebox` instead of the `\vbox`. But that would mean manually computing the offset up front, and hardcoding it into the macro instead of deriving it from the font metrics. I'm not sure this is desirable, particularly seeing how many people (myself included) have asked for KaTeX to also work with different fonts.

kevinb7:
We should try to add ```\kern``` b/c it's used in quite a few different commands.

kevinb7:
I added ```\kern``` to the list of missing horizontal spacing commands in https://github.com/Khan/KaTeX/issues/164.

edemaine:
Fixed by #974 (thanks to previous additions of `\kern` and `\raisebox`).

PR #685, opened 2017-04-27
Support for \raisebox

edemaine:
I started working on `\raisebox` support which will help for #657 and #224.  Currently `\raisebox{...ex}{...text...}` "works", except that the bounding box of the resulting typeset text isn't set correctly, so raised/lowered objects cut into the lines before/after.  I guess I'm still not fully up to speed on how HTML generation works here -- based on other code in `buildHTML.js`, I tried manually updating `.height` or `.depth` according to whether the box moved up or down, but that didn't work.  If anyone has a chance to look at this, I'd appreciate help fixing it.

Here's a sample render with the test server:

raisebox

Here's the same example rendered in TeX, so seems good except for the bounding box issue:

raisebox_tex

Also fixed a typo where parsed sizes are accidentally labeled as "color" nodes instead of "size" nodes.


src/buildHTML.js
```diff
@@ -1491,6 +1491,22 @@ groupTypes.mclass = function(group, options) {
...
+    const span = groupTypes.text(group, options);
+    if (group.value.dy) {
+        const dy = group.value.dy.value;
+        span.style.top = "" + (-dy.number) + dy.unit;
+        // Not working yet:
+        if (dy.number > 0) {
+            span.height += calculateSize(dy.number, options.style);
+        } else {
+            span.depth -= calculateSize(dy.number, options.style);
+        }
```
edemaine: The idea is that the height increases if the box is moved up, and the depth increases if the box is moved down.  At least, my mental model is that height is the extent above the baseline (y-max) while depth is the extent below the baseline (y-min) -- please correct me if I'm wrong.  However, this code doesn't seem to be working, and I'm not really sure how `height` and `depth` end up getting used (or not).

src/buildHTML.js
```diff
@@ -1491,6 +1491,22 @@ groupTypes.mclass = function(group, options) {
     return makeSpan([group.value.mclass], elements, options);
 };
 
+groupTypes.transform = function(group, options) {
+    const span = groupTypes.text(group, options);
+    if (group.value.dy) {
+        const dy = group.value.dy.value;
+        span.style.top = "" + (-dy.number) + dy.unit;
```
edemaine: Good point -- indeed, it currently does not, but would be easy to tweak to.  (A bigger issue is that `calculateSize` only supports mu and em.)

src/buildHTML.js
kevinbarabash: > At least, my mental model is that height is the extent above the baseline (y-max) while depth is the extent below the baseline (y-min) -- please correct me if I'm wrong.

That's right, but if a glyph is shifted vertically, the the extent below the baseline will be different.

What happens if you put the test expression inside a `\sqrt` or a `\frac`?  It may be that just our wrapper container isn't doing the right thing with the `height` and `depth` you're setting on the `span`.

src/buildHTML.js
kevinbarabash: It looks like `calculateSize` supports `ex`.
```
const calculateSize = function(sizeValue, style) {
    let x = sizeValue.number;
    if (sizeValue.unit === "ex") {
        x *= style.metrics.emPerEx;
    } else if (sizeValue.unit === "mu") {
        x /= 18;
    }
    return x;
};
```

src/buildHTML.js
ronkok: Inside a `\frac` or `\sqrt`, the bounding box looks good:
raisebox

This screenshot, by the way, was done with code that was slightly tweaked:

```
groupTypes.transform = function(group, options) {
    const span = groupTypes.text(group, options);
    const dy = calculateSize(group.value.dy.value, options.style);
    if (dy > 0) {
        span.height += dy;
    } else if (dy < 0) {
        span.depth -= dy;
    }
    span.style.top = -dy + "em";
    span.style.position = "relative";
    return span;
};
```

kevinbarabash: @ronkok did your tweaks have any effect on the top level bounding box?

ronkok: Now that you ask me, yes they did. Odd. I was only trying to clean it up, not change how it worked. Here's a screenshot from the original code:
raisebox2
Now that I need to take this more seriously, here’s a version that restores the outer if statement.

```
groupTypes.transform = function(group, options) {
    const span = groupTypes.text(group, options);
    if (group.value.dy) {
        const dy = calculateSize(group.value.dy.value, options.style);
        if (dy > 0) {
            span.height += dy;
        } else if (dy < 0) {
            span.depth -= dy;
        }
        span.style.top = -dy + "em";
        span.style.position = "relative";
    }
    return span;
};
```
That code gives me the good results.

edemaine:
Awesome, thanks @kevinbarabash and @ronkok for finding the solution here!  Looks like it was close indeed, and switching to `calculateSize` at the top level was the missing link.

Here is what my original example `A\raisebox{-0.5em}{B}\raisebox{0.5em}{C}D` looks like with the new code (now pushed):

abcd

It's better, though the top margin is still a lot smaller (zero or even slightly less) than a straight `\text{C}`:

c

ronkok:
You mentioned a smaller margin that a straight `\text{C}`. I wonder if that is why buildCommon.makeVList spends the effort to call makeFontSizer.

edemaine:
Hey, I got it!  I was looking at the definition of the bottom strut in `buildHTML.js`, and it uses a `vertical-align` to put the bottom strut below the top strut.  I didn't know you could put numbers into `vertical-align`.  So I tried that instead of a `top` style (see [this trivial diff](https://github.com/Khan/KaTeX/pull/685/commits/da322228a220256b73eccb2694d66574b5102915)), and it now has exactly the right bounding box!

abcd

kevinbarabash:
> I didn't know you could put numbers into vertical-align

Neither did I.  Cool!

kevinbarabash:
@edemaine now that everything's working correctly, can you add a screenshot test?

edemaine:
In adding a screenshot test similar to @ronkok's, I realize that we're not quite there yet: `vertical-align` behaves well at the top level (and in `\frac`s) but doesn't seem to interact properly with `\sqrt`...

abc

It seems that the positive `vertical-align` in the b is causing the ab to move down... weird...  Hmm, I think this is related to `makeFontSizer` that @ronkok mentioned, as the contents of the sqrt are put in a `vlist` (although so are the `\frac`s...).

ronkok:
I don't think the problem is yours. If I understand matters correctly, KaTeX does not change the radical height in a continuous manner. It's a step function. You happened to land on a bad height.

edemaine:
@ronkok Here's what it looks like, switching back to `top` setting:

abc

The only difference is the vertical shift in the `a\raisebox{0.5em}{b}`.

I believe the issue is [Movement Of the Line Box’s Baseline](http://christopheraue.net/2014/03/05/vertical-align/#movement-of-the-line-boxs-baseline) (quirks of vertical-align).  Still grokking...

edemaine:
Upon further investigation, I think the `top` version (essentially @ronkok's version) was actually correct.  There is a larger space at the top of `\text{a}` simply because of the overall line height.  This margin disappears even in a simple example like `a^{a^{a^a}}`:

aaaa

So I've reverted to that version, and added a screenshot test.  I think we're back to good-to-go.

kevinbarabash:
@edemaine you may want to the texcmp docker to compare the output against LaTeX's output.

edemaine:
One more question: I currently call the parse node "transform", with the thought that it could be used for rotatebox and such in the future. But I wonder if just calling it "raisebox" and keeping it simple would be better. Let me know your opinion.

kevinbarabash:
Good question.  I think having separate `raisebox` and `rotatebox` methods probably makes since we'll need to do something pretty different to implement `rotatebox`.

edemaine:
@kevinbarabash Good call on using texcmp.  I see now that LaTeX resets font size to full size when switching to text mode via `\raisebox` (which `\text` probably is careful to avoid), and I need to do the same:

raisebox

According to [this ImageMagick page](http://www.imagemagick.org/Usage/color_mods/#grayscale), these are probably equivalent, so perhaps we could change that to support more ImageMagick versions?

edemaine:
In principle I fixed it, but I've run into a bug: #709

kevinbarabash:
@edemaine I've merged the fix for #709.

edemaine:
I rebased.  The switch to `\normalsize` is working now, but the bounding box computation seems completely broken (unphased by the `\raisebox`).  In this texcmp, the red (KaTeX) raised b's intersect whatever is above them (`\frac` line or `\sqrt` line).

raisebox

edemaine:
I got another chance to look this over and revise substantially.  I believe this is now working almost perfectly.  I use `makeVList` to do the vertical shift, so it should be well-behaved (and it appears at least reasonably so).  Internal to buildHTML, I use `sizing` and `text` nodes to switch to the equivalent of `\textrm{\normalsize...}`.  texcmp is almost perfect (slight horizontal spacing difference?):

raisebox


I also slightly cleaned up the parse tree: `text` nodes were using a `style` attribute to specify a font, but styles mean something else in KaTeX; switched the attribute to `font`.

Probably best to look at this [as a complete diff](https://github.com/Khan/KaTeX/pull/685/files) but if you prefer you can look at just the last commit for the changes relative to previous versions.

src/buildHTML.js
```diff
@@ -304,7 +304,7 @@ groupTypes.ordgroup = function(group, options) {
 };
 
 groupTypes.text = function(group, options) {
-    const newOptions = options.withFont(group.value.style);
+    const newOptions = options.withFont(group.value.font);
```
kevinbarabash: Calling `withFont` with a font makes more sense.

src/buildMathML.js
```diff
@@ -601,6 +601,10 @@ groupTypes.mclass = function(group, options) {
     return new mathMLTree.MathNode("mstyle", inner);
 };
 
+// Transforms (translation/rotation) don't seem to have a representation
+// in MathML, so just treat them like \text{...}
+groupTypes.raisebox = groupTypes.text;
```
kevinbarabash: :+1:

src/functions.js
```diff
@@ -51,7 +51,7 @@ defineFunction([
     return {
         type: "text",
         body: ordargument(body),
-        style: textFunctionStyles[context.funcName],
+        font: textFunctionStyles[context.funcName],
```
kevinbarabash: Since we're calling these fonts now, should probably rename this to `textFunctionFonts`.

kevinbarabash:
@edemaine thanks for picking this one up again and finishing it off.  The results look great!

edemaine:
Thanks @kevinbarabash for all the reviews today!!

ronkok:
I think that one can get effective presentation MathML for `\raisebox` with code something like:
````
const node = new mathMLTree.MathNode(
        "mpadded", [buildGroup(group.value.body, options)]);
node.setAttribute("voffset", group.value.dy.value);
````

I have not tested that code, but I have put the HTML below into Firefox
````
<math xmlns="http://www.w3.org/198/Math/MathML">
    <mi>a</mi><mpadded voffset="0.5em">b</mpadded><mi>c</mi>
</math>
````

... and it comes out looking like this:
raisebox

This comment is coming very late, I know. Sorry.

PR #689, opened 2017-04-30
Fix spaces before \middle

### LINKED ISSUE #683: Issue around `\middle` and spacing
KaTeX ignores any spacing command on the left of `\middle`:

For example, In the image below, space should be inserted before `|` but no space is rendered, although the space right after `|` is properly rendered:

Even worth, however many spacing commands are inserted before `\middle|`, they are all ignored, e.g. `\left.a\ \ \ \ \ \ \ \ \ \ \middle| \ b\right.` will be rendered same as the above.

kohler:
Shift spaces back into the replacement middle delimiter. Fix #683.


src/buildHTML.js
```diff
@@ -50,6 +50,23 @@ const isBinRightCanceller = function(node, isRealGroup) {
 };
 
 /**
+ * Splice out any spaces from `children` starting at position `i`, and return
+ * the spliced-out array. Returns null if `children[i]` does not exist or is not
+ * a space.
+ */
+const spliceSpaces = function(children, i) {
+    let j = i;
```
edemaine: Perhaps it would be more intuitive to give `i` a default value of 0, via `let j = i || 0`?

src/buildHTML.js
```diff
@@ -74,24 +91,18 @@ const buildExpression = function(expression, options, isRealGroup) {
...
-    let spaces = null;
     for (let i = 0; i < groups.length; i++) {
-        if (isSpace(groups[i])) {
-            spaces = spaces || [];
-            spaces.push(groups[i]);
-            groups.splice(i, 1);
-            i--;
-        } else if (spaces) {
+        const spaces = spliceSpaces(groups, i);
+        if (spaces && i < groups.length) {
```
edemaine: `if (spaces && ...) else if (spaces) ...` seems a little weird / hard to understand.
How about `if (spaces) { if (i < groups.length) ... else ... }`?

src/buildHTML.js
```diff
@@ -1278,11 +1289,17 @@ groupTypes.leftright = function(group, options) {
...
+            const middleDelim = inner[i];
+            if (middleDelim.isMiddle) {
                 // Apply the options that were active when \middle was called
                 inner[i] = delimiter.leftRightDelim(
-                    inner[i].isMiddle.value, innerHeight, innerDepth,
-                    inner[i].isMiddle.options, group.mode, []);
+                    middleDelim.isMiddle.value, innerHeight, innerDepth,
+                    middleDelim.isMiddle.options, group.mode, []);
+                // Add back spaces shifted into the delimiter
+                const spaces = spliceSpaces(middleDelim.children, 0);
```
edemaine: Could use default 0 value of `i` here.

src/buildHTML.js
kohler: No, thanks.

src/buildHTML.js
kohler: No, thanks.

src/buildHTML.js
kohler: Does not seem important.

src/buildHTML.js
edemaine: Really? Why?

kohler: If I thought the code was clearer the way you suggested I would have written it that way. I don't like deeply nested if statements. And it is clearly a matter of taste.

You are asking me for ten minutes or so of work for trivial changes. Like five seconds of typing then ten minutes to rerun the tests just to verify nothing broke, jank around with the GitHub PR, etc. I already spent a lot of time making the fix. As well as writing the original feature—not because I needed it myself, but because I was grateful for KaTeX and wanted to help out with features the main developers wanted.

Projects that want contributions walk a balance between enforcing standards and respecting their contributors' time. As a contributor, I see no standards involved with your comments, and I consider these particular requests less than respectful of my time.

edemaine: I'm sorry that you felt your time was disrespected -- not my intent.  We certainly appreciate the time you've spent in fixing this bug (and in implementing the original feature).  I'll move this agreed-to-be stylistic issue to a separate PR.

kohler: Thanks very much for understanding. I absolutely appreciate your work also, and hope there's no hard feelings. I think it's important that code reviewers for open-source projects understand when they're asking for important functionality changes vs. important stylistic changes vs. tweaks on matters of taste, and thus hope this exchange has been useful for both of us.

PR #697, opened 2017-05-11
Add \iff, \implies, \impliedby support

### LINKED ISSUE #190: Add support for \implies and \iff
These are tricky because they should have space around them (`\;` on either side). We should probably implement these as some sort of internal macro, which can maybe be a precursor to #37.

edemaine:
Fix #190 via @ronkok's code.

I thought about tests but couldn't think of anything that interesting.  A `expect(...).toParseLike` doesn't seem to check anything.  I could add a screenshot test, but it is also pretty "boring".

### DIFF
--- a/src/macros.js
+++ b/src/macros.js
@@ -21,3 +21,12 @@ defineMacro("\\endgroup", "}");
 // \def\overset#1#2{\binrel@{#2}\binrel@@{\mathop{\kern\z@#2}\limits^{#1}}}
 defineMacro("\\overset", "\\mathop{#2}\\limits^{#1}");
 defineMacro("\\underset", "\\mathop{#2}\\limits_{#1}");
+
+//TODO: When implementing \dots, should ideally add the \DOTSB indicator
+//      into the macro, to indicate these are binary operators.
+// \def\iff{\DOTSB\;\Longleftrightarrow\;}
+// \def\implies{\DOTSB\;\Longrightarrow\;}
+// \def\impliedby{\DOTSB\;\Longleftarrow\;}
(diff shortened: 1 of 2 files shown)


ronkok:
That looks good. And kudos to @gagern, who made macros possible.

src/macros.js
```diff
@@ -21,3 +21,10 @@ defineMacro("\\endgroup", "}");
 // \def\overset#1#2{\binrel@{#2}\binrel@@{\mathop{\kern\z@#2}\limits^{#1}}}
 defineMacro("\\overset", "\\mathop{#2}\\limits^{#1}");
 defineMacro("\\underset", "\\mathop{#2}\\limits_{#1}");
+
+// \def\iff{\DOTSB\;\Longleftrightarrow\;}
+// \def\implies{\DOTSB\;\Longrightarrow\;}
+// \def\impliedby{\DOTSB\;\Longleftarrow\;}
```
kevinbarabash: Can you add a TODO above these to indicate that this is what we want the the implementation of these commands to eventually look like?

edemaine: Added.

edemaine:
@kevinbarabash Implemented both your suggestions (and rebased).  Thanks for the feedback!

I also noticed that, while @ronkok's code had a `\mathrel` surrounding the operators, the `amstex.sty` code did not, so I switched to the latter.

ronkok:
I'm good with the change. The span still is coded as a `class="mrel"`.

kevinbarabash:
@edemaine thanks for adding these new commands.

PR #702, opened 2017-05-13
Add Function Support Page

ronkok:
Add an illustrated HTML version of the KaTeX function support page.

Add buttons to the KaTeX home page that link to the proposed function support page.


ronkok:
I’ve put a temporary preview version of the proposed function support page [here](http://copyasmarkup.com/katexdocs/function-support.html).

kevinbarabash:
@ronkok looks good except there seems to be an encoding issue with function-support.html.  Github is showing it as a single line and there are `^M`.  Could you have a look at that?

ronkok:
Done.

kevinbarabash:
Thanks.  Looks great!

screen shot 2017-05-13 at 3 08 40 pm

Issue #708, opened 2017-05-16
texcmp woes


edemaine:
I've never been able to run the texcmp docker via `texcmp.sh`.  I get this permission error:

```$ sudo ./texcmp.sh
+ imgname=katex/texcmp
+ tag=1.1
++ docker images
++ awk '/katex\/texcmp *1\.1/{print $3}'
+ imgid=01e4c2f34b2d
++ dirname ./texcmp.sh
+ cd .
+ npm install
npm WARN texcmp@ No repository field.
+ [[ -z 01e4c2f34b2d ]]
++ cd ../..
++ pwd
+ base=...../KaTeX
+ docker run --rm -v ...../KaTeX:/KaTeX -w /KaTeX/dockers/texcmp katex/texcmp:1.1 nodejs texcmp.js

fs.js:432
  return binding.open(pathModule._makeLong(path), stringToFlags(flags), mode);
                 ^
Error: EACCES, permission denied '/KaTeX/dockers/texcmp/texcmp.js'
    at Object.fs.openSync (fs.js:432:18)
    at Object.fs.readFileSync (fs.js:286:15)
    at Object.Module._extensions..js (module.js:473:44)
    at Module.load (module.js:356:32)
    at Function.Module._load (module.js:312:12)
    at Function.Module.runMain (module.js:497:10)
    at startup (node.js:119:16)
    at node.js:902:3
```

A little more investigation on trying to access that file within the docker:

```$ sudo docker run --rm -v ...../KaTeX:/KaTeX -w /KaTeX/dockers/texcmp katex/texcmp:1.1 ls -l texcmp.js
-rw-r--r-- 1 20028 20028 9115 May 16 13:16 texcmp.js
$ sudo docker run --rm -v ...../KaTeX:/KaTeX -w /KaTeX/dockers/texcmp katex/texcmp:1.1 cat texcmp.js
cat: texcmp.js: Permission denied
```

I tried chmod'ing the files to world readable, but I'm on AFS, so it seems that I need to make the files world readable via `fs sa . system:anyuser read`.  This is odd to me, because the screenshotter docker runs fine without such a setting.  Any chance that the calls to docker could be unified between screenshotter and texcmp to behave like the screenshotter one? (I guess inheriting permissions from the calling user.)

I *can* run `texcmp.js` from the command line without docker, but because my ImageMagick is old, I need to tweak the command line slightly:

```diff
[...]

PR #710, opened 2017-05-16
Adds \notni character ∌

jsmaniac:


jsmaniac:
[clabot:check]

ronkok:
I don’t think this approach will work, because the KaTeX-Main font does not contain a glyph for U+220C. What could work is:

```
defineMacro("\\notni", "\\mathrel\\bgroup\\rlap\\bgroup\\kern\\bgroup0.04em" +
    "\\egroup/\\egroup\\bgroup\\ni\\egroup\\egroup");
```

That results in a character that fits in pretty well:
notni

jsmaniac:
@ronkok Thanks a lot for the solution! I hadn't thought about the font issue.

jsmaniac:
Note that the best might be to support `\not\ni`, which works with AMSMath IIRC, but I'm not sure if KaTeX supports defining macros with arguments. So `\notni` seems a good compromise (its meaning is clear, it exists in (non-AMS) LaTeX packages).

ronkok:
@jsmaniac's original code would give a MathML result that is much superior to the macro that I proposed. Perhaps the thing to do is accept the macro for now, but put U+220C on to the font To Do list, along with `\euro` and `\colonequals`. Also, the macro code should include a TODO comment that suggests that the macro is temporary and that the font glyph should be used when it is available.

jsmaniac:
@ronkok I added the  TODO.

edemaine:
@ronkok Thanks for the input! @jsmaniac Thanks for being responsive to the alternative approach!

If it's not too much trouble, could we see here output from `texcmp` (available in `dockers/texcmp`) comparing this symbol to one of the LaTeX versions of the symbol?  (This involves adding a tiny test to `test/screenshotter/ss_data.yaml`.)  Let me know if you have trouble running `texcmp`.

Incidentally, general `\not` support is in process/almost complete in #140.  (Hi @kevinbarabash!)  But I think there's still value to this PR, given that it is a supported command in some packages.  I guess we could later change the macro to expand to just `\not\ni`... though that might have slightly inferior spacing?

kevinbarabash:
I'll have a look at what's left to do in https://github.com/Khan/KaTeX/pull/140 this evening.

jsmaniac:
@edemaine I know about docker but don't use it in my regular development, and google is proving itself unhelpful at telling me how to start a `Dockerfile`. I tried `docker start dockers/texcmp` and `docker exec -ti dockers/texcmp /bin/sh`, but both fail. Care to remind me what's the one-liner?

edemaine:
@jsmaniac You can try `sudo dockers/texcmp/texcmp.sh YourTestName` which does the launching for you.  (But see #708)  Personally I've had more success with just running `node dockers/texcmp/texcmp.js YourTestName`.

jsmaniac:
@edemaine Sorry, I tried running `sudo dockers/texcmp/texcmp.sh Sqrt` on my local machine, and inside a fresh Ubuntu 14.04 machine (with the default NodeJS and with NodeJS 6 too), and it fails in both cases with `SyntaxError: Use of const in strict mode.`.

I also tried with the `node dockers/texcmp/texcmp.js Sqrt` command directly, and it throws `module.js:471 throw err;`.

You'll have to run texcmp, as I cannot.

edemaine:
Regarding your latter error, you probably need to run `npm install` (which installs all dependencies) first.  (Though hard to tell without seeing the actual error message, which is in the line after `throw err;`.)

I forgot you need to run screenshotter first, though, which I've only gotten to run using Docker.  So you might have had trouble anyway.

Anyway, here is what I got with `\usepackage{newtxmath}` added to `test/screenshotter/test.tex`:

notni

There's a difference here, but it seems like a consistent difference.  Here's a comparison of `\in\ni\notin\notni`:

notni 1

edemaine:
@kevinbarabash Great!  Perhaps we should wait to see if #140 is easy to finalize, and then modify this PR accordingly...

kevinbarabash:
@jsmaniac I'm getting close with https://github.com/Khan/KaTeX/pull/140.  With any luck it should be merged sometime this week and that will unblock this PR.  Thanks for your patience.

kevinbarabash:
@jsmaniac I finally got `\not` merged.  Can you update your diff to us `\not` instead of `/`?

jsmaniac:
@kevinbarabash Sure! I'll look into this next week if you don't mind. Thanks for the update!

edemaine:
@jsmaniac A reminder on this -- should be straightforward. Also, you can use `{` and `}` instead of `\bgroup` and `\egroup`.

src/symbols.js
```diff
@@ -527,6 +527,7 @@ defineSymbol(math, main, rel, "\u2190", "\\gets");
 defineSymbol(math, main, rel, ">", "\\gt");
 defineSymbol(math, main, rel, "\u2208", "\\in", true);
 defineSymbol(math, main, rel, "\u2209", "\\notin", true);
+defineSymbol(math, main, rel, "\u220c", "\\notni", true);
```
kevinbarabash: I think this line should be removed.

src/macros.js
```diff
@@ -298,3 +298,9 @@ defineMacro("\\simcoloncolon", "\\sim\\mathrel{\\mkern-1.2mu}\\dblcolon");
 defineMacro("\\approxcolon", "\\approx\\mathrel{\\mkern-1.2mu}\\vcentcolon");
 defineMacro("\\approxcoloncolon",
             "\\approx\\mathrel{\\mkern-1.2mu}\\dblcolon");
+
+// Present in newtxmath, pxfonts and txfonts
+// TODO: The unicode character U+220C ∌ should be added to the font, and this
+//       macro turned into a propper defineSymbol in symbols.js. That way, the
+//       MathML result will be much cleaner.
+defineMacro("\\notni", "\\not\\ni");
```
kevinbarabash: Should the definition be wrapped in a `\mathrel`?

kevinbarabash: After doing some checking this doesn't look to be necessary as `\ni` is already a mathrel and the `\not` has no effect on the atom type.

kevinbarabash:
@jsmaniac thanks for the PR.  Sorry this took so long to get merged.

jsmaniac:
Hi @kevinbarabash, sorry I didn't remove the `defineSymbol` quick enough. I thought it would allow typing the unicode character in the TeX source, and wanted to ask why it wasn't working when testing with the `npm test` server. I hadn't realised it was meant to translate the other way round (from the TeX command to the unicode char for MathML).

Thanks for fixing this and merging!

kevinbarabash:
@jsmaniac sorry for being over eager to merge your PR.  I think in order to support the unicode char for this you'll have to define a macro, see https://github.com/Khan/KaTeX/pull/950/files for some examples.

PR #718, opened 2017-06-06
Fix #711 issue with multiple superscripts.

### LINKED ISSUE #711: Multiple powers not printed very clearly
I am not sure how this is printed in the reference implementation (not sure what to look at), but `{10^{10}}^{100}` is not displayed in a way that I find particularly readable. Below are the output of KaTeX and the rendering engine used by LyX:

| KaTeX | LyX |
| --- | --- |
| image | image |

kohler:
The previous code incorrectly implemented TeXbook rules. First off,
the font metrics at issue should be those for the superscript/
subscript style, not the main style. Secondly, the metrics should be
scaled by the font size.

This changes some screenshotter tests (and they look more like TeX
now).


kevinbarabash:
Thanks for the PR.  Big improvement in the vertical position of subscripts.

Before:
accents

After:
accents

Before: 
binomtest

After:
binomtest

Before:
limitcontrols

After:
limitcontrols

Before:
oplimits

After:
oplimits

Before:
sqrt

After:
sqrt

Issue #519, opened 2016-08-20
Vertical misalignment when using Katex with transitional and frameset DOCTYPES in HTML4 and XHTML1.0


joelostblom:
I am having issues using Katex with Pandoc, which seems to be related to the DOCTYPE for the HTML4 documents. Pandoc uses the transitional XHTML 1.0 DOCTYPE by default, which causes some characters (such as fraction lines and subset notation) to be offset vertically and displayed above their correct position (as reported in #332 and #517).

I tested all the [doctypes listed at w3schools](http://www.w3schools.com/tags/tag_doctype.asp). The HTML5, strict HTML4, XHTML1.1 and strict XHTML1.0 below all work fine.

```
<!DOCTYPE html>
<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 4.01//EN" "http://www.w3.org/TR/html4/strict.dtd">
<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Frameset//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-frameset.dtd">
<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.1//EN" "http://www.w3.org/TR/xhtml11/DTD/xhtml11.dtd">
```

The transitional and frameset HTML4 and XHTML 1.0 listed below cause the vertical character offset.
```
<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 4.01 Transitional//EN" "http://www.w3.org/TR/html4/loose.dtd">
<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 4.01 Frameset//EN" "http://www.w3.org/TR/html4/frameset.dtd">
<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">
<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Frameset//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-frameset.dtd">
```

I can easily circumvent these issues by using Pandoc HTML5 template with `<!DOCTYPE html>`, but I wonder if there is any recommendation on which DOCTYPES to use for different document formats to have the most robust rendering of Katex elements. Should I just stay away from transitional and frameset DOCTYPES and use any of the other 4, or are there more things to look out for? 

[...]

PR #719, opened 2017-06-06
Fix interaction between styles and sizes.

kohler:
Rather than having both `textstyle` CSS classes and `size5` CSS classes affect the font size (and step on each other), implement sizes more the way TeX does: a command like `\displaystyle` changes the current size.

Remove `displaystyle/textstyle` CSS classes entirely, since they do nothing now.

This is actually a simplification, since now only `size` affects the size. Simplifies CSS and computation. Many screenshotter tests change; most of them (all?) change to be more like TeX.


kevinbarabash:
Thanks for the PR.  I wonder how this interacts with https://github.com/Khan/KaTeX/pull/718.  I'm guessing that after rebasing now that #718 has been merged will result in some additional changes to some of the screenshots.  I'll try to review this over the next couple of days.

kohler:
Nah, there's no interaction between this and #718—this actually assumed #718 :)

kohler:
(Rebased.)

kohler:
c2aaf90 fixes a longstanding bug in `\sqrt` in scriptsizes.

kohler:
I'm going to re-commit with some changes to the mechanism, because of corner cases in the interaction between sizes and styles.

Here's the issue. Consider:

`\scriptstyle\large a` vs. `\large\scriptstyle a`

These commands have no equivalent in TeX, but these work, and have the following result:

```
$\text{\large$a$}$---$\text{\large$\scriptstyle a$}$---$\scriptstyle\text{\large$a$}$
```
image

Note that `\large` overrides the preceding `\scriptstyle` size change. The question is, does `\large` override *all* properties of `\scriptstyle`, or just the size?

In this commit, `Options` has a `baseSize` which determines the sizes chosen by future style changes. A sizing command overrides *both* the currently active size *and* the baseSize, but *does not* override other aspects of the style.

This means that TeX like `\scriptstyle\large a\scriptstyle b` will have possibly-surprising results. Both “a” and “b” will nominally be in `\scriptstyle` (which affects, e.g., math spacing), but “b” will be smaller.

I'm not sure there's a more coherent solution than this (maybe you can think of one), but this seems most TeXlike.

kohler:
After playing with more examples, including `a^{2^{\sqrt{\sqrt{\large 2^{2^2}}}}}\large 2^{2^2}`, I now believe sizing commands should change to `\textstyle` (if the style is currently `\scriptstyle` or `\scriptscriptstyle`), and that's what's there now.

kevinbarabash:
The `\sqrt` changes look awesome.  It's interesting that many of the screenshots changed, but there doesn't appear to be any visual change.  I can only assume that some change in the order of operations of some of the calculations resulted in an imperceptible change to the layout.  I was surprised by the changes to ModScript-chrome.png without changes to ModScript-firefox.png.

> I now believe sizing commands should change to \textstyle (if the style is currently \scriptstyle or \scriptscriptstyle), and that's what's there now.

This is so that both the sizes glyphs and the spaces between them is consistent no matter where the sizing command is used, is that right?

kohler:
> The \sqrt changes look awesome. It's interesting that many of the screenshots changed, but there doesn't appear to be any visual change. I can only assume that some change in the order of operations of some of the calculations resulted in an imperceptible change to the layout. I was surprised by the changes to ModScript-chrome.png without changes to ModScript-firefox.png.

That is my assumption too, and I share your surprise. The main change here is that, e.g., `\scriptsize` changes are applied via the size mechanism, which might be, you know, 0.000001 off from the previous relative fontsize or something.

> This is so that both the sizes glyphs and the spaces between them is consistent no matter where the sizing command is used, is that right?

Yes.

For what it's worth, and it's up to you, but this will probably conflict with #519. Whatever gets merged first I'm happy to help resolve conflicts.

kohler:
Hi @kevinbarabash any other comments on this?

edemaine:
@kohler Did you mean something other than #519?

kohler:
Whoops, thanks, I meant the @ronkok stretchy wide PR, which is #670

edemaine:
Ah, OK, I figured.  Probably #670 wants to go first, as @kevinbarabash has been working on that for a while.

kevinbarabash:
@kohler sorry for the delay.  I want to merge #670 first.  I'm going to spend some more time tonight looking at it and will hopefully be able to merge it.  My hope is to get your PR reviewed either tomorrow or sometime this weekend.

edemaine:
#670 has landed.  I don't see any conflicts, but perhaps tests need to be rerun?

kohler:
The new PR is ready, including necessary changes to the stretchy functions (#670). I also added a screenshotter improvement.

src/Options.js
```diff
@@ -58,22 +68,69 @@ Options.prototype.extend = function(extension) {
...
 };
 
+function sizeAtStyle(size, style) {
+    return style.size < 2 ? size : sizeStyleMap[size - 1][style.size - 1];
+}
+
 /**
- * Create a new options object with the given style.
+ * Return an options object with the given style. If `this.style === style`,
+ * returns `this`.
```
kevinbarabash: This seems functionally equivalent as long as we don't mutate these objects.

src/Options.js
```diff
@@ -18,20 +37,10 @@ function Options(data) {
     this.style = data.style;
```
kevinbarabash: The doc string for this function references a `.reset` function with doesn't currently exist.  It needs to be updated.

src/Options.js
```diff
@@ -104,11 +161,15 @@ Options.prototype.withFont = function(font) {
 };
 
 /**
- * Create a new options object with the same style, size, and color. This is
- * used so that parent style and size changes are handled correctly.
+ * Return the CSS sizing classes required to switch from enclosing options
+ * `oldOptions` to `this`. Returns an array of classes.
```
kevinbarabash: Nice comment.

src/Parser.js
```diff
@@ -483,7 +483,7 @@ Parser.prototype.parseImplicitGroup = function() {
         const body = this.parseExpression(false);
         return new ParseNode("sizing", {
             // Figure out what size to use based on the list of functions above
-            size: "size" + (utils.indexOf(sizeFuncs, func) + 1),
+            size: utils.indexOf(sizeFuncs, func) + 1,
```
kevinbarabash: nice cleanup

src/buildMathML.js
```diff
@@ -489,8 +490,7 @@ groupTypes.sizing = function(group, options) {
     // in, so we can't reset the size to normal before changing it.  Now
     // that we're passing an options parameter we should be able to fix
     // this.
-    node.setAttribute(
-        "mathsize", buildCommon.sizingMultiplier[group.value.size] + "em");
+    node.setAttribute("mathsize", newOptions.sizeMultiplier + "em");
```
kevinbarabash: Does the comment above still apply?

src/delimiter.js
```diff
@@ -317,10 +311,11 @@ const makeStackedDelim = function(delim, heightTotal, center, options, mode,
     inners.push(makeInner(top, font, mode));
 
     // Finally, build the vlist
-    const inner = buildCommon.makeVList(inners, "bottom", depth, options);
+    const newOptions = options.havingBaseStyle(Style.TEXT);
```
kevinbarabash: Why are we reset the base style here?

dockers/Screenshotter/screenshotter.js
```diff
@@ -463,3 +506,21 @@ function promisify(f) {
...
+            console.error(stdout + stderr);
+            err.stdout = stdout;
+            err.stderr = stderr;
+            deferred.reject(err);
+        } else {
+            deferred.fulfill(stdout);
+        }
+    });
+    return deferred.promise;
+}
```
kevinbarabash: These changes seem unrelated to the rest of the PR.  Were you using this to compare the sizing changes with LaTeX output?

kohler: Yes. I'm happy to take them out of the PR but *shrug* it's all going in the same place eventually

src/Options.js
```diff
@@ -58,22 +68,69 @@ Options.prototype.extend = function(extension) {
     return new Options(data);
 };
 
+function sizeAtStyle(size, style) {
+    return style.size < 2 ? size : sizeStyleMap[size - 1][style.size - 1];
```
kohler: No because style.size might be 0 (DISPLAY).

It would be equivalent to `sizeStyleMap[size - 1][Math.max(0, style.size - 1)]` but I find this easier to understand

src/Options.js
```diff
@@ -58,22 +68,69 @@ Options.prototype.extend = function(extension) {
...
 /**
- * Create a new options object with the given size.
+ * Return an options object with a cramped version of the current style. If
+ * the current style is cramped, returns `this`.
  */
-Options.prototype.withSize = function(size) {
-    return this.extend({
-        size: size,
-    });
+Options.prototype.havingCrampedStyle = function() {
```
kohler: `buildHTML` uses it for convenience.

dockers/Screenshotter/screenshotter.js
edemaine: @gagern is probably the most experienced with the screenshotter, so he might want to look at this part of the PR (or separate the PR into two).

src/Options.js
```diff
@@ -5,6 +5,25 @@
  * `.reset` functions.
  */
 
+const Style = require("./Style");
+
+const sizeStyleMap = [ // [textsize, scriptsize, scriptscriptsize]
```
kohler: So these numbers are chosen by applying 0.7x and 0.5x the base size and rounding to an available KaTeX size. But your question inspired an experiment.

In old KaTeX, scriptsize is always 0.7x the size of textsize, and scriptscriptsize is 0.5x the size of textsize. This also holds in LaTeX at 10pt:

```
\documentclass[10pt]{article}
\usepackage[utf8]{inputenc}
\showboxbreadth=\maxdimen
\showboxdepth=\maxdimen
\begin{document}
\tracingonline=1
$a^{b^c}$\showlists
\end{document}
```

shows

```
\OML/cmm/m/it/10 a
\hbox(5.17056+0.0)x7.7638, shifted -3.62892
.\OML/cmm/m/it/7 b
.\hbox(2.15277+0.0)x3.74713, shifted -3.01779
..\OML/cmm/m/it/5 c
```

But at other point sizes these ratios don't hold precisely, since LaTeX was originally built for a fixed palette of fonts at specific sizes. For instance, `11pt` gives sizes 10.95/8/6, and `12pt` gives 12/8/6.

FWIW here are the sizes LaTeX uses at base font size 10pt, compared with the PR's current sizes:

TeX command | TeX Sizes | Current PR sizes
-----------|----------|------------
`\tiny` | 5/5/5 | 5/5/5
`\scriptsize` | 7/5/5 | 7/5/5
`\footnotesize` | 8/6/5 | 8/7/5 [NB KaTeX has no 6pt]
`\small` | 9/6/5 | 9/7/5
`\normalsize` | 10/7/5 | 10/7/5
`\large` | 12/8/6 | 12/8/7
`\Large` | 14.4/10/7 | 14.4/10/7
`\LARGE` | 17.28/12/10 | 17.3/12/9
`\huge` | 20.74/14.4/12 | 20.7/14.4/10
`\Huge` | 24.88/20.74/17.28 | 24.9/17.3/12

So some differences.

So what to do? We could keep the current sizes (& maybe document them). Or we could implement these more TeX-precise sizes in KaTeX, with or without the currently-missing 6pt font. We can also delay this choice (i.e. leave the current sizes in and then change them later).

src/Options.js
kohler: Good catch. The whole comment refers to “parent” functionality that is no longer present or needed. I'll replace this comment with a comment about it being readonly.

src/Options.js
kohler: Sure.

FWIW it would also be OK to use the old `.with*` names. The `.having*` names were very useful during auditing.

src/Options.js
```diff
@@ -58,22 +68,69 @@ Options.prototype.extend = function(extension) {
...
+ * `this.size === size && this.baseSize === size`, returns `this`.
+ */
+Options.prototype.havingSize = function(size) {
+    if (this.size === size && this.baseSize === size) {
+        return this;
+    } else {
+        // Ensure style is at least `\textstyle`.
+        let style = this.style;
+        if (style.size > 1) {
+            style = style.cramped ? Style.TEXT.cramp() : Style.TEXT;
```
kohler: It didn't hurt.

kohler: (I don't have a crisp response. :) )

src/Options.js
```diff
@@ -58,22 +68,69 @@ Options.prototype.extend = function(extension) {
...
+        return this.extend({
+            style: style,
+            size: size,
+            baseSize: size,
+        });
+    }
+};
+
+/**
+ * Like `this.havingSize(5).havingStyle(style)`.
```
kohler: It can create slightly fewer `Options` objects. In particular, imagine `x` has basesize 5 and style SCRIPT. Then `x.havingSize(5).havingStyle(Style.SCRIPT) != x`, but `x.havingBaseStyle(Style.SCRIPT) == x`.

The function was more useful (i.e., did more than just optimize) in a prior version, before I decided that it was best for `.havingSize()` to reset the style.

src/buildMathML.js
kohler: I have no idea and I have no way to check and it's MathML :(

src/delimiter.js
kohler: Because stacked delimiters are taken from the extension font, and in old TeX, the extension font was *only* used at size 10.

However, there is a wrinkle: In AMSLaTeX, the extension font is available (and used) at smaller sizes too. I could see changing the code to do that instead.

dockers/Screenshotter/screenshotter.js
kevinbarabash: It's definitely preferable to to have these as separate commits.  Even though it's all going to the same place, smaller commits will take less time to review and if one review needs revision it doesn't hold up the other.

src/delimiter.js
kevinbarabash: The delimiters look pretty good in the screenshots as is.

src/Options.js
kevinbarabash: For some reason I thought the sizes where started counting from `1`.  I think my reason for confusion is that I was conflating sizing command sizes with style command sizes.  :(  The fact that we're accessing values in `sizeStyleMap` should've clued me.

kohler:
I took out the screenshotter diff and updated some comments as requested.

src/Options.js
kevinbarabash: My understanding of KaTeX fonts is that we only have 10pt size fonts and we scale them appropriately.  What would prevent us from scaling that font to 6pt when necessary?

kevinbarabash: I agree this shouldn't be block, but if we defer please add a comment in the code.

kohler: There is nothing preventing us from scaling the font to 6pt. But because we do things with CSS here, it would mean that “normalsize”, which has been “size5” in CSS classes for a while, would become “size6”. Probably doesn't matter.

kohler:
I'm going to check in another commit that adds the 6pt size and implements TeX's size relationships for scriptstyle/scriptscriptstyle. I also gave KaTeX's sizes more significant digits.

kohler:
The TeX size relationships are in.

src/Options.js
kevinbarabash: Thanks.  I totally missed that file b/c github auto-collapsed it.

kevinbarabash: I'll try to review the changes to `buildHTML.js` over the next couple of days.

kohler:
Hi @kevinbarabash any comments?

kevinbarabash:
@kohler sorry.  I haven't reviewed changes to buildHTML.js.  I'll have time this weekend.

src/buildHTML.js
```diff
@@ -311,43 +308,33 @@ groupTypes.supsub = function(group, options) {
...
 
+    // Rule 18a
+    let supShift = 0;
+    let subShift = 0;
+
     if (group.value.sup) {
-        newOptions = options.withStyle(style.sup());
-        sup = buildGroup(group.value.sup, newOptions);
-        supmid = makeSpan([style.reset(), style.sup().cls()],
-            [sup], newOptions);
```
kevinbarabash: I'm curious why the wrapping isn't necessary anymore.

src/buildHTML.js
```diff
@@ -311,43 +308,33 @@ groupTypes.supsub = function(group, options) {
...
-
-        const substyle = style.sub();
-        subShift = base.depth + substyle.metrics.subDrop
-            * substyle.sizeMultiplier;
+        newOptions = options.havingStyle(style.sub());
+        subm = buildGroup(group.value.sub, newOptions, options);
+        if (!isCharacterBox(group.value.base)) {
+            subShift = base.depth + newOptions.style.metrics.subDrop
+                * newOptions.sizeMultiplier / options.sizeMultiplier;
+        }
```
kevinbarabash: Moving this up here to avoid unnecessary calculations when there isn't both sub and sup parts is a nice optimization.

src/buildHTML.js
```diff
@@ -311,43 +308,33 @@ groupTypes.supsub = function(group, options) {
...
     }
 
-    const base = buildGroup(group.value.base, options.reset());
-    let supmid;
-    let submid;
-    let sup;
-    let sub;
+    const base = buildGroup(group.value.base, options);
+    let supm;
+    let subm;
```
kevinbarabash: Why the name change from `sub` and `sup`?  Keeping those variables names would've reduced the number of changed lines.

src/buildHTML.js
kevinbarabash: I'm okay with keeping this change as it makes reviewing easier.

src/buildHTML.js
```diff
@@ -495,53 +477,49 @@ groupTypes.genfrac = function(group, options) {
...
                 clearance - ((axisHeight - 0.5 * ruleWidth) -
-                             (denom.height - denomShift));
+                             (denomm.height - denomShift));
         }
 
-        const mid = makeSpan(
-            [options.style.reset(), Style.TEXT.cls(), "frac-line"]);
-        // Manually set the height of the line because its height is
-        // created in CSS
-        mid.height = ruleWidth;
```
kevinbarabash: I see `makeLineSpan` takes care of setting the height.

src/buildHTML.js
```diff
@@ -450,20 +436,16 @@ groupTypes.genfrac = function(group, options) {
     const dstyle = style.fracDen();
     let newOptions;
```
kevinbarabash: I wish we weren't reusing this identifier for different options within the frac layout, but since this is the way the code was I guess we'll have to change it later.

src/Options.js
```diff
@@ -42,6 +52,7 @@ Options.prototype.extend = function(extension) {
     const data = {
         style: this.style,
         size: this.size,
+        textSize: this.textSize,
         color: this.color,
         parentStyle: this.style,
         parentSize: this.size,
```
kevinbarabash: Can we get rid of `parentSize`?

src/buildHTML.js
kevinbarabash: It looks like it's being handled by `buildGroup` now that we're passing parent options.

src/buildHTML.js
```diff
@@ -450,20 +436,16 @@ groupTypes.genfrac = function(group, options) {
...
-    const denomreset = makeSpan([style.reset(), dstyle.cls()],
-        [denom], newOptions);
+    newOptions = options.havingStyle(dstyle);
+    const denomm = buildGroup(group.value.denom, newOptions, options);
 
     let ruleWidth;
     if (group.value.hasBarLine) {
         ruleWidth = fontMetrics.metrics.defaultRuleThickness /
-            options.style.sizeMultiplier;
+            options.sizeMultiplier;
```
kevinbarabash: It might be better to call `makeLineSpan` here and then use its `height` below as opposed to calculating the `ruleWidth` here and inside of `makeLineSpan`.

src/buildHTML.js
```diff
@@ -770,18 +748,18 @@ groupTypes.spacing = function(group, options) {
 
 groupTypes.llap = function(group, options) {
     const inner = makeSpan(
-        ["inner"], [buildGroup(group.value.body, options.reset())]);
+        ["inner"], [buildGroup(group.value.body, options)]);
```
kevinbarabash: The `reset()` isn't necessary anymore here b/c `buildGroup` handles that now.

src/buildHTML.js
```diff
@@ -1034,54 +1008,51 @@ groupTypes.katex = function(group, options) {
         ["mord", "katex-logo"], [k, a, t, e, x], options);
 };
 
+const makeLineSpan = function(className, options) {
+    const baseOptions = options.havingBaseStyle();
+    const line = makeSpan(
+        [className].concat(baseOptions.sizingClasses(options)),
```
kevinbarabash: Are adding the sizing classes necessary here?  The span has no children and we're already setting its height explicitly.

src/buildHTML.js
```diff
@@ -1189,35 +1148,37 @@ groupTypes.sqrt = function(group, options) {
...
     // manually. Handle nested size changes.
     for (let i = 0; i < inner.length; i++) {
         const pos = utils.indexOf(inner[i].classes, "sizing");
         if (pos < 0) {
-            inner[i].classes.push("sizing", "reset-" + options.size,
-                                  group.value.size, style.cls());
-            inner[i].maxFontSize = fontSize;
-        } else if (inner[i].classes[pos + 1] === "reset-" + group.value.size) {
+            inner[i].classes.push("sizing", "reset-size" + baseOptions.size,
+                                  "size" + options.size);
```
kevinbarabash: Could we not using `options.sizingClasses(baseOptions)` here?

src/buildMathML.js
kevinbarabash: This should be checkable by rendering a expression with nested sizing commands and verify that the `mathsize` values in the MathML tree in the DOM match the sizes we use for same elements in visible HTML nodes that KaTeX renders.

src/buildHTML.js
```diff
@@ -1232,25 +1193,8 @@ groupTypes.styling = function(group, options) {
...
-        } else {
-            // This is a nested style change, as `\textstyle a\scriptstyle b`.
-            // Only override the old style (the reset class).
-            inner[i].classes[pos] = options.style.reset();
-        }
-    }
-
-    return new buildCommon.makeFragment(inner);
+    const newOptions = options.havingStyle(newStyle);
+    return sizingGroup(group.value.value, newOptions, options);
```
kevinbarabash: I really like how you unified the handle of style groups and sizing groups.

src/buildHTML.js
kohler: So, in the old code, `sub` and `sup` were not decorated with the classes necessary to change size, whereas `supmid` and `submid` were so wrapped.

In the new code, we don't need both versions, because we create the wrapped versions in one step. These are more like the `mid` versions, so I kept a suffix.

src/buildHTML.js
kohler: No: The `.reset()` isn't necessary any more because `Options` objects are immutable.

src/buildHTML.js
kohler: The sizing classes were on rules previously too, I think.

src/Options.js
kohler: Yes, good catch. parentStyle too.

src/buildHTML.js
kohler: OK

src/buildHTML.js
kohler: We could.

kohler:
Thanks for the comments, I checked in some changes.

The MathML values look correct to me in a couple of examples, so I think it's OK.

kohler:
Hi @kevinbarabash, any more comments or requests on this?

(I have some other changes blocked here.)

src/buildHTML.js
kevinbarabash: I was mixing up `options.reset()` with the `reset-` CSS classes.

kevinbarabash:
@kohler thanks for making those changes and for checking the MathML output.  The output looks good so if there are things that I've missed they must be minor.

PR #722, opened 2017-06-09
Update texcmp to ubuntu 17.04 and avoid mounted host directory

### LINKED ISSUE #708: texcmp woes

gagern:
* Switch to linux 17.04 in order to have a version of nodejs which understands ES6 syntax, for the sake of a consistent codebase.
* Avoid mounting the host directory, but use “docker cp” instead to transfer files between host and container.  This should avoid ownership and permission issues, and hopefully fixes #708.
* Support macros with positional arguments.
* Fix one overline example which caused LaTeX failure due to missing braces.

### DIFF
--- a/test/screenshotter/ss_data.yaml
+++ b/test/screenshotter/ss_data.yaml
@@ -95,7 +95,7 @@ LimitControls: |
     \displaystyle\int\limits_2^3 3x^2\,dx + \sum\nolimits^n_{i=1}i +
     \textstyle\int\limits_x^y z
 MathAtom: a\mathrel{\mathop{=}\limits^{\blue ?}}b
-MathAtom2: \mathop{\overline\mathrm{lim}}\limits_{x\to\infty}f(x)
+MathAtom2: \mathop{\overline{\mathrm{lim}}}\limits_{x\to\infty}f(x)
 MathDefaultFonts: Ax2k\breve{a}\omega\Omega\imath+\KaTeX
 MathBb: \mathbb{Ax2k\breve{a}\omega\Omega\imath+\KaTeX}
 MathBf: \mathbf{Ax2k\breve{a}\omega\Omega\imath+\KaTeX}
--- a/dockers/texcmp/Dockerfile
+++ b/dockers/texcmp/Dockerfile
@@ -1,11 +1,30 @@
-# convert from PDF to PNG with -flatten looks really bad on 14.04 LTS
-FROM ubuntu:15.04
+FROM ubuntu:17.04
 
 MAINTAINER Martin von Gagern <gagern@ma.tum.de>
 
-# Disable regular updates, but keep security updates
-RUN sed -i 's/^\(deb.*updates\)/#\1/' /etc/apt/sources.list && apt-get update
+ADD https://github.com/Yelp/dumb-init/releases/download/v1.2.0/dumb-init_1.2.0_amd64 /usr/bin/dumb-init
 
-# Install all required packages, but try not to pull in TOO much
-RUN apt-get -qy --no-install-recommends install \
-  texlive-latex-base etoolbox imagemagick ghostscript nodejs
+# Disable regular updates, keep security updates, avoid intermediate layers
+RUN sed -i 's/^\(deb.*updates\)/#\1/' /etc/apt/sources.list \
+ && apt-get update \
+ && apt-get upgrade -y \
+ && DEBIAN_FRONTEND=noninteractive \
+    apt-get install -qy --no-install-recommends \
+        ca-certificates \
+        etoolbox \
+        ghostscript \
+        imagemagick \
+        nodejs \
+        npm \
+        texlive-fonts-recommended \
+        texlive-latex-base \
+ && apt-get clean \
+ && rm -rf /var/lib/apt/lists/* \
+ && chmod +x /usr/bin/dumb-init
+
+ENTRYPOINT ["/usr/bin/dumb-init", "--"]
+
+RUN useradd --create-home --home-dir /KaTeX katex \
+ && mkdir -p /KaTeX/dockers/texcmp
+
+# USER katex # does not work well with docker cp copying stuff into our dirs
--- a/dockers/texcmp/Dockerfile
+++ b/dockers/texcmp/Dockerfile
@@ -25,6 +25,12 @@ RUN sed -i 's/^\(deb.*updates\)/#\1/' /etc/apt/sources.list \
 ENTRYPOINT ["/usr/bin/dumb-init", "--"]
 
 RUN useradd --create-home --home-dir /KaTeX katex \
- && mkdir -p /KaTeX/dockers/texcmp
+ && mkdir -p /KaTeX/dockers/texcmp /KaTeX/test/screenshotter
 
-# USER katex # does not work well with docker cp copying stuff into our dirs
+ADD package.json /KaTeX/dockers/texcmp/package.json
+
+RUN ( cd /KaTeX/dockers/texcmp; npm install; ) \
+ && ( cd /KaTeX/test/screenshotter; npm install js-yaml; ) \
+ && chown -R katex:katex /KaTeX
+
+USER katex
--- a/dockers/texcmp/package.json
+++ b/dockers/texcmp/package.json
@@ -3,6 +3,7 @@
   "description": "KaTeX helper to compare LaTeX output against screenshots",
   "license": "MIT",
   "dependencies": {
+    "ndarray": "1.0.18",
     "ndarray-fft": "1.0.0",
     "pngparse": "2.0.1",
     "q": "1.4.1"
--- a/dockers/texcmp/texcmp.js
+++ b/dockers/texcmp/texcmp.js
@@ -8,7 +8,7 @@ const path = require("path");
 const Q = require("q"); // To debug, pass Q_DEBUG=1 in the environment
 const pngparse = require("pngparse");
 const fft = require("ndarray-fft");
-const ndarray = require("ndarray-fft/node_modules/ndarray");
+const ndarray = require("ndarray");
 
 const data = require("../../test/screenshotter/ss_data");
 
@@ -68,6 +68,23 @@ function processTestCase(key) {
     }
     if (itm.post) {
         tex = tex + itm.post.replace("<br>", "\\\\");
+    }
+    if (itm.macros) {
+        tex = Object.keys(itm.macros).map(name => {
+            const expansion = itm.macros[name];
+            let numArgs = 0;
+            if (expansion.indexOf("#") !== -1) {
+                const stripped = expansion.replace(/##/g, "");
+                while (stripped.indexOf("#" + (numArgs + 1)) !== -1) {
+                    ++numArgs;
+                }
+            }
+            let args = "";
+            for (let i = 1; i <= numArgs; ++i) {
+                args += "#" + i;
+            }
+            return "\\def" + name + args + "{" + expansion + "}\n";
+        }).join("") + tex;
     }
     tex = template.replace(/\$.*\$/, tex.replace(/\$/g, "$$$$"));
     const texFile = path.join(tmpDir, key + ".tex");
--- a/dockers/texcmp/texcmp.js
+++ b/dockers/texcmp/texcmp.js
@@ -173,11 +173,11 @@ function processTestCase(key) {
         const uh = Math.max(browser.height + by, latex.height + ly); // u. h.
         return execFile("convert", [
             // First image: latex rendering, converted to grayscale and padded
-            "(", pngFile, "-grayscale", "Rec709Luminance",
+            "(", pngFile, "-colorspace", "Gray",
             "-extent", uw + "x" + uh + "-" + lx + "-" + ly,
             ")",
             // Second image: browser screenshot, to grayscale and padded
-            "(", browserFile, "-grayscale", "Rec709Luminance",
+            "(", browserFile, "-colorspace", "Gray",
             "-extent", uw + "x" + uh + "-" + bx + "-" + by,
             ")",
             // Third image: the per-pixel minimum of the first two images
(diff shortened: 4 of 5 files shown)


test/screenshotter/ss_data.yaml
```diff
@@ -95,7 +95,7 @@ LimitControls: |
     \displaystyle\int\limits_2^3 3x^2\,dx + \sum\nolimits^n_{i=1}i +
     \textstyle\int\limits_x^y z
 MathAtom: a\mathrel{\mathop{=}\limits^{\blue ?}}b
-MathAtom2: \mathop{\overline\mathrm{lim}}\limits_{x\to\infty}f(x)
+MathAtom2: \mathop{\overline{\mathrm{lim}}}\limits_{x\to\infty}f(x)
```
edemaine: I assume this is unrelated to the rest of the PR, but just a general bug fix?  I can confirm that the old version doesn't work on LaTeX.

gagern: I guess so. I wonder why the old version works at all in KaTeX. Suggests something is wrong there. Might be worth having a separate ticket for that.

gagern:
The containers not having network access sounds like a problem with docker config. @edemaine, can you try creating a container using `docker run --rm -it ubuntu:17.04 /bin/bash` the exploring its network config? Can you do "apt-get update" therein? What does /etc/resolv.conf look like? Can you reach hosts by IP address instead of host name? Does the internet have any suggestions?

Also check out https://docs.docker.com/engine/userguide/networking/ and its subpages. What kind of networking does your container use? `docker inspect ‹containerid›` should be useful here. On my system the default is to use a bridged network, but I guess it might be different for you. https://docs.docker.com/engine/userguide/networking/default_network/configure-dns/ has details about dns configuration in a bridged network.

edemaine:
All I did was `apt-get install docker` on Ubuntu, so it's not like I'm running a weird Docker config. 😄 (I'm definitely a Docker noob.)  I find it weird that `screenshotter.sh` works fine ... ah, I see, that one doesn't have to download any custom software, just an off-the-shelf selenium image?

As you guessed, within `docker run --rm -it ubuntu:17.04 /bin/bash`, `apt-get update` fails.  Here's what `docker inspect` says about network settings while it's running:

```
    "NetworkSettings": {
        "Bridge": "docker0",
        "Gateway": "172.17.42.1",
        "GlobalIPv6Address": "",
        "GlobalIPv6PrefixLen": 0,
        "IPAddress": "172.17.0.94",
        "IPPrefixLen": 16,
        "IPv6Gateway": "",
        "LinkLocalIPv6Address": "fe80::42:acff:fe11:5e",
        "LinkLocalIPv6PrefixLen": 64,
        "MacAddress": "02:42:ac:11:00:5e",
        "PortMapping": null,
        "Ports": {}
    },
```

`/etc/resolv.conf` looks identical to the host machine.  I tried changing `/etc/apt/sources.list` to use a manual IP (`deb http://91.189.88.149/ubuntu/ zesty main restricted`) but it still won't connect.  So it seems it's really the bridge network that's not working.

Apparently my version of Docker has no `docker network` command, so I'm guessing it's old: 1.6.2.  I guess I should try updating, though it'd be nice to solve this for others on Ubuntu 14.04.5 LTS.

edemaine:
OK, I installed the latest version of Docker (via [these instructions](https://docs.docker.com/engine/installation/linux/ubuntu/#install-using-the-repository)) and now things seem to run, except I think it's not properly extracting the computed file to the host filesystem:

```
Typeset Aligned
Rasterized Aligned
Positioned Aligned: -82, 19
Compared Aligned
+ docker cp a2d1feb26fbc0c6b19b6572a741a7577f80e9fb45139d164c169acf76ca2a912:/KaTeX/test/screenshotter/tex -
+ cd test/screenshotter/tex
+ tar xv
tex/
tex/Aligned-pdflatex.png
tar: tex/Aligned-pdflatex.png: Cannot change ownership to uid 0, gid 0: Operation not permitted
tar: tex: Cannot change ownership to uid 0, gid 0: Operation not permitted
tar: Exiting with failure status due to previous errors
+ exit 2
+ cleanup
+ [[ -n a2d1feb26fbc0c6b19b6572a741a7577f80e9fb45139d164c169acf76ca2a912 ]]
+ false
+ container=
```

Probably because of the original issue, that I'm using AFS, where chown is impossible.  Maybe just some `tar` tweaks would suffice?

gagern:
> All I did was `apt-get install docker.io` on Ubuntu, so it's not like I'm running a weird Docker config.

I think Docker changed their default config several times, usually only affecting new installs. And an Ubuntu which runs on AFS is definitely far from common, right? My `docker inspect` reads like this:

        "NetworkSettings": {
            "Bridge": "",
            "SandboxID": "72…96",
            "HairpinMode": false,
            "LinkLocalIPv6Address": "",
            "LinkLocalIPv6PrefixLen": 0,
            "Ports": {},
            "SandboxKey": "/var/run/docker/netns/72…",
            "SecondaryIPAddresses": null,
            "SecondaryIPv6Addresses": null,
            "EndpointID": "25…9b",
            "Gateway": "172.17.0.1",
            "GlobalIPv6Address": "",
            "GlobalIPv6PrefixLen": 0,
            "IPAddress": "172.17.0.2",
            "IPPrefixLen": 16,
            "IPv6Gateway": "",
            "MacAddress": "02:…:02",
            "Networks": {
                "bridge": {
                    "IPAMConfig": null,
                    "Links": null,
                    "Aliases": null,
                    "NetworkID": "bf…95",
                    "EndpointID": "25…9b",
                    "Gateway": "172.17.0.1",
                    "IPAddress": "172.17.0.2",
                    "IPPrefixLen": 16,
                    "IPv6Gateway": "",
                    "GlobalIPv6Address": "",
                    "GlobalIPv6PrefixLen": 0,
                    "MacAddress": "02:…:02"
                }
            }
        }

So I have this whole `Networks` key which appears to be missing in your setup. The `NetworkID` listed in my setup is also found in the output of `docker network ls`. If you don't even have `docker network`, then perhaps that setting key wasn't meant to be present in your setup either.

> tar: tex: Cannot change ownership to uid 0, gid 0: Operation not permitted

Are you running the command as root? Usually you run `docker` as a regular user, but with `docker.sock` only accessible to the group `docker` so that only users within that group can run `docker`. If you are running the commands as root, then `tar` might assume it *should* be able to “restore” file ownership, and complain for failing to do so.

> […] I'm using AFS, where chown is impossible. Maybe just some tar tweaks would suffice?

The `--no-same-owner` option to `tar` should help. Perhaps combined with `--no-same-permissions`. I fear that permanently adding these flags may cause compatibility problems with tar versions that don't support them, so we should add some checks to verify they are actually supported.

edemaine:
Yes, I'm running as root via `sudo`.  My impression is that running `docker` as a regular user is just as risky as running it as root (though maybe hackers would need to work slightly harder...).

Adding `--no-same-owner` to the tar arguments (after `xv`) was enough for me.  According to https://github.com/npm/npm/issues/1128, this is equivalent to `-o` that works on all (?) tars.  I confirmed that `-o` after `xv` also suffices for me.

gagern:
http://pubs.opengroup.org/onlinepubs/007908799/xcu/tar.html suggests that indeed `-o` is a pretty established command line flag. I added that. Thanks for testing and suggesting that!

edemaine:
Ah, there seems to be a bug in the `tar` calls though.  `test/screenshotter` has a directory `tex` with a *sub*directory `tex` in it, with the rendered `.png` file.  Also `test/screenshotter` has a directory `diff` with a *sub*directory `diff` in it, with another `.png` file.  But all previous files have been removed!  I can't say I like this, as I usually just render one example at a time, but I want my previous renders too, even if out-of-date...

edemaine:
Is there a reason we're using `tar` at all?  Doesn't `[docker cp`](https://docs.docker.com/engine/reference/commandline/cp/#extended-description) support a host destination path?

gagern:
I thought `docker cp` had problems copying more than one file, but it seems to handle whole folders well enough, so I guess we can do without tar on the way out. Building the image I think the tar command is a nice way of copying some files without taking all of them, so there I'd keep it.

I'm wondering whether I should move things like the "npm install" to the image creation, so we can avoid that step on consecutive runs.

edemaine:
> I'm wondering whether I should move things like the "npm install" to the image creation, so we can avoid that step on consecutive runs.

That sounds nice -- it would speed-up startup.  I guess the worry is that `texcmp` changes and we'd need to delete the image?  (I don't even know how to delete the image...)

I wonder if we could rely on the caller running `npm install`, and just copy the `node_modules` directory into the Docker (each call)?  And maybe running `npm install` manually if `node_modules` doesn't yet exist.  So we still cover the common case (no changes to texcmp) being automatic, but it's also easy for someone to update -- just run `npm install` before `texcmp.sh`.

gagern:
> That sounds nice – it would speed-up startup.

For repeated evaluation it would be even better if we could have some texcmp daemon in the background, and could send that rendering requests on some port in a proper server and client kind of way. But probably too much effort for such a rare use case. Note that screenshotter allows reusing selenium containers if you call the jar file directly, not through the bash file.

> I guess the worry is that `texcmp` changes and we'd need to delete the image?

Not necessarily. We'd only have to redo the image if the set of dependencies for `texcmp` changes.

> (I don't even know how to delete the image...)

`docker rmi`, but I try to use the version number in the tag in such a way that whenever I push a change that results in a different image, it gets a new version number and the script will building new images automatically. And if the lower layers remain the same, the resources required for this should be moderate. In the future I should also bump the version when I change the dependencies.

We might want to ship these images from docker hub one day, so people can pull them instead of building them locally.

> I wonder if we could rely on the caller running `npm install`, and just copy the `node_modules` directory into the Docker (each call)?

That's essentially how it worked with the mounted host directory. But some of the modules (either ones we are using now or some we might want to use in the future) could install different stuff depending on the node version in use. I already noticed that the different layout of `node_modules` (nested vs. flat) caused me some grief. So I feel like I'd rather not depend on this too much.

edemaine:
Makes sense.  Putting `npm install` inside the docker build sounds reasonable given the version numbering.  But it's also not bad as is.  (Screenshotter is the really slow one.)

edemaine:
Confirmed that new code works: created new docker (though cached so still pretty fast) and ran `npm install`; and second run was faster, no `npm install`.  Cool!

edemaine:
(Let me know if you're done and I can merge, or feel free to do so yourself.)

gagern:
I think this should be ready for merging now. And I think we can squash it without loosing too much detail. Feel free to do the honors.

Issue #687, opened 2017-04-29
support \setlength and measurement constants


kevinbarabash:
Add a dictionary that tracks the values of
- thinmuskip
- medmuskip
- thickmuskip
- jot

edemaine:
Would be also nice to support `\arraystretch` (see [buildHTML.js](https://github.com/Khan/KaTeX/blob/38ba9f91877cdbc3b687de16803c41d6fffafb74/src/buildHTML.js#L573-L575)) and `\arraycolsep` (in the same buildHTML function, though more spread out).

edemaine:
[Some fun examples of playing with muskip](https://tex.stackexchange.com/questions/83746/keeping-the-distance-between-mathematical-symbols-consistent).  Though that's mainly relevant to `plus` and `minus` settings which we don't currently support, we'll need to keep in mind that we really ought to support things like `\setlength\arraycolsep{0.5\arraycolsep}` (something I do a lot), and possibly `\addtolength\arraycolsep{-0.5em}`.

There's also a question of whether we'd want to support TeX-style `\arraycolsep=0.5\arraycolsep` and especially `\global\arraycolsep=0.5\arraycolsep`.  If we support global and local setting, we might need another data structure.

But even focusing on support for local lengths, I'm not quite sure where this ought to go.  Perhaps it could go in the `options` dictionary in buildHTML, e.g. as a `lengths` dictionary within?  We'll need to parse `\setlength` as defining an implicit group, so that e.g. the following works:

```latex
\setlength\arraycolsep{1em} \begin{array} ... \end{array}
\setlength\arraycolsep{2em} \begin{array} ... \end{array}
```

kevinbarabash:
```
options.length = {
   thinmuskip: '',
   medmuskip: '',
   thickmuskip: '',
   arraycolsep: '',
   jot: '',
   ...,
}
```
Is the implicit group really necessary?  Since the commands listed above are all siblings, they're processed serially.  The `\setlength` commands should update `options`, but not return a group.  Is that what you meant by an "implicit group"?

edemaine:
Ah, I see. So you imagine a `setlength` parse node that specifies that a length change in (the options for) the parent group.  Then all following items in the group will use the new length, until/unless the length gets overridden later.  This is a little tricky, I guess, because the processing of `setlength` nodes needs to happen at the parent level...  (By contrast, what I was imagining was that the modified items in the group would become children of the `setlength` node...)

[...]

Issue #682, opened 2017-04-27
Need `gathered` environment


konn:
LaTeX has a `gather` and `gather*` environments (and also `gathered`, which is analogous to what `aligned` is for `align`) for center-aligned (or un-aligned) multiline equations, but they seems absent in KaTeX.

edemaine:
Thanks, I didn't know about amsmath's `gather`/`gather*`/`gathered`.  KaTeX ought to support `gathered`, whereas `gather`/`gather*` would belong in a text-level wrapper (not yet written) as described in #603.

Intuitively, `\begin{gathered} ... \end{gathered}` seems similar to `\begin{array}{c} ... \end{array}`, so I ran a test:

```latex
$$\begin{gathered} \text{gathered}   \\ x^2+y^2=z^2 \\ x < y \end{gathered} =
  \begin{array}{c} \text{array\{c\}} \\ x^2+y^2=z^2 \\ x < y \end{array}
$$
```
![gathered](https://cloud.githubusercontent.com/assets/2218736/25488183/63c76b3a-2b34-11e7-8cbf-fc8fca8c1c11.png)

The only difference seems to be in the line spacing.  This shouldn't be hard to implement, provided we can figure out what the correct spacing is...  [amsmath](http://mirror.jmu.edu/pub/CTAN/macros/latex/required/amsmath/amsmath.pdf) seems to implement this via `\spread@equation` which I've yet to grok.

kevinbarabash:
I'm guessing the `gathered` implementation changes the line spacing "constant" temporarily while in the environment.  We should add a dictionary for line spacing and other constants using `\setlength`, e.g. `\setlength{\jot}{10pt}` changes the line spacing... then `gathered` could be implemented using a macro.

kevinbarabash:
I've opened an issue for `\setlength` support, see https://github.com/Khan/KaTeX/issues/687.

arnog:
The extra space between rows in the `gathered` environment is the same extra spacing from the `aligned` environment, introduced by the `\jot` command (3pt). 

There's another important difference between `gathered` and `array{x}`, the cells of the former are in `displaystyle` while the cells of the later are in `textstyle`. See for example:
```
\begin{gathered} \text{gathered}   \\ \frac12 x^2+y^2=z^2 \\ x < y \end{gathered}
     =
  \begin{array}{c} \text{array\{c\}} \\ \frac12 x^2+y^2=z^2 \\ x < y \end{array}
```

[...]

PR #725, opened 2017-06-10
Add \jot lineskip to aligned environment, switch contents to displaystyle, and add gathered

### LINKED ISSUE #686: Spacing between rows of an aligned environment too tight
The gap between rows of an array is smaller than what TeX is producing.

**KaTeX**
screen shot 2017-04-27 at 4 46 22 pm
**LaTeX**
screen shot 2017-04-27 at 4 46 29 pm

This might be because line 602 of buildHTML.js should be
`        let gap = arstrutDepth;
`
instead of
`        let gap = 0;
`

edemaine:
AMS environments `aligned` and `gathered` (not yet implemented in KaTeX, see #682) use `\openup\jot` to increase `\baselineskip` by `\jot` (which defaults to 3pt, see #687 for overriding).  This lineskip in turns defines `\strutbox` in the `array` environment, which then gets scaled by `\arraystretch`.  (As reported in #686, which this PR fixes.)

I've added a `rowsAreLines` option to `array` objects in the parse tree that encourages the outputter (here, buildHTML) to add `\jot` (currently hard-coded to 3pt, see #687) to each line.  This seemed simpler than e.g. manually putting the equivalent of 3pt in the environment definition, and will be useful when we implement `gathered` in #682.

`texcmp Aligned` before this change:

aligned_before

`texcmp Aligned` after this change:

aligned_after

Note that other environments like `array` remain unaffected, as they were already correct (vertically anyway):

arrays


edemaine:
While I'm here, I tweaked the definition of the `aligned` environment to put all contents in `\displaystyle`.  This feature was pointed out by @arnog in #682, and I confirmed that it's the case in LaTeX.  For example, `$\begin{aligned} \frac{x}{y} \end{aligned}$` now uses a display fraction, as it does in LaTeX.

I'm going to add the `gathered` environment next, to this PR, as it's almost trivial at this point.

edemaine:
I added the `gathered` environment, fixing #682.  The test case (with `\begin{gathered}` on the left and `\begin{array}{c}` on the right) renders pretty well, but not quite as perfectly:

gathered

Let me know if you can see where the extra space is coming from in LaTeX...

src/buildHTML.js
```diff
@@ -592,6 +592,9 @@ groupTypes.array = function(group, options) {
 
     // Vertical spacing
     const baselineskip = 12 * pt; // see size10.clo
+    // Default \jot from ltmath.dtx
+    // TODO(edemaine): allow overriding \jot via \setlength (#687)
```
kevinbarabash: nice TODO

kevinbarabash:
whoops... I only meant to "comment" not "close and comment".

src/environments.js
```diff
@@ -216,13 +216,20 @@ defineEnvironment("aligned", {
     let res = {
         type: "array",
         cols: [],
+        rowsAreLines: true,
     };
-    res = parseArray(context.parser, res);
+    res = parseArray(context.parser, res, "display");
```
kevinbarabash: Is the `aligned` environment also supposed to be in \displaystyle only?

src/environments.js
```diff
@@ -216,13 +216,20 @@ defineEnvironment("aligned", {
...
+    // Count number of columns = maximum number of cells in each row.
+    // At the same time, prepend empty group {} at beginning of every second
+    // cell in each row (starting with second cell) so that operators become
+    // binary.
```
kevinbarabash: This is so that operators at the start of cell will have the proper spacing around them.  What about operators at the end of a cell, or does that not happen in practice?

src/buildHTML.js
```diff
@@ -632,6 +635,12 @@ groupTypes.array = function(group, options) {
                 gap = 0;
             }
         }
+        // In AMS multiline environments such as aligned and gathered, rows
+        // correspond to lines that have additional \jot added to the
+        // \baselineskip via \openup.
+        if (group.value.rowsAreLines) {
```
kevinbarabash: The naming of this variable is confusing to me.  Aren't rows already lines?  This just adds more space between the rows/lines.  Maybe change this to `addJotBetweenRows`.

src/environments.js
edemaine: I just added documentation here, because it took me a long time to figure out why the existing code was adding empty groups.  I checked and this is indeed how it's implemented in amsmath, [specifically in `\start@aligned`](http://tug.ctan.org/macros/latex/required/amslatex/math/amsmath.pdf#page=46).

src/environments.js
edemaine: Yes, I checked in LaTeX, and `$\begin{aligned} \frac{x}{y} \end{aligned}$` renders a display fraction.

src/buildHTML.js
edemaine: Agreed, that is confusing.  Renamed to `addJot`, which seems clear enough (as `\jot` is an interline spacing).

edemaine:
Comments addressed in https://github.com/Khan/KaTeX/pull/725/commits/3c00b27f263e3d6754114defe1cbb4350ad6299f

kevinbarabash:
@edemaine did you rebase and run `make screenshots`?  I would've expected the LimitControls screenshots to change, see https://github.com/Khan/KaTeX/pull/718/files#diff-e55441e2f8a6090eac1fdb9e12ed86ed.

edemaine:
@kevinbarabash Yes, I rebased.  Almost no screenshot tests involve `aligned` or `gathered` environments --
 just the old `Aligned` test (modified in this PR) and the new `Gathered` test (new to this PR).  Is there a reason `LimitControls` should have changed that I'm missing?

kevinbarabash:
@edemaine sorry... I wasn't looking at things carefully.  I was comparing the ascent of glyphs when I should've been looking at the baselines.  Also, for some reason I thought this was a diff that was opened before the #718, but it was actually only opened a few hours ago.

kevinbarabash:
Thanks for the PR.  Awesome to see how quickly new environments can be added.

PR #727, opened 2017-06-10
Implement \coloneqq, \colonequals, etc. based on mathtools and colonequals

### LINKED ISSUE #726: \mathop{x} should center single-character x
`\mathop` is a TeX feature, and it has some subtlety in its implementation described in [Documentation of \mathop](https://tex.stackexchange.com/questions/368673/documentation-of-mathop/368700).

Specifically, when the argument to `\mathop` is a single character, it should get centered vertically.  Perhaps this could be implemented by tweaking buildHTML's `groupTypes.mclass` (when `group.value.mclass === 'mop'`), e.g. using the in-process `\raisebox` implementation (#685).

It's a minor bug, but comes up in defining custom symbols such as #657.

edemaine:
I implemented the following commands from `mathtools.sty`, using macros with exactly their definitions in `mathtools.sty`:

```latex
\begin{matrix}
x\vcentcolon x \dblcolon x \coloneqq x \coloneq x \Coloneqq x \Coloneq x \\
x \eqqcolon x \Eqqcolon x \eqcolon x \Eqcolon x \\
x \colonapprox x \colonsim x \Colonapprox x \Colonsim x
\end{matrix}
```

coloneqq

I also implemented the following commands from `colonequals.sty`, as aliases to the `mathtools.sty` definitions, and adding a couple analogous definitions when the corresponding symbol wasn't already in `mathtools.sty`:

```latex
\begin{matrix}
x\ratio x \coloncolon x \colonequals x \coloncolonequals x \equalscolon x \equalscoloncolon x \\
x \colonminus x \coloncolonminus x \minuscolon x \minuscoloncolon x \\
x \colonapprox x \colonsim x \coloncolonapprox x \coloncolonsim x \\
x \simcolon x \simcoloncolon x \approxcolon x \approxcoloncolon x
\end{matrix}
```

colonequals

The colon will get more centered once we fix #726, or via manual tweaking via `\raisebox` (#685).  But that is independent of this PR, which at least makes the commands available now, as suggested in #657.

### DIFF
--- a/src/macros.js
+++ b/src/macros.js
@@ -34,3 +34,63 @@ defineMacro("\\iff", "\\;\\Longleftrightarrow\\;");
 defineMacro("\\implies", "\\;\\Longrightarrow\\;");
 defineMacro("\\impliedby", "\\;\\Longleftarrow\\;");
 
+//////////////////////////////////////////////////////////////////////
+// mathtools.sty
+
+//\providecommand\ordinarycolon{:}
+defineMacro("\\ordinarycolon", ":");
+//\def\vcentcolon{\mathrel{\mathop\ordinarycolon}}
+//TODO(edemaine): Not yet centered. Fix via \raisebox or #726
+defineMacro("\\vcentcolon", "\\mathrel{\\mathop\\ordinarycolon}");
+// \providecommand*\dblcolon{\vcentcolon\mathrel{\mkern-.9mu}\vcentcolon}
+defineMacro("\\dblcolon", "\\vcentcolon\\mathrel{\\mkern-.9mu}\\vcentcolon");
+// \providecommand*\coloneqq{\vcentcolon\mathrel{\mkern-1.2mu}=}
+defineMacro("\\coloneqq", "\\vcentcolon\\mathrel{\\mkern-1.2mu}=");
+// \providecommand*\Coloneqq{\dblcolon\mathrel{\mkern-1.2mu}=}
+defineMacro("\\Coloneqq", "\\dblcolon\\mathrel{\\mkern-1.2mu}=");
+// \providecommand*\coloneq{\vcentcolon\mathrel{\mkern-1.2mu}\mathrel{-}}
+defineMacro("\\coloneq", "\\vcentcolon\\mathrel{\\mkern-1.2mu}\\mathrel{-}");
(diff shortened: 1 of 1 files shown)


edemaine:
This should wait for #751 so I can remove the braces.

kevinbarabash:
#751 has been merged.

edemaine:
Rebased and removed braces.  This should be good to go, unless we want to add tests.

kevinbarabash:
I think this is okay to merge without screenshot tests.  There are already a lot of commands which don't have screenshots.  CONTRIBUTING.md provides this guideline:

> If you add a feature that is dependent on the final output looking the way you created it, add a screenshot test.

The word "feature" is a little ambiguous.  It would nice if we had a more explicit guidelines in this area.  I think symbols by themselves probably aren't worth having a screenshot test.  Commands that affect which glyphs are display such as font commands could probably be tested by snapshot tests and probably don't need screenshots.  Same with handling unicode.

Things that do require screenshots are features that implement a specific layout, e.g. array, *matrix, stackrel, mathop, *lap, frac, not, etc.

If someone finds an issue with any of these commands then either the macro is incorrect or one of the commands in the macro isn't behaving correctly and should be fixed.  When it's fixed, we should improve the screenshot test for that command.

kevinbarabash:
We could also add a lot of unit tests to katex-spec.js, but I'd rather be more targeted in our testing and focus on filling in some of the hole in our test coverage.

edemaine:
I guess we need to add these to [function support](https://khan.github.io/KaTeX/function-support.html).  How does this work?  @ronkok do you have time to do this, or would you rather delegate?

ronkok:
I already have that page ready and I plan to make a PR as soon as release 0.8.0 lands. For a advance look, I've put a temporary copy [here](http://copyasmarkup.com/katex/function-support.html). 

Let me know if you see that I've missed anything or if you notice some major problem with organization.

edemaine:
@ronkok Beautiful!!

ronkok:
@kevinbarabash, regarding all the new colonequals functions, a substitute for screenshots is to review them on the function support page, which has a temporary copy [here](http://copyasmarkup.com/katex/function-support.html#relations).

kevinbarabash:
@ronkok LGTM.  We'll probably want to update the function support page after we do a release.

PR #731, opened 2017-06-14
Make auto-render example file run in-place in repository

edemaine:
I ran into this when trying to test #656.  This will make future autorender tests easier, but also make it easier for someone checking out the repo to just try out the example.  I don't think it breaks any other settings for the HTML, but please correct me if I'm wrong (e.g. if it's accessible on `https://khan.github.io/KaTeX/` somehow).

I can't imagine it ever working as written, in particular because `auto-render.js` uses `require`, so it can't run directly in a browser -- you need to run the compiled version that ends up in `dist`.

### DIFF
--- a/contrib/auto-render/index.html
+++ b/contrib/auto-render/index.html
@@ -3,9 +3,9 @@
   <head>
     <meta charset="UTF-8">
     <title>Auto-render test</title>
-    <script src="/katex.js" type="text/javascript"></script>
-    <link href="/katex.css" rel="stylesheet" type="text/css">
-    <script src="./auto-render.js" type="text/javascript"></script>
+    <script src="../../dist/katex.js" type="text/javascript"></script>
+    <link href="../../dist/katex.css" rel="stylesheet" type="text/css">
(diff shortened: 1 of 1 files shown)


gagern:
The file as it stands is intended to be opened by the KaTeX server. Run “make serve” and open http://0.0.0.0:7936/contrib/auto-render/index.html to see it working in the current version. Running HTML files from a local directory is problematic as modern browsers (Chrome in particular) don't allow access to resources like fonts from local disk. Thus we should design all our examples to use the HTTP server. Perhaps we might want to add a comment to that effect somewhere.

xymostech:
+1 to everything @gagern said. There's [a rule in server.js](https://github.com/Khan/KaTeX/blob/master/server.js#L67) for serving auto-render.js as a browserified bundle, which is why the `require()` works. So I don't think we need these changes. That said, it would probably be helpful to add to the documentation about how you're supposed to view the examples for future devs.

edemaine:
Oops, thanks -- that makes sense in hindsight!  I'll add documentation to this effect in a different PR.

Issue #377, opened 2015-11-01
Deprecate IE8 support, based on Microsoft's January 2016 hard deadline.


Pomax:
Based on https://support.microsoft.com/en-us/gp/microsoft-internet-explorer as of January 12, 2016, only IE9 and up will be supported for consumer versions of Windows (IE7/8 will still be supported for very old versions of Windows Embedded, but this is not a platform that KaTeX targets).

This is a hard deadline and a shift in how Microsoft will be supporting browsers going forward (only officially supporting whatever is the latest version for each flavour of Windows) and as such testing with, and support for, IE8 can be deprecrated. 

(Once Vista hits its EOL, IE9 can also be deprecated, but that won't officially be until April 2017)

gagern:
Did KaTeX even work in IE8? https://github.com/Khan/KaTeX/issues/263#issuecomment-128927322 suggests otherwise, but I didn't confirm that suggestion. In any case, the browserstack results discussed there might be useful in formulating official support statements.

spicyj:
It used to, at least. KA has since dropped support for IE8 so it's possible it regressed and no one noticed.

xymostech:
I'm pretty sure it regressed when we introduced MathML support, which was a while ago. There might be a way to bring it back if we really care, but unless we actuallly set up tests it'll probably break again.

kevinbarabash:
@Pomax is there specific work that you think should be done for this?

Pomax:
not really, the docs said KaTeX supported IE8, which sounded like there'd be unnecessary shim code at least in one place, but I've not looked at the codebase in a really long time now >_>

kevinbarabash:
I guess we should update the docs at the very least to reflect that IE8 support no longer works.

PR #737, opened 2017-06-18
Show correct browser support

ronkok:
KaTeX runs in IE 9 but not in IE 8. This PR edits the README and
Contributing files to reflect that fact.

Resolves issue #377.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -92,10 +92,10 @@ created it, add a screenshot test. See
 
 #### Testing in other browsers
 
-KaTeX supports all major browsers, including IE 8 and newer. Unfortunately, it
+KaTeX supports all major browsers, including IE 9 and newer. Unfortunately, it
 is hard to test new changes in many browsers. If you can, please test your
 changes in as many browsers as possible. In particular, if you make CSS changes,
-try to test in IE 8, using [modern.ie](http://modern.ie) VMs.
+try to test in IE 9, using [modern.ie](http://modern.ie) VMs.
 
 ## Style guide
 
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@ KaTeX is a fast, easy-to-use JavaScript library for TeX math rendering on the we
  * **Self contained:** KaTeX has no dependencies and can easily be bundled with your website resources.
  * **Server side rendering:** KaTeX produces the same output regardless of browser or environment, so you can pre-render expressions using Node.js and send them as plain HTML.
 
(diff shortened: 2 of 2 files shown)


edemaine:
This looks great!  Do you think we should list Edge as well, given that that's replacing IE?  I just did a simple test, and it looks OK.  [Your stretchy wide test page](http://copyasmarkup.com/katex/wideelements.html) doesn't fare quite as well -- maybe you know about this already -- but the scaled slashes look ugly for some reason.

ronkok:
I think that Microsoft intends, for Windows 10+, to make Edge an evergreen browser similar to Chrome or Firefox. Since neither of README or Contributing address the other evergreens, I think it makes sense to leave Edge out. I don't feel strongly about that.

Regarding stretchy wide elements, yes, both IE and Edge have problems. The browser support is detailed [here](http://copyasmarkup.com/katex/wideelements.html#browser-support).  

For the `\cancel` series, Edge needs to fix a bug with CSS `calc()` in a `background: linear-gradient`. You can see the bug if you use Edge to go [here](http://copyasmarkup.com/bug/edgeproblems.html) and scroll down a bit.  I made my complaint [here](https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/11326126/). Feel free to add your voice.  In the meantime, the master's fallback uses an SVG, whose lines vary with span size.

And Edge needs to support CSS `mask` before it will support `\color`. I made my complaint [here](https://wpdev.uservoice.com/forums/257854-microsoft-edge-developer/suggestions/6261341-css-masking).

I have a new draft of the [function support page](https://khan.github.io/KaTeX/function-support.html) ready for KaTeX 0.8.0. It includes a note about `\color` and Edge.

kevinbarabash:
I someone has access to a Windows box can they post screenshots of what the issues look like on Edge?

ronkok:
Edge does not support CSS `mask`, so the `\color` fallback is black.
color

The `calc` bug would have done terrible things to `\cancel`, had I not corrected it:
calc

... so the `\cancel` fallback is an SVG, which cannot take color and changes line thickness proportional to span size:
cancel

edemaine:
Cool, glad to hear it's just the same issues you've already worked around.

Regarding evergreen browsers: there is the line `KaTeX supports all major browsers, including Chrome, Safari, Firefox, Opera, and IE 9 - IE 11` (which you edited).  I was wondering whether we should add Edge to that list.

ronkok:
@edemaine , Good point. I've added Edge to the list in README.

edemaine:
Great!  There's apparently a merge conflict now... perhaps you merged with an older version? (I would use `git fetch upstream` then `git rebase -i upstream/master`, if you've set `upstream` to track this repo.)

kevinbarabash:
Looks like the merge conflict is gone so I'm going to merge this.

kevinbarabash:
@ronkok thanks for updating the docs.

PR #740, opened 2017-06-23
Add documentation for how to run auto-render example

edemaine:
Replacement for #731.

Feel free to suggest improved wording! (or complain about duplication -- but I was hoping to catch people who find the example in a couple of different ways)


kevinbarabash:
Thanks for improving the auto-render docs.

PR #743, opened 2017-06-28
Associate font metrics with Options, not Style.

kohler:
Font metrics are associated with a given font and size combination.
Before KaTeX understood sizing commands, sizes were associated with
a Style. That's not true now. So instead of `style.metrics`, use
`options.fontMetrics()`, since `options` knows the font and the
size.

This is a cleanup commit with no direct effect; it will make
other changes possible later.


src/fontMetrics.js
```diff
@@ -55,49 +53,28 @@ const sigmas = {
...
-// This value determines how large a pt is, for metrics which are defined in
-// terms of pts.
-// This value is also used in katex.less; if you change it make sure the values
-// match.
-const ptPerEm = 10.0;
+    // These font metrics are extracted from TeX by using tftopl on cmex10.tfm;
+    // they correspond to the font parameters of the extension fonts (family 3).
+    // See the TeXbook, page 441. In AMSTeX, the extension fonts scale; to
+    // match cmex7, we'd use cmex7.tfm values for script and scriptscript
+    // values.
```
kevinbarabash: I'm surprised that the third value in each of the arrays isn't determined by `cmex5` similar to how `cmsy5` is used for the values above.

src/fontMetrics.js
```diff
@@ -55,49 +53,28 @@ const sigmas = {
...
-    bigOpSpacing2: xi10,
-    bigOpSpacing3: xi11,
-    bigOpSpacing4: xi12,
-    bigOpSpacing5: xi13,
-    ptPerEm: ptPerEm,
-    doubleRuleSep: doubleRuleSep,
+    // The space between adjacent `|` columns in an array definition. From
+    // `\showthe\doublerulesep` in LaTeX. Equals 2.0 / ptPerEm.
+    doubleRuleSep: [0.2, 0.2, 0.2],
 };
```
kevinbarabash: Now that this object also includes `xi`s as well as `sigma`s maybe we should change the name of it.  Maybe `fontMetrics`.

src/fontMetrics.js
```diff
@@ -271,8 +248,33 @@ const getCharacterMetrics = function(character, style) {
...
+    } else {
+        sizeIndex = 2;
+    }
+    if (!fontMetricsBySizeIndex[sizeIndex]) {
+        const metrics = fontMetricsBySizeIndex[sizeIndex] = {};
+        for (const key in sigmas) {
+            if (sigmas.hasOwnProperty(key)) {
+                metrics[key] = sigmas[key][sizeIndex];
+            }
+        }
```
kevinbarabash:
```
Object.keys(sigmas).forEach(key => metrics[key] = sigmas[key][sizeIndex])
```
is another possibility.  I'm okay either way.

src/fontMetrics.js
```diff
@@ -1,6 +1,3 @@
-/* eslint no-unused-vars:0 */
-
-const Style = require("./Style");
```
kevinbarabash: Nice.  I ended up removing this dep in another PR.  I missed the comment at the top.

src/Options.js
```diff
@@ -181,6 +185,16 @@ Options.prototype.baseSizingClasses = function() {
 };
 
 /**
+ * Return the font metrics for this size.
+ */
+Options.prototype.fontMetrics = function() {
+    if (!this._fontMetrics) {
+        this._fontMetrics = fontMetrics.getFontMetrics(this.size);
```
kevinbarabash: When I first looked at this I thought this might create a lot of objects but then I remembered that `getFontMetrics` also stores a copy of each unique object it returns which means there's a small number of objects being created.

src/Options.js
```diff
@@ -24,6 +26,7 @@ const sizeStyleMap = [
 ];
 
 const sizeMultipliers = [
+    // fontMetrics.js:getFontMetrics also uses size indexes.
```
kevinbarabash: I'm not sure what this comment means.

src/fontMetrics.js
kohler: There is no `cmex5`. AMSTeX supplies `cmex7`, `cmex8`, `cmex9` to TeX's `cmex10`.

src/Options.js
kohler: It means that if you change size indexes you should change getFontMetrics

src/fontMetrics.js
kohler: OK

kohler: Called it sigmasAndXis

kohler:
Hi! Responded to the review; any other comments?

edemaine:
FYI, merging your #745 ended up making a small conflict here.

kohler:
Fixed the conflict (assuming a squash-merge later, didn't bother to rebase).

kevinbarabash:
@kohler looks good.  Thanks cleaning this up.

Issue #646, opened 2017-01-25
vertical alignment for \rule seems off


tscholl2:
When a `\rule` has a height argument, its vertical alignment seems to be off:

![image](https://cloud.githubusercontent.com/assets/8905985/22275099/2ac2d8ce-e25f-11e6-9ac6-8812a7e708b5.png)

You can see the issue without `\frac` or `\color`, I just used them to highlight the issue.

Here is a similar expression in latex:

![image](https://cloud.githubusercontent.com/assets/8905985/22275135/6be33f92-e25f-11e6-92c2-4193bf587e0d.png)

rendered with:
```
\documentclass{article}
\usepackage{color}
\begin{document}
$$
\frac{\textcolor{blue}{\rule{2em}{2em}}}{2}
$$
\end{document}
```

The katex screenshot was taken from the katex homepage today, and I've seen the same thing running version `0.7.1` (I'm not sure if the github pages has the most up to date version).

I'd love to help fix this if possible, but my fear is that it's a very subtle and complicated CSS issue. My guess is to look at the rule for `\rule` in the `buildHTML`? Specifically, I thought this function might be important:

https://github.com/Khan/KaTeX/blob/master/src/buildHTML.js#L1319

I don't know if this is a known issue, I couldn't find it searching, so sorry if this is a duplicate.

arnog:
I think the fix might be to set the `style.marginTop` property to `-(height - shift)` and to not set `style.bottom`.

ronkok:
I believe that this issue was resolved by PR #768.
![rule](https://user-images.githubusercontent.com/16403058/29300790-90c699e6-812c-11e7-99d3-2844c7b03e03.PNG)

PR #744, opened 2017-06-28
Set maxFontSize on rules.

kohler:
This is needed in vlists, as in #646, to make fontsize-ensurer
nodes big enough.


ronkok:
Rules can be set to any arbitrary height. This should be tested with some really tall ones. I haven't got time to  test myself this myself at the moment, but when I was working on PR 670, I ran into trouble when I tried to accommodate tall images by a method similar to this. 

If I recall correctly, `makeVList` applied the maxFontSize not only to the element that needed it, but to all the members of the vlist. I ended up in some cases with unexpectedly high spacing above the vlist.  I.e., the vlist pushed the whole line downward on the page and left a large gap above the math.

ronkok:
For `\fbox`, my solution was to (1) set maxFontSize to 1 em because that did not distort the rest of the vlist, and (2) adjust the position of the `fbox` after running `makeVList`.  The adjustment looks like:
````
if (img.height > vlist.maxFontSize) {
        // Correct for an issue in makeVList. It placed the image top at
        // the top of the line box created by a 1 em maxFontSize.
        vlist.children[1].style.top = -(inner.height + pad - 0.9 / scale)
            + "em";
        // The 0.9 in the previous line is there because the KaTeX fonts
        // have an ascent = 0.9 em. We're setting the top of the image
        // relative to the top of that line box.
    }
````
Just a suggestion.

kohler:
Hi @ronkok, that doesn't work when sizes change, even for `\fbox`. E.g. (on master):

image

Or (again on master) try `\fbox{\rule{0.1em}{2em}}`.

The fontsize-ensurer nodes are weird and suboptimal, but given the current CSS coding they appear necessary.

I think this commit should be merged (I've tested with very large rules—thanks!—and they work, although they do, as you suspected, introduce large gaps) and the vlist/fontsize-ensurer issues pushed to another project.

ronkok:
@kohler, Thank you for pointing out the issue with a size change inside an `fbox`. It looks like I have more work to do. Sigh.

edemaine:
Maybe it would be good to add some more test(s) to illustrate, such as the `\fbox{\huge}` example and large `\rule`s?

kohler:
I'd like this PR to concentrate on bug #646, and there could be another PR or issue on issues surrounding fontsize-ensurer?

kevinbarabash:
@kohler could you post a screenshot of `\frac{\textcolor{blue}{\rule{2em}{2em}}}{2}` from the original issue?

kevinbarabash:
screen shot 2017-06-29 at 7 01 53 pm

Curious got the better of me.  It'd be good to have this as a screenshot test as it's more obvious what the problem is in this example as compared to the existing screenshots that change.

One concern I have about adding more screenshot tests is that the tests and taking longer and longer to run.  I hope that the new headless modes for Chrome and Firefox will make it easier to take screenshots.

kohler:
Screenshotter test added. Re: test speed, PR #752 speeds up screenshotter tests 6x

PR #751, opened 2017-06-30
Correct handling of unbraced kerns followed by spaces.

kohler:
Example: `a\mkern-1mu b` should be parsed like `a\mkern{-1mu}b`. It wasn't, now it is.

Did not realize that `Parser.nextToken.text` can contain spaces (it can). Handle that.

This may allow, for instance, #727 to use the actual TeX definitions, rather than adding spurious braces.

### DIFF
--- a/src/Parser.js
+++ b/src/Parser.js
@@ -789,7 +789,7 @@ Parser.prototype.parseSizeGroup = function(optional) {
     let res;
     if (!optional && this.nextToken.text !== "{") {
         res = this.parseRegexGroup(
-            /^[-+]? *(?:$|\d+|\d+\.\d*|\.\d*) *[a-z]{0,2}$/, "size");
+            /^[-+]? *(?:$|\d+|\d+\.\d*|\.\d*) *[a-z]{0,2} *$/, "size");
     } else {
         res = this.parseStringGroup("size", optional);
     }
--- a/test/katex-spec.js
+++ b/test/katex-spec.js
@@ -1004,6 +1004,7 @@ describe("A kern parser", function() {
     const emKern = "\\kern{1em}";
     const exKern = "\\kern{1ex}";
     const muKern = "\\kern{1mu}";
+    const abKern = "a\\kern{1em}b";
     const badUnitRule = "\\kern{1px}";
     const noNumberRule = "\\kern{em}";
 
@@ -1011,10 +1012,12 @@ describe("A kern parser", function() {
         const emParse = getParsed(emKern)[0];
         const exParse = getParsed(exKern)[0];
         const muParse = getParsed(muKern)[0];
+        const abParse = getParsed(abKern)[1];
 
         expect(emParse.value.dimension.unit).toEqual("em");
         expect(exParse.value.dimension.unit).toEqual("ex");
         expect(muParse.value.dimension.unit).toEqual("mu");
+        expect(abParse.value.dimension.unit).toEqual("em");
     });
 
     it("should not parse invalid units", function() {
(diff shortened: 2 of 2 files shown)


src/Parser.js
```diff
@@ -789,7 +789,7 @@ Parser.prototype.parseSizeGroup = function(optional) {
     let res;
     if (!optional && this.nextToken.text !== "{") {
         res = this.parseRegexGroup(
-            /^[-+]? *(?:$|\d+|\d+\.\d*|\.\d*) *[a-z]{0,2}$/, "size");
+            /^[-+]? *(?:$|\d+|\d+\.\d*|\.\d*) *[a-z]{0,2}\s*$/, "size");
```
kevinbarabash: We use `this.consumeSpaces();` in other places.  It might be good to do the same here for consistency.  It should be noted that `consumeSpaces` does not consume all whitespace, only plain old spaces.  I wonder if it should be upgrade?

test/katex-spec.js
```diff
@@ -1037,17 +1040,42 @@ describe("A non-braced kern parser", function() {
     const emKern = "\\kern1em";
     const exKern = "\\kern 1 ex";
     const muKern = "\\kern 1mu";
+    const abKern1 = "a\\mkern1mub";
+    const abKern2 = "a\\kern-1mub";
+    const abKern3 = "a\\kern-1mu b";
```
kevinbarabash: Thanks for filling out the tests.  Only `abKern3` would fail without the changes to Parser.js.

src/Parser.js
kohler: There's no way to call `consumeSpaces()` here, this is before the tokens are created. It would need to be move into `this.parseRegexGroup`. And that would be a bigger project: ti would change the meaning. Please take commit as is.

edemaine:
I didn't even realize that unbraced size arguments were supported by KaTeX!  (because I always added extra spaces, it seems)  Thanks for fixing this.

src/Parser.js
edemaine: How will this `\s*` interact with newlines?  I suppose KaTeX doesn't support the double newline -> `\par` conversion, but if it did, I suppose this regex might cause trouble...

edemaine: Agreed with @kohler on `consumeSpaces` -- that's only for when spaces have already been parsed by the tokenizer.

kohler: To avoid further discussion on hypothetical future problems or inconsistencies, I will make the regex spaces only. :)

kohler: And spaces-only is certainly more consistent with the rest of the regex! Thanks for pushing further.

edemaine: Actually, I think the behavior you've implemented with `\s*` is more consistent with the rest of KaTeX.  (just did some tests on https://khan.github.io/KaTeX/ )

edemaine: I would actually argue for `\s*` in both places.  The following renders like `x\kern 1em y` in LaTeX.

```latex
x\kern
1em
y
```

kohler: Ah, but whitespace conversion is already handled by KaTeX's lexer, which changes all `[ \r\n\t]+` sequences to a single space character.

New commit has test making this clear.

edemaine:
I approve this PR, but I'll leave it to @kevinbarabash to confirm and push the button.

kevinbarabash:
@kohler thanks for the changes.  @edemaine thanks for looking this over.  Always good to have a second pair of eyes.

PR #752, opened 2017-06-30
Screenshotter `--fast` argument speeds it up 6x.

kohler:
The slowest part of screenshotter tests is the page load, probably
because so many assets must be loaded over the slow docker
connection. On my laptop this takes ~4s per test. The `--fast`
argument avoids this cost by rendering new TeX on the existing
page. For second and subsequent tests, use `executeAsyncScript`
to call KaTeX, rather than performing a full page + asset load.
(If too many errors happen in `--verify` mode, we fall back to
full loads.)

On my laptop, a full verify (chrome + FF) used to take 12m20s+.
With `--fast` it takes 2m23s.


kohler:
This does not yet update travis to supply `--fast`. Most likely `--fast` should be the default but that could come later I dunno.

kohler:
Works on Travis too so I made `--fast` the default (old behavior recoverable with `--slow`).

edemaine:
Tested locally and looks good to me!  (3m8s)  I tried tweaking `ss_data.yaml`, introducing errors, etc. and all seems to behave as before.

@gagern, do you want to take a look at this to confirm?

kevinbarabash:
@kohnler I'm super excited about this change.  A lot of time and effort has gone into our current solution so I'm glad that you were able to iterate on it to speed it up in a reliable way.

gagern:
Had a look at the code, didn't run it. Approach looks sane enough, particularly with the fallback in place. Name of the option could be more descriptive. Perhaps `--reload` instead of `--slow`?

kohler:
Sure, good choice—renamed.

test/screenshotter/test.html
```diff
@@ -32,31 +32,40 @@
...
 
-      katex.render(query["tex"], mathNode, settings);
-      document.getElementById("pre").innerHTML = query["pre"] || "";
-      document.getElementById("post").innerHTML = query["post"] || "";
+        katex.render(query["tex"], mathNode, settings);
+        document.getElementById("pre").innerHTML = query["pre"] || "";
+        document.getElementById("post").innerHTML = query["post"] || "";
+
+        if (callback && document.fonts && document.fonts.ready) {
+          document.fonts.ready.then(callback);
```
kevinbarabash: I didn't realize there was a `document.fonts.ready` was a thing.  I've used webfontloader in the past, but this is better for this use.

dockers/Screenshotter/screenshotter.js
```diff
@@ -372,16 +377,32 @@ function takeScreenshot(key) {
...
     const url = katexURL + "test/screenshotter/test.html?" + itm.query;
-    driver.get(url);
-    if (opts.wait) {
-        browserSideWait(1000 * opts.wait);
+    driver.call(loadMath);
+
+    function loadMath() {
+        if (!opts.reload && driverReady) {
+            driver.executeAsyncScript(
+                    "var callback = arguments[arguments.length - 1]; " +
```
kevinbarabash: I had to look this up in the selenium docs.  It seems like such an awkward API for interacting with the browser.  Thanks for figuring out all of this stuff.

Issue #746, opened 2017-06-29
Enable tall items in makeVList


ronkok:
KaTeX's `makeVList` function has trouble with tall items. Given, say, a tall image, it will locate the top of the image no higher than the top of the text line box created by the `fontSizeEnsurer`. This location will be in conflict with the `height` and `depth` passed to `makeVList` if the tall item is taller than `maxFontSize`.

Work-arounds tried so far include: (1) passing a large value of `maxFontSize` to `makeVList`, as in PR #744, and (2) trying a hand-written adjustment to the tall item's location, as in the `\fbox` part of PR #670.

(1) can push the line downward and create large gaps above the math.

(2) will not work properly if there is a font size change inside the `\fbox`.

A more general solution would be helpful.

ronkok:
Question for @kevinbarabash or @xymostech: I am inclined to write a PR that withdraws `\boxed` until I can write a bug-free version or until somebody can make `makeVList` more robust. `\boxed` is not that urgent and I would hate to see a bug in `\boxed` be something that would delay the next KaTeX release.

What do you think of a temporary removal of `\boxed`?

edemaine:
@ronkok Personally I'm OK with a buggy `\boxed` as long as it works in many useful cases.  Many existing features have bugs in some cases -- that shouldn't stop us from releasing them.

kevinbarabash:
@ronkok I'm okay with releasing `\boxed` in its current state, but if there are known issues with a new feature we should include those issues in the release notes.

xymostech:
Nooooooo. vertical alignment bugs strike again. +1 to what @edemaine said, having a feature with known limitations is probably better than no feature.

kohler:
Fixed in #768

PR #754, opened 2017-06-30
Fix \boxed

ronkok:
A temporary patch for `\boxed`. When the box contains a large font, `\boxed`
now will pass a large value of maxFontSize to makeVList. So the box will
always be located correctly relative to the subject matter. On the other hand, 
the whole line may be pushed downward on the screen.

If the all fonts inside the box are <= 1.0 em, then `\boxed` acts just as
in PR #670, a method which does not displace the line.

To resolve all the problems, we first need to resolve issue #746.

### DIFF
--- a/src/buildHTML.js
+++ b/src/buildHTML.js
@@ -1604,7 +1604,7 @@ groupTypes.enclose = function(group, options) {
         {type: "elem", elem: img, shift: imgShift},
     ], "individualShift", null, options);
 
-    if (img.height > vlist.maxFontSize) {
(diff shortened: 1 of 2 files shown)


ronkok:
Perhaps in lieu of Screenshotter, a screen shot may be sufficient:
boxed

ronkok:
I'm closing this PR in preference for a slightly more general approach that I have planned. Stand by for a day or two.

edemaine:
Cool, looking forward to it!

PR #755, opened 2017-07-01
Correct computation of TeX sizes.

kohler:
* Em and mu sizes are relative to current font's quad metric,
  which might not be "1em" in CSS terms.
* We don't need the `emPerEx` metric, because we want to convert
  TeX `ex` to CSS `em`, not TeX `ex` to TeX `em`. The `xHeight`
  metric does that already: it is relative to font size.
* Mu sizes scale with the style, em and ex do not (empirically
  checked; agrees with rule 2 of TeXbook Appendix G).
  This corrects a bug in b866cd5224b143 (in which em scaled),
  and a bug in previous versions (in which mu did not scale).
* Correct rule sizes. Previous comment---"the sizes of rules are
  absolute"---was misleading. Rule sizes are NOT absolute---in
  \large size, a rule denominated in 'em' is larger. But the 'em'
  unit is not sensitive to styles. So now a 1em rule will be the
  same size in super/subscript, as it should be, but an 18mu rule
  will change size in super/subscript, as it should.

Note a TODO.


kohler:
@edemaine This fixes the problem you reported in #732 and makes measurements more texlike. It corrects some bugs in #719 and some bugs in prior versions. I'll leave some comments on #732.

kohler:
NB the screenshotter test will not run in TeX as is because `mu` is not allowed in rule dimensions…

kevinbarabash:
I'll have a look at this tomorrow.

src/buildHTML.js
```diff
@@ -547,14 +547,40 @@ groupTypes.genfrac = function(group, options) {
...
+        } else {
+            unitOptions = options;
+        }
+        // TODO: In TeX these units are relative to the quad of the current
+        // *text* font, e.g. cmr10. KaTeX instead uses values from the
+        // comparably-sized *Computer Modern symbol* font. At 10pt, these
+        // match. At 7pt and 5pt, they differ: cmr7=1.138894, cmsy7=1.170641;
+        // cmr5=1.361133, cmsy5=1.472241. Consider $\scriptsize a\kern1emb$.
+        // TeX \showlists shows a kern of 1.13889 * fontsize;
+        // KaTeX shows a kern of 1.171 * fontsize.
```
kevinbarabash: Should we always be using the size of the quad defined in cmsy for kern/mkern?

src/buildHTML.js
```diff
@@ -547,14 +547,40 @@ groupTypes.genfrac = function(group, options) {
...
+ * a CSS em value. `options` gives the current options.
+ */
 const calculateSize = function(sizeValue, options) {
-    let x = sizeValue.number;
-    if (sizeValue.unit === "ex") {
-        x *= options.fontMetrics().emPerEx;
-    } else if (sizeValue.unit === "mu") {
-        x /= 18;
+    let scale;
+    // `mu` units scale with scriptstyle/scriptscriptstyle.
```
kevinbarabash: This is b/c the value for quad for textstyle is 1.0, right?

src/buildHTML.js
```diff
@@ -547,14 +547,40 @@ groupTypes.genfrac = function(group, options) {
...
+    // Other units always refer to the *textstyle* font in the current size.
```
kevinbarabash: Below it looks like we're only using textstyle if the current style `isTight`.  If that's the correct behavior could you update this comment to reflect that?

src/buildHTML.js
```diff
@@ -1338,15 +1358,11 @@ groupTypes.kern = function(group, options) {
...
-    let dimension = 0;
     if (group.value.dimension) {
-        dimension = calculateSize(group.value.dimension, options);
+        const dimension = calculateSize(group.value.dimension, options);
+        rule.style.marginLeft = dimension + "em";
     }
 
-    dimension /= options.sizeMultiplier;
-
-    rule.style.marginLeft = dimension + "em";
```
kevinbarabash: Not sure why were doing this out side of the `if` statement.  Thanks for cleaning this up.

src/buildHTML.js
kohler: We should be using the size of the quad defined in *cmr*.

src/buildHTML.js
kohler: Hm—`isTight()` is equivalent to `scriptstyle/scriptscriptstyle`?

src/buildHTML.js
kohler: Well, not for all fonts—textstyle in some fonts has a different quad.

For example, under `\tiny`, textstyle is `cmr5`, which has quad:

```
   (QUAD R 1.361133)
```

If the question is referring to the `mu units scale` comment, that comment is trying to explain why the "mu" branch has no `unitOptions`.

src/buildHTML.js
kohler: (But this commit doesn't change that.)

src/buildHTML.js
kohler: I added a comment to that effect.

kohler:
Hey @kevinbarabash, if you’d like to run the tests on texcmp, I could get rid of the 18mu rules. (`\rule` cannot take `mu` units in TeX.)

Here’s what the screenshotter test should show. The top row uses `em` units. The gaps should be (and are) the same in all styles. The middle row uses `mu` units. The gaps should scale with style (i.e., superscript smaller). The left column uses text size, the right column uses `\footnotesize`, so the gaps should be smaller on the right than the left. The bottom row has rules: 1em^1em, then 18mu^18mu. Again, the `em` rules should be the same size in superscript, the `mu` rules should scale.

kohler:
Travis is confused again: this PR builds fine.

https://travis-ci.org/kohler/KaTeX/builds/251605937

src/buildHTML.js
kevinbarabash: I read that backwards.  Thanks.

kevinbarabash:
I ran `dockers/Screenshotter/screenshotter.sh --verify --diff --include=RelativeUnits` locally and ended up with the following images.

Chrome:
relativeunits-chrome-diff

Firefox:
relativeunits-firefox-diff

Not sure why the test run you linked passed and these failed, but it seems to be failing consistently.  I'll send you a PR with updated screenshots.

kevinbarabash:
Remove the `\rule`s containing `mu` units, texcmp renders the following:
relativeunits

The relative distances between `a`s and `b`s appears to be the same.  LGTM.

PR #679, opened 2017-04-09
Upgrade the source to use  ES6 syntax including classes, import and static properties

HosseinAgha:
Discussed at #654.
I also think this fixes the second issue in the project's backlog [here](https://github.com/Khan/KaTeX/projects/5).

PR #756, opened 2017-07-02
Remove endpoint with not transpiled js in dev server.js

HosseinAgha:
In order to merge #679 we cannot continue with not transpiled endpoint in server.js development environment because even most up to date browser still don't support many ES6 features etc yet.   
I did not remove the old code so we can role back when we have a proper browser support.

### DIFF
--- a/server.js
+++ b/server.js
@@ -47,8 +47,9 @@ function serveBrowserified(file, standaloneName, doBabelify) {
 }
 
 function twoBrowserified(url, file, standaloneName) {
-    app.get(url, serveBrowserified(file, standaloneName, false));
-    app.get("/babel" + url, serveBrowserified(file, standaloneName, true));
+    // we should always transpile all js files
+    // in order to support all ES2015 features
+    app.get(url, serveBrowserified(file, standaloneName, true));
 }
 
 function twoUse(url, handler) {


server.js
```diff
@@ -47,8 +47,9 @@ function serveBrowserified(file, standaloneName, doBabelify) {
 }
 
 function twoBrowserified(url, file, standaloneName) {
```
kevinbarabash: Can you change this name to `browserified`?

server.js
```diff
@@ -47,8 +47,9 @@ function serveBrowserified(file, standaloneName, doBabelify) {
...
-    app.get(url, serveBrowserified(file, standaloneName, false));
-    app.get("/babel" + url, serveBrowserified(file, standaloneName, true));
+    // we should always transpile all js files
+    // in order to support all ES2015 features
+    app.get(url, serveBrowserified(file, standaloneName, true));
 }
 
 function twoUse(url, handler) {
```
kevinbarabash: We probably don't need this function anymore since we don't have `/babel` urls for `katex.js`.  Can you replace all uses of this function with `app.use`?

kevinbarabash:
I see you also made this change in https://github.com/Khan/KaTeX/pull/679.  I thought it was going to be more involved, but it's not so let's fix it there.

HosseinAgha:
@kevinbarabash sorry that was an unwanted merge on my side. I made all updated on this PR.

server.js
HosseinAgha: Done

Issue #60, opened 2014-09-16
Square root line position is inconsistent (depending on its contents)


phoenixeliot:
This is on a Retina MBP.

The line above the contents of a square root is inconsistently placed. Interestingly, I just tested this on writelatex.com and they have a similar, but not identical, issue.

Some symbols (seemingly dependent on their height and categorization) cause this issue while others do not. All numerical digits cause this issue, along with parentheses. Short letters (like 'a') result in the whole square root dropping, but I think that is expected. Letters like "h" and "l" cause the rendering to be correct, unless a parenthesis is there. It seems like the tallest character determines the behavior.

I'm not sure if the top of the sqrt symbol is supposed to line up perfectly with the horizontal bar, or if there is supposed to be a little tail sticking out there. writelatex.com also has a slight tail sometimes. I don't have a native LaTeX renderer installed, so I can't test that.

KaTeX: 
![screen shot 2014-09-16 at 6 34 53 pm](https://cloud.githubusercontent.com/assets/491376/4296095/bc617598-3df1-11e4-88c7-6beedd30d4b5.png)

(blown up:)
![screen shot 2014-09-16 at 6 36 01 pm](https://cloud.githubusercontent.com/assets/491376/4296104/dc82591e-3df1-11e4-8604-63be7aed23cb.png)

writeLaTeX.com: (PDF output)
![screen shot 2014-09-16 at 6 33 25 pm](https://cloud.githubusercontent.com/assets/491376/4296098/c2e49d00-3df1-11e4-9fa6-3b38899af5eb.png)

qbolec:
Is there any workaround for Windows?
I don't think I do anything fancy - styles on my site are based on Bootstrap, yet every sqrt sign has the line in a bad place.
I am open to solutions like fixing some particular font/size/whatever combination if that would help.
I can add some custom CSS styles if that's needed.
Is there any temporary solution to this problem, which could let me transition production from MathJax to KaTeX ? :)

qbolec:
It also seems that only particular font-sizes look bad: 
14px - bad
15px - ok
20px - bad
21px - ok
Too bad that Bootstrap uses 14px as the default :/

jannschu:
Looked a bit into this. One problem is, that the line is implemented using a CSS border property, which browsers render differently than fonts.

See this [experiment](https://jsfiddle.net/Lup5o8om/): On my machine I tried to perfectly align the underscore (representing the square root character) and a `<span>` with some border. I can observe that,

[...]

PR #757, opened 2017-07-05
Fix \boxed, add \overbracket

ronkok:
This PR introduces `buildCommon.makeImageVList`, an alternate to `makeVList`,
for use when one element is on the baseline and the other element is an
image. Unlike `makeVList`, the new function aligns the image correctly even if 
image height > 1 em.

The PR then uses `makeImageList` to fix `\boxed` and adds two new
functions: `\overbracket` and `\underbracket`.

This PR is also a stepping stone to work that could resolve issue #60.


ronkok:
Since I have never mastered Screenshotter, I would appreciate any help. It all looks pretty good on my [test page](http://copyasmarkup.com/katex/wideelements.html).

kevinbarabash:
> Since I have never mastered Screenshotter, I would appreciate any help. It all looks pretty good on my test page.

I can help out, but it would be good to diagnose the screenshotter issue as well as other people may run into the same issue.

ronkok:
Oh, my issues run deeper than Screenshotter. I work in Windows. Now, I know that `make` files, `bash` scripts, and Docker are all possible within Windows. They take more effort to learn and the documentation is scattered, incomplete, and often outdated or misleading. But doable, given effort. Which I have not exerted.

Which means that my root problem is sloth.

kevinbarabash:
I thought closing/reopening might fix the travis-ci issue.  I've sent github support an email about the issue.

kevinbarabash:
There we go.

kevinbarabash:
Hmm.. not sure why the `Boxed` screenshots are failing.  I'll look at it this evening.

kevinbarabash:
I see what the issue is.  I didn't commit the changes to ss_data.yaml.  I can fix that.

kevinbarabash:
When comparing the updated images for HorizontalBraces using the swipe feature I noticed that the underbrace is no longer centered when it should be.

horizontalbraces

Another difference from LaTeX that this test points out is that the effect of `\displaystyle` should continue past the group that immediately follows it.  Notice how large the fraction is.  I'll open a separate issue for this.

ronkok:
Re: Centering. I must have introduced a CSS bug with this PR. I'll fix that, but it may be a few days before I get to it. Things are a little hectic right now.

ronkok:
@kevinbarabash, I've reworked `makeImageVList`. It is now possible to center the text argument on a longer image argument. Naturally, the screenshotter images have issues.

I've also made a couple of edits in `buildHML.js` that were necessary to keep up with the rapid changes to other parts of KaTeX.  GitHub is declaring those to be conflicts, and I'm really not sure why.

kevinbarabash:
@ronkok it's possible that github is treating https://github.com/Khan/KaTeX/pull/765, which was merged recently, as a conflict.

ronkok:
Screenshotter error again.

kevinbarabash:
I'm confused as to why the updated files aren't showing up in this PR. :|

ronkok:
Well it you are confused, then I am bewildered. I would be willing to try again, if you want to try another PR.

kevinbarabash:
Looks like it's `StrikeThrough` that I should've regenerate screenshots for.

ronkok:
@kevinbarabash, It's probably not worth spending any more time on this PR. @kohler's PR#768 solves the same problem as `makeImageVList` and #768 does it in a more general way.

I'll leave this PR open for a bit longer, but I expect to close it in preference for #768.

ronkok:
Closing this PR in favor of PR #768.

Issue #760, opened 2017-07-08
Soften the edge of the diagonal lines in cancel


kevinbarabash:
http://jsfiddle.net/nuxcbqqq/26/

with antialiasing
<img width="401" alt="screen shot 2017-07-08 at 6 54 58 pm" src="https://user-images.githubusercontent.com/1044413/27989788-237b021c-640f-11e7-8791-634be4a29bfa.png">

without antialiasing
<img width="401" alt="screen shot 2017-07-08 at 6 55 55 pm" src="https://user-images.githubusercontent.com/1044413/27989791-37d2b232-640f-11e7-99ed-95260e413c67.png">

kevinbarabash:
This is what's already being done.  I guess the difference is in the current code we have a hard edge vs. a soft one.

kevinbarabash:
I've modified the `.cancel` styling in katex.less by subtracting/adding 0.5px to the leading and trailing edges of the cancel line.
```
.cancel {
    background:
        linear-gradient(to top left,
            rgba(0,0,0,0) 0%,
            rgba(0,0,0,0) ~"calc(50% - 0.04em - 0.5px)",
            rgba(0,0,0,1) ~"calc(50% - 0.04em)",
            rgba(0,0,0,1) ~"calc(50% + 0.04em)",
            rgba(0,0,0,0) ~"calc(50% + 0.04em + 0.5px)",
            rgba(0,0,0,0) 100%);
}
```
It seems to look good at a variety of font sizes (12px, 20px, 30px, 72px):
<img width="29" alt="screen shot 2017-07-08 at 7 15 04 pm" src="https://user-images.githubusercontent.com/1044413/27989866-ecbed6ba-6411-11e7-8048-5b43f17efc6d.png">
<img width="52" alt="screen shot 2017-07-08 at 7 15 12 pm" src="https://user-images.githubusercontent.com/1044413/27989868-f1c1a7be-6411-11e7-969f-20375cf3f020.png">
<img width="76" alt="screen shot 2017-07-08 at 7 15 18 pm" src="https://user-images.githubusercontent.com/1044413/27989869-f3b70a28-6411-11e7-8e32-955ccc898d4e.png">
<img width="162" alt="screen shot 2017-07-08 at 7 15 27 pm" src="https://user-images.githubusercontent.com/1044413/27989871-f673a97e-6411-11e7-9116-661a7dc9c119.png">

ronkok:
[...]

PR #761, opened 2017-07-08
Update {b,x}cancel to have a 0.5px transition in gradient so that the…

kevinbarabash:
Addresses https://github.com/Khan/KaTeX/issues/760.  I'm not sure why the lines don't look antialiased in the screenshots.  I tested in Chrome on Mac OS (retina display) and Chrome on Windows.

Chrome on Mac OS
screen shot 2017-07-08 at 7 27 48 pm

Chrome on Windows
screen shot 2017-07-08 at 7 27 35 pm


static/katex.less
```diff
@@ -794,41 +794,41 @@
...
     .cancel {
         background:
-           linear-gradient(to top left,
-               rgba(0,0,0,0) 0%,
-               rgba(0,0,0,0) ~"calc(50% - 0.04em)",
-               rgba(0,0,0,1) ~"calc(50% - 0.04em)",
-               rgba(0,0,0,1) ~"calc(50% + 0.04em)",
-               rgba(0,0,0,0) ~"calc(50% + 0.04em)",
-               rgba(0,0,0,0) 100%);
+            linear-gradient(to top left,
```
kohler: the extra spaces in this diff are slightly annoying but also don't matter. 🤷‍♂️

kevinbarabash: I can fix that.

sthoch:
As I mentioned on the issue (#760), I advise testing on other browsers. I had some problems with gradients involving halfpixels some time ago - could be fixed now, could be not. IIrc, Safari (maybe on iOS) made some troubles.

ronkok:
The screenshots below are, clockwise from top-left, Chrome, Firefox, Opera, and Safari. All are from Windows, on a Microsoft Surface, except the Safari, which is from BrowserStack.
cancel

I don't trust the pixelation shown for Safari, because BrowserStack shows the same pixelation for Opera, and as you can see, that pixelation disappears when taken from my own screen.  Someone with native equipment should re-test Safari.

I didn't test IE or Edge because this CSS doesn't run on IE or Edge. The CSS contains feature-detection that directs IE or Edge to render an SVG instead. Earlier testing showed that those two browsers go berserk when asked to render this linear-gradient combined with a `calc()`.

ronkok:
I just looked at this on iOS Safari on my iPod Touch. All the lines look fine to me.

kevinbarabash:
@ronkok thanks for testing across all of those different browsers.

Issue #762, opened 2017-07-09
define requirements for a plugin architecture


kevinbarabash:
There are a number of issues and PRs that would benefit from having a plugin architecture:
- Add support for Persian & Arabic alphabet in text and eastern-arabic numerals + Proposal Commit #729
- single or double dollar sign as formula delimiter #712
- mhchem extension support #50 
- Cursor #506 
- Class #507 

There are certain functions we could expose:
- `defineSymbol`
- `defineFunction`
- `defineEnvironment`

`defineSymbol` is used for adding support for new symbols or new languages, but if the new glyphs require new fonts currently isn't a way to specify metrics for the new glyphs.  We'd need to provide a way to extend the data stored in fontMetricsData.js.

`defineFunction` is used for adding new commands.  Commands that define implicit groups will be hard to deal and since there aren't that many of them we should probably ignore them for now.  A bigger problem is that any new function will also require code in buildHTML.js and buildMathML.js to generate the correct markup.

Currently, buildHTML.js and buildMathML.js are loose conglomeration of functions.  The only functions that should need to be shared are `buildGroup` and `buildExpression`.  The rest could exist in their own separate files.

I would like to propose grouping the parse node generation functions from `function.js` with functions from `buildHTML.js` and `buildMathML.js` into separate files based on group type.  As part of this, we'd expand the responsibilities of `defineFunction` to add entries to the `groupTypes` objects used by `buildHTML` and `buildMathML`.

We'll want to minimize changes to the parse tree structure once we publish a public plugin API so we should make changes before doing so.

mbourne:
@kevinbarabash Would it also be appropriate to add support for ASCIIMath function in such a plugin architecture?

kevinbarabash:
@mbourne is the idea to convert ASCIIMath syntax to LaTeX syntax and then have KaTeX render the result?  If you're calling KaTeX's render method manually then a plugin probably isn't necessary b/c you can pipe the output of something like https://github.com/tylerlong/asciimath-to-latex to KaTeX's render method.  If you you're looking to tie into the auto render extension then that will have to be updated to include an optional callback that can be run on each string to preprocess it before sending it to KaTeX.

mbourne:
@kevinbarabash Yes, that is the idea. I'm currently piping output from an ASCII to LaTeX converter (similar to tylerlong's), which is fine for me. But I thought I would raise it as a plugin possibility to make it easier for others. MathJax did this after a few years of its existence.

[...]

PR #766, opened 2017-07-18
group all \phantom code, group all delimsizing code, and add introduce flow type checking

kevinbarabash:
This is a first step towards a plug-in system (#762) to allow users to define their own functions.  It augments `defineFunction` to take additional (at the moment optional) args that provide logic to build HTML and MathML expressions necessary to render the `\phantom` command as HTML and MathML respectively.


src/functions/phantom.js
```diff
@@ -0,0 +1,36 @@
...
+        numArgs: 1,
+    },
+    (context, args) => {
+        const body = args[0];
+        return {
+            type: "phantom",
+            value: ordargument(body),
+        };
+    },
+    "phantom",
```
kevinbarabash: It's coincidental that the name of the function and the type matches, most of the time this won't be the case.

kevinbarabash:
I wanted to make sure things worked with defining multiple functions at once so I extracted delimiter sizing functions into their own file.

kevinbarabash:
I'm waiting for #781 to be merged before proceeding with this.  The hope is to convert all over our functions over this kind of a format.  The benefit I see is that `defineFunction` can be pass everything it needs to define a function including the HTML and MathML builders it makes it easier for us to eventually export `defineFunction` so that other developers can define functions at runtime.  In essence this would be a very simple plugin architecture.  Before actually actually exporting `defineFunction` though, I'd like to also clean up our parse tree so it's more consistent and not unnecessarily deep.

src/defineFunction.js
```diff
@@ -0,0 +1,122 @@
+import functions from "./functions";
+import {groupTypes as htmlGroupTypes} from "./buildHTML";
+import {groupTypes as mathmlGroupTypes} from "./buildMathML";
+
+/* This file contains a list of functions that we parse, identified by
```
edemaine: "This file" is no longer correct.  It's now `functions.js`.

Incidentally, is there a particular reason to split `functions.js` into `functions.js` and `defineFunction.js`?  `import` should solve the awkward ordering issue automatically.  But maybe you prefer this conceptually.

src/defineFunction.js
```diff
@@ -0,0 +1,122 @@
...
+ * The function should return an object with the following keys:
+ *  - type: The type of element that this is. This is then used in
+ *          buildHTML/buildMathML to determine which function
+ *          should be called to build this node into a DOM node
+ * Any other data can be added to the object, which will be passed
+ * in to the function in buildHTML/buildMathML as `group.value`.
+ */
+
+export default function defineFunction(
+    names, props, handler, type, htmlBuilder, mathmlBuilder
```
edemaine: I feel like this argument list is starting to get unwieldly.  It's particularly hard to read when looking at the `defineFunction` instances below (which argument am I on?).  I also fear that this precludes other types of building, like SVG or Canvas.  Maybe it would be better to have a single object argument with all the keys currently in `props`, plus `names`, `handler`, `type`, `htmlBuilder`, `mathmlBuilder`?  Then we could easily add other builders too, but not necessarily all at once.

An alternative would be to use `class`es somehow, but I'm still not super familiar with ECMA classes.

src/defineFunction.js
```diff
@@ -0,0 +1,122 @@
...
+    // Set default values of functions
+    const data = {
+        numArgs: props.numArgs,
+        argTypes: props.argTypes,
+        greediness: (props.greediness === undefined) ? 1 : props.greediness,
+        allowedInText: !!props.allowedInText,
+        numOptionalArgs: props.numOptionalArgs || 0,
+        infix: !!props.infix,
+        handler: handler,
+    };
```
edemaine: If we change to a single object argument, perhaps this rewriting of data could be removed?  Or moved to where we manipulate `functions`, if we need to do the default setting.  It feels a bit wasteful... though not a big deal.

src/defineFunction.js
```diff
@@ -0,0 +1,122 @@
...
+    for (let i = 0; i < names.length; ++i) {
+        functions[names[i]] = data;
+    }
+    if (type) {
+        if (htmlBuilder) {
+            htmlGroupTypes[type] = htmlBuilder;
+        }
+        if (mathmlBuilder) {
+            mathmlGroupTypes[type] = mathmlBuilder;
```
edemaine: With the single-object view, these keys could eventually just be checked in `buildHTML`/`buildMathML`.  But maybe just do this for now...

src/defineFunction.js
kevinbarabash: I was thinking that `defineFunction` would be the only shared function, but I see that we also have `ordargument` and we may have other helpers in the future so keeping these functions in `function.js` and `export`ing each of them makes sense.

src/defineFunction.js
kevinbarabash: I agree.  I'll switch to using an object as an arg.  Using `flow` for type checking later will be a low cost way to ensure that we're getting the args we need.

src/defineFunction.js
kevinbarabash: I'm hoping we can do these checks at compile time using `flow`.

kevinbarabash:
> I have one main comment, which would involve changing all defineFunction calls, so is maybe a little drastic... Let me know what you think!

I've side-stepped updating all existing `defineFunction` calls by adding a wrapper around the new one with the change you suggested.

src/defineFunction.js
edemaine: Unpacking the argument to then repack it still feels a little wasteful/verbose here.  Maybe the point is to clone and then set some defaults?  (And not totally sure the cloning is necessary...)  Alternatively, I'd be tempted to add the behavior like `names = [names]` to the old-style `defineFunction` helper, and define the type to require `names: array<string>`.  (Also happy to leave this to a separate PR if you'd prefer.)

src/functions.js
```diff
@@ -1,114 +1,12 @@
...
-const ordargument = function(arg) {
-    if (arg.type === "ordgroup") {
-        return arg.value;
-    } else {
-        return [arg];
-    }
+// Define a convenience function that mimcs the old semantics of defineFunction
+// to support existing code so that we can migrate it a little bit at a time.
+const defineFunction = function(names, props, handler) {
+    _defineFunction({names, props, handler});
```
edemaine: I'm not a huge fan of naming here: if we want to switch to the new style, we need to call `_defineFunction`?  Or is the point that any new stuff should go into its own `.js` file, not this one, and directly import `defineFunction.js`?  Ah yes, I see that below.  Perhaps we should add a comment to the effect of "Don't add new functions here -- follow the style in the `functions` directory".

src/defineFunction.js
kevinbarabash: The cloning isn't necessary, but it can avoid hard to track down issues if they arise.  I think `names: array<string>` would be good.  Good idea moving `names = [names]` to the wrapper.

src/functions.js
kevinbarabash: The main reason for having `defineFunction` in its own file is that nothing in this uses it directly, I didn't want to go changing the calls to `defineFunction` in this file.  Definitely a little confusing.  I'll add some comments including guidance for where to put new functions.

kevinbarabash: > Or is the point that any new stuff should go into its own .js file, not this one, and directly import defineFunction.js?

That's the goal.

kevinbarabash:
@edemaine good call on updating `CONTRIBUTING.md`.  I'll make changes to it this evening.

kevinbarabash:
I'm going to merge this after #833 so that I can put all phantom functions in the same file.

edemaine:
Sounds good.

PR #767, opened 2017-07-18
Supply marginLeft/marginRight as inputs to makeVList.

kohler:
Current code, which modifies the returned elements from makeVList,
is fragile: if makeVList internals changed, the callers would have
to change too.

One place remains where the return value from makeVList is modified, in `groupTypes.enclose`. That will need to be removed in a future commit, because it's not just changing left/right margins.

### DIFF
--- a/src/buildCommon.js
+++ b/src/buildCommon.js
@@ -359,6 +359,12 @@ const makeVList = function(children, positionType, positionData, options) {
             childWrap.height -= shift;
             childWrap.depth += shift;
             childWrap.style.top = shift + "em";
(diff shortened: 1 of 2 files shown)


ronkok:
@kohler, PR #757 already modifies `groupTypes.enclose`. It removes the call to `makeVList` and substitutes a call to new function `makeImageVList`.

kohler:
@ronkok, Thanks for letting me know. I am working on an alternative to `makeImageVList`.

kohler:
Ping on this e.g. @kevinbarabash? Safe; blocking #768.

PR #770, opened 2017-07-21
Test site supports `before=` and `after=` search parameters.

kohler:
They add text before/after the math. Example use: testing baseline alignment.

Example usage:
screen shot 2017-07-21 at 8 41 43 am

### DIFF
--- a/static/main.js
+++ b/static/main.js
@@ -30,6 +30,30 @@ function init() {
         macroString += "&" + match[0].substr(1);
     }
 
+    // The `before` search parameter puts normal text before the math.
+    // The `after` search parameter puts normal text after the math.
+    // Example use: testing baseline alignment.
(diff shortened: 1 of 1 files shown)


static/main.js
```diff
@@ -30,6 +30,30 @@ function init() {
...
+        var mathContainer = math;
+        mathContainer.id = "math-container";
+
+        if ((match = /(?:^\?|&)before=([^&]*)/.exec(window.location.search))) {
+            var child = document.createTextNode(decodeURIComponent(match[1]));
+            mathContainer.appendChild(child);
+            macroString += "&" + match[0].substr(1);
+        }
+
+        math = document.createElement("SPAN");
```
edemaine: As much as I like capital tags personally, I think `"span"` would be more consistent with other KaTeX code (e.g. `src/domTree.js`).

PR #772, opened 2017-07-22
Detect attachEvent() support correctly, fixes #771

### LINKED ISSUE #771: static/main.js assumes attachEvent is supported if input events are not
Hi, just a small issue that came up when I was testing using KaTeX with the PDF generator Prince XML. Here:
https://github.com/Khan/KaTeX/blob/master/static/main.js#L6
the code detects *input event* support using the reasonable shortcut of looking for an `oninput` property. However, it then proceeds to assume that `attachEvent()` is supported if oninput is not defined. The `attachEvent` method should be detected separately.

Thanks for a great script :)

hallvors:


kohler:
Since `main.js` is only used in testing, I see no problem with this. It would be more self-consistent with the code above to write `"attachEvent" in input`, but it's fine either way.

kohler:
I'll wait a bit to merge just in case `"attachEvent" in input` arrives as a PR revision.

edemaine:
As discussed in #711, I'm not convinced this will be useful in isolation -- there's a call to `attachEvent` a few lines below that should be similarly protected, if this one is.  But with a revision (and ideally testing to confirm this works where desired) it seems like a good idea.

kohler:
@edemaine This is a change to a file that is only used in testing KaTeX. I am happy to trust the submitter that the change was useful in his environment.

PR #775, opened 2017-07-30
Improve \sqrt

ronkok:
Use background-image SVGs for square roots. This will ensure a match where the radical meets the viniculum.

Resolves issue #60.


ronkok:
I'll place some screenshots here to illustrate this PR. The first shot was taken this morning before this PR.
sqrt3

Next, we see the same HTML, but the screenshot is taken after this PR.
sqrt4

The new shot has no more mis-alignment between radical and viniculum.

And just to show `\sqrt` in a variety of sizes and styles, here's another screenshot.
sqrt-test

ronkok:
The Screenshotter tests have failed, as expected. @kevinbarabash, your assistance would be appreciated. I'm also curious to see your review comments. At one point, I did something a little sketchy. Perhaps you can think of a better way to pass surd width information.

src/buildHTML.js
```diff
@@ -1122,30 +1120,26 @@ groupTypes.sqrt = function(group, options) {
...
         const rootm = buildGroup(group.value.index, newOptions, options);
 
-        // Figure out the height and depth of the inner part
-        const innerRootHeight = Math.max(delim.height, body.height);
-        const innerRootDepth = Math.max(delim.depth, body.depth);
-
         // The amount the index is shifted by. This is taken from the TeX
         // source, in the definition of `\r@@t`.
-        const toShift = 0.6 * (innerRootHeight - innerRootDepth);
+        const toShift = 0.6 * (body.height - inner.depth);
```
kevinbarabash: Looking at the definition of `innerRootDepth`, shouldn't this be `0.6 * (body.height - body.depth)`?

static/images/Image-Licensing-and-Technical-Notes.txt
```diff
@@ -19,11 +19,16 @@ The license terms above apply to the following images:
     leftrightarrow.svg        U+2194             KaTeX Main
     leftrightharpoons.svg     U+21BC/21B1        KaTeX Main
     mapsto.svg                U+21A6             KaTeX Main
-    overbrace.svg             U+23A9/23A8/23A7   KaTeX Size 4 Regular
+    overbrace.svg             U+23A9/23A8/23A7   KaTeX_Size4-Regular
```
kevinbarabash: bug fix?

src/delimiter.js
```diff
@@ -317,6 +312,71 @@ const makeStackedDelim = function(delim, heightTotal, center, options, mode,
...
+        span = buildCommon.makeSpan(["stretchy"], [], options);
+        const svgHeight = Math.floor(height * 1000);   // scale = 1:1000
+
+        let color = "black";
+        if (options.color) {
+            color = options.color.replace(/#/, "%23");  // url escape
+        }
+
+        // To make a \sqrt that is customized in both height and width, we
+        // have to use an inline SVG. We set the height right now. Then
```
kevinbarabash: Cool

kevinbarabash:
So I ran the screenshotter locally and I'm getting the same error that we're seeing on the travis with `inner.style` being undefined.  I was able to reproduce the error by loading:
```
http://0.0.0.0:7936/test/screenshotter/test.html?tex=%5Cerr%5C%2C%5Cfrac%5Cfracerr3%5C%2C2%5E%5Csuperr_%5Csuberr%5C%2C%5Csqrt%5Csqrterr&noThrow=1
```
For some reason we aren't catching the exception in this situation.

src/delimiter.js
```diff
@@ -317,6 +312,71 @@ const makeStackedDelim = function(delim, heightTotal, center, options, mode,
...
+    let span;
+
+    if (delimData.type === "small") {
+        // Get an SVG that is derived from glyph U+221A in font KaTeX-Main.
+        span = buildCommon.makeSpan(["stretchy", "sqrt-main"], [], options);
+        span.height = 1;  // Inside the font, the glyph is 1000 units tall.
+
+        // span.depth is really zero, but I'll mis-use the .depth attribute
+        // to carry info about surd width back to groupTypes.sqrt
+        span.depth = 0.781;   // surd width, from the font.
```
ronkok: Good idea, to use `span.width`. I'll do that.

Re: multiplier. (1) Because it works empirically.  (2) The function `traverseSequence` has already taken care of that for small delimiters. The relevant code being:
````
        // Small delimiters are scaled down versions of the same font, so we
        // account for the style change size.

        if (sequence[i].type === "small") {
            const newOptions = options.havingBaseStyle(sequence[i].style);
            heightDepth *= newOptions.sizeMultiplier;
        }
````

File: src/delimiter.js [hunk 5]
```diff
@@ -317,6 +312,71 @@ const makeStackedDelim = function(delim, heightTotal, center, options, mode,
...
+        // The code has to be URL-escaped, so it isn't pretty.
+        span.style.backgroundImage = "url('data:image/svg+xml,%3Csvg xmlns=" +
+            "%27http://www.w3.org/2000/svg%27%3E%3Csvg viewBox=%270 0 400000 "
+            + svgHeight + "%27 preserveAspectRatio=%27xMinYMax slice%27%3E"
+            + "%3Cpath fill=%27" + color + "%27 d=%27M591 0h399409v40H631v"
+            + (svgHeight - 54) + "l-4 4-4 4c-.667.667-2 1.5-4 2.5s-4.167 "
+            + "1.833-6.5 2.5-5.5 1-9.5 1h-12l-28-84c-16.667-52-96.667"
+            + "-294.333-240-727l-212 -643 -85 170c-4-3.333-8.333-7.667-13"
+            + "-13l-13-13l77-155 77-156c66 199.333 139 419.667 219 661"
+            + "l218 661zM591 0H400000v40H631z%27/%3E%3C/svg%3E%3C/svg%3E')";
```
ronkok: Done.

src/buildHTML.js
ronkok: I'll change it to `body.depth`.

static/images/Image-Licensing-and-Technical-Notes.txt
ronkok: Yeah, I was writing font names for the surds and I changed this to be consistent.

src/buildHTML.js
```diff
@@ -1122,30 +1120,26 @@ groupTypes.sqrt = function(group, options) {
     if (inner.height === 0 && inner.depth === 0) {
         body = makeSpan();
     } else {
+        inner.style.paddingLeft = surdWidth;
+        // Overlay the image and the argument.
         body = buildCommon.makeVList([
-            {type: "elem", elem: inner},
-            {type: "kern", size: lineClearance},
-            {type: "elem", elem: line},
-            {type: "kern", size: ruleWidth},
```
kohler: It's actually important to preserve this outer kern so that the whole span has a bit extra space on the top.

kohler:
So this is really cool, and a commit like may be the only way to absolutely always match the `\surd` and the rule. However, I still prefer font images (which are already in cache) to remote-loaded SVGs. We can make incremental progress by calculating rule thicknesses more carefully and improving CSS coding somewhat. #776 does that; it's useful whether or not this is merged.

ronkok:
@kohler, I'm ambivalent myself about whether this PR is a good idea. If bar placement can be improved sufficiently, then this PR would be unnecessary. 

I'll clean up the rest of my work tomorrow so a good comparison can be made.

kohler:
@ronkok, not sure. The existing CSS coding requires precise alignment between a font and a CSS border. Seems likely those will be treated differently by browser subpixel rounding algorithms. In #776 I experimented with subpixel offsets; these worked well at larger sizes but poorly at smaller sizes. box-sizing: border-box made no difference. There are, though, perhaps other things to try.

kohler:
A problem with `background-image` is that it often doesn't print.

kohler:
Incidentally, here is the sequence of `\sqrt{a}/\sqrt{1}` at sizes 12,13,14,15,19,20,21pt given #776. It is better than the current state (see #60 for the current state) but not perfect. First at 100%, then 400%.

screen shot 2017-07-31 at 11 51 28 am

screen shot 2017-07-31 at 11 51 49 am

kevinbarabash:
@kohler those screenshots look really good.

ronkok:
@kevinbarabash, I've got a fix for the sizing problem that you found yesterday. (Thanks!) I'll push it later today when I have WiFi access. Then you can choose whether to accept this PR or to accept only PR #776.

If it were me, I'd accept PR #776 only. I don't see that this PR gets you much improvement.  It will remain possible to change to an SVG method later.

ronkok:
@kevinbarabash, This commit selects the correct image size, as shown below. Thank you for pointing out the issue.
sqrtsizing

kohler:
Hi @ronkok, just FWIW. It looks to me like you are scaling the `\sqrt` images by the font multiplier. TeX instead works through a single set of `\sqrt` images, not always scaled by the font multiplier. Honestly I like your look better :) but here's a pdftex image.

screen shot 2017-07-31 at 9 19 47 pm

What I see here is:
* In your image, the superscript versions look to me like scaled-down versions of the main version.
* In pdftex, the superscript images differ. The outermost `\sqrt` in the main version is different from the outermost `\sqrt` in the superscripts. The innermost `\sqrt` versions also differ. I think a smaller `\sqrt` is made available in the superscripts.

ronkok:
In my images, the superscripts are nested three deep. In each group of three, the outermost surd is steeper than the inner surds. That tells you that the outermost surd has been derived from the glyph in font KaTeX_Size2-Regular, not KaTeX Main

So no, these are not all scaled versions of the same glyph. 

That's not to say that my code is perfect. In my images, I get a more narrow line thickness in scriptscript than in textsize. And LaTeX has the same line thickness. So I do have a scaling problem that needs to be fixed.  If @kevinbarabash, decides that the `\sqrt` in PR #776 is not good enough, then I'll dive back in and work it out.

kohler:
Hey @ronkok, not sure I was clear. I can see the differences *within* each group of three, but in pdftex/current katex, there are also differences *across* the groups of three.

In pdftex, the groups of three have the following surd characters, in order outer->inner:
* Largest: cmex10 “r”, cmex10 “q”, cmex10 “p”
* Superscript: cmex7 “q”, cmsy10 “p”, cmsy7 “p”
* Supersuperscript: cmsy10 “p”, cmsy7 “p”, cmsy5 “p”

You can see that in pdftex, the largest surd is differently shaped than any other surd. This is because it is a different character.

In current katex, the groups of three use:
* Largest: KaTeX_Size2, KaTeX_Size1, KaTeX_Main
* Superscript: KaTeX_Size1, KaTeX_Main, KaTeX_Main
* Supersuperscript: KaTeX_Main, KaTeX_Main, KaTeX_Main

SImilarly, in current katex, you can see that the largest surd is differently shaped than any other surd.

In your checkin, I believe you are using KaTeX_Size2 for the outermost surd *at every level*. I just wanted to point out that this is different behavior.

ronkok:
Thank, @kohler, I realized later that you and I were talking at cross purposes. The word *main* sent me off on a different direction.  And your list of characters is helpful.

ronkok:
I believe that I have overcome my sizing problems in script style and scriptscript, as shown by the screenshot below.
sqrtsizing-2

@kohler, thank you for your help. I am now using the same glyphs as current KaTeX, and sizing them properly. Of course, that means that the outer scriptscript is different from LaTeX, and I'm not sure why.

@kevinbarabash, I think I'll rebase when I commit, to give you something less scattered to review. That will be this evening.

ronkok:
@kevinbarabash, I did battle with rebase this morning and I lost, as usual. I'm closing this PR and will make a fresh start with a new PR.

kevinbarabash:
@ronkok rebasing can be a pain especially if there a lots of commits with conflicts.

Issue #696, opened 2017-05-10
poor spacing in fraction inside square root (depends on rendering environment)


jason-s:
This may be related to #408. I have a fraction in a square root and there is no spacing at normal fontsize between the denominator and the vinculum of the fraction.

math in question: `\( K_m = K_e\sqrt{\frac{3}{2R}} \)`

KaTeX:
https://cdn.rawgit.com/jason-s/c913d7810ea9a72138eadbe3f87135dd/raw/1d4afa8441020a0245a463d12ec7d3f61e8c56b1/katex695.html (if you looked at this before May 11, please look again as I mistakenly posted a link to a MathJax implementation rather than KaTeX)
On my Lenovo W520 (1920x1080) running Windows 7, using Google Chrome:
![image](https://cloud.githubusercontent.com/assets/738893/25909277/8dc9545c-3561-11e7-8edc-8603b8850d2a.png)

On my Acer C720 Chromebook (1366 x 768) it does look ok at 100%:
![screenshot 2017-05-11 at 8 15 49 am](https://cloud.githubusercontent.com/assets/738893/25957003/213dc086-3622-11e7-9ef8-a390b42f9866.png)

but not at 70% where it seems to exhibit the same issue as on my Windows 7 laptop:

![screenshot 2017-05-11 at 8 24 33 am](https://cloud.githubusercontent.com/assets/738893/25957517/5d585742-3623-11e7-9186-60b5ba767ee9.png)

MathJax:
https://cdn.rawgit.com/jason-s/c913d7810ea9a72138eadbe3f87135dd/raw/37cf37aa54b4c83fe3c80e5fa013cab764173a3b/katex695-mathjax.html
![image](https://cloud.githubusercontent.com/assets/738893/25909710/e13cfb74-3562-11e7-9fd6-69b43c007936.png)

jason-s:
At 150% zoom it looks ok:

![image](https://cloud.githubusercontent.com/assets/738893/25909979/be28173a-3563-11e7-9f55-485012d15751.png)

jason-s:
(same example case as in #695)

edemaine:
Can you add LaTeX rendering for comparison?

jason-s:
hmm... will try, I have MiKTeX installed but I use it extremely rarely, just for TiKZ work.

jason-s:
here we go:

```
\documentclass{article}
\begin{document}
[...]

PR #776, opened 2017-07-31
Improve rule coding, including for `\sqrt`.

kohler:
* Rule widths (overline, underline, fraction, sqrt) actually scale
  with the current font size in AMS-TeX. Implement that. (Sqrt is
  a special case: the rule width depends on the font size *of the
  surd*.)
* Change the CSS coding for rules. The old, complex coding prevented
  variable-width lines and may have contributed to issues like #696.
  Its purpose, according to 0a3a2271f4b207feb14fe366cb4e72e48d79dd93,
  was IE8 support; but KaTeX no longer supports IE8.


kohler:
Example difference, using `\frac` towers (an example from @ronkok). Before:
screen shot 2017-07-30 at 11 41 21 pm

After:
screen shot 2017-07-30 at 11 40 47 pm

Note that the rules are less unruly.

kevinbarabash:
@kohler I'll review this evening.  The screenshots look good.

kevinbarabash:
> I'll review this evening

I was overly optimistic.  I'll try to get to it this weekend.

kevinbarabash:
The \sqrt rule looks a lot better.  I've found the smaller font sizes to be more troublesome.  Here are some renderings using 12px sized font.

Safari
screen shot 2017-08-04 at 10 25 13 pm

Safari @ 400%
screen shot 2017-08-04 at 10 30 48 pm

Chrome
screen shot 2017-08-04 at 10 25 32 pm

Chrome @ 400%
screen shot 2017-08-04 at 10 30 57 pm

I tried out some other font sizes.  18px was the worst.

Safari @ 400%
screen shot 2017-08-04 at 10 34 17 pm

Chrome @ 400%
screen shot 2017-08-04 at 10 34 34 pm

My guess is that the `\sqrt` rule is a bit to high on Chrome for 18px and is getting clipped.  This also happens at 11px and might happen for other font sizes.  I zoomed to 100% just in case the zoom was causing an issue and the rule was still missing.  This should be fixed before merging.

kohler:
Hi @kevinbarabash, I rebased on top of the current master & pushed. I cannot replicate a missing \sqrt rule on Chrome, with either 11px or 18px, at any mag. Here's the diff I made to main.css. Version 59.0.3071.115 (Official Build) (64-bit) on Mac. Can you help me replicate? There's no clipping in the CSS coding, so your proposed explanation doesn't make sense to me.

```
--- a/static/main.css
+++ b/static/main.css
@@ -1,7 +1,7 @@
 body {
     margin: 0px;
     padding: 0px;
-    font-size: 72px;
+    font-size: 11px;
 }
 
 #input {
```

kevinbarabash:
I'm running a dev version of Chrome.  It said it was ready to upgrade so I upgraded and the problem went away.

src/buildHTML.js
```diff
@@ -1078,25 +1079,33 @@ groupTypes.sqrt = function(group, options) {
...
+        lineClearance + theta) * options.sizeMultiplier;
 
     // Create a \surd delimiter of the required minimum size
-    const delim = makeSpan(["sqrt-sign"], [
-        delimiter.customSizedDelim("\\surd", minDelimiterHeight,
-                                   false, options, group.mode)],
-                         options);
+    const delimChar = delimiter.customSizedDelim("\\surd", minDelimiterHeight,
+            false, options, group.mode);
+    const delim = makeSpan(["sqrt-sign"], [delimChar], options);
```
kevinbarabash: thanks for breaking this up

src/buildHTML.js
```diff
@@ -1078,25 +1079,33 @@ groupTypes.sqrt = function(group, options) {
     // and line
     const inner = buildGroup(group.value.body, options.havingCrampedStyle());
 
-    const line = makeLineSpan("sqrt-line", options);
-    const ruleWidth = line.height;
+    // Calculate the minimum size for the \surd delimiter
+    const metrics = options.fontMetrics();
+    const theta = metrics.defaultRuleThickness;
```
kevinbarabash: Is `theta` the official name for this metric in the TeXBook?

src/buildHTML.js
```diff
@@ -432,12 +432,16 @@ groupTypes.genfrac = function(group, options) {
     const denomm = buildGroup(group.value.denom, newOptions, options);
 
     let rule;
+    let ruleSpacing;
+    let ruleWidth;
     if (group.value.hasBarLine) {
         rule = makeLineSpan("frac-line", options);
+        ruleWidth = ruleSpacing = rule.height;
```
kevinbarabash: Can you change this to:
```
ruleWidth = rule.height;
ruleSpacing = ruleWidth;
```
I misread this and thought that `ruleSpacing` wasn't being defined in this block.

src/buildHTML.js
```diff
@@ -1020,13 +1024,10 @@ groupTypes.katex = function(group, options) {
         ["mord", "katex-logo"], [k, a, t, e, x], options);
 };
 
-const makeLineSpan = function(className, options) {
-    const baseOptions = options.havingBaseStyle();
-    const line = makeSpan(
-        [className].concat(baseOptions.sizingClasses(options)),
```
kevinbarabash: Are the sizing classes not necessary b/c we're manually figuring setting the `borderBottomWidth`?

kohler: Yes. The sizing classes *were* necessary because previously we wanted all rules to have the same thickness, and that thickness was specified in `katex.less` in *base-sized* `em` units. Now that we're specifying the thickness in local `em` units, no sizing classes.

src/buildHTML.js
kohler: Yes

kohler:
When a nucleus has both sub & sub, the `defaultRuleThickness` is part of the calculation involved in separating the subscript from the superscript. The screenshotter subscript positions changed in scriptscript mode because now `defaultRuleThickness` is relatively larger for script and scriptscript fonts—e.g., the `script` `defaultRuleThickness` is 0.049em, not 0.04em.

The previous KaTeX implementation differed. It used 0.04em as the thickness for purposes of separating sup+sub, but it should have used 0.04em *base-size* units for consistency with the TeXbook (since previously all rules were 0.04em base-size units thick). Instead it used 0.04em *script-size* units (in scriptscript mode).

src/buildHTML.js
kohler: done

kevinbarabash:
@kohler thanks for the PR.  The updated horizontal rules look great.

PR #781, opened 2017-08-03
Improve \sqrt

ronkok:
Make \sqrt out of SVG background images, to ensure a perfect match at the junction between surd and viniculum.


ronkok:
This consolidates the work in my previous PR, as shown here:
sqrtsnip

ronkok:
The question is still open on whether this PR ought be accepted. My own inclination is to wait. Accept PR #776  now, and try it for six months. If it turns out to be insufficiently precise, this PR can be merged at any time.

kevinbarabash:
@ronkok due to the variability in the position of the `\sqrt` rule that remains after https://github.com/Khan/KaTeX/pull/776 I think we'll still want to accept this PR.

ronkok:
@kevinbarabash, It's still unclear to me whether this PR should be approved. In case you do want to merge it, I have updated it to accommodate PR #776.

kohler:
Hi @ronkok, I agree with @kevinbarabash that this PR should be merged eventually, but I printed the test page to PDF; see screenshot. I doubt printing KaTeX is common, so some messiness would be acceptable, but this is pretty extreme.

If you generated inline SVG would this be better?

Ideally KaTeX should be copy-pasteable/screenreadable too. Must the surd character completely disappear from the markup?
​​
screen shot 2017-08-08 at 7 03 53 am

ronkok:
@kohler, The surds have to be *background*-images, in order to get the horizontal stretchiness of the viniculum. Background images are, alas, not part of the DOM. 

Chrome and Firefox both have print settings that allow you include background-images.  I don't see anything similar in Edge.

kevinbarabash:
It's unfortunate about the background image not printing out, but we're using `background-image` for all of the stretchy accents and overline/underline.  In for a penny in for a pound.

kohler:
@kevinbarabash The stretchy accents and overline/underline are all new features with no preceding support; \sqrt is core. I think this should be taken seriously. My users print out KaTeX math with some frequency.

@ronkok Why do you say that only css background images can be made to stretch? With tables, positioning, etc., pretty much any element can be made to stretch, right?

ronkok:
Well, in no particular order,

I agree that `\sqrt` is core and that everything in PR #670 is much less vital. I am in no hurry to see this PR merged. 

You can stretch any image to any width and height, **if** you know a numeric value in advance.  I sought to meet the challenge of issue #587 by creating images that would stretch to match a span width that is not known in advance, that is instead defined by the  content of the span. That particular trick can be done with a background-image but not with ordinary images.  And so, everything in PR 670 is made of either span borders, as in `\boxed`, or background-images. 

You *can* print anything in this PR or in PR #670. For instance in Chrome, the setting is at: ⋮ | Print ... |  + More Settings ... |  Options: □ Background graphics. That's more complicated than I would like, but it is certainly within reach.

Here is a screenshot from Chrome's Print dialog:
print

kohler:
> You can stretch any image to any width and height, **if** you know a numeric value in advance. I sought to meet the challenge of issue #587 by creating images that would stretch to match a span width that is not known in advance, that is instead defined by the content of the span. *That particular trick can be done with a background-image but not with ordinary images.*

Really? Here's something like what you want, with inline SVG. The `span.c` sets the width and height of the SVG.

screen shot 2017-08-10 at 7 40 49 am

```html
<html>
<head>
<style>
span.a {
        display: inline-block;
        position: relative; /* establishes new flow root for contents */
}
span.b {
        position: absolute; /* absolute relative to containing span.a */
        width: 100%; /* 100% width of span.a */
        height: 100%; /* 100% height of span.a */
        background: rgba(255,0,0,0.4);
        z-index: -1;
}
svg {
        display: block;
        position: absolute; /* again, absolute relative to parent */
        width: 100%;
        height: 100%;
        stroke-width: 5;
}
</style>
</head>
<body>
<p><span class="a"><span class="b"><svg viewBox="0 0 1000 100" preserveAspectRatio="xMinYMid slice"><path d="M 0 100 L 20 2.5 L 1000 2.5" stroke="purple" fill="none"></svg></span><span class="c"> XYZ</span></span></p>

<p><span class="a"><span class="b"><svg viewBox="0 0 100 100" preserveAspectRatio="none"><path d="M 0 0 L 25 100 L 50 0 L 75 100 L 100 0" stroke="purple" fill="none"></svg></span><span class="c" style="font-size: 300%"> XYZ</span></span></p>

<p><span class="a"><span class="b"><svg viewBox="0 0 1000 100" preserveAspectRatio="xMinYMid slice"><path d="M 0 100 L 20 2.5 L 1000 2.5" stroke="purple" fill="none"></svg></span><span class="c" style="font-size: 300%"> XYZ ABC</span></span></p>

<p><span class="a"><span class="b"><svg viewBox="0 0 1000 100" preserveAspectRatio="xMinYMid slice"><path d="M 0 100 L 20 2.5 L 1000 2.5" stroke="purple" fill="none"></svg></span><span class="c" style="font-size: 300%"> XYZ ABC DEF</span></span></p>

<p><span class="a"><span class="b"><svg viewBox="0 0 1000 100" preserveAspectRatio="xMinYMid slice"><path d="M 0 100 L 20 2.5 L 1000 2.5" stroke="purple" fill="none"></svg></span><span class="c" style="font-size: 300%"> XYZ ABC DEF GHI JKL</span></span></p>
</body>
</html>
```

ronkok:
I stand corrected. That arrangement of nested spans is terrific, and I thank you for showing it to me.

It can be adjusted so that the SVG code need not be written in line. When I edit one of the paragraphs to read like this:
````
<p><span class="a">
    <span class="b">
        <img src="images/sqrt-main.svg" alt="sqrt" width="100%" height="100%">
    </span>
    <span class="c" style="font-size: 300%">
        <span style="padding-left:1em;">XYZ ABC</span>
    </span>
</span></p>
````
The result looks like this:
sqrt-cached

The SVG can thus be cached and it won't clutter up the HTML. Also, being in the DOM, it won't have the Edge `\color` problem that I got from background-images.

I'm going to start working up a PR that uses this method. I may be back with a question or two.

**Comment by kohler** [comment]
That's great.

FWIW, @ronkok, I still think we should prefer inline SVG to out-of-line SVG, for render-performance reasons (better in general to avoid external assets), and because my understanding is that many large sites have settled on inline SVG. For example:

https://github.com/blog/2112-delivering-octicons-with-svg

> We tried a number of approaches when adding SVG icons to our pages. Given the constraints of GitHub’s production environment, some were dead-ends.
> 
> 1. External .svg — We first attempted to serve a single external “svgstore”. We’d include individual sprites using the <use> element. With our current cross-domain security policy and asset pipeline, we found it difficult to serve the SVG sprites externally.
> 2. SVG background images — This wouldn’t let us color our icons on the fly.
> 3. SVGs linked via `` and the src attribute — This wouldn’t let us color our icons on the fly.
> 4. Embedding the entire “svgstore” in every view and using `<use>` — It just didn’t feel quite right to embed every SVG shape we have on every single page throughout GitHub.com especially if a page didn’t include a single icon.

The #1 issue “current cross-domain security policy and asset pipeline” is something that might bite some KaTeX users—or it may not.

**Comment by ronkok** [comment]
You know, I could have sworn that I was once able to color an external SVG, linked via an `` tag, by setting `<path fill="currentColor" …` inside the SVG file. It doesn't work today. I must have been hallucinating.

I guess inline SVG is the order of the day.

**Comment by kohler** [comment]
FWIW, the `span.b` doesn’t seem to be necessary. The `display:block;position:absolute` `svg` element can be a direct child of `span.a`.

**Comment by ronkok** [comment]
I've updated the [test page](http://copyasmarkup.com/katex/Investigate.html) with the latest commit. I think you'll find better printing behavior.

**Comment by ronkok** [comment]
@glebm, I don't know if you have monitored the conversation in this PR. I think you should be given the opportunity to state the anti-inline case. Do you have any thoughts?

**Comment by kevinbarabash** [comment]
@ronkok I'll send you a PR with screenshots for these changes.

**Comment by kevinbarabash** [comment]
sqrt

Green is LaTeX and red is KaTeX.  It looks like the surd is one size too big for the scriptscript root on the right.

**Comment by ronkok** [comment]
That's a problem, all right. I'll get to work.

**Comment by kevinbarabash** [comment]
@ronkok thanks for fixing the surd size.  I'll send you updated screenshots later today.

**Comment by glebm** [comment]
@ronkok WRT inlining.

External SVGs:

* Trigger a re-render of the katex block when loading the page.
* May slightly complicate using KaTeX from frameworks that generate resource filenames on deployment.

Inline SVGs:

* Approximately the same size because gzip eliminates much of the duplication.
* The browser has to parse each instance anew. This can be mitigated with `<use>`, but not completely as there may be multiple katex blocks / server-side rendering.
* Do not benefit from browser-side caching in the case of server-side rendering.

If the SVGs are small (no more than a few KiB), I'd go with inline, but no strong opinion.

**Comment by ronkok** [comment]
@glebm I'm glad your're okay with it, because I'm pretty sure this PR will be merged.

**Comment by kevinbarabash** [comment]
sqrt

There's still something a little wonky with the superscripts, but the sizes are matching much better.  I'm okay with merging this and tweaking this latter.  The problem this diff fixes is much worse than the minor differences I'm seeing now.  I won't get to reviewing the code until tomorrow evening.

**Comment by ronkok** [comment]
I wonder if there is a cramped/uncramped issue.

**Comment by kevinbarabash** [comment]
That might explain why the subscript looks good by the superscript is off.

File: static/katex.less [hunk 1]
```diff
@@ -399,23 +399,22 @@
...
         > .root {
             // These values are taken from the definition of `\r@@t`,
             // `\mkern 5mu` and `\mkern -10mu`.
             margin-left: 5*@mu;
             margin-right: -10*@mu;
         }
+
+        > span > span > .vlist {
+            text-align: left;
+        }
```
kevinbarabash: What is this for?

src/buildHTML.js
```diff
@@ -1095,33 +1101,30 @@ groupTypes.sqrt = function(group, options) {
...
-    // Create a \surd delimiter of the required minimum size
-    const delimChar = delimiter.customSizedDelim("\\surd", minDelimiterHeight,
-            false, options, group.mode);
-    const delim = makeSpan(["sqrt-sign"], [delimChar], options);
+    // Create a sqrt SVG of the required minimum size
+    const img = delimiter.customSizedDelim("\\surd", minDelimiterHeight,
+                    false, options, group.mode);
+
+    const surdWidth = img.width + "em";
+    // Actually img width will be set by CSS to match argument width.
```
kevinbarabash: nit: this comment should go above line 1108 to match the position of the other comments and remove "Actually"

src/delimiter.js
```diff
@@ -318,6 +313,112 @@ const makeStackedDelim = function(delim, heightTotal, center, options, mode,
         Style.TEXT, options, classes);
 };
 
+const sqrtInnerSVG = {
+    main : `<svg viewBox=\'0 0 400000 1000\' preserveAspectRatio=\'xMinYMin
```
kevinbarabash: Escaping isn't necessary in template strings.  This could be:
```
main: `<svg viewBox='0 0 400000 1000' preserveAspectRatio='xMinYMinslice'>
...
</svg>`
```
nit: we don't put a space before the colon in the rest of the code.

src/delimiter.js
```diff
@@ -318,6 +313,112 @@ const makeStackedDelim = function(delim, heightTotal, center, options, mode,
...
+        const refHeight = height / sizeMultiplier;
+        let newOptions = options.havingBaseStyle(Style.TEXT);
+
+        //In the font, the glyph is 1000 units tall. The font scale is 1:1000.
+        if (1 / newOptions.sizeMultiplier < refHeight) {
+            newOptions = options.havingBaseStyle(Style.SCRIPT);
+            if (1 / newOptions.sizeMultiplier < refHeight) {
+                newOptions = options.havingBaseStyle(Style.SCRIPTSCRIPT);
+            }
+        }
```
ronkok: Well, `delim.style` is set by function `traverseSequence` after it has iterated through a sequence of sizes and styles, so I too thought that `options.havingBaseStyle(delim.style)` would give me the correct sizeMultiplier. But it isn't reliable. Believe me, I tried. It would often give me surds that were too large.

In the end, I had to do my own test to get the style that would give the correct sizeMultiplier. I don't know if `traverseSequence` is passing bad style choices to other delimiters, but I do know that I can't trust it tp pick script or scriptscript style for a surd.

static/images/Image-Licensing-and-Technical-Notes.txt
```diff
@@ -19,11 +19,16 @@ The license terms above apply to the following images:
...
+    overbrace.svg             U+23A9/23A8/23A7   KaTeX_Size4-Regular
     rightarrow.svg            U+2192             KaTeX Main
     rightharpoon.svg          U+21C0             KaTeX Main
     rightharpoondown.svg      U+21C1             KaTeX Main
     rightleftharpoons.svg     U+21CC             KaTeX Main
+    sqrt-main.svg             U+221A             KaTeX Main
+    sqrt-size1.svg            U+221A             KaTeX_Size1-Regular
+    sqrt-size2.svg            U+221A             KaTeX_Size2-Regular
+    sqrt-size3.svg            U+221A             KaTeX_Size3-Regular
+    sqrt-size4.svg            U+221A             KaTeX_Size4-Regular
```
ronkok: Good point. I'll move the information.

static/katex.less
ronkok: In a numerator or a denominator, `text-align` is set to center. When I place a `\sqrt` into either one, I have to protect the surd from that alignment or else I get this: 
align

src/delimiter.js
ronkok: Re: template strings. Yes, indeed. And that's true for the comments below, as well.

src/delimiter.js
ronkok: BTW, the difference between my test and the one inside `traverseSequence` is that my test used `height / options.sizeMultiplier` instead of `height` as the reference for comparison.

src/buildHTML.js
```diff
@@ -1132,30 +1132,26 @@ groupTypes.sqrt = function(group, options) {
     if (inner.height === 0 && inner.depth === 0) {
         body = makeSpan();
     } else {
+        inner.style.paddingLeft = img.surdWidth + "em";
+        // Overlay the image and the argument.
         body = buildCommon.makeVList([
-            {type: "elem", elem: inner},
-            {type: "kern", size: lineClearance},
-            {type: "elem", elem: line},
-            {type: "kern", size: ruleWidth},
```
kevinbarabash: Where did this kern on top of the horizontal rule go?

src/delimiter.js
kevinbarabash: We're using `traverseSequence` to determine `delim.type`, but if it's doing the wrong thing (at least for image surds) then maybe `delim.type` is wrong in some cases.  Maybe we should determine whether it's `small`, `large`, or `stacked` using this technique as well.

src/buildHTML.js
ronkok: It hasn't gone anywhere that I know of. The proposed `sqrt` images are overlaid on the arguments. That is, the image spans vertically from above the argument to below the argument. They are not stacked above the arguments like the old `sqrt-line` span. So I called `makeVList` in individualShift mode instead of in firstBaseline mode. 

So the question is: What is the proper value of `imgShift`? I used:
```
const imgShift = img.height - inner.height - lineClearance - ruleWidth;
````
... which is my attempt to reproduce the previous behavior.

ronkok: Oh, excuse me. You said **on top of** the horizontal rule. You're right, I missed that one. I'll fix it.

ronkok:
@kevinbarabash I am going to close this PR temporarily. As a Git klutz, I am able to have only one PR at a time that is pointed at `master`. Right now, I want to fix stretchy element color. I'll reopen this PR when that effort is resolved.

kevinbarabash:
@ronkok you may want to try creating a branch per PR.  You can always delete the branch after the PR closes.

PR #787, opened 2017-08-12
Update function-support page for release 0.8.1

ronkok:


ronkok:
For a preview of the page, I've put a temporary copy [here](http://copyasmarkup.com/katex/function-support.html). I am especially interested in any comments on the [Units](http://copyasmarkup.com/katex/function-support.html#units) section.

edemaine:
Your documentation looks good, but your (*) footnote is no longer accurate, as [@kohler discovered](https://github.com/Khan/KaTeX/pull/732#issuecomment-312408720). In fact, `mu` units scale with scriptstyle/scriptscriptstyle, while other relative units always refer to the textstyle font in the current size.  (This is a quote from the source code. But these are all behaviors we inherit from TeX, so I don't know how much we need to say here...)

ronkok:
@edemaine, I've revised the units section at the [preview page](http://copyasmarkup.com/katex/function-support.html#units).  As a structural engineer, my tendency is to just draw the picture.

What do you think?

edemaine:
@ronkok Nice approach!

kevinbarabash:
@ronkok looks great.  I like how you called out the change in behavior for `\color`.

kevinbarabash:
I've opened https://github.com/Khan/KaTeX/pull/789 to update the demo page to 0.8.1 as well.  We should try to land these two PRs at the same time.

mbourne:
@kevinbarabash There seems to be a problem with swapped up and down triangles in the demo page.

image

\triangledown looks big, and \bigtriangleup looks small.

kevinbarabash:
@mbourne the issue with the triangles is an existing issue.  I did git blame on the lines in question and they haven't been changed in over 2 years.  Can you open a separate issue for the triangles?

ronkok:
@mbourne, Is your screenshot from the function-support page? When I look at that page, either [current](https://khan.github.io/KaTeX/function-support.html) or [proposed](http://copyasmarkup.com/katex/function-support.html), it looks to me like this:
triangles

kevinbarabash:
I thought there was supposed to be a size difference between `\triangle` and `\bigtriangleup` but that isn't actually the case.

ronkok:
Right. And `\triangleleft` really is smaller than `\triangle`.

edemaine:
`fdsymbol` defines `\triangle` smaller than `\bigtriangleup` (but still bigger than `\triangleleft`).  See [symbols-a4, Table 135](http://tug.ctan.org/info/symbols/comprehensive/symbols-a4.pdf#page=69).  So this issue depends on package use...  Personally, I'd be fine if we just made the `\bigtriangleup/down` bigger than they are now.  (We could even hack this now, by defining a macro `\bigtriangleup` &rarr; `\large\triangleup`.)

Reminds me to finally make issue #793 about listing packages that best reproduce KaTeX behavior.

edemaine:
When mentioning `\text` under the Style section, is it worth mentioning that nested `$...$` math is now also supported?

ronkok:
@edemaine Yes, I think that the nested `$...$` definitely deserves a mention. I'll write it in.

mbourne:
@ronkok My screen shot came from the page linked earlier in this thread, http://copyasmarkup.com/katex/function-support.html#units

But now that I look at it again today, I see the same as what your screen shot has. Did someone change something?

kevinbarabash:
@ronkok thanks for updating the function support page.  It looks great!

Issue #204, opened 2015-03-13
MathML for primes (') is wrong


NSoiffer:
On the [definite integral page](https://www.khanacademy.org/math/integral-calculus/indefinite-definite-integrals/definite_integrals/e/evaluating-definite-integrals), the encoding of `F'` is not quite right -- while the MathML properly uses `<msub>` for the prime, the prime character is in a `<mi>` but should instead be in `<mo>` as it is an operator. The MathML for the entire expr is:

``` html
<math><semantics>
  <mrow><mstyle displaystyle="true" scriptlevel="0">
     <msup>
        <mi>F</mi>
        <mrow> <mi mathvariant="normal">′</mi> </mrow>
     </msup>
     <mo>(</mo>
     ...
     </mstyle></mrow>
<annotation encoding="application/x-tex">\displaystyle F'(x) =  f(x)\,</annotation>
</semantics></math>
```

Although using `<mo>` for the prime char isn't mentioned in chapter 3 of the MathML spec, there is an example of it in [chapter 7](http://www.w3.org/Math/draft-spec/chapter7.html#chars.pseudo-scripts).

kevinb7:
@NSoiffer what tex code are you using to generate this MathML?

kevinb7:
For reference MathJax renders the prime in ```F'``` as:
```
<msup>
    <mi>F</mi>
    <mo>&#x2033;</mo>
</msup>
```
and ```F^\prime``` as:
```
<msup>
    <mi>F</mi>
    <mi class="MJX-variant" mathvariant="normal">&#x2032;<!-- ′ --></mi>
</msup>
```

PR #788, opened 2017-08-12
Fix MathML output for ' and large operators with limits

kevinbarabash:
Summary:
This diff extracts MathML builder tests out into a separate file and
updates them to use jest snapshot testing.  It updates the output of
prime ' from using identifier nodes <mi> to operator nodes <mo>.  It
also updates large operators with limits to use munderover instead
of msupsub.

Test Plan:
- make test w/o errors
- verify mathml snapshots contain the desired markup

Fixes: #204 and #198.


test/__snapshots__/mathml-spec.js.snap
```diff
@@ -0,0 +1,135 @@
...
+    <mrow>
+      <msup>
+        <mi>
+          f
+        </mi>
+        <mrow>
+          <mo mathvariant="normal">
+            ′
+          </mo>
+        </mrow>
```
edemaine: I was surprised to see the `mo` in an `mrow`.  Otherwise this markup matches the recommended markup [described in this Mozilla post](https://bugzilla.mozilla.org/show_bug.cgi?id=140439#c7).  Is it worth checking for a size-1 `mrow` element and removing it?  (I think we do this elsewhere.)

test/__snapshots__/mathml-spec.js.snap
```diff
@@ -0,0 +1,135 @@
...
+  <semantics>
+    <mrow>
+      <mi>
+        sin
+      </mi>
+      <mrow>
+        <mi>
+          x
+        </mi>
+      </mrow>
```
edemaine: See below. (though maybe this is for another PR...)

test/__snapshots__/mathml-spec.js.snap
kevinbarabash: Good point.  I'll see what I can do to remove the `mrow`.

test/__snapshots__/mathml-spec.js.snap
kevinbarabash: I added an option to remove unnecessary `mrow`s to `buildGroup`.  Right now it's only used for by `groupTypes.supsub`.  I'll see if it can be used elsewhere (everywhere?) in a follow up PR.

edemaine: Nice!  These new Jest snapshots sure make it easy to review the quality of the MathML.

PR #789, opened 2017-08-12
Update demo to v0.8.1

kevinbarabash:


bower_components/katex/dist/images/Image-Licensing-and-Technical-Notes.txt
```diff
@@ -0,0 +1,70 @@
+Copyright and Licensing
```
edemaine: Maybe I'm late to the party here, but is it worth calling this file `README.txt`?

ronkok: Well, I began that file strictly as a copyright notice. It seemed reasonable at the time to call it "Image Licensing.txt". The technical comments came later. Nothing makes any call to this file, so I have no strong feelings on what the file name is.

The KaTeX home page, https://khan.github.io/KaTeX/ and the test page http://utensil-site.github.io/available-in-katex/ both get their KaTeX by calling 
````
<script src="bower_components/katex/dist/katex.min.js" type="text/javascript"></script>
````

edemaine: Maybe it should be called `LICENSE.txt`?  I see, this isn't related to this PR at all, sorry!

Issue #786, opened 2017-08-12
Change "main" in package.json?


edemaine:
0.8.x is the first release using modern JavaScript.  This seems to behave poorly with NPM:

```
$ node -v
v6.9.4

$ node
> k=require('katex')
C:\Users\edemaine\Projects\coauthor\node_modules\katex\katex.js:10
import ParseError from "./src/ParseError";
^^^^^^
SyntaxError: Unexpected token import

> k=require('katex/dist/katex.js')
{ render: [Function: render],
  renderToString: [Function: renderToString],
  __parse: [Function: generateParseTree],
  ParseError: [Function: ParseError] }
>
```

Should we change the `"main"` pointer in `package.json` from `katex.js` to `dist/katex.js`?  Or is there another standard approach to this?

PR #791, opened 2017-08-13
Update package.json to point 'main' at dist/katex.js

kevinbarabash:
Summary:
Now that we're using ES6 import statements in our source code
it makes it difficult for people to consume the katex package
as a node module unless they're using node.js version 8.x.

This diff points the node entry point to dist/katex.js which
is the compiled version of katex.js.

Test Plan:
- publish
- create a new project node project and add the new npm katex package as a deep
- require katex in a test file and run it using node.js version 6.x

Fixes: #786


edemaine:
Testing:

* Install Node 6 (in my case, I use nvm so it's easy to switch between Node versions)
* Within `checkout point_main_to_dist`, run `npm link`
* Then in another directory, `npm link katex` and then try `node -e "require('katex')"`

But I get the same error.  If I change the `main` key to point to `lib/katex.js`, it works.

package.json
```diff
@@ -8,7 +8,7 @@
     "url": "git://github.com/Khan/KaTeX.git"
   },
   "files": [
-    "katex.js",
+    "dist/katex.js",
```
edemaine: In fact, I think this line edit should be reverted.  It's probably good to include the root (ES6) `katex.js` in the NPM package (why not? then people can choose). `dist` is already fully included 3 lines below.  It's the `main` key, 6 lines above, that seems to need to change.

kevinbarabash: Good call.  I totally messed this up.  I meant to change `main`.

edemaine:
Tested and it works.  Thanks @kevinbarabash for implementing this!

kevinbarabash:
@edemaine thanks for testing it.

kevinbarabash:
I should probably do an 0.8.2 release now that this has been merged.

edemaine:
Yeah, makes sense, as this will affect all NPM users.

kevinbarabash:
I was thinking that for future major release it might be good to do a couple of RC release to work out the kinks when there are big changes.

edemaine:
Given our experience with 0.8 and 0.7, that sounds like a good idea. 😄

Issue #28, opened 2014-09-07
Get rid of custom colors


xymostech:
Currently, we have functions like "\blue", which produce things in line with KA exercise colors, but these colors shouldn't be enforced/available for everyone. Instead, we should make a way to dynamically add extensions (like macros maybe).

schovi:
Is there any support (or plans) for color usage as in Tex/Matjax?
`\color{blue}x = 1,\color{255,0,0}y = 2`

Thanks

edemaine:
Revitalizing this issue after discussion in #792. `Options.js` has an extensive `colorMap` with comment `TODO(emily): Remove this when we have real macros`, and `functions.js` has an extensive `Single-argument color functions` (`\blue`, \orange`, `\pink`, `\red`, ...).  These could both be easily replaced with a macro object like:

```javascript
{`\\blue': `\\textcolor{#6495ed}{#1}`,
 ...}
```

or, in `macros.js`,

```javascript
defineMacro(`\\blue', `\\textcolor{#6495ed}{#1}`);
...
```

What remains unclear to me is where these macros belong.  They're not commands in LaTeX (except in many author's preambles), so my intuitive says "not in base KaTeX".  Maybe this is a good test for plugin infrastructure as discussed elsewhere?  If we made `defineMacro` accessible to the outside world (e.g., as `katex.defineMacro`), then we could make a `ka-colors.js` or `katex-colors.js` plugin, perhaps in a new `plugins` directory or just `contrib`.  When included in a project, you'd get those macros automatically, instead of having to pass them in.  Thoughts?

xymostech:
Thank you for revitalizing my TODO! I'm all for removing those color constants now that macro support exists. These colors were added to the initial implementation to maintain compatibility with our MathJax color macros: https://github.com/Khan/khan-MathJax/blob/master/kathjax-config.js#L21 They were definitely never intended to be part of base KaTeX, so removing them seems very reasonable.

KA only has a couple of entry points that call KaTeX, so we could manually register the macros in those few places and we'd be fine. I guess the bigger problem is that non-KA users are probably using them now, so figuring out a strategy for users to easily add back the colors to their versions would be needed before we fully remove them. Pulling them out into `macros.js` as a first step would be good. :)

kevinbarabash:
[...]

PR #792, opened 2017-08-13
Fix color support for stretchy, strikethrough, and fbox

xsznix:
Summary:
Stuff like `\red{\overbrace{AB}}` works now in addition to `\color{red}{\overbrace{AB}}`. Strikethrough now respects color. The Firefox in the screenshotter doesn't seem to support `background-image` + `mask`, but I manually tested that the latest Firefox does.

Test plan:
Ran `make`, then tested in latest Chrome and Firefox to ensure color support was working, then ran `make screenshots`.


ronkok:
Interesting. I did not realize that KaTeX supported that syntax. It does, though. `\red{A}` renders a red A.

Are there any LaTeX packages that support this syntax?

edemaine:
Nor did I!  [Here's the code](https://github.com/Khan/KaTeX/blob/master/src/functions.js#L352-L378).  And I guess there are matching `katex-...` colors in the CSS.  I'm not aware of a LaTeX package implementing this -- probably a KA thing. IMO, these might make more sense as macros...

ronkok:
Ah, I see. Well, @xymostech has written issue #28, which states that these custom colors were meant to be KA-only and not available for everyone. There is some push back in the discussion.

I suppose one approach would be to accept this PR, so that KaTeX has consistent behavior. Then, if a PR is ever written to resolve #28, it would have to address the whole issue.

spicyj:
Yes, the intent was always to move those to macros as soon as was possible.

I think this change seems pretty unambiguously good especially in the existing presence of the options.getColor method although I don't understand the significance of the CSS change.

ronkok:
The CSS change fixes a bug which has crept into function `stretchy.encloseSpan()`. At one time, I pushed `mask` onto the list of classes for the relevant span. Somehow, I let that item slip out. The CSS in this PR corrects for my error.

Thank you, @xsznix! 

Also, Firefox 53 was the first Firefox release with full `mask` support. I suppose Screenshotter is using an older version.

edemaine:
In this case, perhaps we should accept this PR and then make a new one to rewrite `\red` etc. in terms of macros?  I could use some clarification on whether those macros belong in KaTeX (e.g. for backwards compatibility) or in another KA project.

kevinbarabash:
> In this case, perhaps we should accept this PR and then make a new one to rewrite \red etc. in terms of macros? 

I think that makes sense.

> I could use some clarification on whether those macros belong in KaTeX (e.g. for backwards compatibility) or in another KA project.

As per the comments in #28, we definitely want to move KA custom colors out of KaTeX but in a thoughtful way.

kevinbarabash:
@xsznix thanks for the PR.  @edemaine thanks for reviewing.

PR #794, opened 2017-08-13
Advanced macro support and magic \dots

edemaine:
This is a re-implementation of #599 (`\dots` support) but extending the MacroExpander as necessary to build it as a macro instead of a function, following [@gagern's outline](https://github.com/Khan/KaTeX/pull/599#issuecomment-271854903).

Along the way, we get some nice new macro features:
1. Macros (e.g. given by `defineMacro`) can now be JavaScript functions, with `this` bound to the `MacroExpander`.  (Note the slightly different interface from functions, which get a complex context object -- up for discussion!)  This lets us extend the `MacroExpander` in more powerful ways, and use the following features:
2. `future()` returns the next token without expansion (similar to `\futurelet` but without the "let" aspect of assigning to a variation)
3. `expandAfterFuture()` returns the next token after one level of expansion (equivalent to a careful use of `\expandafter` and `\futurelet`)
4. `\relax` is now implemented!  It stops expansion, but doesn't actually get returned from the `MacroExpander`.

We also get fixes to old KaTeX bugs (with new tests to confirm, some of which used to fail):
1. `\text{\textellipsis !}` used to include a space in the middle. The trouble is that we were only eating spaces when in math mode, but in text mode we need to be more careful to eat the spaces that make up part of the command name. This is now consumed during macro expansion, even though it's not technically a macro expansion (`\textellipsis` is an unexpandable character).
2. Ditto for `\text{\foo }` where `\foo` is defined to something -- we got a space, but shouldn't have.
3. Macros did not handle command arguments correctly.  `\def\foo#1{(#1)}\foo\bar` should expand to `(\bar)` and then expand `\bar`, whereas the old macro expander would first expand `\bar` and then use its tokens for arguments.  For example, this didn't work if `\def\bar{}` or `\def\bar{ }` (especially the latter because of bad space ignoring).  Also, if `\foo` is a multiargument function, `\bar` should constitute one argument instead of one per token.  (I did a bunch of testing in LaTeX to confirm this behavior.)

A few other miscellaneous changes:

* `Lexer.js` optionally exports `Token`, so that I can do `instanceof` tests.
* `consumeSpaces()` irreversibly consumes space tokens.  This is used to fix the first bug listed above.
* `nextToken()` got split into `expandOnce()` and `expandNextToken()`.  So, e.g., `expandAfterFuture()` is equivalent to `expandOnce()` and then `future()`.  (This makes for some awkward diffs -- sorry!)
* The cdots symbol is now `\cdots@`, because amsmath defines a more complex `\cdots` in terms of this symbol.
* `@` is now considered a valid character for commands, as if `\makeatletter` is in effect.  I don't think we currently define `\@` but if we did this change would be annoying.
* A rewrite of #599 to handle `\dots`, `\cdots`, etc. correctly.

Here's a texcmp confirming I got at least the tested cases right:

dots


kevinbarabash:
@edemaine this is awesome.  I don't have a very good understanding of macros so hopefully @gagern and others can help out with the review.

src/macros.js
```diff
@@ -38,6 +43,156 @@ defineMacro("\\iff", "\\;\\Longleftrightarrow\\;");
...
+// starting with \DOTSX implies \dotso, and then \extra@ detects this case
+// and forces the added `\,`.
+defineMacro("\\dotsx", "\\ldots\,");
+
+// \let\DOTSI\relax
+// \let\DOTSB\relax
+// \let\DOTSX\relax
+defineMacro("\\DOTSI", "\\relax");
+defineMacro("\\DOTSB", "\\relax");
+defineMacro("\\DOTSX", "\\relax");
```
edemaine: amsmath defines these "null macros" for other macros to indicate the behavior of `\dots` preceding them.  Indeed, I should have added use of these macros to `\iff`, `\implies`, and `\impliedby`.  See new commit.

src/Lexer.js
```diff
@@ -75,14 +75,14 @@ const tokenRegex = new RegExp(
     "([ \r\n\t]+)|" +                                 // whitespace
     "([!-\\[\\]-\u2027\u202A-\uD7FF\uF900-\uFFFF]" +  // single codepoint
     "|[\uD800-\uDBFF][\uDC00-\uDFFF]" +               // surrogate pair
-    "|\\\\(?:[a-zA-Z]+|[^\uD800-\uDFFF])" +           // function name
+    "|\\\\(?:[a-zA-Z@]+|[^\uD800-\uDFFF])" +          // function name
```
edemaine: Good to have at least one negative reaction -- I was curious what others would think.  I see three alternatives:

1. Add a `Lexer` option for whether to include `@` as a letter.  Then we could actually support `\makeatletter` and `\makeatother` (presumably this would need to be carried in the `Settings` object...).
2. Have the `cdots` macro return manually parsed `Token`s instead of strings.  Then we don't need to touch the `Lexer`.
3. Rename `\@cdots` to something else like `\cdotsINTERNAL` or `\latexcdots`.

Thoughts/preferences?

edemaine: Incidentally, the regex change won't affect `@{...}` support in `tabular` (the only use I know of `@` as an active char), nor will it affect the most typical usage of `\@`, namely, `\@.`.  What it would break is if you used `\@` immediately followed by a letter, which ... I can't imagine doing.  So the other option is to leave this change as is for now.

kohler: That is where I came down and for a similar reason: keep `@` as is in this commit.

kevinbarabash:
I'm going to try to get this reviewed over the next couple of days.

edemaine:
Thanks @kevinbarabash!  I just rebased to fix a conflict with the CJK tweak on `\dots` commands.

src/MacroExpander.js
```diff
@@ -17,114 +17,194 @@ class MacroExpander {
...
+        const name = topToken.text;
+        const macro = (name.charAt(0) === "\\");
+        if (macro) {
+            // Consume all spaces after \macro
+            this.consumeSpaces();
+        }
+        if (!(macro && this.macros.hasOwnProperty(name))) {
+            // Fully expanded
+            this.stack.push(topToken);
+            return topToken;
```
kevinbarabash: This is a little weird to pop `topToken` at the start just to push it back on.  I can't think of demonstrably better way to do this though.

src/MacroExpander.js
```diff
@@ -17,114 +17,194 @@ class MacroExpander {
...
+        }
+        // Concatenate expansion onto top of stack.
+        this.stack.push.apply(this.stack, expansion);
+        return expansion;
+    }
+
+    /**
+     * Expand the next token once (ignoring initial spaces like `get`),
+     * without removing anything from the stack, and return the top token
+     * on the stack.  Similar in behavior to TeX's `\expandafter\futurelet`.
```
kevinbarabash: How does it ignore initial spaces if it calls `expandOnce` right away which only ignores trailing spaces?

src/MacroExpander.js
```diff
@@ -17,114 +17,194 @@ class MacroExpander {
...
+    /**
+     * Recursively expand first token, then return first non-expandable token.
+     */
+    expandNextToken() {
+        for (;;) {
+            const expanded = this.expandOnce();
+            // expandOnce returns Token if and only if it's fully expanded.
+            if (expanded instanceof Token) {
+                // \relax stops the expansion, but shouldn't get returned (a
+                // null return value couldn't get implemented as a function).
```
kevinbarabash: I'm not sure what `a null return value couldn't get implemented as a function` means.  Could you give an example?

src/macros.js
```diff
@@ -3,6 +3,9 @@
  * This can be used to define some commands in terms of others.
  */
 
+import symbols from "./symbols";
+import utils from "./utils";
+
 // This function might one day accept additional argument and do more things.
 function defineMacro(name, body) {
```
kevinbarabash: Once we get flow in place we'll be able to type this as:
```
function defineMacro(name: string, body: string | () => string) {
```

src/Lexer.js
ronkok: FWIW, The MathJax [mhchem extension](https://github.com/mhchem/MathJax-mhchem/blob/master/mhchem.js) makes heavy use of `@`. I'm no expert, so I can't tell you if it is being used as a catcode.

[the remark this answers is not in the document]
ronkok: Correction. `mhchem` only uses `@` internally in its JavaScript code. `@` is not part of its [macro syntax](https://mhchem.github.io/MathJax-mhchem/).

Never mind.

test/katex-spec.js
```diff
@@ -2300,13 +2305,62 @@ describe("A macro expander", function() {
...
+        compareParseTree("\\text{\\foo 1 2}", "\\text{12end}", {"\\foo": "#1#2end"});
+        compareParseTree("\\text{\\foo {1} {2}}", "\\text{12end}", {"\\foo": "#1#2end"});
+    });
+
     it("should allow for multiple expansion", function() {
         compareParseTree("1\\foo2", "1aa2", {
             "\\foo": "\\bar\\bar",
             "\\bar": "a",
         });
     });
```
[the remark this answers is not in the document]
edemaine: For fun, I added such a test in 56d8900583406e0dd309de784cd8d39610e1057e

src/MacroExpander.js
```diff
@@ -17,114 +17,194 @@ class MacroExpander {
     }
 
     /**
-     * Recursively expand first token, then return first non-expandable token.
+     * Return the next unexpanded token without removing anything from the
+     * stack.  Similar in behavior to TeX's `\futurelet`.
+     */
```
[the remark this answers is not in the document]
edemaine: Nice rewrite.  Implemented.

src/MacroExpander.js
```diff
@@ -17,114 +17,194 @@ class MacroExpander {
...
+                this.stack.pop();
+            } else {
+                break;
+            }
+        }
+    }
+
+    /**
+     * Expand next token only once, and leave it on the stack.
+     * Returns the token or its expansion.
```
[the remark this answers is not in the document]
edemaine: Revised accordingly.

src/MacroExpander.js
```diff
@@ -17,114 +17,194 @@ class MacroExpander {
...
      * See the TeX book page 202ff. for details on how those should behave.
      */
-    nextToken() {
-        for (;;) {
-            if (this.stack.length === 0) {
-                this.stack.push(this.lexer.lex());
+    expandOnce() {
+        const topToken = this.popToken();
+        const name = topToken.text;
+        const macro = (name.charAt(0) === "\\");
```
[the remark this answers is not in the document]
edemaine: Changed.

src/MacroExpander.js
edemaine: I agree, unfortunately on both counts.  If I didn't have to `consumeSpaces`, I could use `future()` and then `popToken` after the fully expanded case.  Maybe there's another place to put the `consumeSpaces`... but at least this works.

src/MacroExpander.js
```diff
@@ -17,114 +17,194 @@ class MacroExpander {
...
+        }
+        if (!(macro && this.macros.hasOwnProperty(name))) {
+            // Fully expanded
+            this.stack.push(topToken);
+            return topToken;
+        }
+        let expansion = this.macros[name];
+        if (typeof expansion === "function") {
+            expansion = expansion.call(this);
+        }
```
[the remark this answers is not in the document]
edemaine: Technically, it could also return an `Array` that's a reverse-ordered list of tokens.  This was functionality from before that I'm preserving, though I'm not sure we'd want to use it exactly that way.  (In particular, the reverse-order aspect is weird.)  But it is a nice way for a macro to be able to force strange parsing.

src/MacroExpander.js
```diff
@@ -17,114 +17,194 @@ class MacroExpander {
...
-                let numArgs = 0;
-                if (expansion.indexOf("#") !== -1) {
-                    const stripped = expansion.replace(/##/g, "");
-                    while (stripped.indexOf("#" + (numArgs + 1)) !== -1) {
-                        ++numArgs;
+            expansion.reverse(); // to fit in with stack using push and pop
+            expansion.numArgs = numArgs;
+            // Old approach saved the preprocessed expansion for future use.
+            // But this is incompatible with new function-based macros.
+            //this.macros[name] = expansion;
```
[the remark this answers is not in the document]
edemaine: Replaced with a TODO.  I'm not sure this is important, as TeX doesn't do any such caching (as far as I'm aware), but could be done in the future.

src/MacroExpander.js
edemaine: That comment was out-of-date.  Updated.

src/macros.js
edemaine: Added.

src/MacroExpander.js
edemaine: Functions (as defined by `defineFunction`) return a `ParseNode` or an object that gets converted into a `ParseNode` by the parser.  It's not possible for a function to return `null` that would just disappear in the final parse tree.  Hence it would be impossible to implement a `\relax` that expands to nothing (which is different to how `{}` expands).

edemaine:
Sorry for the delay.  Finally went through and implemented all the comments.  Thanks for the review!!  

Incidentally, this PR also fixes a bug in master with `\text{\v x}`.  Currently, because of incorrect space handling, this renders as an accent (over a space) followed by an `x`.  With the PR, it renders correctly (same as `\text{\v{x}}` does now).

src/macros.js
```diff
@@ -3,8 +3,11 @@
  * This can be used to define some commands in terms of others.
  */
 
+import symbols from "./symbols";
+import utils from "./utils";
+
 // This function might one day accept additional argument and do more things.
-function defineMacro(name, body) {
+function defineMacro(name: string, body: string | () => string) {
```
marcianx: Cool! Mind adding `@flow` at the top of this file?

edemaine: Oops, indeed!  Clearly still a flow noob...

marcianx: Me too, for all that's worth. ;)

src/MacroExpander.js
kevinbarabash: So if we return `null` the parser would think that it's an object an convert it to a `ParseNode`.  I wonder when would we ever want to write a function that returns a `null`?

edemaine: Long-term, I vaguely hope we can transition functions over to macros, as that's what TeX really does.  So maybe not too critical... though certainly could be done by tweaking group parsing.

Issue #638, opened 2017-01-20
Text-mode accent functions


edemaine:
KaTeX already supports [math-mode accents](https://en.wikibooks.org/wiki/LaTeX/Special_Characters#Math_mode), but doesn't support any of the usual accent commands in text mode.  According to [this list of text-mode LaTeX accent commands](https://en.wikibooks.org/wiki/LaTeX/Special_Characters#Escaped_codes).  We're missing
```\`{o}, \'{o}, \^{o}, \"{o}, \H{o}, \~{o}, \c{c}, \k{a}, \l{}, \={o}, \b{o}, \.{o}, \d{u}, \r{a}, \u{o}, \v{s}, \t{oo}, \o```.  A lot of these should be straightforward by translating into Unicode, using combining characters.

gagern:
Looking at [a text version of KaTeX-Main-Regular](https://github.com/gagern/KaTeX/blob/b8d05bc2266e81f4c54c6d02b27fed1c287c70bb/static/fonts/KaTeX_Main-Regular.ttx), the font should contain combining characters for (in order of map code)

* `` \` `` U0+300 grave
* `\'` U+0301 acute
* `\^` U+0302 circumflex
* `\~` U+0303 tilde
* `\=` U+0304 macron
*  `\u` U+0306 breve
* `\.` U+0307 dot above
* `\"` U+0308 diaresis
* `\r` U+030A ring above
* `\H` U+030B double acute
* `\v` U+030C caron

I checked that inserting one of these combining characters after a text symbol does indeed render correctly, i.e. has correct horizontal placement. As these already work for math formulas, we should have correct metrics there, too. So I'd say yes, adding these should be simple.

We also have U+0338, the combining *long* solidus, but for lower-case `\o` we'd need the combining *short* solidus instead, I think. But the new fonts currently discussed in #632 do include precomposed glyphs called `oslash` and `Oslash`, so we should be able to use those pretty soon.

I don't see support for the following symbols:

* `\c` U+0327 cedilla
* `\k` U+0328 ogonek
* `\l` U+0142 barred l
* `\b` U+0331 macron below
* `\d` U+0323 dot below
* `\t` U+0361 double inverted breve

@kevinbarabash, do you know which of these symbols are available in the TeX fonts from which our font building pipeline takes its input? It would be nice to include as many of these as possible.

[...]

PR #796, opened 2017-08-16
Add latin-1 letters as math symbols

chrismessiah:
Currently katex throws an EOF parse error when receiving an input of latin-1 letters that is not inside a `\text{}` clause. To prevent this the latin-1 letters are added as math-symbols as well. 

skarmavbild 2017-08-16 kl 09 47 12

### DIFF
--- a/src/symbols.js
+++ b/src/symbols.js
@@ -671,16 +671,19 @@ for (let i = 0; i < letters.length; i++) {
 // Latin-1 letters
 for (let i = 0x00C0; i <= 0x00D6; i++) {
     const ch = String.fromCharCode(i);
+    defineSymbol(math, main, mathord, ch, ch);
     defineSymbol(text, main, textord, ch, ch);
 }
 
 for (let i = 0x00D8; i <= 0x00F6; i++) {
     const ch = String.fromCharCode(i);
+    defineSymbol(math, main, mathord, ch, ch);
     defineSymbol(text, main, textord, ch, ch);
 }
 
 for (let i = 0x00F8; i <= 0x00FF; i++) {
     const ch = String.fromCharCode(i);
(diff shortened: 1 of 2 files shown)


spicyj:
I can't tell, is that Computer Modern in your screenshot? Not sure if our fonts support these chars.

chrismessiah:
Good catch. Didn't notice that they were using different fonts. What is the procedure for adding such letters to the font? 

skarmavbild 2017-08-16 kl 21 46 54

kevinbarabash:
Oh noes... we aren't using CM for latin-1 chars inside `\text` either.  Having this work correctly will require #632 and the some additional work to actually render combining marks for each of the accent types.  We'll also need to update how we calculate the height (ascent) and depth (descent) for the combined glyphs that we construct.  There's a fair amount of work involved so it's probably best to split it up.  I'm okay with this diff as is, but we should aim for rendering CM glyphs for latin-1 eventually.

edemaine:
Given that KaTeX already supports math-mode accents, we should be able to macro each Latin-1 Unicode character in math mode to the corresponding accent command (ideally with a loop over the characters and a call to some Unicode splitting).

To fix Latin-1 characters in text mode, we should first do #638 (text-mode accent functions).  Then we can follow the above approach.

kevinbarabash:
@edemaine can a macro be defined for something without a leading `\`?

kevinbarabash:
I've added a project to track work relating to latin-1 support.

edemaine:
@kevinbarabash No, macros currently need to start with `\`. But we could easily add support for single-character macros (in the spirit of TeX active characters). I'll give it a try...

edemaine:
Moving discussion to #801

Issue #735, opened 2017-06-17
Misplacement of top arrow in Safari


ylian:
The following is displayed incorrectly in the latest version of Safari on macOS Sierra, retina MBP.

```
\vec x
```
![image](https://user-images.githubusercontent.com/984895/27251567-e71052c6-5349-11e7-940a-b90284017876.png)

It works fine on Chrome.

edemaine:
Just to make sure, are you using `<!DOCTYPE html>`? Does the same behavior appear on our demo page?

kevinbarabash:
I was able to reproduce this using the demo page in Safari 10.

ronkok:
`\vec` uses a combining character, so it has its own unique CSS with `position: relative` and `left: 0.326em`.  The `buildHTML.js` file contains a TODO line suggesting 
````
Fix this in a better way, like by changing the font
````

I'm not volunteering to change the font, but if you're interested, I could give you a temporary solution using an SVG image copied directly from the `\vec` character.

kevinbarabash:
I've noticed while working on https://github.com/Khan/KaTeX/pull/140 that Safari doesn't like combining characters.  :sad:

kevinbarabash:
MathJax is having the same issue, https://github.com/mathjax/MathJax/issues/1737.  There's a link to a fix at the bottom of that issue.

ylian:
Yes indeed. I actually noticed the issue on MathJax before I tried it on KaTeX. Thanks for looking into the issue.

ronkok:
Yesterday, I  tried inserting a non-breaking space:
````
if (group.value.label === "\\vec") {
    accent.value = "\u00A0" + accent.value;
}
````
... and then adjusting the advance value in CSS class `accent-vec` by 0.25em (because nbsp is 0.25 em wide). I found that the arrow came in too low. I don't think I'll get any farther with it. I've got other issues at the moment.

kevinbarabash:
@ronkok thanks for giving it a try.  I'll play around with it when I have some time.

edemaine:
#802 may now fix this bug.

ronkok:
@ylian I believe that release 0.9.0-beta has fixed this bug. Can you confirm?

ylian:
@ronkok Yes, I can confirm this bug is fixed in release 0.9.0-beta. Thanks!

PR #802, opened 2017-08-17
Support for \' \` \^ \~ \= \u \. \" \r \H \v text-mode accents

edemaine:
I took a stab at implementing the text-mode accents that were easy to implement using current fonts, as outlined by @gagern in #638.  Here's a sample screenshot:

accents
```latex
\begin{array}{l}
\text{e \'e \`e \^e \~e \u{e} \.e \"e \r{e} \H{e} \v{e}}\\
\text{l \'l \`l \^l \~l \u{l} \.l \"l \r{l} \H{l} \v{l}}\\
\text{X \'X \`X \^X \~X \u{X} \.X \"X \r{X} \H{X} \v{X}}
\end{array}
```

This seems to work, but feels a bit hacky because the accents are combining characters so don't behave well in isolation, so there's a magic number in a CSS style (similar but different to the `\vec` accent).  It also misses key accents like `\c` because they're not in the font.  Makes me wonder whether this is worthwhile, or we should correct the font somehow to make non-combining characters for these accents like we have for math mode.  On the other hand, this support is better than nothing for now.


ronkok:
Does this work properly on Safari? Issue #735 is a combining character problem.

edemaine:
@ronkok Probably not -- should have the same issue.  Another reason to "wonder whether this is worthwhile, or we should correct the font somehow to make non-combining characters for these accents like we have for math mode. On the other hand, this support is better than nothing for now."

kevinbarabash:
We should look at porting this fix https://github.com/mathjax/MathJax/pull/1775/files.  I feel like that's a separate issue/PR.

ronkok:
Yesterday, I  tried inserting a non-breaking space:
````
if (group.value.label === "\\vec") {
    accent.value = "\u00A0" + accent.value;
}
````
... and then adjusting the advance value in CSS class `accent-vec` by 0.25em (because nbsp is 0.25 em wide). I found that the arrow came in too low. I don't think I'll get farther with it. I've got other issues at the moment.

**Comment by edemaine** [comment]
Cool, I hadn't seen that.  Now that I understand the accent generation code, I pushed a fix which ought to work for Safari.  Can someone test?  In my experiments, `0.262em` seems a closer approximation to the width of a space -- how did you compute `0.25em`?

**Comment by ronkok** [comment]
I got the 0.25 from http://glyphrstudio.com/online/. I don't take it as conclusive.
spacewidth

**Comment by ronkok** [comment]
Safari on the left, Chrome on the right.
accents

**Comment by ronkok** [comment]
@edemaine Instead of combining characters, could you use modifier characters?

**Comment by edemaine** [comment]
Hmm, not sure why that's not working.  Perhaps the combination of `left` and `margin-left`?  I don't think we have a choice of the type of character, if we fix the font as-is.

**Comment by kevinbarabash** [comment]
screen shot 2017-08-19 at 1 22 40 am

The Chrome changes look good for me.  I started playing around with the styles on Safari and it looks like `margin-left` should be `-spaceWidth + xMin` for each of the accents.  The `xMin` for the accents is a negative number b/c it's to the left of the origin.  

The width of the space character is 0.250em according to opentype.js (I uploaded KaTeX_Main-Regular.ttf).  For each of the accents the average of `xMin` and `yMin` is -0.250em which ends of centering it perfectly over the center of the space.  Since Chrome renders the accent character correctly over the space, all we have to do is to correct for the space alone which is why the `margin-left: -0.250em` does the trick.

Chrome however treats the accent as a character with a real width from `xMin` to the origin.  In the case of the umlaut which has an `xMin` of -0.405em, the resulting correction is `margin-left: -0.655em`.

screen shot 2017-08-19 at 1 34 01 am

screen shot 2017-08-19 at 1 34 46 am

Notice that this doesn't quite fix the single dot accent to the left of the umlaut.  This is because its `xMin` is a different value.

This means that the solution is different for Safari and Chrome.  The MathJax hack special cases Safari as well.  The unfortunate result of this is that server side rendering will only work if you return different markup based on the browser's user agent.  For client side rendering special casing shouldn't pose a problem.

One last note: fontMetricsData.js doesn't actually contain the values we need for the accent characters to fix Safari so we're going to need to extract those from the font.

**Comment by kevinbarabash** [comment]
@edemaine let's not worry about fixing Safari accent positioning in this diff.

File: static/katex.less [hunk 1]
```diff
@@ -513,6 +513,15 @@
             // it left. We center it by shifting it half way right again.
             left: 0.326em;
         }
+
+        .accent-body.accent-text > span {
+            position: relative;
+            left: 0.262em;  // width of space, apparently
```
kevinbarabash: The width, at least for KaTeX_Main-Regular, is 0.250em.  Thankfully it appears that this value is the same, at least for the fonts I checked which were KaTeX_Main-Bold and KaTeX_SansSerif-Regular.

src/buildHTML.js
```diff
@@ -1401,10 +1401,29 @@ groupTypes.accent = function(group, options) {
...
+            "\\.", '\\"', "\\r", "\\H", "\\v"], group.value.label)) {
+            accentClass = "accent-text";
+        }
 
-        accentBody = makeSpan(["accent-body", vecClass], [
-            makeSpan([], [accent])]);
+        // Some browsers (Safari in particular) don't like combining
+        // characters without a preceding character.  For each such accent,
+        // we add an artificial nonbreaking space with a negative margin
+        // to compensate for its width.
```
kevinbarabash: It's probably worth leaving this code in here with a TODO describe what's necessary to get the accents lining up on Chrome that way other contributors can build on top of the foundation you started here.

edemaine:
@kevinbarabash I'm a little confused about whether I should include the space hack in this patch at all.  My impression was that it wasn't necessary at all for Chrome, but perhaps your investigation has found that it makes it easier to get the horizontal offset more consistent?  Still grokking.

kevinbarabash:
@edemaine I didn't try removing the hack.  If all the new accents work on Chrome without the hack, let's take it and get this merged and we can deal with Safari separately.

edemaine:
[opentype.js](https://opentype.js.org/glyph-inspector.html) was insightful!  I realized that most of these accent characters have noncombining versions in the current font.  I switched to those, so they should work in Safari.  The only missing one is `\H`.  I still use combining characters for that one.

I found that the width of a space (0.250em) looked best for `\H`.  I would have thought something like: offset by -xMin to start at position 0; then shift back by half of the width (xMax - xMin) to center the accent.  This worked well for e.g. `\"` (not that I need that anymore), but I think only because it ended up resulting in an almost identical offset of 0.2505em...  @kevinbarabash If you understand this better, let me know if you think this offset is slightly off...

Other changes:
* Removed the Safari hack (no more space before the accent)
* Added `allowedInMathMode` to forbid these commands in math mode (as they don't work there -- and nor do they in LaTeX).  This will interact with #766.
* Added a screenshot test

kevinbarabash:
To fix `\H` for Safari we could look at updating the font to create a non-combining version of the glyph, but that's another PR.  Also, if we end up not using the combining versions of these glyph we might look at removing them completely from the font to save bytes.

src/Parser.js
```diff
@@ -561,6 +561,11 @@ class Parser {
                     throw new ParseError(
                         "Can't use function '" + func + "' in text mode",
                         baseGroup.token);
+                } else if (this.mode === "math" &&
+                    funcData.allowedInMath === false) {
+                    throw new ParseError(
+                        "Can't use function '" + func + "' in math mode",
```
kevinbarabash: For future diffs, this could be written as:
```
`Can't use function '${func}' in math mode`,
```

test/errors-spec.js
```diff
@@ -140,6 +140,11 @@ describe("Parser:", function() {
                 "Can't use function '\\sqrt' in text mode" +
                 " at position 7: \\text{\\̲s̲q̲r̲t̲2 is irrational…");
         });
+        it("rejects text-mode-only functions in math mode", function() {
+            expect("\\'echec").toFailWithParseError(
+                "Can't use function '\\'' in math mode" +
+                " at position 1: \\̲'̲echec");
+        });
```
kevinbarabash: Thanks for enforcing this and writing a test to verify enforcement.

src/Parser.js
edemaine: Indeed. (And thanks for teaching me ES interpolation strings in the first place!) You can tell I was copy/pasting from above. 😄

edemaine:
@kevinbarabash Thanks for the review!  I agree that we don't need combining characters in the font, and it'd be nice to add `\H`, `\"`, and `\vec` accents as non-combining characters.

PR #803, opened 2017-08-19
Add a maxSize option to limit user-specified sizes

### LINKED ISSUE #109: safe mode (user input mode)
I don't know what the policy is about LaTeX content posted to UGC sites, whether or not you're patching bugs that deal with this.

However the following TeX was found by my users and they used it to (temporarily) break my site until I filtered \rule:

```
\rule{999em}{999em}
```

I think it should have an upper maximum.

xsznix:
This fixes #109 by capping the maximum size that can be calculated.


src/Settings.js
```diff
@@ -24,6 +24,7 @@ class Settings {
         this.errorColor = utils.deflt(options.errorColor, "#cc0000");
         this.macros = options.macros || {};
         this.colorIsTextColor = utils.deflt(options.colorIsTextColor, false);
+        this.maxSize = utils.deflt(options.maxSize, 0);
```
kevinbarabash: Maybe this could be `Math.max(0, utils.deflt(options.maxSize, Infinity))`.  I'm not sure if anything bad will happen from having a negative `maxSize`, but it doesn't make sense so we may as well guard against it.

src/units.js
```diff
@@ -90,7 +90,12 @@ const calculateSize = function(sizeValue, options) {
             scale *= unitOptions.sizeMultiplier / options.sizeMultiplier;
         }
     }
-    return sizeValue.number * scale;
+    const scaledSize = sizeValue.number * scale;
+    if (options.maxSize && scaledSize > options.maxSize) {
+        return options.maxSize;
+    } else {
+        return scaledSize;
+    }
```
kevinbarabash: If `options.maxSize` is clamped between `0` and `Infinity`, this could become:
```
return scaleSize > options.maxSize ? options.maxSize : scaledSize;
```

kevinbarabash:
@xsznix thanks for the PR.  This features should come in really handy for people wanting to sanitize user input a bit.

kevinbarabash:
@xsznix friendly ping.

xsznix:
@kevinbarabash sorry for dropping off the radar, I've been traveling/busy for the past week. I'll get the requested changes in by this weekend at the latest.

kevinbarabash:
@xsznix thanks for updating the PR.  The tests look great.

PR #807, opened 2017-08-19
Use inline SVG for stretchy elements

ronkok:
Replace all background-images with inline SVG code.

Pros:

* `\color` works in all browsers, even IE/Edge
* Better printing
* Much simpler CSS
    * No links to background-images
    * No `mask`
    * No browser-detection
* No external SVG files
* Faster first rendering

Cons

* No image caching
* Heavier HTML load
* Larger JavaScript file
* ~~`\cancel` line is in `px` units, not `em` units~~


ronkok:
I've put up a [test page](http://copyasmarkup.com/katex/wideelements.html).

kevinbarabash:
The test page looks great!  I'll regenerate the screenshots later today.

kevinbarabash:
> `\cancel` line is in px units, not em units

The stroke-width can also be controlled by inline CSS, e.g. `<line style="stroke-width: 0.2em;"/>`.

src/stretchy.js
```diff
@@ -9,39 +9,39 @@ const mathMLTree = require("./mathMLTree");
 const utils = require("./utils");
 
 const stretchyCodePoint = {
-    widehat : "^",
-    widetilde : "~",
-    undertilde : "~",
-    overleftarrow : "\u2190",
-    underleftarrow : "\u2190",
-    xleftarrow : "\u2190",
+    widehat: "^",
```
kevinbarabash: I hadn't noticed the difference in `:` position before.  Thanks for aligned this with the style that's used elsewhere.

src/stretchy.js
```diff
@@ -52,130 +52,512 @@ const mathMLnode = function(label) {
...
+//    whose edges appear to be sharper.
+
+const innerSVG = {
+    // Since bcancel's SVG is inline and it omits the viewBox attribute,
+    // the browswer will render stroke-width='1.5px' in the screen context.
+    bcancel: `<line x1='0' y1='0' x2='100%' y2='100%' stroke-width='1.5px'
+stroke='currentcolor'/>`,
+
+    cancel: `<line x1='0' y1='100%' x2='100%' y2='0' stroke-width='1.5px'
+stroke='currentcolor'/>`,
```
kevinbarabash: Why does this require a `>` at the start like many of the innerSVG values?

src/stretchy.js
```diff
@@ -52,130 +52,512 @@ const mathMLnode = function(label) {
...
-    } else {
-        classArray.push(fileName);             // Set image and span height.
-        node = buildCommon.makeSpan(classArray, [], options);
+    const span = buildCommon.makeSpan([], [], options);
+    span.height = height;
+    span.depth = depth;
+    const totalHeight = height + depth;
+    span.style.height = totalHeight + "em";
+    if (minWidth > 0) {
+        span.style.minWidth = minWidth + "em";
```
kevinbarabash: Why is this necessary?

src/stretchy.js
```diff
@@ -52,130 +52,512 @@ const mathMLnode = function(label) {
...
+ 0-12.2.5-14.5 1.5-2.3 1-4.8 4.5-7.5 10.5-49.3 97.3-121.7 169.3-217 216-28
+ 14-57.3 25-88 33-6.7 2-11 3.8-13 5.5-2 1.7-3 4.2-3 7.5s1 5.8 3 7.5
+c2 1.7 6.3 3.5 13 5.5 68 17.3 128.2 47.8 180.5 91.5 52.3 43.7 93.8 96.2 124.5
+ 157.5 9.3 8 15.3 12.3 18 13h6c12-.7 18-4 18-10 0-2-1.7-7-5-15-23.3-46-52-87
+-86-123l-10-10h399738v-40H218c328 0 0 0 0 0l-10-8c-26.7-20-65.7-43-117-69 2.7
+-2 6-3.7 10-5 36.7-16 72.3-37.3 107-64l10-8h399782v-40z
+m8 0v40h399730v-40zm0 194v40h399730v-40z'/></svg>`,
+
+    // doubleleftrightarrow is from glyph U+21D4 in font KaTeX Main
+    doubleleftrightarrow: `><svg width='50.1%' viewBox='0 0 400000 549'
```
kevinbarabash: Why not `width='50%'`?

src/buildHTML.js
```diff
@@ -1420,16 +1420,18 @@ groupTypes.accent = function(group, options) {
...
-            accentBody.style.backgroundSize = adjSize;
-        }
-
         accentBody = buildCommon.makeVList([
             {type: "elem", elem: body},
             {type: "elem", elem: accentBody},
         ], "firstBaseline", null, options);
+
+        const styleSpan = accentBody.children[0].children[0].children[1];
+        styleSpan.classes.push("svg-align");  // text-align: left;
```
kevinbarabash: I couldn't find the definition for the `svg-align` class.  Is this still needed?

src/stretchy.js
ronkok: I originally wrote both `svgSpan` and `encloseSpan` so the the first closing `>` was in the template string, not in `innerSVG`.  But in `svgSpan`, I found that some of the SVGs were not nested, so I had to remove the `>` from of the template string and put it into `innerSVG`.

In encloseSpan, I left that closing `>` in the template string. I think it makes more sense that way.

src/stretchy.js
ronkok: If we try for an exact 50%-50% match, then floating point rounding error can sometimes create a 1 screen-pixel vertical white line in the middle of the arrow shaft. Setting one side to 50.1% is enough overlap to avoid this. Except one element, I think the braces, where Firefox needed something larger.

src/stretchy.js
ronkok: In cases where no text argument is given, I am trying to render an element that is the same length as the base glyph. That matches LaTeX behavior.

I did the same thing with the original stretchy elements, but the min-width was in the static CSS.

src/buildHTML.js
ronkok: Very much needed. If that `text-align: left` is not applied on just the right span, then the elements render smaller, lower, and off to the right.

ronkok:
I'll try experimenting with `<line style="stroke-width: 0.2em;"/>`. But I am trying to avoid a line that changes stroke-width with the area of the span. That means I have to omit the viewBox attribute. And so I'm not sure what `em` means in the screen context.

src/stretchy.js
ronkok: Exception: minWidth for the braces is set large enough to avoid a discontinuity where sloping parts of the brace match up.

src/buildHTML.js
kevinbarabash: Sorry... katex.less was collapsed and I didn't click to expand it.

ronkok: Yes, you have to open it to see the glorious sight of 575 lines of deleted code.

kevinbarabash:
I checked out the changes and ran `make build`.  The resulting minified JS + CSS is 203K vs 193K.  With gzipping the minified files, the difference is 49K vs 44K.

ronkok:
Well, I'm content with that the size change. It's a better product when the SVG is inline.

PR #810, opened 2017-08-20
Improve \sqrt

ronkok:
Make `\sqrt` out of inline SVGs to ensure a perfect match at the junction between surd and viniculum.


kevinbarabash:
sqrt

Green is LaTeX, red is KaTeX.  The spacing between the horizontal rules in the superscripts is improved over the last iteration of the inline SVG.

kevinbarabash:
Some of the radical symbols seem to be missing padding on the left that's built into the glyph.

before:
screen shot 2017-08-22 at 12 08 54 am

after:
screen shot 2017-08-22 at 12 09 05 am

Highlighting the outermost sqrt of the first superscript shows that there is no left padding in that sqrt.
screen shot 2017-08-22 at 12 06 18 am

The large sqrt symbols seem to be the ones that are affected as there is a gap between the first superscript and the second.

I looked at KaTeX_Size1-Regular and KaTeX_Size2-Regular fonts in opentype.js to confirm that there is indeed left padding built into the glyphs.

screen shot 2017-08-22 at 12 06 49 am

@ronkok can you update the inline SVGs to include the missing left padding?  Sorry for not catching this in in the original review.

ronkok:
I was hoping to avoid this. I didn't catch it before I minified the SVG code. Before that point, I have things somewhat automated, but afterwards a change is pretty tedious. No matter. I'll do it.

I have already added the correct leading to the `main` surd.  I'll go and do the other five.

kevinbarabash:
> Before that point, I have things somewhat automated, but afterwards a change is pretty tedious. No matter. I'll do it.

Do you have a script?  Would it be possible to add the padding via bounding box change?

ronkok:
I have scripts that do matrix transformations of the original geometry. I shift all of them so that the SVG has no negative coordinates, because IE 9 reportedly doesn't deal well with negative coordinates. 

But then there are non-scripted changes to make the viniculum absurdly long. And the tall surd got some custom changes of its own. 

Don't worry about it. I have found copies from an intermediate step in the process. That will make things pretty easy for sizes 1 thru 4.

ronkok:
@kevinbarabash It turned out to be pretty easy. Thank you for the nudge.

kevinbarabash:
sqrt

Left-padding on surds looks good now.

kevinbarabash:
screen shot 2017-08-22 at 9 13 24 pm

MathJax output for comparison.

src/delimiter.js
```diff
@@ -374,22 +374,12 @@ const sqrtSpan = function(height, delim, options) {
...
-        if (1 / newOptions.sizeMultiplier < refHeight) {
-            newOptions = options.havingBaseStyle(Style.SCRIPT);
-            if (1 / newOptions.sizeMultiplier < refHeight) {
-                newOptions = options.havingBaseStyle(Style.SCRIPTSCRIPT);
-            }
-        }
-        sizeMultiplier = newOptions.sizeMultiplier;
-
-        span.height = 1 / sizeMultiplier;
+        span.height = 1 * sizeMultiplier;
```
kevinbarabash: I'm glad you were able to simplify this.

ronkok:
Yes, it took a few tries to focus in on the target, but we got there eventually. Thanks for your help and your patience.

edemaine:
Nice work guys!

jpzwarte:
Very cool!

Issue #645, opened 2017-01-22
Copy/paste support


edemaine:
If you use select/copy/paste on a webpage containing KaTeX objects (in Chrome at least), you get the math expression twice (in some format), because of the MathML and HTML versions.  What is the "correct" behavior?

1. At the least, we can remove one copy (e.g. MathML) via [CSS user-select: none](https://css-tricks.com/almanac/properties/u/user-select/).  This seems natural given that MathML is already visually hidden.  This lets you select substrings of the formula... but it's still pretty hit-or-miss what you might get (sub/superscripts/lines/etc. disappear, lots of invisible space characters get copied).

2. I think it might be nice to make the LaTeX source code get copied, perhaps in combination with [CSS user-select: all](https://css-tricks.com/almanac/properties/u/user-select/) to make selection all-or-nothing.  (Selecting substrings wouldn't make a lot of sense in this context.)  I'm not sure whether the current `<annotation>` field can be made to be copyable, though, and I'm not sure it's worth a second copy of the LaTeX source.  (The `<annotation>` tag is also missing `$`s or `$$`s for display mode.  This seems correct for MathML, but isn't what I'd expect when copy/pasting...)

Anyway, I thought we could discuss desirable behavior before implementing anything.  In limited experiments, I haven't gotten either approach to work...

ronkok:
Yes, it would be cool if the LaTeX code could be copied. Even Microsoft Word [now accepts LaTeX input](https://blogs.msdn.microsoft.com/murrays/2017/07/30/latex-math-in-office/).

One way to accomplish this would be to make the entire KaTeX span into a toggle switch that would, when clicked, display the LaTeX code. One way to do this needs only some CSS and a unique ID by means of the [checkbox hack.](https://css-tricks.com/the-checkbox-hack/).

edemaine:
@ronkok In my application, there is a mix of text and math, with several equations, and I'd like to be able to copy/paste the entire thing.  Clicking each equation would be tedious, and also somewhat counterintuitive...

Your message prompted me to look up the HTML5 clipboard APIs. Amusingly, [Section 3.3](https://www.w3.org/TR/clipboard-apis/#Mathematic) gives our exact motivation:

[...]

PR #813, opened 2017-08-21
copy-tex contrib module

edemaine:
Solution to #645.  We wouldn't want this in base KaTeX, as it globally modifies copy/paste behavior, and a user might want to tweak it further.  But it's usable as-is for anyone not wanting to otherwise modify the copy/paste behavior, and a useful starting point in general.

From the README:

> This extension modifes the copy/paste behavior in any browser supporting the [Clipboard API](https://developer.mozilla.org/en-US/docs/Web/API/ClipboardEvent) so that, when selecting and copying whole KaTeX-rendered elements, the text content of the resulting clipboard renders KaTeX elements as their LaTeX source surrounded by specified delimiters.  (The HTML content remains unchanged.) The default delimiters are `$...$` for inline math and `$$...$$` for display math, but you can easy switch them to e.g. `\(...\)` and `\[...\]` by modifying `copyDelimiters` in [the source code](copy-tex.js).

[Demo on jsfiddle](https://jsfiddle.net/edemaine/hazzrd1d/) (slightly older version of the code)


kevinbarabash:
I've got stung a couple of times pushing changes with lint in it.  Maybe we could run lint as a pre-commit or pre-push hook.

kevinbarabash:
http://caniuse.com/#search=clipboard

screen shot 2017-08-21 at 2 33 36 pm

edemaine:
Now fully tested and working in Chrome (Windows and Android), Firefox (Windows), and Edge.  Can someone test Safari?

I found that Firefox had trouble with the modern ES I used (particularly [`replaceWith`](https://developer.mozilla.org/en-US/docs/Web/API/ChildNode/replaceWith), and in some older versions, `Array.forEach`), so I implemented some workarounds.

I found that Edge didn't like being given a DocumentFragment, so I converted it to HTML.  Also, given two versions of the copy/paste buffer, it just kept the latter one.  ([This is a bug in Edge](https://developer.microsoft.com/en-us/microsoft-edge/platform/status/clipboardapi/).)  So I now include the HTML before the plain text, so the latter gets used for Edge.  (This actually also makes the code more efficient.)

@kevinbarabash Agreed that pre-commit or pre-push hook would be nice!

xymostech:
Yay! Thank you for the cross-browser testing! I have a Safari that I have access to and can test if nobody else does.

+1 to the pre-commmit hooks. I'm always a little sad because I think that when you clone a repository, you can't make it also add some commit hooks by default, so you have to manually set them up. :(

edemaine:
Googling around, apparently [shared-git-hooks](https://github.com/kilianc/shared-git-hooks) can enforce Git hooks via `npm install`.

mbourne:
Regarding

> when selecting and copying whole KaTeX-rendered elements

My experience using Chrome/Windows of the fiddle mentioned earlier:

If I copy the display math portion only I get this:

> x^2+y^2=z^2 x 
> 2
>  +y 
> 2
>  =z 
> 2

If I include just a bit of text above and below the block, I get this:

> math:
>  $$ x^2+y^2=z^2 $$
> The

I suspect most users will copy just the math, and will not get the desired result.

kevinbarabash:
I ran into the same issue.  Maybe we can walk up the tree until we hit a `.katex` node so that it's a little more forgiving with the selection.

edemaine:
I thought it might be useful to allow people to copy/paste part of an expression (e.g. a `\text` chunk), and in that case we can't get the source... but perhaps there's a situation in which you select an entire math block but it doesn't give us the div.  I'll do some more experimentation.  We could use `user-select: all` to force all-or-nothing selection behavior, though that might also be annoying to the user...

edemaine:
OK, when selecting exactly a full formula, you don't get the containing `.katex` div (even though it's entirely selected!).  I've separated the handling of `.katex-mathml` and `.katex-html` so that we don't need the containing `.katex` div.  So now, if you fully select the display formula in Chrome, you'll get `$x^2+y^2=z^2 $`.  I can't detect the `$$` because I don't have the containing `.katex-display` either...  Also, if you select by clicking on the x and then dragging right, you don't end up selecting the MathML part at all, so you're back to selecting a subset of the formula.  Ouch, and this works only in Chrome, not Firefox.

At this point, I've gone as far as I can with this approach, `window.getSelection.getRangeAt(0).cloneContents()`.  There are two alternatives:

1. Use a `TreeWalker`.  See [this guide](https://superdevelopment.com/2016/02/08/working-with-the-html-selection-api/) which I've been following.  I'm not quite sure what behavior to target, though.
2. Define `user-select: all` on `.katex` and `.katex-display`.  Presumably just within this `contrib` package.  I've done some testing and this seems to work pretty well.  Just might annoy some users trying to select a piece of a big formula.

Probably option 2 is best, but let me know your thoughts.  Of course, a developer can always choose to use the JavaScript and not this CSS, and deal with the consequences.

kevinbarabash:
Safari's `select-all` behavior isn't very good.  It copies the right thing, but what it shows for its selection highlight is ridiculous.
screen shot 2017-08-26 at 10 16 04 pm

kevinbarabash:
screen shot 2017-08-26 at 10 22 40 pm

The `KaTeX` logo selection works fine.  My guess is probably some issue with vertical layouts.

contrib/copy-tex/README.md
```diff
@@ -0,0 +1,26 @@
+# Copy-tex extension
+
+This extension modifes the copy/paste behavior in any browser supporting the
+[Clipboard API](https://developer.mozilla.org/en-US/docs/Web/API/ClipboardEvent)
+so that, when selecting and copying whole KaTeX-rendered elements, the text
+content of the resulting clipboard renders KaTeX elements as their LaTeX source
+surrounded by specified delimiters.  (The HTML content remains unchanged.)
```
kevinbarabash: Maybe: "The rendered HTML content on the page remains unchanged."

contrib/copy-tex/copy-tex.js
```diff
@@ -0,0 +1,71 @@
+// Set these to how you want inline and display math to be delimited.
+const copyDelimiters = {
+    inline: ['$', '$'],    // alternative: ['\(', '\)']
+    display: ['$$', '$$'], // alternative: ['\[', '\]']
+};
+
+// Replace .katex elements with their TeX source (<annotation> element).
+// Modifies fragment in-place.  Useful for writing your own 'copy' handler.
+window.katexReplaceWithTex = function(fragment) {
```
kevinbarabash: `katexReplaceWithTex` doesn't need to go on `window`.  In fact this whole file could be wrapped in an IIFE to avoid polluting the global namespace.

Makefile
```diff
@@ -48,7 +49,7 @@ build/katex.css: static/katex.less $(wildcard static/*.less) $(NIS)
 	./node_modules/.bin/lessc $< $@
 
 build/katex.min.css: build/katex.css
-	./node_modules/.bin/cleancss -o $@ $<
+	$(CLEANCSS) -o $@ $<
```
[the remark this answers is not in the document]
edemaine: Exactly.  Just assigning it a variable so that it gets passed down (with necessary `..`s) into the subdirectory makes, like the other variables.

contrib/copy-tex/README.md
edemaine: Actually, I meant that the HTML content of the copy/paste buffer remains unchanged.  Will revise.

contrib/copy-tex/copy-tex.js
edemaine: The whole thing is already wrapped in an IIFE because it gets babelified into `build/contrib/copy-tex*.js`.  I was polluting the namespace in case someone wanted to build their own copy handler using this function.  In hindsight, that's not a very nice way to do it.

Instead I've split the function into its own module, `katex2tex.js`, which is the re-usable part.  (This is similar to how `auto-render` does it.)  Babelify merges these two files into one.  No more namespace pollution.

You still can directly change the delimiters only by manually editing the build.  (This is necessary if we don't want to expose any pollution.)  But now you could copy/paste the rather short `copy-tex.js` into your own build environment (and `require('katex/contrib/copy-tex/katex2tex.js')`).

edemaine:
New commits should deal with all requested changes, in particular in documentation.  Also dealt with a small conflict on `Makefile` by merging master in.

contrib/copy-tex/copy-tex.js
kevinbarabash: right... I forgot we started babelifying the contrib folder.

Issue #812, opened 2017-08-21
Support for Chinese punctuation marks in \text{}


zhzhzoo:
Hello everybody, I'm using KaTeX and discovered that it doesn't support Chinese punctuation marks in \text{}. For example, rendering `\text{，}` raises "KaTeX parse error: Expected '}', got '，' at position 7". In China we use '。' for period, '？' for question mark, '！' for exclamation mark, '，' for comma, '、' for dun hao, '；' for semicolon, '：' for colon, '‘’“”' for quotation marks (seems already supported), '（）［ ］〔〕【】' for parentheses, '——' for dash, '……' for ellipsis, '《》〈〉' to indicate book titles. ([General rules for punctuation](http://www.moe.edu.cn/ewebeditor/uploadfile/2012/06/01/20120601102833791.pdf)) Could you help supporting Chinese punctuation marks in \text{}? Thank you very much!

edemaine:
Thanks @zhzhzoo for the detailed request!  This looks relatively easy to do, mostly by changing some regexes.  PR #814 should do it.

PR #814, opened 2017-08-21
Support CJK full-width punctuation + Unicode dots

edemaine:
Fix for #812.  Screenshot:

cjk

I found two unicode groups that were missing for the symbols that @zhzhzoo listed in #812: 
* [CJK symbols and punctuation](https://en.wiktionary.org/wiki/Appendix:Unicode/CJK_Symbols_and_Punctuation): `[\u3000-\u303F]`
* [Fullwidth punctuation](https://en.wiktionary.org/wiki/Appendix:Unicode/Halfwidth_and_Fullwidth_Forms): `[\uFF00-\uFF60]`

I also reformatted the regex more minimally.  Happy to revert if that's not preferred, but I think the comment covers the intent pretty clearly. (Perhaps we should add the wiktionary links above?)

@zhzhzoo also listed "…" (U+2026) for ellipsis.  This was easy to add using @kevinbarabash's new Unicode character functionality.  While I was at it, I added support for all already-supported forms of ellipsis.


kevinbarabash:
The height for some of these characters is going to be wrong, but I think that's okay b/c there will almost always be non-punctuation CJK characters around them.

kevinbarabash:
I'm glad that this was easy to add.  Thanks @edemaine!

PR #815, opened 2017-08-21
Pre-commit hook to call make lint

edemaine:
As discussed in #813. A simple `npm install` will enable this.

Tested on Ubuntu Linux so far. (Do Git hooks work on Windows?)

### DIFF
--- a/hooks/pre-commit
+++ b/hooks/pre-commit
@@ -1,0 +1,2 @@
+#!/bin/sh
+exec make lint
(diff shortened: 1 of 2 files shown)


kevinbarabash:
> (Do Git hooks work on Windows?)

According to https://stackoverflow.com/questions/18277429/executing-git-hooks-on-windows:
```
By default, Git for Windows executes hook scripts using its own Windows port of the bash shell. 
```

kevinbarabash:
Thanks for adding this.  It should save some time next time I forget `;`.

PR #819, opened 2017-08-24
Reduce SVG

ronkok:
1. Edit to reduce redundancy in SVG code. Save 5 KB in size of stretchy.js.

2. Bug fix: Use correct path geometry for `\xleftharpoondown`.


ronkok:
Re: the bug fix:

I've put a copy of the proposed version of `\xleftharpoondown` at the top of the [test page](http://copyasmarkup.com/katex/wideelements.html). The existing version can be seen, alas, on the function support page on the right-hand side of the [extensible arrows](https://khan.github.io/KaTeX/function-support.html#extensible-arrows).

src/stretchy.js
```diff
@@ -137,280 +137,232 @@ const katexImagesData = {
...
     tilde4: ` viewBox='0 0 2340 312' preserveAspectRatio='none'>
-<path fill='currentcolor' d='M786 58C457 58 32 177.487 13 177.487c-6 0-10-3.345
+<path d='M786 58C457 58 32 177.487 13 177.487c-6 0-10-3.345
 -11-10.035L.15 143c-1-7 3-12 10-13l22-6.7C381.2 35 637.15 0 807.15 0c337 0 409
  177 744 177 328 0 754-127 773-127 5 0 10 3 11 9l1 14.794c1 7.805-3 13.38-9
  14.495l-20.7 5.574c-366.85 99.79-607.3 139.372-776.3 139.372-338 0-409
  -175.236-744-175.236z'/>`,
 
     // tofrom is from glyph U+21C4 in font KaTeX AMS Regular
     tofrom: `><svg width='50.1%' viewBox='0 0 400000 528'
```
kevinbarabash: There's a second `currentColor` is this SVG snippet that can also be removed.

src/stretchy.js
```diff
@@ -137,280 +137,232 @@ const katexImagesData = {
...
     // hookleftarrow is from glyph U+21A9 in font KaTeX Main
     hookleftarrow: `><svg width='50.1%' viewBox='0 0 400000 522'
-preserveAspectRatio='xMinYMin slice'><path fill='currentcolor' d='M400000 241
-H110l3-3c68.7-52.7 113.7-120 135-202 4-14.7 6-23 6-25 0-7.3-7-11-21-11-8 0
--13.2.8-15.5 2.5-2.3 1.7-4.2 5.8-5.5 12.5-1.3 4.7-2.7 10.3-4 17-12 48.7-34.8 92
--68.5 130S65.3 228.3 18 247c-10 4-16 7.7-18 11 0 8.7 6 14.3 18 17 47.3 18.7
- 87.8 47 121.5 85S196 441.3 208 490c.7 2 1.3 5 2 9s1.2 6.7 1.5 8c.3 1.3 1 3.3 2
- 6s2.2 4.5 3.5 5.5c1.3 1 3.3 1.8 6 2.5s6 1 10 1c14 0 21-3.7 21-11 0-2-2-10.3-6
--25-20-79.3-65-146.7-135-202l-3-3h399890zM100 241v40h399900v-40z'/></svg>
+preserveAspectRatio='xMinYMin slice'>${svgPath["leftarrow"]}</svg>
```
kevinbarabash: Even though this change doesn't save much after minification and gzip compression, it's good to de-duplicate strings so that if we have to modify the path there's only one place to modify it.

src/stretchy.js
kevinbarabash: @ronko I've removed the second `fill='currentcolor'` and updated your branch.

static/katex.less
```diff
@@ -547,6 +547,14 @@
...
         position: absolute; // absolute relative to parent
         width: 100%;
+
+        path {
+            fill: currentColor;
+        }
+
+        line {
+            stroke: currentColor;
+        }
```
kevinbarabash: nice!

kevinbarabash:
@ronkok thanks for the PR.

PR #827, opened 2017-08-29
Require \[text]color HTML colors to be well-formed.

### LINKED ISSUE #825: Color regex seems too permissive
I noticed in `Parser#parseColorGroup()`
https://github.com/Khan/KaTeX/blob/2fbb2b0128f062f6627581d250977a6ff9604964/src/Parser.js?utf8=%E2%9C%93#L776
that the color regex is defined as
`/^(#[a-z0-9]+|[a-z]+)$/i`
which seems too permissive for HTML hex colors. To test, I tried invalid colors like `#aar` and `#aaaa` on the main site's [test playground](https://khan.github.io/KaTeX/) and the parser accepted them, but didn't colorize the content in any way. Perhaps the regex should be changed to
`/^(#[a-f0-9]{3}|#[a-f0-9]{6}|[a-z]+)$/i`
to require hex digits (0-9a-f) and the correct number of them (3 or 6).

This change would be backward incompatible, of course, rejecting invalid colors that were silently accepted but ignored on conversion to MathML before.

marcianx:
Previously, the parser accepted all letters of the
alphabet for hex digits in HTML colors, and allowed any
number of digits. Now, only well-formed HTML hex colors
are allowed.

fixes #825

### DIFF
--- a/src/Parser.js
+++ b/src/Parser.js
@@ -773,7 +773,7 @@ class Parser {
         if (!res) {
             return null;
         }
-        const match = (/^(#[a-z0-9]+|[a-z]+)$/i).exec(res.text);
+        const match = (/^(#[a-f0-9]{3}|#[a-f0-9]{6}|[a-z]+)$/i).exec(res.text);
         if (!match) {
             throw new ParseError("Invalid color: '" + res.text + "'", res);
         }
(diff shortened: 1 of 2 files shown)


kevinbarabash:
I'm ignoring khanbot b/c it doesn't know how to verify people from organizations who have signed.

PR #830, opened 2017-08-29
Add better support for using webpack

### LINKED ISSUE #828: Using contrib with webpack - sources not included in NPM
Currently, the source files from the `contrib` directory are included in the published NPM package, meaning that I can't use webpack to import the auto-render plugin.

The way I understand it, it should be simple enough to just add `contrib` to the `files` array in `package.json`.

schmuli:
This makes it possible to consume the contrib extensions using webpack.

Closes #828

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,6 +11,7 @@
     "katex.js",
     "cli.js",
     "src/",
+    "contrib/",
     "dist/"
   ],
   "license": "MIT",


schmuli:
[clabot:check]

kevinbarabash:
@edemaine it'll be nice to be able to use modern JS in contrib code.

kevinbarabash:
@schmuli thanks for the PR.

PR #831, opened 2017-08-29
Update function support page for v0.8.3

ronkok:


function-support.html
```diff
@@ -321,6 +319,43 @@ <h2 id="accents">Accents</h2>
 			<td></td>
 		</tr>
 	</table>
+	
+	<h4 id="accents-inside-\text…}">Accents inside \text{…}</h4>
```
edemaine: This ID could maybe use a tweak -- either an `{` or maybe a simplification to avoid fancy Unicode characters...

function-support.html
```diff
@@ -321,6 +319,43 @@ <h2 id="accents">Accents</h2>
...
+			<td>\(\text{\H{a}}\)</td>
+			<td>\H{a}</td>
+		</tr>
+		<tr>
+			<td>\(\text{\`{a}}\)</td>
+			<td>\`{a}</td>
+			<td>\(\text{\={a}}\)</td>
+			<td>\={a}</td>
+			<td>\(\text{\&quot;{a}}\)</td>
+			<td>\&quot;{a}</td>
```
edemaine: I was surprised to see `&quot;` here.  `"` should be safe here, no?

ronkok: Software again. It's written to always replace quotation marks with either `&quot;` or `<q>` tags.

edemaine:
Hmm, looking at https://khan.github.io/KaTeX/function-support.html, it seems the new "accents inside \text{...}" section is missing `<code>` delimiters.  Another PR?

edemaine:
Fixed by #832

Issue #270, opened 2015-06-25
\llap and \rlap should render their content in text mode


gagern:
In LaTeX, `\llap`and `\rlap` render their content in text mode. I found this while creating LaTeX screenshots as part of #268. The image has KaTeX to the left, LaTeX to the right, and a diff of both below.

![lap3](https://cloud.githubusercontent.com/assets/190759/8351954/6bf921dc-1b31-11e5-926a-700c9dffc6a7.png)

The `RlapBug` test case also looks different in LaTeX: its numerator is centered. Is this a consequence of the mode change, in some way?

_(With a bit of luck, fixing this issue will also get rid of the weird non-determinism of the `Lap` example when creating screenshots.)_

spicyj:
Wow! Today I learned. I guess I now remember that `\mathllap` and friends are provided by `mathtools`: https://www.tug.org/TUGboat/tb22-4/tb72perlS.pdf

kevinbarabash:
This is fixed on master.  Still need to do a release with it.

Issue #164, opened 2014-11-03
Horizontal spacing commands: \>, \hfill


xymostech:
We have some spacing commands, but it appears we missed one. There might be more that we missed.

kevinb7:
We currently support: ```\!```, ```\ ```, ```\,```, ```\;```, ```\:```, ```~```, ```\space```, ```\enspace```, ```\quad```, ```\qquad```, and ```\phantom```.

We're missing: ```\thinspace```, ```\hspace```, ```\hfill```, ```\kern```, and ```\hphantom```.

Notes:
- ```\thinspace``` is an alias for ```\,```
- ```\hphantom``` is not an alias for ```\phantom```, according to TeXBook pg 178:

> There’s also \hphantom, which has the width of a \phantom, but its
height and depth are zero.

Reference: http://tex.stackexchange.com/questions/74353/what-commands-are-there-for-horizontal-spacing

gagern:
The way I see it, `\kern` is often followed by a dimension without braces around it. Would we want to support that as well? If so, we would have to adjust the [parsing of sizes in my macros branch](https://github.com/gagern/KaTeX/blob/b81fcb8765bbdb77c0514c566d651a8726dde15f/src/Parser.js#L609-L617), at least if we actually follow some variant of that branch to implement macros.

kevinb7:
I guess we'd have to.  Should probably have a ```lexSize``` function.

gagern:
Nah, we shouldn't do this in the lexer, but on the token stream. That's what TeX does, as far as I understand things.

kevinb7:
That could work to.  It seems like using a regex directly on the input stream would be easier as opposed to having to concatenate tokens would be trickier.  I know there's code there that does this already, but it relies on a hard stop which we wouldn't have in this situation.

gagern:
I think a length would be finished after its two letter unit has been parsed. But I'd have to read The TeX Book to verify this first.

kevinb7:
> I think a length would be finished after its two letter unit has been parsed.

For sure, it's a little trickier than checking for a ```"]"```.

gagern:
True, but it has the benefit of working even if there is some macro expansion involved in the length notation.

michaelcadilhac:
Any update on that?  It seems that `hspace` is a reasonable thing to have indeed.  Cheers.

kevinbarabash:
@michaelcadilhac we do have macros now, and we do support `\kern` now

[...]

Issue #50, opened 2014-09-16
mhchem extension support


bartolsthoorn:
I'm switching from _MathJax_ to _KaTeX_, but support for TeX extensions is currently missing. More specifically, I'd really like to use the `mhchem` extension: http://www.ctan.org/pkg/mhchem

For the record, this is how _MathJax_ supports it: http://docs.mathjax.org/en/latest/tex.html#mhchem

mhchem:
This is the author of the original mhchem LaTeX package. I will create a JavaScript parser for mhchem syntax in 2016 that probably also could be used for KaTeX, if you are interested.

xymostech:
@mhchem that sounds awesome! I think we would definitely be interested, if that gets done! Thanks!

brunovcosta:
I had to make an fast adaptation of mchem code to work with KaTeX:
https://github.com/brunovcosta/mchemJS

I can fork the katex in order to create the parser extension

mhchem:
@brunovcosta could you get in touch with me via e-mail? I am interested in what you changed. A lot of time and effort went into fine-tuning the spacing (and this might continue), so that I am not too excited to hear you got rid of all of these spacing details. I'd like to hear your reasons and learn from that (and maybe vice versa).

brunovcosta:
@mhchem, i can't find your email. Can you send me one? brunovcostadev@gmail.com

I made this adaptation because i needed mchem suport quickly. I understand the importance of the spacing details and i'm adapting my katex fork to support the required functions (\vphantom, \hphantom, \mathrel,\mkern and \mskip).

It's not intended to be an official pull request, neither an mchem replacement. Sorry about any mistake, i'll be happy to receive your email and help appropriately port mchem for katex

arjun810:
Is there still a chance that this'll get ported over to KaTeX? Would be really useful! We might be able to help out too, depending on how much knowledge of the underlying package is involved.

mhchem:
Yes, of course. mhchem's processing can be described as a 4-step process.
1. The LaTeX parser recognizes the `\ce` command and passes its argument on.
2. The mhchem parser parses the argument into an internal format
3. The mhchem texifier converts the internal format into a LaTeX expression
4. The LaTeX parser handles the output of step 3.

Step 1 can be done, proven by brunovcosta.

Step 2 is mhchem internal, no problem.

Step 3 depends on step 4. Currently, the output of step 3 cannot be handled by KaTeX on step 4.

[...]

PR #833, opened 2017-08-29
Add \smash, laps, spaces, and phantoms

ronkok:
1. Support `\smash`, including the optional argument from AMS.

2. Change `\llap` and `\rlap` so that they render in text style. Repeat: This *changes* KaTeX behavior.

3. Add `\mathllap` and `\mathrlap`. These will act as they do  in `mathtools` and as in previous KaTeX versions of `\llap` and `\rlap`.

4. Add `\mathclap` and `\clap`.

5. Add `\hphantom` and `\vphantom`.

6. Add `\thinspace`, `\medspace`, `\thickspace`

7. Add `\hspace`.

This work will resolve issue #270 and parts of #50 and #164.

A. Perlis has written a [concise description](https://www.math.lsu.edu/~aperlis/publications/mathclap/perlis_mathclap_24Jun2003.pdf) of items 1 thru 5. Except for `\smash`'s optional argument. It's described in the [AMS User's Guide](http://texdoc.net/texmf-dist/doc/latex/amsmath/amsldoc.pdf).

Item 6 also comes from the AMS User's Guide.


ronkok:
My apologies for the disconnectedness in this PR. And @edemaine thank you the help in getting restarted.

@kohler I haven't yet looked into your comment about calling `makeVList` from `groupTypes.smash`. I'll do that later this evening.

src/buildCommon.js
```diff
@@ -413,14 +413,26 @@ const spacingFunctions = {
         size: "0.277778em",
         className: "thickspace",
     },
+    "\\thickspace": {
+        size: "0.277778em",
+        className: "thickspace",
+    },
```
kevinbarabash: Are these separate commands or would it make sense to use `defineMacro` to define `\thickspace` in terms of `\;`?

static/katex.less
```diff
@@ -367,10 +368,16 @@
...
-    .rlap > .inner {
+    .rlap > .inner,
+    .clap > .inner {
         left: 0;
     }
 
+    .clap > .inner  > span {
+        margin-left: -50%;
+        margin-right: 50%;
+    }
```
kevinbarabash: I feel like there might be an easier way to center things.  I'll check out the branch and experiment.

test/katex-spec.js
```diff
@@ -715,7 +715,7 @@ describe("A text parser", function() {
     const textExpression = "\\text{a b}";
     const noBraceTextExpression = "\\text x";
     const nestedTextExpression =
-        "\\text{a {b} \\blue{c} \\textcolor{#fff}{x} \\llap{x}}";
+        "\\text{a {b} \\blue{c} \\textcolor{#fff}{x} \\mathllap{x}}";
```
kevinbarabash: Why change all `\lap`s in this file to `\mathlap`s?

test/screenshotter/ss_data.yaml
```diff
@@ -107,7 +107,7 @@ KaTeX: \KaTeX
 Kern:
     tex: \frac{a\kern{1em}b}{c}a\kern{1em}b\kern{1ex}c\kern{-0.25em}d
     nolatex: LaTeX fails to typeset this, “Missing number, treated as zero.”
-Lap: ab\llap{f}cd\rlap{g}h
+Lap: ab\mathllap{f}cd\mathrlap{g}h
```
kevinbarabash: We should probably also verify the changed behavior of `llap` et al.

src/functions.js
```diff
@@ -516,15 +517,42 @@ defineFunction([
...
+// Horizontal overlap functions
+defineFunction(["\\mathllap", "\\mathrlap", "\\mathclap"], {
     numArgs: 1,
     allowedInText: true,
 }, function(context, args) {
     const body = args[0];
     return {
-        type: context.funcName.slice(1),
+        type: "lap",
+        className: context.funcName.slice(5),
```
kevinbarabash: I'd like to avoid including CSS specifics at this level, maybe this could be called `alignment`.

src/buildMathML.js
```diff
@@ -645,6 +659,20 @@ groupTypes.phantom = function(group, options) {
...
+    const node = new mathMLTree.MathNode("mphantom", inner);
+    node.setAttribute("height", "0px");
+    return node;
+};
+
+groupTypes.vphantom = function(group, options) {
+    const inner = buildExpression(group.value.value, options);
+    const node = new mathMLTree.MathNode("mphantom", inner);
+    node.setAttribute("width", "0px");
+    return node;
```
kevinbarabash: Thanks for your attention to detail with this.

src/buildHTML.js
```diff
@@ -762,20 +762,68 @@ groupTypes.spacing = function(group, options) {
...
+
+    if (group.value.tb.length === 0) {
+        smashHeight = true;
+        smashDepth = true;
+    } else {
+        // Optional [tb] argument is engaged.
+        // ref: amsmath: \renewcommand{\smash}[1][tb]{%
+        //               def\mb@t{\ht}\def\mb@b{\dp}\def\mb@tb{\ht\z@\z@\dp}%
+        smashHeight = /t/.test(group.value.tb);
+        smashDepth = /b/.test(group.value.tb);
```
kevinbarabash: I see, the `tb` optional param is used to "smash" the top and/or the bottom of the group.

kevinbarabash: These tests could be moved into functions.js so that we don't have repeat the here and in buildMathML.js.

src/functions.js
```diff
@@ -516,15 +517,42 @@ defineFunction([
...
+        body: body,
+    };
+});
+
+// smash, with optional [bt], as in AMS
```
kevinbarabash: nit: the comment uses `bt` but everywhere else it's `tb`

src/functions.js
```diff
@@ -516,15 +517,42 @@ defineFunction([
...
+                tb += letter;
+            }
+        }
+    }
+
+    const body = args[1];
+    return {
+        type: "smash",
         body: body,
+        tb: tb,
```
kevinbarabash:
```
smashHeight: /t/.test(tbArg),
smashDepth: /b/.test(tbArg),
```

kevinbarabash: Do we care if there are other values in `tbArg`?

src/buildCommon.js
ronkok: Using `defineMacro` would work very well. I'll change it.

test/katex-spec.js
ronkok: It seemed to me that `\mathllap` is a core command. I found `\llap` to be less interesting. But your point is well taken. This line is supposed to be testing a *text* parser.  I'll change it back.

test/screenshotter/ss_data.yaml
ronkok: I'll add some tests.

src/functions.js
ronkok: Yes, I suppose we do care. I just checked and AMS LaTeX ignores the `\smash` if I put another character in there. Here's an example and the first exponent has `\smash[tab]{2}`


I'll change the code to act in the same way.

src/buildHTML.js
ronkok: I'll do that.

test/katex-spec.js
ronkok: Well, now I have a better answer to your question about the test. Since `\lap` is a test mode function, we should not expect it to work properly when given math mode arguments, such as `\left`. So the tests really should be revised to test `\mathllap` and not `\lllap`.

kevinbarabash:
@ronkok I've updated the `\llap`, `\clap`, and `\rlap` to use `textrm` instead of `mathrm`.  The unit tests are passing now.  I'll regenerate the `Lap` screenshots.

static/katex.less
kevinbarabash: I never got around to experimenting.  This does the trick and anything else would be marginally smaller so I'm going to merge this as is.

ronkok:
@kevinbarabash, It just occurred to me that there is another breaking change in release 0.9.0-alpha, one that is not yet listed in the release notes.  

The `\llap` and `\rlap` functions now render in text style, which follows LaTeX behavior. Also, KaTeX now has functions `\mathllap` and `\mathrlap`. These will act as they do  in `mathtools` and as in previous KaTeX versions of `\llap` and `\rlap`.

My [current draft](http://copyasmarkup.com/katex/function-support.html#overlap) of the function support page has been so updated. But I did not emphasize the change as much as I did when `\color` changed. If you want that draft revised, let me know.

kevinbarabash:
Good call.  I can up date the release notes.

kevinbarabash:
I've updated the [release notes](https://github.com/Khan/KaTeX/releases/tag/v0.9.0-alpha).

PR #834, opened 2017-08-30
Fix exponential behavior in accent production

sophiebits:
src/functions.js was returning two properties referring to the base; since buildHTML runs `JSON.parse(JSON.stringify(tree))` to get an immutable copy, that meant we'd traverse and serialize and parse an exponentially-sized tree.

Test Plan: `\breve{\breve{\breve{\breve{\breve{\breve{\breve{\breve{\breve{\breve{\breve{\breve{\breve{\breve{\breve{\breve{\breve{\breve{\breve{\breve{\breve{\breve{\breve{\breve{\breve{\breve{\breve{\breve{\breve{\breve{\breve{\breve{\breve{\breve{A}}}}}}}}}}}}}}}}}}}}}}}}}}}}}}}}}}` renders instantly; previously, it reliably timed out with even half that depth.

### DIFF
--- a/src/stretchy.js
+++ b/src/stretchy.js
@@ -442,6 +442,14 @@ c-1 5-5 9-11 9h-2L532 67 19 159h-2c-5 0-9-4-11-9l-5-22c-1-6 2-12 8-13z'/>`,
 <line x1='0' y1='100%' x2='100%' y2='0' stroke-width='0.046em'/>`,
 };
 
+const groupLength = function(arg) {
+    if (arg.type === "ordgroup") {
+        return arg.value.length;
+    } else {
+        return 1;
+    }
+};
+
 const svgSpan = function(group, options) {
     // Create a span with inline SVG for the element.
     const label = group.value.label.substr(1);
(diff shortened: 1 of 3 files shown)


ronkok:
I like it.

src/stretchy.js
```diff
@@ -453,7 +461,7 @@ const svgSpan = function(group, options) {
     if (utils.contains(["widehat", "widetilde", "undertilde"], label)) {
         // There are four SVG images available for each function.
         // Choose a taller image when there are more characters.
-        const numChars = group.value.value.length;
+        const numChars = groupLength(group.value.base);
```
kevinbarabash: Stretchy accents use `body` instead of base.  Probably best to make them use the same thing.  I have a slight preference for `body` because `base` makes me think of exponents, but either would be fine.

sophiebits:
Opted for `base` since `groupTypes.supsub` has:

```
if (shouldHandleSupSub(group, options)) {
    return groupTypes[group.value.base.type](group, options);
}
```

which currently relies on op, accent, horizBrace all having `.base`. Should be good now though. Screenshots pass this time.

kevinbarabash:
@sophiebits thanks for finding this issue and fixing it.

Issue #809, opened 2017-08-19
support for using KaTeX in a wysiwyg math editor


kevinbarabash:
There are a couple of approaches to implementing an editor:
- edit the input string directly
- edit the parse tree

Both have their challenges.  In the case of editing the LaTeX string, determining which characters from the original source should be updated when the user makes an edit can be tricky especially when macros are involved.  Editing the parse tree is complicated by the fact that we haven't stabilized the parse tree yet.

There are various things that an editor needs to be able to do:
- move a cursor around via the keyboard and mouse
- render a cursor at the correct location
- make a selection of multiple characters and render it
- insert and delete nodes

All of this seems incredible difficult to do if we try to map back to the original input string.  If we go with the parse tree approach we'd want to be able to navigate through the tree.  The parse tree doesn't support anyway to navigate to sibling nodes or to navigate to a parent node, but it could be made to do so.

I think the biggest challenge will be generating new nodes in the format expected by `buildTree`.

In order to facilitate using the mouse to position the cursor and make selections we'll want someway to link nodes in the DOM with nodes in the parse tree.  We could add an `id` prop to each node in the parseTree and matching `katex-id` properties to corresponding DOM nodes.

@spontaliku-softaria has been experimenting with adding `katex-id`s in https://github.com/khan/KaTeX/compare/master...spontaliku-softaria:math-input.

kevinbarabash:
Even if the editor is modifying parse tree nodes, we'd probably eventually want to serialize those nodes so being able to convert back to LaTeX at some point would is important.

spontaliku-softaria:
I believe editing tree is better approach because it would allow to express editing as tree transformations. They could be done in various sane and expressive ways. In general, I think having good tree would be better for most external usages of Katex like #800 or #554. 

So, this task looks to me like a successor of #376. I'd like to make a proposal of changes to achieve #376 and then for this task.

edemaine:
Incidentally, once complete, this would likely produce a bidirectional map between editing/rendered subtrees and portions of the LaTeX text, which would be handy for the copy-tex module (#813): we could reasonably copy/paste portions of the equation and get the resulting subset of the source code.  A pretty minor effect, but it's currently on my mind. 😃

spontaliku-softaria:
I see there is existing support for arbitrary attributes in constructed HTML and MathML nodes. So, naive idea how to implement this issue:
[...]

Issue #800, opened 2017-08-17
update buildHTML to output HAST


sandhose:
Hi,

We're doing a [markdown engine](https://github.com/zestedesavoir/zmarkdown) using [`remark`](https://github.com/wooorm/remark/) with the [`rehype-katex`](https://github.com/Rokt33r/remark-math/tree/master/packages/rehype-katex) plugin.
The thing is, `rehype-katex` is using KaTeX's [`renderToString`](https://github.com/Rokt33r/remark-math/blob/master/packages/rehype-katex/index.js#L35) function, and [parses the output](https://github.com/Rokt33r/remark-math/blob/master/packages/rehype-katex/index.js#L58) using [`rehype-parse`](https://github.com/wooorm/rehype/tree/master/packages/rehype-parse) (which uses `parse5` to parse the HTML string into [HAST](https://github.com/syntax-tree/hast), `rehype`'s AST).

This is really inefficient (haven't done benchmarks, but it adds ~250Kb to the bundle), and could be improved by allowing to render to a [hyperscript](https://github.com/hyperhype/hyperscript) like function. By doing this, it would allow to render far more efficiently KaTeX into frameworks like [React](https://github.com/mlmorg/react-hyperscript) or [Vue.js](https://vuejs.org/v2/guide/render-function.html).

I'll write a proof of concept today ; I might not have the time to write the doc nor tests though.

I know there's a lot of links up here because of our use case, but **TL;DR**: it would be nice to be able to render KaTeX directly through a hyperscript like function.

kevinbarabash:
@sandhose in https://github.com/Khan/react-components/blob/master/js/tex.jsx we use `dangerouslySetInnerHTML` to side step this issue in `React`.  I'm not sure if Vue.js has a similar mechanism or not.

sandhose:
I'm not using Vue.js in this case, but the issue is quite the same. You could avoid using `dangerouslySetInnerHTML` by using `React.createElement` (with a small wrapper) as an hyperscript function to the renderer. By doing this, you're creating real React elements, and you take advantage of React's virtual dom.

In my case, the output of the markdown engine is in one stage represented by a `HAST` node (`HAST` is a HTML AST). 
The way the plugin that parses the math elements in the markdown works, is that it passes the raw math input to KaTeX through the `renderToString` function, and then uses `rehype-parse` to parse the HTML string to a HAST node. 
This parse is heavy, and the parser itself adds ~250Kb to the bundle.

[...]

PR #835, opened 2017-08-31
Propagate custom attributes from parse tree to DOM nodes

spontaliku-softaria:
Add an ability to propagate attributes from parsed tree nodes into DOM nodes. Essential for #809.

Current concerns:
* I've changed `color`, `mod`, `sizing`, `styling` and `phantom` to be spans instead of fragments to have single root node where I could write attributes. Is it acceptable change? All I could see what could be broken is some CSS rules with `>` in Katex's users code. But I guess relying on Katex's styles and DOM structure in your site should not be supported case anyway.
* Test is too simple. I'll think about something better. I'd like some advice about it if any.
* I can do MathML later when we evaluate proper approach to this change.


spontaliku-softaria:
[clabot:check]

spontaliku-softaria:
Oops, forgot to push one commit. Should be ok now.

test/katex-spec.js
```diff
@@ -2330,3 +2336,40 @@ describe("The maxSize setting", function() {
...
+            "katex-frac-id": "frac",
+        };
+
+        tree[0].value.numer.attributes = {
+            "katex-numer-id": "numer",
+        };
+
+        tree[0].value.denom.attributes = {
+            "katex-denom-id": "denom",
+        };
```
kevinbarabash: You're going to want to add `id`s to each node automatically eventually though.  I wonder if provide some sort of callback that is passed each parse tree node and its associated HTML tree node might be better.  The callback would be called whenever a method on groupTypes in buildHTML.js returns.

spontaliku-softaria: I moved attributes logic into post-processing funciton. Currently it is set to Options.postProcessor, which makes Options non-serializable. Should it be serialzable? If yes, I can move post-processing into some wrapper around Options. Or into signatures of every method of buildHTML.js, but I guess it would be a bit dirty.

Simple numeric IDs are currently assigned inside test. I can make an option to set them automatically.

spontaliku-softaria:
About improving test: last time I was doing something with trees I had nice experience with QuickCheck-style tests. I used [quickcheckjs](https://quickcheckjs.readme.io/) and I was able to detect huge amount of bugs in my tree transformations. I could add some infrastructure for generating Katex tree nodes suitable for tests. Then I'll add attributes testing for all types of nodes simultaneously.

spontaliku-softaria:
Still now sure why tests fail. I was able to run them locally and they fail locally too, but I shouldn't change actual rendering.

src/buildHTML.js
```diff
@@ -1228,7 +1230,7 @@ function sizingGroup(value, options, baseOptions) {
         inner[i].depth *= multiplier;
     }
 
-    return buildCommon.makeFragment(inner);
+    return buildCommon.makeSpan(null, inner, null, attributes);
```
kevinbarabash: This will change the spacing between atoms at sizing group boundaries.  This is because the spacing styles are based on sibling relationships.  If the start of a sizing group is one type of atom and the end is another type of atom we can't apply the existing classes and have the spacing on both ends match.  I think it makes sense to give sizing groups an element in the HTML.  There might be nodes within the LaTeX parse tree which don't have nodes in the HTML.  I think this is okay, but you'll need ways to get this information, e.g. functions to get previous/next sizing command and possible build selections of other nodes based on the location of sizing commands.

kevinbarabash: All changes of `makeFragment` to `makeSpan` will likely have similar results on the spacing of atoms.

test/katex-spec.js
```diff
@@ -2378,3 +2384,54 @@ describe("The maxSize setting", function() {
...
+
+        let counter = 1;
+        const putId = function(treeNode) {
+            if (!treeNode.attributes) {
+                treeNode.attributes = {};
```
kevinbarabash: I would probably do:
```
let counter = 1;
const putId = (treeNode, htmlNode) => {
    const id = count++;
    treeNode.id = id;
    if (htmlNode instanceof BaseNode) {
         htmlNode.properties['katex-id'] = id;
    }
}
```
This assumes that htmlNode is a HAST node, see https://github.com/syntax-tree/hast and https://github.com/Khan/KaTeX/issues/800.  This allows us to avoid having to create `defaultPostProcessor` in buildHTML.js which splits up the logic of what we're trying to do across multiple files.

src/domTree.js
```diff
@@ -25,20 +25,73 @@ const createClass = function(classes) {
...
+class baseNode {
+    constructor() {
+        this.attributes = {};
+    }
+
+    attributesToNode(node) {
+        for (const attr in this.attributes) {
+            if (Object.prototype.hasOwnProperty.call(this.attributes, attr)) {
+                node.setAttribute(attr, this.attributes[attr]);
+            }
```
kevinbarabash: This is really confusing.  Too many things are node here.  In this case the node being passed in a DOM node, right?

kevinbarabash: I think ideally, if we using something like HAST to build a representation of the HTML we want using JS objects, we'd probably want to use the visitor pattern to build the actually DOM nodes.

kevinbarabash:
@spontaliku-softaria there are few different things going on in this diff.  I'd like to try to split things up a bit.  In particularly I'd to try and tackle #800 (update buildHTML to output HAST) first.  After that's in place you we should only need to modify `buildGroup` as you've done in this PR.

src/domTree.js
spontaliku-softaria: Not sure I got it about node. I've made it just to put common code here. Maybe the name `baseNode` is misleading?

But I got it about HAST. I guess you would want visitor now because I already see svgNode, pathNode and lineNode.

src/buildHTML.js
spontaliku-softaria: I think sizing and color could be treated somehow sane in an editor, but `mod` would be bigger problem because it changes structure and semantics. I've reverted chages for former fragments. Maybe I would be able to say more about it later when we have an actual editor.

src/domTree.js
kevinbarabash: Looking at the method `attributesToNode(node)` on the class `baseNode` I assumed that the `node` being passed in was an instance of `baseNode`.  Renaming the parameter to `domNode` would've helped.  Also, `attributesToNode` should probably be a helper function.

spontaliku-softaria: Oh, I get it now. I'll move it to helper function and get rid of baseNode since this classes are going to change I guess.

kevinbarabash: @spontaliku-softaria you may want to hold off on revisions to this PR.  I think we should do #800 first.

spontaliku-softaria: OK, I'll stop spamming you with comments.

kevinbarabash: The comments are fine.  I just don't think I'll be able to merge this right away and the changes proposed in #800 will require significant rework of this PR anyways.

spontaliku-softaria: That's OK to me since I'm not fully in context of your plans. It's just we are experimenting with editor and I need to do all this anyway, here or in a private fork.

PR #837, opened 2017-09-02
Update Parser comments to match implementation.

marcianx:
I'd been a little confused by some parts of Parser.js, I'd updated some comments for clarity, which I've included in this PR.

I also can't help but wondering whether using a transpiler that supports more compile-time checks (e.g. Google's closure compiler or Microsoft's Typescript) might be worth using to catch some of these mismatches between `@param`/`@return` definitions and actual implementation.

### DIFF
--- a/src/Parser.js
+++ b/src/Parser.js
@@ -43,15 +43,23 @@ import ParseError from "./ParseError";
  * standalone object which can be used as an argument to another function.
  */
 
-/**
- * An initial function (without its arguments), or an argument to a function.
- * The `result` argument should be a ParseNode.
- */
-function ParseFuncOrArgument(result, isFunction, token) {
-    this.result = result;
-    // Is this a function (i.e. is it something defined in functions.js)?
-    this.isFunction = isFunction;
-    this.token = token;
+/** A function name or an argument to a function. */
+class ParseFuncOrArgument {
+    /**
+     * @param {ParseNode|string} result It's a string if `isFunction=true` or if
+     *     `result="$"` for switching into math mode; otherwise, it's a ParseNode
+     *     and `isFunction` must be false. If it's a function, the string should
+     *     be a name defined by defineFunction, e.g. "\\frac".
+     * @param {boolean} isFunction True when this is a function. False when it's a
+     *     function argument or "$" to switch into math mode.
+     * @param {Token} token
+     */
+    constructor(result, isFunction, token) {
+        this.result = result;
+        // Is this a function (i.e. is it something defined in functions.js)?
+        this.isFunction = isFunction;
+        this.token = token;
+    }
 }
 
 class Parser {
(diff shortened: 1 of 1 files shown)


kevinbarabash:
@marcianx Parser.js definite needs some cleanup.  Thanks for improving the comments and changing that call to `handleUnsupportedCmd`.  I've started to use `flow` for type checking in https://github.com/Khan/KaTeX/commit/d8116bdc64a8be77a7153a57965c5e6e40b24fad.  I'd eventually like most/all of our code be statically typed.

marcianx:
Nice! When is that landing? I'm happy to help convert things, though I haven't yet read past Parser and its dependencies so far.

kevinbarabash:
`flow` support landed on `master` earlier today.  Adding type additions, even if it's just within Parser.js, would be greatly appreciated.

src/Parser.js
```diff
@@ -45,7 +45,13 @@ import ParseError from "./ParseError";
...
- * The `result` argument should be a ParseNode.
+ *
+ * @param {ParseNode|string} result It's a string if `isFunction=true` or if
+ *     `result="$"` for switching into math mode; otherwise, it's a ParseNode
+ *     and `isFunction` must be false.
+ * @param {boolean} isFunction True when this is a function. False when it's a
+ *     function argument or "$" to switch into math mode.
+ * @param {Token} token
  */
 function ParseFuncOrArgument(result, isFunction, token) {
```
kevinbarabash: This should really be specified as a class since that's how it's being used.  The fact that `result` can sometimes be a `string` and sometimes a `ParseNode` makes things hard to reason about.  No need to do anything about it in this PR though.

src/Parser.js
```diff
@@ -45,7 +45,13 @@ import ParseError from "./ParseError";
 
 /**
  * An initial function (without its arguments), or an argument to a function.
- * The `result` argument should be a ParseNode.
+ *
+ * @param {ParseNode|string} result It's a string if `isFunction=true` or if
+ *     `result="$"` for switching into math mode; otherwise, it's a ParseNode
+ *     and `isFunction` must be false.
```
kevinbarabash: It's probably worth noting that when it is a string, it should be a function name defined by `defineFunction`, e.g. `"\\frac"`.

src/Parser.js
marcianx: I completely agree about the point of confusion. Once we have a static type system in place, I'm rather inclined to do a little refactor. But I'd open an issue on that first to open up discussion.

src/Parser.js
```diff
@@ -140,7 +146,7 @@ class Parser {
      *                  should end with, or `null` if something else should end the
      *                  expression.
      *
-     * @return {ParseNode}
+     * @return {?Array<ParseNode>}
```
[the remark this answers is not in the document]
marcianx: Good point!

src/Parser.js
marcianx: Done.

PR #839, opened 2017-09-03
To @flow: Token, Lexer, ParseError, and ParseNode.

marcianx:
Here's a start of converting the parsing pipeline to `@flow`.
I have a concern: I'd screwed up the member definition for `ParseNode` -- specifically, I'd declared `firstToken` and `lastToken` properties instead of declaring `start`, `end`, and `lexer` which are actually assigned in the constructor. And `make test` didn't notice -- it just compiled successfully in contradiction to the documentation here: https://flow.org/en/docs/types/classes/#toc-class-fields-properties.

I know the `@flow` pipeline is working since it complains about missing or unused types (which is why I had to split off `Token` into a separate file and introduce `LexerInterface`).
I don't see anything fundamentally wrong with the babel set up per the documentation. (This is my first time using babel+flow).
Are you by any chance able to reproduce this?

### DIFF
--- a/src/ParseError.js
+++ b/src/ParseError.js
@@ -1,3 +1,7 @@
+// @flow
+import ParseNode from "./ParseNode";
+import {Token} from "./Token";
+
 /**
  * This is the ParseError class, which is the main error thrown by KaTeX
  * functions when something has gone wrong. This is used to distinguish internal
@@ -5,17 +9,20 @@
  *
  * If possible, a caller should provide a Token or ParseNode with information
  * about where in the source string the problem occurred.
- *
- * @param {string} message  The error message
- * @param {(Token|ParseNode)=} token  An object providing position information
  */
-class ParseError {
-    constructor(message, token) {
+class ParseError extends Error {
+    position: number|void; // Error position based on passed-in Token or ParseNode.
+
+    constructor(
+        message: string,         // The error message
+        token?: Token|ParseNode, // An object providing position information
+    ) {
         let error = "KaTeX parse error: " + message;
         let start;
-        let end;
 
-        if (token && token.lexer && token.start <= token.end) {
+        if (token && token.lexer &&
+            token.start != null && token.end != null &&
+            token.start <= token.end) {
             // If we have the input and a position, make the error a bit fancier
 
             // Get the input
@@ -23,7 +30,7 @@ class ParseError {
 
             // Prepend some information
             start = token.start;
-            end = token.end;
+            const end = token.end;
             if (start === input.length) {
                 error += " at end of input: ";
             } else {
@@ -47,20 +54,12 @@ class ParseError {
                 right = input.slice(end);
             }
             error += left + underlined + right;
+
         }
 
-        // Some hackery to make ParseError a prototype of Error
-        // See http://stackoverflow.com/a/8460753
-        const self = new Error(error);
-        self.name = "ParseError";
-        self.__proto__ = ParseError.prototype;
-
-        self.position = start;
-        return self;
+        super(error);
+        this.position = start;
     }
 }
-
-// More hackery
-ParseError.prototype.__proto__ = Error.prototype;
 
 module.exports = ParseError;
--- a/src/ParseError.js
+++ b/src/ParseError.js
@@ -10,7 +10,7 @@ import {Token} from "./Token";
  * If possible, a caller should provide a Token or ParseNode with information
  * about where in the source string the problem occurred.
  */
-class ParseError extends Error {
+class ParseError {
     position: number|void; // Error position based on passed-in Token or ParseNode.
 
     constructor(
@@ -57,9 +57,19 @@ class ParseError extends Error {
 
         }
 
-        super(error);
-        this.position = start;
+        // Some hackery to make ParseError a prototype of Error
+        // See http://stackoverflow.com/a/8460753
+        const self = new Error(error);
+        self.name = "ParseError";
+        // $FlowFixMe
+        self.__proto__ = ParseError.prototype;
+        // $FlowFixMe
+        self.position = start;
+        return self;
     }
 }
+
+// $FlowFixMe More hackery
+ParseError.prototype.__proto__ = Error.prototype;
 
 module.exports = ParseError;
--- a/.babelrc
+++ b/.babelrc
@@ -2,7 +2,6 @@
     "presets": ["es2015", "flow"],
     "plugins": [
         "transform-runtime",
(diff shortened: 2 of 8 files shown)


marcianx:
Ah, per the travis failures, `npm test` does the trick (which runs `make lint flow test`). Cool, I'll fix the existing failures. Not sure how to deal with the `match-at` failure yet, though, and I'd love any insights on that.

marcianx:
I've updated and squashed fixes for most of these. Please note that I had to include [babel-plugin-transform-builtin-extend](https://github.com/loganfsmyth/babel-plugin-transform-builtin-extend) to allow `ParseError` to extend `Error` without annoying @flow.

I still have to figure out the inclusion of "match-at":

    Error: src/Lexer.js:15
     15: import matchAt from "match-at";
                             ^^^^^^^^^^ match-at. Required module not found

Simple workaround is to remove `@flow` from `Lexer.js` and removing the flow types gets rid of this error. "match-at" already uses `@flow`, so I was hoping it would output the right things (`.flow` file per https://javascriptplayground.com/blog/2017/01/npm-flowjs-javascript/) for this to work.

Still investigating, but, again, if you have any insights, I'd love to hear.

kevinbarabash:
@marcianx unfortunately `babel-plugin-transform-builtin-extend` doesn't work with IE10.  I tried the `approximate` option and that didn't work either.  I created https://github.com/marcianx/KaTeX/pull/2 which fixes the flow errors.  The errors in `ParseError.js` required `$FlowFixMe` unfortunately.  Hopefully that's the only places where we need to do that.

marcianx:
Ok, so the `.flowconfig` ignores `match-at` in `node_modules`. Un-ignoring it results in errors in `match-at` because the transpiled js still includes `@flow` even though the types have been stripped. Manually removing the `@flow` makes it compile, which is equivalent to adding
     "prepublish": "babel -d lib/ src/; sed -i 1d lib/matchAt.js",
in `match-at`'s `package.json`. I'm not sure they'd want to take on a `gnu sed` dependency on their build.

Sadness...

`match-at` also seems to be using a very old version of "babel" and things, and trying to upgrade meant running into a bunch of new things that I'm not yet familiar with in the realm of `babel`. So I decided to not go down that rabbit hole, as that also requires getting `match-at` to publish a new version, etc.

It might be worth making our own Regex matching helper. I've done something similar before when writing a parser in js.

marcianx:
Also, thanks for the `$FlowFixMe` fix. That definitely works. :)

marcianx:
(BTW, I was also editing in parallel, but it seems that our changes would overlap. So we can just discuss on your branch to avoid confusion.)

src/ParseError.js
```diff
@@ -1,29 +1,36 @@
...
  *
  * If possible, a caller should provide a Token or ParseNode with information
  * about where in the source string the problem occurred.
- *
- * @param {string} message  The error message
- * @param {(Token|ParseNode)=} token  An object providing position information
  */
-class ParseError {
-    constructor(message, token) {
+class ParseError extends Error {
```
kevinbarabash: This doesn't work for IE10.  Please revert.

src/ParseError.js
```diff
@@ -47,20 +54,12 @@ class ParseError {
                 right = input.slice(end);
             }
             error += left + underlined + right;
-        }
 
-        // Some hackery to make ParseError a prototype of Error
-        // See http://stackoverflow.com/a/8460753
-        const self = new Error(error);
-        self.name = "ParseError";
-        self.__proto__ = ParseError.prototype;
```
kevinbarabash: `$FlowFixMe`

src/ParseError.js
```diff
@@ -47,20 +54,12 @@ class ParseError {
...
 
-        self.position = start;
-        return self;
+        super(error);
+        this.position = start;
     }
 }
 
-// More hackery
-ParseError.prototype.__proto__ = Error.prototype;
```
kevinbarabash: `$FlowFixMe`

marcianx:
@kevinbarabash: A few questions before I update the PR with the suggested fixes, as I'm new to collaboration on github with PRs and also to KaTeX's conventions.

1) Whenever I've been updating the `flow` branch that I've been smashing the commits together into a single commit as that's how you'd want to commit them. However, does this hurt incremental review and your ability to fork? It also presumably also hurts your ability to add your edits on the PR? More specifically, what is your preference: should I push additional commits with fixes on the `flow` branch or just smash the commits together?

2) I noticed in your edits on the PR that you'd changed some `(Type|void)` to `?Type`. I did the former since it was less work (optional parameters were already `void`-able). Do you have a preference for one or the other?

3) IMHO, regarding the use of `NaNs`: I think it would be extremely useful for the type system to prevent us from using optional numbers without first checking them hence preferring the use of `(number|void)`. I agree that it's inconvenient in the case or `start`, `end`, `lexer`, all of which would be `void/null`-able but to accommodate we should (separately) refactor to use `?{start: number, end: number, lexer: LexerInterface}` as the type so that there's only a single `null` check. WDYT?

marcianx:
(I pushed the fixes as a separate commit for now.)

kevinbarabash:
1. I have a slight preference for reviewing separate commits, although I'm okay with squashed commits too.  Sometimes people have split their commits into logically separate changes and so revisions to those usually end up rewriting history which is cool too.

2. Your reason for going with `Type|void` makes sense to me.  Let's go with it.

3. I'd probably combine `start` and `end` in a `range` prop, but am open to further discussion on the matter.

kevinbarabash:
@marcianx thanks for the PR.  I'm excited to have more static typing in our codebase.

marcianx:
@kevinbarabash I agree, we should avoid `*` unless it's thoroughly obvious. I've only used it where the property matches the well-documented constructor parameters which appear immediately after it. Otherwise, I'm very inclined to avoid it. I'm also fine with banning `*`. I too am a fan of explicitness.

kevinbarabash:
If we do ban `*` then we should probably lint against it.

marcianx:
@kevinbarabash: How do we avoid stepping over others' toes while porting to `@flow`? `Parser.js` has a *lot* of dependencies to port and there's a danger that two people could be doing this independently. I'd like to keep plowing through.

kevinbarabash:
@marcianx probably the best way is to do it in small chunks if possible and open issues for each part that you're working on that way there's better visibility.

edemaine:
FWIW, I'm adding types to `defineMacro` in `MacroExpander.js` now.

PR #841, opened 2017-09-04
Improve SVG Performance

ronkok:
This PR introduces three new classes:  `svgNode`, `pathNode`, and `lineNode`. SVG data is then loaded into the domTree exclusively via instances of these classes.

`innerHTML` is banished.


ronkok:
@kevinbarabash I believe that this is one step in a direction where you wan to go. Let me know if I have mis-interpreted things.

ronkok:
I've put up a [test page](http://copyasmarkup.com/katex/wideelements.html).

kevinbarabash:
This will help big time with https://github.com/Khan/KaTeX/issues/800.  We haven't done much perf work, but perf is one KaTeX's main features so this is very much appreciated.

kevinbarabash:
The height on some of the `<svg>` elements on the test page is `1800em` which causes the page to be super long.  This is probably what's causing failures with the screenshot tests.

ronkok:
Oops. Multiplied the wrong variable by 1000. I hate it when that happens.

kevinbarabash:
@ronkok my assumption is that these changes shouldn't change the output.  I'll run the screenshot tests tomorrow and hopefully that'll shed some light on the failures.

ronkok:
When I do this sort of thing, every line of SVG code passes through my hands, and I can't help but tweak things as they go by. In this PR, I changed the `min-length` on many arrows and I rewrote the paths for `\xlinesegment`, `\xequals`, and the left hand side of `\xmapsto`.

I have no idea why `\sqrt` would fail its screenshots. I calculate the same things, but the calculations are done in a different order. Perhaps floating point rounding error?

ronkok:
@kevinbarabash Thank you once again for the help with Screenshotter. I know that you're busy just now with flow reviews and I appreciate the time you take to help here.

src/domTree.js
```diff
@@ -344,8 +336,164 @@ class symbolNode {
     }
 }
 
+/**
+ * SVG nodes are used to render stretchy wide elements.
+ */
+class svgNode {
+    constructor(children, width, height, x, viewBoxWidth, viewBoxHeight, align) {
```
kevinbarabash: I think combining all of the arguments after `children` into a single arg called `attributes` would simplify things greatly.  This should simplify `toNode` and `toMarkup` b/c we can iterate through `attributes` instead requiring all the `if` statements.  One slight downside is that now the caller will have to provide `viewBox` as a single value, but I think that's a small price to pay.

src/domTree.js
```diff
@@ -344,8 +336,164 @@ class symbolNode {
...
+
+        return markup;
+
+    }
+}
+
+class pathNode {
+    constructor(pathName, alternate) {
+        this.pathName = pathName || "";
+        this.alternate = alternate || "";  // Used only for tall \sqrt
```
kevinbarabash: The `|| ""` isn't necessary.

src/domTree.js
```diff
@@ -344,8 +336,164 @@ class symbolNode {
...
+        if (this.pathName !== "sqrtTall") {
+            return `<path d='${svgGeometry.path[this.pathName]}'/>`;
+        } else {
+            return `<path d='${this.alternate}'/>`;
+        }
+    }
+}
+
+class lineNode {
+    constructor(x1, y1, x2, y2, strokeWidth) {
```
kevinbarabash: This could be simplified to (not critical though):
```
constructor(x1 = 0, y1 = 0, x2 = 0, y2 = 0, strokeWidth = 0) {
    Object.assign({x1, y1, x2, y2, strokeWidth});
}
```

src/stretchy.js
```diff
@@ -137,309 +92,52 @@ const katexImagesData = {
...
+// For example, \overrightarrow uses the same arrowhead as glyph U+2192
+// from the KaTeX Main font. The scaling factor is 1000.
+// That is, inside the font, that arrowhead is 522 units tall, which
+// corresponds to 0.522 em inside the document.
 
-    xcancel: `<line x1='0' y1='0' x2='100%' y2='100%' stroke-width='0.046em'/>
-<line x1='0' y1='100%' x2='100%' y2='0' stroke-width='0.046em'/>`,
+const katexImagesData = {
+                   //   path(s), minWidth, height, align
+    overrightarrow: [["rightarrow"], 0.888, 522, "xMaxYMin"],
```
kevinbarabash: Why has the `minWidth` changed for some of these?  Previously it was `0.5` for `overrightarrow`.

src/stretchy.js
```diff
@@ -453,48 +151,86 @@ const groupLength = function(arg) {
...
+            } else if (numSvgChildren === 2) {
+                // small overlap to prevent a 1 pixel gap.
+                width = ["50.1%", "50%"][i];
+                x = [null, "50%"][i];
+                align = ["xMinYMin", "xMaxYMin"][i];
+            } else {
+                // 3 inner SVGs, as in a brace
+                width = ["25.5%", "50%", "25.1%"][i];
+                x = [null, "25%", "74.9%"][i];
+                align = ["xMinYMin", "xMidYMin", "xMaxYMin"][i];
```
kevinbarabash: Interesting way to grab all of these different values.

src/svgGeometry.js
```diff
@@ -0,0 +1,266 @@
+/**
+ * This file provides support to domTree.js
+ * It's a storehouse of path geometry for SVG images.
+ */
+
+const path = {
```
kevinbarabash: I like that these are all paths.

src/svgGeometry.js
```diff
@@ -0,0 +1,266 @@
...
+
+    widehat3: `M1181 0h2l1171 236c6 0 10 5 10 11l-2 23c-1 6-5 10
+-11 10h-1L1182 67 15 280h-1c-6 0-10-4-11-10l-2-23c-1-6 4-11 10-11z`,
+
+    widehat4: `M1181 0h2l1171 296c6 0 10 5 10 11l-2 23c-1 6-5 10
+-11 10h-1L1182 67 15 340h-1c-6 0-10-4-11-10l-2-23c-1-6 4-11 10-11z`,
+};
+
+module.exports = {
+    path: path,
```
kevinbarabash: Are you considering exporting other types of geometry at some point in the future?

src/svgGeometry.js
```diff
@@ -0,0 +1,266 @@
...
+-40.5-119-75.5s-58.3-74.8-73-119.5c-4.7-14-8.3-27.3-11-40-1.3-6.7-3.2-10.8-5.5
+-12.5-2.3-1.7-7.5-2.5-15.5-2.5-14 0-21 3.7-21 11 0 2 2 10.3 6 25 20.7 83.3 67
+ 151.7 139 205zm0 0v40h399900v-40z`,
+
+    rightbrace: `M400000 542l
+-6 6h-17c-12.7 0-19.3-.3-20-1-4-4-7.3-8.3-10-13-35.3-51.3-80.8-93.8-136.5-127.5
+s-117.2-55.8-184.5-66.5c-.7 0-2-.3-4-1-18.7-2.7-76-4.3-172-5H0V214h399571l6 1
+c124.7 8 235 61.7 331 161 31.3 33.3 59.7 72.7 85 118l7 13v35z`,
+
+    rightbraceunder: `M399994 0l6 6v35l-6 11c-56 104-135.3 181.3-238 232-57.3
```
kevinbarabash: Why do some paths start with `M400000` while others are a little shy of that value?

src/stretchy.js
```diff
@@ -453,48 +151,86 @@ const groupLength = function(arg) {
...
+        svgNode = new domTree.svgNode([path], "100%", height,
+            null, viewBoxWidth, viewBoxHeight, align);
+
     } else {
         const imgData = katexImagesData[label];
-        height = imgData[0];
-        depth = imgData[1];
-        imageName = imgData[2];
-        minWidth = imgData[3];
+        const numSvgChildren = imgData[0].length;
```
kevinbarabash: Instead of indexing into `imgData` for each value, using destructuring would help the reader know which value was which without having to scroll up to the definition of `katexIamgeData`, e.g.
```
const [paths, minWidth, height, align] = katexImagesData[label];
```

src/delimiter.js
```diff
@@ -313,107 +314,59 @@ const makeStackedDelim = function(delim, heightTotal, center, options, mode,
...
-    // tall is from glyph U23B7 in the font KaTeX_Size4-Regular
-    tall: `l-4 4-4 4c-.667.667-2 1.5-4 2.5s-4.167 1.833-6.5 2.5-5.5 1-9.5 1h
--12l-28-84c-16.667-52-96.667 -294.333-240-727l-212 -643 -85 170c-4-3.333-8.333
--7.667-13 -13l-13-13l77-155 77-156c66 199.333 139 419.667 219 661 l218 661z
-M702 0H400000v40H742z'/></svg>`,
+const sqrtSvg = function(sqrtName, height, viewBoxHeight, options) {
+    let alternate;
+    if (sqrtName === "sqrtTall") {
+        // sqrtTall is from glyph U23B7 in the font KaTeX_Size4-Regular
+        const vertSegment = viewBoxHeight - 54;
```
kevinbarabash: Where does the `54` come from?

src/buildHTML.js
```diff
@@ -1598,15 +1598,17 @@ groupTypes.xArrow = function(group, options) {
...
     const arrowBody = stretchy.svgSpan(group, options);
 
-    const arrowShift = -options.fontMetrics().axisHeight + arrowBody.depth;
-    const upperShift = -options.fontMetrics().axisHeight - arrowBody.height -
-        0.111;    // 2 mu. Ref: amsmath.dtx: #7\if0#2\else\mkern#2mu\fi
+    const arrowShift = -options.fontMetrics().axisHeight +
+        0.5 * arrowBody.height;
+    // 2 mu kern. Ref: amsmath.dtx: #7\if0#2\else\mkern#2mu\fi
+    const upperShift = -options.fontMetrics().axisHeight -
+        0.5 * arrowBody.height - 0.111;
```
kevinbarabash: Why the change in the shifts?

src/domTree.js
ronkok: Okay, I'll do that.

src/domTree.js
ronkok: A single `attributes` arg might be simpler in the constructor. But it also might be more complex in `stretchy.js` I'll look at it.

src/stretchy.js
ronkok: I discovered `opentype.js`. It made it easier to quickly look at the width of each glyph. So instead of using the width from one glyph and applying it several times, I got a number that matched that specific glyph. For instance, `\overrightarrow` is exactly 888 units width inside the font.

src/stretchy.js
ronkok: Will do.

src/svgGeometry.js
ronkok: When I wrote the geometry for the absurdly long arrow shafts, I made the end of each shaft take sharp 90° corners. So those points usually landed on x=400000.

In contrast, the KaTeX font glyphs don't have many sharp corners. If the font designer started the path at the right end of the glyph, it was never *exactly* on the edge of the border box. There is always a radius there.

src/svgGeometry.js
ronkok: I can't think of any other types at the moment, but I left it open in case I thought of something later.

src/delimiter.js
ronkok: From the glyph geometry.  The vertical path length is variable, but it lands on a point that is 54 units from the bottom of the surd.
landingpoint

And why, you ask, isn't there another calculation for the vertical length on the left edge? That's because I sequenced the points so that the path starts at the top of that left edge. The final command of a path `z` tells the browser to close the path back to the beginning point. So I've made the browser do that calculation for me.

src/buildHTML.js
ronkok: In previous versions, I treated extensible arrows slightly differently than the other SVGs. If an extensible arrow had a total height of 0.522 ems, then the values I returned to `buildHTML` had a `arrowBody.height` of 0.261 and a `arrowBody.depth` of 0.261. Now the code in `stretchy.js` is so array-driven, I found it easier to make things more consistent. Now, even extensible arrows return `.height` with their overall height value and return `.depth` as zero.

So the code inside `buildHTML` had to adjust and split that value in half, so that the arrow is centered on the math axis.

ronkok:
After picking up the review comments, I've refreshed the [test page](http://copyasmarkup.com/katex/wideelements.html)

src/stretchy.js
kevinbarabash: Awesome!  Are there situations in which the different `minWidth` will have an impact on the rendering?

src/domTree.js
kevinbarabash: In this case I probably wouldn't bother with a single `attributes` arg even though it deviates from the other constructors.  The reason is that x1, y1, x2, y2 is a well-known ordering of args and you're never going to skip one... it's probably all or none.

src/svgGeometry.js
kevinbarabash: I see... so some of these have the non-stretchy part on the right side and then stretchy line extends out to the left?

src/stretchy.js
ronkok: A function with an empty argument, say `\overrightarrow{}` will be rendered with the min-width. I suppose that min-width might also come into play if the argument is a single, narrow letter like `\imath`.

src/delimiter.js
kevinbarabash: I opened up `KaTeX-Size4` in `opentype.js to look at the coordinates of the surd's outline, but I couldn't find any coordinates near the bottom of the surd where subtracting would give `54`.

screen shot 2017-09-08 at 7 20 29 pm

src/buildHTML.js
kevinbarabash: > Now the code in stretchy.js is so array-driven, I found it easier to make things more consistent. Now, even extensible arrows return .height with their overall height value and return .depth as zero.

Could you add a comment describing this in stretchy.js?

src/svgGeometry.js
ronkok: Right. I would start with path geometry from a glyph. Say something like this:
shortarrow

Then I would:
1.  Run a little vector transformation script that would shift the whole arrow to the right until its right tip was at x=400000.
2. Edit the points at the left end of the path so they were back at x=0.

The result is a very long arrow:
longarrow

Okay, so it's hard to see. But on the gray background, that thin line is an arrow and there is an arrow head at one end, I promise. That absurdly long arrow is the inner SVG. The outer SVG is a window that only shows one small section at one end of the long arrow. The arrow head geometry is untouched, relatively. So I'm not really stretching an arrow; I'm really just revealing just enough arrow shaft to match the width of the subject matter.

src/delimiter.js
ronkok: Correction. That landing point is 14 units from the bottom of the surd (885 - 871 = 14). The path edge in question runs from the bottom of the viniculum to the landing point. The viniculum is 40 units thick. So the math is:

Relative distance from point to point = Overall height - 14 - 40 =  Overall height - 54

src/buildHTML.js
ronkok: Sure. I'll leave a note in both stretchy.js and in in buildHTML.js.

PR #845, opened 2017-09-05
To @flow: MacroExpander.

### LINKED ISSUE #842: Port MacroExpander to @flow
Port `MacroExpander.js` and `macro.js` to `@flow`.

I volunteer as tribute! (...assuming no one else is already on this) ;)

marcianx:
Fixes #842.

1) For sanity, I moved some logic into a private method, which I prefixed by `_` as I see that convention in a couple of other places. It kinda helps me to know more readily what methods are not intended to be called from outside the module. Should I put the convention in other places as well?
2) I re-added the caching for macros defined using strings and not functions.
3) In the recently-submitted updates to `MacroExpander.js`, I noticed that `expansion.call(this)` was used as opposed to just `expansion(this)` and taking an explicitly-defined parameter. For type safety, I'm inclined toward the latter, but it's not obvious for me why `expansion.call(this)` was preferred; any insights? @edemaine @kevinbarabash E.g. Currently, if I mistakenly change my code to `lookup()` instead of `lookup.call(this)`, flow doesn't give a type error since the function's signature doesn't indicate any required `this`.

### DIFF
--- a/flow-typed/object-assign.js
+++ b/flow-typed/object-assign.js
@@ -1,0 +1,6 @@
+declare module 'object-assign' {
+    declare module.exports:
+        <T>(target: {[string]: T}, ...sources: Array<{[string]: T}>)
+            => {[string]: T};
+}
+
--- a/src/MacroExpander.js
+++ b/src/MacroExpander.js
@@ -10,7 +10,7 @@ import builtinMacros from "./macros";
 import ParseError from "./ParseError";
 import objectAssign from "object-assign";
 
-import type {MacroContextInterface, MacroMap, ParsedExpansion} from "./macros";
+import type {MacroContextInterface, MacroMap, MacroExpansion} from "./macros";
 
 export default class MacroExpander implements MacroContextInterface {
     lexer: Lexer;
@@ -91,8 +91,8 @@ export default class MacroExpander implements MacroContextInterface {
             this.stack.push(topToken);
             return topToken;
         }
-        const {expansion: exp, numArgs} = this._getExpansion(name);
-        let expansion = exp;
+        const {tokens, numArgs} = this._getExpansion(name);
+        let expansion = tokens;
         if (numArgs) {
             const args: Token[][] = [];
             // obtain arguments, either single token or balanced {…} group
@@ -139,9 +139,7 @@ export default class MacroExpander implements MacroContextInterface {
                     if (tok.text === "#") { // ## → #
                         expansion.splice(i + 1, 1); // drop first #
                     } else if (/^[1-9]$/.test(tok.text)) {
-                        // expansion.splice(i, 2, arg[0], arg[1], …)
-                        // to replace placeholder with the indicated argument.
-                        // TODO: use spread once we move to ES2015
+                        // replace the placeholder with the indicated argument
                         expansion.splice(i, 2, ...args[tok.text - 1]);
                     } else {
                         throw new ParseError(
@@ -195,9 +193,10 @@ export default class MacroExpander implements MacroContextInterface {
      * argument count.
      * Caches macro expansions for those that were defined simple TeX strings.
      */
-    _getExpansion(name: string): ParsedExpansion {
-        const lookup = this.macros[name];
-        const expansion = typeof lookup === "function" ? lookup(this) : lookup;
+    _getExpansion(name: string): MacroExpansion {
+        const definition = this.macros[name];
+        const expansion =
+            typeof definition === "function" ? definition(this) : definition;
         if (typeof expansion === "string") {
             let numArgs = 0;
             if (expansion.indexOf("#") !== -1) {
@@ -214,10 +213,10 @@ export default class MacroExpander implements MacroContextInterface {
                 tok = bodyLexer.lex();
             }
             tokens.reverse(); // to fit in with stack using push and pop
-            const expanded = {expansion: tokens, numArgs};
+            const expanded = {tokens, numArgs};
             // Cannot cache a macro defined using a function since it relies on
             // parser context.
-            if (typeof lookup !== "function") {
+            if (typeof definition !== "function") {
                 this.macros[name] = expanded;
             }
             return expanded;
--- a/src/MacroExpander.js
+++ b/src/MacroExpander.js
@@ -10,9 +10,9 @@ import builtinMacros from "./macros";
 import ParseError from "./ParseError";
 import objectAssign from "object-assign";
 
-import type {MacroMap, ParsedExpansion} from "./macros";
+import type {MacroContextInterface, MacroMap, ParsedExpansion} from "./macros";
 
-export default class MacroExpander {
+export default class MacroExpander implements MacroContextInterface {
     lexer: Lexer;
     macros: MacroMap;
     stack: Token[];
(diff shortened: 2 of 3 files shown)


marcianx:
Regarding the `@flow` typechecker failing to catch the error in (3): https://github.com/facebook/flow/issues/4809

edemaine:
Regarding Point 3: This was my call: I thought it was nice to be given the `MacroExpander` as `this` in the macro function (jquery style? feels more object oriented?), but I could easily see switching to an argument style -- the latter would actually be more consistent with the rest of the code base (in particular `functions.js`).  So happy to go either way.  @kevinbarabash opinion?

kevinbarabash:
I think we should strive for consistency where possible.  Also, many of people find `this` confusing calling functions directly instead of using `.call` will hopefully make our code easier to understand.

marcianx:
Thank you both for the background. I do prefer keeping code as familiar as possible. Though `this` is reminiscent of a familiar object-oriented style, its frequent rebinding in javascript can be confusing to newcomers to the language who don't expect to see `this` outside the context of a class definition. As such, I avoid taking advantage of it if there's a viable/ergonomic alternative. Per your feedback, I've switched to using an explicit parameter.

As there are numerous methods in `MacroExpander`, it's not immediately obvious to me which are intended to be part of its public interface.

1) I've defined a `MacroContextInterface` in `macros.js` to encapsulate the ones that are currently used by macro definitions. Please let me know if any more of the existing methods should be added there.

2) As I didn't receive any feedback regarding (1), I've prefixed the other methods that I _think_ are intended to be private by `_`.

I've pushed my changes. Please correct me if any of the methods I marked private are actually intended to be public, or if the `_` notation is stylistically undesirable.

src/macros.js
```diff
@@ -6,10 +6,39 @@
...
+    /**
+     * Expand the next token only once (if possible), and return the resulting
+     * top token on the stack (without removing anything from the stack).
+     * Similar in behavior to TeX's `\expandafter\futurelet`.
+     */
+    expandAfterFuture(): Token;
+}
+
+/** Macro tokens (in reverse order). */
+export type ParsedExpansion = {expansion: Token[], numArgs: number};
```
kevinbarabash: Now that we've given this thing a name, `ParsedExpansion`, it might make sense to change it to be
```
{tokens: Token[], numArgs: number}
```
then
```
const {expansion: exp, numArgs} = this._getExpansion(name);
```
will become
```
const {tokens, numArgs} = this._getExpansion(name);
```
which is less awkward.

src/MacroExpander.js
```diff
@@ -200,6 +184,46 @@ class MacroExpander {
...
+        throw new Error(); // eslint-disable-line no-unreachable
+    }
+
+    /**
+     * Returns the expanded macro as a reversed array of tokens and a macro
+     * argument count.
+     * Caches macro expansions for those that were defined simple TeX strings.
+     */
+    _getExpansion(name: string): ParsedExpansion {
+        const lookup = this.macros[name];
```
kevinbarabash: Maybe `definition` would be a more apt variable name.  Then maybe `MacroExpansion` could be renamed to `MacroDefinition`.  I think this will better help separate the concepts of expanding macro definitions from the definitions themselves.  Also, renaming `ParsedExpansion` should probably be renamed to `MacroExpansion` as the parser hasn't done anything with the tokens yet.

TL;DR:
```
lookup -> definition
MacroExpansion -> MacroDefinition
ParsedExpansion -> MacroExpansion
```

src/MacroExpander.js
```diff
@@ -156,9 +142,7 @@ class MacroExpander {
                         // expansion.splice(i, 2, arg[0], arg[1], …)
                         // to replace placeholder with the indicated argument.
                         // TODO: use spread once we move to ES2015
-                        expansion.splice.apply(
-                            expansion,
-                            [i, 2].concat(args[tok.text - 1]));
+                        expansion.splice(i, 2, ...args[tok.text - 1]);
```
kevinbarabash: Can run `make build` and check what's being output for the spread operator.  Also, if we go with this, the `TODO` will no longer be necessary.

flow-typed/object-assign.js
```diff
@@ -0,0 +1,6 @@
+declare module 'object-assign' {
+    declare module.exports:
+        <T>(target: {[string]: T}, ...sources: Array<{[string]: T}>)
+            => {[string]: T};
```
kevinbarabash: It's too bad `flow` doesn't do this on its own.

src/MacroExpander.js
```diff
@@ -200,6 +184,46 @@ class MacroExpander {
                 }
             }
         }
+
+        // Flow unable to figure out that this pathway is impossible.
+        // https://github.com/facebook/flow/issues/4808
+        throw new Error(); // eslint-disable-line no-unreachable
```
kevinbarabash: I wonder if uglyifyjs would strip this out automatically?  It's probably easier to quiet `flow`'s warning about with a well-placed `$FlowFixMe` comment than it is to determine if this is getting into the output or not.

edemaine:
(moving discussion here) I'm not a huge fan of making those methods. (Sorry for not responding to (1) earlier.) I think *all* of these operations could be useful in implementing sophisticated macros... But could go either way.

src/macros.js
```diff
@@ -6,10 +6,39 @@
...
+
+type MacroExpansion = string | (MacroContextInterface => string) | ParsedExpansion;
```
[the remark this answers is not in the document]
marcianx: Yes. Would you prefer I rename it to `CachedExpansion` to make that clear?

src/MacroExpander.js
marcianx: Ooops! I meant to remove it. I'd failed to do this part when manually merging in my changes with Erik's previous commit. Indeed the results produced by `make` seem sane:
```javascript
var _expansion;
// replace placeholder with the indicated argument.
(_expansion = expansion).splice.apply(_expansion, [_i, 2].concat((0, _toConsumableArray3.default)(args[_tok.text - 1])));
```

src/MacroExpander.js
marcianx: It didn't. My concern with `$FlowFixMe` is that there's no single line within the method to put this on, and it would be put on the whole method, which means type safety would not be checked there. Would you still prefer I do it it? "throw new Error" already appears in the string, and I just tested and it makes no difference in the compressed size:

Sizes before gzip:
```
199950 katex1.min.js
199935 katex2.min.js
```
Sizes after gzip:
```
52621 katex1.min.js.gz
52621 katex2.min.js.gz
```

src/macros.js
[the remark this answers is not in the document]
marcianx: Ah nevermind, I see you've suggested calling it `MacroExpansion` below.

src/MacroExpander.js
kevinbarabash: It's too bad there's no good place to put the `$FlowFixMe`.  But since it doesn't make a difference when gzipped let's leave it in.  gzip is magic.

src/macros.js
marcianx: Done.

src/MacroExpander.js
marcianx: Done.

marcianx:
O wait. Gonna rebase to new master.

marcianx:
I'm seeing "Merging is blocked.
Merging can be performed automatically once the requested changes are addressed"

Any idea why? Do I have to mark any comments as resolved?

kevinbarabash:
> Any idea why? Do I have to mark any comments as resolved?

I have to click "Approve changes" after reviewing them.

kevinbarabash:
We can save some space by using the `loose` option on [transform-es2015-spread](https://babeljs.io/docs/plugins/transform-es2015-spread/) babel plugin.  I'm going to approve this, but we should fix #850 before publishing.

kevinbarabash:
@marcianx thanks for PR.  I'm excited about how quickly we're updating our code to use flow.

PR #846, opened 2017-09-05
To @flow: fontMetrics, fontMetricsData, Options, Settings, Style

xsznix:
I've inferred the intended types as best as I can from the code, but it'd be worthwhile to double check.

I'm not sure how to fix old commits showing up in this PR… any ideas?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,6 +26,7 @@ export UGLIFYJS = $(realpath ./node_modules/.bin/uglifyjs) \
 	--mangle \
 	--beautify \
 	ascii_only=true,beautify=false
+export CLEANCSS = $(realpath ./node_modules/.bin/cleancss)
 
 # The prepublish script in package.json will override the following variable,
 # setting it to the empty string and thereby avoiding an infinite recursion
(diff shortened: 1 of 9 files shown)


kevinbarabash:
@xsznix thanks for expanding our use of flow.  I can give you a hand with git this evening.

kevinbarabash:
The issue is that even though some of the commits in that list have landed in `Khan/master`, they've been rebased which means that they been given different commit ids and git thinks they're different commits.

In general I try to make all PRs from a branch.  That way I can easily update my `master` branch to match what's on `Khan/master`.

I suggest you do the following:
- create a feature branch from your current master: `git commit -b flow1`
- then push your branch to your remote repo: `git push origin flow1`
- then delete your local copy of `master` by doing: `git branch -D master`
- then checkout `Khan/KaTeX`'s master by doing: `git checkout khan/master` (assumes you have remote called `khan` pointing at https://github.com/Khan/KaTeX`) followed by `git checkout -b master`.  Now your local `master` has the same thing that's in `khan/master`.
- then do `git push origin master --force`, this will forcefully update `xsznix/master` to match what's locally in master
- create another feature branch from your current master: `git commit -b flow2`
- then cherry pick the commits you want: `git cherry-pick 5d2325f` followed by `git cherry-pick d250de0`
- finally do `git push origin flow2`, you should be able to do a clear PR using `flow2`.

Please note: I am not a git guru.  There are probably other more succinct ways of doing this.  You may also want to create a copy of the repo before starting just in case something goes wrong.

kevinbarabash:
@xsznix I can also squash this into a single commit when I merge this and then you could do a fresh fork of `Khan/KaTeX` and start using feature branches from a clean repo instead of having to do what I suggested above.

xsznix:
Cool, thanks for the tips! I'll make sure to work on feature branches from now on.

PR #847, opened 2017-09-06
Fix match-at flow errors

ariabuckles:
Upgrade match-at and flow versions to resolve the match-at flow
errors, allowing us to remove match-at from the .flowconfig ignore

Test plan:

run `jest` and get the following output:

    PASS  test/unicode-spec.js
    PASS  test/errors-spec.js
    PASS  test/mathml-spec.js
    PASS  test/katex-spec.js (5.104s)

run `flow` and get:

    No errors!


ariabuckles:
thanks :D

PR #848, opened 2017-09-06
To @flow: fontMetrics, fontMetricsData, Options, Settings, Style

xsznix:
Supersedes #846.

I've inferred the intended types as best as I can from the code, but it'd be worthwhile to double check.


kevinbarabash:
@xsznix glad to see you got things working with a feature branch.

src/Options.js
```diff
@@ -31,10 +32,20 @@ const sizeMultipliers = [
     0.5, 0.6, 0.7, 0.8, 0.9, 1.0, 1.2, 1.44, 1.728, 2.074, 2.488,
 ];
 
-const sizeAtStyle = function(size, style) {
+const sizeAtStyle = function(size: number, style: StyleInterface): number {
     return style.size < 2 ? size : sizeStyleMap[size - 1][style.size - 1];
 };
 
+export type OptionsData = $Shape<{
```
kevinbarabash: I don't think we should use `$Shape`.  This will require all props on `OptionData` to be marked as optional, e.g. `color?: string` which is okay imo b/c in a couple of places we're missing some existence checks.

src/Settings.js
```diff
@@ -1,10 +1,20 @@
+// @flow
 /**
  * This is a module for storing settings passed into KaTeX. It correctly handles
  * default settings.
  */
 
 import utils from "./utils";
 
+type SettingsOptions = $Shape<{
```
kevinbarabash: All of the props on `SettingsOptions` should be optional as well.

src/Options.js
```diff
@@ -43,10 +54,25 @@ const sizeAtStyle = function(size, style) {
...
 class Options {
-    constructor(data) {
+    style: StyleInterface;
+    color: string;
+    size: number;
+    textSize: number;
+    phantom: boolean;
+    font: string;
+    sizeMultiplier: number;
+    maxSize: number;
```
kevinbarabash: These should all be optional as well.

src/fontMetricsData.js
```diff
@@ -1,3 +1,4 @@
+// @flow
 module.exports = {
```
marcianx: I recommend giving it an explicit tuple `[number, number, number, number]` type instead of an array type `number[]`. This allowed me to catch a (mostly harmless) error in `fontMetrics.js`. 
E.g. via a local type:
```
type FontMetrics = {[string]: [number, number, number, number]};

// character code => [depth, height, italic, skew]
const fontMetricsData: {[string]: FontMetrics} = {
...
};

export default fontMetricsData;
```

src/fontMetrics.js
```diff
@@ -226,21 +227,32 @@ const extraCharacterMap = {
...
     let ch = character.charCodeAt(0);
     if (character[0] in extraCharacterMap) {
         ch = extraCharacterMap[character[0]].charCodeAt(0);
     } else if (cjkRegex.test(character[0])) {
         ch = 'M'.charCodeAt(0);
     }
-    const metrics = metricMap[style][ch];
+    const metrics = metricMap[font][ch];
     if (metrics) {
         return {
```
marcianx: Ah, sorry, I'd started porting this without paying attention to the existence of this commit. In any case, if you use a tuple type (that I commented on further below), flow will point out that `width: metrics[4]` is doing an out-of-bounds access.
So remove `width:` here. I grepped through the code base and I couldn't find any uses of it either. ;)

The now-closed [PR 853](https://github.com/Khan/KaTeX/pull/853) contains the work I did on `frontMetrics*` and `unicodeRegex`.

src/fontMetrics.js
```diff
@@ -252,12 +264,14 @@ const getCharacterMetrics = function(character, style) {
     }
 };
 
-const fontMetricsBySizeIndex = {};
+export type FontMetrics = {[charCode: number]: Array<number>};
```
marcianx: I recommend something along the lines of
```javascript
type FontSizeIndex = 0 | 1 | 2;
type FontSizeMetrics = {[string]: number};
const fontMetricsBySizeIndex: {[FontSizeIndex]: FontSizeMetrics} = {};
```
and using `let sizeIndex: FontSizeIndex` below. The now defunct [PR 853](https://github.com/Khan/KaTeX/pull/853) demonstrates this.

src/Options.js
```diff
@@ -36,15 +36,15 @@ const sizeAtStyle = function(size: number, style: StyleInterface): number {
...
-    color: string;
-    size: number;
-    textSize: number;
-    phantom: boolean;
-    font: string;
+    color?: ?string;
+    size?: ?number;
+    textSize?: ?number;
+    phantom?: ?boolean;
+    font?: ?string;
```
kevinbarabash: These only need the first `?`.  The second `?` means that they can be assigned the value `null`, but I don't think we do that anywhere.

src/Options.js
```diff
@@ -290,11 +290,16 @@ class Options {
      * Gets the CSS color of the current options object, accounting for the
      * `colorMap`.
      */
-    getColor(): string {
+    getColor(): ?string {
```
kevinbarabash: I think this can still be `getColor(): string`.

src/Options.js
```diff
@@ -85,7 +85,7 @@ class Options {
      * Returns a new options object with the same properties as "this".  Properties
      * from "extension" will be copied to the new options object.
      */
-    extend(extension: OptionsData): Options {
+    extend(extension: $Shape<OptionsData>): Options {
```
kevinbarabash: Can you remove `$Shape` here as well?  I think it may be hiding some null checking we should be doing.

src/Settings.js
```diff
@@ -6,14 +6,14 @@
...
-    colorIsTextColor: boolean;
-    maxSize: number;
-}>;
+type SettingsOptions = {
+    displayMode?: ?boolean;
+    throwOnError?: ?boolean;
+    errorColor?: ?string;
+    macros?: ?{[macroName: string]: string};
+    colorIsTextColor?: ?boolean;
+    maxSize?: ?number;
```
kevinbarabash: Only the first `?` should be necessary here too.

src/Options.js
xsznix: I think this `$Shape` here is necessary since you should able to do stuff like `newOptions = options.extend({color: 'white'})`, which doesn't set the required `style` and `maxSize` properties in `OptionData`. You still wouldn't be able to do stuff like `options.extend({style: null})` (see [Flow docs](http://sitr.us/2015/05/31/advanced-features-in-flow.html#shapet)).

src/fontMetrics.js
```diff
@@ -252,27 +251,30 @@ const getCharacterMetrics = function(
...
     }
-    const metrics = metricMap[font][ch];
+    const metrics = metricMap[font]['' + ch];
     if (metrics) {
         return {
             depth: metrics[0],
             height: metrics[1],
             italic: metrics[2],
             skew: metrics[3],
-            width: metrics[4],
```
kevinbarabash: `width` was here for experimental reasons.  By default fontMetricsData.js doesn't include the width of the characters, but if you do `make extended_metrics` you can build a copy that does.

test/katex-spec.js
```diff
@@ -17,6 +17,7 @@ const defaultSettings = new Settings({});
 const defaultOptions = new Options({
     style: Style.TEXT,
     size: 5,
+    maxSize: Infinity,
```
kevinbarabash: nice

kevinbarabash: hmm... this looks like a default, it should probably go inside the `Options` constructor.

src/fontMetrics.js
xsznix: should this be left in then?

src/Options.js
xsznix: Not if `Options.color` is nullable…

src/Options.js
xsznix: `color` and `font` need the second `?` in order to pass type checking in the last line in `extend`, since those properties are nullable on `Options`.

src/fontMetrics.js
edemaine: I think so? I believe #632 plans to (eventually) add this to master.

src/Options.js
```diff
@@ -43,12 +54,27 @@ const sizeAtStyle = function(size, style) {
  * different properties, call a `.having*` method.
  */
 class Options {
-    constructor(data) {
+    style: StyleInterface;
+    color: ?string;
```
kevinbarabash: I this is changed to `color: string | void` then the `color` prop in `OptionsData` could be `style?: string` like the other optional properties.  This would also match existing code in https://github.com/Khan/KaTeX/blob/588f5a1ee64b1c4ebde18b1af139ad1e6d00ce63/src/ParseNode.js#L19-L21 for optional properties on classes.

src/Options.js
```diff
@@ -43,12 +54,27 @@ const sizeAtStyle = function(size, style) {
...
+    size: number;
+    textSize: number;
+    phantom: boolean;
+    font: ?string;
```
kevinbarabash: same

src/Options.js
```diff
@@ -265,18 +290,18 @@ class Options {
      * Gets the CSS color of the current options object, accounting for the
      * `colorMap`.
      */
-    getColor() {
+    getColor(): ?string {
```
kevinbarabash: If the `color` prop is defined as `color: string | void` on the class, then the return value would have to change to `string | void` here as well.

src/fontMetrics.js
kevinbarabash: @edemaine how useful are horizontal metrics after the introduction of @ronkok's SVG code?

src/Settings.js
```diff
@@ -1,10 +1,20 @@
...
 
 import utils from "./utils";
 
+type SettingsOptions = {
+    displayMode?: boolean;
+    throwOnError?: boolean;
+    errorColor?: string;
+    macros?: {[macroName: string]: string};
+    colorIsTextColor?: boolean;
+    maxSize?: number;
```
kevinbarabash: This could be `maxSize: number` since we always set `maxSize` to a number in the constructor.

src/Options.js
```diff
@@ -59,7 +85,7 @@ class Options {
      * Returns a new options object with the same properties as "this".  Properties
      * from "extension" will be copied to the new options object.
      */
-    extend(extension) {
+    extend(extension: $Shape<OptionsData>): Options {
```
kevinbarabash: The `$Shape` is no longer necessary.

src/fontMetrics.js
ronkok: `\rotatebox` and `\cancelto` cannot be done with CSS trickery, at least not any that I can think of. So it may yet be necessary to know glyph widths and also something equivalent to `makeVList` to find layout widths. 

Other than those two functions, I can think of no uses for horizontal font metrics

edemaine: I guess if we would ever want to support `\width`, \wd`, etc.? (though only approximately)  There are zeros in there now, I don't see why not populate them.  I'm also not sure about advanced `tabular` support (e.g. `\multicolumn) -- it might have use there.

src/Settings.js
xsznix: I'm not sure I'm following. `SettingsOptions` is what the user passes into the `Settings` constructor, and as we established earlier, all the properties here should be optional.

test/katex-spec.js
xsznix: There would be a bit of code duplication if we added this to the `Options` constructor because the default is already being parsed in the `Settings` constructor:

https://github.com/Khan/KaTeX/blob/ccf09786cc747080cdaa091c2ba2251861d1d138/src/Settings.js#L27

Is that okay, or is there a more elegant solution?

src/Options.js
xsznix: I'm pretty sure it's necessary. Taking it out leads to the following Flow errors:

```
Error: Options.js:116
                        v------------
116:             return this.extend({
117:                 style: style,
118:                 size: sizeAtStyle(this.textSize, style),
119:             });
                 -^ call of method `extend`
 88:     extend(extension: OptionsData): Options {
                           ^^^^^^^^^^^ property `maxSize`. Property not found in
                                    v
116:             return this.extend({
117:                 style: style,
118:                 size: sizeAtStyle(this.textSize, style),
119:             });
                 ^ object literal

Error: Options.js:139
                        v------------
139:             return this.extend({
140:                 style: this.style.text(),
141:                 size: size,
142:                 textSize: size,
143:             });
                 -^ call of method `extend`
 88:     extend(extension: OptionsData): Options {
                           ^^^^^^^^^^^ property `maxSize`. Property not found in
                                    v
139:             return this.extend({
140:                 style: this.style.text(),
141:                 size: size,
142:                 textSize: size,
143:             });
                 ^ object literal

Error: Options.js:158
                        v------------
158:             return this.extend({
159:                 style: style,
160:                 size: wantSize,
161:             });
                 -^ call of method `extend`
 88:     extend(extension: OptionsData): Options {
                           ^^^^^^^^^^^ property `maxSize`. Property not found in
                                    v
158:             return this.extend({
159:                 style: style,
160:                 size: wantSize,
161:             });
                 ^ object literal

Error: Options.js:169
                    v------------
169:         return this.extend({
170:             color: color,
171:         });
             -^ call of method `extend`
 88:     extend(extension: OptionsData): Options {
                           ^^^^^^^^^^^ property `maxSize`. Property not found in
                                v
169:         return this.extend({
170:             color: color,
171:         });
             ^ object literal

Error: Options.js:169
                    v------------
169:         return this.extend({
170:             color: color,
171:         });
             -^ call of method `extend`
 88:     extend(extension: OptionsData): Options {
                           ^^^^^^^^^^^ property `style`. Property not found in
                                v
169:         return this.extend({
170:             color: color,
171:         });
             ^ object literal

Error: Options.js:178
                    v------------
178:         return this.extend({
179:             phantom: true,
180:         });
             -^ call of method `extend`
 88:     extend(extension: OptionsData): Options {
                           ^^^^^^^^^^^ property `maxSize`. Property not found in
                                v
178:         return this.extend({
179:             phantom: true,
180:         });
             ^ object literal

Error: Options.js:178
                    v------------
178:         return this.extend({
179:             phantom: true,
180:         });
             -^ call of method `extend`
 88:     extend(extension: OptionsData): Options {
                           ^^^^^^^^^^^ property `style`. Property not found in
                                v
178:         return this.extend({
179:             phantom: true,
180:         });
             ^ object literal

Error: Options.js:187
                    v------------
187:         return this.extend({
188:             font: font || this.font,
189:         });
             -^ call of method `extend`
 88:     extend(extension: OptionsData): Options {
                           ^^^^^^^^^^^ property `maxSize`. Property not found in
                                v
187:         return this.extend({
188:             font: font || this.font,
189:         });
             ^ object literal

Error: Options.js:187
                    v------------
187:         return this.extend({
188:             font: font || this.font,
189:         });
             -^ call of method `extend`
 88:     extend(extension: OptionsData): Options {
                           ^^^^^^^^^^^ property `style`. Property not found in
                                v
187:         return this.extend({
188:             font: font || this.font,
189:         });
             ^ object literal

File: src/Options.js (hunk 13 above)
Comment (xsznix) [reply src/Options.js]: I've switched to `T | void` for nullable types, but this still doesn't allow `color?: string` in `OptionsData`.

In `Options.extend`, `data` must be compatible with the type `OptionsData` in order to be passed into `new Options(data)`, but if `this.color` is of type `?string`, you cannot assign it to property `color` of an `OptionsData`, which must have type `string`. The same applies for `font`.

File: src/Settings.js (hunk 16 above)
Comment (kevinbarabash) [reply src/Settings.js]: Sorry, my bad.  I'm confounding `SettingOptions` with the properties on the `Settings` class.

File: test/katex-spec.js (hunk 12 above)
Comment (kevinbarabash) [reply test/katex-spec.js]: That's too bad about the duplication.  Let's leave this as is for now.

File: src/Options.js (hunk 17 above)
Comment (kevinbarabash) [reply src/Options.js]: I was thinking all the props could be optional in `OptionsData`, but I guess that's the purpose of `$Shape` so we may as well use it.  I was thinking that by having all props in `OptionsData` look like `prop?: T` that would communicate to people that each prop was optional, but it should be pretty obvious by the existing usage.

Comment (kevinbarabash) [reply src/Options.js]: Let's use `$Shape` for now and if people find it confusing we can always change it later.

File: src/Options.js [hunk 18]
```diff
@@ -158,7 +183,7 @@ class Options {
     /**
      * Create a new options objects with the give font.
      */
-    withFont(font) {
+    withFont(font: ?string): Options {
         return this.extend({
             font: font || this.font,
```
kevinbarabash: All b/c we use `"\\text": undefined` in `textFunctionFonts`.  That's something we should maybe look at changing in the future.

kevinbarabash:
@xsznix thanks for the PR.  Sorry about the back-and-forth on how to type some of these things.

src/Options.js
xsznix: Yeah; it doesn't seem like `style` could be made optional because `Options.havingCrampedStyle` and `Options.havingSize` both use `style` without null checks.

xsznix:
@kevinbarabash np, it's always good to have a second set of eyes to double check everything

edemaine:
One question now that you have expertise on this: given the move to Flow, is it still worthwhile separating a `Settings` class from a settings object?  I've always felt like this was redundant, and just to get some kind of type checking, but now that we do type-checking elsewise... maybe the `Settings` class is no longer needed?  Curious of your opinion before opening an issue.

xsznix:
Yeah, I think getting rid of the separate `Settings` class would be good. Right now each property is repeated three times in the code (once in `SettingsOptions`, once as fields on `Settings`, and once in the constructor to set those fields), and we can reduce that to two by doing something like:

```js
// Settings.js
export type Settings = {...};
export const defaultSettings = {...};

// somewhere else
settings = Object.assign({}, Settings.defaultSettings, settings);
```

Issue #844, opened 2017-09-04
add lint check for `// @flow` at the top of files that use flow type annotations


kevinbarabash:
https://github.com/gajus/eslint-plugin-flowtype#no-types-missing-file-annotation

PR #849, opened 2017-09-06
eslint-plugin-flowtype & upgrade to eslint@4

edemaine:
I added `eslint-plugin-flowtype` as requested in #844.  I used the recommended settings for plugin, which involved adding some spaces to some existing union types.  Hopefully doesn't cause much trouble with new type PRs...

I also upgraded `eslint` to version 4.  This led to finding some escape bugs.  I switched to `indent-legacy` to allow e.g. comments to have extra indents -- otherwise there were lots of errors, not all of which I agreed with.


kevinbarabash:
@edemaine thanks for avoiding massive changes to current comment indentation.

PR #855, opened 2017-09-07
Port unit.js to @flow.

### LINKED ISSUE #854: Port `units.js` to @flow
Technically, this soft-requires #848 since `units.js` uses the `Option` type. However, it seems to work even without that.

marcianx:
Fixes #854.

### DIFF
--- a/src/units.js
+++ b/src/units.js
@@ -1,4 +1,4 @@
-/* eslint no-console:0 */
+// @flow
 
 /**
  * This file does conversion between units.  In particular, it provides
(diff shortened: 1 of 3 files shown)


src/units.js
```diff
@@ -1,17 +1,18 @@
...
 
 import ParseError from "./ParseError";
+import Options from "./Options";
 
 // This table gives the number of TeX pts in one of each *absolute* TeX unit.
 // Thus, multiplying a length by this number converts the length from units
 // into pts.  Dividing the result by ptPerEm gives the number of ems
 // *assuming* a font size of ptPerEm (normal size, normal style).
-const ptPerUnit = {
+const ptPerUnit: {[string]: number} = {
```
kevinbarabash: Is this necessary?

xsznix: No; Flow will infer the type automatically. Ditto for line 34 as well.

xsznix: hm, on second thought, Flow isn't as good at inferring types when you use computed properties, so I think it's probably good to keep this annotation and the one below.

marcianx: Removed this and the one on line 34.

PR #856, opened 2017-09-07
Port unicodeRegexes to @flow.

marcianx:
This is extremely simple.

### DIFF
--- a/src/unicodeRegexes.js
+++ b/src/unicodeRegexes.js
@@ -1,4 +1,5 @@
-const hangulRegex = /[\uAC00-\uD7AF]/;
+// @flow
+export const hangulRegex = /[\uAC00-\uD7AF]/;
 
 // This regex combines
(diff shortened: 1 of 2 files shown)


src/unicodeRegexes.js
```diff
@@ -8,10 +9,5 @@ const hangulRegex = /[\uAC00-\uD7AF]/;
 // - Hangul syllables: [\uAC00-\uD7AF]
 // - Fullwidth punctuation: [\uFF00-\uFF60]
 // Notably missing are halfwidth Katakana and Romanji glyphs.
-const cjkRegex =
+export const cjkRegex: RegExp =
```
xsznix: This is redundant; you don't have to explicitly annotate consts for Flow to recognize their type.

marcianx: Removed. And updated commit description.

PR #858, opened 2017-09-09
Port symbol.js to @flow.

### LINKED ISSUE #857: Port symbols.js to @flow

marcianx:
Fixed #857.

### DIFF
--- a/src/ParseNode.js
+++ b/src/ParseNode.js
@@ -1,5 +1,6 @@
 // @flow
 import {LexerInterface, Token} from "./Token";
+import type {Mode} from "./types";
 
 /**
  * The resulting parse tree nodes of the parse tree.
@@ -12,7 +13,7 @@ import {LexerInterface, Token} from "./Token";
 export default class ParseNode {
     type: *;
     value: *;
-    mode: *;
+    mode: Mode;
     // TODO: We should combine these to ({lexer, start, end}|void) as they
     // should all exist together or not exist at all. That way, only a single
     // void check needs to be done to see if we have metadata.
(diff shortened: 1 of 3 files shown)


marcianx:
Actually, wait a sec. The `mode` is used in `ParserNode` as well, so I'll pull it out into a separate file that just defines the `mode` type.

marcianx:
Pulled out  `Mode`. I wasn't sure whether I should put it in `types.js` or `Mode.js`. Types cannot be default exported, so `import {Mode} from "./Mode";` didn't seem as nice. But that's what I've gone with.

marcianx:
Changed my mind again and renamed `Mode.js` to `types.js` since I foresee adding a `ParserInterface` type to break circular dependencies.

src/symbols.js
```diff
@@ -16,20 +17,28 @@
...
-function defineSymbol(mode, font, group, replace, name, acceptUnicodeChar) {
-    module.exports[mode][name] = {
-        font: font,
-        group: group,
-        replace: replace,
-    };
+/** `acceptUnicodeChar = true` is only applicable if `replace` is set. */
+function defineSymbol(
+    mode: Mode, font: Font, group: Group, replace: ?string, name: string,
+    acceptUnicodeChar?: boolean) {
```
kevinbarabash: Ideally I'd like function args to formatted as so it's easier to read the args:
```
function foo(
   arg0: T0,
   arg1: T1,
   ...
   argn: Tn,
) {
   console.log('bar');
   ...
}
```
Unfortunately, I don't believe our babel setup supports trailing commas in function defintions and calls so let's leave this for now.  I opened an issue for this: https://github.com/Khan/KaTeX/issues/861.

src/types.js
```diff
@@ -0,0 +1,9 @@
+// @flow
+
+/**
+ * This file consists only of basic flow types used in multiple places.
+ * For types with javascript, create separate files by themselves.
```
kevinbarabash: Nice!

src/ParseNode.js
```diff
@@ -23,7 +24,7 @@ export default class ParseNode {
     constructor(
         type: string,       // type of node, like e.g. "ordgroup"
         value: mixed,       // type-specific representation of the node
```
kevinbarabash: I just noticed this was `mixed`.  Hopefully we'll be able to make this more precise in the future.

marcianx: I agree. A lot of refactoring needs to be done before we can do that.

PR #862, opened 2017-09-09
Port environments.js to @flow.

### LINKED ISSUE #859: Port environments.js to @flow

marcianx:
Fixes #859.
Sits on top of [PR 860](https://github.com/Khan/KaTeX/pull/860), which needs to be submitted first!

### DIFF
--- a/src/environments.js
+++ b/src/environments.js
@@ -13,10 +13,10 @@ import type {ArgType, Mode, StyleStr} from "./types";
  *  - positions: the positions associated with these arguments from args.
  */
 type EnvContext = {
-  mode: Mode,
-  envName: string,
-  parser: Parser,
-  positions: number[],
+    mode: Mode,
+    envName: string,
+    parser: Parser,
+    positions: number[],
 };
 
 /**
@@ -36,7 +36,7 @@ type EnvHandler = (context: EnvContext, args: ParseNode[]) => ParseResult;
  *  - numArgs: (default 0) The number of arguments after the \begin{name} function.
  *  - argTypes: (optional) Just like for a function
  *  - allowedInText: (default false) Whether or not the environment is allowed
- *  i                nside text mode (not enforced yet).
+ *                   inside text mode (not enforced yet).
  *  - numOptionalArgs: (default 0) Just like for a function
  */
 type EnvProps = {
@@ -58,18 +58,18 @@ const environments: {[string]: EnvData} = {};
 export default environments;
 
 // Data stored in the ParseNode associated with the environment.
-type ArrayColSpec = { type: "separator", separator: string } | {
+type AlignSpec = { type: "separator", separator: string } | {
     type: "align",
     align: string,
     pregap?: number,
     postgap?: number,
 };
-type EnvNodeData = {
+type ArrayEnvNodeData = {
     type: "array",
     hskipBeforeAndAfter?: boolean,
     arraystretch?: number,
     addJot?: boolean,
-    cols?: ArrayColSpec[],
+    cols?: AlignSpec[],
     // These fields are always set, but not on struct construction
     // initialization.
     body?: ParseNode[][], // List of rows in the (2D) array.
@@ -84,7 +84,7 @@ type EnvNodeData = {
  */
 function parseArray(
     parser: Parser,
-    result: EnvNodeData,
+    result: ArrayEnvNodeData,
     style: StyleStr,
 ) {
     let row = [];
@@ -123,22 +123,14 @@ function parseArray(
 
 /*
  * An environment definition is very similar to a function definition:
- * it is declared with a name or a list of names, a set of properties
- * and a handler containing the actual implementation.
- *
- * - props can either be an EnvProps struct, or a number indicating the numArgs.
+ * it is declared with a list of names, a set of properties and a handler
+ * containing the actual implementation.
  */
 function defineEnvironment(
-    names: string | string[],
-    props: EnvProps | number,
+    names: string[],
+    props: EnvProps,
     handler: EnvHandler,
 ) {
-    if (typeof names === "string") {
-        names = [names];
-    }
-    if (typeof props === "number") {
-        props = ({ numArgs: props }: EnvProps);
-    }
     // Set default values of environments
     const data = {
         numArgs: props.numArgs || 0,
@@ -274,7 +266,7 @@ defineEnvironment([
 // except it operates within math mode.
 // Note that we assume \nomallineskiplimit to be zero,
 // so that \strut@ is the same as \strut.
-defineEnvironment("aligned", {
+defineEnvironment(["aligned"], {
 }, function(context) {
     let res = {
         type: "array",
@@ -319,7 +311,7 @@ defineEnvironment("aligned", {
 // A gathered environment is like an array environment with one centered
 // column, but where rows are considered lines so get \jot line spacing
 // and contents are set in \displaystyle.
-defineEnvironment("gathered", {
+defineEnvironment(["gathered"], {
 }, function(context) {
     let res = {
         type: "array",
--- a/src/environments.js
+++ b/src/environments.js
@@ -1,6 +1,80 @@
-/* eslint no-constant-condition:0 */
+// @flow
 import ParseNode from "./ParseNode";
 import ParseError from "./ParseError";
+
+import type Parser from "./Parser";
+import type {ArgType, Mode, StyleStr} from "./types";
+
+/**
+ * The context contains the following properties:
+ *  - mode: current parsing mode.
+ *  - envName: the name of the environment, one of the listed names.
+ *  - parser: the parser object.
+ *  - positions: the positions associated with these arguments from args.
+ */
+type EnvContext = {
+  mode: Mode,
+  envName: string,
+  parser: Parser,
+  positions: number[],
+};
+
+/**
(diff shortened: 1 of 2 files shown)


src/environments.js
```diff
@@ -1,18 +1,96 @@
...
+ *  - envName: the name of the environment, one of the listed names.
+ *  - parser: the parser object.
+ *  - positions: the positions associated with these arguments from args.
+ */
+type EnvContext = {
+  mode: Mode,
+  envName: string,
+  parser: Parser,
+  positions: number[],
+};
```
kevinbarabash: nit: 4 space tab.

src/environments.js
```diff
@@ -1,18 +1,96 @@
...
+
+/**
+ * List of ParseNodes followed by an array of positions.
+ * Returned by `Parser`'s `parseArguments`.
+ */
+type ParseResult = (ParseNode | number[])[] | ParseNode;
```
kevinbarabash: When do we return a result with type `number[][]`?

src/environments.js
```diff
@@ -1,18 +1,96 @@
...
+};
+
+type EnvData = {
+    numArgs: number,
+    argTypes?: ArgType[],
+    greediness: number,
+    allowedInText: boolean,
+    numOptionalArgs: number,
+    handler: EnvHandler,
+};
```
kevinbarabash: This looks suspiciously similar to how functions are typed.  We might be able to dedupe in future.

src/environments.js
```diff
@@ -1,18 +1,96 @@
...
+ *  - context: information and references provided by the parser
+ *  - args: an array of arguments passed to \begin{name}
+ */
+type EnvHandler = (context: EnvContext, args: ParseNode[]) => ParseResult;
+
+/**
+ *  - numArgs: (default 0) The number of arguments after the \begin{name} function.
+ *  - argTypes: (optional) Just like for a function
+ *  - allowedInText: (default false) Whether or not the environment is allowed
+ *  i                nside text mode (not enforced yet).
```
kevinbarabash: `i nside => inside`

src/environments.js
```diff
@@ -1,18 +1,96 @@
...
+export default environments;
+
+// Data stored in the ParseNode associated with the environment.
+type ArrayColSpec = { type: "separator", separator: string } | {
+    type: "align",
+    align: string,
+    pregap?: number,
+    postgap?: number,
+};
+type EnvNodeData = {
```
kevinbarabash: `ArrayEnvNode` or `ArrayNode` would be better as this type is specific to arrays.  For consistency `ArrayColSpec` could be `AlignNode` and we could even introduce a `SeparateNode` type.

src/environments.js
```diff
@@ -1,18 +1,96 @@
...
+function parseArray(
+    parser: Parser,
+    result: EnvNodeData,
+    style: StyleStr,
+) {
     let row = [];
     const body = [row];
     const rowGaps = [];
-    while (true) {
+    for (;;) {
```
kevinbarabash: Why the change?

src/environments.js
```diff
@@ -1,18 +1,96 @@
...
  * columns delimited by &, and create a nested list in row-major order
  * with one group per cell.  If given an optional argument style
  * ("text", "display", etc.), then each cell is cast into that style.
  */
-function parseArray(parser, result, style) {
+function parseArray(
+    parser: Parser,
+    result: EnvNodeData,
+    style: StyleStr,
+) {
```
kevinbarabash: Thank you.

src/environments.js
```diff
@@ -42,35 +120,24 @@ function parseArray(parser, result, style) {
...
- *  - envName: the name of the environment, one of the listed names.
- *  - parser: the parser object
- *  - lexer: the lexer object
- *  - positions: the positions associated with these arguments from args.
- * The handler must return a ParseResult.
+ * - props can either be an EnvProps struct, or a number indicating the numArgs.
  */
-function defineEnvironment(names, props, handler) {
+function defineEnvironment(
+    names: string | string[],
```
kevinbarabash: We updated `defineFucntion` to only take an array of `strings`.  We should probably do the same here.

src/environments.js
```diff
@@ -42,35 +120,24 @@ function parseArray(parser, result, style) {
...
+    props: EnvProps | number,
```
kevinbarabash: `defineFunction` doesn't support taking a `number` for `props`, we should do the same here too.

src/environments.js
marcianx: Oi! Time to change my vim defaults. I keep forgetting to set it to 4 spaces in every new session. ;)

src/environments.js
marcianx: `Parser`'s `parseArgument` always returns the list of positions as the last element.
So it returns something of the format `[ParseNode, ParseNode, ..., ParseNode, number[]]`. So this happens any time you see a `defineFunction` with code like
```javascript
    res = parseArray(context.parser, res, "display");
    return res;
```
We need to refactor this at some point, but I'd rather do it right before or during `ParseNode` being ported. It would help to have maximal support from type-safety while doing this since `ParseResult` is simply stored here and used various places (including possible buildHtml, but I haven't read that part of the code base yet).

src/environments.js
marcianx: Ah, that was just to get rid of the `eslint` suppression at the top (personally a fan of maximal linting). However, that's not my favorite lint anyway; and I can revert this if you like.

src/environments.js
marcianx: Agreed.

src/environments.js
marcianx: Fixed.

src/environments.js
marcianx: `ArrayEnvNode` is more descriptive. So I'm using that.
Is `AlignNode` the most appropriate name? It's not a `ParseNode`.

marcianx: Also, per `ArrayEnvNode`, it's not a `ParseNode`, but the data stored in the `ParseNode`. Would you still prefer to call it `ArrayEnvNode`? Or would `ArrayEnvNodeData` (a little long, I know) be more appropriate to avoid confusion?

marcianx: In the currently-pushed version, I went with `AlignSpec` and `ArrayEnvNodeData`. I'm very open to changing per your feedback.

src/environments.js
marcianx: Done.

src/environments.js
marcianx: Done.

src/environments.js
kevinbarabash: Gotcha.  This is fine.  We can re-evaluate lint rules separately.

src/environments.js
kevinbarabash: I see, it's documented in the comment above.

> We need to refactor this at some point, but I'd rather do it right before or during ParseNode being ported.

Sounds good.

src/environments.js
kevinbarabash: I don't see a need to differentiate parse nodes from environment nodes as they end up being part of the same tree at the end of the day.  I think eventually we'll types for each of the different nodes the parser can generate and eventually `ParseNode` will become:
```
ParseNode = FracNode | ExpNode | ... | ArrayNode | AlignNode | ... 
```
Environments are just one type of thing that the parser parses as it does its work, it just happens that the environment parsing code is in a separate file.

marcianx: I'm sorry, I didn't quite understand this comment and the future evolution you intend. What I meant was: for `result: ArrayEnvNodeData`, we have
```javascript
    return new ParseNode(result.type, result, parser.mode);
```
So the `ParseNode` instance contains a _payload_ that is of type `ArrayEnvNodeData` struct, which is why I was thinking about the `Data` suffix.

So did you mean that the payload would become a union type (e.g. `ParseNodeData` or something briefer)?
I'm guessing you didn't mean that the `ParseNode` class would be dissolved in favor of a union type (given that we still want to encapsulate the common tokenizer metadata for error reporting)?

src/environments.js
marcianx: FYI: Looking at `defineFunction.js` I don't currently see any benefit in coalescing.
1. We don't really need any subtype relationship.
2. For object/struct types, there's not really an `extends` type behavior AFAICS (please correct me if I'm wrong), so we would need to repeat the fields anyway when defining it.

src/environments.js
kevinbarabash: I was thinking that eventually we'd have a union type to specify the different types in the tree.  It's possible that `ParseNode` wraps these data nodes and adds additional data like error reporting.  I have spent any time thinking trough the details.  The high-level goal is for nodes with in the parse tree to be more regular and for all them to be statically typed.

marcianx: Ok, so what you're saying is that the _payload_ of `ParseNode` is still considered an undecorated "Node". And hence we should use the `Node` suffix rather than `NodeData`?
Is that your preference?

kevinbarabash:
@marcianx thanks for the PR.  It's exciting to see all of these implicit types being made explicit.

src/environments.js
marcianx: Ah, I see that you merged it. Please LMK what nomenclature you had in mind. I'm happy to update these as we continue in the future.

Issue #863, opened 2017-09-09
Port functions.js to @flow


marcianx:

marcianx:
In the process, it would help to make `defineFunction` types more stringent. There's a lot of `any`, which can make it difficult to reason about what precisely is expected to go into or come out of the functions here.

marcianx:
@kevinbarabash I'm trying to figure out the specific type of the `positions` argument in the `defineFunction`'s `handler`'s `context` argument. We noted this tangentially in the `environment.js`. As mentioned in that PR, `parseArguments` returns a list of positions by pushing it into the args array before returning it. However, this array is populated with `this.pos` _in the `Parser`_. I don't see anywhere where `this.pos` is ever set in `Parser.js`! Here's a pair of greps (the latter's result is empty).
```
$ grep -R '\.pos' src
src/buildHTML.js:        outrow.pos = totalHeight;
src/buildHTML.js:            const shift = row.pos - offset;
src/Lexer.js:        this.pos = 0;
src/Lexer.js:        const pos = this.pos;
src/Lexer.js:        const start = this.pos;
src/Lexer.js:        this.pos += match[0].length;
src/Lexer.js:        const end = this.pos;
src/Parser.js:            return [[this.pos]];
src/Parser.js:        const positions = [this.pos];
src/Parser.js:                    positions.push(this.pos);
src/Parser.js:            positions.push(this.pos);
$ grep -R 'pos:\b' src 
```
I `console.log()`ed `positions` computed in `Parser`'s `parseArguments()` and ran the tests, and it's always an array of `undefined`s.

So, was this meant to be `this.gullet.lexer.pos`? If so, it's reaching really deep into another layer of abstraction.

Attempting to look manually through all code paths was a bit exhausting, but I did another grep to look for actual uses and I found none.
```
$ grep -R positions src
src/environments.js: *  - positions: the positions associated with these arguments from args.
src/environments.js:    positions: number[],
src/environments.js: * List of ParseNodes followed by an array of positions.
src/defineFunction.js://    positions: ?number[],
src/defineFunction.js:    //   - positions: the positions in the overall string of the function
src/Parser.js:                positions: args.pop(),
src/Parser.js:    callFunction(name, args, positions, token) {
src/Parser.js:            positions: positions,
src/Parser.js:     * @return the array of arguments, with the list of positions as last element
src/Parser.js:        const positions = [this.pos];
src/Parser.js:                    positions.push(this.pos);
[...]

PR #864, opened 2017-09-10
Remove the positions array computed by Parser's parseArguments().

marcianx:
This array is computed incorrectly (it's always an array of undefineds) and it's not used anywhere. See #863 for details.

### DIFF
--- a/src/environments.js
+++ b/src/environments.js
@@ -10,27 +10,19 @@ import type {ArgType, Mode, StyleStr} from "./types";
  *  - mode: current parsing mode.
  *  - envName: the name of the environment, one of the listed names.
  *  - parser: the parser object.
- *  - positions: the positions associated with these arguments from args.
  */
 type EnvContext = {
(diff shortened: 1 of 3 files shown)


src/environments.js
```diff
@@ -10,27 +10,19 @@ import type {ArgType, Mode, StyleStr} from "./types";
...
     envName: string,
     parser: Parser,
-    positions: number[],
 };
 
 /**
- * List of ParseNodes followed by an array of positions.
- * Returned by `Parser`'s `parseArguments`.
- */
-type ParseResult = (ParseNode | number[])[] | ParseNode;
```
marcianx: FYI: I'd confused myself in the last PR where I used this as the type: I'd confused the local `parseArray` method with `Parser`'s `parseArguments` method. Sorry about that.

kevinbarabash:
@marcianx thanks for the PR.

PR #865, opened 2017-09-10
Make defineFunction and delimSizing stricter and refactor types in the former for functions.js.

marcianx:
This is toward #863. 

On a parenthetical note, I've actually "done" `functions.js` as well, but tests fail in a manner currently inexplicably to me. I'll look at it with a fresher pair of eyes maybe tomorrow. Switching to ES6 exports and the fact that `functions.js` requires `defineFunctions.js` and not vice versa meant that the `functions` object was being used by `defineFunctions` before being initialized. So I had to move the `functions` map into `defineFunctions.js` to get rid of some basic test errors. But now others still exist.

### DIFF
--- a/src/defineFunction.js
+++ b/src/defineFunction.js
@@ -3,75 +3,82 @@ import functions from "./functions";
 import {groupTypes as htmlGroupTypes} from "./buildHTML";
 import {groupTypes as mathmlGroupTypes} from "./buildMathML";
 
+import type ParseNode from "./ParseNode" ;
 import type Options from "./Options";
 import type {ArgType} from "./types" ;
+import type {Parser} from "./Parser" ;
+import type {Token} from "./Token" ;
 
-type FunctionSpec<T> = {
+/** Context provided to function handlers for error messages. */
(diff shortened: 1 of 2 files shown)


src/defineFunction.js
```diff
@@ -3,92 +3,101 @@ import functions from "./functions";
...
-    //   - lexer: the lexer object
-    // The latter three should only be used to produce error messages.
     //
     // The function should return an object with the following keys:
     //   - type: The type of element that this is. This is then used in
     //          buildHTML/buildMathML to determine which function
     //          should be called to build this node into a DOM node
     // Any other data can be added to the object, which will be passed
     // in to the function in buildHTML/buildMathML as `group.value`.
-    handler: (context: any, args: any) => T,
```
marcianx: Note that the output value of `handler` is actually unrelated to the `group` parameter passed into `htmlBuilder`.

marcianx:
I'm going to close this as it's already being included in and reviewed as part of [PR 867](https://github.com/Khan/KaTeX/pull/867).

PR #866, opened 2017-09-10
Support \operatorname. Resolve #145

### LINKED ISSUE #145: Support \operatorname
We should support this so you can do, e.g., `\operatorname{span}` (see #144). Its implementation is subtler than it may appear:

http://tex.stackexchange.com/a/84308/9

ronkok:


kevinbarabash:
operatorname

LaTeX is in green, KaTeX is in red.  Screenshot looks great!

src/functions/operators.js
```diff
@@ -0,0 +1,98 @@
...
+import defineFunction, {ordargument} from "../defineFunction";
+import buildCommon from "../buildCommon";
+import mathMLTree from "../mathMLTree";
+
+import * as html from "../buildHTML";
+import * as mml from "../buildMathML";
+
+// \operatorname
+// amsopn.dtx: \mathop{#1\kern\z@\operator@font#3}\newmcodes@
+defineFunction({
```
kevinbarabash: Thanks for defining these new functions using the new `defineFunction`.  Let me know if you have any ideas of how to make this new structure better.

src/functions/operators.js
```diff
@@ -0,0 +1,98 @@
...
+
+    mathmlBuilder: (group, options) => {
+        // The steps taken here are similar to the html version.
+        let output = [];
+        if (group.value.value.length > 0) {
+            let temp = mml.buildExpression(group.value.value, options);
+
+            let word = "";
+            for (let i = 0; i < temp.length; i++) {
+                word += temp[i].children[0].text;;
```
kevinbarabash: nit: extra `;`

src/functions/operators.js
```diff
@@ -0,0 +1,98 @@
...
+            }
+            word = word.replace(/\u2212/g, "-");
+            word = word.replace(/\u2217/g, "*");
+            output = [new mathMLTree.TextNode(word)];
+        }
+        const node = new mathMLTree.MathNode("mi", output);
+        node.setAttribute("mathvariant", "normal");
+        return node;
+        // See defineFunction below.
```
kevinbarabash: This comment seems out of place.

src/functions/operators.js
```diff
@@ -0,0 +1,98 @@
...
+    },
+});
+
+defineFunction({
+    type: "functionapply",
+
+    // The \operatorname macro expands as: \opNameMain{#1}\functionapply
```
kevinbarabash: Is this is to get around the fact that `mathmlBuilder` only returns a single node?  I wonder if we could have it return a document fragment containing MathML nodes?

ronkok: Correct. I would be happy to get rid of the macro if I knew how to make `mathmlBuilder`  return multiple nodes.

ronkok: Yes, that works out well. I can eliminate the macro entirely.

PR #867, opened 2017-09-10
Port functions.js to @flow.

### LINKED ISSUE #863: Port functions.js to @flow

marcianx:
Fixes #863. _Requires_ [PR 865](https://github.com/Khan/KaTeX/pull/865) on top of which this sits.


src/defineFunction.js
```diff
@@ -1,94 +1,127 @@
...
+type FunctionSpec = {|
+    numArgs: number,
+    argTypes?: ArgType[],
+    greediness: number,
+    allowedInText: boolean,
+    allowedInMath: boolean,
+    numOptionalArgs: number,
+    infix: boolean,
+    // Must be specified unless it's handled directly in the parser.
+    handler: ?FunctionHandler,
```
kevinbarabash: Thanks for calling this out here.  I think we should try to move the handlers for implicit group functions into the function definition.  I opened #868 for this purpose.

src/defineFunction.js
```diff
@@ -1,94 +1,127 @@
...
+|};
+
+/**
+ * All registered functions. Import this via 'functions.js' since that ensures
+ * that all functions are defined.
+ */
+export const _functions: {[string]: FunctionSpec} = {};
```
kevinbarabash: This should be `functions` since it's a public.

src/functions.js
```diff
@@ -1,24 +1,34 @@
...
+    default as _defineFunction,
+    ordargument,
+    _functions,
+} from "./defineFunction";
 
-// WARNING: New functions should be added to src/functions.
+import type {FunctionPropSpec, FunctionHandler} from "./defineFunction" ;
+
+// WARNING: New functions should be added to src/functions and imported from
+// here.
```
kevinbarabash: `here => there`

src/defineFunction.js
```diff
@@ -1,94 +1,127 @@
...
-    htmlBuilder: (group: T, options: Options) => any,
+    // TODO: Port buildHTML to flow and make the group and return types explicit.
+    htmlBuilder?: (group: *, options: Options) => *,
 
     // This function returns an object representing the MathML structure to be
     // created when rendering the defined LaTeX function.
-    mathmlBuilder: (group: T, options: Options) => any,
-}
+    // TODO: Port buildMathML to flow and make the group and return types explicit.
+    mathmlBuilder?: (group: *, options: Options) => *,
```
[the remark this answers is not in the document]
marcianx: Yep, I mentioned this in [PR 865](https://github.com/Khan/KaTeX/pull/865), though I suppose you're reviewing that part here as well. I'd said that these two types are unrelated, whereas the actual truth is that handler returns a value, where as `htmlBuilder` and `mathmlBuilder` receive a `ParseNode` (IIUC) _containing_ that value; this will flesh out better once `buildHTML` and `buildMathML` are ported and we understand its types better. We are not currently set up to specify this relationship since `ParseNode` is not parameterized by its stored value type.

src/defineFunction.js
```diff
@@ -1,94 +1,127 @@
...
+|};
+
+/**
+ * Final function spec for use at parse time.
+ * This is almost identical to `FunctionPropSpec`, except it
+ * 1. includes the function handler, and
+ * 2. requires all arguments except argTypes.
+ * It is generated by `_defineFunction`.
```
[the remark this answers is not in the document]
marcianx: Oops! I'd copy-pasted that from `functions.js`. Updated.

src/defineFunction.js
marcianx: I just did this since we don't want _anyone_ but `functions.js` to import this. I updated the comment above this to emphasize that fact a bit more.
I could also name this to `importOnlyInFunctionsJs` or `functionsDoNotImportDirectly` or some such to make it clearer. ;)

src/functions.js
marcianx: Sorry, my meaning didn't come out clearly. I changed it to "imported here". The point is that whoever imports `functions.js` should be guaranteed that all definitions are then available; and that's assured only by `functions.js` importing all the definitions.

I am open to other rewordings.

src/defineFunction.js
marcianx: SGTM! That would be much more satisfying.
That being said, spreading out parsing responsibilities a lot can get confusing, so as things become clearer.

src/defineFunction.js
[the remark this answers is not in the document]
kevinbarabash: I see now.  I guess that was wishful thinking on my part.

src/defineFunction.js
kevinbarabash: `functions.js` just exports this same dictionary again and makes it public.  `Parser.js` requires this dictionary.

marcianx: Done.

kevinbarabash: The comment is good, but it doesn't prevent people from access the `functions` map.  In the past we haven't really had issues with people accessing `functions`, but if you're concerned that people might do so in the future then maybe we should take additional steps.  One possibility might be to have functions.js export a `getFunction` method instead of `functions` itself.

marcianx: 1) I'm sorry, did I merge before this was resolved? If so, sorry!
2) I'd very much considered wrapping this in a method, although I was thinking `getFunctions` (which has no benefits over this) as opposed to `getFunction`, which would indeed be better. We could do this as follow-ups as we refactor further.

kevinbarabash: It's my fault for hitting "accept" early.

marcianx: Ah, I'll just let you merge it next time.
Perhaps I have an unreasonable fear of merge conflicts. ;)

PR #875, opened 2017-09-12
refactor defineEnvironment to also accept HTML and MathML builders

### LINKED ISSUE #870: refactor defineEnvironment to look more like the new defineFunction
When defining an environment, we need to specify a handler as well as functions to build appropriate HTML and MathML markup.  Currently these functions are spread over 3 different files.  Ideally, we'd be able to define all functions for an environment (or related environments) in a single file.  These environment definition files should go in a new created `src/environments` folder.  See `defineFunction.js` for details on what the function signature should look like.

kevinbarabash:
fixes #870


src/environments.js
```diff
@@ -1,318 +1,16 @@
...
-    postgap?: number,
+import {_environments} from "./defineEnvironment";
+
+const environments = {
+    has(envName: string) {
+        return _environments.hasOwnProperty(envName);
+    },
+    get(envName: string) {
+        return _environments[envName];
+    },
```
kevinbarabash: @marcianx here's what I did for the wrapper around `_environments`.

src/defineEnvironment.js
```diff
@@ -0,0 +1,117 @@
...
+};
+
+/**
+ * Final enviornment spec for use at parse time.
+ * This is almost identical to `EnvDefSpec`, except it
+ * 1. includes the function handler
+ * 2. requires all arguments except argType
+ * It is generated by `defineEnvironment()` below.
+ */
+type EnvSpec = {|
```
kevinbarabash: I tightened up some of these types by updating them to use exact object types.

kevinbarabash:
I thought we'd end up with separate files for `\*matrix`, `\*cases`, etc., but it turns out they all use the same `htmlBuilder` and `mathmlBuilder` so for now they're all in the same file.  I still think this is the right direction because it groups all the related code in the same file and it also opens the door to eventually expose `defineEnvironment` at some point which wasn't possible the way things were structured before.

src/buildHTML.js
```diff
@@ -581,166 +581,6 @@ groupTypes.genfrac = function(group, options) {
         options);
 };
 
-groupTypes.array = function(group, options) {
```
kevinbarabash: Copied to `const htmlBuilder = function(group, options) { ...` in array.js.

src/environments.js
```diff
@@ -1,326 +1,16 @@
 // @flow
```
kevinbarabash: All the red in this file was copied over to array.js.

src/environments/array.js
```diff
@@ -0,0 +1,475 @@
...
+                        "mtd", [mml.buildGroup(cell, options)]);
+                }));
+        }));
+};
+
+// Arrays are part of LaTeX, defined in lttab.dtx so its documentation
+// is part of the source2e.pdf file of LaTeX2e source documentation.
+// {darray} is an {array} environment where cells are set in \displaystyle,
+// as defined in nccmath.sty.
+defineEnvironment({
```
kevinbarabash: The calls to `defineEnvironment` look a little different from how they looked in environments.js.  All of the existing parameters have been converted props in a single object argument.  This was done so it's easier to know which param is which.

src/environments/array.js
```diff
@@ -0,0 +1,475 @@
...
+        let res = {
+            type: "array",
+            cols: cols,
+            hskipBeforeAndAfter: true, // \@preamble in lttab.dtx
+        };
+        res = parseArray(context.parser, res, dCellStyle(context.envName));
+        return res;
+    },
+    htmlBuilder,
+    mathmlBuilder,
```
kevinbarabash: These are the two "new" parameters.

src/environments.js
kevinbarabash: Everything except for `defineEnvironment` and the type definitions have been copied over to `array.js` while the remaining code has gone to `defineEnvironment.js`.

src/defineEnvironment.js
```diff
@@ -0,0 +1,117 @@
...
+    };
+    for (let i = 0; i < names.length; ++i) {
+        _environments[names[i]] = data;
+    }
+    if (htmlBuilder) {
+        htmlGroupTypes[type] = htmlBuilder;
+    }
+    if (mathmlBuilder) {
+        mathmlGroupTypes[type] = mathmlBuilder;
+    }
```
kevinbarabash: The body of `defineEnvironment` has remained largely the same.  The new code adds registers the `htmlBuilder` and `mathmlBuilder` args for the given `type`.  This was previously being done in `buildHTML.js` and `buildMathML.js` respectively.

src/environments.js
```diff
@@ -1,326 +1,16 @@
...
-
-// Data stored in the ParseNode associated with the environment.
-type AlignSpec = { type: "separator", separator: string } | {
-    type: "align",
-    align: string,
-    pregap?: number,
-    postgap?: number,
+import {_environments} from "./defineEnvironment";
+
+const environments = {
```
rileyjshaw: Cool! Is the motivation behind this abstraction to limit access to the `_environments` object? Or do you expect for the bodies of `has` and `get` to eventually become more complicated?

If this API was added solely for clarity, I don't think `has` and `get` improve readability over `hasOwnProperty` and `[]`.

src/buildHTML.js
rileyjshaw: Thanks for leaving hints like this!

src/environments/array.js
```diff
@@ -0,0 +1,475 @@
...
+ */
+function parseArray(
+    parser: Parser,
+    result: ArrayEnvNodeData,
+    style: StyleStr,
+): ParseNode {
+    let row = [];
+    const body = [row];
+    const rowGaps = [];
+    for (;;) {
```
rileyjshaw: I know this is a copypasta, but `while(true)` would be clearer here.

src/defineEnvironment.js
rileyjshaw: Meta-comment on GitHub reviews vs. Phabricator: I'm finding it frustrating that comments need to pin to a single line.

src/environments/array.js
```diff
@@ -0,0 +1,475 @@
...
+    type: "array",
+    names: [
+        "matrix",
+        "pmatrix",
+        "bmatrix",
+        "Bmatrix",
+        "vmatrix",
+        "Vmatrix",
+    ],
+    props: {},
```
rileyjshaw: Should `props` be optional?

src/defineEnvironment.js
```diff
@@ -0,0 +1,117 @@
...
+
+/**
+ *  - numArgs: (default 0) The number of arguments after the \begin{name} function.
+ *  - argTypes: (optional) Just like for a function
+ *  - allowedInText: (default false) Whether or not the environment is allowed
+ *                   inside text mode (not enforced yet).
+ *  - numOptionalArgs: (default 0) Just like for a function
+ */
+type EnvProps = {
+    numArgs?: number,
```
rileyjshaw: I only see `numArgs` used in this review, but I imagine the rest are used elsewhere / externally?

src/environments.js
kevinbarabash: The motivation was to limit access.  The plan is to do the same for `_functions` and `_symbols`.  Not sure how necessary it is.

src/environments/array.js
kevinbarabash: It was originally `while(true)`, but a previous commit changed it after removing a top level eslint-disable.  I can add it back in with a eslint-disable-line.

src/environments/array.js
kevinbarabash: That's a good question.  `defineFunction` requires `numArgs` on `props` even if it's `0`.  I'm not sure how useful that is, but I'd rather be consistent so I'll make it mandator here too.

src/defineEnvironment.js
kevinbarabash: Good catch.  We may want to add those at some point in the future, but if they're not being used no reason to include them now.

PR #876, opened 2017-09-12
Use mpadded for \raisebox MathML

ronkok:


src/buildMathML.js
```diff
@@ -604,9 +604,13 @@ groupTypes.mclass = function(group, options) {
...
-// Transforms (translation/rotation) don't seem to have a representation
-// in MathML, so just treat them like \text{...}
-groupTypes.raisebox = groupTypes.text;
+groupTypes.raisebox = function(group, options) {
+    const node = new mathMLTree.MathNode(
+        "mpadded", [buildGroup(group.value.body, options)]);
+    const dy = group.value.dy.value.number + group.value.dy.value.unit;
+    node.setAttribute("voffset", dy);
+    return node;
+};
```
kevinbarabash: Can you add a snapshot test for this?

ronkok:
You know, I'm not sure I know how to reliably document MathML outcomes. For what it's worth, here's a screenshot.
raiseboxml

kevinbarabash:
@ronkok see https://github.com/Khan/KaTeX/blob/master/test/mathml-spec.js for examples of snapshot tests.  After you add a test mathml-spec.js that uses `.toMatchSnapshot()`, run `npm test` and `jest` add a new snapshot to `__snapshots__/mathml-spec.js.snap`.

ronkok:
> Can you add a snapshot test for this?

It turns out that the answer to that question is no, I can't. My attempt to `npm test` failed on the `make lint flow test`.

kevinbarabash:
> It turns out that the answer to that question is no, I can't. My attempt to npm test failed on the make lint flow test.

I can send you a PR for the snapshot test.  Can you post the console output of the failure from `npm test`?

ronkok:
Output as requested: 
>Windows PowerShell
>Copyright (C) 2016 Microsoft Corporation. All rights reserved.
>
>PS C:\Users\ronko> cd C:\Users\ronko\OneDrive\Documents\GitHub\KaTeX
>PS C:\Users\ronko\OneDrive\Documents\GitHub\KaTeX> npm test
>
> katex@0.8.4-pre test C:\Users\ronko\OneDrive\Documents\GitHub\KaTeX
> make lint flow test
>
>make i info Invoking lint target
(node:8744) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 1): undefined
>make × ERR  not found: make-$(NIS)

kevinbarabash:
@ronkok thanks.  It looks like `make` isn't found.  Does that sounds right?  I think we should be able to modify the scripts in `package.json` to run commands in `node_modules/.bin` directly.  The benefit of the Makefile at the moment is that it ensures that dependencies are up to date.  I wonder if there's a way to do without using `make`.

ronkok:
I think that's right. `make` is not, I believe, part of Windows. There are ways to get equivalents running in Windows and no doubt someone like @edemaine knows all about them. The problem here is that you are dealing with an amateur.

Go ahead and search for a `make` alternative if that is something with long term value. Don't do that on my account. I have 3 or 4 small PRs left to submit and then my work here will be done.

ronkok:
One small step forward.

kevinbarabash:
@ronkok thanks for PR and for working through difficulties generating the snapshot.

PR #882, opened 2017-09-14
Add read/write file cli parameters

raoulb:
Read from and write to file paths given on command line.


kevinbarabash:
@raoulb I haven't be able to actually run `cli.js`.  How are you running it and what version of node are you using?

raoulb:
@kevinbarabash I installed the latest version of katex with `npm` and edited the file. I failed to run the `cli.js` from within the git clone on Debian `sid`.

raoulb:
[clabot:check]

kevinbarabash:
@raoulb I'll try doing `npm install` again with the most recent version and will try applying your changes.  It would be nice to if we could run `cli.js` from git clone so that it's easier to test changes.  I've opened #894 to track this.

raoulb:
@kevinbarabash Yes, running `cli.js` from git would help. (I'm not a JS developer and new to node, hence assumed I'm doing it the wrong way.) I see the value of this tool as a filter for static site generators that want to render math on site build time.

cli.js
```diff
@@ -4,29 +4,95 @@
...
 
     console.log("\n" +
                 "Options:");
-    console.log("  --help            Display this help message");
-    console.log("  --display-mode    Render in display mode (not inline mode)");
+    console.log("  --help              Display this help message");
+    console.log("  --display-mode      Render in display mode (not inline mode)");
+    console.log("  --no-throw-on-error Render unsupported commands as text");
+    console.log("  --error-color       Color unsupported commands are rendered in");
+    console.log("  --max-size          User-specified sizes are capped to maxSize");
```
kevinbarabash: Should specify what units the cap is in.

cli.js
```diff
@@ -4,29 +4,95 @@
...
+               );
 
     console.log("\n" +
                 "Options:");
-    console.log("  --help            Display this help message");
-    console.log("  --display-mode    Render in display mode (not inline mode)");
+    console.log("  --help              Display this help message");
+    console.log("  --display-mode      Render in display mode (not inline mode)");
+    console.log("  --no-throw-on-error Render unsupported commands as text");
+    console.log("  --error-color       Color unsupported commands are rendered in");
```
kevinbarabash: The help message should describe what format the color must be in and give an example.

cli.js
```diff
@@ -4,29 +4,95 @@
...
+    console.log("  --max-size          User-specified sizes are capped to maxSize");
+    console.log("  --input file        Read from given file");
+    console.log("  --output file       Write to given file");
     process.exit();
 }
 
-process.stdin.on("data", function(chunk) {
-    input += chunk.toString();
-});
+function getOptions(args) {
```
kevinbarabash: If you have time, would you be able switch the option parsing code to use `nomnom`.  We use this in a couple of other places in the project, e.g. https://github.com/Khan/KaTeX/blob/master/dockers/Screenshotter/screenshotter.js.

cli.js
```diff
@@ -4,29 +4,95 @@
...
+            input += chunk.toString();
+        });
+
+        process.stdin.on("end", function() {
+            writeOutput(input);
+        });
+    }
+}
+
+function writeOutput(input) {
```
kevinbarabash: It would be good to include an trailing `\n` in the output.

cli.js
raoulb: Initially I had all that, but then the line length was larger than 84 characters, so I dropped it.
If we readd the information, then the description must span multiple lines.

cli.js
raoulb: Same as above and I agree.

cli.js
raoulb: I did not want to introduce a new dependency. Using a proper option parsing library will make things much easier.

cli.js
raoulb: Ok

cli.js
raoulb: Done, but not sure if this is better.

cli.js
raoulb: Done

raoulb: Hmm: nomnom is deprecated: https://github.com/harthur/nomnom

kevinbarabash: Thanks for switching to nomnom.  That's too bad about it's deprecation.  I created #900 to switch from nomnom to commander.js everywhere.  It's not super pressing b/c nomnom does the job.

cli.js
kevinbarabash: Whenever we upgrade the option parser it would be nice if we could find one which supported short and long help strings and nicely formatted the longs ones like descriptions underneath the short one and properly aligned them.  Maybe commander.js already does, but it doesn't show in the given examples.

kevinbarabash:
@raoulb nice changes.  The abbreviations are a nice touch.  For the macros, would the following work:
```
cli.js --input test.tex --macros "\\foo": "foo", "\\bar": "bar"
```
where test.tex contains `\foo\bar`?

raoulb:
@kevinbarabash Thanks. For the macros, your version would not work with the 'list' property of the '-m' option. We have to write:
`
-m '\a:\sqrt{2}' -m '\b:\frac{3}{5}'
`
One could however change to option parsing to split up something like:
`
-m '\a:\sqrt{2} \b:\frac{3}{5}'
`
where we need an arbitrary separator character between the individual macros.

I prefer to write multiple '-m' and I think I'll add an option for reading a list of macros from a text file.

kevinbarabash:
@raoulb I saw your most recent change.  Is that b/c the macro itself can have a `:` in it?

raoulb:
@kevinbarabash Exactly, a macro could be:
```
\index : [G : H]
```
as in [https://en.wikipedia.org/wiki/Index_of_a_subgroup](https://en.wikipedia.org/wiki/Index_of_a_subgroup)

The character ":" was an ad-hoc choice, chosen because it often has the semantics of assignment (f.e. in dicts / hash tables). It should be pretty save as macro names can not contain special characters, see:

* [http://www.tex.ac.uk/FAQ-whatmacros.html](http://www.tex.ac.uk/FAQ-whatmacros.html)
* [http://www.tex.ac.uk/FAQ-linmacnames.html](http://www.tex.ac.uk/FAQ-linmacnames.html)

Probably one could break it with something similar to `makeatletter`.

raoulb:
Ok, I think I'm done with adding features here.
Let me know what needs to be done next.
If you want some test cases, please provide some examples.

kevinbarabash:
> If you want some test cases, please provide some examples.

@raoulb I'm going to do some manual testing of the recent changes and will merge it if things look good.  It would be great to have some automated test cases, but I feel like that would require us being able to run cli.js from the git repo.  If you have time, a followup PR would be greatly appreciated.

raoulb:
I wonder what the review status of this PR is?

kevinbarabash:
@raoulb sorry, got busy with other things.  I'll have another look at this tomorrow.

PR #884, opened 2017-09-15
Add support for comments, fixes #20

### LINKED ISSUE #20: Add support for % comments

kevinbarabash:
As part of this PR, I also updated the local test page to use a `<textarea>` so that you can enter newlines.

screen shot 2017-09-15 at 12 32 45 am

### DIFF
--- a/src/Lexer.js
+++ b/src/Lexer.js
@@ -19,6 +19,7 @@ import {LexerInterface, Token} from "./Token";
 
 /* The following tokenRegex
  * - matches typical whitespace (but not NBSP etc.) using its first group
+ * - matches comments (must have trailing newlines)
  * - does not match any control character \x00-\x1f except whitespace
  * - does not match a bare backslash
  * - matches any ASCII character except those just mentioned
@@ -32,15 +33,19 @@ import {LexerInterface, Token} from "./Token";
  * If there is no matching function or symbol definition, the Parser will
  * still reject the input.
  */
+const commentRegexString = "%[^\n]*[\n]";
 const tokenRegex = new RegExp(
     "([ \r\n\t]+)|" +                                 // whitespace
-    "([!-\\[\\]-\u2027\u202A-\uD7FF\uF900-\uFFFF]" +  // single codepoint
+    `(${commentRegexString}|` +                       // comments
+    "[!-\\[\\]-\u2027\u202A-\uD7FF\uF900-\uFFFF]" +   // single codepoint
     "|[\uD800-\uDBFF][\uDC00-\uDFFF]" +               // surrogate pair
     "|\\\\verb\\*([^]).*?\\3" +                       // \verb*
     "|\\\\verb([^*a-zA-Z]).*?\\4" +                   // \verb unstarred
     "|\\\\(?:[a-zA-Z@]+|[^\uD800-\uDFFF])" +          // function name
     ")"
 );
+
+const commentRegex = new RegExp(commentRegexString);
 
 /** Main Lexer class */
 export default class Lexer implements LexerInterface {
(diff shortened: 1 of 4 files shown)


test/__snapshots__/katex-spec.js.snap
```diff
@@ -0,0 +1,11 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`A comment parser should not include comments in the output 1`] = `
```
kevinbarabash: I should include the input string in the title so that it's easy to see what input produce this parse tree.

edemaine:
@kevinbarabash `<textarea>` would be very handy when testing arrays and such, thanks!

You should test spacing behavior:
```latex
\text{hello% comment
world}
```
should parse like `\text{helloworld}`, but not like
```latex
\text{hello% comment

world}
```

If I understand line 67 of `src/Lexer.js`, the comment gets converted into a space.  This is not what we want.  Comments are used in particular to hide newlines from the bowels of TeX.

src/Lexer.js
```diff
@@ -68,6 +69,10 @@ export default class Lexer implements LexerInterface {
         const start = this.pos;
         this.pos += match[0].length;
         const end = this.pos;
-        return new Token(text, start, end, this);
+        if (/%[^\n]*[\n]/.test(text)) {
```
xymostech: Is there a way to factor out this regex so we don't have to duplicate it here and above? Or maybe we could just test for a single `^%` so it doesn't have to be duplicated?

test/katex-spec.js
```diff
@@ -1484,6 +1484,35 @@ describe("A font parser", function() {
...
+    it("should parse comments at the start of a line", () => {
+        expect("% comment\n").toParse();
+    });
+
+    it("should parse multiple lines of comments in a row", () => {
+        expect("% comment 1\n% comment 2\n").toParse();
+    });
+
+    it("should not parse a comment that isn't followed by a newline", () => {
+        expect("x%y").toNotParse();
```
xymostech: This seems a little strange to me. Maybe we could support comments that go to the end of the input string as well as ones that end with a newline? I feel like this will catch people up otherwise.

xymostech:
Oh, I had another comment separate from this. MathJax parses `\text{x % y}` incorrectly, producing the text `x % y` instead of erroring out because we don't close `\text`. Not sure what to do about that, but it's a discrepency to keep note of.

kevinbarabash:
@xymostech thanks for the review.  Maybe a MathJax migration guide or at the very least a "Difference from MathJax" section in the README.md would help with this common differences.

kevinbarabash:
I'll make the requested changes before merging this.

test/katex-spec.js
kevinbarabash: There was recent discussion in #20 about this.  If we allow comments without the newline then some KaTeX strings will break when copied over to a LaTeX document.

src/Lexer.js
```diff
@@ -32,16 +33,20 @@ import {LexerInterface, Token} from "./Token";
  * If there is no matching function or symbol definition, the Parser will
  * still reject the input.
  */
+const commentRegexString = "%[^\n]*[\n]";
```
marcianx: I know this is already submitted, but can the author or any of the reviewers explain why the comment character eats all subsequent blank lines instead of just the first end-of-line? That's not like LaTeX, is it?

marcianx: Ah, sorry, I just completely misread this as `%(^\n)*[\n]`. Sorry. This is just consuming non-line breaks. Ignore me. ;)

Issue #516, opened 2016-08-03
support sizing commands within optional params


kevinbarabash:
e.g. `h(x)=\sqrt[\Large3]{2x-15}`

kevinbarabash:
I think I have this figured out:
<img width="860" alt="screen shot 2017-09-15 at 1 27 47 am" src="https://user-images.githubusercontent.com/1044413/30467769-205457ce-99b5-11e7-9953-743d2e1485a1.png">

`\small` looks a little better:
<img width="855" alt="screen shot 2017-09-15 at 1 30 37 am" src="https://user-images.githubusercontent.com/1044413/30467847-8ca9f528-99b5-11e7-8e95-7ee967b6560e.png">

Without sizing:
<img width="668" alt="screen shot 2017-09-15 at 1 30 46 am" src="https://user-images.githubusercontent.com/1044413/30467857-9d231dd0-99b5-11e7-8b49-74fede62c4dd.png">

kevinbarabash:
@camsc the behavior's a bit different from MathJax's behavior.

kevinbarabash:
This has been done for a while.

PR #885, opened 2017-09-15
allow sizing commands inside optional groups

### LINKED ISSUE #650: support size commands inside optional params
```
g(x)=2\sqrt[\large 3]{x}
```
results in the following error:
```
KaTeX parse error: Expected ']', got 'EOF' at end of input: …rt[\large 3]{x}
```

kevinbarabash:
Fixes #650 and #516.

TODO:
- [x] unit tests
- [x] screenshot tests
- [x] \color 
- [x] style commands e.g. \textstyle
- [x] old font functions, e.g. \tt


src/Parser.js
```diff
@@ -165,7 +165,8 @@ class Parser {
             if (breakOnInfix && functions[lex.text] && functions[lex.text].infix) {
                 break;
             }
-            const atom = this.parseAtom();
+            const optional = breakOnTokenText === "]";
+            const atom = this.parseAtom(optional);
```
edemaine: I haven't thought about this deeply, but is there a reason to just pass in `]` and not just `breakOnTokenText` in general?  I think the latter might be useful for more general macro support in the future... Maybe worth reviewing what other values this currently takes on in the code, if any.

kevinbarabash: Good idea.  I'll look into this evening.

test/__snapshots__/katex-spec.js.snap
```diff
@@ -0,0 +1,44 @@
...
+      "index": ParseNode {
+        "mode": "math",
+        "type": "ordgroup",
+        "value": Array [
+          ParseNode {
+            "mode": "math",
+            "type": "sizing",
+            "value": Object {
+              "size": 5,
+              "value": Array [
```
kevinbarabash: This is kind of weird.  I hope that in the future that `sizing` nodes could maybe look like:
```
ParseNode: {
   mode: "math",
   type: "sizing",
   size: 5,
   body: Array [ ... ],
}
```

kevinbarabash:
I'm going to update the jsDocs to include the new `breakOnTokenText`.  @marcianx would you be able to review this PR?

marcianx:
Will do. I'll do one pass now. And if there are any changes. Then one in the morning.

src/Parser.js
```diff
@@ -165,7 +165,8 @@ class Parser {
             if (breakOnInfix && functions[lex.text] && functions[lex.text].infix) {
                 break;
             }
-            const atom = this.parseAtom();
+            // const optional = breakOnTokenText === "]";
```
marcianx: Should this now be removed?

src/Parser.js
```diff
@@ -308,10 +309,10 @@ class Parser {
      *
      * @return {?ParseNode}
      */
-    parseAtom() {
+    parseAtom(breakOnTokenText) {
```
marcianx: nit: Mind adding a `@param` documenting the type of the new parameter?

src/Parser.js
```diff
@@ -425,7 +426,7 @@ class Parser {
      *
      * @return {?ParseNode}
      */
-    parseImplicitGroup() {
+    parseImplicitGroup(breakOnTokenText) {
```
marcianx: Same here, please.

test/__snapshots__/katex-spec.js.snap
```diff
@@ -0,0 +1,44 @@
...
+                  "mode": "math",
+                  "type": "textord",
+                  "value": "3",
+                },
+              ],
+            },
+          },
+        ],
+      },
+      "type": "sqrt",
```
marcianx: (optional) I don't know whether an arbitrary strict order is enforced here, but if not, moving this up would make it much easier to see which level it is in.

src/Parser.js
```diff
@@ -308,10 +309,10 @@ class Parser {
      *
      * @return {?ParseNode}
      */
-    parseAtom() {
+    parseAtom(breakOnTokenText) {
         // The body of an atom is an implicit group, so that things like
         // \left(x\right)^2 work correctly.
-        const base = this.parseImplicitGroup();
+        const base = this.parseImplicitGroup(breakOnTokenText);
```
marcianx: It might be worth adding a comment somewhere as to why it's okay not to pass this parameter down to other methods below. The reason seems to be subtle: as `handleSupSubscript` can call `parseGroup` which can call `parseExpression`, but `parseExpression` is only called in `parseGroup` only if a `{` or `[` is encountered, opening a deeper layer. But I haven't yet done an exhaustive check.

I'm also fine with completely deferring this if we intend to formalize the various non-obvious invariants in this file some point to aid comprehensibility so that they aren't violated in the future.

test/__snapshots__/katex-spec.js.snap
kevinbarabash: I can set up a custom formatter for the snapshots to always put `"type"` first.

src/Parser.js
kevinbarabash: I'll add a comment.  `handleSupSubscript` should always be parsing non-optional groups afaict because `x^[2]` isn't valid (or rather it's valid it's treated as `x^{[}2]`).

src/Parser.js
```diff
@@ -834,7 +836,7 @@ export default class Parser {
         if (this.nextToken.text === (optional ? "[" : "{")) {
             // If we get a brace, parse an expression
             this.consume();
-            const expression = this.parseExpression(false, optional ? "]" : null);
+            const expression = this.parseExpression(false, optional ? "]" : "}");
```
kevinbarabash: I'm not sure why this was `null`.

edemaine: Parsing always ends on `}`, so I'm not sure specifying `"}"` is necessary... hence the `null`.  (It's been a while since I've looked at this code, so I forget the exact details...)

kevinbarabash: It feels nice to have symmetry the `? "[" : "{"` check above.

src/Parser.js
```diff
@@ -527,7 +529,7 @@ export default class Parser {
         } else if (utils.contains(Parser.sizeFuncs, func)) {
             // If we see a sizing function, parse out the implicit body
             this.consumeSpaces();
-            const body = this.parseExpression(false);
+            const body = this.parseExpression(false, breakOnTokenText);
```
marcianx: Why does this only apply to this block and not to the rest below: `styleFuncs`, `oldFontFuncs`, and `\color`? Are they not similarly affected? Or do none of our functions currently want to allow these in optional groups?
Otherwise, this PR mostly LGTM.

kevinbarabash: Good question. I was fixing only the issue, but I'm sure there are places where KA also uses `\color` inside of optional groups (apparently LaTeX supports this too).  LaTeX also supports style functions within optional blocks too.

kevinbarabash:
@marcianx feel free to merge PRs as well.

PR #886, opened 2017-09-15
Support \colorbox and \fcolorbox

ronkok:
These are functions from the `color` package. They accept text, not math. They also have padding similar to `\fbox`.

Because of the padding, the code in `buildHTML` is intermixed with the `\fbox` code. I have not (yet) created a new file in the functions folder. This way, the reviewer gets a cleaner diff.


src/buildHTML.js
```diff
@@ -1533,43 +1533,63 @@ groupTypes.accentUnder = function(group, options) {
...
-        const isCharBox = (isCharacterBox(group.value.body));
+        let vertPad = 0;
         // ref: LaTeX source2e: \fboxsep = 3pt;  \fboxrule = .4pt
         // ref: cancel package: \advance\totalheight2\p@ % "+2"
-        pad = (label === "fbox" ? 0.34 : (isCharBox ? 0.2 : 0));
-        imgShift = inner.depth + pad;
+        if (isColorbox || label === "fbox") {
+            vertPad = (label === "colorbox" ? 0.3 : 0.34);
+        } else {
+            vertPad = (isCharacterBox(group.value.body) ? 0.2 : 0);
```
kevinbarabash: The parens around the ternaries aren't necessary.

src/buildHTML.js
```diff
@@ -1533,43 +1533,63 @@ groupTypes.accentUnder = function(group, options) {
...
+        inner.classes.push(/cancel/.test(label) ? "cancel-pad" : "boxpad");
 
         // Add vertical padding
-        const isCharBox = (isCharacterBox(group.value.body));
+        let vertPad = 0;
         // ref: LaTeX source2e: \fboxsep = 3pt;  \fboxrule = .4pt
         // ref: cancel package: \advance\totalheight2\p@ % "+2"
-        pad = (label === "fbox" ? 0.34 : (isCharBox ? 0.2 : 0));
-        imgShift = inner.depth + pad;
+        if (isColorbox || label === "fbox") {
```
kevinbarabash: This could be `/box/.test(label)`.

src/functions.js
```diff
@@ -88,6 +88,42 @@ defineFunction(["\\color"], {
...
+    allowedInText: true,
+    greediness: 3,
+    argTypes: ["color", "text"],
+}, function(context, args) {
+    const color = args[0];
+    const body = args[1];
+    return {
+        type: "enclose",
+        label: context.funcName,
+        bkgrndColor: color,
```
kevinbarabash: I have a slight preference for `backgroundColor` mainly b/c I type it a lot so it easier to type.

src/stretchy.js
```diff
@@ -254,11 +254,13 @@ const encloseSpan = function(inner, label, pad, options) {
     let img;
     const totalHeight = inner.height + inner.depth + 2 * pad;
 
-    if (label === "fbox") {
+    if (/(fbox)|(color)/.test(label)) {
         img = buildCommon.makeSpan(["stretchy", label], [], options);
-        if (options.color) {
+
+        if (label === "fbox" && options.color) {
             img.style.borderColor = options.getColor();
```
kevinbarabash: Why do we use `options` for the color whereas in `buildHTML.js` we use `borderColor` and `bkgrndColor` directly from the group?

src/buildHTML.js
```diff
@@ -1533,43 +1533,63 @@ groupTypes.accentUnder = function(group, options) {
...
+    } else {
+        vlist = buildCommon.makeVList([
+            // Write the \cancel stroke on top of inner.
+            {type: "elem", elem: inner, shift: 0},
+            {type: "elem", elem: img, shift: imgShift},
+        ], "individualShift", null, options);
     }
 
     if (/cancel/.test(label)) {
+        vlist.children[0].children[0].children[1].classes.push("svg-align");
```
kevinbarabash: Why was this added?

ronkok: This line is not *added* in this PR. It just moved a bit.

src/stretchy.js
ronkok: The `fbox` border takes the current text color. The `options.getColor()` will do that successfully even if the color was defined using the KA one argument custom colors.  See PR #792.

I don't think we want to support behavior that for `\colorbox` and `\fcolorbox`. I think these two proposed functions should get only the colors specified in the arguments to the function

kevinbarabash:
@ronkok thanks for the PR.  The tests look great!

PR #887, opened 2017-09-17
Implement correct macros for liminf and limsup, fixes #111

### LINKED ISSUE #111: \limsup, etc should have a small space in the middle

kevinbarabash:
TODO:
- [x] fix test failures
- [x] add unit tests
- [x] add screenshot tests

screen shot 2017-09-16 at 10 57 08 pm


kevinbarabash:
Here's the LaTeX output:
screen shot 2017-09-16 at 11 00 59 pm

katex.js
```diff
@@ -24,6 +24,7 @@ let render = function(expression, baseNode, options) {
     const settings = new Settings(options);
 
     const tree = parseTree(expression, settings);
+    console.log(tree);
```
kevinbarabash: remove

static/main.js
```diff
@@ -21,7 +21,7 @@ function init() {
     }
 
     var macros = {};
-    var options = {};
+    var options = {displayMode: true};
```
kevinbarabash: Add a toggle for this.

src/functions/operators.js
```diff
@@ -30,25 +30,33 @@ defineFunction({
             let mode = "";
 
             // Consolidate Greek letter function names into symbol characters.
-            const temp = html.buildExpression(group.value.value, options, true);
+            const temp = html.buildExpression(
+                group.value.value, options.withFont("mathrm"), true);
```
kevinbarabash: Do the same thing for MathML.

src/Parser.js
```diff
@@ -165,7 +165,7 @@ class Parser {
             if (breakOnInfix && functions[lex.text] && functions[lex.text].infix) {
                 break;
             }
-            const atom = this.parseAtom();
+            const atom = this.parseAtom(body);
```
kevinbarabash: This is maybe a little hacky, but some commands need to know what was the previous atom in the current group.

src/functions.js
```diff
@@ -365,6 +364,28 @@ defineFunction(["\\mathop"], {
     };
 });
 
+defineFunction(["\\limits"], {
```
kevinbarabash: I decided to implement limits (and nolimits) so that they could be used elsewhere.  According to [latex.ltx](https://www.tug.org/svn/texlive/trunk/Master/texmf-dist/tex/latex/base/latex.ltx?view=co) it looks `\mathop` defaults to enabling limits.  Maybe we should do the same.

src/functions/operators.js
```diff
@@ -59,7 +67,16 @@ defineFunction({
...
-                word += temp[i].children[0].text;
+                if (temp[i].type === "mi") {
+                    word += temp[i].children[0].text;
+                } else if (temp[i].type === "mspace") {
+                    // TODO(kevinb) differentiate space width
+                    word += " ";
+                } else {
+                    const pos = temp[i].lexer.pos;
+                    throw new Error(
+                        `unrecognized group in \\operatorname at ${pos}`);
```
kevinbarabash: Does it make sense to raise a `ParseError` inside a MathML builder?  It's not really a parser error, it's just something we don't know how to deal with.

src/functions/operators.js
kevinbarabash: Done.

src/macros.js
```diff
@@ -298,3 +298,6 @@ defineMacro("\\simcoloncolon", "\\sim\\mathrel{\\mkern-1.2mu}\\dblcolon");
 defineMacro("\\approxcolon", "\\approx\\mathrel{\\mkern-1.2mu}\\vcentcolon");
 defineMacro("\\approxcoloncolon",
             "\\approx\\mathrel{\\mkern-1.2mu}\\dblcolon");
+
+defineMacro("\\limsup", "\\DOTSB\\mathop{\\operatorname{lim\\,sup}}\\limits");
+defineMacro("\\liminf", "\\DOTSB\\mathop{\\operatorname{lim\\,inf}}\\limits");
```
kevinbarabash: [latex.ltx](https://www.tug.org/svn/texlive/trunk/Master/texmf-dist/tex/latex/base/latex.ltx?view=co) seems to indicate that `\mathop` has limits defined by default.  Perhaps that's something we should consider adopting.

src/functions/operators.js
```diff
@@ -30,36 +31,56 @@ defineFunction({
             let mode = "";
 
             // Consolidate Greek letter function names into symbol characters.
-            const temp = html.buildExpression(group.value.value, options, true);
+            const temp = html.buildExpression(
+                group.value.value, options.withFont("mathrm"), true);
```
ronkok: That space on line 76 of the mathBuilder section should be a \u2006.

src/functions/operators.js
```diff
@@ -30,36 +31,56 @@ defineFunction({
...
+                group.value.value, options.withFont("mathrm"));
 
             let word = "";
             for (let i = 0; i < temp.length; i++) {
-                word += temp[i].children[0].text;
+                if (utils.contains(["mi", "mn", "mo"], temp[i].type)) {
+                    word += temp[i].children[0].text;
+                } else if (temp[i].type === "mspace") {
+                    // TODO(kevinb) differentiate space width
+                    word += " ";
```
ronkok: That space should be a \u2006.

test/__snapshots__/mathml-spec.js.snap
```diff
@@ -74,6 +74,41 @@ exports[`A MathML builder should make prime operators into <mo> nodes 1`] = `
 
 `;
 
+exports[`A MathML builder should output \\limsup_{x \\rightarrow \\infty} correctly 1`] = `
+
+<math>
+  <semantics>
+    <mrow>
+      <msub>
```
kevinbarabash: I think this should actually be `munder`.

kevinbarabash: Actually this is right b/c the tests in mathml-spec.js don't use display mode by default.

test/__snapshots__/mathml-spec.js.snap
```diff
@@ -74,6 +74,76 @@ exports[`A MathML builder should make prime operators into <mo> nodes 1`] = `
...
+</math>
+
+`;
+
+exports[`A MathML builder should output \\limsup_{x \\rightarrow \\infty} in displaymode correctly 1`] = `
+
+<math>
+  <semantics>
+    <mrow>
+      <munder>
```
kevinbarabash: Now we're getting `munder` for limits with subscripts only, but we have to be in display mode for this.

src/macros.js
kevinbarabash: TODO: remove `\DOTSB`

edemaine: Why remove `\DOTSB`? Maybe I'm missing some context -- if this is in the definition in AMS LaTeX, it makes sense to keep it here, so preceding `\dots` behave correctly.

kevinbarabash: Hmm... maybe I saw the `\DOTSB` when I was inspecting commands using pidflatex as described in https://github.com/Khan/KaTeX/wiki/Examining-TeX.

edemaine: Hmm, I checked the source in `amsopn.sty`, and uses of `\DOTSB` in `amsmath.sty`, and checked in pdflatex... and I now agree that `\DOTSB` should be removed. 😄

src/functions/operators.js
```diff
@@ -30,36 +31,60 @@ defineFunction({
...
+                    word += temp[i].children[0].text;
+                } else if (temp[i].type === "mspace") {
+                    if (group.value.value[i].value === "\\,") {
+                        word += "\u2006";
+                    } else {
+                        // TODO(kevinb) differentiate based on space width
+                        word += " ";
+                    }
+                } else {
+                    if (temp[i].lexer) {
```
marcianx: Why does this not throw with a more generic message when lexer is not set?

src/functions/operators.js
```diff
@@ -30,36 +31,60 @@ defineFunction({
...
             let word = "";
             for (let i = 0; i < temp.length; i++) {
-                word += temp[i].children[0].text;
+                if (utils.contains(["mi", "mn", "mo"], temp[i].type)) {
+                    word += temp[i].children[0].text;
+                } else if (temp[i].type === "mspace") {
+                    if (group.value.value[i].value === "\\,") {
+                        word += "\u2006";
+                    } else {
+                        // TODO(kevinb) differentiate based on space width
```
marcianx: Does it make sense to add an issue for this for others to pick up and add the issue number here? Or is this something you expect to address immediately?

static/main.js
marcianx: Sorry, I'm a little unclear. Is this a TODO for this PR? Or something longer term?

test/__snapshots__/mathml-spec.js.snap
marcianx: Is it worth saying `in \textstyle` in the `exports` key above to avoid this confusion in the future?

src/functions/operators.js
```diff
@@ -30,36 +31,60 @@ defineFunction({
...
         let output = [];
         if (group.value.value.length > 0) {
-            const temp = mml.buildExpression(group.value.value, options);
+            const temp = mml.buildExpression(
+                group.value.value, options.withFont("mathrm"));
 
             let word = "";
             for (let i = 0; i < temp.length; i++) {
-                word += temp[i].children[0].text;
+                if (utils.contains(["mi", "mn", "mo"], temp[i].type)) {
```
marcianx: 1. This is not backward-compatible.
2. It's not obvious to me that these are the only allowable possibilities inside an `operatorname`, and not, say, `menclose` `mrow` or `mpadded` or any of the stuff on the left sidebar [here](https://developer.mozilla.org/en-US/docs/Web/MathML/Element/mpadded). I see examples of sub/superscripts within `operatorname` [here](https://tex.stackexchange.com/a/100064) and [here](https://tex.stackexchange.com/q/67949) among others.

kevinbarabash: Good point.  I'll try to make this more general.

src/functions/operators.js
kevinbarabash: Good point.  I more generic error would make more sense here.

static/main.js
kevinbarabash: I think it's okay to do it as a separate commit.  I'll open an issue once this lands.

kevinbarabash:
@marcianx I'm on vacation now so I have time to pick this up.

marcianx:
Please go ahead. :)
I've had an uncharacteristically busy couple of weeks. ;)

kevinbarabash:
@marcianx I rebased.  I'll start addressing the issues that you raised.

kevinbarabash:
@marcianx I think I've addressed the issues you raised.  Could you have another look at this when you have some time?

marcianx:
Will do.

src/functions/operatorname.js
```diff
@@ -30,37 +30,44 @@ defineFunction({
             let mode = "";
 
             // Consolidate Greek letter function names into symbol characters.
-            const temp = html.buildExpression(group.value.value, options, true);
+            const temp = html.buildExpression(
+                group.value.value, options.withFontFamily("mathrm"), true);
 
             // All we want from temp are the letters. With them, we'll
             // create a text operator similar to \tan or \cos.
             for (let i = 0; i < temp.length; i++) {
```
marcianx: (optional) You could change this to `for (const child of temp) {` if we have a preference for that. It makes it more up front that the index is not used anywhere else.

src/functions/operatorname.js
```diff
@@ -30,37 +30,44 @@ defineFunction({
...
+                    mode = (/[\u0391-\u03D7]/.test(letter) ? "math" : "text");
+                    output.push(buildCommon.mathsym(letter, mode));
+                } else {
+                    output.push(child);
+                }
             }
         }
-        return buildCommon.makeSpan(["mop"], output, options);
+        const span = buildCommon.makeSpan(["mop"], output, options);
+        return span;
```
marcianx: (optional) Any particular reason this was split into 2 lines? For debugging perhaps?

Issue #883, opened 2017-09-15
\sqrt doesn't scale properly when browser is zoomed in or out on Safari


kevinbarabash:
`\sqrt{\frac{1}{x}}` produces the following output when zoom in:
<img width="261" alt="screen shot 2017-09-15 at 12 07 12 am" src="https://user-images.githubusercontent.com/1044413/30466220-53afcbf4-99aa-11e7-9c85-ddae5dc31db7.png">
and out:
<img width="89" alt="screen shot 2017-09-15 at 12 07 18 am" src="https://user-images.githubusercontent.com/1044413/30466217-4ec1cd90-99aa-11e7-897e-efea30db24c9.png">

kevinbarabash:
It works fine in Chrome and Firefox.

ronkok:
I'll look into it.

ronkok:
It seems to be a problem with all the nested SVGs. The non-nested SVGs (`\cancel`, `widehat`, etc) seem to be working okay.  But in nested SVGs, the inner SVG is scaling taller than the outer SVG. The bottom of each one is being cut off and what can be seen is too large. I'll keep looking and report as I learn more.

ronkok:
I think it probably is possible to re-write all the nested SVGs and instead use two or three single-level SVGs placed side by side. The variable-length arrow shafts will require CSS `calc(...)` if this scheme is to work. 

During very early drafts of PR #670, I tried this approach and had it working, sort of. I ended up using nested SVGs because I was getting better quality images. Maybe the side-by-side, CSS-controlled approach can succeed if I work it a little smarter.

This will take time. I'll make a progress report in a couple of days. I sure wish that Safari was better.

kevinbarabash:
I'll try this out in the nightly webkit build and see if there's any info on the status of this bug anywhere.  If the bug's being fixed in Safari, no need to rework things.

ronkok:
@kevinbarabash PR #890 fixes `sqrt` and single-ended arrows, but it does not fix double-ended arrows or horizontal braces. It may be a couple of weeks before I can address them.

kevinbarabash:
@ronkok sounds good.  We can close this issue when #980 is merged and open another issue just for double-ended arrow and horizontal braces.

ronkok:
I believe that this issue is now resolved.

Issue #889, opened 2017-09-17
largest \sqrt is scaled incorrectly when used with sizing commands


kevinbarabash:
```
{\sqrt{\sqrt{\sqrt{\sqrt{\sqrt{\sqrt{x}}}}}}} \Huge {\sqrt{\sqrt{\sqrt{\sqrt{x}}}}}
```
<img width="495" alt="screen shot 2017-09-17 at 12 17 44 pm" src="https://user-images.githubusercontent.com/1044413/30522689-4015cd60-9ba2-11e7-9dcb-b9489a6c9295.png">

PR #890, opened 2017-09-17
Upgrade \sqrt zoom and width

### LINKED ISSUE #888: \sqrt{} doesn't render anything
LaTeX show the square root symbol, but KaTeX shows nothing.

ronkok:
This PR does two things:

1. It evades a Safari bug which improperly zooms nested SVGs. The fix applies to `\sqrt` and to single-ended stretchy arrows. These SVGs are now single level. Their tails are sliced off using CSS `overflow: hidden`. Double-ended stretchy arrows and horizontal braces are not fixed. Fixes part of #883.

2. It adds a `min-width` to `\sqrt` so that a `\sqrt{}` with an empty argument will still render an SVG whose width matches the font glyph width. Fixes #888.


ronkok:
I've updated the [test page](http://copyasmarkup.com/katex/wideelements.html).

ronkok:
I've added a commit which fixes. #889.

kevinbarabash:
I assumed that `DisplayStyle` and `Smash` screenshots hadn't actually changed b/c the screenshots were different for Chrome only.  I should've looked at travis-ci output before making the P.  @ronkok I'll send you another PR later this evening with these screenshots.

src/delimiter.js
```diff
@@ -329,14 +329,12 @@ const sqrtSvg = function(sqrtName, height, viewBoxHeight, options) {
     }
     const pathNode = new domTree.pathNode(sqrtName, alternate);
 
-    let attributes = [["width", "100%"], ["height", height + "em"]];
+    const attributes = [["width", "400em"], ["height", height + "em"]];
     attributes.push(["viewBox", "0 0 400000 " + viewBoxHeight]);
```
kevinbarabash: It'd be good to compute the `width` attribute in terms of the `viewBox`'s width (or vice versa) so that they stay in sync.

src/delimiter.js
```diff
@@ -353,21 +351,24 @@ const sqrtSpan = function(height, delim, options) {
         sizeMultiplier = newOptions.sizeMultiplier / options.sizeMultiplier;
         spanHeight = 1 * sizeMultiplier;
         span = sqrtSvg("sqrtMain", spanHeight, viewBoxHeight, options);
+        span.style.minWidth = "0.781em";
         span.surdWidth = 0.833 * sizeMultiplier;   // from the font.
```
kevinbarabash: Why is `minWidth` different from `surdWidth`?

src/delimiter.js
```diff
@@ -353,21 +351,24 @@ const sqrtSpan = function(height, delim, options) {
...
         span = sqrtSvg("sqrtSize" + delim.size, spanHeight, viewBoxHeight, options);
+        span.style.minWidth = "0.909em";
         span.surdWidth = 1.0 / sizeMultiplier; // from the font
 
     } else {
         // Tall sqrt. In TeX, this would be stacked using multiple glyphs.
         // We'll use a single SVG to accomplish the same thing.
         spanHeight = height / sizeMultiplier;
-        viewBoxHeight = Math.floor(1000 * spanHeight);
+        viewBoxHeight = Math.floor(1000 * height);
```
kevinbarabash: This fixes the issue with the tall `\sqrt` being the wrong size when sizing functions are involved?

src/delimiter.js
```diff
@@ -329,14 +329,12 @@ const sqrtSvg = function(sqrtName, height, viewBoxHeight, options) {
     }
     const pathNode = new domTree.pathNode(sqrtName, alternate);
 
-    let attributes = [["width", "100%"], ["height", height + "em"]];
+    const attributes = [["width", "400em"], ["height", height + "em"]];
```
kevinbarabash: Why does this have to be `400em` as opposed to `100%`?

src/buildHTML.js
```diff
@@ -1159,26 +1163,16 @@ groupTypes.sqrt = function(group, options) {
     // Shift the sqrt image
     const imgShift = img.height - inner.height - lineClearance - ruleWidth;
 
-    // We add a special case here, because even when `inner` is empty, we
-    // still get a line. So, we use a simple heuristic to decide if we
-    // should omit the body entirely. (note this doesn't work for something
-    // like `\sqrt{\rlap{x}}`, but if someone is doing that they deserve for
-    // it not to work.
```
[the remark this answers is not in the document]
ronkok: I take no credit for that comment. It was there before I got here.

src/delimiter.js
ronkok: I'll give `100%` a try, but I don't think it will work. I'm now defining the `viewBox` dimensions in the same SVG as the width attribute. The height has a factor of 1000, so I think the width needs consistent treatment.

src/delimiter.js
ronkok: I've consistently used 400 ems for all the wide paths. I suppose that might not be obvious to someone looking at this cold. I'll look at it.

src/delimiter.js
ronkok: It's not the best choice of variable names. For the CSS `minWidth`, I used the actual width of the ink in the glyph. What I have called `surdwidth` is actually the horizontal advance value from the font.  Horizontal advance value is of course, a little wider than the ink.

I might adjust the `surdwidth` name and re-commit.

src/delimiter.js
ronkok: Correct.

src/buildHTML.js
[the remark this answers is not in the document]
kevinbarabash: That's explains why I missed in the previous reviews.  😅

src/delimiter.js
ronkok: I will also adjust the values of `min-width`. The paths in these surds have the `xMin` value baked into the SVG paths. So I'll increase each `min-width` by that amount.

src/delimiter.js
ronkok: I was wrong. It does work to use `width='100%'`. Not only that but when `width='100%'`, I can remove the CSS `oveflow:hidden` and it still works. So consider that done.

I wonder if that will cause sub-pixel changes in all the screenshots.

ronkok: I retract my last statement. When I use `width='100%`, IE does not slice off the long tail. I going back to `width='400em'` and CSS `overflow:hidden`.

ronkok: Final word. It turns out that it doesn't matter if I write `width='400em'` or `width='100%'`. It would matter, except that IE will not work unless I include CSS `overflow:hidden`. Once I include the CSS condition, both widths work. I'm going with 400em.

src/delimiter.js
```diff
@@ -351,25 +352,25 @@ const sqrtSpan = function(height, delim, options) {
         sizeMultiplier = newOptions.sizeMultiplier / options.sizeMultiplier;
         spanHeight = 1 * sizeMultiplier;
         span = sqrtSvg("sqrtMain", spanHeight, viewBoxHeight, options);
-        span.style.minWidth = "0.781em";
-        span.surdWidth = 0.833 * sizeMultiplier;   // from the font.
+        span.style.minWidth = "0.853em";
+        span.advanceWidth = 0.833 * sizeMultiplier;   // from the font.
```
kevinbarabash: This is clearer.  Thank you.

src/delimiter.js
kevinbarabash: The comment is good enough for now.

kevinbarabash:
@ronkok thanks for the PR.

src/delimiter.js
ronkok: > The comment is good enough for now.

Let's suppose that someone other than me comes along and writes a PR with an SVG in it. They will need to coordinate `viewBox` geometry and document geometry. They will need to realize this early, when they are arranging the `<path>` geometry.  So I am trying to leave multiple comments regarding the 1000:1 ratio. 

Such is my reasoning for writing `400em`. Reasoning with which you may disagree, but there it is.

edemaine: Would it make sense to define a `const hugeWidth = '400em'` somewhere and use that variable repeatedly, instead of repeating a constant like this?  (And one comment could explain why this choice is sensible.)  Sorry if this suggestion doesn't make sense, coming from out of context.

ronkok: A constant for 400em would do no harm. But the more critical concept is the 1000:1 ratio between ems and `viewBox` geometry. Some future contributor could waste a lot of time if they don't realize that early in their work. So any constant, defined in one place, would be improved by adding multiple comments, scattered widely.

Issue #892, opened 2017-09-18
Port Parser.js to @flow


marcianx:

kevinbarabash:
@marcianx I think this is going to be the hardest one.  Thanks for tackling this.

marcianx:
Yes. Yes it is. ;) I've run into at least 4 challenges.

The first two a are just a bunch of grunge work, but doable:
- `functionHandler`'s return type (in `defineFunction`) is `*`. It needs to be enumerated.
- `ParseNode`'s value type is `*`, and that needs to be enumerated `*`.
Although, who knows what surprises await. ;)

The third one is more fundamental. Many uses of `ParseFuncOrArgument` (which I've turned to `ParsedFunc | ParsedArg | ParsedDollar`) predicate strictly on whether or not it is a function, an they don't distinguish `ParsedArg` from `ParsedDollar`. It's a locally non-obvious (to me at least) implementation artifact that `ParsedDollar` cannot appear in these places and so the implementation doesn't check it. I would, of course, increase the final js size by adding thrown `Error`s, but I'm guessing that's not desired, and we might just want to `$FlowFixMe` those places to suppress the errors.

The fourth thing is that `parseImplicitGroup` methods call functions (`parseFunction`), but treat the return value as non-nullable and directly dereference it. It's not locally-obvious (at least to me) why that should be the case. I was going to try to produce cases that would fail based on local reasoning and see if they don't when running the full parser. Maybe tomorrow.

kevinbarabash:
Yup, definitely a bit of a ball of yarn.  If possible, please try to separate this into multiple PRs.  The `ParsedDollar` is quite implicit so anything we can do to make that more explicit is probably good.  We might want to switch out `isFunction: boolean` for `type: "function" | "argument" | "dollar"`.

Is `functionHandler`'s return type the same as `ParseNode`'s `value`'s type? 

We should try to get rid of `parseImplicitGroup`.  This should be doable at this point since we're including a reference to the Parser instance in the `context` we pass to the function handlers.  This might actually be a good first step to take as its own PR.

marcianx:
This is quite the adventure! I'm trying to go a bit more deeply into the weeds to understand the extent of the changes needed to have strong guarantees and convertibility to `@flow` without resorting to type-unsafe features like `any`.

[...]

PR #896, opened 2017-09-20
Correct @flow types. Refactor some Parser code for stricter typing.

marcianx:
Toward #892.


src/defineFunction.js
```diff
@@ -16,6 +16,12 @@ export type FunctionContext = {|
 |};
 
 // TODO: Enumerate all allowed output types.
+
+// TODO: The real type of `args` is `(?ParseNode)[]`, but doing that breaks
+//       compilation since some of these arguments are passed to `ordargument`
+//       below which requires a non-nullable argument.
```
kevinbarabash: I like this idea.  It should be pretty easy to do too.

src/Parser.js
```diff
@@ -543,47 +543,49 @@ class Parser {
...
-                return baseGroup.result;
+    /**
+     * Same as parseFunction(), except that the base is provided, guaranteeing a
+     * non-nullable result.
+     *
+     * @param {ParseFuncOrArgument} baseGroup
+     * @return {ParseNode}
+     */
+    parseGivenFunction(baseGroup) {
+        if (baseGroup.isFunction) {
```
kevinbarabash: I feel like by storing `isFunction` on `ParseFuncOrArgument` it leads to unnecessary wrapping and unwrapping.  What if we create a function that given a `ParseNode` would tell us if it was a function?  With that, we should be able to get rid of `ParseFuncOrArgument` and just use `ParseNode` everywhere.  This is maybe something to consider for a future PR.

src/Parser.js
```diff
@@ -543,47 +543,49 @@ class Parser {
...
-            } else {
-                return baseGroup.result;
+    /**
+     * Same as parseFunction(), except that the base is provided, guaranteeing a
+     * non-nullable result.
+     *
+     * @param {ParseFuncOrArgument} baseGroup
+     * @return {ParseNode}
+     */
+    parseGivenFunction(baseGroup) {
```
kevinbarabash: Good idea separate this from the `parseFunction` which might return `null`.

src/Parser.js
marcianx: Yes, I intend to do more changes within `Parser.js` itself in subsequent PRs while the `ParseNode` refactor waits on other PRs to complete. `ParseFuncOrArgument` will be replaced with a union of 3 structs per our discussion in #892. We can see whether it makes sense to have a function for an is-function check or simply use `foo.type === "fn"` (or `"$"` or `"arg"`) or some such.

kevinbarabash:
@marcianx thanks for keeping the changes small.

PR #901, opened 2017-09-22
Replace ParseFuncOrArgument with type-safer alternative.

marcianx:
In the process, document consequences of implementation
details that make type-safety difficult to implement.

Towards #892.

~~This one might spawn a bit of a discussion. ;)~~ (Well, perhaps not so after the second commit.)

### DIFF
--- a/src/Parser.js
+++ b/src/Parser.js
@@ -38,28 +38,39 @@ import ParseError from "./ParseError";
  *
  * The earlier functions return ParseNodes.
  * The later functions (which are called deeper in the parse) sometimes return
- * ParseFuncOrArgument, which contain a ParseNode as well as some data about
+ * ParsedFuncOrArgOrDollar, which contain a ParseNode as well as some data about
  * whether the parsed object is a function which is missing some arguments, or a
  * standalone object which can be used as an argument to another function.
  */
 
-/** A function name or an argument to a function. */
-class ParseFuncOrArgument {
-    /**
-     * @param {ParseNode|string} result It's a string if `isFunction=true` or if
-     *     `result="$"` for switching into math mode; otherwise, it's a ParseNode
-     *     and `isFunction` must be false. If it's a function, the string should
-     *     be a name defined by defineFunction, e.g. "\\frac".
-     * @param {boolean} isFunction True when this is a function. False when it's a
-     *     function argument or "$" to switch into math mode.
-     * @param {Token} token
-     */
-    constructor(result, isFunction, token) {
-        this.result = result;
-        // Is this a function (i.e. is it something defined in functions.js)?
-        this.isFunction = isFunction;
-        this.token = token;
-    }
+/* TODO: Uncomment when porting to flow.
+type ParsedType = "fn" | "arg" | "$"
+type ParsedFunc = {|
+    type: "fn",
+    result: string, // Function name defined via defineFunction (e.g. "\\frac").
+    token?: ?Token,
+|};
+type ParsedArg = {|
+    type: "arg",
+    result: ParseNode,
+    token?: ?Token,
+|};
+type ParsedDollar = {|
+    // Math mode switch
+    type: "$",
+    result: "$",
+    token?: ?Token,
+|};
+type ParsedFuncOrArgOrDollar = ParsedFunc | ParsedArg | ParsedDollar;
+*/
+
+/**
+ * @param {ParseNode} result
+ * @param {?Token=} token
+ * @return {ParsedArg}
+ */
+function newArgument(result, token) {
+    return {type: "arg", result, token};
 }
 
 export default class Parser {
@@ -238,6 +249,8 @@ export default class Parser {
 
     /**
      * Handle a subscript or superscript with nice errors.
+     * @param {string} name For error reporting.
+     * @return {ParsedNode}
      */
     handleSupSubscript(name) {
         const symbolToken = this.nextToken;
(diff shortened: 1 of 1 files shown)


marcianx:
Added another commit to resolving the TODO in the first code. This results in some of the errors changing from "$ within math mode" to, e.g. "Unexpected $ at position 3: a^$". This resolves the type-safety issues mentioned in the first commit's TODOs.

src/Parser.js
```diff
@@ -38,28 +38,50 @@ import ParseError from "./ParseError";
...
+*/
+
+/**
+ * @param {ParseNode} result
+ * @param {?Token=} token
+ * @return {ParsedArg}
+ */
+function newArgument(result, token) {
+    return {type: "arg", result, token};
+}
```
kevinbarabash: I would be nice to have symmetry between how `ParsedFunc`, `ParsedArg`, and `ParsedDollar` get created.

src/Parser.js
```diff
@@ -38,28 +38,50 @@ import ParseError from "./ParseError";
...
+type ParsedArg = {|
+    type: "arg",
+    result: ParseNode,
+    token?: ?Token,
+|};
+type ParsedDollar = {|
+    // Math mode switch
+    type: "$",
+    result: "$",
+    token?: ?Token,
```
kevinbarabash: `token` can be mandatory for `ParsedDollar`.

src/Parser.js
```diff
@@ -38,28 +38,50 @@ import ParseError from "./ParseError";
...
-        this.result = result;
-        // Is this a function (i.e. is it something defined in functions.js)?
-        this.isFunction = isFunction;
-        this.token = token;
+/* TODO: Uncomment when porting to flow.
+type ParsedType = "fn" | "arg" | "$"
+type ParsedFunc = {|
+    type: "fn",
+    result: string, // Function name defined via defineFunction (e.g. "\\frac").
+    token?: ?Token,
```
kevinbarabash: `token` can be mandatory for `ParsedFunc`.

src/Parser.js
```diff
@@ -38,28 +38,50 @@ import ParseError from "./ParseError";
...
+    type: "$",
+    result: "$",
+    token?: ?Token,
+|};
+type ParsedFuncOrArgOrDollar = ParsedFunc | ParsedArg | ParsedDollar;
+*/
+
+/**
+ * @param {ParseNode} result
+ * @param {?Token=} token
```
kevinbarabash: Is the reason we don't pass a token sometimes is that an argument might span multiple tokens?

src/Parser.js
marcianx: Done. I was just keeping the generated code low a concern you'd mentioned in a previous PR.

src/Parser.js
marcianx: Not really. Tokens can be merged from mutliple tokens. Here are the places where a token isn't supplied to the constructor:

- `arg = newArgument(this.handleUnsupportedCmd());`
- `return newArgument(new ParseNode("color", match[0], this.mode));`
- `return newArgument(new ParseNode("size", data, this.mode));`

src/Parser.js
marcianx: Done.

src/Parser.js
marcianx: Done.

marcianx:
@kevinbarabash 
> It feels like things are getting more complicated, but I think that's a function of making implicitly complicated things more explicit. Any ideas of how we might simplify args vs. fn vs. $ parsing in the future?

There's no new complexity added here; only documentation of existing assumptions and an adherence to type strictness that wasn't there before -- for various inputs, we actually got "$" (string) where a `ParseNode` was expected, only to throw an error later. If you haven't, please read the massive and detailed `TODO` at the bottom of the first commit (which I removed in the second commit) about the unintuitive way the `Parser` behaved before the second commit.
I can't think of any obvious way to fix this without actually rethinking parts of the `Parser`.
Let's first get static typechecking in place. ;)

src/Parser.js
marcianx: Actually, I can return a token in all of those existing cases.

But I also missed
- `return newArgument(new ParseNode("ordgroup", expression, this.mode, firstToken, lastToken));`
for which there isn't an obvious token text. Specifically, I don't know what to put in as the second parameter to `firstToken.range(lastToken, ???)`. I used `firstToken.text` for now and added a commit.

These changes are included in the latest commit.

kevinbarabash: I think that's fine for now.  Eventually, we'll want to pass the token when constructing a `ParseNode` from the arg's `result`.  It may make sense for this token to include text that spans from the start of the first token to the end of the last token, but that can be in a future PR.

marcianx: I was trying to think of how to do that. While that's trivial if `firstToken` and `lastToken` come from the same lexer -- we can use `firstToken.loc.lexer.text.substring(firstToken.loc.start, lastToken.loc.end)` -- it's not so straightforward when they're not.

marcianx:
Rebased.

src/Parser.js
kevinbarabash: I thinking I was being overly picky in that previous PR, sorry about that.  We definitely want to keep the size down, but I doubt this will have much of an impact especially after uglify.  As we continue to make things more explicit in Parser.js I'm sure the code size will fluctuate up and down a bit.  In the other diff you dedupe `parseArguments` so we saved some bytes so it's all good.  ^_^

Issue #805, opened 2017-08-19
switch to webpack-dev-server for server.js


kevinbarabash:
Right now we build katex.js on each page load.  On my machine this can take from 20 to 40 seconds.  With webpack-dev-server, the server would build the bundle on file change instead of whenever the browser is reloaded.  It also opens the possibility of hot module reloading.

edemaine:
Sounds great!  I've noticed the slowness of this lately when testing.

rrandallcainc:
I'd be happy to take a stab at this for you.

rrandallcainc:
Solved by: https://github.com/Khan/KaTeX/pull/902

PR #902, opened 2017-09-23
Webpack dev server

rrandallcainc:
As part of this PR:

1) Migrated everything to ES6 modules (helps w/ tree-shaking)
2) Cleaned up some of the imports
3) Created initial functioning webpack dev server & removed the old server.js
4) Added a couple scripts to package.json, including one for building the css

This PR is in response to: https://github.com/Khan/KaTeX/issues/805


rrandallcainc:
[clabot:check]

edemaine:
@rrandallcainc Did you intentionally close this?

rrandallcainc:
@edemaine Whoops, no. Not sure how that happened! Will take a look at the failing tests either tonight or tomorrow.

.babelrc
```diff
@@ -1,7 +1,19 @@
 {
-    "presets": ["es2015", "flow"],
+    "presets": [
+        ["es2015", { "modules": false }],
```
kevinbarabash: Is setting `modules: false` to reduce overhead from the default `"commons"` setting?

contrib/auto-render/index.html
```diff
@@ -9,7 +9,7 @@
     <title>Auto-render test</title>
     <script src="/katex.js" type="text/javascript"></script>
     <link href="/katex.css" rel="stylesheet" type="text/css">
-    <script src="./auto-render.js" type="text/javascript"></script>
+    <script src="/renderMathInElement.js" type="text/javascript"></script>
```
kevinbarabash: It's be nice to get away without renaming these files as people may be relying on the existing script name.

static/katex.css
```diff
@@ -0,0 +1 @@
...ngarian>span{position:relative;left:.25em}.katex .mtable .vertical-separator{display:inline-block;margin:0 -.025em;border-right:.05em solid #000}.katex .mtable .arraycolsep{display:inline-block}.katex .mtable .col-align-c>.vlist-t{text-align:center}.katex .mtable .col-align-l>.vlist-t{text-align:left}.katex .mtable .col-align-r>.vlist-t{text-align:right}.katex .svg-align{text-align:left}.katex svg{display:block;position:absolute;width:100%}.katex svg path{fill:currentColor}.katex svg line{stroke:currentColor}.katex .stretchy{width:100%;display:block}.katex .stretchy:after,.katex .stretchy:before{content:""}.katex .hide-tail{width:100%;position:relative;overflow:hidden}.katex .x-arrow-pad{padding:0 .5em}.katex .mover,.katex .munder,.katex .x-arrow{text-align:center}.katex .boxpad{padding:0 .3em 0 .3em}.katex .fbox{box-sizing:border-box;border:.04em solid #000}.katex .fcolorbox{box-sizing:border-box;border:.04em solid}.katex .cancel-pad{padding:0 .2em 0 .2em}.katex .mbin+.cancel-lap,.katex .mord+.cancel-lap{margin-left:-.2em}.katex .cancel-lap+.mbin,.katex .cancel-lap+.mord,.katex .cancel-lap+.msupsub{margin-left:-.2em}.katex .sout{border-bottom-style:solid;border-bottom-width:.08em}
```
kevinbarabash: Would be it be possible to build this file when the server starts instead of checking it into source control?

.babelrc
rrandallcainc: Setting to `modules: false` enables the use of tree-shaking with webpack.

contrib/auto-render/index.html
rrandallcainc: Agreed. Will look into it.

static/katex.css
rrandallcainc: Nice catch. Meant to take care of this prior to opening.

rrandallcainc:
@kevinbarabash I brought back server.js, and it's now leveraging webpack-dev-server. It's still failing however. I'll look into this tomorrow.

kevinbarabash:
@rrandallcainc one option might be to revert the changes to server.js and only use server.js with screenshotter.js for now and then we could figure out how to get webpack-dev-server working with the screenshot separately.

rrandallcainc:
@kevinbarabash Agreed. Selenium tests are now back to running off of server.js.

kevinbarabash:
@rrandallcainc I checked out your branch.  I was having troubles loading 0.0.0.0:7936 in Safari 10.x.  I think it may have something to do with https://github.com/webpack/webpack-dev-server/issues/1090.  I tried Safari 11.x and that solved the issue.  I think it's probably okay to merge with this issue, but we'll want to call this out in the release notes.

I also tried running `make build` and ran into the following issue at the end:
```
~/KaTeX/node_modules/browserify/bin/cmd.js -t [ babelify ] katex.js --standalone katex > build/katex.js

~/KaTeX/katex.js:10
import ParseError from "./src/ParseError";
^
ParseError: 'import' and 'export' may appear only with 'sourceType: module'
make: *** [build/katex.js] Error 1
```

kevinbarabash:
Sorry about some scripts being in package.json while others are in `Makefile`.  We use `make build` as part of the release.sh script (it's called indirectly via `make dist`).

rrandallcainc:
@kevinbarabash Build should now be working with the makefile. Perhaps now that we leverage webpack, do you think it would be worth leveraging it for builds as well? Just a thought.

rrandallcainc:
I reverted back to commonjs modules in babelrc. At this point there's no point in leveraging ES6 modules since it will need to be transpiled to commonjs for the consuming app. 

Perhaps in the future, there could be two distributed builds (bundle.js & bundle.es6.js).

kevinbarabash:
LGTM.  Reloading is fast.  The source maps work well in Chrome, but not in Safari (I blame Safari).

kevinbarabash:
@rrandallcainc thanks for the PR.  I think we'll want to use webpack eventually for the dist build.  If we had two dist builds, would we need a second npm package?

rrandallcainc:
@kevinbarabash I would agree with you that the source-maps issue is related to Safari/Webpack.

I'm happy to hear that things are building much faster now! There are a couple other things I'd like to work on, but I would be happy to help with more build related items in the future (if need be).

caiofcm:
I am a beginner user of katex (as in web dev). I am not being able to import auto-render.js using the es6 module scheme.
It only works if I change the line in auto-render.js (https://github.com/Khan/KaTeX/blob/master/contrib/auto-render/auto-render.js)

-`module.exports = renderMathInElement;`
to 
+`export default renderMathInElement;`
Analogously to what is done in #902

kevinbarabash:
@caiofcm I was under the impression that you could use `import` statements when importing `commonjs` modules, but maybe that's only if you're using babel to smooth over the rough edges.

rrandallcainc:
@caiofcm What are you using to build? Are you getting any errors?

caiofcm:
I am building with webpack (3.8.1) and .babelrc with "presets": [ "es2015"] }.

With the original code line `module.exports = renderMathInElement` I get the error:

Importing with:
`var renderMathInElement = require('katex/contrib/auto-render/auto-render.js').renderMathInElement`

Error: `Uncaught TypeError: Cannot assign to read only property 'exports' of object '#<Object>'`

Importing with (es6 style):
`import renderMathInElement from 'katex/contrib/auto-render/auto-render.js'`
error is as above and with the warning:
`"export 'default' (imported as 'renderMathInElement') was not found in 'katex/contrib/auto-render/auto-render.js'`

rrandallcainc:
Yeah, this is a bug that exists with webpack. I remembered that this issue existed with webpack, but then forgot it when creating this PR. 

Basically you can't mix es6 with commonjs in the same file. See: https://github.com/almende/vis/issues/2934

We cannot change to ES6 format due to backwards compatibility. The solution would be to change the imports to commonjs format.

I've been extremely busy lately, but I can see if I can get this in maybe next week. Should be a simple fix. If you want, you can open a PR, and I can take a brief look.

Basically 
`import Blah from 'blah'; => `const Blah = require('Blah').default;`

caiofcm:
@rrandallcainc thank you for your reply.
I wouldn't now how to properly create this PR.

But, if the other files in the project uses the export default syntax, such as [parseTree.js](https://github.com/Khan/KaTeX/blob/fbffdc5fc713f26d84ad00bacb5279abb9529bfd/src/parseTree.js#L20) and also from the contrib folder [katex2tex.js](https://github.com/Khan/KaTeX/blob/fbffdc5fc713f26d84ad00bacb5279abb9529bfd/contrib/copy-tex/katex2tex.js#L52), shouldn't auto-render.js also follow it ?

PR #904, opened 2017-09-24
Add SourceLocation to encapsulate Token/ParseNode debug information.

marcianx:
Towards changes discussed in #892.
I made SourceLocation a class instead of a struct in order to contain an associated static method for merging two (nullable) locations representing a range. This coalesces some logic that gets very slightly more involved due to lexer/start/end moving one layer deeper.

### DIFF
--- a/src/MacroExpander.js
+++ b/src/MacroExpander.js
@@ -140,7 +140,7 @@ export default class MacroExpander implements MacroContextInterface {
                         expansion.splice(i + 1, 1); // drop first #
                     } else if (/^[1-9]$/.test(tok.text)) {
                         // replace the placeholder with the indicated argument
-                        expansion.splice(i, 2, ...args[tok.text - 1]);
+                        expansion.splice(i, 2, ...args[+tok.text - 1]);
                     } else {
                         throw new ParseError(
                             "Not a valid argument number",
--- a/src/ParseNode.js
+++ b/src/ParseNode.js
@@ -1,5 +1,6 @@
 // @flow
-import {LexerInterface, Token} from "./Token";
+import {Token} from "./Token";
+import SourceLocation from "./SourceLocation";
 import type {Mode} from "./types";
 
 /**
@@ -14,12 +15,7 @@ export default class ParseNode {
     type: *;
     value: *;
     mode: Mode;
-    // TODO: We should combine these to ({lexer, start, end}|void) as they
-    // should all exist together or not exist at all. That way, only a single
-    // void check needs to be done to see if we have metadata.
-    lexer: LexerInterface | void;
-    start: number | void;
-    end: number | void;
+    loc: ?SourceLocation;
 
     constructor(
         type: string,       // type of node, like e.g. "ordgroup"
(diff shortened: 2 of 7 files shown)


src/MacroExpander.js
```diff
@@ -140,7 +140,7 @@ export default class MacroExpander implements MacroContextInterface {
                         expansion.splice(i + 1, 1); // drop first #
                     } else if (/^[1-9]$/.test(tok.text)) {
                         // replace the placeholder with the indicated argument
-                        expansion.splice(i, 2, ...args[tok.text - 1]);
+                        expansion.splice(i, 2, ...args[+tok.text - 1]);
```
kevinbarabash: Why the `+`?

src/ParseNode.js
```diff
@@ -33,10 +29,6 @@ export default class ParseNode {
         this.type = type;
         this.value = value;
         this.mode = mode;
-        if (firstToken && (!lastToken || lastToken.lexer === firstToken.lexer)) {
-            this.lexer = firstToken.lexer;
-            this.start = firstToken.start;
-            this.end = (lastToken || firstToken).end;
-        }
+        this.loc = SourceLocation.range(firstToken, lastToken);
```
kevinbarabash: nice

src/SourceLocation.js
```diff
@@ -0,0 +1,42 @@
...
+export default class SourceLocation {
+    lexer: LexerInterface; // Lexer holding the input string.
+    start: number;         // Start offset, zero-based inclusive.
+    end: number;           // End offset, zero-based exclusive.
+
+    constructor(lexer: LexerInterface, start: number, end: number) {
+        this.lexer = lexer;
+        this.start = start;
+        this.end = end;
+        Object.freeze(this); // Immutable to allow sharing in range().
```
kevinbarabash: Apparently this impacted performance in the past, but modern browsers seem to be unaffected.

src/SourceLocation.js
```diff
@@ -0,0 +1,42 @@
...
+        second?: {loc: ?SourceLocation},
+    ): ?SourceLocation {
+        if (!second) {
+            return first && first.loc;
+        }
+        if (!first || !first.loc || !second.loc ||
+            first.loc.lexer !== second.loc.lexer) {
+            return null;
+        }
+        return new SourceLocation(first.loc.lexer, first.loc.start, second.loc.end);
```
kevinbarabash: Minor style nit: using `if/else if/else` makes it easier to see that only one of these blocks will be executed.

src/MacroExpander.js
marcianx: To convert the string to a number (see second commit). When making the type of `Token`'s `text` explicitly `string` (which we thought was previously unnecessary) instead of `*`, this line through a type error (`string` and `number` couldn't be added).
I was surprised, as I expected `Token`'s `text` to have been inferred as `string` due to its constructor. Flow seems to be doing some global type inference, giving some surprising results here. Hence, we should avoid using `*` for maximal type safety.
Thanks for asking. ;)

src/SourceLocation.js
marcianx: Done.

src/SourceLocation.js
marcianx: Ok, LMK if you'd like me to remove.

kevinbarabash: Let's leave it for now.

PR #909, opened 2017-09-29
Convert nested SVGs to single-level SVGs

ronkok:
This PR evades a Safari bug which causes nested SVGs to zoom improperly. All nested SVGs are converted to single-level SVGs. Arrow tails are trimmed to proper length by use of CSS `overflow: hidden` on a container span.

This builds on the work of PR #890 and fixes the remainder of issue #883.


ronkok:
I've updated the [test page](http://copyasmarkup.com/katex/wideelements.html) with the code from this PR.

kevinbarabash:
@ronkok thanks for fixing this issue.  I've sent you a PR with updated screenshots.  I'll have a look at the code tomorrow morning when I'm fresh.

kevinbarabash:
@ronkok sorry.  Today's been a super busy day and I haven't had a chance to look at the code yet.  I should hopefully get to it Monday.  The weather's supposed to be bad where I'm at so it'll be a good day for code reviewing.  ^_^

ronkok:
No worries.

static/katex.less
```diff
@@ -581,6 +581,41 @@
         position: relative;   // ditto
         overflow: hidden;     // This line applies to all browsers.
     }
+	
+	.halfarrow-left {
```
kevinbarabash: Please use spaces instead of tabs.

static/katex.less
```diff
@@ -581,6 +581,41 @@
...
+		position: absolute;
+		left: 0;
+		width: 50.1%;
+		overflow: hidden;
+	}
+
+	.halfarrow-right {
+		position: absolute;
+		right: 0;
+		width: 50%;
```
kevinbarabash: Why is 50% okay when positioning things from the right?

static/katex.less
```diff
@@ -581,6 +581,41 @@
...
+		overflow: hidden;
+	}
+
+	.brace-left {
+		position: absolute;
+		left: 0;
+		width: 25.1%;
```
kevinbarabash: The extra 0.1% here and in brace-right is to overlap the center part of the brace, right?  Did certain browsers exhibit a gap without the 0.1%?

src/stretchy.js
```diff
@@ -193,61 +192,52 @@ const svgSpan = function(group, options) {
...
-        const innerSVGs = [];
         height = vbHeight / 1000;
         minWidth = gWidth;
 
         for (let i = 0; i < numSvgChildren; i++) {
             path = new domTree.pathNode(paths[i]);
-            attributes = [];
+
+            attributes = [["width", "400em"], ["height", height + "em"]];
+            attributes.push(["viewBox", `0 0 ${viewBoxWidth} ${vbHeight}`]);
```
kevinbarabash: It'd be nice to make these variable names more consistent, maybe `vbWidth` and `vbHeight`.

src/stretchy.js
```diff
@@ -193,61 +192,52 @@ const svgSpan = function(group, options) {
...
+                widthClass = ["halfarrow-left", "halfarrow-right"][i];
                 align = ["xMinYMin", "xMaxYMin"][i];
             } else {
                 // 3 inner SVGs, as in a brace
-                if (i > 0) {
-                    attributes.push(["x", [null, "25%", "74.9%"][i]]);
-                }
-                width = ["25.5%", "50%", "25.1%"][i];
+                widthClass = ["brace-left", "brace-center", "brace-right"][i];
                 align = ["xMinYMin", "xMidYMin", "xMaxYMin"][i];
```
kevinbarabash: This works b/c the SVG for each part of the brace extends out the right, in both directions, and to the left if we were looking at each pieces from left to right.

static/katex.less
kevinbarabash: Answering my own question: it's b/c we only want a 0.1% overlap.

kevinbarabash:
@ronkok sorry for the delay in reviewing this.

static/katex.less
ronkok: Yes, I have seen browsers exhibit a one screen-pixel gap when I tried to set widths to exactly 50%. It happened occasionally, not frequently. It happened in multiple browsers, not any particular browser. I think the behavior is probably due to floating point rounding error.

Now to be precise, I observed those gaps when I was using SVGs in background images. I suppose that the current inline SVGs, written directly into the DOM, might act differently. But I feel a lot more secure knowing that I have that 0.1% overlap.

src/stretchy.js
ronkok: I can make the names consistent as `viewBoxWidth` and `viewBoxHeight`. To do that, I'll have to remove the `const` from the destructuring statement:
````
const [paths, gWidth, vbHeight, alignOne] = katexImagesData[label];
````

src/stretchy.js
ronkok: Exactly.

kevinbarabash:
@ronkok thanks for making all of the SVGs work the same.

static/katex.less
edemaine: FWIW, this is a general rendering/antialiasing problem with most SVG renderers I've encountered (including Chrome): abutting features generally leave tiny gaps.  So definitely approve of the slight overlap (assuming 0.1% is never too big?).

Issue #910, opened 2017-10-01
Newline (\\) eats non-size argument surrounded by brackets


konn:
The following code is valid in LaTeX language, but KaTeX interprets `[b]` to be size specification for `\\` and throws error:

```latex
\begin{gathered}
a \\ [b]
\end{gathered}
```

The same holds even if one add whitespaces (including newline).
For now, one can avoid this bug by replace `\\` by `\\{}` or `[b]` by `{[b]}`.

edemaine:
KaTeX is not properly recognizing the space between `\\` and `[b]`.  Note that

```latex
\begin{gathered}
a \\[b]
\end{gathered}
```

is invalid LaTeX.  Taking a look at this now.

edemaine:
Fix in #912. Thanks @konn for the bug report!

konn:
Great news! Thank you for your rapid support!

hrldcpr:
I just encountered this bug, it seems to have returned from the dead!

You can easily reproduce on katex.org with:

```latex
a \\ [b]
```

edemaine:
Thanks for the report. I'm surprised our automated tests didn't catch this regression...

PR #912, opened 2017-10-01
Fix space handling

edemaine:
This PR fixes several issues with space handling, in particular fixing #910:
1. "Control symbols" (as they're called in the TeXbook), such as `\\`, should
   not have spaces eaten after them (only "control words" such as `\foo`).
2. In math mode, spaces should be consumed at the parser level, not the
   gullet level.  This enables `\\ [x]` to parse differently from `\\[x]`
3. Eat spaces between arguments, so `\frac x y` still works.
   (This used to work only because math mode ate all spaces.
    The analog in text mode wouldn't have worked.)

Notably, `Parser.consume` no longer eats spaces when in math mode; rather, I get rid of them in `parseExpression` (which seems more natural -- though there is a worry that some code somewhere assumed that spaces would already be consumed). This implies that we never use `MacroExpander.get` with a `true` argument, which could simplify the code of both `get` and `unget`. It also means that `switchMode` no longer does anything useful.  Is it worth removing any of this code?  I'm unsure.


kevinbarabash:
It looks like there's an issue with the `Sqrt` screenshot test.  I'm guessing whatever the issue is with that test, it's probably something that we could write a unit test for.

edemaine:
@kevinbarabash Thanks for spotting that. Indeed, it was a bug, and I added a test for it.  Spaces needed to be checked for (and ignored) also in the sup/subscript handling of `parseAtom`.  I fixed it, and added a test that failed otherwise.  I've searched for other places where spaces need to be ignored, but I haven't found any...

test/katex-spec.js
```diff
@@ -2431,10 +2438,30 @@ describe("A macro expander", function() {
...
+    it("should consume spaces after macro with \\relax", function() {
+        compareParseTree("\\text{\\foo }", "\\text{}", {"\\foo": "\\relax"});
+    });
+
+    it("should consume spaces after \\relax", function() {
+        compareParseTree("\\text{\\relax }", "\\text{}");
+    });
+
+    it("should consume spaces after control-word function", function() {
+        compareParseTree("\\text{\\KaTeX x}", "\\text{\\KaTeX\\relax x}");
```
kevinbarabash: So these two things parse the same, but I thought control words weren't allowed inside `\text{}`.  Why is it important that these two parse the same?

test/katex-spec.js
```diff
@@ -2475,13 +2502,37 @@ describe("A macro expander", function() {
...
+
     it("should allow for empty macro argument", function() {
         compareParseTree("\\foo\\bar", "()", {
             "\\foo": "(#1)",
             "\\bar": "",
         });
     });
 
+    // The following is not currently possible to get working, given that
+    // functions and macros are dealt with separately.
```
kevinbarabash: What's the current behavior?  Can you open an issue for this?

test/katex-spec.js
```diff
@@ -2353,6 +2355,11 @@ describe("An aligned environment", function() {
             .toParse();
     });
 
+    it("should allow cells in brackets", function() {
+        expect("\\begin{aligned}[a]&[b]\\\\ [c]&[d]\\end{aligned}")
+            .toParse();
```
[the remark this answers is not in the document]
edemaine: Yes.  This matches LaTeX behavior, based on testing.  (`\@ifnextchar[` must get a space instead of a `[`.)

test/katex-spec.js
edemaine: I wanted to make sure that `\text{\KaTeX x}` does not render a space.  Ideally I'd say that `\text{\KaTeX x}` renders the same as `\text{{\KaTeX}x}` but that generates another group... perhaps I should tweak the test to actually look for features in the parse tree.  (Control words like `\KaTeX` definitely work in text mode.)

src/Parser.js
```diff
@@ -140,9 +140,13 @@ export default class Parser {
      * and fetches the one after that as the new look ahead.
      */
     consume() {
-        this.nextToken = this.gullet.get(this.mode === "math");
+        this.nextToken = this.gullet.get(false);
```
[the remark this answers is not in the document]
edemaine: Yes, I wanted a second opinion on this.  As I wrote above: "we never use `MacroExpander.get` with a `true` argument, which could simplify the code of both `get` and `unget`."  We no longer need any of the space saving/restoring mechanics, so I'll get rid of that.

"It also means that `switchMode` no longer does anything useful." (in the parser)  This one I'm less sure of.  Maybe `switchMode` would be useful in the future, e.g. if we can ever tweak catcodes in other ways? (e.g. verbatim or url modes?)

test/katex-spec.js
edemaine: Opened #924

edemaine:
Thanks for the review!  All comments should be addressed in the two new commits.

src/Parser.js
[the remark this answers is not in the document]
kevinbarabash: If you think `switchMode` will come in handy in the future let's keep it, but please add a comment as to why we're keeping it.

test/katex-spec.js
kevinbarabash: I got an error when I tried `\text{\KaTeX}` on the demo page so maybe it's just some control words don't work.

src/Parser.js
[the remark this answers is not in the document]
kevinbarabash: I think we still need `swtichMode` because we store the current mode in the parse nodes and that value gets used in the builders.  Ignore my comment about adding a comment.

edemaine:
Oops, I merged this before simplifying `get`.  Perhaps useful to look at the diff separately anyway.  It's in a new PR: #928.

PR #918, opened 2017-10-05
ensure that build/ exists

kevinbarabash:
release.sh runs `git clean -fdx build dist` before `make setup dist` and so the `build` folder isn't there when the `build/katex.js`, `build/katex.min.js:`, etc. build targets try to get made.  I had considered adding a line `mkdir -p build` in release.sh, but the build targets should be buildable without having to run release.sh and the `build` folder doesn't exist on checkout so that's why I fixed things the way I did.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -41,15 +41,19 @@ lint: $(NIS)
 	$(NPM) run lint
 
 build/katex.js: katex.js $(wildcard src/*.js) $(NIS)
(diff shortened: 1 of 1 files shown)


kevinbarabash:
I'm merging with this without a review b/c it's a safe change, doesn't affect KaTeX code itself, and I really want to get a release out.

PR #920, opened 2017-10-06
Fixing CJS modules and Globals

rrandallcainc:
**Problem:** 

When loading any of the katex bundles either through commonjs imports, or a script tag, the module will be embedded inside a "default" property, instead of attached to the modules object itself. 

This regression was caused due to: 
https://github.com/Khan/KaTeX/pull/902/files#diff-7db782e7d1aa172755d2943d78caeac6R78
https://github.com/Khan/KaTeX/pull/902/files#diff-b34825135eb977dc3f05a6e4b5cf6944R103

See image below for how it manifests:

screen shot 2017-10-06 at 9 35 36 am

**Why?**

The introduction of babel 6 changed the way es6 modules interact with commonjs modules and globals.

When you mix them,  it will resolve to something like this:

// Module A.js
```
export default A;
```
//Some other module
```
const A = require('A.js');
```
The value of A will be: { default: A }... in other words I would need to do 
```
const A = require('A').default;
```
Similarly for script tags:

```
<script src="A.js"></script>
<script>
    console.log(A.default);
</script>
```
There's a good description of this here: https://blog.kentcdodds.com/misunderstanding-es6-modules-upgrading-babel-tears-and-a-solution-ad2d5ab93ce0

**Possible Solutions:**

1) Bite the bullet. Have consumers using commonjs, just use the .default property. Ideally we'd want to move towards ES6 modules anyway. That being said, this would be a breaking change. 
2) I came across: https://github.com/59naga/babel-plugin-add-module-exports, but it was highly recommended not to use it. It basically reverts to Babel5's behavior which was wrong and had issues.
3) Leverage webpack for builds. Webpack provides a "libraryExport" property which can be used to ensure the .default property is not present when importing a umd. See more here: https://github.com/webpack/webpack/issues/3929. That being said, I'm not sure how standard this feature is.
4) Leverage the commonjs syntax for the entry file. This is likely the easiest option for now.

For this case, I went with number **4** as it is probably the cleanest non-breaking solution. I think in the long-term we'd want to consider 1.


kevinbarabash:
@rrandallcainc thanks for figuring out what the issue was and submitting a fix.

contrib/auto-render/auto-render.js
```diff
@@ -1,7 +1,7 @@
 /* eslint no-console:0 */
 /* global katex */
 
-import splitAtDelimiters from "./splitAtDelimiters";
+const splitAtDelimiters = require("./splitAtDelimiters").default;
```
kevinbarabash: It should be possible to mix `import` syntax with `module.exports`.

kevinbarabash: @rrandallcainc I double checked this by reverting this line, running `make dist`, and doing `const renderMathInElement = require("./dist/contrib/auto-render.js")`.  The result was a function.  Can you change the `require` statements back to `import`s?

rrandallcainc:
@kevinbarabash Good point. Definitely cleaner without using .default

ronkok:
The auto-render extension still won't work for me. I'm getting the error message: "renderMathInElement is not a function".  I have a [test page](http://copyasmarkup.com/katex/Investigate.html) up.

kevinbarabash:
@ronkok the test page you linked to is still using v0.9.0-alpha.  I haven't published a new release yet.

ronkok:
Yes, of course. Silly me.

xzel23:
Is this still unresolved in 0.10.0? I get following exception:

> netscape.javascript.JSException: ReferenceError: Can't find variable: renderMathInElement

for this code:

> renderMathInElement(document.body);

What is the suggested workaround to access `renderMathInElement`?

edemaine:
@xzel23 This PR is merged. Please open another issue, with details about how you're trying to use KaTeX.

PR #922, opened 2017-10-09
Edit link to Function Support page

ronkok:
On README, change a link so that it directs not to the wiki's function support page, but to the HTML function support page.


kevinbarabash:
@ronkok thanks.

Issue #90, opened 2014-09-19
Support for non-standard macros \class and \cssId


hastebrot:
Hi!

Support for the [TeX HTML extension from MathJax](http://docs.mathjax.org/en/latest/tex.html#html) in the output of KaTeX would be nice.

---

Excerpt from the docs:

The HTML extension gives you access to some HTML features like styles, classes, element ID’s and clickable links. It defines the following non-standard macros: 
- `\class{name}{math}`: Attaches the CSS class name to the output associated with math when it is included in the HTML page. This allows your CSS to style the element.
- `\cssId{id}{math}`: Attaches an id attribute with value id to the output associated with math when it is included in the HTML page. This allows your CSS to style the element, or your javascript to locate it on the page.

rrandallcainc:
@kevinbarabash How do we feel about bringing this back from the dead? 

I know there's a few people who are looking for it.

If you're okay with the concept, I can see what I can about the implementation.

kevinbarabash:
@rrandallcainc since this is non-standard it should probably go in contrib/.  In order to do that though, we need a way to add functions from outside of KaTeX core.  You could expose `defineFunction` as `__defineFunction`.  There should also be automated tests for this so that we can catch any changes to `defineFunction` that break these commands in contrib/.

ylemkimon:
@kevinbarabash With strict mode in work, I think we can revisit this and consider enabling them in non-strict mode instead of `contrib`.

PR #923, opened 2017-10-09
Implemented `\href' command

konn:
Implemented `\href` command in LaTeX hyperref package.
It seems that LaTeX simply ignores special commands for hyper-linking and processes spacing as usual.
Since I don't make out how to simulate this spacing behaviour in KaTeX, I adopt the following strategy:

1. If it has the only element, then use its math-type;
2. if it has more than two elements, and the types of both ends of elements coincides, then use it as a class for the anchor tag;
3. otherwise, we inject empty span-tag aside both ends of generated anchor tag, both with the same classes as the first and the last ones respectively.


edemaine:
Thanks @konn for working on this.  `\href` is definitely something we should support.

I don't understand what you mean by your three rules though.  `\href` should not be setting class, should it?  Isn't that the purpose of `\class` (see #90)?

Isn't the second argument of `\href` treated as if it's in `\text`?  You should be able to use code similar to `\text`, `\raisebox`, etc. which switch to text mode.

konn:
> I don't understand what you mean by your three rules though. \href should not be setting class, should it? Isn't that the purpose of \class (see #90)?

Well, if you don't specify any math-type via class of an anchor tag, you get the different spacing behaviour than latex's default.

Consider the following input:

```latex
\documentclass[a4]{article}
\usepackage{hyperref}
\usepackage{amsmath,amssymb}
\begin{document}
$0 \href{http://en.wikipedia.org/wiki/Equals\_sign}{=} 1$.

$0 = 1$.
\end{document}
```

If you typeset the above by xelatex, lualatex (or latex + dvipdfmx with `dvipdfmx` argument for `hyperref` package), you get the following result, with the upper equals-sign linked to Wikipedia:

default

Here, we see that both `\href`'ed version of `=` and plain `=` result in the same spacing.
On the other hand, if we drop the math-type handling in KaTeX, we get the following result:



So, to get the same spacing adjustment, we have to set a class of anchor tag to `mord`, `mrel` or so.

> Isn't the second argument of \href treated as if it's in \text? You should be able to use code similar to \text, \raisebox, etc. which switch to text mode.

As above, LaTeX's (more precisely, `hyperref`'s) `\href` commands accept a math as its second argument, not only text. So we can't treat them as if they're in `\text`.

edemaine:
Ah, now I understand where you're getting the classes.  (I was worried that they were being specified by the user, but rather they're grabbed from the rendered arguments.)  Now this seems like the right approach to me.  Cool!

src/functions/href.js
```diff
@@ -0,0 +1,81 @@
...
+            }
+        } else { // No elements at all, just ignore.
+            cls = [];
+        }
+        if (!cls) {
+            const anc = buildCommon.makeAnchor(href, [], elements, options);
+            const elts = [new buildCommon.makeSpan([fst], [], options),
+                anc,
+                new buildCommon.makeSpan([lst], [], options),
+            ];
```
kevinbarabash: Super minor nit, the following formatting might better reflect the new `span` sandwich we got going on here:
```
return new buildCommon.makeFragments([
    new buildCommon.makeSpan([fst], [], options),
    anchor,
    new buildCommon.makeSpan([lst], [], options),
]);
```

src/functions/href.js
```diff
@@ -0,0 +1,81 @@
...
+
+import * as html from "../buildHTML";
+import * as mml from "../buildMathML";
+
+defineFunction({
+    type: "href",
+    names: ["\\href"],
+    props: {
+        numArgs: 2,
+        argTypes: ["original", "original"],
```
kevinbarabash: Should we maybe be using `"text"` for the href so that if we're in math mode and we include a URL with spaces in it, the spaces don't get eaten.

edemaine: Since #912, I don't think the space behavior will be different, but `"text"` does feel slightly more correct.  I'm guessing, in LaTeX, the argument is treated like a verbatim. For example, are two spaces preserved as two spaces or merged into one?  (In KaTeX, they will currently be merged.)  Does `\foo bar` have a space (I guess so) or remove the space (as it will probably currently behave).  This might require a new argument type of `"url"`...

src/domTree.js
```diff
@@ -143,6 +143,133 @@ class span {
...
+    /**
+     * Convert the a into an HTML markup string
+     */
+    toMarkup() {
+        let markup = "<a";
+
+        // Add the href
+        markup += " anchor=\"";
+        markup += utils.escape(this.href);
+        markup += "\"";
```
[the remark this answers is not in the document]
konn: > Since we're adding href to attributes in the constructor shouldn't the loop below that adds all other attributes also add this one?

Ah, yes, that's right.
And I mistyped `anchor` instead of `href` above, and this mistakenly prevents the duplication of `href` attribute. I must fix this.

src/domTree.js
```diff
@@ -143,6 +143,133 @@ class span {
...
+                markup += "\"";
+            }
+        }
+
+        markup += ">";
+
+        // Add the markup of the children, also as markup
+        for (let i = 0; i < this.children.length; i++) {
+            markup += this.children[i].toMarkup();
+        }
```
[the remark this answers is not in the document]
konn: Neat! I will adopt it.

src/functions/href.js
konn: It seems that the first argument of `\href` will be treated as if it's inside verbatim, but both `\href{http://en.wikipedia.org/wiki/Equals\_sign}{=}` and `\href{http://en.wikipedia.org/wiki/Equals_sign}{=}` generate the same link to Wikipedia.

In the current implementation, it rejects `\href{http://en.wikipedia.org/wiki/Equals_sign}{=}` and accepts `\href{http://en.wikipedia.org/wiki/Equals\_sign}{=}`, but it links to `http://en.wikipedia.org/wiki/Equals/_sign` instead of `http://en.wikipedia.org/wiki/Equals_sign`. So it might be appropriate to create new argument type `url` for proper parsing, as @edemaine said.

src/functions/href.js
```diff
@@ -0,0 +1,81 @@
...
+         * 1. if it has the only element, use its class;
+         * 2. if it has more than two elements, and the classes
+         *    of its first and last elements coincide, then use it;
+         * 3. otherwise, we will inject an empty <span>s at both ends,
+         *    with the same classes of both ends of elements.
+         */
+
+        let cls = []; // null if the type of both ends differs.
+        let fst; // mathtype of the first child
+        let lst; // mathtype of the last child
```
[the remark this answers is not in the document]
konn: Oh sorry. This naming convention is one of the bad habit of Haskellers... I will make them more readable.

src/functions/href.js
```diff
@@ -0,0 +1,81 @@
...
+        // Invariants: both fst and lst must be non-null if cls is null.
```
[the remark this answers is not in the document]
konn: I didn't notice that. In that case. I will add the code to handle that case and default to `mord`.

src/functions/href.js
konn: This seems much more appropriate. I will reflect it.

edemaine:
@konn Just checking: Is this ready for review again?

konn:
@edemaine Yes. I think it’s ready.

kevinbarabash:
I'll have another look this weekend.

src/Parser.js
```diff
@@ -853,6 +856,23 @@ export default class Parser {
...
+    parseUrlGroup(optional) {
+        const res = this.parseStringGroup("url", optional);
+        if (!res) {
+            return null;
+        }
+        const raw = res.text;
+        // hyperref package allows backslashes alone in href, but doesn't generate
+        // valid links in such cases; we interpret this as "undefiend" behaviour,
+        // and keep them as-is. In some environment, they're replaced by slashes
+        // in url by browser.
```
kevinbarabash: Are the backslashes being replaced by forward slashes?  In which environments does this happen?

src/functions/href.js
```diff
@@ -0,0 +1,77 @@
...
+
+        const href = group.value.href;
+
+        /**
+         * Determining class for anchors.
+         * 1. if it has the only element, use its class;
+         * 2. if it has more than two elements, and the classes
+         *    of its first and last elements coincide, then use it;
+         * 3. otherwise, we will inject an empty <span>s at both ends,
+         *    with the same classes of both ends of elements.
```
kevinbarabash: Sneaky.  In this third case does the element in between the empty spans have no class on it?

src/functions/href.js
```diff
@@ -0,0 +1,77 @@
...
+                classes = [first];
+            } else { // Case 3: both ends have different types.
+                classes = null;
+            }
+        } else { // No elements at all, just ignore.
+            classes = [];
+        }
+        if (!classes) {
+            const anc = buildCommon.makeAnchor(href, [], elements, options);
+            return new buildCommon.makeFragment([
```
kevinbarabash: This works, but I'd probably move this up to where you set `classes = null` so that we can remove an extra `if` statement.

src/functions/href.js
```diff
@@ -0,0 +1,77 @@
...
+        } else if (elements.length >= 2) {
+            first = html.getTypeOfDomTree(elements[0]) || 'mord';
+            last = html.getTypeOfDomTree(elements[elements.length - 1]) || 'mord';
+            if (first === last) { // Case 2 : type of both ends coincides
+                classes = [first];
+            } else { // Case 3: both ends have different types.
+                classes = null;
+            }
+        } else { // No elements at all, just ignore.
+            classes = [];
```
kevinbarabash: `classes` is already `[]` from `let classes = []` so this `else` block isn't necessary.

src/domTree.js
```diff
@@ -143,6 +143,131 @@ class span {
...
+    toMarkup() {
+        let markup = "<a";
+
+        // Add the href
+        markup += `href="${markup += utils.escape(this.href)}"`;
+        // Add the class
+        if (this.classes.length) {
+            markup += " class=\"";
+            markup += utils.escape(createClass(this.classes));
+            markup += "\"";
```
kevinbarabash: This could be simplified to 
```
markup += ` class="${utils.escape(createClass(this.classes))}"`;
```

src/domTree.js
```diff
@@ -143,6 +143,131 @@ class span {
...
+            markup += " style=\"" + utils.escape(styles) + "\"";
+        }
+
+        // Add the attributes
+        for (const attr in this.attributes) {
+            if (attr !== "href" &&
+                Object.prototype.hasOwnProperty.call(this.attributes, attr)) {
+                markup += " " + attr + "=\"";
+                markup += utils.escape(this.attributes[attr]);
+                markup += "\"";
```
kevinbarabash:
```
markup += ` ${attr}="${utils.escape(this.attributes[attr])}"`
```

kevinbarabash:
@konn sorry about the delay.  Please let me know if you have any questions/concerns about any of the feedback.

src/Parser.js
konn: At least in Safari 11.0.1 (13604.3.5) on macOS High Sierra 10.13.1.

src/domTree.js
konn: Thanks. Template literal looks nicer!

src/domTree.js
konn: I will adopt template literal also here.

src/functions/href.js
konn: Oh, I should elaborate on it. The third case we insert a kind of glue, with the first one having the same class as the first element of body and the second one the same as the last. I will add this description.

src/functions/href.js
konn: Exactly. I will drop this clause.

src/functions/href.js
konn: Oh, yes. It makes program much simpler. Thanks.

konn:
@kevinbarabash No problem, thank you for your review!

> This could also use some unit tests to verify that `\hrefs` parse and don't parse in the expected way, see katex-spec.js. Also, it'd be nice to have a snapshot test for the MathML that we're outputting, see mathml-spec.js (running npm test -- -u will update the snapshot file).

OK, I will try to write some unit-tests.

src/Parser.js
kevinbarabash: Ah... I was getting confused between LaTeX environments and browser environments.  Maybe reword that last sentence to be: `Some browser will replace backslashes with forward slashes.`

kevinbarabash:
> OK, I will try to write some unit-tests.

@konn let me know if you need help with this.

src/Parser.js
konn: Oh yes, it was really confusing expression. I will replace them.

test/katex-spec.js
```diff
@@ -2420,6 +2420,36 @@ describe("An aligned environment", function() {
...
+    it("should not allow unbalanced brace(s) in url", function() {
+        expect("\\href{http://example.com/{a}{bar}").toNotParse();
+        expect("\\href{http://example.com/}a}{bar}").toNotParse();
+    });
+
+    it("should allow escape for letters [#$%&~_^{}]", function() {
+        const url = "http://example.org/~bar/#top?foo=$}foo{&bar=bar^r_boo%20baz";
+        const input = url.replace(/([#$%&~_^{}])/g, '\\$1');
+        const ae = getParsed(`\\href{${input}}{\\alpha}`)[0];
+        expect(ae.value.href).toBe(url);
```
kevinbarabash: Cool!

test/mathml-spec.js
```diff
@@ -58,4 +58,10 @@ describe("A MathML builder", function() {
     it('should use <menclose> for colorbox', () => {
         expect(getMathML("\\colorbox{red}{b}")).toMatchSnapshot();
     });
+
+    it('should set href attribute for href appropriately', () => {
+        expect(getMathML("\\href{http://example.org}{\\alpha}")).toMatchSnapshot();
+        expect(getMathML("p \\Vdash \\beta \\href{http://example.org}{+ \\alpha} \\times \\gamma"))
+            .toMatchSnapshot();
+    });
```
kevinbarabash: I didn't realize that you could have two snapshots inside the same `it` block, cool!

konn:
@kevinbarabash Is there any time to look into this?

src/Parser.js
```diff
@@ -807,6 +810,52 @@ export default class Parser {
...
+                        firstToken.range(this.nextToken, str));
+                } else {
+                    nest -= 1;
+                }
+            }
+            this.consume();
+        }
+        this.mode = outerMode;
+        this.expect(optional ? "]" : "}");
+        return firstToken.range(lastToken, str);
```
kevinbarabash: I understand why we're tracking `lastToken` now and why a recursive solution isn't as easy as I thought it might be.

kevinbarabash:
@konn I've reviewed `parseStringGroupWithBalancedBraces`, looks good.  I will merge this after the tests pass.  Sorry for taking so long with this.

konn:
Thanks!

Issue #921, opened 2017-10-08
linter should complain about tabs in .less files


kevinbarabash:
Currently we use eslint for our .js but we don't do any linting of our .less files.

edemaine:
Perhaps [stylelint](https://stylelint.io/)?  See [this css-tricks.com post](https://css-tricks.com/stylelint/).  Looks like you can turn off tab support via the [indentation rule](https://stylelint.io/user-guide/rules/indentation/).

PR #925, opened 2017-10-09
Lint .less files with stylelint

edemaine:
In particular, prevent tabs, fixing #921.

### DIFF
--- a/.flowconfig
+++ b/.flowconfig
@@ -5,6 +5,7 @@
 
 [libs]
 flow-typed
+./node_modules/stylelint/decls
 
 [lints]
 
--- a/.stylelintrc
+++ b/.stylelintrc
@@ -1,0 +1,6 @@
+{
+    //"extends": "stylelint-config-standard",
+    "rules": {
+        "indentation": 4,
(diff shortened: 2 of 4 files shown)


rrandallcainc:
Should node_modules be added as an ignore in the flowconfig?

edemaine:
@rrandallcainc I don't think so? Wouldn't it be good to test typing across all available modules?

Anyway, not sure that's relevant to this particular PR.

rrandallcainc:
It looks like the new stylelint is causing the build to fail here when running flow. Perhaps a simple solution would be to add an ignore for stylelint for now?

Looks like an open thread on this issue here: https://github.com/stylelint/stylelint/issues/2322

edemaine:
Oops, sorry @rrandallcainc, now I see what you mean!  (I was looking at the Github post on my phone, which isn't the best interface, and didn't see the failed checks.)

As suggested in one of those threads, I added a link to `node_modules/stylelint/decls` to the `libs` section of `.flowconfig` and now it seems to work.

PR #871, opened 2017-09-11
Migrate genfrac, lap, smash, op, mod, and katex into their own files in src/functions

kevinbarabash:
Fixes #869

PR #926, opened 2017-10-09
Append ApplyFunction to math functions

ronkok:
Append a `<mo>&ApplyFunction;</mo>` to the MathML of math functions such as `\sin` or `\tan`.

Ref:  https://www.w3.org/TR/REC-MathML/chap3_2.html#sec3.2.4

### DIFF
--- a/test/__snapshots__/mathml-spec.js.snap
+++ b/test/__snapshots__/mathml-spec.js.snap
@@ -27,6 +27,9 @@ exports[`A MathML builder should generate the right types of nodes 1`] = `
       <mi>
         sin
       </mi>
+      <mo>
+        &amp;ApplyFunction;
+      </mo>
       <mrow>
         <mi>
           x
--- a/test/__snapshots__/mathml-spec.js.snap
+++ b/test/__snapshots__/mathml-spec.js.snap
@@ -28,7 +28,7 @@ exports[`A MathML builder should generate the right types of nodes 1`] = `
         sin
       </mi>
       <mo>
-        &amp;ApplyFunction;
+        ⁡
       </mo>
       <mrow>
         <mi>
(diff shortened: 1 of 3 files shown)


test/__snapshots__/mathml-spec.js.snap
```diff
@@ -27,6 +27,9 @@ exports[`A MathML builder should generate the right types of nodes 1`] = `
       <mi>
         sin
       </mi>
+      <mo>
+        &amp;ApplyFunction;
```
edemaine: I think this should be `&ApplyFunction;`, right? `makeText` above must be escaping the ampersand in a way we don't want...

ronkok: I missed that. Thanks for the comment.

edemaine: Is there a reference to show that `&ApplyFunction;` is equivalent to Unicode Character 'FUNCTION APPLICATION' (U+2061) in MathML?  It seems to be, but I couldn't find one...

ronkok: This [cheat sheet](http://www.html-5.com/cheat-sheet/html-character-codes.html) gives the equivalence. You can also type in `\sin x`  into the LaTeX box at [iMathEQ](http://www.imatheq.com/imatheq/com/imatheq/math-equation-editor-latex-mathml.html) and ask it to give you their version of the equivalent MathML.

ronkok: [This page](https://www.w3.org/TR/xml-entity-names/bycodes.html) is from a better source.

edemaine:
LGTM! Thanks for the pointers.

ronkok:
@kevinbarabash, I think I will shut this PR down and re-submit after PR #871 lands. That will be easier than trying to coordinate the two PRs.

kevinbarabash:
@ronkok thanks.  I'll see what I can do to expedite that review.

PR #928, opened 2017-10-10
Simplify get() now that we don't need it to ignorespaces

edemaine:
Continuation of #912 (forgot to do this change there before merging).

Open question: Should we keep the `get()` alias for `expandNextToken()`, or just use the latter?  `unget()` seems like a useful thing to keep (though it's no longer used), but we could also rename it (e.g. `pushToken`).

### DIFF
--- a/src/MacroExpander.js
+++ b/src/MacroExpander.js
@@ -29,7 +29,7 @@ export default class MacroExpander implements MacroContextInterface {
      */
     future(): Token {
         if (this.stack.length === 0) {
-            this.stack.push(this.lexer.lex());
+            this.pushToken(this.lexer.lex());
         }
         return this.stack[this.stack.length - 1];
     }
@@ -86,7 +86,7 @@ export default class MacroExpander implements MacroContextInterface {
         }
         if (!(isMacro && this.macros.hasOwnProperty(name))) {
             // Fully expanded
-            this.stack.push(topToken);
+            this.pushToken(topToken);
             return topToken;
         }
         const {tokens, numArgs} = this._getExpansion(name);
@@ -224,17 +224,10 @@ export default class MacroExpander implements MacroContextInterface {
     }
 
     /**
-     * Recursively expand first token, then return first non-expandable token.
-     * Equivalent to expandNextToken().
+     * Add a given token to the token stack.  In particular, this get be used
+     * to put back a token returned from one of the other methods.
      */
-    get(): Token {
-        return this.expandNextToken();
-    }
-
-    /**
-     * Undo the effect of the preceding call to the get method.
-     */
-    unget(token: Token) {
+    pushToken(token: Token) {
         this.stack.push(token);
     }
 }
--- a/src/MacroExpander.js
+++ b/src/MacroExpander.js
@@ -16,13 +16,11 @@ export default class MacroExpander implements MacroContextInterface {
     lexer: Lexer;
     macros: MacroMap;
     stack: Token[];
-    discardedWhiteSpace: Token[];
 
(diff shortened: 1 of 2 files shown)


src/Parser.js
```diff
@@ -140,17 +140,18 @@ export default class Parser {
...
+     * Switches between "text" and "math" modes.
+     * This used to reconsume nextToken in the new mode because whitespace
+     * handling differed between the two nodes; now it just updates this.mode.
      */
     switchMode(newMode) {
-        this.gullet.unget(this.nextToken);
+        //this.gullet.unget(this.nextToken);
         this.mode = newMode;
-        this.consume();
+        //this.consume();
```
kevinbarabash: remove commented out code

src/Parser.js
```diff
@@ -140,17 +140,18 @@ export default class Parser {
...
-        this.nextToken = this.gullet.get(false);
+        this.nextToken = this.gullet.get();
     }
 
     /**
-     * Switches between "text" and "math" modes, reconsuming nextToken
-     * in case it would be read differently in the new mode.
+     * Switches between "text" and "math" modes.
+     * This used to reconsume nextToken in the new mode because whitespace
+     * handling differed between the two nodes; now it just updates this.mode.
```
kevinbarabash: Describing the previous behavior isn't necessary.

src/Parser.js
kevinbarabash: I'm curious though why the `unget` and `consume` are no longer necessary.

edemaine: It used to be that `get` got passed true or false according to whether math mode was on, i.e., it ate spaces if it was in math mode.  But that's no longer the case (having moved that into parser, where it belongs according to TeX behavior), so now get doesn't depend on whether math mode is on, so there's no effect to ungetting and regetting.

edemaine:
@kevinbarabash I've removed the excess documentation and commented-out old behavior.

kevinbarabash:
I'll finish reviewing this evening.

src/MacroExpander.js
```diff
@@ -227,36 +225,17 @@ export default class MacroExpander implements MacroContextInterface {
...
     }
 
     /**
      * Undo the effect of the preceding call to the get method.
-     * A call to this method MUST be immediately preceded and immediately followed
-     * by a call to get.  Only used during mode switching, i.e. after one token
-     * was got in the old mode but should get got again in a new mode
-     * with possibly different whitespace handling.
      */
     unget(token: Token) {
```
kevinbarabash: Since it's not being used and it's easy enough to do `this.stack.push(token)`, let's get rid of this.

edemaine: I think it's still a nice interface to keep, in case we want to use it in macros.js. Though still happy to rename it...

kevinbarabash: If you rename it, please change all `this.stack.push(token)`s to use it.

edemaine:
I dropped the `get()` alias (opting for `expandNextToken` instead), renamed `unget()` to `pushToken()`, and used it in two places as suggested by @kevinbarabash (good catch!).  There is one exception: I couldn't easily change `this.stack.push(...expansion)` given the different argument type.  Should I change `pushToken()` to use `arguments`...?  There's also a new doc string for `pushToken` that'd be worth reading.

kevinbarabash:
I think it's fine for `pushToken` to only accept a single token for now.  If we find we're needing to push multiple tokens at once in more places in the future we can revisit at that time.

kevinbarabash:
@edemaine thanks the PR.

PR #930, opened 2017-10-12
Implements the alignedat environment for flexible math spacing

konn:
Implemented `alginat` and `alignat*` environments from `amsmath` package.
They are display math environments, similar to `align` and `align*` environments, but adds no spacing commands such as `\qquad` between maths, which permits users to adjust spacing manually.

Use cases:

```latex
\begin{alignat*}{2} % alignat requires number of maths per row as an argument.
  1 &= 2 &\quad &(\because \bot)\\
  xyz &= zyx && (\text{brabra})
\end{alignat*}

\begin{alignat*}{3}
  x &= 2 &\quad zzz &< 5 & \hspace{1cm} \int_{-\infty}^\infty e^{-x^2} \mathrm{d}x &= \mathord{???}\\
  z - 5 &= w + q && &\hspace{1cm} 1+2+3+4&< 20
\end{alignat*}
```

With LaTeX, we get:

latex-image-2


edemaine:
Given how similar `aligned` and `alignat` are, do you think the code could be shared between the two?  You can list multiple functions in `defineFunction`, then branch within the code (adding space only when in `aligned` mode, say).

Another question: if I understand correctly, `alignat` can appear only in text mode, whereas `alignedat` is the equivalent within math mode.  If this is correct, I think you are implementing the latter (in the same way that KaTeX currently supports `aligned` but not `align`) and it should be renamed accordingly.

konn:
@edemaine Thank you for your rapid reply!

> Given how similar aligned and alignat are, do you think the code could be shared between the two? You can list multiple functions in defineFunction, then branch within the code (adding space only when in aligned mode, say).

I think they can share code. I will try.

> Another question: if I understand correctly, alignat can appear only in text mode, whereas alignedat is the equivalent within math mode. If this is correct, I think you are implementing the latter (in the same way that KaTeX currently supports aligned but not align) and it should be renamed accordingly.

Ah, I overlooked `alignedat` environment. Yes, this should definitely be that.

Anyway, I will resolve these soon. Thanks!

konn:
I've just renamed `alignat` to `alignedat` and factored out duplicated code.
I couldn't find the way to vary the `numArgs` in one `defineEnvrionment` block, so I create new local function `alignedHandler` for that purpose.

kevinbarabash:
Cool!  @konn would you be able to add a screenshot test for this new environment?

konn:
@kevinbarabash OK, I've just added a code fragment for screenshot testing. Unfortunately, I couldn't run docker image appropriately and couldn't get the diff.

kevinbarabash:
@konn yeah getting docker set up is a bit of work.  I can run the tests and send you a PR with the images.

konn:
OK, finally here comes the screenshot diff of `alignedat`s:

alignedat

It seems that KaTeX renders `alignedat` appropriately.

By the way, it seems that KaTeX fails to render `aligned` environment correctly, even before this pull req is applied:

aligned

Should we fix this in a separate pull request, or within this?

kevinbarabash:
> It seems that KaTeX renders alignedat appropriately.

The screenshots look great.

> Should we fix this in a separate pull request, or within this?

Thanks for checking the output against LaTeX's.  Since it's an existing issue a separate PR is appropriate.  Can you open a new issue for this?

kevinbarabash:
@konn I'm going try to review this later this week.

konn:
@kevinbarabash Thank you for everything!

> Since it's an existing issue a separate PR is appropriate. Can you open a new issue for this?
OK, I've just separate this as Issue #941.

> I'm going try to review this later this week.
Thanks!

konn:
Just to caught up with master branch, I've just made a little rebasing on this PR.

kevinbarabash:
Sorry for not having reviewed the code for this yet.  😞  I've been rather busy lately.  Going to try to get to it this weekend though.

kevinbarabash:
@konn thanks for the PR.  Sorry again about the delay.

konn:
@kevinbarabash No problem. Thank you again!

Issue #916, opened 2017-10-02
KaTeX css is not rendering correctly \overline


thiagopnobre:
Sometimes katex just not renders the line of the \overline, depending on the zoom or the font-size.
- Example without zoom:
![bug-katex](https://user-images.githubusercontent.com/5624432/31087556-d6a26e00-a773-11e7-8d5d-1c0cfc55eda7.png)

- Example with zoom:
![bug-katex-02](https://user-images.githubusercontent.com/5624432/31087561-ddba5586-a773-11e7-9ab7-c76465d63a50.png)

thiagopnobre:
Maybe the problem is because of the `em` used in `<span class="overline-line" style="border-bottom-width:0.04em;"></span>`. I think that might be a good idea to change it to `1px` or something like that.

edemaine:
I suspect that this is a further consequence of #824.  We're still at a loss for the right fix.

Note that `1px` in CSS does not mean 1 pixel -- it means 1/96 inches.  (On the other hand, [1px is supposed to be guaranteed to look nice](https://www.w3.org/Style/Examples/007/units.en.html).)  We'd like to match LaTeX's lengths, but we obviously want to get the browser to round up instead of down.

trimbonz:
I can confirm that fraction lines do not appear in old Android webView clients up to and including API 19 (KitKat).  I can work around this by making the fraction line a bit bigger - in katex-min.js, I changed the first entry of the DefaultRuleThickness array from 0.04 to 0.06.  This changes the in-line style for the line to 0.06em, which is apparently enough for it to show up in a WebView.  0.059 doesn't work but 0.5999 does, so there is a threshold there somewhere.

edemaine:
Just-merged #931 should fix the frac line issue, thanks to @ronkok! Please test against master if you can.

kevinbarabash:
@thiagopnobre this should be fixed with the release of [v0.9.0-alpha1](https://github.com/Khan/KaTeX/releases/tag/v0.9.0-alpha1).  Please re-open if this is still an issue.

thiagopnobre:
Sorry for the delay guys. It worked. :)

Issue #824, opened 2017-08-27
Dividing lines for \frac and overhead lines for \sqrt sometimes disappear


akalin:
When trying to upgrade to 0.8.x, I noticed that it sometimes omitted the dividing lines for \frac and the overhead lines for \sqrt.

Here's a test file showing the problem (for dividing lines): https://rawgit.com/akalin/60e78169d60908f8111652540f427607/raw/f53ae1650b12bf6db46b264e8852ded3d82ca7e7/test-0.8.2.html At default zoom, the dividing line for the \fracs in the 2nd expression isn't visible, and if you zoom in and out, you can make one or more of the lines disappear.

Here's the same file using 0.7.1: https://rawgit.com/akalin/60e78169d60908f8111652540f427607/raw/f53ae1650b12bf6db46b264e8852ded3d82ca7e7/test-0.7.1.html If you zoom out far enough, the lines do disappear, but it seems somehow more robust then 0.8.x.

I suspect that the problem isn't unique to 0.8.x, but the change in rendering the superscripts exposed it.

Gist for the test file using other versions: https://gist.github.com/akalin/60e78169d60908f8111652540f427607

kevinbarabash:
Fixed in #810.  See #799.  I'm going to leave this issue open until we publish a new version with the fix.

kevinbarabash:
I should mention that the #810 does not fix the fraction bar issue.  I have not been able to reproduce either issue.  @akalin what browser and OS versions are you using?

akalin:
I'm using Chrome Version 62.0.3192.0 (Official Build) dev (64-bit) on Linux and macOS. Just tried Safari 10.1.2 and Firefox 55.0.3 and  it works fine there. So it might be a Chrome-only issue, or even a dev-channel-only issue.

akalin:
Also works fine on Chrome Version 60.0.3112.101 (Official Build) (64-bit) on Windows.

akalin:
On Windows, works fine for Version 61.0.3163.59 (Official Build) beta (64-bit) but not Version 62.0.3192.0 (Official Build) dev (64-bit). So it looks like a dev-channel-only issue.

kevinbarabash:
The `\sqrt` issue is fixed in [v0.8.3](https://github.com/Khan/KaTeX/releases/tag/v0.8.3).

akalin:
[...]

PR #931, opened 2017-10-12
Prevent disappearing fraction lines.

ronkok:
Set a CSS min-height that will ensure that fraction lines will display at least one screen pixel.

Addresses issues #824 and #916.

### DIFF
--- a/static/katex.less
+++ b/static/katex.less
@@ -313,7 +313,7 @@
 
     // Modify min-height for retina displays. Don't force two screen pixels.
     @media screen and (-webkit-min-device-pixel-ratio: 2.0),
-        screen and (min-resolution: 2dppx) {
+        screen and (min-resolution: 192dpi) {
         .mfrac .frac-line,
         .overline .overline-line,
         .underline .underline-line {
--- a/static/katex.less
+++ b/static/katex.less
@@ -301,6 +301,26 @@
         }
     }
 
+    // Set a min-height for some horizontal lines so their display
+    // will show at least one screen pixel.
+    @media screen {
+        .mfrac .frac-line,
+        .overline .overline-line,
+        .underline .underline-line {
+            min-height: 1px;
+        }
+    }
+
+    // Modify min-height for retina displays. Don't force two screen pixels.
+    @media screen and (-webkit-min-device-pixel-ratio: 2.0),
+        screen and (min-resolution: 2dppx) {
+        .mfrac .frac-line,
+        .overline .overline-line,
+        .underline .underline-line {
+            min-height: 0.5px;
+        }
+    }
+
     .mspace {
         display: inline-block;
 


kevinbarabash:
@ronkok cool!  When zoomed out the some layouts may look a little more cramped.  Could you post some screenshots to show what that looks like at small font sizes and/or zoomed out?

ronkok:
I'll have some screenshots ready tomorrow. Also, I should point out that IE does not support `dppx`. So IE's `min-height` will always be 1px. I'm not troubled by that. I expect few IE users have a high resolution display.

kevinbarabash:
> Also, I should point out that IE does not support `dppx`.

Do you know if Edge supports `dppx`?

ronkok:
Yes, Edge does support `dppx`.

ronkok:
This post shows some screen shots to demonstrate the effect of this PR. The image in the screen shots is a series of fractions, written with font-size from 16px down to 7px in 1px increments. All of the screenshots were taken with the browser set to 100% magnification. Then each shot was enlarged to 400% in Microsoft Paint.

The first two screen shots are done to show the efficacy of this PR. The first screen shot shows the current KaTeX master in Chrome, Firefox, and IE 11 on a screen with resolution = 166 dpi. Notice the missing fraction bars in Chrome.
master-166dpi

The second screen shot is on the same browsers, on the same screen, but includes this PR. All the bars are present. Hurray!
pr-166dpi

The next two screen shots are done to investigate whether this PR creates frac bars that are too thick on a higher resolution screen. Remember, the goal of this PR is ensure at least one screen pixel, but other than that, leave all frac-bar thickness determination to the 0.04 em calculation in KaTeX master.

The next (third) screen shot shows the current KaTeX master in Chrome, Firefox, Opera, and Edge on a screen with resolution = 272 dpi. It all looks good.
master-277dpi
The final screen shot is on the same screen. It shows the result of this PR. It all looks good except that some Edge lines have thickened to two screen pixels.
pr-277dpi
So we have a tradeoff of some thickened Edge lines vs. some missing Chrome lines. I think this PR is a net gain. 

I've also put up a [test page](http://copyasmarkup.com/katex/Investigate.html) so others can look at the results. I'm particularly interested in Safari.

kevinbarabash:
@ronkok thanks for all the screenshots comparing all the browsers before/after.  I agree, this is definitely a net win.

static/katex.less
```diff
@@ -301,6 +301,26 @@
...
+        .mfrac .frac-line,
+        .overline .overline-line,
+        .underline .underline-line {
+            min-height: 1px;
+        }
+    }
+
+    // Modify min-height for retina displays. Don't force two screen pixels.
+    @media screen and (-webkit-min-device-pixel-ratio: 2.0),
+        screen and (min-resolution: 192dpi) {
```
kevinbarabash: I'm going to merge this, but I'm curious about the choice of dpi... is it b/c Window's default 1x resolution is 96dpi?

ronkok: It's not just a Windows thing. According to [this post](https://www.w3.org/blog/CSS/2012/06/14/unprefix-webkit-device-pixel-ratio/),192 dpi and 2 dppx are the same thing. Now, I know that a blog post from W3C is not as authoritative as a W3C standard, but I still take it as pretty good information.

And by using dpi, we get support from IE, which does not support dppx.

kevinbarabash: Interesting.  Hmm... I wonder if we can we can supply both `dppx` values and `dpi` values for `min-resolution` in that order.  Not sure if there's a browser out there that supports `dppx` but doesn't do the correct thing with `dpi` as described in the "caveats" section of the blog post.

edemaine:
Great idea @ronkok!

Given that this was just an issue on Chrome (and maybe recent Chrome?), I wonder if we should submit an upstream bug report too?

PR #934, opened 2017-10-15
Port stretchy to @flow.

marcianx:


src/stretchy.js
```diff
@@ -227,13 +234,12 @@ const svgSpan = function(group, options) {
...
-                spans.push(span);
             }
+            spans.push(span);
         }
 
-        if (numSvgChildren > 1) {
-            span = buildCommon.makeSpan(["stretchy"], spans, options);
-        }
+        span = numSvgChildren === 1 ? spans[0] :
+            buildCommon.makeSpan(["stretchy"], spans, options);
```
marcianx: This is to convince the compiler that `span` is indeed set. Also, unlike before this change, `span` is set also for `(numSvgChildren === 0)`. Unsure if that ever happens.

kevinbarabash: We should probably be checking if `span` is defined below instead.  It's possible that we've messed up somewhere and the SVG path we're looking for doesn't exist.  We should probably throw on that and ask people to report it as a bug.

src/stretchy.js
```diff
@@ -227,13 +234,12 @@ const svgSpan = function(group, options) {
             } else {
                 span = buildCommon.makeSpan([widthClass], [svgNode], options);
                 span.style.height = height + "em";
-                spans.push(span);
             }
+            spans.push(span);
```
kevinbarabash: Why did this get moved out of the `else` block?

src/stretchy.js
```diff
@@ -9,43 +10,47 @@ import buildCommon from "./buildCommon";
...
-    xtwoheadleftarrow: "\u219e",
-    xtwoheadrightarrow: "\u21a0",
-    xLongequal: "=",
-    xtofrom: "\u21c4",
+import type Options from "./Options";
+import type ParseNode from "./ParseNode";
+import type {span} from "./domTree";
+
+const stretchyCodePoint: {[string]: string} = {
+    "widehat": "^",
```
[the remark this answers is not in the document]
marcianx: Removed all added quotes.

src/stretchy.js
```diff
@@ -98,56 +103,58 @@ const mathMLnode = function(label) {
...
-    xrightharpoonup: [["rightharpoon"], 0.888, 522, "xMaxYMin"],
-    xrightharpoondown: [["rightharpoondown"], 0.888, 522, "xMaxYMin"],
-    xLongequal: [["longequal"], 0.888, 334, "xMinYMin"],
-    xtwoheadleftarrow: [["twoheadleftarrow"], 0.888, 334, "xMinYMin"],
-    xtwoheadrightarrow: [["twoheadrightarrow"], 0.888, 334, "xMaxYMin"],
-
-    overleftrightarrow: [["leftarrow", "rightarrow"], 0.888, 522],
-    overbrace: [["leftbrace", "midbrace", "rightbrace"], 1.6, 548],
-    underbrace: [["leftbraceunder", "midbraceunder", "rightbraceunder"],
+    "overrightarrow": [["rightarrow"], 0.888, 522, "xMaxYMin"],
```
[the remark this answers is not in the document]
marcianx: Same here.

src/stretchy.js
marcianx: So that I can access it in the `span =` assignment below. Flow wasn't convinced that `span` was guaranteed to be assigned otherwise.

kevinbarabash: That seems really weird as the `span` assignment is in both the `if` and `else` block.  I'll check out the branch and play around with it.

marcianx: I agree, I don't like that either. I can actually fix that really easily by using a block local `span` variable to make that clearer.

src/stretchy.js
ronkok: `numSvgChildren` is always 1, 2, or 3. If I get creative, may I can think of reasons for `numSvgChildren` > 3 but I can't think of any reasons for `numSvgChildren` = 0.

marcianx: `numSvgChildren > 3` would mean that `widthClass` would be `undefined`. Is that okay?
For now, I've updated the code with an error requiring 1 <= numSvgChildren <= 3.

src/stretchy.js
```diff
@@ -223,17 +235,16 @@ const svgSpan = function(group, options) {
             svgNode = new domTree.svgNode([path], attributes);
 
             if (numSvgChildren === 1) {
-                span = buildCommon.makeSpan(["hide-tail"], [svgNode], options);
+                spans.push(buildCommon.makeSpan(["hide-tail"], [svgNode], options));
```
kevinbarabash: What if we just return `buildCommon.makeSpan(["hide-tail"], [svgNode], options)` here immediately?  Then we wouldn't have to special case things below.

ronkok: We can't return the span yet at that point. We still have to set some height and min-width properties later in the function.

kevinbarabash: My bad, I didn't see that github had collapsed some of the code.  Nevermind.

kevinbarabash:
@marcianx sorry for the delay.  Thank you for continue to push forward `@flow` conversion work.

PR #935, opened 2017-10-15
Port utils to @flow.

marcianx:


src/utils.js
```diff
@@ -27,21 +28,21 @@ const indexOf = function(list, elem) {
...
-const contains = function(list, elem) {
+const contains = function<T>(list: Array<T>, elem: T) {
     return indexOf(list, elem) !== -1;
 };
 
 /**
  * Provide a default value if a setting is undefined
  */
-const deflt = function(setting, defaultIfUndefined) {
+const deflt = function<T>(setting: T | void, defaultIfUndefined: T): * {
```
marcianx: Unfortunately, I couldn't use `T` as the output type itself as I ran into https://github.com/facebook/flow/issues/5022.

https://flow.org/try/#0MYewdgzgLgBAJgUwGYBsoEYYF4ZIK5jBQCW4APACoB8AFBAlCWAOYBcMFMAPjAG4jE4AGnjIAhnjQBJJAFUwiJMTAI47CgEp2AKhgBvAFAxjMAE4M8psDHqNlzbFhwFFy1TAD8opBOlyFyG5wMOy2TMwA3AYAvlGgkLCKaABM2LgERKRglLRh9urcfALC3r5QMvKuKmocWhz6RibmUJbWeSyOzgFK1Z6lkuX+Ve6hDOFRsQYG8dAwxBAAQmKm6OwARiAgKAhi1jhJGDQugdUiPij0GnHgs-NLpsnrm9u7aQfJR91BZ2IXCFcGIA

src/utils.js
```diff
@@ -55,18 +56,16 @@ const ESCAPE_LOOKUP = {
 
 const ESCAPE_REGEX = /[&><"']/g;
 
-function escaper(match) {
+// Type-safe only if the input string matches ESCAPE_REGEX.
+function escaper(match: string): string {
```
kevinbarabash: It's possible that `escaper` could return `undefined`.

src/utils.js
```diff
@@ -27,21 +28,21 @@ const indexOf = function(list, elem) {
 /**
  * Return whether an element is contained in a list
  */
-const contains = function(list, elem) {
+const contains = function<T>(list: Array<T>, elem: T) {
```
kevinbarabash: The return value for this could be `boolean`.

marcianx: Whoops. I wish flow would catch missing types.

src/utils.js
marcianx: Done.

src/utils.js
marcianx: Not given the comment I'd added above `escaper()`. It's used only for `escape()`. In fact, I removed this function and just inlined it where it's used.

src/utils.js
kevinbarabash: We could start using https://github.com/gajus/eslint-plugin-flowtype#eslint-plugin-flowtype-rules-require-return-type.

marcianx: Yes please!

PR #936, opened 2017-10-15
Export svgGeometry to @flow.

marcianx:
Straightforward constant-only file. I was explicit with the type as I've become more paranoid about flow's type inference. However, I'm also happy to remove it if you prefer.

### DIFF
--- a/src/svgGeometry.js
+++ b/src/svgGeometry.js
@@ -1,9 +1,10 @@
+// @flow
 /**
  * This file provides support to domTree.js
  * It's a storehouse of path geometry for SVG images.
  */
 
-const path = {
+const path: {[string]: string} = {
     // sqrtMain path geometry is from glyph U221A in the font KaTeX Main
     sqrtMain: `M95 622c-2.667 0-7.167-2.667-13.5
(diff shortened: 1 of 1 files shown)


src/svgGeometry.js
```diff
@@ -1,51 +1,52 @@
...
 /**
  * This file provides support to domTree.js
  * It's a storehouse of path geometry for SVG images.
  */
 
-const path = {
+const path: {[string]: string} = {
     // sqrtMain path geometry is from glyph U221A in the font KaTeX Main
-    sqrtMain: `M95 622c-2.667 0-7.167-2.667-13.5
+    "sqrtMain": `M95 622c-2.667 0-7.167-2.667-13.5
```
kevinbarabash: I tried this in the flow demo and it works without the quotes: https://flow.org/try/#0PQKgBAAgZgNg9gdzCYAoVBjOA7AzgFzCjjgC4wBvAbQICcBLbAcwF1y7GmBfMAXktRgwAD3IByYWIA0gsAE9xc6ai4BuVEA

marcianx: Updated to remove all the quotes I introduced in this commit. I was just a little surprised that all these unquoted keys worked as I'm used to the Google Closure Compiler that renames struct fields but not string keys.

PR #938, opened 2017-10-15
Port domTree and buildCommon to @flow.

### LINKED ISSUE #939: Port buildCommon and domTree to @flow.

marcianx:
Fixes #939.

### DIFF
--- a/src/buildCommon.js
+++ b/src/buildCommon.js
@@ -14,12 +14,7 @@ import type Options from "./Options";
 import type ParseNode from "./ParseNode";
 import type {CharacterMetrics} from "./fontMetrics";
 import type {Mode} from "./types";
-import type {
-    DomContainerNode,
-    DomChildNode,
-    CombinableDomNode,
-} from "./domTree";
-
+import type {DomChildNode, CombinableDomNode} from "./domTree";
 
 // The following have to be loaded from Main-Italic font, using class mainit
 const mainitLetters = [
@@ -34,6 +29,7 @@ const mainitLetters = [
  */
 const lookupSymbol = function(
     value: string,
+    // TODO(#963): Use a union type for this.
     fontFamily: string,
     mode: Mode,
 ): {value: string, metrics: ?CharacterMetrics} {
@@ -55,6 +51,7 @@ const lookupSymbol = function(
  * TODO: make argument order closer to makeSpan
  * TODO: add a separate argument for math class (e.g. `mop`, `mbin`), which
  * should if present come first in `classes`.
+ * TODO(#953): Make `options` mandatory and always pass it in.
  */
 const makeSymbol = function(
     value: string,
@@ -101,12 +98,14 @@ const makeSymbol = function(
 /**
  * Makes a symbol in Main-Regular or AMS-Regular.
  * Used for rel, bin, open, close, inner, and punct.
+ *
+ * TODO(#953): Make `options` mandatory and always pass it in.
  */
 const mathsym = function(
     value: string,
     mode: Mode,
-    options: Options,
-    classes: string[],
+    options?: Options,
+    classes?: string[] = [],
 ): domTree.symbolNode {
     // Decide what font to render the symbol in by its entry in the symbols
     // table.
@@ -231,25 +230,22 @@ const tryCombineChars = function(
  * Calculate the height, depth, and maxFontSize of an element based on its
  * children.
  */
-const sizeElementFromChildren = function(elem: DomContainerNode) {
+const sizeElementFromChildren = function(
+    elem: domTree.span | domTree.documentFragment,
+) {
     let height = 0;
     let depth = 0;
     let maxFontSize = 0;
 
-    if (elem.children) {
-        for (let i = 0; i < elem.children.length; i++) {
-            if (elem.children[i].height != null &&
-                elem.children[i].height > height) {
-                height = elem.children[i].height;
-            }
-            if (elem.children[i].depth != null &&
-                elem.children[i].depth > depth) {
-                depth = elem.children[i].depth;
-            }
-            if (elem.children[i].maxFontSize != null &&
-                elem.children[i].maxFontSize > maxFontSize) {
-                maxFontSize = elem.children[i].maxFontSize;
-            }
+    for (const child of elem.children) {
+        if (child.height > height) {
+            height = child.height;
+        }
+        if (child.depth > depth) {
+            depth = child.depth;
+        }
+        if (child.maxFontSize > maxFontSize) {
+            maxFontSize = child.maxFontSize;
         }
     }
 
@@ -261,8 +257,8 @@ const sizeElementFromChildren = function(elem: DomContainerNode) {
 /**
  * Makes a span with the given list of classes, list of children, and options.
  *
- * TODO: Ensure that `options` is always provided (currently some call sites
- * don't pass it) and make the type below mandatory.
+ * TODO(#953): Ensure that `options` is always provided (currently some call
+ * sites don't pass it) and make the type below mandatory.
  * TODO: add a separate argument for math class (e.g. `mop`, `mbin`), which
  * should if present come first in `classes`.
  */
@@ -317,9 +313,17 @@ const makeFragment = function(
 };
 
 
+// These are exact object types to catch typos in the names of the optional fields.
 type VListElem = {|
     type: "elem",
     elem: DomChildNode,
+    marginLeft?: string,
+    marginRight?: string,
+|};
+type VListElemAndShift = {|
+    type: "elem",
+    elem: DomChildNode,
+    shift: number,
     marginLeft?: string,
     marginRight?: string,
 |};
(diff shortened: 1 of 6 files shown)


marcianx:
This PR is now ready as well.

src/buildCommon.js
```diff
@@ -248,11 +299,21 @@ const makeFragment = function(children) {
 };
 
 
-// TODO(#939): Uncomment and use VListParam as the type of makeVList's first param.
-/*
-type VListElem =
-    {type: "elem", elem: DomChildNode, marginLeft?: string, marginRight?: string};
-type VListKern = {type: "kern", size: number};
+// These are exact object types to catch typos in the names of the optional fields.
```
kevinbarabash: Love it!

src/buildCommon.js
```diff
@@ -248,11 +299,21 @@ const makeFragment = function(children) {
...
+type VListElem = {|
+    type: "elem",
+    elem: DomChildNode,
+    marginLeft?: string,
+    marginRight?: string,
+|};
+type VListElemAndShift = {|
```
kevinbarabash: That's unfortunate that intersections don't work with exact types.

src/buildCommon.js
```diff
@@ -261,45 +322,44 @@ type VListChild = VListElem | VListKern;
...
-        const oldChildren = children;
-        children = [oldChildren[0]];
 
-        // Add in kerns to the list of children to get each element to be
+// Computes the updated `children` list and the overall depth.
+//
+// This helper function for makeVList makes it easier to enforce type safety by
+// allowing early exits (returns) in the logic.
+const getVListChildrenAndDepth = function(params: VListParam): {
+    children: (VListChild | VListElemAndShift)[] | VListChild[],
```
kevinbarabash: This should probably be defined as a type somewhere.  Also, what's the benefit of typing both the situation where there's a mix of `VListChild` and `VListElemAndShift` in the `children` array as well as the the case where `children` is all `VListChild`s.

src/buildCommon.js
```diff
@@ -308,30 +368,50 @@ const makeVList = function({positionType, positionData, children}, options) {
...
-            if (children[i].type === "kern") {
-                bottom -= children[i].size;
-            } else {
-                bottom -= children[i].elem.height + children[i].elem.depth;
-            }
+        let bottom = params.positionData;
+        for (const child of params.children) {
+            bottom -= child.type === "kern" ?
+                child.size :
+                child.elem.height + child.elem.depth;
```
kevinbarabash: Style nit, if a ternary is too long to fit on a single line can you format it as:
```
result = condition
   ? truthyResult
   : falsyResult;
```
I find it makes it easier to tell that it's a ternary.

src/buildCommon.js
```diff
@@ -308,30 +368,50 @@ const makeVList = function({positionType, positionData, children}, options) {
...
-    } else if (positionType === "firstBaseline") {
-        depth = -children[0].elem.depth;
+    } else if (params.positionType === "bottom") {
+        depth = -params.positionData;
     } else {
-        depth = 0;
+        const firstChild = params.children[0];
+        if (firstChild.type !== "elem") {
+            throw new Error('First child must have type "elem".');
+        }
```
kevinbarabash: This is can happen if `firstChild.type === "kern"`?

src/buildCommon.js
```diff
@@ -308,30 +368,50 @@ const makeVList = function({positionType, positionData, children}, options) {
...
+    return {children: params.children, depth};
+};
+
+/**
+ * Makes a vertical list by stacking elements and kerns on top of each other.
+ * Allows for many different ways of specifying the positioning method.
+ *
+ * See VListParam documentation above.
+ */
+const makeVList = function(params: VListParam, options: Options): domTree.span {
```
kevinbarabash: `makeVList` does a lot of stuff, thanks for breaking it up a bit.

src/buildCommon.js
```diff
@@ -423,7 +505,7 @@ const makeVerb = function(group, options) {
 
 // A map of spacing functions to their attributes, like size and corresponding
 // CSS class
-const spacingFunctions = {
+const spacingFunctions: {[string]: {| size: string, className: string |}} = {
```
kevinbarabash: I was confused for a second until I realize that `[string]` was the key.

src/domTree.js
```diff
@@ -199,8 +243,25 @@ const iCombinations = {
...
+    height: number;
+    depth: number;
+    italic: number;
+    skew: number;
+    maxFontSize: number;
+    classes: string[];
+    style: {[string]: string};
+
+    constructor(
+        value?: string,
```
kevinbarabash: This doesn't need to be optional, all call sites provide this information.

src/domTree.js
```diff
@@ -339,13 +401,23 @@ class symbolNode {
 /**
  * SVG nodes are used to render stretchy wide elements.
  */
-class svgNode {
-    constructor(children, attributes) {
+class svgNode implements VirtualDomNode {
+    children: SvgChildNode[];
+    attributes: Attribute[];
```
kevinbarabash: In other places we provide `attributes` as `attributes: {[string]: string};` we should do the same here.  Not sure how much work is involved with that.  If it looks like a lot feel free to add a TODO and tackle it later.

marcianx: Done.

marcianx:
Since a few other PRs made related files type-safe, there needed to be more edits to fix these files and others to address flow type errors.

src/stretchy.js
```diff
@@ -259,52 +268,57 @@ const svgSpan = function(group: ParseNode, options: Options): span {
...
     } else {
         // \cancel, \bcancel, or \xcancel
         // Since \cancel's SVG is inline and it omits the viewBox attribute,
         // its stroke-width will not vary with span area.
 
-        let attributes = [["x1", "0"]];
+        let attributes: {[string]: string} = {"x1": "0"};
         const lines = [];
 
         if (label !== "cancel") {
```
marcianx: Honestly, I don't understand the structure of the original logic. However, I've replicated it herein the conversion to object.

src/stretchy.js
```diff
@@ -154,98 +153,108 @@ const groupLength = function(arg: ParseNode): number {
...
-            svgNode = new domTree.svgNode([path], attributes);
-
-            if (numSvgChildren === 1) {
-                spans.push(buildCommon.makeSpan(["hide-tail"], [svgNode], options));
-            } else {
                 const span = buildCommon.makeSpan([widthClass], [svgNode], options);
-                span.style.height = height + "em";
-                spans.push(span);
+                if (numSvgChildren === 1) {
+                    return {span, minWidth, height};
```
marcianx: You suggested this in the stretchy PR. I did this anyway to resolve some of the type errors. (It was otherwise interpreting the returned span to have type `span | svgNode`.

src/stretchy.js
```diff
@@ -154,98 +153,108 @@ const groupLength = function(arg: ParseNode): number {
...
-    let attributes = [];
-    let height;
-    let viewBoxWidth = 400000;  // default
-    let viewBoxHeight = 0;
-    let minWidth = 0;
-    let path;
-    let paths;
-    let pathName;
-    let svgNode;
-    let span;
```
marcianx: I made all these variables as local as possible since it was getting difficult for me to otherwise reason about their area of influence of any assignment to any of these variables.

src/functions/op.js
```diff
@@ -132,48 +152,29 @@ const htmlBuilder = (group, options) => {
...
                 ],
             }, options);
-        } else if (!supGroup && !subGroup) {
+        } else {
             // This case probably shouldn't occur (this would mean the
             // supsub was sending us a group with no superscript or
             // subscript) but be safe.
             return base;
-        } else {
-            bottom = options.fontMetrics().bigOpSpacing5 +
```
marcianx: I moved this to the top to make all the if statements positive so that type refinement could guarantee what was non-nullable in each if-then statement body.

src/functions/op.js
```diff
@@ -90,38 +90,58 @@ const htmlBuilder = (group, options) => {
...
         if (supGroup) {
-            newOptions = options.havingStyle(style.sup());
-            supm = html.buildGroup(supGroup, newOptions, options);
+            const elem = html.buildGroup(
+                supGroup, options.havingStyle(style.sup()), options);
 
-            supKern = Math.max(
-                options.fontMetrics().bigOpSpacing1,
-                options.fontMetrics().bigOpSpacing3 - supm.depth);
+            sup = {
```
marcianx: I coupled the `elem` (originally `subm`, `supm`) variable with the corresponding `kern` (originally `subKern`, `supKern`) so that type refinement (`if` checks) below could guarantee when both these values were set and usable.

src/buildCommon.js
```diff
@@ -9,6 +10,12 @@ import fontMetrics from "./fontMetrics";
 import symbols from "./symbols";
 import utils from "./utils";
 
+import type Options from "./Options";
+import type ParseNode from "./ParseNode";
```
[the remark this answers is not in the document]
kevinbarabash: I didn't realize that you could import a class from another module as a `type`, cool.

src/buildCommon.js
kevinbarabash: I guess flow would need to introduce some new operate to extend exact types.

src/buildCommon.js
kevinbarabash: One thing we might consider doing is to switch from `string[]` to `Array<string>` for arrays.  I've opened https://github.com/Khan/KaTeX/issues/987 to discuss it.

src/stretchy.js
```diff
@@ -154,98 +153,107 @@ const groupLength = function(arg: ParseNode): number {
...
-    }
+            for (let i = 0; i < numSvgChildren; i++) {
+                const path = new domTree.pathNode(paths[i]);
+
+                const svgNode = new domTree.svgNode([path], {
+                    "width": "400em",
+                    "height": height + "em",
+                    "viewBox": `0 0 ${viewBoxWidth} ${viewBoxHeight}`,
+                    "preserveAspectRatio": aligns[i] + " slice",
+                });
```
kevinbarabash: This feels more idiomatic.

src/stretchy.js
kevinbarabash: I think this is the "strikethrough" case.

kevinbarabash: This probably could be simplified by moving not reusing the same `attributes` object for both diagonals.  Maybe
```
if (/^x?cancel$/.test(label)) {
    lines.push(new domTree.lineNode({
        x1: 0,
        y1: 0,
        x2: "100%",
        y2: "100%",
    });
}

if (/^[bx]cancel$/.test(label)) {
    lines.push(new domTree.lineNode({
        x1: 0,
        y1: "100%",
        x2: "100%",
        y2: 0,
    });
}
```

marcianx: Ok, I'll simplify it as a tiny PR separately after doing more paranoid checks about what pathways can actually end up in the `else` block that contains this code.

PR #940, opened 2017-10-16
Change buildCommon.makeVList params to struct for better type-safety.

marcianx:
This is towards #939.

The expected structure of the makeVList params depends on the `positionType` parameter. To allow this to be strictly-typed using flow, this PR combines the co-related parameters into a single struct.

### DIFF
--- a/src/buildCommon.js
+++ b/src/buildCommon.js
@@ -286,7 +286,7 @@ const makeFragment = function(children) {
  *  - options: An Options object
  *
  */
-const makeVList = function(children, positionType, positionData, options) {
+const makeVList = function({positionType, positionData, children}, options) {
     let depth;
     let currPos;
     let i;
--- a/src/buildCommon.js
+++ b/src/buildCommon.js
@@ -247,44 +247,45 @@ const makeFragment = function(children) {
     return fragment;
 };
 
+
+// TODO(#939): Uncomment and use VListParam as the type of makeVList's first param.
+/*
+type VListElem =
+    {type: "elem", elem: DomChildNode, marginLeft?: string, marginRight?: string};
+type VListKern = {type: "kern", size: number};
+
+// A list of child or kern nodes to be stacked on top of each other (i.e. the
+// first element will be at the bottom, and the last at the top).
+type VListChild = VListElem | VListKern;
+
+type VListParam = {|
+    // Each child contains how much it should be shifted downward.
+    positionType: "individualShift",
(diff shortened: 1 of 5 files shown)


kevinbarabash:
Why does adding flow types require that we switch to a struct for the args?

marcianx:
The types of the other parameters depends on the value of `positionType`. For example,

- `"shift"`, `"top"`, `"bottom"`, and other require `positionData: number`.
- `"positionShift"` requires `children` to be an array of `{type: "elem", elem: DomChildNode, shift: number}`.

(`DomChildNode` is defined in a local WIP commit.) All these implicit constraints are in the implementation of `makeVList`. I don't know of a way to constrain one parameter like this based on another. But we can achieve the same thing using type union. E.g.,

```javascript
type VListParam = {|
  positionType: "positionShift",
  children: {type: "elem", elem: DomChildNode, shift: number},
|} | {|
  positionType: "top" | "bottom",
  positionData: number,
  children: ...
|} | ...;

const makeVList = function(param: VListParam, options: Options): domTree.span {
   ...
};
```

kevinbarabash:
> The types of the other parameters depends on the value of `positionType`

That makes sense.

kevinbarabash:
Not sure why some of the screenshot tests are failing, but my guess is that something is causing an exception somewhere.  If that's the cause we should add some unit tests the trigger the same issue and then fix the problem.  Also, it might be nice to have `type VListParam` defined as part of this diff, wdyt?

marcianx:
Added `VListParam` (in comments until ported to flow). I did local (non-docker) screenshots, and [the diff of the before and the after](https://github.com/marcianx/KaTeX/compare/marcianx:screenshots-orig...marcianx:screenshots) shows that something's wrong with one or more of the VList kerns. I looked through the code carefully and I couldn't spot my error.

I'll look again tomorrow, but if you know of any ways to quickly check precisely where the problem is based on your knowledge of the implementation or the testing tools, I'm all ears.

It better not be something easily caught by the type-system. ;)
If so, I could rebase my other commits on it and see if flow throws an error.

marcianx:
:D There's nothing like using a _real_ diff tool (`meld`) to highlight only the parts that have changed to make the typo painfully obvious. I had a `positonType` instead of a `positionType`. I don't think it's worth adding a test as flow will easily catch this.

kevinbarabash:
@marcianx thanks for the PR.

PR #948, opened 2017-10-24
Change \undertilde to \utilde

ronkok:
In PR #670, I made an error. The function that should have been `\utilde` I named instead `\undertilde`.

There is an `\undertilde` from the `wsuipa` package, but it is a text-mode non-stretchy function. The proper command name is `\utilde`, a math-mode, stretchy function from packge `undertilde`.

This PR fixes my error.

### DIFF
--- a/src/functions.js
+++ b/src/functions.js
@@ -593,7 +593,7 @@ defineFunction([
 // Stretchy accents under the body
 defineFunction([
     "\\underleftarrow", "\\underrightarrow", "\\underleftrightarrow",
-    "\\undergroup", "\\underlinesegment", "\\undertilde",
+    "\\undergroup", "\\underlinesegment", "\\utilde",
 ], {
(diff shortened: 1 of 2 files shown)


kevinbarabash:
Let's do a version bump to 0.11.x for this even though it's a bug fix so that people if anyone was using `\undertilde` it'll be less likely they'll be caught off guard by the change.

kevinbarabash:
@edemaine good question.  I was thinking of jumping to 0.10.0-alpha... we could also do 0.9.0 and just call out this breaking change and the one that was in 0.9.0-alpha.

edemaine:
@kevinbarabash Any objection to merging this to master? (now 1 month later) I don't want to overstep if you're about to release. We could do another 0.9.0 alpha/beta... (In any case, I think releasing is important to get the array bug fix out.)

Issue #688, opened 2017-04-30
How to use katex on angular2 projects?


nitishkumarnitc:
Can somebosy please help me by telling how to use katex in angular 2 app?

kevinbarabash:
I don't have any angular2 experience, but I found http://stackoverflow.com/questions/36370826/how-to-get-mathjax-working-with-angular2 which might help. Instead of calling 
```
MathJax.Hub.Queue(["Typeset", MathJax.Hub, this.el.nativeElement]);
```
in `ngOnChanges` you'd call 
```
katex.render("c = \\pm\\sqrt{a^2 + b^2}", this.el.nativeElement);
```

garciparedes:
Based on this question, I started the development of an open source project to typeset TeX math expressions with Angular. The project is on https://github.com/garciparedes/ng-katex.

You can install the module with:

    npm install ng-katex --save

To add the module to your proyect add the `KatexModule` to `import`'s field of your parent module:
```js
import { NgModule } from '@angular/core';
import { BrowserModule } from '@angular/platform-browser';
import { AppComponent } from './app/app.component';

import { KatexModule } from 'ng-katex';

@NgModule({
  imports: [BrowserModule, KatexModule],
  declarations: [AppComponent],
  bootstrap: [AppComponent]
})

class AppModule {}

platformBrowserDynamic().bootstrapModule(AppModule);
```

And then you can use it as follows:

```js
import { Component } from '@angular/core';

@Component({
  selector: 'my-app',
  template: `<ng-katex [equation]="equation">`
})
export class AppComponent {
  equation: string = ''\sum_{i=1}^nx_i'';
}
```

nitishkumarnitc:
@garciparedes. Thanks a lot.

edemaine:
This sounds great! We should consider linking to this project from KaTeX.

garciparedes:
Hi @edemaine, thank you very much. Do I make a pull request to the `readme.md` or do you prefer to be yours?

kevinbarabash:
@garciparedes please open a pull request to a link to the README.md.

garciparedes:
I made the changes in `readme.md` and opened the https://github.com/Khan/KaTeX/pull/959 pull request.

PR #959, opened 2017-11-11
Added ng-katex link to readme.md

garciparedes:
From issue https://github.com/Khan/KaTeX/issues/688


garciparedes:
[clabot:check]

README.md
```diff
@@ -86,6 +86,12 @@ the same aspect ratio in KaTeX as in LaTeX.  However, because most browsers
 default to a larger font size, this typically means that a 1cm kern in KaTeX
 will appear larger than 1cm in browser units.
 
+## Libraries
```
kevinbarabash: Thanks for adding this section.  I'm sure we'll be expand this to include other libraries/components.

kevinbarabash:
@garciparedes thanks for the PR.

README.md
ronkok: Perhaps [katex-ruby](https://github.com/glebm/katex-ruby) from @glebm would be a good addition.

edemaine: Agreed. Also, wasn't there a KA plugin of some relevance?

PR #960, opened 2017-11-12
Append ApplyFunction to math functions

ronkok:
As described at: https://www.w3.org/TR/REC-MathML/chap3_2.html#sec3.2.4

Character U+2061 is the same as &ApplyFunction;. Ref: https://www.w3schools.com/charsets/ref_html_entities_a.asp


src/functions/operatorname.js
```diff
@@ -69,7 +69,7 @@ defineFunction({
         identifier.setAttribute("mathvariant", "normal");
 
         const operator = new mathMLTree.MathNode("mo",
-            [mml.makeText("&ApplyFunction;", "text")]);
+            [mml.makeText("\u2061", "text")]);
```
edemaine: Maybe worth a comment that this is `<mo>&ApplyFunction;</mo>` as above?

ronkok: Will do.

edemaine:
Thanks again for the PR!

kevinbarabash:
It might be nice to have an option to control the output of MathML.  I like `&ApplyFunction;` because it's easier see what's going on, but the unicode char takes up less space which is preferable in production.

PR #933, opened 2017-10-14
Support Unicode relations

ronkok:
This is the first in a series of PRs to give KaTeX the ability to recognize Unicode character input. The code in this PR follows the style of PR #410.

All the characters in this PR will produce `rel` atoms. I’ll submit PRs for other atom types later.

PR #961, opened 2017-11-12
Support Unicode Big Operators

ronkok:
This is the sixth in a series of PRs to give KaTeX the ability to recognize Unicode character input.

This is the final PR in this series that will submit characters in bulk. After this, there come only a few characters whose alias or atom type was not obvious. I'll turn in PRs for some of them, too, but only after they are vetted.


ronkok:
As in PR #933, the following table shows: (1) the KaTeX function to which each character is mapped, (2) confirmation whether [unicode math](http://ctan.sharelatex.com/tex-archive/macros/latex/contrib/unicode-math/unicode-math-table.tex) and/or John Cook’s [conversion page](https://www.johndcook.com/unidata.js) map it the same way, (3) possible alternates that I did not use.

  

| **Character** | **Code point** | **Maps to** | **unicode math** | **Cook** | **Not used**        |
|:-------------:|:--------------:|:------------|:----------------:|:--------:|:-------------------:|
| ∏             | U+220F         | \\prod      | ✓                | ✓        |                     |
| ∐             | U+2210         | \\coprod    | ✓                | ✓        |                     |
| ∑             | U+2211         | \\sum       | ✓                | ✓        |                     |
| ∫             | U+222B         | \\int       | ✓                | ✓        | \\intop, \\smallint |
| ∬             | U+222C         | \\iint      | ✓                | ✓        |                     |
| ∭             | U+222D         | \\iiint     | ✓                | ✓        |                     |
| ∮             | U+222E         | \\oint      | ✓                | ✓        |                     |
| ⋀             | U+22C0         | \\bigwedge  | ✓                | ✓        |                     |
| ⋁             | U+22C1         | \\bigvee    | ✓                | ✓        |                     |
| ⋂             | U+22C2         | \\bigcap    | ✓                | ✓        |                     |
| ⋃             | U+22C3         | \\bigcup    | ✓                | ✓        |                     |
| ⨀             | U+2A00         | \\bigodot   | ✓                | ✓        |                     |
| ⨁             | U+2A01         | \\bigoplus  | ✓                | ✓        |                     |
| ⨂             | U+2A02         | \\bigotimes | ✓                | ✓        |                     |
| ⨄             | U+2A04         | \\biguplus  | ✓                | ✓        |                     |
| ⨆             | U+2A06         | \\bigsqcup  | ✓                | ✓        |                     |

src/functions.js
```diff
@@ -297,15 +297,24 @@ defineFunction([
...
 }, function(context) {
+    let fName = context.funcName;
+    if (fName.length === 1) {
+        fName = {
+            "∫": "\\int",
+            "\u222c": "\\iint",
+            "\u222d": "\\iiint",
+            "\u222e": "\\oint",
+        }[fName];
+    }
```
edemaine: This feels a little weird, to have the constant in the code there.  It's a little inefficient if this gets run a lot (though perhaps slightly faster as is if it never gets used...).  Thoughts on moving it to a `const` outside the function?

src/functions/op.js
```diff
@@ -221,17 +221,36 @@ defineFunction({
...
+                "\u22c1": "\\bigvee",
+                "\u22c2": "\\bigcap",
+                "\u22c3": "\\bigcap",
+                "\u2a00": "\\bigodot",
+                "\u2a01": "\\bigoplus",
+                "\u2a02": "\\bigotimes",
+                "\u2a04": "\\biguplus",
+                "\u2a06": "\\bigsqcup",
+            }[fName];
+        }
```
edemaine: Same comment as above.

src/functions.js
```diff
@@ -297,15 +297,24 @@ defineFunction([
 
 // No limits, symbols
 defineFunction([
-    "\\int", "\\iint", "\\iiint", "\\oint",
+    "\\int", "\\iint", "\\iiint", "\\oint", "∫", "\u222c", "\u222d", "\u222e",
```
edemaine: Is there a reason that `∫` gets embedded in Unicode and the others are `\u` escape sequences?  Something about the character being available in a default font?

ronkok: Your guess is on the money. I tend to code in the Consolas font, in which ∫ is available, but not the other characters in this PR. If you would like to see some of any of them changed, let me know.

src/functions.js
ronkok: Sure. I'll move it to a constant outside the function.

src/functions/op.js
ronkok: Ditto.

src/functions.js
edemaine: I feel like it'd be more consistent to put them all as `\u` escape sequences, but I don't feel strongly about it.

ronkok: Done.

edemaine:
Thanks! Maybe the wrong code, though because the test fails now.

ronkok:
You would think that I could type ten keystrokes without a typo. But apparently not.

Fixed now.

edemaine:
Yay! Unicode mission nearing completion!

kevinbarabash:
@edemaine thanks for reviewing and merging.

PR #968, opened 2017-11-16
Implements `\mathchoice` function for choosing appropriate definition with regard to the current math style

konn:
This PR implements `\mathchoice` function, which is particularly useful when one defines custom macro for "big operators".
For example:

```latex
\newcommand{\infdisj}{%
  \mathop{%
    \mathchoice{% display
      \bigvee\hspace{-2ex}\bigvee%
    }{          % inline
      \bigvee\hspace{-1.75ex}\bigvee%
    }{          % script
      \bigvee\hspace{-1.4ex}\bigvee%
    }{          % scriptscript
      \bigvee\hspace{-1ex}\bigvee%
}}}
```

Screenshot test shows that there are still differences between LaTeX's and KaTeX's behaviour, but I think it is a problem of the current implementations of `\scriptstyle` and `\scriptscriptstyle` in KaTeX, as there is already mismatch in `StyleSpacing` screenshot test.


konn:
Oops, it seems that I created the pull request on wrong branch... Sorry, I will re-create it.

PR #969, opened 2017-11-16
Implements \mathchoice command

konn:
This PR implements `\mathchoice` function.
I once created PR on the wrong branch. Sorry for the mess.

This is particularly useful when one defines custom macro for "big operators".
For example:

```latex
\newcommand{\infdisj}{%
  \mathop{%
    \mathchoice{% display
      \bigvee\hspace{-2ex}\bigvee%
    }{          % inline
      \bigvee\hspace{-1.75ex}\bigvee%
    }{          % script
      \bigvee\hspace{-1.4ex}\bigvee%
    }{          % scriptscript
      \bigvee\hspace{-1ex}\bigvee%
}}}
```

Screenshot test shows that there are still differences between LaTeX's and KaTeX's behaviour:
mathchoice
mathchoice2

I think it is a problem of the current implementations of `\scriptstyle` and `\scriptscriptstyle` in KaTeX, as there is already mismatch in `StyleSpacing` screenshot test:
stylespacing


test/screenshotter/ss_data.yaml
```diff
@@ -144,6 +144,10 @@ LowerAccent: |
     \end{matrix}
 MathAtom: a\mathrel{\mathop{=}\limits^{\blue ?}}b
 MathAtom2: \mathop{\overline{\mathrm{lim}}}\limits_{x\to\infty}f(x)
+MathChoice: |
+  {\displaystyle\mathchoice{D}{T}{S}{SS}} {\textstyle\mathchoice{D}{T}{S}{SS}} {\scriptstyle \mathchoice{D}{T}{S}{SS}} {\scriptscriptstyle\mathchoice{D}{T}{S}{SS}}
+MathChoice2: |
+  \displaystyle X_{\mathchoice{D}{T}{S}{SS}_{\mathchoice{D}{T}{S}{SS}}}
```
edemaine: Can you merge these two tests (e.g. side-by-side)? I think we try to keep down the number of snapshot images, and these tests render small enough (and they're related) that it's easy to combine them.

src/functions/mathchoice.js
```diff
@@ -0,0 +1,62 @@
...
+        let body = group.value.text;
+        if (style === Style.DISPLAY) {
+            body = group.value.display;
+        } else if  (style === Style.TEXT) {
+            body = group.value.text;
+        } else if  (style === Style.SCRIPT) {
+            body = group.value.script;
+        } else if  (style === Style.SCRIPTSCRIPT) {
+            body = group.value.scriptscript;
+        }
```
edemaine: Given that these lines (44-54) are repeated (25-35), perhaps it makes sense to factor them out as a shared function defined just within and for this module?  Not strictly necessary, but could make for cleaner code.

konn: Exactly. I will define new `chooseMathStyle` function.

test/screenshotter/ss_data.yaml
konn: Sure. I've just combined them into one, and here is the result:

mathchoice

konn:
Oops, I forgot to remove them. Just deleted!

kevinbarabash:
Cool!

Issue #967, opened 2017-11-14
Render smart quotes in math as regular quotes


aterenin:
Many Markdown preprocessors, such as the one that Jekyll and GitHub Pages use, have a "smart quotes" feature.

This changes quotes such as 
```
'
```
to fancier ones such as
```
’
```
which unfortunately interferes with KaTeX' ability to render prime symbols correctly.

I propose either ignoring smart quotes and rendering them as ordinary quotes (i.e. if KaTeX encounters ```’```, it renders it as if it had encountered ```'```) , or introducing an option to do so.

ccorn:
Just a comment: If such things happen, wouldn't such processors also interfere with usage of `_` and the like?
Or are you referring to (Liquid) filters such as [jekyll-katex-block](https://github.com/drewsberry/jekyll-katex-block)?
In both cases the actual problem is that a language intended for only one parser is interpreted by a second parser. This should be addressed by preventing the wrong parser from trying to fiddle with the output from the first.
(I currently work on a server-side KaTeX implementation that cooperates better with kramdown, so that might fix your issue. The current code works, but I consider factoring out the KaTeX-Ruby interface to a separate gem.)

aterenin:
If they do, the interference is minimal - I have lots of large formulas on my blog with `^_&` and other syntax and it all works correctly. Smart quotes is the only thing that gets through.

My site is built using GitHub pages, and I do not have the kind of control over the environment that would allow me to modify Jekyll or enable 3rd party plugins to work around the issue. Even if I did, I wouldn't use any due to potential security implications.

My current solution is to add a small bit of JavaScript to automatically undo the problematic smart quotes client-side. This works, but I feel that smart quotes in particular are common enough that they shouldn't cause KaTeX to throw an error, especially when there is no valid TeX using them in math mode that I'm aware of.

edemaine:
I believe the "right" way to do KaTeX in Markdown is to extract the math separately, run Markdown processing, then stick the math back in (formatted by KaTeX).  This is what I do [here](https://github.com/edemaine/coauthor/blob/master/lib/formats.coffee), and eventually plan to release as a `contrib` module.

[...]

PR #973, opened 2017-11-19
Single-character macros like active characters

edemaine:
This tiny PR adds support for redefining single characters by arbitrary macros, while also simplifying the code slightly.  I don't see a reason not to allow this, and it would be immediately useful for resolving #967, as well as e.g. users adding support for Unicode characters that KaTeX doesn't already support, or which KaTeX already supports but the user wants to override in a different way.  Overall seems like a win.


ronkok:
Cool. This will be helpful for double-struck Unicode characters like ℂ or ℕ.

edemaine:
@ronkok Great! More motivation for this change.

README.md
```diff
@@ -48,7 +48,7 @@ You can provide an object of options as the last argument to `katex.render` and
...
 - `throwOnError`: `boolean`. If `true`, KaTeX will throw a `ParseError` when it encounters an unsupported command. If `false`, KaTeX will render the unsupported command as text in the color given by `errorColor`. (default: `true`)
 - `errorColor`: `string`. A color string given in the format `"#XXX"` or `"#XXXXXX"`. This option determines the color which unsupported commands are rendered in. (default: `#cc0000`)
-- `macros`: `object`. A collection of custom macros. Each macro is a property with a name like `\name` (written `"\\name"` in JavaScript) which maps to a string that describes the expansion of the macro.
+- `macros`: `object`. A collection of custom macros. Each macro is a property with a name like `\name` (written `"\\name"` in JavaScript) which maps to a string that describes the expansion of the macro. The behavior of single characters can also be redefined by a macro (similar to TeX active characters).
```
kevinbarabash: The tests allow single characters for the names, but reading this it sounds like only things like `\name` are accepted.  Maybe change the last sentence to read like: 
```
Single characters keys can also be included in which case the character will be redefined as the given macro (similar to TeX active characters).
```

test/katex-spec.js
```diff
@@ -2603,6 +2603,12 @@ describe("A macro expander", function() {
         expect("X \\implies Y").toBuild();
         expect("X \\impliedby Y").toBuild();
     });
+
+    it("should allow aliasing characters", function() {
+        compareParseTree("x’=c", "x'=c", {
+            "’": "'",
+        });
```
kevinbarabash: Nice test.

README.md
edemaine: Good idea!  Revised.

kevinbarabash:
@edemaine thanks for make this change.  It's really going to help with the unicode effort.

Issue #793, opened 2017-08-13
List LaTeX packages that make LaTeX behave most like KaTeX


edemaine:
KaTeX supports a lot of features from a lot of packages.  We should make a list, either in the README or in a `katex.sty` style file, so that it's easy to port code from KaTeX to LaTeX.  Feel free to list packages you're aware of here.  For starters:

* `amsmath`

(I've mentioned this idea before, but never turned it into an issue.)

edemaine:
I had an idea: extract all the math from the [function-support page](https://khan.github.io/KaTeX/function-support.html) and test it out in LaTeX.  Here's a little script, which I should add to `contrib`:

```javascript
fs = require('fs');
let html = fs.readFileSync('function-support.html', {encoding: 'utf8'});
html = html.replace(/<script>[^]*?<\/script>/ig, '');
const regex = /\\\(.*?\\\)/g;
const maths = [];
let match;
while ((match = regex.exec(html))) {
    maths.push(match[0]);
}
const tex = `\
\\documentclass{article}
\\usepackage{amsmath,amssymb,latexsym,color,MnSymbol,overrightarrow,cancel,extpfeil}

\\begin{document}

${maths.join('\n\n')}

\\end{document}
`;
fs.writeFileSync('function-support.tex', tex);
```

This set of `\usepackage`s mostly works, though may not be minimal (I don't think `extpfeil` should be necessary but I get more errors without it...).  Still get some errors, though:

* `! LaTeX Error: Command \Rightarrowfill@ already defined.` (package incompatibility)
* `\mapstochar` used in `\xmapsto` is undefined (package incompability)
* `\undertilde` doesn't seem to be defined anywhere.  `undertilde.sty` defines `\utilde`...
* `\Bbbk` isn't loading for some reason, though shouldn't it be in `amssymb`?
* `\sout` isn't defined anywhere.
* `\lt` and `\gt` are not defined anywhere?  Weird.  We could add them to `katex.sty`
* Of course `\KaTeX` isn't defined.

ronkok:
I'd suggest making this list into a page on the wiki.

ronkok:
@edemaine  I've added a [wiki page](https://github.com/Khan/KaTeX/wiki/Package-Emulation) that addresses some of your ideas.

Also, your note re: `\undertilde` has made me realize that I made a mistake. The stretchy accent that I added should have been named `\utilde`, not `\undertilde`. I'll submit a PR to fix the error. 

[...]

PR #974, opened 2017-11-19
\kern fixes, \hskip support, \TeX, \LaTeX, \KaTeX

edemaine:
This PR does a lot of cleanup on `\kern`-related functions, and defines `\TeX` and `\LaTeX` macros (fixing #224).  It also proposes a possible replacement to `\KaTeX` (needs discussion).

In more detail:
* Move `\kern`, `\mkern` into `functions` directory (and use new `defineFuction` interface)
* Add `\hskip`, `\mskip` support (but without supporting plus/minus -- this should be a future issue)
* Properly separate `\kern`, `\hskip` from `\mkern`, `\mskip`.
  (The former work in both modes, and don't support mu units.
  The latter work only in math mode and only support mu units.)
* Render horizontal spacing commands (such as `\kern`) using MathML `<mspace>`.
  Previously this was rendered as an empty `<mrow>` with a TODO by @kevinbarabash to figure it out.
* Implement `\TeX` macro identical to LaTeX's definition using `\raisebox` and `\kern`.
* Implement `\LaTeX` macro roughly identical to LaTeX's definition (using `\raisebox` instead of a vertical box with a fill to get top alignment).  Comments welcome, but I think this is a pretty good solution for now.
* I tried implementing a new `\KaTeX` logo using a similar macro (now that we have all these features), instead of the large body of code in `src/functions/katex.js` and `static/katex.less`.  It's currently defined as `\katex` so that you can check out this branch and compare the two:

  katex

  I'd like to suggest that this definition replace the old `\KaTeX`, to reduce code complexity, but this definitely needs discussion.  Note the spacing is a little different, because it seemed natural to me to more closely match the spacing in `\LaTeX`.  Feel free to disagree!

  Incidentally, this is also a step toward implementing `\KaTeX` in LaTeX, as mentioned in #793.  (The existing definition would be challenging to port.)


kevinbarabash:
Cool!

> Properly separate \kern, \hskip from \mkern, \mskip.
(The former work in both modes, and don't support mu units.
The latter work only in math mode and only support mu units.)

I think we'll need to bump the version again as this may break some math that was working for people.

> Render horizontal spacing commands (such as \kern) using MathML <mspace>.
Previously this was rendered as an empty <mrow> with a TODO by @kevinbarabash to figure it out.

Thanks for figuring this out.

> Implement \LaTeX macro roughly identical to LaTeX's definition (using \raisebox instead of a vertical box with a fill to get top alignment). Comments welcome, but I think this is a pretty good solution for now.

Cool.  I think some people will be pretty stoked about this.

> I tried implementing a new \KaTeX logo using a similar macro (now that we have all these features), instead of the large body of code in src/functions/katex.js and static/katex.less. It's currently defined as \katex so that you can check out this branch and compare the two:

I think the new \katex implementation is closer to the \LaTeX rendering, especially the kerning between the `a` and the `T`.  It might look better if there was more space between the `K` and `a`.  Looking at \LaTeX, there's an even amount of space between the `L` and the `a` and the `a` and the `T`.

src/functions/hskip.js
```diff
@@ -0,0 +1,65 @@
+//@flow
+// Horizontal spacing commands
```
kevinbarabash: Maybe call this file `kern.js` b/c `type: "kern"` below... or maybe `horizontal-spacing.js`.

src/functions/hskip.js
```diff
@@ -0,0 +1,65 @@
...
+
+import defineFunction from "../defineFunction";
+import buildCommon from "../buildCommon";
+import mathMLTree from "../mathMLTree";
+import { calculateSize } from "../units";
+import ParseError from "../ParseError";
+
+// TODO: \hskip and \mskip should support plus and minus in lengths
```
kevinbarabash: Good call splitting this out, tbh I'm not sure how to do this, at least for our current HTML layout strategy.

src/macros.js
```diff
@@ -233,6 +229,31 @@ defineMacro("\\thickspace", "\\;");   //   \let\thickspace\;
...
+//          \vbox to\ht\z@{\hbox{\check@mathfonts
+//                               \fontsize\sf@size\z@
+//                               \math@fontsfalse\selectfont
+//                               A}%
+//                         \vss}%
+//         }%
+//         \kern-.15em%
+//         \TeX}
+// This code roughly aligns the top of the A with the T.  Here we approximate
+// this vertical alignment with a \raisebox{.45ex}, which seems close.
```
kevinbarabash: What font size is being selected for the `A`?  Perhaps we could compute the `raisebox` amount based on the difference in size between the `A` and the `T`.

src/macros.js
```diff
@@ -233,6 +229,31 @@ defineMacro("\\thickspace", "\\;");   //   \let\thickspace\;
...
+defineMacro("\\LaTeX", "\\text{L\\kern-.36em\\raisebox{.45ex}{\\scriptsize A}" +
+                       "\\kern-.15em\\TeX}");
+
+// New KaTeX logo based on tweaking LaTeX logo
+defineMacro("\\katex", "\\text{K\\kern-.25em\\raisebox{.45ex}{\\scriptsize A}" +
+                       "\\kern-.15em\\TeX}");
```
kevinbarabash: Does `K\\kern-.36em` even out the spaces around the `A`?

ronkok:
> Properly separate \kern, \hskip from \mkern, \mskip.

I have, so far, purposely omitted `\mkern` from the [function support page](https://khan.github.io/KaTeX/function-support.html) because I thought it might break. Once the behavior of `\mkern` and `\mskip` is more reliable, I'll put them into the documentation.

edemaine:
I need to respond about the other issues (thanks for the review, @kevinbarabash!), but one question I have is whether the errors on misuse of `\kern` (especially) and `\mkern` (less so) should be turned into warnings.  We don't really care about the separation for our purposes, it's just to encourage people to use LaTeX correctly, so maybe a warning is better?  (We could still bump the version number, but maybe unnecessary to break existing code that uses `\kern{..mu}`.)

src/functions/hskip.js
edemaine: Good point -- switched to `kern.js`.

src/functions/hskip.js
edemaine: Yeah, definitely for another time.  Supporting `fil`/`fill`/`filll` units in `plus` at least should be possible using Flexbox.  That'd be pretty useful -- but for another issue.  I've opened #990 to track.

src/macros.js
edemaine: Nice objective to even out those top spaces.  By manual search, I found `K\\kern-.16em` to balance it out pretty evenly:

katex
http://localhost:7936/?text=%5Cbegin%7Barray%7D%7Br%7D%5Ckatex%5C%5C%5CLaTeX%5C%5C%5CKaTeX%5Cend%7Barray%7D

I feel like the this puts the bottom of the A a little far from the K, though.  I tried tweaking the A slightly more to the left, with `K\\kern-1.17em`:

katex

Preference?  This may be nit-picky, but if we're redesigning the logo, we might as well try to get it right.  I do find these far more readable than the old logo, which is nice (but also more readable than the LaTeX logo).

src/macros.js
edemaine: The `A` should be in the equivalent of `\scriptsize`, which has a scale of 70%.  The height of a T is 0.68333em (according to `fontMetricsData.js`), so the raisebox should be `0.3 * 0.68333 = 0.204999`em.  I tried this, and it puts the A slightly higher than what I did, but it matches the LaTeX rendering better!  Good thinking.  I've pushed code that computes this automatically from the font data instead of hard-coding constants (other than 0.3 which seems safe enough).

Compare this screenshot with others on this page:
katex

edemaine:
I addressed @kevinbarabash comments (thanks!) and switched from errors to warning when misusing `mu` units/`mkern` (and rebased to remove conflict).  I kept it an error to use `\mkern`/`\mskip` outside math mode, because this seems natural, and previously no kern/skip commands worked in text mode at all.

src/macros.js
kevinbarabash: I think the top one looks the best.  In the bottom one the `A` is too cramped.

edemaine: Just to be clear, we're talking about the top logo in the top image vs. the top logo in the bottom image, right?  (Bottom logo is the same in both images, the old KaTeX logo.)

kevinbarabash: I can't tell the difference between the top logos in the two images... whichever you think is better.

edemaine:
Here's a texdiff for `\LaTeX` and `\TeX` (red is KaTeX, green is LaTeX):

latex

It appears to me that the A is slightly smaller in KaTeX, which causes the horizontal shift of the L.  I don't understand why; I confirmed that the A is in a font size exactly 70% of the T, and LaTeX says the same in its rendering:

```
\hbox(0.0+0.0)x15.0
\OT1/cmr/m/n/10 L
\kern -3.6
\vbox(6.83331+0.0)x5.90282, glue set 2.04997fil
.\hbox(4.78334+0.0)x5.90282
..\OT1/cmr/m/n/7 A
.\glue 0.0 plus 1.0fil minus 1.0fil
\kern -1.49994
\OT1/cmr/m/n/10 T
\kern -1.66702
\hbox(6.83331+0.0)x6.80557, shifted 2.15277
.\OT1/cmr/m/n/10 E
\kern -1.25
\OT1/cmr/m/n/10 X
spacefactor 1000
```

I also notice that there's some difference between Chrome vs. Firefox rendering (notice height of A):

katex-chrome vs. katex-firefox


Personally I'd be OK with leaving it as is, as I'm running out of ideas (and time) for how to improve it, it matches the definition, and the rendering is reasonably close now.

I've committed this as a new screenshot test so that it can be explored later.  I also switched over `\KaTeX` to the new macro, and deleted the old function code.  So now the latest choice for the KaTeX logo is in the screenshot test.

kevinbarabash:
@edemaine I'm not sure why all of the fonts screenshot tests are failing. 😕

edemaine:
OK, should be all fixed now.

kevinbarabash:
@edemaine thanks for investing and fixing the issue.

PR #975, opened 2017-11-19
\hspace*, \@ifstar, \@ifnextchar, \@firstoftwo

edemaine:
This PR starts to add some more advanced LaTeX macro support using existing features of `MacroParser`.  Getting fun!
* Support both `\hspace` and `\hspace*` (still aliasing to `\kern`) using new `\@ifstar`
* `\@ifstar` is a macro using new `\@ifnextchar` and `\@firstoftwo` (same definition as LaTeX)
* `\@firstoftwo` and `\@ifnextchar` use available `MacroParser` features to act as they do in LaTeX.
* Also new method `pushTokens` (which I almost used but didn't end up), and moved `pushToken` next to it and `popToken`.  Could remove these or the changes, but slightly cleaner?


src/macros.js
```diff
@@ -25,22 +25,50 @@ export interface MacroContextInterface {
...
+    const nextToken = context.future();
+    if (args[0].length === 1 && args[0][0].text === nextToken.text) {
+        return {tokens: args[1], numArgs: 0};
+    } else {
+        return {tokens: args[2], numArgs: 0};
+    }
+});
+
+// \def\@ifstar#1{\@ifnextchar *{\@firstoftwo{#1}}}
+defineMacro("\\@ifstar", "\\@ifnextchar *{\\@firstoftwo{#1}}");
```
kevinbarabash: Why does `\\@firstoftwo{#1}` need to be wrapped in `{...}`?

src/macros.js
```diff
@@ -235,8 +263,8 @@ defineMacro("\\thickspace", "\\;");   //   \let\thickspace\;
 
 // \DeclareRobustCommand\hspace{\@ifstar\@hspacer\@hspace}
 // \def\@hspace#1{\hskip  #1\relax}
-// KaTeX doesn't do line breaks, so \hspace is the same as \kern
-defineMacro("\\hspace", "\\kern{#1}");
+// KaTeX doesn't do line breaks, so \hspace and \hspace* are the same as \kern
+defineMacro("\\hspace", "\\@ifstar\\kern\\kern");
```
kevinbarabash: I would've assumed that `\hspace*` would be aliased to `\hspace` by doing:
```
defineMacro("\\hspace*", "\\hspace")
```
My hope was to one day we might have a script that scrapes the source code to determine all of the commands we support.  This makes it harder.  If this is the way LaTeX defines `*` functions then let's roll with it.

kevinbarabash: If a person wrote:
```
defineMacro("\\hspace", "\\@ifstar{\\kern{#1}\\kern{#1}}\\kern{#1}");
```
would that result in `\hspace*` producing double the amount of kerning?

src/macros.js
edemaine: `\@firstoftwo{#1}` is the second (of three) arguments to `\@ifnextchar`.  If it wasn't wrapped in braces, `\@firstoftwo` would be the second argument and `{#1}` would be the third argument.  (Same as the original TeX source.)

src/macros.js
edemaine: Yes, this is how LaTeX does it.  `\hspace*` is parsed by TeX as two tokens, a control word `\hspace` and a character `*`, and the former detects the latter and adapts its behavior.  Of course, LaTeX also does this (at least internally) to detect optional arguments, and KaTeX deals with optional arguments to functions in a different way, so this is open to debate...  But I don't think LaTeX offers another way to detect asterisks, while it does offer another way to parse optional arguments.

Anyway, `defineMacro("\\hspace*", "\\hspace")` would be difficult to get working, as it leads to ambiguous parsing.  When KaTeX sees an `\hspace` token, it doesn't know whether `*` is an argument to `\hspace` or it's just the beginning of `\hspace*`.

Regarding your second question:
```latex
defineMacro("\\hspace", "\\@ifstar{\\kern{#1}\\kern{#1}}{\\kern{#1}}");
```
should behave as you describe.  (Needed some extra braces, I think.)

src/macros.js
kevinbarabash: @edemaine I don't have much experience with TeX macros.  I appreciate you taking the time to answer my questions.

src/macros.js
kevinbarabash: Good point about the ambiguity of the parsing.

edemaine:
I added some quick tests.  Let me know if you'd like more!

kevinbarabash:
Not sure why the firefox screenshot tests didn't work as I don't think any of the screenshot tests would be impacted by these changes.  I restarted the travis-ci build, hopefully it'll pass this time.

kevinbarabash:
Firefox tests passed this time.  ¯\_(ツ)_/¯

Issue #716, opened 2017-05-26
The componenents of formula cut each other under High Contrast mode in Windows [and Solution]


GeeLaw:
Environment: Windows 10.

## Reproduction Step

1. Go to [KaTeX demo page](https://khan.github.io/KaTeX);
2. Type `f(x) = \int_{-\infty}^\infty\hat f(\xi)\,e^{2 \pi i \xi x} \,d\xi`, which is actually the default formula;
3. Turn on High Contrast and observe.

I am using my own High Contrast scheme. The result is that some parts of the formula are cut by other parts. As seen below:

![Example](https://cloud.githubusercontent.com/assets/4462444/26514603/12dbebf0-4273-11e7-80e1-b8a99ca19361.png)

## Cause

Under High Contrast mode, it is assumed that any background graphics should be removed.* Especially those with transparency or gradient that might reduce visual recognisability. Elements KaTeX uses to build the formula has `transparent` background set and as a result, they are restored back to the scheme-defined background colour.

\* There is a longer story regarding background images. Windows is now smart enough not to remove some images so that the widely used technique to display an image without `img`, without setting appropriate properties in CSS, will still work under High Contrast mode.

## Solution

The solution is simple. We just need to disable this adjust. To do this, add the following rule to CSS:

```css
.katex * { -ms-high-contrast-adjust: none !important; }
```

This disables resetting the background. Note that using this rule will still let `.katex` change its `color` and `background` according to the scheme, which is expected and desired. It can be seen that with the rule added, the formula is now readable.

![Repairment](https://cloud.githubusercontent.com/assets/4462444/26514673/947b2356-4273-11e7-8e71-35b501eea065.png)

## Finally…

I am just too lazy to make a pull request or anything. Please help yourself :-)

edemaine:
Thanks for the report plus solution! I have Windows 10 so will give this a try to confirm, then turn it into a PR.

ronkok:
This is terrific. It fixes an Edge problem I have had in PR #670, where I use `background: linear-gradient` to draw the lines for `\cancel`, `\bcancel`, and `\xcancel`. Thank you for this information.

edemaine:
Just confirmed.  I found that it was enough to add the rule

```css
.katex * { -ms-high-contrast-adjust: none !important; }
```

because [according to the documentation](https://msdn.microsoft.com/en-us/library/hh771863(v=vs.85).aspx) the property is inherited.

PR now available as #724.

[...]

PR #976, opened 2017-11-20
Change frac-line from border to full span

ronkok:
Change `frac-line`  and `vertical-separator`.  Instead of using span borders to render the lines, fill an entire span using a ~~box-shadow~~ __SVG__.

This change will enable more dependable engagement of the `min-height` CSS.


marcianx:
FYI: I'd be wary of using box-shadow for anything since in Windows high contrast mode, browsers remove all box shadows (and also background colors/images). I've totally been bitten by this when working on making accessible web UIs.

ronkok:
Pity. I already knew that background colors were a bad idea when printing. I suppose I could use an SVG image. Well, this will take a little longer.

edemaine:
Is this true even if the settings are marked `!important`?  (relevant to both backgrounds and shadow boxes) It seems a shame to have to use such advanced features for something as simple as a fraction... though I'm not coming up with an alternative?

marcianx:
`!¡mportant` is specifically designed to override CSS precedence. I'm not aware of it offering any other use.

edemaine:
@marcianx It's at least sometimes relevant, but maybe only in ms-specific rules... See https://github.com/Khan/KaTeX/issues/716#issuecomment-308818105

ronkok:
Interesting.

I want to avoid `background-color`, even if there is a way to avoid the high contrast problem. When printing, the default setting is to omit background colors.

But regarding box-shadow, as @edemaine points out, the KaTeX CSS already includes
````
.katex * {
  -ms-high-contrast-adjust: none !important;
}
````

I just finished writing the code for a frac-line via inline SVG image. The code works fine, but I have not made a commit. First, I'll check if box-shadow works when high-contrast is set and also works when printing.

**Comment by ronkok** [comment]
High contrast results are in. The image on the left is the box-shadow method. On the right is the inline SVG image.

box shadow
SVG is the clear winner. I'll turn in a commit.

Thank you to @marcianx for the warning about high contrast.

**Comment by ronkok** [comment]
Well, now there’s a working PR and, thanks to @marcianx, we have overcome one hurdle. At this point, I am the wrong person to do any further detailed testing on this work. I have not been able to reproduce [the problem](https://github.com/Khan/KaTeX/issues/824#issuecomment-344051809) on my own screen and I have never mastered Docker for the Screenshotter tests. Any help would be appreciated.

**Comment by trimbonz** [comment]
Both the box-shadow and SVG methods in this PR fix the issue for my old Android webView clients, that would previously not render any fraction lines up to and including API 19 (KitKat).  Previously I had to tweak the line thicknesses manually to make fraction lines appear, but they all show up without further intervention after I started using the fraclinespan branch.

**Comment by ronkok** [comment]
@trimbonz Cool! Thank you for the test and the report.

**Comment by kevinbarabash** [comment]
@ronkok I can regenerate the screenshots.

File: src/stretchy.js [hunk 1]
```diff
@@ -315,8 +315,22 @@ const encloseSpan = function(
...
+    // Get a big square image. The parent span will hide the overflow.
+    const pathNode = new domTree.pathNode('bigRule');
+    const attributes = [
+        ["width", "400em"],
+        ["height", "400em"],
+        ["viewBox", "0 0 400000 400000"],
+        ["preserveAspectRatio", "xMinYMin slice"],
+    ];
+    const svg =  new domTree.svgNode([pathNode], attributes);
+    return buildCommon.makeSpan([className, "hide-tail"], [svg], options);
```
kevinbarabash: How is the height (or width) set for the `ruleSpan`?

ronkok: The `hide-tail` class contains the CSS `overflow: hidden;`. So the height that gets rendered is the  `line.style.height` set by `buildHTML.makeLineSpan()`.

Class `frac-line` has always contained the CSS `width: 100%;`, so the width is being set the way it always has been set.

ronkok: Clarification: My response above pertained to `\frac`, `\overline`, and `\underline`. I did not change `\rule` in this PR, so it is still using a `top-border` to render a line.

static/katex.less
```diff
@@ -546,7 +544,18 @@
...
+            // Set a min-width so it will display at least one screen pixel.
+            @media screen {
+                min-width: 1px;
+            }
+
+            // Modify min-width for retina displays. Don't force two screen pixels.
+            @media screen and (-webkit-min-device-pixel-ratio: 2.0),
+                screen and (min-resolution: 192dpi) {
+                min-width: 0.5px;
+            }
```
kevinbarabash: Should `.frac-line` have its `min-height` set in a similar fashion?

ronkok: `frac-line` [already has](https://github.com/Khan/KaTeX/blob/master/static/katex.less#L297) that `min-height` set. That was the purpose of PR #931.

kevinbarabash: Thanks for refreshing my memory.  😅

kevinbarabash:
@ronkok thanks for making these changes.

Issue #836, opened 2017-08-31
\! results in extra root element width


dennisss:
When rendering `\$10,\!000,\!000,\!000,\!000`, I expect it to show up with a tight horizontal spacing between the commas and zeros. This works fine, but the root element shows up with extra width (note the blank space in the right side of the blue bounding box below). So, when rendering it inline with text after it, there is extra space between the text and the math. This shows up without the extra space in MathJax

Browser: Chrome, Mac OS X

<img width="804" alt="screen shot 2017-08-31 at 3 33 29 pm" src="https://user-images.githubusercontent.com/1341104/29942627-9235beca-8e64-11e7-8108-4f78cce77d8a.png">

kevinbarabash:
@dennisss I can't reproduce this on https://khan.github.io/KaTeX/.  What version are you using?

dennisss:
@kevinbarabash I'm on Chrome 60.0.3112.113 (the latest version as of now), and the screenshot is from the currently served copy of https://khan.github.io/KaTeX/. It seems to work on Firefox, so possibly isolated to only Chrome.

kevinbarabash:
@dennisss my bad.  I was selecting things using the selection cursor in the browser window instead of using the inspector to select the DOM element.  Yes, I'm seeing this behavior too.

edemaine:
`\fbox{\$10,\!000,\!000,\!000,\!000}` further confirms this.  Time to read up on `width: auto` rules, I guess...

kohler:
“1\!2” is rendered as `<span.mord>1</span><span.mord><span.mspace></span><span.mord>2</span></span>`.

Giving the second top-level `<span.mord>` `display:inline-block`, instead of `display:inline`, solves the problem.

edemaine:
Confirmed. Nice solution @kohler. Given that we currently forbid line breaks in KaTeX renderings, is there any reason not to make all `span.mord`s into `display:inline-block` (same as `span.mspace`)?  Apparently the main difference between `inline` and `inline-block` is that the latter "can have a defined width" which I'm guessing is what's making this work (behaving correctly with negative-width children).

kohler:
I found a much simpler fix.

kohler:
I was wrong about the simple fix. It would be nice to default all katex spans to `inline-block` but this breaks accents.

kohler:
[...]

PR #984, opened 2017-11-23
Avoid negative space in \bmod.

### LINKED ISSUE #982: \bmod in inline math mode adds extra following space
Test file: https://cdn.rawgit.com/akalin/98bc47f5e54f7529462f12a1d7336858/raw/28a50452296c9d7bd4dc9e5023056cf64412a913/test-0.8.3.html

Using \bmod in inline math mode seems to add some extra following space, which seems like a bug. Not sure if it's a problem with other binary operators.

edemaine:
This is a workaround for #836 for the command case of using \bmod, without changing any spacing (and making the code and generated output slightly more efficient).  In particular, fixes #982.


kevinbarabash:
Looks like the `mod` screenshots need to be regenerated.

src/functions/mod.js
```diff
@@ -12,13 +12,16 @@ const htmlModBuilder = (group, options) => {
     const inner = [];
 
     if (group.value.modType === "bmod") {
-        // “\nonscript\mskip-\medmuskip\mkern5mu”
+        // “\nonscript\mskip-\medmuskip\mkern5mu”, where \medmuskip is
+        // 4mu plus 2mu minus 1mu, translates to 1mu space in
+        // display/textstyle and 5mu space in script/scriptscriptstyle.
```
kevinbarabash: Took me a bit to grok what's going on, but it makes sense.

edemaine:
Hmm, seems there's more investigation to do.  Here's the diff between old and new screenshots (green is new):

mod-firefox-diff

More clear is to look at the texdiff.  Before this PR is near-perfect in the top left:

mod-old

After this PR is worse:

mod-new

edemaine:
Turns out I just failed to properly add the muspace CSS rule.  (Alternatively, we could add 1mu as a manual style value instead of using a CSS rule.)  I believe the Mod screenshots are now unaffected; this will be confirmed once Travis finishes.

PR #988, opened 2017-11-24
Simplify HTML generation logic for \\cancel, \\bcancel, \\xcancel.

marcianx:


src/stretchy.js
```diff
@@ -276,7 +276,7 @@ const encloseSpan = function(
     let img;
     const totalHeight = inner.height + inner.depth + 2 * pad;
 
-    if (/(fbox)|(color)/.test(label)) {
+    if (/fbox|color/.test(label)) {
```
kevinbarabash: I wonder if this should maybe be `/^(fbox|color)$/`.

kevinbarabash: I checked the code and we want to match `colorbox` and `fcolorbox` too.

kevinbarabash:
@marcianx thanks for the quick PR.

PR #989, opened 2017-11-24
Port katex, buildTree, parseTree to @flow.

marcianx:


src/buildTree.js
```diff
@@ -1,11 +1,19 @@
...
 import Settings from "./Settings";
 import Style from "./Style";
 
-const buildTree = function(tree, expression, settings) {
+import type ParseNode from "./ParseNode";
+import type domTree from "./domTree";
+
+const buildTree = function(
+    tree: ParseNode[],
+    expression: string | String,
```
kevinbarabash: `string` should be enough.

katex.js
```diff
@@ -14,11 +15,18 @@ import buildTree from "./src/buildTree";
...
+import type ParseNode from "./src/ParseNode";
+
 /**
  * Parse and build an expression, and place that expression in the DOM node
  * given.
  */
-let render = function(expression, baseNode, options) {
+let render = function(
+    expression: string,
+    baseNode: Node,
```
kevinbarabash: Should this be `ParseNode`?

marcianx: This is an HTML [`Node`](https://developer.mozilla.org/en-US/docs/Web/API/Node). Should it perhaps be an [`HTMLElement`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement) instead?

src/buildTree.js
marcianx: Done.

katex.js
kevinbarabash: I didn't read the surrounding code.  `Node` is fine since that's what `applyNode` is defined on.

PR #991, opened 2017-11-25
Support Unicode ∴ and ∵

ronkok:
Map ∴ and ∵ to `\therefore` and `\because`.


ronkok:
Unicode work will now proceed much more slowly than in the bulk phase. The PRs will be small and each one will explain the choices made.

In this PR, it's obvious which aliases are the correct ones. [unicode-math](https://github.com/wspr/unicode-math/blob/master/unicode-math-table.tex) agrees that these two characters map to `\therefore` and `\because`.

I’ve withheld these two characters until now because there is a conflict as to the proper type of atom. KaTeX `symbols.js` says they are `rel` atoms. **unicode-math** says they are `\mathord`. I think KaTeX got this one right. They come from the [amssymb](http://ctan.math.utah.edu/ctan/tex-archive/fonts/amsfonts/doc/amssymb.pdf) package, which contains: `\DeclareMathSymbol{\therefore} {\mathrel}{AMSa}{"29}`.

So this PR merely sets the flag in `symbols.js`, without changing the atom type.

ronkok:
Why not? I'll edit the comments.

edemaine:
Great!  Does anyone else want to weigh in on the atom types before this gets merged?

kevinbarabash:
Let's go with the AMS.

Issue #895, opened 2017-09-19
"KaTeX parse error: Expected 'EOF', got '<bla>' at position 1" when input pure Chinese character(s)


Lin-Buo-Ren:
![_257](https://user-images.githubusercontent.com/13408130/30611159-2a01f13a-9db3-11e7-8e41-2483e80bd39e.png)

control case: 

![_258](https://user-images.githubusercontent.com/13408130/30612763-36a304a6-9db8-11e7-8d14-9faf9a4f7b31.png)

## Bug Reproducing Instructions
1. Browse [KaTeX – The fastest math typesetting library for the web](https://khan.github.io/KaTeX/)
1. Input any C(JK) characters, this attempt uses "測試"

## Expected Behavior
"測試" printed in "See how it renders with KaTeX: " field, presumably to be in italic just as any regular text

## Current Behavior
The "See how it renders with KaTeX: " field spews out the following error: 

```
KaTeX parse error: Expected 'EOF', got '測' at position 1: 測̲試
```

kevinbarabash:
@Lin-Buo-Ren awesome bug report.  You can use CJK characters within `\text`.  I'm curious about the use cases of natural language strings outside of `\text`?

Lin-Buo-Ren:
@kevinbarabash 

> I'm curious about the use cases of natural language strings outside of \text?

I'm not sure as I'm not really acquainted with LaTeX and simply using it to write math equations in Markdown documents on GitBook, never used `\text` before ;)

07pepa:
I know it is utf8 characters related... looked into sources of katex on font and there are only printable english chars maybe it is related to multy-byte-ness of non ascii chars
 
PS. czech chars  like "ěščřžýáíéťňůú" and theri capitalizated version"ĚŠČŘŽÝÁÍÉŤŇŮÚ" do not work also and  utf-8 emoji also not work 😭.....only sotlution  is ( if input box is utf-8) do utf 8 filter if it thinking it is procesing ascii)

it can look like this ( is not corect im writing in hurry) //will return nuber of bytes that )
`function utf8(char input)
var locarvar=0;
{
if (input>128){
shr input; // shift right input
for (var i=0;i<7;i++)
{
if (odd(input)){
localvar++;
}
shr input;
}
return localwar;
}else return 0;
 }
`

edemaine:
[...]

PR #992, opened 2017-11-25
Unicode accents

### LINKED ISSUE #801: Latin character support
As realized in #796, Latin-1 characters are supported as input but render in the wrong font.

One approach is to allow single-character macros (like TeX active characters) and macro `é` to `\acute{e}`.
  [Here's an experiment in this direction.](https://github.com/edemaine/KaTeX/commit/2f3b09ca76c8b786c874df8de5416350045613d3)  This works, but we unfortunately need different macros for text and math mode.  Effectively we need `\ifmmode` support.  This can't be done in the macro expander, but could be done in the parser...  Alternatively, we could add support for `\acute` in text mode, even though LaTeX doesn't allow that.

Of course, actual text-mode support in this way would require first resolving #638 (text-mode accent commands)

edemaine:
This PR fixes #801.
* Lexer now looks for combining dicritical marks and adds them to the same character
* Parser's `parseSymbol` now recognizes both combined and uncombined forms of Unicode accents, and builds accent objects just like the accent functions
* ~~Added CJK support to math mode (not just text mode). #895.~~


test/unicode-spec.js
```diff
@@ -66,12 +66,12 @@ describe("unicode", function() {
...
     });
 
-    it("should parse Latin-1 inside \\text{}", function() {
-        expect('\\text{ÀàÇçÉéÏïÖöÛû}').toParse();
+    it("should parse all lower case Greek letters", function() {
+        expect("αβγδεϵζηθϑικλμνξοπϖρϱςστυφϕχψω").toParse();
     });
 
-    it("should parse Latin-1 outside \\text{}", function() {
-        expect('ÀàÇçÉéÏïÖöÛû').toParse();
```
kevinbarabash: Can we keep these tests but only check for `Çç`?  Eventually we'll want to test for `œ`, `Œ`, `æ`, `Æ`, and `ß` when they get added.

test/katex-spec.js
```diff
@@ -2680,15 +2680,54 @@ describe("A parser taking String objects", function() {
...
-
-    it("should parse 'ΓΔΘΞΠΣΦΨΩ'", function() {
-        expect("ΓΔΘΞΠΣΦΨΩ").toParse();
+describe("Unicode accents", function() {
+    it("should parse Latin-1 letters in math mode", function() {
+        // TODO(edemaine): Unsupported Latin-1 letters in math: ÅåÆÇÐØÞßæçðøþ
+        expect("ÀÁÂÃÄÈÉÊËÌÍÎÏÑÒÓÔÕÖÙÚÛÜÝàáâãäèéêëìíîïñòóôõöùúûüýÿ")
+        .toParseLike(
+            "\\grave A\\acute A\\hat A\\tilde A\\ddot A" +
+            "\\grave E\\acute E\\hat E\\ddot E" +
```
kevinbarabash: I guess expanding to `\grave A` makes sense.  Can you add a MathML fixture test?  I'm curious to see if we recombine the accent with the letter when we render to MathML.

src/Lexer.js
```diff
@@ -36,11 +36,16 @@ import {LexerInterface, Token} from "./Token";
 const commentRegexString = "%[^\n]*[\n]";
 const controlWordRegexString = "\\\\[a-zA-Z@]+";
 const controlSymbolRegexString = "\\\\[^\uD800-\uDFFF]";
+const combiningDiacriticalMarkString = "[\u0300-\u036f]";
```
kevinbarabash: This contains many more characters than we support.  Where do we generate an error if someone uses an unsupported combining mark?  Maybe add a test for this.

src/Parser.js
```diff
@@ -1062,8 +1063,61 @@ export default class Parser {
...
+            return newArgument(new ParseNode(symbols[this.mode][text].group,
+                text, this.mode, nucleus), nucleus);
+        } else if (cjkRegex.test(text)) {
+            this.consume();
+            return newArgument(
+                new ParseNode("textord", text, this.mode, nucleus));
+        }
+        // Decompose symbol into base and combining characters
+        // if the combined symbol is not recognized.
+        text = text.normalize('NFD');
```
kevinbarabash: This is super cool.  I wish I would've known about this sooner.  There's no support for this in IE and Safari only got this in version 10.  We might consider using https://github.com/walling/unorm instead.

src/Parser.js
```diff
@@ -1062,8 +1063,61 @@ export default class Parser {
...
+        // At this point, we should have a symbol, possibly with accents.
+        // If it's in the symbol table or CJK, we're done.
+        if (symbols[this.mode][text]) {
+            this.consume();
+            return newArgument(new ParseNode(symbols[this.mode][text].group,
+                text, this.mode, nucleus), nucleus);
+        } else if (cjkRegex.test(text)) {
+            this.consume();
+            return newArgument(
+                new ParseNode("textord", text, this.mode, nucleus));
```
kevinbarabash: If we bailing here do we need to include the `combiningDiacriticalMarkString` with surrogate pairs in the regex in Lexer.js?

src/Parser.js
```diff
@@ -1062,8 +1063,61 @@ export default class Parser {
...
+            return null;  // EOF, ^, _, {, }, etc.
+        }
+        this.consume();
+        // Transform combining characters into accents
+        if (match) {
+            for (let i = 0; i < match[0].length; i++) {
+                const accent = match[0][i];
+                if (!accents[accent]) {
+                    throw new ParseError(`Unknown accent ' ${accent}'`, nucleus);
+                }
```
kevinbarabash: Good enough for me.

src/symbols.js
```diff
@@ -705,25 +705,6 @@ for (let i = 0; i < letters.length; i++) {
...
-    defineSymbol(math, main, mathord, ch, ch);
-    defineSymbol(text, main, textord, ch, ch);
-}
-
-for (let i = 0x00F8; i <= 0x00FF; i++) {
-    const ch = String.fromCharCode(i);
-    defineSymbol(math, main, mathord, ch, ch);
-    defineSymbol(text, main, textord, ch, ch);
-}
-
```
kevinbarabash: Nice clean up!

src/Parser.js
```diff
@@ -1062,8 +1063,61 @@ export default class Parser {
...
+            symbol = new ParseNode("textord", text, this.mode, nucleus);
         } else {
-            return null;
+            return null;  // EOF, ^, _, {, }, etc.
+        }
+        this.consume();
+        // Transform combining characters into accents
+        if (match) {
+            for (let i = 0; i < match[0].length; i++) {
+                const accent = match[0][i];
```
kevinbarabash: Are there any unicode glyphs that include multiple accents?  If so, we should test that this works correctly.

src/Parser.js
edemaine: `combiningDiacriticalMarkString` is already included in Lexer's `tokenRegex` both after a single codepoint and after a surrogate pair.

src/Parser.js
edemaine: Ah, I should have checked the [browser compatibility matrix for String.normalize](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/normalize#Browser_compatibility).  The only problem with a polyfill is that it increases the `katex.min.js` build by 60%, from 226,836 bytes to 363,641 bytes (as any polyfill requires a huge table of Unicode symbols).

So I think this raises a serious question of whether we should:

1. Include the polyfill.
2. Not include the polyfill, and support Unicode in all supported browsers *except* IE.
3. Rewrite the PR to avoid `String.normalize`, at least in the built form.  We actually only want symbols that expand to accents within our supported set, so the symbol table we could generate at build time could be smaller than `unorm`'s.
4. Drop this PR altogether (though this is probably a desirable feature, and IE probably isn't evolving).

I think option 2 or 3 is best.  Thoughts?  I can look into option 3.

src/Lexer.js
edemaine: I added an error test (in `error-spec.js` because I wanted `toFailWithParseError`). Incidentally, most of my tests are in `katex-spec.js` because I wanted `toParseLike`, but they would make more sense in `unicode-spec.js`...  Not sure how to best share code between these test modules.

test/katex-spec.js
edemaine: They render exactly how accent commands do, which I learned is via `<mover accent="true">`.  I think this makes sense...  Test added.

test/unicode-spec.js
edemaine: This PR actually removes support for `Çç`.  I believe they were added prematurely when forcing support for Latin-1.  I just verified that none of the Latin-1 glyphs, `Çç` included, are in the KaTeX fonts.  (I believe this actually makes sense, and the purpose of this PR is to map them into the accent macros, like LaTeX would.)  Here's an example of incorrect looking `Çç` vs `Cc` in the current KaTeX:

image

edemaine: I opened https://github.com/KaTeX/katex-fonts/issues/2 to track the cedilla issue.

src/Parser.js
kevinbarabash: It's in Edge so my vote would be for option 2.  If people need to support IE then they can add the polyfill themselves.  We should document this requirement.

edemaine: Sounds good.  That's the easiest!  I'll add a fallback when `normalize` isn't implemented, maybe with a warning, and add documentation.

src/Lexer.js
kevinbarabash: I've created #1021 to extract helper functions into test-utils.js.  I can do that this weekend and then we could rebase this with the tests in the proper suite.

src/Parser.js
kevinbarabash: We have to lex the tokens before we can test the tokens against the cjkRegex here, d'oh.

kevinbarabash:
@edemaine this is looking pretty good.  It looks like some of the screenshot tests use `Ç` and `ç` so those will have to be changed or we could add support for them even though we don't have the correct glyphs and fix that issue in a separate PR.  Your call.

src/symbols.js
```diff
@@ -684,27 +684,33 @@ defineSymbol(text, main, spacing, "\u00a0", "~");
 // There are lots of symbols which are the same, so we add them in afterwards.
 
 // All of these are textords in math mode
-const mathTextSymbols = "0123456789/@.\"";
-for (let i = 0; i < mathTextSymbols.length; i++) {
-    const ch = mathTextSymbols.charAt(i);
+for (const ch of "0123456789/@.\"") {
```
kevinbarabash: Nice cleanup, but I think this may actually cause the bundle size to increase quite a bit b/c it requires a polyfill for the string iterator which I believe uses .codePointAt() which also ends up be polyfilled.  Can you generate a build and see how much the bundle size has increased?

edemaine:
@kevinbarabash Looking at the history, I realize that the "weak" support for Latin-1 characters (using the wrong font) is actually really old, so we should keep support for those characters for backwards compatibility.  I added a TODO comment about how they're currently supported "incorrectly".

I also updated the Unicode screenshot, which shows proper support for many characters but not all (like `Çç`).  It also highlights an issue I've known about but we should probably discuss: `ï` renders with an accent on top of a dot (and similarly for all accented `i`s and `j`s).  To fix this in text mode, we need `\i` and `\j` support (https://github.com/KaTeX/katex-fonts/issues/2), but I added a fix for it now in math mode because we have `\imath` and `\jmath`.

src/symbols.js
edemaine: I don't think there are polyfills involved, but this is what it expands to:

```js
var _iteratorNormalCompletion = true;
var _didIteratorError = false;
var _iteratorError = undefined;

try {
    for (var _iterator = (0, _getIterator3.default)("0123456789/@.\""), _step; !(_iteratorNormalCompletion = (_step = _iterator.next()).done); _iteratorNormalCompletion = true) {
        var _ch = _step.value;

        defineSymbol(math, main, textord, _ch, _ch);
    }

    // All of these are textords in text mode
} catch (err) {
    _didIteratorError = true;
    _iteratorError = err;
} finally {
    try {
        if (!_iteratorNormalCompletion && _iterator.return) {
            _iterator.return();
        }
    } finally {
        if (_didIteratorError) {
            throw _iteratorError;
        }
    }
}
```

Or minified:

`var M=true;var _=false;var S=undefined;try{for(var z=(0,n.default)('0123456789/@."'),T;!(M=(T=z.next()).done);M=true){var A=T.value;u(o,f,k,A,A)}}catch(e){_=true;S=e}finally{try{if(!M&&z.return){z.return()}}finally{if(_){throw S}}}var C=true;var O=false;var N=undefined;try{for(var j=(0,n.default)('0123456789!@*()-=+[]<>|";:?/.,'),L;!(C=(L=j.next()).done);C=true){var E=L.value;u(s,f,k,E,E)}}catch(e){O=true;N=e}finally{try{if(!C&&j.return){j.return()}}finally{if(O){throw N}}}`

So it does add a few hundred bytes.  Shall I switch back?

kevinbarabash: The code that it generates is pretty awful.  We should experiment with `loose` mode for the `for-of` plugin for babel sometime, see https://babeljs.io/docs/plugins/transform-es2015-for-of.  Interestingly, when iterating over an array literal it will output a basic `for` loop.  Maybe we could do `for (const letter of "abc...".split()) { ... }` to force this optimization.

kevinbarabash:
Not sure why travis isn't running the tests.  I had to restart the tests manually.

src/symbols.js
edemaine: How about the following, which Babel seems to leave as-is?

```js
"0123456789/@.\"".split().forEach(function () {
    defineSymbol(math, main, textord, ch, ch);
});
```

Wow, this is actually fastest!  [jsperf benchmark](https://jsperf.com/iterate-over-string)

jsperf output

kevinbarabash: I wasn't expecting that.  Awesome!

edemaine: Oops, that's because I was using `split` wrong.  It's now the slowest...  So perhaps I should go back to a regular for loop.

Corrected jsperf output

edemaine:
I added a test and warning for `normalize()`, and tested that it occurs when using IE (but not otherwise). And switched back to for loops.

src/symbols.js
kevinbarabash: right, b/c it still has to call `charCodeAt`... yeah, let's go back to regular for loops.

edemaine:
I think this is ready for final review.  We can move the tests around later (?).

kevinbarabash:
@edemaine sorry I've been holding off on approving this.  I've been doing some more thinking about making people add their own polyfill for IE support and I wonder how many people are still supporting IE.  For the record KA is still supporting IE10+.  Although we haven't upgraded KaTeX recently, if we were to do so we probably wouldn't want to include the whole `normalize` polyfill and would end up writing our polyfill which covers what we using in KaTeX.  At that point we may as well include the code in KaTeX.  Later, we could add a build setting to leave it out for people who aren't supporting IE.

edemaine:
That makes sense. Maybe I should try to get it working with normalize only during build, and not requiring it at run time, then.

Syndesi:
Hi,
if I get it right, the width and length of supported characters are currently hardcoded and therefore unknown characters won't get rendered or break the formula.
Wouldn't it be better when the major characters are hardcoded (to improve the performance) and all other strings are put into an invisible paragraph which is then tested for its width & height? With this solution it should be possible to use every language and even those whose width depends on their context (e.g. Arabic).

edemaine:
@Syndesi I'd suggest opening a separate issue (or PR) for this idea.  Actually, there might be one; I'd need to search.  Rendering repeatedly in the browser to measure box sizes is the general approach of MathJax, which is the main reason (I believe) it's so slow in comparison to KaTeX.  It also prevents server-side rendering, which KaTeX currently supports.  So it's unlikely that this will happen, but feel free to open/find a discussion about it.

kevinbarabash:
> Maybe I should try to get it working with normalize only during build, and not requiring it at run time, then.

@edemaine were you thinking of using `normalize` at build time to generate a table or something containing the split characters?

edemaine:
@kevinbarabash Yes, exactly. Maybe there's a `src/unicode.js`, which can be recompiled (e.g. if Unicode changes) via `make src/unicode.js`, which runs some code (e.g. `src/make-unicode.js`, or located elsewhere) not included in the KaTeX build, and only the latter uses `normalize`/`unorm`. Similar to fontMetrics making, I suppose -- I can look more at how that's done.

I'd set this up only to recognize accents that KaTeX can handle, so it'd hopefully be not too large a list.  We can still handle combining characters in exactly the way of this PR, just skipping the normalization.  Instead, I think I'd define (in `src/unicode.js`) macros that map e.g. `á` to `a` followed by an acute combining character.

edemaine:
OK, I gave a crack at this.  Multi-accented characters were challenging to get right (glad we added a test for them!), and it turns out there *are* characters with two accents.  I checked and there aren't any triple-accented characters.

I'm not sure of the best location for `unicodeMake.js`.  We don't have a `util` directory, which would be a natural place.  It could also be in the root directory.  But in `src` it immediately gets linted and such.

src/unicodeSymbols.js
```diff
@@ -0,0 +1,290 @@
+// This file is GENERATED by unicodeMake.js. DO NOT MODIFY.
```
kevinbarabash: Cool!

src/unicodeSymbols.js
```diff
@@ -0,0 +1,290 @@
...
+
+export const unicodeSymbols = {
+    "\u00e1": "\u0061\u0301",  // á = \'{a}
```
kevinbarabash: I love the comments.

src/unicodeMake.js
```diff
@@ -0,0 +1,59 @@
...
+        }
+        output += `\\u${hex}`;
+    }
+    output = `${output}"`;
+    return output;
+};
+
+console.log("// This file is GENERATED by unicodeMake.js. DO NOT MODIFY.");
+console.log("");
+console.log("export const unicodeSymbols = {");
```
kevinbarabash: Do we need this in addition to the `export default`?

src/unicodeMake.js
```diff
@@ -0,0 +1,59 @@
...
+// whose purpose is to generate unicodeSymbols.js in this directory.
+// In this way, only this tool, and not the distribution/browser,
+// needs String's normalize function.
+
+const fs = require('fs');
+
+// Read supported accents from Parser.js
+const Parser = fs.readFileSync('./Parser.js', {encoding: 'utf8'});
+const match = /accents = [^;]*;/.exec(Parser);
+eval(match[0]);
```
kevinbarabash: Maybe we could have `Parser.js` export `accents` instead.

src/unicodeMake.js
```diff
@@ -0,0 +1,59 @@
...
+console.log("// This file is GENERATED by unicodeMake.js. DO NOT MODIFY.");
+console.log("");
+console.log("export const unicodeSymbols = {");
+
+const letters = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ";
+for (const letter of letters) {
+    for (const accent of Object.getOwnPropertyNames(accents)) {
+        const combined = letter + accent;
+        const normalized = combined.normalize('NFC');
+        if (normalized.length === 1) {
```
kevinbarabash: This is super cool.

test/katex-spec.js
```diff
@@ -2697,15 +2697,64 @@ describe("A parser taking String objects", function() {
     });
 });
 
-describe("Unicode", function() {
-    it("should parse all lower case Greek letters", function() {
-        expect("αβγδεϵζηθϑικλμνξοπϖρϱςστυφϕχψω").toParse();
-    });
-
-    it("should parse 'ΓΔΘΞΠΣΦΨΩ'", function() {
-        expect("ΓΔΘΞΠΣΦΨΩ").toParse();
```
kevinbarabash: I see, these got moved down to unicode-spec.js.

test/katex-spec.js
```diff
@@ -2697,15 +2697,64 @@ describe("A parser taking String objects", function() {
...
+            "\\grave o\\acute o\\hat o\\tilde o\\ddot o" +
+            "\\grave u\\acute u\\hat u\\ddot u" +
+            "\\acute y\\ddot y");
+    });
+
+    it("should parse Latin-1 letters in text mode", function() {
+        // TODO(edemaine): Unsupported Latin-1 letters in text: ÆÇÐØÞßæçðøþ
+        expect("\\text{ÀÁÂÃÄÅÈÉÊËÌÍÎÏÑÒÓÔÕÖÙÚÛÜÝàáâãäåèéêëìíîïñòóôõöùúûüýÿ}")
+        .toParseLike(
+            "\\text{\\`A\\'A\\^A\\~A\\\"A\\r A" +
```
kevinbarabash: `\\\"` caught me off-guard until I realized that the `"` need to be escaped as well.

test/unicode-spec.js
```diff
@@ -87,8 +97,8 @@ describe("unicode", function() {
         expect('\\text{여보세요}').toParse();
     });
 
-    it("should not parse CJK outside \\text{}", function() {
-        expect('私はバナナです。').toNotParse();
-        expect('여보세요').toNotParse();
+    it("should parse CJK outside \\text{}", function() {
+        expect('私はバナナです。').toParse();
+        expect('여보세요').toParse();
```
kevinbarabash: This change isn't really related to accents and should probably be in a separate PR.

src/unicodeMake.js
```diff
@@ -0,0 +1,59 @@
+/* eslint no-console:0 no-undef:0 */
+
+// This is an internal Node tool, not part of the KaTeX distribution,
+// whose purpose is to generate unicodeSymbols.js in this directory.
+// In this way, only this tool, and not the distribution/browser,
+// needs String's normalize function.
```
kevinbarabash: Can you include when this tool should be run?

Makefile
```diff
@@ -126,6 +126,10 @@ metrics:
 extended_metrics:
 	cd metrics && $(PERL) ./mapping.pl | $(PYTHON) ./extract_tfms.py | $(PYTHON) ./extract_ttfs.py | $(PYTHON) ./format_json.py --width > ../src/fontMetricsData.js
 
+unicode:
+	cd src && $(NODE) unicodeMake.js >unicodeSymbols.js
+src/unicodeSymbols.js: unicode
```
kevinbarabash: I wonder if we can run this as part of the build and then not even bother with committing the result.  That we don't have to worry about whether people are running it when they should or not.

src/unicodeMake.js
edemaine: I'm still somewhat new to ES6, so I'm not sure what's normal here. I still find it weird that `export default const` is forbidden. I believe we could just do `export default {...}` if we don't need to name the object, so maybe that's best? The current system allows importing in two different ways (default or nondefault), but that's probably not necessary.

src/unicodeMake.js
edemaine: I should add a comment about this. I originally tried it that way, but Node doesn't support `import`, and I'm running `unicodeMake.js` in Node, so Babel would have to get involved, and the accents are not available in built `katex.js`.  I'm open to better suggestions!  We could, for example, put `accents` here in `unicodeMake.js` and/or `unicodeSymbols.js` and import them in `Parser.js` -- perhaps that's best.

Makefile
edemaine: My motivation for committing it is so that people wouldn't worry about when to run it.  (You need to run it whenever we add support for a new accent, or when the Unicode spec changes, neither of which I imagine being very common.)  Another advantage is that we can directly see what characters are supported, so it's easy to complain about a missing one, and for PRs to show what additional characters are being supported.

If we do it during build, would we also have to do it when running the test server...?  This approach felt more complicated to me, but I'm happy to have a specific suggestion.

src/unicodeMake.js
edemaine: Added.

src/unicodeMake.js
edemaine: OK, I tried moving them. It's not especially pretty -- I still `eval`d the code in `unicodeMake`, which led to extra `\` escaping in the `accents` definition... Let me know if you see a cleaner way! But at least I'm no longer parsing `Parser.js` with a regex!

test/unicode-spec.js
edemaine: It's relevant in that this PR rewrites how we parse Unicode. The relevant line change is from `if (this.mode === "text" && cjkRegex.test(nucleus.text))` to `if (cjkRegex.test(text))`. I can't make a separate PR now because they'd both change that line, but if you like, I can remove that change from here and add a second PR once this one is accepted.

Makefile
kevinbarabash: Let's leave it like this for now.  I definitely like it committing it from a documentation point of view.  I don't think we have much work left to do on accents so there shouldn't be too many PRs which is affected by this.

src/unicodeMake.js
```diff
@@ -0,0 +1,74 @@
...
+    '\\u0303': {text: '\\\\~', math: '\\\\tilde'},
+    '\\u0304': {text: '\\\\=', math: '\\\\bar'},
+    '\\u0306': {text: '\\\\u', math: '\\\\breve'},
+    '\\u030c': {text: '\\\\v', math: '\\\\check'},
+    '\\u0302': {text: '\\\\^', math: '\\\\hat'},
+    '\\u0307': {text: '\\\\.', math: '\\\\dot'},
+    '\\u030a': {text: '\\\\r'},
+    '\\u030b': {text: '\\\\H'},
+}`;
+eval(accentsCode);
```
kevinbarabash: I'm confused why the `eval` is necessary, can't we just `console.log(accentsCode);` this as is?

kevinbarabash: A better solution might be to do:
```
const {accents} = require('./unicodeSymbols.js');

const encodedAccents = {};
for (const [key, value] of Object.values(accents)) {
    const encodedEntry = {};
    encodedAccents[encode(key)] = encodedEntry;
    if (value.text) {
        encodedEntry.text = encode(value.text);
    }
    if (value.math) {
        encodedEntry.math = encode(value.math);
    }
}
```
This way we have one source of truths for the accents.

test/unicode-spec.js
kevinbarabash: > I can remove that change from here and add a second PR once this one is accepted.

I think that makes the most sense.  Thanks. :)

src/unicodeMake.js
kevinbarabash: I find the lack of support for `export default const` weird as well.  You can do:
```
export { foo as default };
```
I guess it's a moot point now since there no longer seems to be an `export default` in this file.

kevinbarabash:
@edemaine do you know what package we need to load to render the `Unicode` test case in LaTeX?

kevinbarabash:
The good news:
- the base characters are using computer modern now, yay!
- grave, acute, and circumflex accents have better shapes than before

Areas for improvement:
- grave, acute, and circumflex accents on capitals need to be rotated or squished a bit
- umlaut circles seem a bit too large
- `i` should use the dotless i when it has an accent*
- the horizontal spacing around `i` and `I` is too much*

The items that are starred are things that should probably be resolved before merging this PR.

kevinbarabash:
@edemaine thanks for refactoring `unicodeAccents` and revert the CJK change.  :)

edemaine:
@kevinbarabash I guess for backwards compatibility, we could just let `ï` pass through as a character and render in Times.  Would that be acceptable?  We can't otherwise fix the double accent issue in text mode (it already is fixed in math mode) until https://github.com/KaTeX/katex-fonts/issues/2 gets solved.

I'm guessing that the squishing of the accents on capital letters is also some kind of font issue (another symbol that we need), but I don't actually see where it's implemented in LaTeX...  I just see `\DeclareTextAccentDefault{\'}{OT1}`.  Dot size is presumably also a font issue...

The horizontal spacing is definitely something that can be fixed.  It's an existing issue with `\acute{i}`, `\acute{\imath}`, and `\text{\'i}` in KaTeX, not new to this PR (but this PR exacerbates it to apply to `ï` as well).  I think we just never handled the case of the innards being less wide than the accent.  I'll take a look... But we could also leave this to another PR, if we hackily let `ï` pass through.

edemaine:
I looked into the horizontal spacing issue, and summarized in #1028. I don't think this will be a quick fix, unfortunately, because I think it needs the accents' width to be added to `fontMetricsData.js`.

Back to this PR, I just discovered Latin-1 characters listed in `fontMetrics.js`.  They should be removed during this PR, with the possible exception of `ï`.

edemaine:
The last three commits fix the following:
* Accents on i now use `\i` in text mode and `\imath` in math mode; actually they do this via the Unicode mechanism. This fixes one of your starred issues.
* Remove the extra character mappings in `fontMetrics.js` that I pointed out in my previous post.
* Fix a mistake in the Latin-1 coverage test (wasn't testing math mode), which led to fixing an error for Åå (which we want for backwards compatibility, but can't do "properly" because KaTeX (and LaTeX) lack this accent in math mode).

Incidentally, I tried `xelatex` (with `\usepackage{unicode-math}`) on the following simple test: `Héllo Åå $éÅå$` and none of the Unicode characters in math mode worked... maybe I'm missing something?

So that mainly leaves the width of accented `i` issue.

kevinbarabash:
> So that mainly leaves the width of accented i issue.

Cool!  Once #1030 is merged I'll open a PR that adds the width to the glyph metrics which shouldn't take too long.  You can build them already by doing `make extended_metrics`.  I think it's best to have them in a separate PR as it's going to be a big change and this PR is already quite big.

kevinbarabash:
#1032 contains the horizontal metrics.

edemaine:
#1033 should fix the `ï` width issue.

edemaine:
New texcmp with #1033 merged (via rebase):

image

Ready to merge?

I also added support for accented Greek letters, as I recalled/noticed there are a handful of those as well.  (I didn't add e.g. Greek Capital Letter A(lpha) with acute accent because we don't support Greek Capital Letter A(lpha) yet.)  @ronkok Does `unicode-math` or something alias that character to a Latin A?

ronkok:
@edemaine, I am not aware of a stand-alone Unicode character for Greek capital alpha with acute accent, so `unicode-math` does not have such an alias. There is Ά, that is U+0386, Greek capital alpha with tonos, but `unicode-math` does not have an alias for it, either.

kevinbarabash:
> Ready to merge? Still the cedilla issue and accent-on-capital letter issue, which we can add. (Neither specific to this PR.)

I think so.  We can deal with the cedilla and `\H` issues separately.

kevinbarabash:
@edemaine I removed some of the letters from the math mode test since we aren't rendering them with the correct font anyways.

kevinbarabash:
@edemaine thanks for all of the hard work on this.  I'm glad it's finally merged.

edemaine:
@ronkok I was indeed referring to U+0386 Ά which is "Greek Capital Letter A with acute accent" according to [this list](https://en.wikipedia.org/wiki/List_of_Unicode_characters). (I assume A means Alpha here, though I'm no expert.) I see that other lists call it "GREEK CAPITAL LETTER ALPHA WITH TONOS", but it does canonicalize into U+0391 (Greek Capital Alpha) followed by U+0301 (combining acute accent). There are similar "Latin-looking" (and thus no-LaTeX-macro) characters Έ Ή Ί Ό Ύ.  But if LaTeX doesn't support them, fine for us not to.

@kevinbarabash Thanks for pushing and helping to make this PR so much better! Together with @ronkok's Unicode symbol support, the next KaTeX release will be really Unicode friendly!

ronkok:
Very cool stuff. I'll start to write it into the KaTeX function support page. 

@edemaine, when I submit the PR for the updated function support page, I would appreciate if you would look it over. I'm not sure that I have picked up on everything that you and @kevinbarabash  are doing here.

edemaine:
@ronkok Sounds good! For this PR, the new single symbols should all be in [src/unicodeSymbols.js](https://github.com/Khan/KaTeX/blob/master/src/unicodeSymbols.js). They currently work in both text and math mode (but I'm not sure they should work in math mode -- see #1046). In addition, the combining characters for all the supported text and math accents work, as listed in [src/unicodeAccents.js](https://github.com/Khan/KaTeX/blob/master/src/unicodeAccents.js).  In addition, #1030 added `\ae, \AE, \oe, \OE, \o, \O, \ss` and their Unicode equivalents to text mode.

ronkok:
>  I'm not sure they should work in math mode

For now, I'll write it up to suggest that they work only in text mode.  I would like to keep math mode out of the docs until we're sure that it won't break future work. We can always change the docs later to include math mode.

PR #993, opened 2017-11-26
Support Unicode \mid

ronkok:
Map ∣, U+2223, to `\mid`


ronkok:
This character is U+2223, DIVIDES. I’ve withheld it until now only because it looks just like `|`, U+007C. So the question here is do we include any confusables at all. If we do, then the path forward is reasonably clear. [unicode-math](https://github.com/wspr/unicode-math/blob/master/unicode-math-table.tex) maps this character to `\mid`, the vertical line with `rel` spacing, useful for set builder notation. [John Cook](https://www.johndcook.com/unicode_latex.html) maps it the same way, as does Microsoft Word.

PR #994, opened 2017-11-27
Update function support page for 0.9.0-alpha2

ronkok:


ronkok:
Remember, the function support page is a stand-along file. If you want to see how it renders, you can just download it and take a look.

kevinbarabash:
@ronkok thanks for updating this.  I like how you have "direct input" sections for each of the existing groups.

PR #997, opened 2017-11-28
Change \xLongequal to \xlongequal

ronkok:
Fix a spelling mistake from PR #670.


ronkok:
I noticed it while looking at the [test page](http://utensil-site.github.io/available-in-katex/). The test page, by the way, is looking much better these days.

PR #998, opened 2017-11-28
extract color and text commands into their own files in functions sub-directory

kevinbarabash:
Continue to work towards having all functions use the new `defineFunction` command.  Previous functions were migrated in #871.

### DIFF
--- a/src/functions/color.js
+++ b/src/functions/color.js
@@ -1,0 +1,88 @@
+// @flow
+import defineFunction, {ordargument} from "../defineFunction";
+import buildCommon from "../buildCommon";
+import mathMLTree from "../mathMLTree";
+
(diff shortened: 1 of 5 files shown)


src/functions/color.js
```diff
@@ -0,0 +1,88 @@
...
+defineFunction({
+    type: "color",
+    names: ["\\textcolor"],
+    props: {
+        numArgs: 2,
+        allowedInText: true,
+        greediness: 3,
+        argTypes: ["color", "original"],
+    },
+    handler: (context, args) => {
```
marcianx: (optional) Here and below (`hander`, `htmlBuilder`, `mathmlBuilder`), you can just do

```
handler(context, args) {
...
```

and save a few characters.

PR #999, opened 2017-11-28
extract overline, underline, and rule into their own files

kevinbarabash:
Continue working towards having all functions use the new defineFunction command. Previous functions were migrated in #871 and #998.

### DIFF
--- a/src/functions/overline.js
+++ b/src/functions/overline.js
@@ -1,0 +1,57 @@
+// @flow
+import defineFunction from "../defineFunction";
+import buildCommon from "../buildCommon";
+import mathMLTree from "../mathMLTree";
+
+import * as html from "../buildHTML";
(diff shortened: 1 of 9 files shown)


src/functions/overline.js
```diff
@@ -0,0 +1,57 @@
...
+import * as html from "../buildHTML";
+import * as mml from "../buildMathML";
+
+defineFunction({
+    type: "overline",
+    names: ["\\overline"],
+    props: {
+        numArgs: 1,
+    },
+    handler: (context, args) => {
```
marcianx: (optional) As in the other PR, the following is slightly shorter

```
handler(context, args) {
    ...
}
```

kevinbarabash:
@marcianx thanks for the quick review.

Issue #970, opened 2017-11-19
Dividing lines for array environment sometimes disappear


akalin:
Test file for 0.8.3: https://rawgit.com/akalin/cc8023a41d1fc1cd0250bc50d122680f/raw/ed169e5a427382c12543a0b941ae66c34eda0003/test-0.8.3.html

Test file for 0.9.0-alpha1:
https://rawgit.com/akalin/cc8023a41d1fc1cd0250bc50d122680f/raw/e7b2ba03faa7b28b93951dc21402fc5ab443f0eb/test-0.9.0-alpha1.html

Vertical divider isn't visible at 100% zoom with Chrome Version 64.0.3269.3 (Official Build) dev (64-bit) on macOS 10.13.1, but they show up at 110%.

Related to #824, but @ronkok says he wants to fix that one first before doing this one.

Gist: https://gist.github.com/akalin/cc8023a41d1fc1cd0250bc50d122680f

ronkok:
@akalin In PR #976, I went ahead and gave `vertical-separator` the same full-span treatment that I gave to `frac-line`. We have one report in from @trimbonz that the work is effective making frac-lines visible. Hopefully, the same will be true for vertical array separator lines.

akalin:
Not sure if 0.9.0-alpha2 is supposed to fix this too, but https://rawgit.com/akalin/cc8023a41d1fc1cd0250bc50d122680f/raw/e7b2ba03faa7b28b93951dc21402fc5ab443f0eb/test-0.9.0-alpha2.html is a test file that behaves the same way as 0.9.0-alpha1.

akalin:
0.9.0-beta fixes this completely for me: https://rawgit.com/akalin/cc8023a41d1fc1cd0250bc50d122680f/raw/81cbe2e2c359d591b61260c832490c0f1d2c87cc/test-0.9.0-beta.html

Thanks!

kevinbarabash:
@akalin thanks for verify that the new release fixes the issue.

PR #1004, opened 2017-12-01
Fix frac-line with wide span

ronkok:
Render frac-line, underline, overline, and vertical-separator using an SVG path that fills the middle third of an SVG image and its containing span.

Because the span is three times as thick as the line, we hope that Chrome will consistently render it.


ronkok:
@kevinbarabash, When I test this PR locally, I get math rendered, but I also get a number of console warnings about missing fonts.

ronkok:
I'm not 100% sure that I got all the kern distances correct. A careful test with Screenshotter is called for.

kevinbarabash:
@ronkok in #891 I moved fonts into a submodule. Running `git submodule update --init --recursive` should fix the issue.  There are additional details of working with submodules in https://github.com/Khan/KaTeX/blob/master/CONTRIBUTING.md#working-with-submodules.

kevinbarabash:
screen shot 2017-12-02 at 12 06 35 am

The fraction lines are a bit blurry on non-retina displays.  This is to be expected.  We might experiment with setting `shape-rendering: crispEdges` on the fraction bar SVGs.

kevinbarabash:
screen shot 2017-12-02 at 12 37 14 am

Even though the fraction line is 0.12em high, at small fonts it looks that's still somewhere between 1 and 2 pixels.  I wonder if we should render an empty span for the fraction bar and then a second span with that contains the SVG which is at least 0.04em + 2px.

ronkok:
> We might experiment with setting `shape-rendering: crispEdges`

It may be worth an experiment, but I expect the same sort of failure that we get from borders. We tried `crispEdges` during PR #670 and some horizontal lines disappeared on a low resolution monitor.

ronkok:
> at least 0.04em + 2px tall

I'll give that some thought. It might be easier just to make the line span 7 lines tall or 9 lines tall instead of 3 lines tall.

ronkok:
>  a bit blurry on non-retina displays

I'll experiment a bit with changing the [two brush strokes](https://github.com/Khan/KaTeX/blob/master/src/stretchy.js#L86) to three brush strokes.

ronkok:
Since we're experimenting, I'm also going to try CSS [border-image](https://developer.mozilla.org/en-US/docs/Web/CSS/border-image). Unlike a `border`, a `border-image-width` can be a percentage. That would make a `min-height` value on the span effectively a `min-height` on the frac-line.

ronkok:
Results are in re: `crispEdges`. Both images below are from Firefox. The top image includes `crispEdges`. The bottom image does not:
crispedges
nocrispedges

Notice that the fifth and sixth `crispEdges` examples, in 12 pt font and 11 pt font, omit the `fracline`.  Chrome and Opera were okay in the larger font sizes, but they omitted `frac-lines` in the smaller sizes. 

I am going to avoid `crispEdges`. I'll try to find some other way to avoid blurriness of the line edges.

ronkok:
The most recent commit sets the container span to 5 times the line thickness. My tests show `frac-lines` that are:
1. visible until font sizes get so small that I can't read the numerator,
2.  blurry on non-retina displays.

I can't think of much more to help that blurriness. I don't want to use `crispEdges` for reasons described above and a triple brush stroke did not help. I think this is the best I can do short of changing the `defaultRuleThickness` to something larger than 0.04em.

ronkok:
@kevinbarabash, It seems to me that the blurriness is worse in scriptstyle than in textstyle. And I think there is something I can do about that, but it may be a few days before I get to it.

ronkok:
The most recent commit changes from SVG paths to SVG lines. The change makes no difference that I can see on high-resolution screens, but the lines will get darker/sharper when two lines are written in Chrome on a Windows 8 box with a low-res screen. 

If fact, when I tested 4 brush strokes, Chrome would actually make the lines much thicker on the low-res screen. Strange.  

For the commit, I stayed with two brush strokes.

kevinbarabash:
@ronkok should I regenerate the screenshots?

ronkok:
Yes, thank you. I've done everything with this PR that I can think to do.

kevinbarabash:
@ronkok I've sent you a PR with the updated screenshots.  There also seems to be a `flow` issue:
```
Error: src/functions/genfrac.js:165
                        v----------------------
165:             frac = buildCommon.makeVList({
166:                 positionType: "individualShift",
167:                 children: [
...:
176:             }, options);
                 ----------^ call of method `makeVList`
173:                     {type: "elem", elem: rule,   shift: midShift + 2 * ruleWidth},
                                              ^^^^ null. This type is incompatible with
346:     elem: DomChildNode,
               ^^^^^^^^^^^^ union: span | anchor | svgNode | symbolNode. See: src/buildCommon.js:346
```

src/buildCommon.js
```diff
@@ -330,11 +330,13 @@ const makeLineSpan = function(
     className: string,
     options: Options,
 ) {
-    // Fill the entire span instead of just a border. That way, the min-height
-    // value in katex.less will ensure that at least one screen pixel displays.
-    const line = stretchy.ruleSpan(className, options);
-    line.height = options.fontMetrics().defaultRuleThickness;
-    line.style.height = line.height + "em";
+    // Return a span with an SVG image of a horizontal line. The SVG path
+    // fills the middle third of the span. We want an extra tall span
```
kevinbarabash: Should this be `fills the middle fifth of the span.`?

src/functions/genfrac.js
```diff
@@ -166,8 +166,10 @@ defineFunction({
                 positionType: "individualShift",
                 children: [
                     {type: "elem", elem: denomm, shift: denomShift},
-                    // $FlowFixMe `rule` cannot be `null` here.
-                    {type: "elem", elem: rule,   shift: midShift},
+                    // The next line would ordinarily contain "shift: midShift".
+                    // But we put the rule into a a span that is 5 rules tall,
+                    // to overcome a Chrome rendering issue.
+                    {type: "elem", elem: rule,   shift: midShift + 2 * ruleWidth},
```
kevinbarabash: It might be good to further elaborate on the comment by stating that the top/bottom padding is 2 spans in height which corresponds to the `2 * ruleWidth` we're adding to the `shift`.

src/functions/overline.js
```diff
@@ -34,9 +34,11 @@ defineFunction({
             positionType: "firstBaseline",
             children: [
                 {type: "elem", elem: innerGroup},
-                {type: "kern", size: 3 * line.height},
-                {type: "elem", elem: line},
+                // The kern on the next line would ordinarily be 3 * line.height
+                // But we put the line into a span that is 5 lines tall, to
+                // overcome a Chrome rendering issue.
```
kevinbarabash: Shouldn't there be a kern that's `3 * line.height - 2 * ruleWidth` or is the change so slight that it's not worth bothering about?

src/functions/underline.js
```diff
@@ -24,17 +24,19 @@ defineFunction({
...
             positionType: "top",
             positionData: innerGroup.height,
             children: [
-                {type: "kern", size: line.height},
+                // The SVG image is 5x as tall as the line.
+                // The bottom 2/5 of the image is blank and acts like a kern.
+                // So we omit the kern that would otherwise go at the bottom.
                 {type: "elem", elem: line},
-                {type: "kern", size: 3 * line.height},
+                {type: "kern", size: 5 * line.height},
```
kevinbarabash: Are these `elem` and `kern` children overlapping somehow?

src/stretchy.js
```diff
@@ -326,16 +326,53 @@ const encloseSpan = function(
...
+const ruleSpan = function(className: string, lineThickness: number,
+    options: Options): domTree.span {
+
+    // Get a span with an SVG line that fills the middle fifth of the span.
+    // We're using an extra wide span so Chrome won't round it down to zero.
+
+    const lines = [];
+    let svgNode;
+    if (className === "vertical-separator") {
+        // Apply 2 brush strokes for sharper edges on low-res screens.
```
kevinbarabash: Nice

static/katex.less
```diff
@@ -574,20 +554,9 @@
...
-
-            // Modify min-width for retina displays. Don't force two screen pixels.
-            @media screen and (-webkit-min-device-pixel-ratio: 2.0),
-                screen and (min-resolution: 192dpi) {
-                min-width: 0.5px;
-            }
-        }
+            margin: 0 -0.125em;     // The ink in the rule is actually 0.05em wide.
+            width: 0.25em;          // We put it in a 0.25em wide span to overcome
+        }				            // a Chrome rendering issue.
```
kevinbarabash: Were the vertical separators disappearing at times?  Won't this change to the styling affect the horizontal layout slightly?

src/buildCommon.js
ronkok: Yep.

src/functions/genfrac.js
ronkok: Done.

src/functions/overline.js
ronkok: I don't see a difference in that distinction. From `buildCommon.makeLineSpan`, we have 
~~~
line.height = options.fontMetrics().defaultRuleThickness;
~~~
So the proposed 2 lines of space in the SVG plus the one-line-high kern sum up to the same distance as the previous `3 * line height`. I'll amplify the comment, but I don't see any other action to take. Am I missing something?

src/functions/underline.js
ronkok: Yes. 
underline

static/katex.less
ronkok: Issue #970 reported a disappearing vertical separator line, so this is a valid action to take.

Re: horizontal layout: Good catch. I changed the negative `margin-left` amount, so the line has the correct space to its left. But the wider span must widen the space to its right.  No instant, obvious fix comes to mind. I'll think it over.

ronkok:
Well, I've corrupted another branch with commits from another PR. I'm closing this PR so that I can submit the same material in a fresh one. Have I mentioned lately that I hate Git?

kevinbarabash:
☹️

src/functions/overline.js
kevinbarabash: I was missed the connection between `lineHeight` and `ruleWidth`.

PR #1005, opened 2017-12-02
Fix frac-line via border-image

ronkok:
Use a border-image instead of a border to render a frac-line. That way, a `min-height` will apply to the line.


ronkok:
This is an alternate to the previous PR.

ronkok:
This PR fails. On a high-resolution monitor, the CSS `min-height: 0.5px` was not sufficient to guarantee a visible line in all browsers. I could guarantee a line if I set `min-height: 0.265mm`, but then some lines are too thick if one zooms in, as in:
border-image

I am going to shut this PR down and spend a little more time on PR #1004.

Issue #676, opened 2017-04-06
Text font commands don't stack right


edemaine:
`\textbf{\textit{x}}` should produce a bold-italic "x", but it currently produces a nonbold italic "x" because all KaTeX commands reset the entire font.  My understanding is that this is correct for math font commands (`\mathbf{\mathit{x}}` correctly produces a nonbold italic "x"), but incorrect for text font commands.  (How asymmetric...)  By contrast, `\text{\textbf{\it x}}` (using `\it` implemented in #675) should produce a nonbold italic "x", because old font commands like `\it` *should* reset the font.

~~This would be a moot point if there weren't bold italic fonts available, but there are.~~ This would also require creating Main-BoldItalic fonts; see below.

kevinbarabash:
Nice find.

ronkok:
@kevinbarabash Many issues, like this one, are blocked by missing font glyphs. It would be helpful if we had a single list that identified all of the desired glyphs. Then, if someone took on the font task, they would have a ready-made TO DO list. I wouldn't mind putting together such a list.

Would you rather have it in an issue or a project?

kevinbarabash:
@ronkok a project would be a good way to track this kind of stuff.  Updating fonts is currently blocked on creating a separate repo for fonts.  I'll try to tackle this this weekend.

rrandallcainc:
Was going to open an issue on this, but just came across this. 

I'm attempting to do: \textsf{\textbf{Hello world}}, and while it renders in bold, it does not render in Sans Serif font.

Is there any update on this issue? What can I do to help?

kevinbarabash:
@rrandallcainc it looks like we have the fonts.  I think `Option#withFont` needs to be updated to so that instead of simply replacing the `.font` property, we'd modify it.  It might even make sense to turn `.font` into `.fonts` so that we can toggle the active fonts by adding/removing them from the list of fonts.  We'll also need to apply one or more font classes from https://github.com/Khan/KaTeX/blob/master/static/katex.less.

rrandallcainc:
@kevinbarabash Okay, I'll give this a shot. Thanks.

rrandallcainc:
@kevinbarabash I noticed in: https://github.com/Khan/KaTeX/wiki/Function-Support-in-KaTeX that there's no support yet for a bold/itatlic Sans Serif font (along with others), but it looks like they are contained inside of the "fonts" folder. Do we know what's blocking these from being exposed?

[...]

PR #1009, opened 2017-12-04
Stacking text commands

rrandallcainc:
Per: https://github.com/Khan/KaTeX/issues/676

Sorry for the delay on this one. Been busy lately :(

Note: This does not include `Main-BoldItalic`.


kevinbarabash:
@rrandallcainc I'll try to look at it this weekend.

src/buildCommon.js
```diff
@@ -139,12 +139,15 @@ const mathDefault = function(
     } else if (type === "textord") {
         const font = symbols[mode][value] && symbols[mode][value].font;
         if (font === "ams") {
+            const fontName = retrieveTextFontName("amsrm", options.fontStyles);
             return makeSymbol(
-                value, "AMS-Regular", mode, options, classes.concat(["amsrm"]));
+                value, fontName, mode, options,
+                classes.concat(options.fontStyles, "amsrm"));
```
kevinbarabash: I feel like we should just pass `fontStyles` to `makeSymbol` so that the caller isn't responsible for calling `retrieveTextFontName` and concatenating classes.

src/Options.js
```diff
@@ -60,6 +61,7 @@ class Options {
     textSize: number;
     phantom: boolean;
     font: string | void;
+    fontStyles: Array<string>;
```
kevinbarabash: I think changing `font` to `baseFont` or `fontFamily` would make it clearer that the two are necessary and work together.

src/buildCommon.js
```diff
@@ -536,6 +548,60 @@ const makeVerb = function(group: ParseNode, options: Options): string {
     return text;
 };
 
+// Takes an Options object, and returns the appropriate fontLookup
+const retrieveTextFontName = function(
+    font: string,
+    fontStyles: Array<string>
+): string {
+    const baseFontName = retrieveBaseFontName(font);
+    const fontStylesName =  retrieveFontStylesName(fontStyles);
```
kevinbarabash: nit: extra space

src/buildCommon.js
```diff
@@ -536,6 +548,60 @@ const makeVerb = function(group: ParseNode, options: Options): string {
...
+    return baseFontName;
+};
+
+const retrieveFontStylesName = function(fontStyles: Array<string>): string {
+    let fontStylesName = '';
+    if (fontStyles.indexOf("textbf") > -1 || fontStyles.indexOf("textit") > -1) {
+        fontStyles.indexOf("textbf") > -1 && (fontStylesName += "Bold");
+        fontStyles.indexOf("textit") > -1 && (fontStylesName += "Italic");
+    } else {
+        fontStylesName += "Regular";
```
kevinbarabash: It's kind of hard to grok this, maybe:
```
let fontStylesName = '';
if (fontStyles.indexOf("textbf")) {
    fontStylesName += "Bold";
}
if (fontStyles.indexOf("textit")) {
   fontStylesName += "Italic";
}
return fontStylesName || "Regular";
```

test/katex-spec.js
```diff
@@ -1590,31 +1590,38 @@ describe("An HTML font tree-builder", function() {
...
         const markup = katex.renderToString("\\textsf{R}");
         expect(markup).toContain("<span class=\"mord mathsf\">R</span>");
     });
 
+    it("should render \\textsf{\\textit{R}G\\textbf{B}} with the correct font", function() {
+        const markup = katex.renderToString("\\textsf{\\textit{R}G\\textbf{B}}");
+        expect(markup).toContain("<span class=\"mord textit mathsf\">R</span>");
+        expect(markup).toContain("<span class=\"mord mathsf\">G</span>");
+        expect(markup).toContain("<span class=\"mord textbf mathsf\">B</span>");
+    });
```
kevinbarabash: You might want to try writing snapshot tests for new tests that verify things about the HTML markup we're producing.  See mathml-spec.js for examples.

test/katex-spec.js
```diff
@@ -1590,31 +1590,38 @@ describe("An HTML font tree-builder", function() {
 
     it("should render \\textit{R} with the correct font", function() {
         const markup = katex.renderToString("\\textit{R}");
-        expect(markup).toContain("<span class=\"mord textit\">R</span>");
+        expect(markup).toContain("<span class=\"mord textit mathrm\">R</span>");
```
kevinbarabash: It's kind of strange to see the `mathrm` class being applied to text that's inside `\textit`.  Is this necessary?

test/katex-spec.js
kevinbarabash: It would be good to have some tests that compare the markup produced for things that should be similar, e.g.

I would expect
```
\textsf{\textbf{mathrm{A}}}
```
to produce the same markup as
```
\mathrml{A}
```

A little bit more complicated but
```
\textsf{\textbf{\mathrm{\textsf{A}}}}
```
is equivalent to
```
\textsf{\textbf{A}}
```
I checked using quicklatex.com which I believe calls actual LaTeX on its backend.

src/Options.js
kevinbarabash: According to the answer to https://tex.stackexchange.com/questions/5008/when-to-use-bold-italics-small-caps-typewriter-etc there are three axes for fonts:
- The shape axis covers: \textup, \textit, \textsc and \textsl
- The family axis covers: \textrm, \textsf, \texttt
- The weight axis covers: \textbf, \textmd

Maybe we should change from `font` + `fontStyles` array to three props:
- `fontShape`
- `fontFamily`
- `fontWeight`

rrandallcainc:
@kevinbarabash Thanks for the quick feedback. Will take a look tomorrow or the wekeend.

rrandallcainc:
@kevinbarabash This should be ready for another look. I opted to keep the old testing method as it looked more verbose/readable. I did end up adding the tests you suggested however.

test/__snapshots__/katex-spec.js.snap
```diff
@@ -10,6 +10,10 @@ exports[`An HTML font tree-builder should render \\textbf{R} with the correct fo
...
 
+exports[`An HTML font tree-builder should render \\textsf{\\textbf{$\\mathrm{\\textsf{A}}$}} with the correct font 1`] = `"<span class=\\"katex\\"><span class=\\"katex-mathml\\"><math><semantics><mrow><mstyle scriptlevel=\\"0\\" displaystyle=\\"false\\"><mrow><mtext>A</mtext></mrow></mstyle></mrow><annotation encoding=\\"application/x-tex\\">\\\\textsf{\\\\textbf{$\\\\mathrm{\\\\textsf{A}}$}}</annotation></semantics></math></span><span class=\\"katex-html\\" aria-hidden=\\"true\\"><span class=\\"strut\\" style=\\"height:0.69444em;\\"></span><span class=\\"strut bottom\\" style=\\"height:0.69444em;vertical-align:0em;\\"></span><span class=\\"base\\"><span class=\\"mord text\\"><span class=\\"mord text\\"><span class=\\"mord\\"><span class=\\"mord text\\"><span class=\\"mord textsf textbf\\">A</span></span></span></span></span></span></span></span>"`;
```
kevinbarabash: I'll have to see if I can make it printout nice snapshots like the ones from mathml-spec.js.

test/katex-spec.js
kevinbarabash: I thought the HTML trees rendered would be equivalent between `\textsf{\textbf{$\mathrm{A}$}}` and `\mathrm{A}`, but they're not b/c the first one has extra spans.  It's just that the visual output looks the same... unfortunately that's hard to test.  :(

test/__snapshots__/katex-spec.js.snap
```diff
@@ -1,23 +1,5 @@
...
-
-exports[`An HTML font tree-builder should render \\textsf{R} with the correct font 1`] = `"<span class=\\"katex\\"><span class=\\"katex-mathml\\"><math><semantics><mrow><mtext>R</mtext></mrow><annotation encoding=\\"application/x-tex\\">\\\\textsf{R}</annotation></semantics></math></span><span class=\\"katex-html\\" aria-hidden=\\"true\\"><span class=\\"strut\\" style=\\"height:0.69444em;\\"></span><span class=\\"strut bottom\\" style=\\"height:0.69444em;vertical-align:0em;\\"></span><span class=\\"base\\"><span class=\\"mord text\\"><span class=\\"mord textsf\\">R</span></span></span></span></span>"`;
-
```
kevinbarabash: I totally get why you reverted these.  Thanks for trying.  I'll try to make the snapshots better in the future.  The tests you have are good for now.

metrics/mapping.pl
```diff
@@ -1004,6 +1004,94 @@
...
+    [9,0xA] => 0x3A8,       # \Psi, \Omega
+
+    0x10 => 0x131,          # \imath (roman)
+    0x11 => 0x237,          # \jmath (roman)
+    0x12 => [0x300,-550,0], # \grave (combining)
+    0x13 => [0x301,-550,0], # \acute (combining)
+    0x14 => [0x30C,-550,0], # \check (combining)
+    0x15 => [0x306,-550,0], # \breve (combining)
+    0x16 => [0x304,-550,0], # \bar (combining)
+    0x17 => [0x30A,-608,0], # ring above (combining)
```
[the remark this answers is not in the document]
kevinbarabash: Did you copy/paste this from https://github.com/Khan/MathJax-dev/blob/master/fonts/OTF/TeX/makeFF#L1400-L1405?

package.json
```diff
@@ -58,13 +59,14 @@
...
     "jest": "jest",
+    "jest-update": "jest --updateSnapshot",
     "coverage": "jest --coverage",
     "copy": "cp -a static/. build/ && cp -a contrib build/",
     "clean": "rm -rf build/* node_modules/",
     "clean-install": "npm run clean && npm i",
     "test": "check-dependencies && npm run lint && npm run flow && npm run jest",
     "build-css": "lessc --clean-css static/katex.less build/katex.css",
-    "prestart": "npm run build-css && npm run copy",
+    "prestart": "rimraf build && npm run build-css && npm run copy",
```
kevinbarabash: What was happening if `build` was still around?

src/Options.js
```diff
@@ -59,7 +61,9 @@ class Options {
     size: number;
     textSize: number;
     phantom: boolean;
-    font: string | void;
+    fontFamily: string | void;
```
kevinbarabash: I think it's fine leaving this as is, but we might want to change this to `fontFamily: string` in a follow up diff to match `fontWeight` and `fontShape`.  Could you add a `TODO`?

metrics/mapping.pl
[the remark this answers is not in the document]
rrandallcainc: Ah, yes that's where I grabbed them from :). Thought you were saying there was something wrong with the shifts.

src/Options.js
rrandallcainc: Done

package.json
rrandallcainc: https://stackoverflow.com/questions/20705677/mv-cannot-overwrite-directory-with-non-directory

Couldn't overwrite a directory with the same name.

src/functions/text.js
```diff
@@ -7,17 +7,28 @@ import * as html from "../buildHTML";
 import * as mml from "../buildMathML";
 
 // Non-mathy text, possibly in a font
-const textFunctionFonts = {
-    "\\text": undefined, "\\textrm": "mathrm", "\\textsf": "mathsf",
-    "\\texttt": "mathtt", "\\textnormal": "mathrm", "\\textbf": "mathbf",
+const textFontFamilies = {
+    "\\text": undefined, "\\textrm": "textrm", "\\textsf": "textsf",
+    "\\texttt": "texttt", "\\textnormal": "textrm",
```
kevinbarabash: The reason for using `mathrm`, `mathsf`, etc. here is that that's what's used in `fontMap` in `buildCommon`.  Not quite sure how this was working without updating `fontMap`, it'd be good to figure that out.  Maybe we can get rid of it.  We probably should separate the font family from whether we're in math mode or text mode.  Maybe we should change this to:
```
"\\text": undefined, "\\textrm": "rm", "\\textsf": "sf", ...
```

Same thing for font weights and font shapes as we have both `\\textbf` and `\\mathbf` as well as `\\textit` and `\\mathit` commands.

We'll also need to update `fontMap` if it's still needed.

rrandallcainc: Font map is now strictly for math functions, and old text commands (\rm, \sf, etc.). The reason is the two are treated differently.

Text allows the fontFamily/fontWeight/fontShape to stack. In other words you can't simply use the last function to determine what the font is.

i.e. `\textsf{\textbf{H}}` combines the two to make a bold SansSerif H.

This means that the old fontMap won't work either wont work well for text, or will need to be very verbose (contain every permutation of family, weight, shape).

Math and old font commands change the font as a whole.

i.e. `\mathsf{\mathbf{H}}` will make a bold, non-SansSerif H.

The old fontMap works fine here.


The determination for which route to go with happens here:

https://github.com/Khan/KaTeX/pull/1009/files#diff-91067d2948380956002a57296faf3c18R208

kevinbarabash: Cool.  I didn't realize that the math font commands behaved differently from the text commands.  Thanks for the explanation.

rrandallcainc:
Thanks for the review @kevinbarabash

kevinbarabash:
@rrandallcainc great work!

PR #1017, opened 2017-12-11
Exposing the build tree.

rrandallcainc:
As a user I would like to expose Katex's buildTree for more advanced use cases (like rendering to a custom output other than DOM)

While this [thread](https://github.com/Khan/KaTeX/issues/376) tracks an intermediate representation, it would be nice if users could leverage what exists in the meantime (with the caveat that it's likely to change).

This PR follows a similar convention to the parseTree in that user's are welcome to use it, but are given an appropriate warning message.

### DIFF
--- a/katex.js
+++ b/katex.js
@@ -28,12 +28,7 @@ let render = function(
     options: SettingsOptions,
 ) {
     utils.clearNode(baseNode);
-
-    const settings = new Settings(options);
-
-    const tree = parseTree(expression, settings);
-    const node = buildTree(tree, expression, settings).toNode();
-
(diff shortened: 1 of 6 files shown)


kevinbarabash:
@rrandallcainc looks reasonable to me.

katex.js
```diff
@@ -86,5 +91,6 @@ module.exports = {
      * to change. Use at your own risk.
      */
     __parse: generateParseTree,
+    __getBuildTree: generateBuildTree,
```
kevinbarabash: Can you add a warning above this function as well?

kevinbarabash:
Thanks for the PR.

PR #1019, opened 2017-12-12
\kern generates right-margin instead of left-margin. Fix #995

### LINKED ISSUE #995: \mbin{...}\kern{...} doesn't stack spacing
Originally reported by @akalin in #982:
```latex
\begin{aligned}
& (a \cdot b) \bmod 257 \\
& (a \cdot b) \mathbin{\mathrm{mod}} 257 \\
& (a \cdot b) \kern1mu\mathbin{\mathrm{mod}}\kern1mu 257
\end{aligned}
```
produces:
screen shot 2017-11-23 at 2 04 04 pm

The last row is incorrect (relative to LaTeX).  I've figured out the reason, but I'm not sure of the solution:

* We have a CSS rule `.katex .mbin+.mord { margin-left: .22222em; }` that should give the object after an mbin a space before it of `.22222em`.
* But the `\kern1mu` makes an element with class `mord rule` with a *manual style* of `margin-left: 0.0555556em;`.
* So instead of the spaces adding up as they do in LaTeX, the second one overrides the first one. ☹️ 

Thoughts on how to fix this?

edemaine:
Fix #995 by making `\kern` generate `padding-left` instead of `margin-left` manual CSS.  Then the padding stacks with the margin.  This also makes some intuitive sense to me: a kern (space) makes sense as a padding not a margin.

I don't know for sure that this captures *all* aspects of accidental non-stacking of margins or paddings, but I looked through the code, and I couldn't think of any other ways it could fail (before or after this change).


edemaine:
Well, the screenshot tests revealed the problem: negative padding isn't possible, so negative kerns fail.

edemaine:
I found an actual fix (included in the latest commit), though this one feels a bit more hacky: represent `\kern` as an object with manual `margin-right`.  This should avoid stacking conflicts because the CSS spacing rules always assign a `margin-left` to the second object.  And because it's a margin, it can be negative.  Screenshot tests now seem to work.

kevinbarabash:
@edemaine we should have a screenshot test for:
```
\begin{aligned}
& (a \cdot b) \bmod 257 \\
& (a \cdot b) \mathbin{\mathrm{mod}} 257 \\
& (a \cdot b) \kern1mu\mathbin{\mathrm{mod}}\kern1mu 257
\end{aligned}
```

edemaine:
Screenshot test added (with slightly different order of lines).  Here's a texcmp, which looks pretty identical:

texcmp

kevinbarabash:
The tests failed because there's an issue with eslint 4.14.  I committed my local package-lock.json file which had an earlier version of eslint.  Hopefully that should fix the issue.

PR #1022, opened 2017-12-14
Add buildHTMLTree

edemaine:
This is a followup to #1017 (Exposing the build tree).  It includes a new `generateBuildHTMLTree` function that skips the MathML step, and exposes that function as `__getBuildHTMLTree`. See #800 for discussion.

I'd like to have a discussion about the naming of these functions.  I find the naming of these functions, and the `generateBuildTree` and `__getBuildTree` that come from #1017, to be "incorrect".  I don't think these were ever supposed to be called "build trees".  "Build" in `buildTree` was meant as a verb, as in "build a tree".  I think it would make more sense to rename `buildTree` to `domTree` or something.  Alternatively, we could rename `generateBuildTree` to `parseAndBuildTree` or something like this, which is what it's doing.  I also am not a fan of the `get` in `__get`.  If there's consensus, I can rename in this PR.

### DIFF
--- a/src/buildTree.js
+++ b/src/buildTree.js
@@ -9,24 +9,19 @@ import Style from "./Style";
 import type ParseNode from "./ParseNode";
 import type domTree from "./domTree";
 
-const buildTree = function(
+const optionsFromSettings = function(settings: Settings) {
+    return new Options({
+        style: (settings.displayMode ? Style.DISPLAY : Style.TEXT),
+        maxSize: settings.maxSize,
+    });
+};
+
+export const buildTree = function(
     tree: ParseNode[],
     expression: string,
     settings: Settings,
 ): domTree.span {
-    settings = settings || new Settings({});
-
-    let startStyle = Style.TEXT;
-    if (settings.displayMode) {
-        startStyle = Style.DISPLAY;
-    }
-
-    // Setup the default options
-    const options = new Options({
-        style: startStyle,
-        maxSize: settings.maxSize,
-    });
-
+    const options = optionsFromSettings(settings);
     // `buildHTML` sometimes messes with the parse tree (like turning bins ->
     // ords), so we build the MathML version first.
     const mathMLNode = buildMathML(tree, expression, options);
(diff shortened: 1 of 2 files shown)


src/buildTree.js
```diff
@@ -43,4 +38,19 @@ const buildTree = function(
...
 };
 
+export const buildHTMLTree = function(
+    tree: ParseNode[],
+    expression: string,
+    settings: Settings,
+): domTree.span {
+    const options = optionsFromSettings(settings);
+    const htmlNode = buildHTML(tree, options);
+    const katexNode = buildCommon.makeSpan(["katex"], [htmlNode]);
```
rrandallcainc: Is the katexNode necessary here? Why not just return `htmlNode`?

edemaine:
@rrandallcainc I think because we haven't released a version with this change (right?) it's probably OK, though obviously not ideal.

src/buildTree.js
edemaine: Hmm, good question.  `htmlNode` would be natural too.  My reasoning here was that there are various CSS rules that only apply with the `.katex-display .katex` nesting...

rrandallcainc: Ah okay.

rrandallcainc:
Yup, agreed that this should not be an issue. Forgot it hasn't been released.

rrandallcainc:
@edemaine These changes look fine to me. What were you thinking for the new naming convention?

edemaine:
How about the following?
* `__getBuildTree` &rarr; `__renderToDomTree` (by analogy to `renderToString`)
* `generateBuildTree` &rarr; `renderToDomTree`
* `__getBuildHTMLTree` &rarr; `__renderToHTMLTree`
* `generateBuildHTMLTree` &rarr; `renderToHTMLTree`

Hmm, maybe we should have `renderToMathMLTree` and/or `renderToMathML`.  I guess we can wait for someone to request that feature.

rrandallcainc:
Using the existing "render" nomenclature sounds good. I agree that consistency would be nice here.

edemaine:
OK, I did the renames. I also added some clarifying documentation so it's clear what the different functions provide.

It occurs to me that a "no MathML" option would still be neat in that it would allow `renderToString` to also render HTML only.  But maybe it's best not to encourage that behavior...  One can achieve it now via `__renderToHTMLTree(...).toMarkUp()`.

kevinbarabash:
:tada:

edemaine:
@rrandallcainc Thanks for the quick review!

Issue #1003, opened 2017-11-29
conditional support in macros


kevinbarabash:
Add support for `\if`, `\fi`, `\ifx`, `\else`, `\ifmmode`.  Not sure what other conditionals may be useful.

kevinbarabash:
We already have some conditionals thanks to @edemaine in #975.  In particular `\@firstoftwo`, `\@ifstar`, and `\@ifnextchar` were added.  Maybe we should also add `\@secondoftwo` as well.

kevinbarabash:
The goal here is to provide a solid base of macro support so that people can extend KaTeX functionality by using existing macros with minor or no modification from existing packages (or their own personal macros).

kevinbarabash:
All functions as macros sounds like a big change, but it would make things more regular.  Are there any other macro abilities a change like this would enable beyond allowing us to implement `\TextOrMath`?

edemaine:
Conditionals that involve changing things might be hard to get correct without more exactly matching LaTeX semantics.

Reviewing some older issues, I noticed that #924 is a (probably minor?) bug that probably can't be fixed without changing functions into macros.

In general, I believe it's all in the subtle details.  But makes it harder to implement certain functionality as macros.

kevinbarabash:
Cool.  We should probably make this change at some point.  I think we should finishing add flow types to Parser.js and extracting functions into the `functions/` folder before starting this.  I'm not sure how we'd do this though.  Macros are all about modifying which tokens get sent to the parser.  We still need to define handlers to transform those tokens to into chunks of the parse tree.  How does TeX handle this if it's macros all the way down?

edemaine:
@kevinbarabash I've admittedly never dug this deep, but I believe the answer is that there are some commands which are not truly macros -- they have side-effects -- in particular one called `\output` that sends stuff to the DVI/PDF output file.

Thinking more, there does seem to be a real difference between `\frac`, whose arguments *must* be complete expressions, and `\TextOrMath` or other conditionals.  So maybe the separation between macros and functions isn't *so* bad... but I think we need tighter integration between `MacroExpander` and `Parser`.  If `MacroExpander` knew about functions, it could plausibly grab functions and their arguments (instead of it being `Parser`'s job) to solve #924.  And maybe `Parser` could reveal to `MacroExpander` whether we're in math vs. text mode, to enable e.g. a `\TextOrMath` macro.

kevinbarabash:
[...]

Issue #924, opened 2017-10-09
Difficult interaction between MacroExpander and Parser


edemaine:
If `\foo` expands to a space (`\def\foo{ }`) or nothing (`\def\foo{}`), LaTeX will treat `\frac\foo\foo` just like `\frac{}{}`, i.e., two empty arguments. In KaTeX, in both cases, the arguments will get ignored (in math mode), so `\frac\foo\foo x y` will act just like `\frac xy`.

Another example: with `\def\foo{foo}`, `\text\foo` is equivalent to `\text{foo}` in LaTeX, but is treated as `\text foo` i.e. `\text{f}oo` in KaTeX.

One way to fix this is to move all function support into `MacroExpander` instead of `Parser`.  I think a possible simpler fix is for `MacroExpander` to grab a function and its arguments, and then pass them on to `Parser` for actual handling.

PR #1024, opened 2017-12-16
Implement \TextOrMath, \@secondoftwo

edemaine:
Partially addressing #1003.
* Parser now tells MacroExpander about mode switching.  (This seems preferable to a circular reference between Parser and MacroExpander.)
* Implement `\TextOrMath`
* Improve when we switch modes so that this actually works, in all cases except single-symbol arguments.
* Add `\@secondoftwo` to match `\@firstoftwo`.
* Add comments documenting all the conditional macros


test/katex-spec.js
```diff
@@ -2672,11 +2672,61 @@ describe("A macro expander", function() {
         expect("\\@ifnextchar!{yes}{no}?!").toParseLike("no?!");
     });
 
-    it("\\@firstoftwwo should consume star but nothing else", function() {
+    it("\\@ifstar should consume star but nothing else", function() {
```
kevinbarabash: good catch

test/katex-spec.js
```diff
@@ -2672,11 +2672,61 @@ describe("A macro expander", function() {
...
+    it("\\TextOrMath should work immediately after $ ends", function() {
+        expect("\\text{$\\TextOrMath{text}{math}$\\TextOrMath{text}{math}}")
+            .toParseLike("\\text{$math$text}");
+    });
+
+    it("\\TextOrMath should work in a macro", function() {
+        compareParseTree("\\mode\\text{\\mode$\\mode$\\mode}\\mode",
+            "math\\text{text$math$text}math",
+            {"\\mode": "\\TextOrMath{text}{math}"});
+    });
```
kevinbarabash: Great test suite!

test/katex-spec.js
```diff
@@ -2672,11 +2672,61 @@ describe("A macro expander", function() {
...
+
+    // TODO(edemaine): This doesn't work yet.  Parses like `\text math`,
+    // which doesn't even treat all four letters as an argument.
+    //it("\\TextOrMath should work in a macro passed to \\text", function() {
+    //    compareParseTree("\\text\\mode", "\\text{text}",
+    //        {"\\mode": "\\TextOrMath{text}{math}"});
```
kevinbarabash: What does this currently result in?

src/MacroExpander.js
```diff
@@ -16,13 +16,22 @@ export default class MacroExpander implements MacroContextInterface {
...
         this.lexer = new Lexer(input);
         this.macros = objectAssign({}, builtinMacros, macros);
+        this.mode = mode;
         this.stack = []; // contains tokens in REVERSE order
     }
 
+    /**
+     * Switches between "text" and "math" modes.
+     */
+    switchMode(newMode) {
```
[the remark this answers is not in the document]
edemaine: Added.

test/katex-spec.js
edemaine: It results in "m" formatted in text mode followed by "ath" formatted in math mode. So it's wrong in two ways: it thinks it's in math mode when it's actually in text mode, and the argument is just the first character instead of all four. This is a (kind of known) bug with interaction between `MacroExpander` and `Parser`.

src/Parser.js
```diff
@@ -976,6 +989,7 @@ export default class Parser {
                     firstToken.range(lastToken, firstToken.text));
         } else {
             // Otherwise, just return a nucleus, or nothing for an optional group
+            // TODO: Not yet switching modes for single-symbol group.
```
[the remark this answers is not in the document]
edemaine: I'm not actually sure how to do it -- it's pretty challenging, because by the time `parseGroup` gets called, `nextToken` *already* has the single-symbol group, which means that it already got expanded in the wrong mode.  In the old way of doing things, we could "unget" the symbol, change modes, and reget it, but that seems challenging/impossible to do if there are side-effects/errors caused by the first execution in the wrong mode.  So I think we'd need to call `parseGroup` before actually reading the first symbol of the group, which might cause a rather big change across all of `Parser.js`.

So in summary, I'm not sure how much work it'll be.  It's surely possible, but hard.  I thought I'd commit what I have so far, and open another issue for fixing this further.  On the plus side, it took me a lot of playing around with tests before I could even get this situation to arise, and in some sense #924 is even more important than this missing case.  So maybe this PR is good to merge for now?  Up to you -- I can look at this more when I have time.

kevinbarabash:
Code changes look good to me, but there's something causing the `Colorbox` and `DashesAndQuotes` screenshot tests to fail.

**ColorBox:**
screen shot 2017-12-19 at 8 39 02 pm
The `x` should be in typewriter.

**DashesAndQuotes:**
screen shot 2017-12-19 at 8 38 30 pm
Three dashes in a row should be an em dash.

edemaine:
Wow, the power of tests. DashesAndQuotes was a nice easy fix pointing out incorrect mode detection for the ligatures (obvious in hindsight).  Unfortunately, the Colorbox issue is exactly the missing part of this PR, switching to text mode for a single-symbol argument (here, `x`).  So maybe I *should* try to do that before this PR gets merged; otherwise, this PR would regress behavior a bit.

kevinbarabash:
@edemaine the following seems to be sufficient to fix the `ColorBox` screenshot test:
```
        } else {
            // Otherwise, just return a nucleus, or nothing for an optional group
            if (mode) {
                this.switchMode(mode);
            }
            const result = optional ? null : this.parseSymbol();
            if (mode) {
                this.switchMode(outerMode);
            }
            return result;
        }
```

kevinbarabash:
There may still be other places where we need to switch modes before calling `parseSymbol`, but we can deal with them in a separate PR as necessary.

edemaine:
@kevinbarabash Thanks for investigating!  This partially fixes the single-symbol support, basically whenever the single symbol is not a macro (not expanded by the `MacroExpander`).  I'm glad this makes it mergable (backwards-compatible); now I'll open an issue for the remaining challenge of single-symbol macro.

PR #1025, opened 2017-12-19
Fix frac-line

ronkok:
For frac-line, use an SVG line in an extra tall span.

I still need to think of a way to adust `vertical-separator`.


ronkok:
> Were the vertical separators disappearing at times? Won't this change to the styling affect the horizontal layout slightly?

Revisiting my response to this comment:
1. Yes. Issue #970 
2. No. It turns out that `vertical-separator` lines do nothing to the inter-column spacing of an `{array}`. In the screen shot below, I've drug part of a no-line array over and overlaid it onto an `{array}{c|c}`. The columns still align perfectly.

In the code for this PR, it was only necessary to adjust the `vertical-separator` negative left margin. No action is necessary to further fix the horizontal layout.

I believe that we're good to go.

fracline

kevinbarabash:
@ronkok you'll have to merge `master` into your branch and resolve the conflict.  I can regenerate those screenshots afterwards.

kevinbarabash:
@ronkok thank you for tackling this this issue with the fraction bars disappearing.  I'm going to publish a release over the next couple of days.  With this change I think it's safe to make it a beta release or release candidate.  :)

ronkok:
@kevinbarabash You're very welcome. I certainly think that we've now got something better than when we started. I hope it's enough.

PR #1029, opened 2017-12-22
add \i and \j for text mode

kevinbarabash:
screen shot 2017-12-22 at 11 31 00 am


kevinbarabash:
@rrandallcainc I added a test to verify that `\\text{\\i\\j}` parses.  A screenshot test is probably overkill for this as there a lots of symbols we don't have screenshot tests for.

PR #1030, opened 2017-12-22
add \ae, \AE, \oe, \OE, \o, \O, \ss with unicode support

kevinbarabash:
screen shot 2017-12-22 at 12 00 52 pm


test/katex-spec.js
```diff
@@ -2843,4 +2843,10 @@ describe("Symbols", function() {
     it("should parse \\text{\\i\\j}", () => {
         expect("\\text{\\i\\j}").toParse();
     });
+
+    it("should render ligature commands like their unicode characters", () => {
+        const commands = getBuilt("\\text{\\ae\\AE\\oe\\OE\\o\\O\\ss}");
+        const unicode = getBuilt("\\text{æÆœŒøØß}");
+        expect(commands).toEqual(unicode);
```
edemaine: You can use `toParseLike` instead of this: `expect("\\text{...}").toParseLike("\\text{...}")`

kevinbarabash: I tried that, but they don't actually parse the same.

kevinbarabash:
> Technically, some lines in fontMetrics.js such as this one should be removed. You can also feel free to leave this to #992 if you'd like.

I can remove those.

test/katex-spec.js
edemaine: Ah, I see, `getBuilt` renders to HTML.  I didn't realize that -- it suggests some other tests we could do!

src/fontMetrics.js
```diff
@@ -103,7 +103,6 @@ const extraCharacterMap = {
     'Ã': 'A',
     'Ä': 'A',
     'Å': 'A',
-    'Æ': 'A',
```
kevinbarabash: There weren't any entries for `Œ` or `œ`.

edemaine:
@kevinbarabash What are your thoughts about supporting these commands in text mode?

kevinbarabash:
@edemaine I assume you mean "math mode".  I didn't realize that they appear in both.  I'll put up another PR for that.

edemaine:
@kevinbarabash "LaTeX throws warnings (`LaTeX Warning: Command \ae invalid in math mode on input line 7.`) if we use these commands in math mode, but technically it does support them (generating textords)."

kevinbarabash:
@edemaine I was testing using http://quicklatex.com.  I tried it again with pdflatex locally and am seeing the same issue you are.  In that case let's not bother.  The change was non trivial to b/c while we have the glyphs in KaTeX_Main-Italic they need to either be added to KaTeX_Math-Italic or we would've had to special case things in the code.

edemaine:
If they're specified as textords, I thought they'd be rendered in text mode using text fonts. I could be wrong though. (That is also what LaTeX does: it renders them in Roman font by default.)

kevinbarabash:
For text mode they are textords and they do get rendered using `KaTeX_Main-*` which are the text fonts.

PR #1032, opened 2017-12-22
add glyph widths to fontMetricsData.js

kevinbarabash:


kevinbarabash:
@edemaine if any of the widths are off for the accents let me know.

src/fontMetricsData.js
```diff
@@ -1,1998 +1,1998 @@
...
+        "120": [0, 0.43056, 0.12042, 0, 0.46389],
+        "121": [0.19444, 0.43056, 0.08847, 0, 0.48555],
+        "122": [0, 0.43056, 0.12292, 0, 0.40889],
+        "126": [0.35, 0.31786, 0.11585, 0, 0.51111],
+        "163": [0, 0.69444, 0, 0, 0.76909],
+        "305": [0, 0.43056, 0, 0.02778, 0.32246],
+        "567": [0.19444, 0.43056, 0, 0.08334, 0.38403],
+        "768": [0, 0.69444, 0, 0, 0.51111],
+        "769": [0, 0.69444, 0.09694, 0, 0.51111],
+        "770": [0, 0
```
edemaine: I tried verifying these two characters manually with [opentype.js Glyph Inspector](https://opentype.js.org/glyph-inspector.html).  0.27778 seems like a good approximation to the `advanceWidth` of `!` which is 278 (after dividing by 1000), and ditto for the next few characters.  But I'm confused that the space character (32) has width 0.  It seems to have `advanceWidth` of 250...  Should we fix this now?

kevinbarabash: The glyph inspector shows xMin, yMin, xMax, yMax to be `undefined` for the `space` character so a width and height of 0 for that character makes sense to me.  There is no black mark to measure the width or height of.

I'm not really sure why these metrics have more decimal places then should occur when dividing an integer by 1000.

PR #1033, opened 2017-12-22
 Make accents zero width

### LINKED ISSUE #1028: \ddot\imath is too wide
`\ddot\imath` should render `ï` with the width of an `i`, not the width of the umlaut accent.  I think the fix would be to render the accent with style `width:0px`, but then we need to fix the horizontal offset to be half the width of the accent character.  So I believe this depends on #587 / #632 to add the width of characters (specifically the accent characters) to `fontMetricsData.js`.

edemaine:
Make the width of an accented character equal to the width of the character, by making accents zero width.  In particular, fixes #1028 (`\ddot\imath`).

This involved reworking all of the accent offset machinery, in particular skew and accent-hungarian.  A bit cleaner now.

Also added support for the new width character metrics in symbolNode.

A bunch of screenshots involving skew correction have changed in an invisible way (diffs look entirely black and white).  I assume this is a consequence of switching from `margin-left` to `left` as the mechanism for skew shift on accents.


kevinbarabash:
You can see the difference in the spacing in the font tests.

kevinbarabash:
@edemaine does this mean that we don't use combining accent glyphs for anything (besides `\H`) anymore?  If so, we'll want to update all fonts to have all non-combining accents glyphs, many don't at the moment.  Also, we should be able to remove all combining accent glyphs.

kevinbarabash:
@edemaine the screenshots look good for `\H`, but are off for Safari.  https://github.com/KaTeX/katex-fonts/pull/5 will add non-combining glyph for `\H`.  I made other changes in that PR so it's probably best to land this as is and deal with the `\H` Safari issue separately.

src/buildCommon.js
```diff
@@ -733,6 +733,8 @@ const staticSvg = function(value: string, options: Options): domTree.span {
     const svgNode = new domTree.svgNode([path], {
         "width": width + "em",
         "height": height + "em",
+        // Override CSS rule `.katex svg { width: 100% }`
+        "style": "width:" + width + "em",
         "viewBox": "0 0 " + 1000 * width + " " + 1000 * height,
         "preserveAspectRatio": "xMinYMin",
     });
```
kevinbarabash: We should be able to get rid of this once we have non-combining `\vec` as well.  Then all of the non-stretchy accent code should use the same path.

src/buildHTML.js
```diff
@@ -688,6 +688,7 @@ groupTypes.accent = function(group, options) {
             // So now we use an SVG.
             // If Safari reforms, we should consider reverting to the glyph.
             accent = buildCommon.staticSvg("vec", options);
+            accent.width = parseFloat(accent.style.width);
```
kevinbarabash: This is kind of hacky, but since it should disappear once we have non-combining `\vec` I'm okay with this.

src/buildHTML.js
```diff
@@ -696,22 +697,26 @@ groupTypes.accent = function(group, options) {
...
         // The \H character that the fonts use is a combining character, and
-        // thus shows up much too far to the left. To account for this, we add a
-        // specific class which shifts the accent over to where we want it.
+        // thus shows up much too far to the left. To account for this, we add
+        // a manual shift of the width of one space.
         // TODO(emily): Fix this in a better way, like by changing the font
-        let accentClass = null;
         if (group.value.label === '\\H') {
-            accentClass = "accent-hungarian";
+            left += 0.5;  // width of space
```
kevinbarabash: I don't think this will work for all fonts, but since we're going to be adding non-combining double acute accents to all fonts soon it's okay to leave this as is for now.

kevinbarabash: Also, I don't think this comment is correct as the width of a space is 250 (or thereabouts depending on the font).

src/buildHTML.js
```diff
@@ -696,22 +697,26 @@ groupTypes.accent = function(group, options) {
...
         }
 
-        accentBody = makeSpan([], [accent]);
-        accentBody = makeSpan(["accent-body", accentClass], [accentBody]);
-
-        // Shift the accent over by the skew. Note we shift by twice the skew
-        // because we are centering the accent, so by adding 2*skew to the left,
-        // we shift it to the right by 1*skew.
-        accentBody.style.marginLeft = 2 * skew + "em";
+        accentBody.style.left = left + "em";
```
kevinbarabash: I found that change the `left` value in the browser didn't actually change the position of the accent body.  It appears that `.katex .accent>.vlist-t` is the style that's actively centering accents.  Maybe we don't need the glyph width after all.

src/buildCommon.js
edemaine: Agreed. That makes me feel better about the hacky duplication of `width`.

src/buildHTML.js
edemaine: It's true that `vlist` is doing horizontal centering.  That's how accents were able to work before.  The catch is that we need to set the accent to zero width so that it doesn't affect the bounding box.  When that happens, the left edge of the accent gets centered, and the accent itself extends to the right.  Hence the need for the horizontal offset.

I do wonder whether there's a way to do this with CSS, without needing the width of the character...

src/buildHTML.js
edemaine: Ah, right, maybe it's double the width of a space. (I was copy/pasting the comment from the `accent-hungarian` CSS.) It might also be the width of the character, which is 0.5...

edemaine:
@kevinbarabash Can you confirm whether `\H` used to work correctly in Safari? I would have guessed not (at one point I tried to copy the old mechanism from `\H` to `\vec`, and I don't think that worked, even though they're effectively the same).  If my guess is correct, then I think it's safe to merge this PR.  Otherwise we have a small regression, and I leave the decision (whether to wait for `\H` noncombining symbol) up to you.

I pushed a tweak to the comment.

kevinbarabash:
@edemaine let's deal with the non-combining `\H` separately.

ronkok:
@edemaine, You're correct. `\H` does not work in the current release.
combinh

PR #1020, opened 2017-12-12
Add smart quote example, as discussed in #967

edemaine:
This small PR adds a sentence of documentation out of the resolution of #967 (Render smart quotes in math as regular quotes).

PR #1034, opened 2017-12-25
Add a 'common issues' section to the README

kevinbarabash:
Addressed #967, replaces #1020.


README.md
```diff
@@ -86,6 +86,20 @@ the same aspect ratio in KaTeX as in LaTeX.  However, because most browsers
...
+  math containing primes, e.g. `f'`.  This can be worked around by defining a
+  single character macro which changes them back, e.g. `{"’", "'"}`.
+- KaTeX follows LaTeX's rendering of aligned and *matrix environments.  When
+  displaying fractions one above another in these vertical layouts there may
+  not be enough space between rows for people who are used to MathJax's rendering.
+  The distance between rows can be adjust by using `\\[0.1em]` instead of the
+  standard line separator.
+- KaTeX does not support the `align` environment because LaTeX doesn't support
+  `align` in math mode.  You can define a macro that maps `align` to `aligned`
+  to work aroudn this.
```
edemaine: Typo "aroudn". Also some readers may not be familiar with `aligned` (partly because it's amsmath, not regular LaTeX -- at least I didn't know about it until I learned KaTeX), so maybe a short introduction of "`aligned` offers the same functionality but in math mode, so use that instead" before mentioning macro option?

README.md
```diff
@@ -86,6 +86,20 @@ the same aspect ratio in KaTeX as in LaTeX.  However, because most browsers
 default to a larger font size, this typically means that a 1cm kern in KaTeX
 will appear larger than 1cm in browser units.
 
+### Common Issues
+- Many Markdown preprocessors, such as the one that Jekyll and GitHub Pages use,
+  have a "smart quotes" feature.  This changes `'` to `’` which is an issue for
+  math containing primes, e.g. `f'`.  This can be worked around by defining a
+  single character macro which changes them back, e.g. `{"’", "'"}`.
+- KaTeX follows LaTeX's rendering of aligned and *matrix environments.  When
```
edemaine: Extra `*` before "matrix".  Should we add "unlike MathJax" here?

kevinbarabash: I'm going to put `*matrix` inside backticks to get around the italics issues.

edemaine: What is `*matrix`? Maybe it's `matrix*`? (I'm just guessing based on LaTeX conventions.) Does this comment not apply to regular `matrix`?  Also I'd suggest adding backticks to aligned to match.

README.md
edemaine: Nice rewrite!

mbourne:
Small nit:

The distance between rows can be adjust**ed** by using...

README.md
kevinbarabash: Doesn't `*matrix` cover `matrix`, `bmatrix`, `pmatrix`, etc.?  I guess in this case `*` is overloaded... I can just write **all** `matrix` environments.

kevinbarabash: I'm going to leave the **all** out.

PR #1040, opened 2017-12-27
extract sqrt, enclose, and verb into their own files

kevinbarabash:
This follows the same pattern as #999 and #998.


src/buildHTML.js
```diff
@@ -203,52 +202,6 @@ const shouldHandleSupSub = function(group, options) {
     }
 };
 
-/**
- * Sometimes we want to pull out the innermost element of a group. In most
- * cases, this will just be the group itself, but when ordgroups and colors have
- * a single element, we want to pull that out.
- */
-const getBaseElem = function(group) {
```
kevinbarabash: moved to utils.js

src/buildHTML.js
```diff
@@ -203,52 +202,6 @@ const shouldHandleSupSub = function(group, options) {
...
-        return group;
-    }
-};
-
-/**
- * TeXbook algorithms often reference "character boxes", which are simply groups
- * with a single character in them. To decide if something is a character box,
- * we find its innermost group, and see if it is a single character.
- */
-const isCharacterBox = function(group) {
```
kevinbarabash: moved to utils.js

src/functions/enclose.js
```diff
@@ -0,0 +1,188 @@
...
+                {
+                    type: "elem",
+                    elem: inner,
+                    shift: 0,
+                },
+                {
+                    type: "elem",
+                    elem: img,
+                    shift: imgShift,
+                    wrapperClasses: /cancel/.test(label) ? ["svg-align"] : [],
```
kevinbarabash: We were doing `vlist.children[0].children[0].children[1].classes.push("svg-align");` before, but flow didn't like it because not all `children` have great grand children.  This specifies classes to be added to the wrapper spans up front instead of afterwards.

marcianx: Not to mention the original code is much harder to reason about. This quite an improvement.

src/utils.js
```diff
@@ -89,6 +89,85 @@ function clearNode(node: Node) {
...
+        body: Group,
+    |},
+|} | BaseElem;
+
+/**
+ * Sometimes we want to pull out the innermost element of a group. In most
+ * cases, this will just be the group itself, but when ordgroups and colors have
+ * a single element, we want to pull that out.
+ */
+const getBaseElem = function(group: Group) {
```
marcianx: What's the type of the return value? I'm seen at least 2 different types being returned.

src/utils.js
```diff
@@ -89,6 +89,85 @@ function clearNode(node: Node) {
...
+        return group;
+    }
+};
+
+/**
+ * TeXbook algorithms often reference "character boxes", which are simply groups
+ * with a single character in them. To decide if something is a character box,
+ * we find its innermost group, and see if it is a single character.
+ */
+const isCharacterBox = function(group: Group) {
```
marcianx: Missing `boolean` return type annotation.

src/functions/verb.js
```diff
@@ -0,0 +1,53 @@
...
+                const rule = buildCommon.makeSpan(["mord", "rule"], [], newOptions);
+                rule.style.marginLeft = "0.525em";
+                body.push(rule);
+            } else {
+                body.push(buildCommon.makeSymbol(text[i], "Typewriter-Regular",
+                    group.mode, newOptions, ["mathtt"]));
+            }
+        }
+        buildCommon.tryCombineChars(body);
+        // $FlowFixMe: CombinableDomNode[] is not compatible with DomChildNode[]
```
marcianx: Where did `CombinableDomNode[]` get inferred by the type system? I don't see anything explicitly requiring that. Does declaring `body` above explicitly as `DomChildNode[]` resolve this?

marcianx:
(Given the difficulty of checking diffs, could you possibly include subsequent edits in commits separate from the base?)

marcianx:
Also, you might either wish to reference Issue 892 or indicate that this fixes it. Not sure whether we should include the `ParseNode` refactor as part of the issue since it was discussed there.

kevinbarabash:
> Not sure whether we should include the ParseNode refactor as part of the issue since it was discussed there.

We should probably have a separate issue for the refactor.  I think adopting flow types in Parser.js is enough of a change for one PR. :)

src/functions/verb.js
kevinbarabash: The problem is that `tryCombineChars` expects `CombinableDomNode[]` but the call to `makeSpan` below expects `DomChildNode[]` and the two aren't compatible.  I'll update the comment to be more detailed.

src/utils.js
```diff
@@ -127,7 +127,7 @@ type Group = {|
  * cases, this will just be the group itself, but when ordgroups and colors have
  * a single element, we want to pull that out.
  */
-const getBaseElem = function(group: Group) {
+const getBaseElem = function(group: Group): Group | boolean {
```
marcianx: (optional) Just a side node: I think it's more canonical to return `?Group` rather than `Group | boolean`.

src/functions/verb.js
marcianx: There's got to be some way to using bounded polymorphism or some other way of specifying type variance to handle this. However, that requires experimentation, so we can do it separately.

src/utils.js
kevinbarabash: `?Group` would allow the return value to be `null`, but not `false`.

marcianx: That's exactly what I meant (sorry, I wasn't clear). This would require changing the return value to `null` and changing where it's used. That's more in line with what we do elsewhere -- if we're conditionally returning a value, we typically use a nullable.

kevinbarabash: Sorry for the misunderstanding.  I'll change that in a followup PR.

PR #1041, opened 2017-12-28
flow typings for parser.js

### LINKED ISSUE #892: Port Parser.js to @flow

kevinbarabash:
Add flow typings to parser.js.  Fixes #892.


src/Parser.js
```diff
@@ -481,17 +493,38 @@ export default class Parser {
...
             return this.parseFunction();
+        } else if (start.type === "arg") {
+            // Defer to parseGivenFunction if it's not a function we handle
+            return this.parseGivenFunction(start);
         }
 
         const func = start.result;
 
-        if (func === "\\left") {
+        if (func === "$") {
```
kevinbarabash: I moved this up to the top so that we handle the case when `parseSymbol` returns a `Dollar` object first so that `flow` doesn't complain when it sees `oldFontFuncs[func]`.

src/Parser.js
```diff
@@ -511,13 +547,13 @@ export default class Parser {
             // begin...end is similar to left...right
             const begin = this.parseGivenFunction(start);
             const envName = begin.value.name;
-            if (!environments.has(envName)) {
+            if (!environments.hasOwnProperty(envName)) {
```
kevinbarabash: I got rid of the `has`/`get` methods from before b/c `flow` doesn't know how to analyze checks using those methods.

src/Parser.js
```diff
@@ -626,7 +645,7 @@ export default class Parser {
      * @param {ParsedFuncOrArgOrDollar} baseGroup
      * @return {ParseNode}
      */
-    parseGivenFunction(baseGroup) {
+    parseGivenFunction(baseGroup: ParsedFuncOrArgOrDollar) {
```
kevinbarabash: add return type

src/environments/array.js
```diff
@@ -72,6 +72,9 @@ function parseArray(
             break;
         } else if (next === "\\\\" || next === "\\cr") {
             const cr = parser.parseFunction();
```
kevinbarabash: In a followup PR, I'd like to have `parseFunction` throw if it was going return `null`.

src/Parser.js
```diff
@@ -1010,24 +1039,27 @@ export default class Parser {
...
+    formLigatures(group: ParseNode[]) {
         let n = group.length - 1;
         for (let i = 0; i < n; ++i) {
             const a = group[i];
             const v = a.value;
             if (v === "-" && group[i + 1].value === "-") {
                 if (i + 1 < n && group[i + 2].value === "-") {
                     group.splice(i, 3, new ParseNode(
+                        // $FlowFixMe: update ParseNode to accept a location
                         "textord", "---", "text", a, group[i + 2]));
```
kevinbarabash: The last two args are supposed to be `Token`s, but we're passing `ParseNode`s.  Unfortunately, there's no way to get the `Token`s here.

kevinbarabash:
@marcianx thanks for all of the work you did before on this, it made this PR much easier than it otherwise would have been.

src/Parser.js
```diff
@@ -653,18 +672,24 @@ export default class Parser {
...
-     * @param {Token=} token
      */
-    callFunction(name, args, optArgs, token) {
-        const context = {
+    callFunction(
+        name: string,
+        args: ParseNode[],
+        optArgs: (?ParseNode)[],
+        token?: Token,
+    ) {
```
marcianx: Return type is missing. I'm unsure, but it might have to be something generic like `*` before the `ParseNode` refactor.

src/Parser.js
```diff
@@ -891,7 +920,7 @@ export default class Parser {
     /**
      * Parses a color description.
      */
-    parseColorGroup(optional) {
+    parseColorGroup(optional: boolean) {
```
marcianx: Return type missing.

src/Parser.js
```diff
@@ -906,7 +935,7 @@ export default class Parser {
     /**
      * Parses a url string.
      */
-    parseUrlGroup(optional) {
+    parseUrlGroup(optional: boolean) {
```
marcianx: Return type missing.

src/Parser.js
```diff
@@ -923,7 +952,7 @@ export default class Parser {
     /**
      * Parses a size specification, consisting of magnitude and unit.
      */
-    parseSizeGroup(optional) {
+    parseSizeGroup(optional: boolean) {
```
marcianx: Return type missing.

src/Parser.js
marcianx: Those parameter can be changed from `Token`s to `{loc: ?SourceLocation}`, just like the parameters of `SourceLocation.range`.

marcianx:
Also, I forgot to mention: **Nice!**

src/Parser.js
```diff
@@ -529,13 +565,14 @@ export default class Parser {
...
-            if (end.value.name !== envName) {
+            if (!end) {
+                throw new ParseError("failed to parse function after \\end");
+            } else if (end.value.name !== envName) {
                 throw new ParseError(
                     "Mismatch: \\begin{" + envName + "} matched " +
                     "by \\end{" + end.value.name + "}",
                     endNameToken);
             }
-            result.position = end.position;
```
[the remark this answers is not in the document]
kevinbarabash: That's correct.

src/Parser.js
```diff
@@ -62,46 +68,52 @@ type ParsedDollar = {|
...
 |};
 type ParsedFuncOrArgOrDollar = ParsedFunc | ParsedArg | ParsedDollar;
-*/
+type ParsedFuncOrArg = ParsedFunc | ParsedArg;
 
 /**
  * @param {ParseNode} result
  * @param {Token} token
  * @return {ParsedArg}
  */
```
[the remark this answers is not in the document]
kevinbarabash: Requiring return type annotations should be a separate PR as lots of existing code is missing the types.

src/Parser.js
```diff
@@ -481,17 +493,38 @@ export default class Parser {
...
-    parseImplicitGroup(breakOnTokenText) {
+    parseImplicitGroup(breakOnTokenText?: "]" | "}" | "$"): ?ParseNode {
         const start = this.parseSymbol();
 
         if (start == null) {
             // If we didn't get anything we handle, fall back to parseFunction
             return this.parseFunction();
+        } else if (start.type === "arg") {
+            // Defer to parseGivenFunction if it's not a function we handle
+            return this.parseGivenFunction(start);
```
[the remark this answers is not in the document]
kevinbarabash: `flow` was complaining about `func.length` lower down in this method.  Although we check if `func` is contained in an array of strings using `utils.contains`, `flow` still thinks that `func`, which is `start.result`, could be a `ParseNode`.

src/Parser.js
kevinbarabash: That might of have fixed it, but this also brings `environments` in line with `functions`.

src/Parser.js
```diff
@@ -252,7 +264,7 @@ export default class Parser {
             }
         }
 
-        if (overIndex !== -1) {
+        if (overIndex !== -1 && funcName) {
```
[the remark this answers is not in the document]
kevinbarabash: `overIndex` is used below to `slice` the `body` into `numerBody` and `denomBody`.

src/Parser.js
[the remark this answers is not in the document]
marcianx: Sorry, wasn't clear. I didn't intend for it to be incorporated into this PR. ;)

src/Parser.js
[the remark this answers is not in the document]
marcianx: Whoops, not sure how I missed that. ;)

PR #1044, opened 2017-12-28
Update testing infrastructure NPM packages

edemaine:
Travis is currently failing all tests.  I believe it's an issue with lint, as documented in https://github.com/eslint/eslint/issues/9767 .  I took the opportunity to update all lint/flow/jest related tools.

We should consider updating other NPM packages, e.g. browserify/babelify before release.  (See `npm outdated`.)


kevinbarabash:
@edemaine thanks for updating.  Can you run `npm install` and add the changes to the `package-lock.json` file?  If you don't have this file you may have to update your version of `npm`.

edemaine:
@kevinbarabash Added. I don't usually use NPM 5; I'll have to start. Luckily, Travis does, which will avoid this "surprise failing" issue in the future.

PR #1047, opened 2017-12-29
extract handling of implicit leftright groups

kevinbarabash:
This is a first step in moving `parseImplicitGroup` into the handlers of functions themselves.


src/functions/delimsizing.js
```diff
@@ -128,12 +128,30 @@ defineFunction({
...
+                type: "leftright",
+                body: body,
+                left: delim.value,
+                right: right.value.value,
+            };
+        } else {
+            return {
+                type: "leftright",
+                value: delim.value,
+            };
```
kevinbarabash: This is a little weird.  We return this object which gets turned into a `ParseNode` which gets returned by `const right = parser.parseFunction();` up above.  I should add a comment here to explain this.

kevinbarabash:
There's some shell script snippets in #871 that may help when reviewing this diff.

src/functions/delimsizing.js
```diff
@@ -128,12 +128,30 @@ defineFunction({
...
+            return {
+                type: "leftright",
+                body: body,
+                left: delim.value,
+                right: right.value.value,
+            };
+        } else {
+            return {
+                type: "leftright",
+                value: delim.value,
```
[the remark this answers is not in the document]
kevinbarabash: I would like to make the parse tree better, eventually.  For now though, I'm just trying to break up the monoliths that are functions.js, buildHTML.js, and buildMathML.js.

src/functions/delimsizing.js
kevinbarabash: Good idea.  I'll add a TODO for that.

src/functions/delimsizing.js
```diff
@@ -4,12 +4,12 @@ import defineFunction from "../defineFunction";
 import delimiter from "../delimiter";
 import mathMLTree from "../mathMLTree";
 import ParseError from "../ParseError";
+import ParseNode from "../ParseNode";
```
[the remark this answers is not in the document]
kevinbarabash: Good call.  Not sure why I change that.

src/functions/delimsizing.js
kevinbarabash: Looking at a bit more, splitting into `left` and `right` handlers would be tricky b/c the left/right delimiters are matched and the contents in between is assigned an `minner` class.  Ensuring the matching and setting the correct class on the contents would probably be more difficult if they were separate handlers.

kevinbarabash:
> Also, I see that there are lots of \lefts and \rights in the tests, which increases my confidence that this isn't breaking anything.

Yeah, the test suite is pretty good. ^_^

PR #1048, opened 2017-12-29
extract accent and accentunder into their own files

kevinbarabash:


edemaine:
Great! When I was working on this code for #1033, I felt it would be nicely separated. Leaving to @marcianx  to review with his fancy diff tool (though I should learn to do it too).

kevinbarabash:
There's some shell script snippets in #871 that may help when reviewing this diff.

src/functions/accentunder.js
```diff
@@ -0,0 +1,60 @@
...
+            label: context.funcName,
+            base: base,
+        };
+    },
+    htmlBuilder: (group, options) => {
+        // Treat under accents much like underlines.
+        const innerGroup = html.buildGroup(group.value.base, options);
+
+        const accentBody = stretchy.svgSpan(group, options);
+        const kern = (/tilde/.test(group.value.label) ? 0.12 : 0);
```
davidflanagan: nit: are there functions other than \\utilde that would match this and we just haven't implemented them yet? If not, this seems like a silly place to use a RegExp. Maybe change to:

```
const kern = (group.value.label === '\\utilde') ? 0.12 : 0;
```

src/functions/accent.js
```diff
@@ -0,0 +1,246 @@
...
+    props: {
+        numArgs: 1,
+    },
+    handler: (context, args) => {
+        const base = args[0];
+
+        const isStretchy = !utils.contains([
+            "\\acute", "\\grave", "\\ddot", "\\tilde", "\\bar", "\\breve",
+            "\\check", "\\hat", "\\vec", "\\dot",
+        ], context.funcName);
```
davidflanagan: For performance, consider moving the array of stretchy function names outside the function so it doesn't get re-created on each invocation. 

According to https://jsperf.com/static-arrays-inside-functions/1 moving the array out of the function will make this operation twice as fast.

src/functions/accent.js
```diff
@@ -0,0 +1,246 @@
...
+
+        const isShifty = !isStretchy || utils.contains([
+            "\\widehat", "\\widetilde",
```
davidflanagan: Unless this set of two functions is likely to grow, consider just doing a normal funcName === '\\widehat' || funcName === '\\widetilde' comparison

src/functions/accent.js
davidflanagan: A regexp (outside the function) seems like it would also be a good choice here instead of an array of strings.

src/buildCommon.js
```diff
@@ -318,8 +318,9 @@ const makeSpan = function(
     classes?: string[],
     children?: DomChildNode[],
     options?: Options,
+    style?: {[name: string]: string},
```
davidflanagan: Is it worth defining a flow type for this string-to-string map?

davidflanagan: Also, you should probably update the comment to mention this argument (and maybe also explain when to use styles instead of classes?)

src/functions/accent.js
```diff
@@ -127,19 +126,19 @@ const htmlBuilder = (group, options) => {
...
+                        ? {
+                            width: `calc(100% - ${2 * skew}em)`,
+                            marginLeft: `${(2 * skew)}em`,
+                        }
+                        : undefined,
+                },
             ],
         }, options);
-
-        // TODO: pass "svg-align" to stretchy.svgSpan
```
davidflanagan: This TODO seems to have gone away. Just want to make sure that was intentional. Similarly in accentunder.js

src/domTree.js
```diff
@@ -68,13 +68,14 @@ class span implements CombinableDomNode {
...
         options?: Options,
+        style?: {[string]: string},
     ) {
         this.classes = classes || [];
         this.children = children || [];
         this.height = 0;
         this.depth = 0;
         this.maxFontSize = 0;
-        this.style = {};
+        this.style = style || {};
```
davidflanagan: `this.style` is mutated below, so it seems dangerous to use the passed in style object without copying it-it probably works now, but could be a source of difficult bugs the the calling code changes to reuse the style object. 

Does this project have babel set up so that you can use `{...style}` here?

The rest of my comments are just nits, but for this one I'm actually going to request changes...

src/functions.js
```diff
@@ -222,55 +221,7 @@ defineFunction([
...
-
-    return {
-        type: "accent",
-        label: context.funcName,
-        isStretchy: false,
-        isShifty: true,
-        base: base,
-    };
-});
+import "./functions/accent";
```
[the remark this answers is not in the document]
kevinbarabash: This is good feedback.  The position of the `import` statements was b/c I wanted to show what was being replace, but that's not a very good reason.  Eventually there should only be `import` statements so things will end up at the top, but right now it looks kind of funky.

Relying on side-effects is not ideal.  I really like your suggestion of having the modules in the functions folder export objects and then this file can load those definitions.

src/functions/accentunder.js
kevinbarabash: Good catch.

src/buildCommon.js
kevinbarabash: I think I'm to change this to `cssStyle` to differentiate from LaTeX's concept of styles.  I'll add a comment, but basically this should only be used by dynamic styles which we can't define a class for.

src/functions/accent.js
```diff
@@ -0,0 +1,246 @@
...
+        let accent;
+        let width: number;
+        if (group.value.label === "\\vec") {
+            // Before version 0.9, \vec used the combining font glyph U+20D7.
+            // But browsers, especially Safari, are not consistent in how they
+            // render combining characters when not preceded by a character.
+            // So now we use an SVG.
+            // If Safari reforms, we should consider reverting to the glyph.
+            accent = buildCommon.staticSvg("vec", options);
+            width = parseFloat(accent.style.width);
```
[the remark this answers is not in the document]
kevinbarabash: Good call with `buildCommon.svgData['vec'][1]`, I'm going to do that.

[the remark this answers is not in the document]
kevinbarabash: Unfortunately we can't know the width of spans without knowing the metrics for every character which we don't at the moment.

src/domTree.js
kevinbarabash: Whoops, missed the one comment I actually need to address.  Fixing now. :)

kevinbarabash:
@davidflanagan this is ready for review again.

kevinbarabash:
Okay, now it's ready.  I forgot to push that last commit.

src/domTree.js
```diff
@@ -49,6 +49,8 @@ export type DomChildNode = span | anchor | svgNode | symbolNode;
 
 export type SvgChildNode = pathNode | lineNode;
 
+export type CssStyle = {[name: string]: string};
```
kevinbarabash: I'll probably add an `Attributes` type in a future PR.

src/domTree.js
```diff
@@ -61,20 +63,21 @@ class span implements CombinableDomNode {
...
         options?: Options,
+        style?: CssStyle,
     ) {
         this.classes = classes || [];
         this.children = children || [];
         this.height = 0;
         this.depth = 0;
         this.maxFontSize = 0;
-        this.style = {};
+        this.style = Object.assign({}, style);
```
kevinbarabash: I double checked the build output and the polyfill for this is included.  I think this will be smaller than object spread, but not sure.

src/functions/accent.js
```diff
@@ -180,6 +180,11 @@ const mathmlBuilder = (group, options) => {
     return node;
 };
 
+const NON_STRETCHY_ACCENT_REGEX = new RegExp([
+    "\\acute", "\\grave", "\\ddot", "\\tilde", "\\bar", "\\breve",
+    "\\check", "\\hat", "\\vec", "\\dot",
+].map(accent => `\\${accent}`).join("|"));
```
davidflanagan: I would have forgotten to double the backslashes here. Glad you go this right!

src/domTree.js
davidflanagan: TIL: that the second argument to Object.assign() can be null or undefined. This is great!

src/functions/accent.js
edemaine: Is there a reason to use a regex here instead of an object like
```js
nonstretchyAccent = {"\\acute": true, ...}
...
if (context.funcName in nonstretchyAccent)
```

In general, KaTeX should do this sort of thing in place of most/all `utils.contains`.  Faster.  But that's another issue.

In this instance, another option would be to have a mapping defining properties like `stretchy` for each accent type... could go either way. But RegExp seems slightly unnecessary (though also fine).

PR #1054, opened 2017-12-31
add perf tests

### LINKED ISSUE #1053: Write a performance test harness and test suite
We should have some way of measuring changes in performance in KaTeX.  To start I think it's enough to test the performance of `renderToString()` with difference TeX strings, maybe the ones in `ss_data.yaml`.  https://benchmarkjs.com looks like a good tool to use for this purpose so that the results are somewhat consistent (at least on the same machine).

kevinbarabash:
Here's the output I get when I run `npm run perf-test`:
```
compiling katex...

AccentsText x 2,416 ops/sec ±4.79% (80 runs sampled)
ArrayMode x 344 ops/sec ±1.35% (85 runs sampled)
GroupMacros x 7,374 ops/sec ±5.01% (82 runs sampled)
MathBb x 1,976 ops/sec ±1.10% (84 runs sampled)
SqrtRoot x 1,424 ops/sec ±0.95% (86 runs sampled)
StretchyAccent x 370 ops/sec ±1.64% (83 runs sampled)
Units x 298 ops/sec ±3.04% (78 runs sampled)
Fastest is GroupMacros
(khan27)kb-no-mbp:KaTeX kevinbarbash$ git st
On branch perf_test
```
This fixes #1053.

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,6 +24,7 @@
     "babel-preset-es2015": "^6.18.0",
     "babel-preset-flow": "^6.23.0",
     "babel-register": "^6.26.0",
+    "benchmark": "^2.1.4",
     "check-dependencies": "^1.1.0",
     "css-loader": "^0.28.8",
     "eslint": "^4.14.0",
@@ -33,7 +34,7 @@
     "flow-bin": "^0.62.0",
     "jest": "^22.0.4",
     "jest-serializer-html": "^4.0.1",
-    "js-yaml": "^3.3.1",
+    "js-yaml": "^3.10.0",
     "json-stable-stringify": "^1.0.1",
     "jspngopt": "^0.2.0",
     "less": "~2.7.1",
@@ -66,6 +67,7 @@
     "start": "check-dependencies && webpack-dev-server --open --config webpack.dev.js",
     "build": "check-dependencies && rimraf build/* && webpack",
     "watch": "npm run build -- --watch",
+    "perf-test": "check-dependencies && NODE_ENV=test node test/perf-test.js",
     "prepublishOnly": "make NIS= dist"
   },
   "pre-commit": [
--- a/test/perf-test.js
+++ b/test/perf-test.js
@@ -1,0 +1,59 @@
+/* eslint-disable no-console */
+/**
+ * Performance Tests
+ *
+ * This file measures the performance of renderToString using a number of strings
+ * from ss_data.yaml.
+ *
+ * TODO:
+ * - allow users to specify a different key or keys from ss_data.yaml
+ * - allow users to specify a different string or strings
+ * - provide a way to test the performance against different branches
+ */
+require('babel-register');
+const Benchmark = require('benchmark');
+const yaml = require('js-yaml');
+const fs = require('fs');
+const path = require('path');
+
+const filename = path.resolve(__dirname, 'screenshotter/ss_data.yaml');
+const data = yaml.load(fs.readFileSync(filename, 'utf-8'));
+
+console.log('compiling katex...');
+const katex = require('../katex').default;
+console.log('');
+
+// Benchmark is a performancing testing library.  It allows you to define a
+// suite of tests.  After adding tests to the suite with the .add() method they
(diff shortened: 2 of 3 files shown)


kevinbarabash:
I picked tests which I thought were interesting in some way, e.g. use of macros, stretchy accents, environments, etc.  Each test takes a few seconds to run and I wasn't able to find a way to reduce the number of runs sampled so I limited the number of strings we test.

package.json
```diff
@@ -35,7 +36,7 @@
     "glob": "^7.1.1",
     "jest": "^22.0.4",
     "jest-serializer-html": "^4.0.1",
-    "js-yaml": "^3.3.1",
+    "js-yaml": "^3.10.0",
```
kevinbarabash: I didn't realize we were already using `js-yaml`.  I ran `npm install js-yaml --save-dev` and it must've upgraded it.

test/perf-test.js
```diff
@@ -0,0 +1,53 @@
+/* eslint-disable no-console */
```
kevinbarabash: TODO: top level docstring explaining the purpose of the file and how to run the tests.

test/perf-test.js
```diff
@@ -0,0 +1,53 @@
...
+for (const key of testsToRun) {
+    const value = data[key];
+    if (typeof value === "string") {
+        suite.add(key, () => katex.renderToString(value));
+    } else {
+        if (value.macros) {
+            const options = {
+                macros: value.macros,
+            };
+            suite.add(key, () => katex.renderToString(value.tex, options));
```
davidflanagan: This line and the three lines above might work in both the if and else cases... Maybe you can omit the nested if/else

test/perf-test.js
davidflanagan: Don't forget!

test/perf-test.js
```diff
@@ -0,0 +1,53 @@
...
+        } else {
+            suite.add(key, () => katex.renderToString(value.tex));
+        }
+    }
+}
+
+suite.on('cycle', function(event) {
+    console.log(String(event.target));
```
davidflanagan: Is this the line that prints out the ops/sec for each test? That is such a non-intuitive event name that I think a comment is warrented here.

test/perf-test.js
```diff
@@ -0,0 +1,53 @@
...
+});
+
+suite.on('complete', function() {
+    console.log('Fastest is ' + this.filter('fastest').map('name'));
```
davidflanagan: This doesn't seem like useful output. I'd cut it.

test/perf-test.js
```diff
@@ -0,0 +1,53 @@
...
+});
+
+suite.run({
+    async: true,
```
davidflanagan: What does async do? None of the tests are async, so is this needed? (I guess I'm wondering if this affects the stability of timing or anything).

test/perf-test.js
```diff
@@ -0,0 +1,53 @@
...
+const path = require('path');
+
+const filename = path.resolve(__dirname, 'screenshotter/ss_data.yaml');
+const data = yaml.load(fs.readFileSync(filename, 'utf-8'));
+
+console.log('compiling katex...');
+const katex = require('../katex');
+console.log('');
+
+const suite = new Benchmark.Suite;
```
davidflanagan: Nits: This line seems like it is worth a comment for readers who don't already know about this benchmark package. I skimmed over it the first time and then had to come back to it. Also, I think that the variable name 'tests' would make the code below a little clearer.

kevinbarabash:
> I've already got feature requests, however:

I've added these as TODOs in perf-test.js and will created GH issues for them.

test/perf-test.js
kevinbarabash: I copied it from the benchmark.js example.  I removed and the tests still pass.  It shouldn't make a difference either way b/c all of our code is synchronous.

kevinbarabash:
@davidflanagan thanks for the quick review.  I'm happy this in on `master` now.  It'll make understanding the perf impact of future changes much easier.

PR #1057, opened 2018-01-04
Browserify hotfix

nickkolok:
Greetings from Russia, comrads!

`0.8.0` and higher are incompatible with `browserify`. `0.7.1` is compatible.
The problem is described in detail [here](https://github.com/browserify/browserify/issues/374).

I suggest this solution as a hotfix. Real solution should use `standalone` option as described in a link above.

I believe that this PR should be merged into `0.8.x` ASAP, because otherwise upgrading from `0.7.x` to `0.8.x` breaks any application which uses `katex` through `browserify`.


kevinbarabash:
@nickkolok thanks for the PR.  This should be pretty easy to back port to 0.8.x.  I'll have to set up a 0.8.x branch to do the deploy from.  I think I'll merge it into master and then back port it afterwards.

nickkolok:
@kevinbarabash Can I do something to fastenate integration of this PR into `0.8.x` ? I need `browserify`-compatible version of `katex` in production, and I'd definitely prefer `0.8.x` to `0.7.x`.

kevinbarabash:
Still need to create a 0.8.x branch and merge this in there.

nickkolok:
@kevinbarabash I've backported this patch on `v0.8.3` (patch + version bump to `v0.8.4`): https://github.com/nickkolok/KaTeX/tree/browserify-hotfix-0.8.4
Hope this helps.

PR #1060, opened 2018-01-12
Add basic support for Indic scripts in addition to CJK.

davidflanagan:
This patch just makes KaTeX recognize Unicode codepoints in the
range \u0900-\u109f so that those South and Southeast Asian scripts
do not get automatically rejected.

The patch also generalizes the way that Unicode blocks are handled
to make it easier to add support for new scripts in the future.
src/unicodeRegexes.js is replaced with the new file src/unicodeScripts.js


davidflanagan:
What do you think, Kevin?

.gitignore
```diff
@@ -13,3 +13,4 @@ diff.png
 /test/symgroups.pdf
 /test/screenshotter/unicode-fonts
 coverage/
+*~
```
davidflanagan: This is for emacs backup files.

src/fontMetrics.js
```diff
@@ -198,10 +198,15 @@ const getCharacterMetrics = function(
     let ch = character.charCodeAt(0);
     if (character[0] in extraCharacterMap) {
         ch = extraCharacterMap[character[0]].charCodeAt(0);
-    } else if (cjkRegex.test(character[0])) {
+    } else if (supportedCodepoint(ch) && !metricMap[font][ch]) {
```
davidflanagan: Note the `&& !metricMap` bit I added on this line... I was thinking that this allows the addition of other font metrics in the future. Not sure whether that makes sense, though.

src/unicodeScripts.js
```diff
@@ -0,0 +1,58 @@
...
+                return script;
+            }
+        }
+    }
+    return null;
+}
+
+export function supportedCodepoint(codepoint: number): boolean {
+    return scriptFromCodepoint(codepoint) !== null;
+}
```
davidflanagan: I could make this faster by statically concatening all of the ranges for all scripts into a single array. Then there wouldn't be a nested loop. I'm assuming that this has similar performance to using a regex. But could also compile the ranges into a regexp at startup time and use that here on a character instead of a codepoint if performance is a concern.

src/fontMetrics.js
davidflanagan: Actually, I think I need to think some more about this.  In Devangari (and probably most of these other Brahmic scripts) there are a lot of combining characters which generally turn into ligatures and have no intrinsic width on their own.  Using a full em width is wrong given that somewhere around half of the characters are zero width vowels.

At a minimum I should choose a narrower character for Brahmic scripts. Really though, I think we can't really measure characters from those scripts in isolation: to actually be accurate we have to measure the whole string of them. I haven't looked yet at how this code is used, but I worry that this is a larger architectural change required here for complete accuracy.

For our needs at KA we only need to be able to display these characters in plain \text{} environments. Not trying to center them above a big capital sigma character or anything, so I'd guess that super-accurate measurements are not all that necessary.

kevinbarabash:
Not sure why the Unicode screenshot test is failing.  It renders using the test page.  I'm going to run `make screenshots` locally and try to figure out what the change was.

kevinbarabash:
The reason why the tests are failing is because the browser in the docker isn't loading a font with the glyphs for the Hangul and CJK parts of the Unicode screenshot test.

unicode-firefox

This is because the spans containing the Hangul and CJK text no long have `hangul_fallback` and `cjk_fallback` as part of their `class`.  This class is used in test.html to load appropriate fonts.

src/domTree.js
```diff
@@ -408,19 +408,16 @@ class symbolNode implements CombinableDomNode {
...
+        // Mark text from non-Latin scripts with specific classes so that we
+        // can specify which fonts to use.  This allows us to render these
+        // characters with a serif font in situations where the browser would
+        // either default to a sans serif or render a placeholder character.
+        // We use CSS class names like cjk_fallback, hangul_fallback and
+        // indic_fallback. See ./unicodeScripts.js for the set of possible
+        // script names
+        const script = scriptFromCodepoint(this.value.charCodeAt[0]);
+        if (script) {
+            this.classes.push(script + "_fallback");
```
kevinbarabash: For some reason this isn't getting applied as expected.

src/domTree.js
```diff
@@ -408,19 +408,16 @@ class symbolNode implements CombinableDomNode {
...
-                this.classes.push('cjk_fallback');
-            }
+        // Mark text from non-Latin scripts with specific classes so that we
+        // can specify which fonts to use.  This allows us to render these
+        // characters with a serif font in situations where the browser would
+        // either default to a sans serif or render a placeholder character.
+        // We use CSS class names like cjk_fallback, hangul_fallback and
+        // indic_fallback. See ./unicodeScripts.js for the set of possible
+        // script names
+        const script = scriptFromCodepoint(this.value.charCodeAt[0]);
```
kevinbarabash: I can't believe `flow` didn't pick this up.

kevinbarabash: I was worried that our collating didn't respect script boundaries, but it does.  I tested `\text{私はバナナです여보세요}` and it does the right thing by rendering two spans, one for CJK and one for Hangul.

src/fontMetrics.js
kevinbarabash: The only place we actually use the `width` of any glyph is when we're positioning accent glyphs.  We still need some sort of `height`/`depth` for these so what you've done here I think is okay for now.

src/unicodeScripts.js
```diff
@@ -0,0 +1,58 @@
...
+        0x3000, 0x30FF,  // CJK symbols and punctuation, Hiragana, Katakana
+        0x4E00, 0x9FAF,  // CJK ideograms
+        0xFF00, 0xFF60,  // Fullwidth punctuation
+        // TODO: add halfwidth Katakana and Romanji glyphs
+    ],
+
+    // Korean
+    hangul: [0xAC00, 0xD7AF],
+
+    // The Brahmic scripts of South and Southeast Asia
```
kevinbarabash: So many different scripts... cool!

src/unicodeScripts.js
```diff
@@ -0,0 +1,58 @@
...
+ * of a block of codepoints that are part of the script.
+ */
+const scriptData = {
+    // Chinese and Japanese.
+    // The "k" in cjk is for Korean, but we've separated Korean out
+    cjk: [
+        0x3000, 0x30FF,  // CJK symbols and punctuation, Hiragana, Katakana
+        0x4E00, 0x9FAF,  // CJK ideograms
+        0xFF00, 0xFF60,  // Fullwidth punctuation
+        // TODO: add halfwidth Katakana and Romanji glyphs
```
kevinbarabash: Maybe change these to:
```
cjk: [
    [0x3000, 0x30FF],
    [0x4E00, 0x9FAF],
    [0xFF00, 0xFF60],
],
brahmic: [ 
    [0x0900, 0x109F],
],
```

kevinbarabash: It simplifies the logic in `scriptFromCodepoint` and makes it more obvious what's going on when there are multiple ranges.

src/unicodeScripts.js
kevinbarabash: I think it's fine.  Although perf is important we aren't measuring it yet.  Once we start measuring it, and if we notice a drop in perf, then we can start optimizing things.  If you have time maybe have a look at https://github.com/Khan/KaTeX/pull/1054.

.gitignore
kevinbarabash: Should probably add this to your global gitignore

test/unicode-spec.js
```diff
@@ -101,4 +102,46 @@ describe("unicode", function() {
         expect('私はバナナです。').toNotParse();
         expect('여보세요').toNotParse();
     });
+
+    it("should parse Devangari inside \\text{}", function() {
+        expect('\\text{नमस्ते}').toParse();
+    });
```
kevinbarabash: Thanks for adding tests.

src/unicodeScripts.js
davidflanagan: I ended up changing this all around to make Flow happy and to speed up perf. (I was making those changes before you started the review... sorry about that.) But now that I've made codepointSupported() fast, I can probably make this change to make the data structure clearer...

davidflanagan:
@kevinbarabash: sorry for asking for a review and then continuing to work on this... I believe I've addressed all of your requests. The new unicodeScripts.js file has changed substantially since you reviewed it, so you might want to take another look.

src/fontMetrics.js
```diff
@@ -198,10 +198,21 @@ const getCharacterMetrics = function(
...
+    if (!metrics) {
+        // We don't typically have font metrics for Asian scripts.
+        // So if the character is in a script we support but we
+        // dont have metrics for it, just use the metrics for
+        // the Latin capital letter M. This is close enough because
+        // we (currently) only care about the height of the glpyh
+        // not its width.
+        if (supportedCodepoint(ch)) {
+            metrics = metricMap[font][77]; // 77 is the charcode for 'M'
+        }
```
kevinbarabash: I see, we fall back to the metrics for `'M'` for any code point.  Good enough for now.

test/unicode-spec.js
```diff
@@ -101,4 +102,46 @@ describe("unicode", function() {
...
+    const hangulRE = /[\uAC00-\uD7AF]/;
+    const brahmicRE = /[\u0900-\u109F]/;
+    const allRE =
+        /[\u3000-\u30FF\u4E00-\u9FAF\uFF00-\uFF60\uAC00-\uD7AF\u0900-\u109F]/;
+
+    it("supportedCodepoint() should return the correct values", () => {
+        for (let codepoint = 0; codepoint <= 0xffff; codepoint++) {
+            expect(supportedCodepoint(codepoint)).toBe(
+                allRE.test(String.fromCharCode(codepoint))
+            );
```
kevinbarabash: wow

test/unicode-spec.js
```diff
@@ -101,4 +102,46 @@ describe("unicode", function() {
...
+        }
+    });
+
+    it("scriptFromCodepoint() should return correct values", () => {
+        for (let codepoint = 0; codepoint <= 0xffff; codepoint++) {
```
kevinbarabash: hopefully these tests don't take that long to run.

kevinbarabash: I had a look and unicode-spec.js take 7s vs 9s for katex-spec.js so I think we're okay.

src/fontMetrics.js
ronkok: This is really cool and yet I have a question. Many, perhaps most, of these characters are not in Times New Roman, so we don't even know which font will display. An yet at least some of these characters are taller than an "M".
height1

Is there a widely-used system font that contains all these characters?  Then we could specify it in KaTeX CSS and get some font metrics, at least `height` and `depth`.

PR #1061, opened 2018-01-12
Add section, paragraph and copyright as symbol

ysawa:
I have known some symbols in TeX do not exist in KaTeX. I have added

* Section mark
* Paragraph mark
* Copyright mark

to the symbols list. Would you be able to check them?


ysawa:
[clabot:check]

kevinbarabash:
@ysawa thanks for the PR.  I don't think these symbols are actually in any of our fonts.  Do you know what TeX font (or package) they appear in?

ysawa:
@kevinbarabash Those symbols may be LaTeX fonts 
 ( https://tex.stackexchange.com/questions/255127/why-is-the-copyright-symbol-in-helvet-asymmetric ) . Actually, they are not in KaTeX fonts, and I think it's better to use those text as normal HTML. Thank you.

edemaine:
I've created https://github.com/KaTeX/katex-fonts/issues/9 to add these to the fonts.

PR #1067, opened 2018-01-15
Add katex-ruby to the libraries list

glebm:


glebm:
I can't sign the CLA, at least not without a possibly lengthy approval process from Google that is not guaranteed to succeed.

glebm:
It has been brought to my attention that Google already has a CLA on file for Khan Academy!
I can't sign an individual CLA but it should not be necessary.

kevinbarabash:
@glebm I didn't realize you were from Google.  Google has already signed our CLA.

Issue #929, opened 2017-10-10
webpackDevServer does not reload CSS or main.js


edemaine:
When I change `static/katex.css`, the webpack server doesn't notice the change.  I have to manually kill and restart the server to get the CSS to update.  This makes it difficult to develop CSS changes.

The same is true if I change `static/main.js`.

rrandallcainc:
This shouldn't be too hard to integrate for the css. 

Is static/main.js really something that is changed often? Does it need to be watched?

edemaine:
@rrandallcainc FWIW, I was actively development `static/main.js` the other day (for testing of #899). I've seen it change at least twice in the past 6 months, so it does happen, and would be useful.  There is close to zero cost for watching, right?

rrandallcainc:
Nope, was just curious if it get changed frequently :)

PR #1068, opened 2018-01-16
Use webpack to build files and webpack-dev-server for testing

### LINKED ISSUE #692: Suggestion: don't force global katex for auto-render to work
At the moment you have to declare katex globally for auto-render to work. The easiest way not to force this would be to add katex as a property of the renderMathInElement options which could be used by the user to supply an instance of katex, although this is perhaps slightly cumbersome. Alternatively you could do a conditional require if require is defined and katex is not defined globally, using a relative path so katex would be pulled in automatically assuming the directory structure is correct (and make sure katex isn't bundled during the build).

### LINKED ISSUE #983: Make contrib/auto-render importable
It appears Webpack >= 2.2.0 [does not support](https://github.com/webpack/webpack/issues/4039) `import` mixed with `module.exports = ...` syntax. Perhaps I have misread that thread or am mistaken? I'm unable to figure out a way to import `renderMathInElement` without getting `TypeError: "exports" is read-only`.

### LINKED ISSUE #1063: switch from browserify to webpack for building distribution files
Reasons:
- we already have a webpack.config.js with settings for the unminified distribution files
- webpack has cool plugins we can use to better optimize our code, e.g. https://webpack.js.org/plugins/module-concatenation-plugin/ to dedupe helper code that babel inserts, as discussed in #872.

ylemkimon:
Fixes #1063 
Fixes partially #929 (webpack-dev-server reloads when `katex.less` is changed)
Fixes #692
Fixes #983

Old | New
--- | ---
`browserify`(`make *.js`) | `webpack`
browserify transform `babelify` | webpack `babel-loader`
`UglifyJS` CLI(`make *.min.js`) | webpack `UglifyJsPlugin`
`Less` CLI, copy .css(`make *.css`) | webpack `less-loader`, `css-loader`, and `ExtractTextPlugin`
`clean-css` CLI(`make *.min.css`) | `cssnano` included in webpack `css-loader`
fonts(`make build/fonts`) | webpack `file-loader`
contrib(`make contrib` and Makefiles for each contrib) | webpack entry points in `webpack.common.js`[1]
`make build/katex.js build/katex.min.js build/katex.css build/katex.min.css build/fonts build/contrib` | Depends on `make` target `webpack`, which does `npm run build`(`webpack`)
`make` recompile | `webpack --watch`(`npm run watch`)
Local testing using `server.js`(Removed in #902), `webpack-dev-server`(`webpackDevServer.js`) | Updated `webpack-dev-server` configuration(renamed to `webpack.dev.js`), updated documentations
Build stylesheets before starting dev server(npm script `prestart`) | Load(inline) stylesheets using webpack `style-loader`[2]
Screenshotter using `server.js` | Use `webpack-dev-server`
Export CommonJS module | Export ES6 module[3], expose default export

### Possible Breaking Changes:

Ref | Changes | Impact | Migration Path
--- | ------- | ------ | --------------
[1] | Removed make target `contrib` | `make contrib` | `make build/contrib` or `npm run build`(`webpack`)
[2] | Removed npm script `build-css` | `npm run build-css` | `make [.css filename]` or `npm run build`(`webpack`)
[3] | Switched to ES6 module export |  Directly requiring `./katex.js`<br>(`require('./katex.js)`) | Use `default` property<br>(`require('./katex.js).default`)

I'm not sure there are use cases of above cases.


webpackDevServer.js
```diff
@@ -1,15 +1,19 @@
-/* eslint no-console:0 */
-
-const WebpackDevServer = require("webpack-dev-server");
-const webpack = require("webpack");
-const webpackConfig = require("./webpack.config.js");
+// @flow
+const { targets, createConfig } = require('./webpack.common');
+const path = require('path');
 const PORT = 7936;
```
ylemkimon: Is there any specific reason using the port 7936?

contrib/mathtex-script-type/Makefile
```diff
@@ -1,9 +0,0 @@
-.PHONY: build integrity
-
-build: $(BUILDDIR)/contrib/mathtex-script-type.min.js
-
-$(BUILDDIR)/contrib/mathtex-script-type.min.js: $(BUILDDIR)/mathtex-script-type.js
-	$(UGLIFYJS) < $< > $@
-
-$(BUILDDIR)/mathtex-script-type.js: mathtex-script-type.js
```
ylemkimon: Is outputting `mathtex-script-type.js` to `$(BUILDDIR)`, not `$(BUILDDIR)/contrib`, and not including in the distribution archive an intended behavior?

contrib/copy-tex/Makefile
```diff
@@ -1,15 +0,0 @@
-.PHONY: build
-
-build: $(BUILDDIR)/contrib/copy-tex.min.js $(BUILDDIR)/contrib/copy-tex.css $(BUILDDIR)/contrib/copy-tex.min.css
-
-$(BUILDDIR)/contrib/copy-tex.min.js: $(BUILDDIR)/contrib/copy-tex.js
-	$(UGLIFYJS) < $< > $@
-
-$(BUILDDIR)/contrib/copy-tex.js: copy-tex.js
-	$(BROWSERIFY) -t [ babelify ] $< --standalone renderMathInElement > $@
```
ylemkimon: Is exporting the module to `renderMathInElement` an intended behavior?

webpackDevServer.js
kevinbarabash: We wanted to avoid a collision with other services that are usually running in the normal dev environment at Khan Academy.

contrib/mathtex-script-type/Makefile
kevinbarabash: Good catch.

contrib/copy-tex/Makefile
kevinbarabash: Thanks for catching this.  I guess nobody's been using this as a standalone script.

webpack.common.js
```diff
@@ -0,0 +1,123 @@
...
+            }),
+        ],
+        devtool: dev && 'inline-source-map',
+    };
+
+    if (minimize) {
+        config.plugins.push(new UglifyJsPlugin({
+            uglifyOptions: {
+                output: {
+                    ascii_only: true,
```
kevinbarabash: The Makefile used `--mangle --beautify ascii_only=true,beautify=false`.  Does `UglifyJsPlugin` support the same options?

webpack.common.js
```diff
@@ -0,0 +1,123 @@
...
+        plugins: [
+            new webpack.EnvironmentPlugin({
+                NODE_ENV: dev ? 'development' : 'production',
+            }),
+            new ExtractTextPlugin({
+                filename: minimize ? '[name].min.css' : '[name].css',
+                disable: dev,
+            }),
+        ],
+        devtool: dev && 'inline-source-map',
```
kevinbarabash: Cool

webpack.common.js
```diff
@@ -0,0 +1,123 @@
...
+                            loader: 'less-loader',
+                        }],
+                    }),
+                },
+                {
+                    test: /\.(ttf|woff|woff2)$/,
+                    use: [{
+                        loader: 'file-loader',
+                        options: {
+                            name: 'fonts/[name].[ext]',
```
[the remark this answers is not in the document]
kevinbarabash: I've never handled fonts like this before.  If we didn't do this we'd need a separate way to serve font files when using the dev server correct?

webpack.common.js
```diff
@@ -0,0 +1,123 @@
...
+
+/**
+ * Create a webpack config for given target
+ */
+function createConfig(target /*: Target */, dev /*: boolean */,
+        minimize /*: boolean */) /*: Object */ {
+    const cssLoader = {
+        loader: 'css-loader',
+        options: {
+            minimize, // cssnano
```
kevinbarabash: What's `cssnano`?

webpack.common.js
```diff
@@ -0,0 +1,123 @@
...
+        library: 'katex',
+    },
+    {
+        name: 'contrib/auto-render',
+        entry: './contrib/auto-render/auto-render.js',
+        library: 'renderMathInElement',
+    },
+    {
+        name: 'contrib/copy-tex',
+        entry: './contrib/copy-tex/copy-tex.js',
```
kevinbarabash: I assume you left of the `library` property for this and `mathtex-script-type` since we weren't exporting the right thing in the browserify config which implies nobody was using it.  Is that correct?

webpack.config.js
```diff
@@ -1,61 +1,8 @@
...
-    },
-};
+// @flow
+const { targets, createConfig } = require('./webpack.common');
+
+module.exports = targets.reduce((configs, target) => {
+    configs.push(createConfig(target, false, false));
+    configs.push(createConfig(target, false, true));
+    return configs;
+}, []);
```
kevinbarabash: I have a preference for avoiding `reduce`.  Maybe we could do:
```
module.exports = [
    ...targets.map(target => createConfig(target, false, false)),
    ...targets.map(target => createConfig(target, false, true)),
];
```
It isn't necessary to to interleave the configs is it?

kevinbarabash: It might also be good to have comments here to remind people what the booleans are for.

package.json
```diff
@@ -25,50 +25,47 @@
...
-    "copy": "cp -a static/. build/ && cp -a contrib build/",
     "clean": "rm -rf build/* node_modules/",
     "clean-install": "npm run clean && npm i",
     "test": "check-dependencies && npm run lint && npm run flow && npm run jest",
-    "build-css": "lessc --clean-css static/katex.less build/katex.css",
-    "prestart": "rimraf build && npm run build-css && npm run copy",
-    "start": "check-dependencies && node webpackDevServer.js",
+    "start": "webpack-dev-server --open --config webpackDevServer.js",
+    "build": "webpack",
+    "watch": "webpack --watch",
```
kevinbarabash: I wonder if we should use `--watch` for `start` to maintain the same behavior?  The current server supports hot reloading (except for the CSS which was kind of annoying).

katex.webpack.js
```diff
@@ -0,0 +1,11 @@
+/**
+ * This is the webpack entry point for KaTeX. As flow[1] and jest[2] doesn't support
+ * CSS modules natively, a separate entry point is used and it is not flowtyped.
+ *
+ * [1] https://gist.github.com/lambdahands/d19e0da96285b749f0ef
+ * [2] https://facebook.github.io/jest/docs/en/webpack.html
+ */
+import './static/katex.less';
+import katex from './katex.js';
```
kevinbarabash: Importing `katex.less` is necessary so that we can build a .css file from it based on the `katex` entry point, right?

katex.webpack.js
```diff
@@ -0,0 +1,11 @@
...
+
+module.exports = katex;
```
kevinbarabash: This is so that the build file will be commonjs?

Makefile
```diff
@@ -40,48 +31,25 @@ $(NIS) setup: package.json
...
-build/katex.css: static/katex.less $(wildcard static/*.less) $(NIS)
-	mkdir -p build
-	./node_modules/.bin/lessc $< $@
-
-build/katex.min.css: build/katex.css
+webpack: katex.js $(wildcard src/*.js) $(wildcard static/*.less) $(NIS)
 	mkdir -p build
-	$(CLEANCSS) -o $@ $<
+	mkdir -p build/contrib
+	mkdir -p build/fonts
```
kevinbarabash: Doesn't webpack create the directories it needs if they don't exist?

Makefile
```diff
@@ -40,48 +31,25 @@ $(NIS) setup: package.json
...
+	@# Since everything in build/contrib and build/fonts is put in the built files,
+	@# make sure there's nothing in there we don't want.
+	rm -rf build/contrib/*
+	rm -rf build/fonts/*
```
kevinbarabash: I don't if we should just do `rm -rf build`?

contrib/auto-render/README.md
```diff
@@ -29,9 +29,9 @@ before the close body tag:
 ```
 
 See [index.html](index.html) for an example.
-(To run this example from a clone of the repository, run `make serve`
+(To run this example from a clone of the repository, run `npm start`
 in the root KaTeX directory, and then visit
-http://0.0.0.0:7936/contrib/auto-render/index.html
+http://localhost:7936/contrib/auto-render/index.html
```
kevinbarabash: Thanks for updating the docs.

contrib/auto-render/index.html
```diff
@@ -1,15 +1,14 @@
...
     in the root KaTeX directory and then visit with your web browser:
-    http://0.0.0.0:7936/contrib/auto-render/index.html
+    http://localhost:7936/contrib/auto-render/index.html
   -->
 <html>
   <head>
     <meta charset="UTF-8">
     <title>Auto-render test</title>
     <script src="/katex.js" type="text/javascript"></script>
-    <link href="/katex.css" rel="stylesheet" type="text/css">
```
kevinbarabash: Don't we still need `katex.css`?

contrib/copy-tex/index.html
```diff
@@ -1,17 +1,15 @@
...
-    http://0.0.0.0:7936/contrib/copy-tex/index.html
+    http://localhost:7936/contrib/copy-tex/index.html
   -->
 <html>
   <head>
     <meta charset="UTF-8">
     <title>Copy-tex test</title>
     <script src="/katex.js" type="text/javascript"></script>
-    <link href="/katex.css" rel="stylesheet" type="text/css">
-    <link href="./copy-tex.css" rel="stylesheet" type="text/css">
```
kevinbarabash: Why no .css files?

package.json
kevinbarabash: Sorry, `webpack-dev-server` does hot reloading, everything is fine.

kevinbarabash:
> Fixes partially #929 (webpack-dev-server reloads when katex.less is changed)

Why is it a partial fix?

kevinbarabash:
I restarted the travis-ci build b/c I was really confused by the output of the previous run.  It should be running `npm test`, but I couldn't find any of the usual output that I expect from that command.  I ran `npm test` locally on your branch and it seems to work fine.

ylemkimon:
@kevinbarabash Currently Travis seems to be [not working and backlogged](https://www.traviscistatus.com/).

Makefile
ylemkimon: @kevinbarabash Yes, it does. Will remove them.

webpack.common.js
ylemkimon: Yes. `mangle` is [enabled by default](https://webpack.js.org/plugins/uglifyjs-webpack-plugin/#uglifyoptions) and `beautify=false` was necessary in order to pass `ascii_only=true` to the CLI(https://github.com/mishoo/UglifyJS2/issues/54).

ylemkimon:
@kevinbarabash 
> Why is it a partial fix?
As `main.js` is not handled by webpack, it's not watched and doesn't reload if changed.

Makefile
ylemkimon: @kevinbarabash I think preserving other files in `build/` is no longer necessary as dev server doesn't generate or use files in `build/`. Will change to `rm -rf build/*`

webpack.common.js
```diff
@@ -0,0 +1,123 @@
...
+const UglifyJsPlugin = require('uglifyjs-webpack-plugin');
+const ExtractTextPlugin = require('extract-text-webpack-plugin');
+const webpack = require('webpack');
+
+/*::
+type Target = {|
+    name: string, // the name of output JS/CSS
+    entry: string, // the path to the entry point
+    library?: string // the name of the exported module
+|};
```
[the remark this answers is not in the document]
ylemkimon: @kevinbarabash Yes 😄

webpack.common.js
ylemkimon: @kevinbarabash Yes, but also because `copy-tex` doesn't export anything and just adds a copy event listener to the document, so only its side effect is needed.

webpack.common.js
ylemkimon: @kevinbarabash [`cssnano`](https://github.com/ben-eb/cssnano) is a CSS minifier equivalent to `clean-css` and it is included in the `css-loader`, which can be enabled by passing `minimize: true` to its options.

webpack.config.js
ylemkimon: @kevinbarabash No, the order doesn't matter as it runs them in parallel. Will change and add comments.

katex.webpack.js
ylemkimon: @kevinbarabash `webpack` `style-loader` inlines imported stylesheets into the output JS files, and when JS files are loaded they are injected to the style tag. This is why css files are not linked in html files.

Using `extract-text-plugin`, they can be extracted to .css files. So in production build where `extract-text-plugin` is not enabled, it builds .css files.

katex.webpack.js
ylemkimon: @kevinbarabash Yes, I think we can (and should) switch to ES6 module `export`, but it seems it's out of scope of this PR.

webpack.common.js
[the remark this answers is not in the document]
ylemkimon: @kevinbarabash It loads included font files in the stylesheet(`fonts.less`) and outputs to `fonts/`.

katex.webpack.js
ylemkimon: @kevinbarabash Actually, build file format is specified in webpack configuration `output.libraryTarget` and it's set to UMD (Universal Module Definition). Changed to ES6 module.

katex.webpack.js
kevinbarabash: Cool.  Thanks for the explanation.

contrib/auto-render/auto-render.js
```diff
@@ -1,6 +1,6 @@
 /* eslint no-console:0 */
-/* global katex */
 
+import katex from "katex";
```
kevinbarabash: Won't this result in the bundle that's built for auto-render.js also containing all of KaTeX?

contrib/mathtex-script-type/mathtex-script-type.js
```diff
@@ -1,24 +1,22 @@
-/* global katex */
+import katex from "katex";
```
kevinbarabash: Same concern as in auto-render.js.

katex.webpack.js
rrandallcainc: I believe this would be a breaking change. Users would just need to be aware of that.

webpack.common.js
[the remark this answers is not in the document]
rrandallcainc: For this to get used, you'd have to import the font into the dependency tree, right? Is that being done?

webpack.config.js
```diff
@@ -1,61 +1,7 @@
-const path = require('path');
+// @flow
+const { targets, createConfig } = require('./webpack.common');
```
rrandallcainc: Would we want to rename this to `webpack.prod.js`?

contrib/copy-tex/index.html
rrandallcainc: I believe webpack will now inject these with either a script tag (extract plugin) or directly into the JS.

dockers/Screenshotter/screenshotter.js
```diff
@@ -315,7 +318,7 @@ function findHostIP() {
 
     // Now we need to find an IP the container can connect to.
     // First, install a server component to get notified of successful connects
-    app.get("/ss-connect.js", function(req, res, next) {
+    devServer.app.get("/ss-connect.js", function(req, res, next) {
```
rrandallcainc: Nice

katex.js
```diff
@@ -95,7 +95,7 @@ const renderToHTMLTree = function(
     return buildHTMLTree(tree, expression, settings);
 };
 
-module.exports = {
+export default {
```
rrandallcainc: Is this a breaking change for CJS users?

server.js
```diff
@@ -1,106 +0,0 @@
-/* eslint no-console:0 */
```
rrandallcainc: 👍

webpack.common.js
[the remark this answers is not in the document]
ylemkimon: @rrandallcainc Yes. Current dependency tree is: `katex.wepack.js` imports `static/katex.less`, `static/katex.less` imports `static/fonts.less`, and `fonts.less` imports(using `@font-face`) font files.

contrib/mathtex-script-type/mathtex-script-type.js
ylemkimon: @kevinbarabash The module `katex` is declared as `externals` in the webpack configuration. As a result, webpack doesn't bundle the KaTeX, but tries to require the module by `require("katex")` in CJS, `define(["katex"], ...)` in AMD, and using global variable `katex`(old behavior). Basically, it add support for module loaders.

webpack.config.js
ylemkimon: @rrandallcainc As `webpack`'s default configuration filename is `webpack.config.js`, if `webpack` is called without any arguments, production build will be run.

katex.js
ylemkimon: @rrandallcainc As for (i) `dist/katex.{min.js,.js}`, `katex` module, webpack exports an UMD module and default export is exposed, and for (ii) `katex.js`, transpilers and/or bundlers uses `interopDefault`, which imports default export when importing ES6 module, I think there'll be no breaking changes.

ylemkimon: @rrandallcainc I'm don't know there is a use case of directly requiring `katex.js` (from the root), e.g., `const katex = require('./node_modules/katex/katex.js);`, as it needs transpiling, but in that case, the user has to use `katex.default`, e.g., `const katex = require('./node_modules/katex/katex.js).default;`.

contrib/mathtex-script-type/mathtex-script-type.js
kevinbarabash: Oh wow.  That's super cool!

package.json
```diff
@@ -3,7 +3,6 @@
   "version": "0.10.0-pre",
   "description": "Fast math typesetting for the web.",
   "main": "dist/katex.js",
-  "browser": "dist/katex.min.js",
```
kevinbarabash: Why was this reverted?

ylemkimon: @kevinbarabash #1057 is no longer needed, since webpack doesn't uses name `require` and https://github.com/browserify/browserify/issues/374 is not an issue.

ylemkimon:
Updated PR description 😄

.babelrc
```diff
@@ -1,7 +1,21 @@
...
+    ],
     "plugins": [
         "transform-runtime",
         "transform-class-properties"
-    ]
+    ],
+    "env": {
+      "test": {
+        "presets": ["es2015", "flow"],
+        "plugins": [
```
ylemkimon: This is for Jest, as it [doesn't support ES6 modules](https://facebook.github.io/jest/docs/en/getting-started.html).

Babel has an issue merging env configurations in `.babelrc`(https://github.com/babel/babel/issues/5276), so plugins configuration has to be duplicated.

.babelrc
```diff
@@ -1,7 +1,19 @@
...
+    ],
     "plugins": [
         "transform-runtime",
         "transform-class-properties"
-    ]
+    ],
+    "env": {
+      "test": {
+        "plugins": [
+            "transform-es2015-modules-commonjs"
```
ylemkimon: This is for Jest, as it [doesn't support ES6 modules](https://facebook.github.io/jest/docs/en/getting-started.html).
In Babel 6, env configurations doesn't overrides configuration, rather are merged to configuration(babel/babel#5276).

.babelrc
```diff
@@ -1,7 +1,19 @@
...
+    ],
     "plugins": [
         "transform-runtime",
         "transform-class-properties"
-    ]
+    ],
+    "env": {
+        "test": {
+            "plugins": [
+                "transform-es2015-modules-commonjs"
```
ylemkimon: This is for Jest, as it [doesn't support ES6 modules](https://facebook.github.io/jest/docs/en/getting-started.html).
In Babel 6, env configurations doesn't overrides configuration, rather are merged to the root configuration(babel/babel#5276).

katex.js
rrandallcainc: Hmm, but I still think katex.js serves as the entry point, right? Even though this will be wrapped in a UMD, I think it's going to be a breaking change when requiring the built artifact. This is due to ES6 modules always being named. When CJS users try and import Katex, they'll need to change their import from `require('katex')` to `require('katex').default`. 

Perhaps I am missing something though.

.babelrc
rrandallcainc: Could you avoid adding this dependency by adding the `.default` to your imports instead?

katex.js
ylemkimon: @rrandallcainc Webpack has option `libraryExport`, which configures which module or modules will be exposed via UMD and it is set to `default`. So requiring `katex` module(`require('katex')`) will expose default export and work same.

.babelrc
ylemkimon: @rrandallcainc Jest doesn't understand ES6 `import`/`export` syntaxes, so I think adding `.default` won't help it. Also this plugin is part of `babel-preset-es2015`, so there won't be any additional dependencies.

webpack.common.js
```diff
@@ -0,0 +1,121 @@
+// @flow
```
ylemkimon: It seems it's possible to
(i) use ES6 in `webpack.config.js` by using the extension `.babel.js`, which transpiles the file before loading
(ii) put all configurations in one file by `webpack.config.js` exporting a function, which takes environment variables as an argument and returns the webpack configuration

kevinbarabash:
@ylemkimon thanks for listing out the breaking changes.  I think the only breaking change that might affect users would be the third one, but my guess would be that most users are using the compiled bundle so this shouldn't be an issue.  We can add it to the breaking changes part of the release notes when we do a release.

From the "old/new" section:
> Export CommonJS module | Export ES6 module[3], expose default export

This sounds scarier than it is b/c the bundle is built in such a way that it exposes a commonjs interface as well as a `katex` global as per the discussion in the comments.  Could you update your comment to make it more explicit that the interface the published module exports remains unchanged?

webpack.common.js
kevinbarabash: I think having the whole config in a single file and using environment variables to control things might be nice.  I'm going land this as is though and any additional work could happen in a followup PR.

edemaine:
Sounds great! Just checking: Does `require('katex')` still work as it did?  We'll definitely want to do another beta (not final) release with this new system.

kevinbarabash:
I did the following as a quick check:
- `npm run build`
- `cd build`
- `node`
- `const katex = require('katex')` // this returned the `katex` object as expected

kevinbarabash:
@davidflanagan just let me know that this change "shaved 40Kb off of katex.min.js!"

Issue #1066, opened 2018-01-15
Poorly rendered Å


DominikRocek:
Hi,
in **v0.9.0-beta**, the symbol **Å** is not rendered correctly.
Thank you for the repair!

![a](https://user-images.githubusercontent.com/35472032/34965606-a137de50-fa55-11e7-8c2d-e2465e3ab743.jpg)

librilex:
Which font did you use? On the [demo site](https://khan.github.io/KaTeX/) entering 
```latex 
m=10^{-10} \cdot \text{Å}
```
works perfectly fine:
![grafik](https://user-images.githubusercontent.com/26027208/34988319-ac67c880-fabe-11e7-874c-8e6910e4a6c8.png)

So I assume it is a problem with the font you are using.

edemaine:
@librilex I believe the demo site is not on 0.9.0-beta yet, which is why `\text{Å}` is rendering differently from `\text{\r A}` in your examples.  (It shouldn't, and doesn't on master.)

@DominikRocek `Å` is currently supported in text mode, but not math mode, because the `\r` accent is supported in text mode but not math mode.  Are you wrapping it in `\text`?

Developers: Should we consider making `Å`, and other false Latin-1 characters that we added for backward compatibility, cause a warning in math mode, given that they render incorrectly?

ronkok:
Below I show a test from 0.9.0-beta.  It appears that KaTeX does currently support `\text{Å}` and `\text{å}`. It just does a poor rendering of both.  The math-mode rendering looks good.
![latin1](https://user-images.githubusercontent.com/16403058/35003569-12af515c-faa2-11e7-94b1-6e8b9c5646b7.PNG)
Of all the Latin-1 characters added by PR #796, we've regressed only for characters  `æ`, `ø`, and `Ø`. I can't get them to render in math mode

DominikRocek:
Hi,
thank you for the answers. I used **\textsf{Å}**.

When I use: **\text{Å}** .. Result is:
![a](https://user-images.githubusercontent.com/35472032/35003259-8fb001a2-faec-11e7-8e17-edbd152e87a4.jpg)

When I use: **\textsf{Å}** .. Result is:
![2](https://user-images.githubusercontent.com/35472032/35003568-8179b51e-faed-11e7-89a5-499671109747.jpg)

When I use: **Å** .. Result is:
[...]

PR #1069, opened 2018-01-16
Add support for \aa, \AA, \lq, \rq, \lbrack, \rbrack

edemaine:
The lack of `\aa` and `\AA` was reported in #1066.  Added these via simple macros, with supporting tests.

Also add `\lq`, `\rq`, `\lbrack`, `\rbrack` as aliases for "\`", `'`, `[`, `]` which were in the same area of latex.ltx.

### DIFF
--- a/src/macros.js
+++ b/src/macros.js
@@ -96,11 +96,27 @@ defineMacro("\\TextOrMath", function(context) {
 });
 
 //////////////////////////////////////////////////////////////////////
-// basics
+// Grouping
+// \let\bgroup={ \let\egroup=}
 defineMacro("\\bgroup", "{");
 defineMacro("\\egroup", "}");
 defineMacro("\\begingroup", "{");
 defineMacro("\\endgroup", "}");
+
+// Symbols from latex.ltx:
+// \let\bgroup={ \let\egroup=}
+// \def\lq{`}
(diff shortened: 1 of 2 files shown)


edemaine:
Thanks for the typo fix, and quick review!

kevinbarabash:
@edemaine thanks for adding all of these commands.

ronkok:
I realize that I'm late here, but I want to point out that `\lbrack` and `\rbrack` were supported by KaTeX before this PR.
lbrack

kevinbarabash:
@ronkok good catch.  Since there are multiple ways to define things, we might want to have some sort of central dictionary of all the commands regardless of whether they were defined as a symbol, function, macro, or environment.  In LaTeX, if you have something that you've defined as a command and then you redefine it as something else, like a color or a length, is LaTeX cool with that?

kevinbarabash:
Anyway, the goal of having a central lookup would be to provide a way to audit for duplicates so that we're not shipping code we don't have to.

ronkok:
Well, I caught it because I was updating the function support page and found that those two functions were already there. So in a sense, the function support page already serves as a low tech dictionary. No doubt you're looking for something more robust.

PR #1070, opened 2018-01-16
Use JS for spacing between atoms instead of CSS

kevinbarabash:
This is the first step towards creating an intermediate representation
that can be used to generate HTML, SVG, and Canvas commands for rendering.
By generating spans that contain the width of the spaces instead of
relying on CSS sibling rules we'll be able to one day replaces the spans
with intermeidate 'Glue' nodes (in a later PR).  This was described in https://github.com/Khan/KaTeX/issues/376#issuecomment-354567190.

An added benefit of this approach is that is enables us to programmatically
change the values for thinspace, mediumspace, and thickspace which will
allow us to implement the \setlength command.


kevinbarabash:
There are some failing tests for this change that I have to work through, in particular:
```
BinCancellation BoldSymbol Not OperatorName OverUnderset SizingBaseline StrikeThrough StrikeThroughColor
```

kevinbarabash:
Somethings have improved in the screenshots:
- spaces between operators and strike through items is closer to what LaTeX does

Somethings are worse:
- Not is in the wrong position when followed by a group
- spaces around `+` in SizingBaseline now depend on the sizing, but shouldn't
- spaces around dot operators in StyleSwitchings should be symmetric but aren't

Somethings are neutral:
- changes to OverUnderset are minimal

kevinbarabash:
strikethrough

I added
```
\usepackage{cancel}
\usepackage{ulem}
```
to test.tex in order to generate the image above.  The spacing around binary operators is better than it used to be.  It seems like the addition of the `\cancel` bar doesn't affect the bounding box of `\cancel` in the LaTeX rendering (green).  The bounding box of `\cancel` groups is essentially the same as the bounding box of the contents.  This is probably most noticeably in `{\cancel B}_1^2`.

kevinbarabash:
not

`\not` is now looking right on for Chrome and Firefox.  We need to change the glyph to a non-combining character to fix Safari.

kevinbarabash:
sizingbaseline

`SizingBaseline` seems to be better too, but not perfect.

kevinbarabash:
All tests are passing now, but there's still a bit of cleanup work that I'd like to do first.  In particular:
- [ ] extract helper function for generating glue
- [ ] do the TODO for handling styling commands the same as sizing commands in `buildExpression`

src/buildHTML.js
```diff
@@ -75,81 +160,80 @@ export const spliceSpaces = function(children, i) {
...
-                container.style.paddingLeft = "0.277771em";
-            }
-
-            // Ensure that the \u0338 is positioned relative to the container.
-            container.style.position = "relative";
-            groups.splice(i, 2, container);
+        if (groups[i].value === "\u0338") {
+            groups[i].style.position = "absolute";
+            // TODO(kevinb) fix this for Safari by switching to a non-combining
+            // character for \not.
```
kevinbarabash: I think TODO makes the most sense in a separate PR since font changes are involved.

src/buildHTML.js
```diff
@@ -75,81 +160,80 @@ export const spliceSpaces = function(children, i) {
...
+            // 0.194440 is the width for the \not character
+            groups[i].style.paddingLeft = `${1.0 - 0.194440}em`;
```
kevinbarabash: Should grab the width metric for this character from fontMetricsData.js.

src/buildHTML.js
```diff
@@ -75,81 +160,80 @@ export const spliceSpaces = function(children, i) {
...
+
+                    glue.style.marginRight = `${dimension}em`;
+                    groups.push(glue);
+                }
+            }
+            j++;
         }
     }
 
     // Process \\not commands within the group.
```
kevinbarabash: I'm happy how this became a lot simpler.

src/functions/delimsizing.js
```diff
@@ -182,15 +183,20 @@ defineFunction({
...
 
+        const lastChildType = html.getTypeOfDomTree(inner[inner.length - 1]);
+        if (lastChildType && html.spacings[lastChildType]["mclose"]) {
+            const glue =
+                buildCommon.makeSpan(["mord", "rule"], [], options);
+            const dimension =
+                calculateSize(html.spacings[lastChildType]["mclose"], options);
+            glue.style.marginRight = `${dimension}em`;
+            inner.push(glue);
+        }
```
kevinbarabash: I wish that delimsizing.js didn't have to know about this.  I was thinking about moving space inserting to be a post-processing step on the HTML tree, but then we don't have access to sizing/styling data so I think we'll just have to live with this.

kevinbarabash:
This is ready for review now.

kevinbarabash:
@kohler if you have time, could you look at this PR since you did original work to fix the spacing?

src/buildCommon.js
```diff
@@ -596,6 +585,14 @@ const makeVerb = function(group: ParseNode, options: Options): string {
     return text;
 };
 
+const makeGlue = (measurement: Measurement, options: Options): domTree.span => {
```
rrandallcainc: I think a comment here would be useful. The term glue might be confusing to people.

src/buildHTML.js
```diff
@@ -75,81 +167,80 @@ export const spliceSpaces = function(children, i) {
  */
 export const buildExpression = function(expression, options, isRealGroup) {
     // Parse expressions into `groups`.
-    const groups = [];
+    const rawGroups = [];
```
rrandallcainc: What was the reasoning for the rename?

src/buildHTML.js
```diff
@@ -75,81 +167,80 @@ export const spliceSpaces = function(children, i) {
...
     for (let i = 0; i < expression.length; i++) {
         const group = expression[i];
         const output = buildGroup(group, options);
         if (output instanceof domTree.documentFragment) {
-            Array.prototype.push.apply(groups, output.children);
+            rawGroups.push(...output.children);
```
rrandallcainc: 👍

static/katex.less
```diff
@@ -149,122 +149,6 @@
     @mediumspace: 0.22222em;    // 4mu
     @thickspace: 0.27778em;     // 5mu
 
-    // These spacings apply in textstyle and displaystyle.
-    .mord {
```
rrandallcainc: ❤️ Love that this is gone

src/buildHTML.js
```diff
@@ -164,6 +255,8 @@ export const getTypeOfDomTree = function(node) {
                 node.children[node.children.length - 1]);
         }
     } else {
+        // This makes a lot of assumptions as to where the type of atom
+        // appears.  We shoul do a better job of enforcing this.
```
rrandallcainc: Nit: typo shoul -> should

src/buildHTML.js
```diff
@@ -174,6 +267,17 @@ export const getTypeOfDomTree = function(node) {
     return null;
 };
 
+export const isTight = function(node) {
+    if (node instanceof domTree.documentFragment) {
+        if (node.children.length) {
+            return isTight(node.children[node.children.length - 1]);
```
rrandallcainc: Why is the recursive call necessary? Is documentFragment just a wrapper? Also, why is the last child a safe assumption?

src/buildHTML.js
```diff
@@ -75,81 +167,80 @@ export const spliceSpaces = function(children, i) {
...
+
+                    groups.push(glue);
+                }
+            }
+            j++;
         }
     }
 
     // Process \\not commands within the group.
-    // TODO(kevinb): Handle multiple \\not commands in a row.
```
rrandallcainc: Assuming this comment handled by this PR?

src/buildHTML.js
```diff
@@ -75,81 +167,80 @@ export const spliceSpaces = function(children, i) {
...
-    // TODO(kevinb): Handle \\not{abc} correctly.  The \\not should appear over
```
rrandallcainc: Same as above

src/buildHTML.js
```diff
@@ -66,6 +66,98 @@ export const spliceSpaces = function(children, i) {
...
+const mediumspace = {
+    number: 4,
+    unit: "mu",
+};
+const thickspace = {
+    number: 5,
+    unit: "mu",
+};
+
+export const spacings = {
```
rrandallcainc: This all looking super clean 👍 . Since this is just data with no dependencies. Perhaps it could be extracted outside of buildHTML into it's own module? Maybe a `spacing.js`?

src/buildHTML.js
```diff
@@ -75,81 +167,80 @@ export const spliceSpaces = function(children, i) {
...
-                    groups[i] = makeSpan([].concat(groups[i].classes),
-                        [groups[i]]);
-                }
-                buildCommon.prependChildren(groups[i], spaces);
-            } else {
-                // Otherwise, put any spaces back at the end of the groups.
-                Array.prototype.push.apply(groups, spaces);
-                break;
+    // spacing (e.g., "add thick space between mord and mrel").
+    const nonSpaces =
```
rrandallcainc: The comment above indicates that this is spacing, but the variable name is `nonSpaces`. Is that intentional?

src/buildHTML.js
```diff
@@ -66,6 +66,98 @@ export const spliceSpaces = function(children, i) {
...
+        mop: thinspace,
+        mbin: mediumspace,
+        mrel: thickspace,
+        mopen: thinspace,
+        mpunct: thinspace,
+        minner: thinspace,
+    },
+};
+
+const tightSpacings = {
```
rrandallcainc: What's the reasoning for distinguishing these from `spacings`?

src/buildHTML.js
```diff
@@ -66,6 +66,98 @@ export const spliceSpaces = function(children, i) {
     }
 };
 
+const thinspace = {
+    number: 3,
+    unit: "mu",
```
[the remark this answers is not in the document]
kevinbarabash: I originally used `em` and had comments and math.  This seemed a lot simpler and maps directly to LaTeX definitions.

src/buildHTML.js
kevinbarabash: The spacings are different when we're in `\scriptstyle` or `\scriptscriptstyle`.  I'm open to suggestions of other ways to organize these alternate values.

src/buildHTML.js
kevinbarabash: I thought about insert spaces into this array, but inserts mess with indices and I thought the code would be cleaner copying non space groups over to a new array as the spaces were being inserted.

src/buildHTML.js
kevinbarabash: I'll expand on this comment.  Basically, we want to look at the relationship between non space groups to determine what spaces to insert.  This means we want to disregard explicit spaces when determining which implicit spaces to add.

src/buildHTML.js
kevinbarabash: See the new screenshots for `Not`. :)

src/buildHTML.js
kevinbarabash: This follows the same pattern as `getTypeOfDomTree`.  Both of these functions need docstrings that explain that if the node is a document fragment then we return the value for the right most node in that fragment.  A document fragment is a tree so it could contain nested nodes hence the recursion.

kevinbarabash:
@rrandallcainc thanks for the review.  I'll make the requested changes this evening.

src/buildHTML.js
```diff
@@ -75,81 +167,80 @@ export const spliceSpaces = function(children, i) {
...
-                        [groups[i]]);
-                }
-                buildCommon.prependChildren(groups[i], spaces);
-            } else {
-                // Otherwise, put any spaces back at the end of the groups.
-                Array.prototype.push.apply(groups, spaces);
-                break;
+    // spacing (e.g., "add thick space between mord and mrel").
+    const nonSpaces =
+        rawGroups.filter(group => group && group.classes[0] !== "mspace");
```
ylemkimon: I think we can remove using dummy `span`s in the enclosing element like:
https://github.com/Khan/KaTeX/blob/732f2a84ccb000099d29fb1f61dd02e611fab64b/src/functions/href.js#L34-L65
and handle spacing here.

kevinbarabash: Good catch.  I'm guessing that this diff breaks `\href` spacing which means we should probably have a test for it.

kevinbarabash: I should probably update `getTypeOfDomTree` to be able to get the class of the rightmost or leftmost element in the tree.

kevinbarabash: @ylemkimon as for removing the dummy spans, it's still possible that the anchor wraps a fragment.  `makeAnchor` always returns a single element even when it has multiple children so I believe we still need to do this otherwise I don't see how we'd communicate there being atoms of different classes at either end of the anchor's content.

kevinbarabash:
@rrandallcainc this is ready for another pass.

src/functions/href.js
```diff
@@ -54,6 +54,8 @@ defineFunction({
             if (first === last) { // Case 2 : type of both ends coincides
                 classes = [first];
             } else { // Case 3: both ends have different types.
+                // TODO(kevinb): figure out a better way to communicate this
+                // information to buildHTML.js#buildExpression.
```
ylemkimon: I think we can just `return new buildCommon.makeAnchor(href, [], elements, options);` here

src/buildHTML.js
```diff
@@ -75,95 +83,111 @@ export const spliceSpaces = function(children, i) {
...
         }
     }
 
     return groups;
 };
 
 // Return math atom class (mclass) of a domTree.
-export const getTypeOfDomTree = function(node) {
+export const getTypeOfDomTree = function(node, side = "right") {
     if (node instanceof domTree.documentFragment) {
```
ylemkimon: and do same for `node instanceof domTree.anchor`.

ylemkimon: Also I think we can define a special class name for `span` that encloses elements without affecting spacings.

src/buildHTML.js
```diff
@@ -75,95 +83,111 @@ export const spliceSpaces = function(children, i) {
...
-                if (groups[i] instanceof domTree.symbolNode) {
-                    groups[i] = makeSpan([].concat(groups[i].classes),
-                        [groups[i]]);
-                }
-                buildCommon.prependChildren(groups[i], spaces);
-            } else {
-                // Otherwise, put any spaces back at the end of the groups.
-                Array.prototype.push.apply(groups, spaces);
-                break;
+    // Ignore explicit spaces (e.g., \;, \,) when determine what implicit
```
kohler: “when determining”

src/buildHTML.js
```diff
@@ -75,95 +83,111 @@ export const spliceSpaces = function(children, i) {
...
+            // Get the type of the current non-space node.  If it's a document
+            // fragment, get the type of the rightmost node in the fragment.
+            const left = getTypeOfDomTree(nonSpaces[j], "right");
+
+            // Get the type of the next non-space node.  If it's a document
+            // fragment, get the type of the leftmost node in the fragment.
+            const right = getTypeOfDomTree(nonSpaces[j + 1], "left");
+
+            // We use buildExpression inside of sizingGroup, but it returns a
+            // document fragement of elements.  sizingGroup sets `isRealGroup`
```
kohler: “fragment”

kevinbarabash:
@kohler thanks!

I'll fix all the nits before merging.

rrandallcainc:
@kevinbarabash This is looking good. I'll do one more pass tomorrow morning.

src/buildCommon.js
```diff
@@ -596,6 +585,17 @@ const makeVerb = function(group: ParseNode, options: Options): string {
     return text;
 };
 
+// Glue is a concept from TeX which is a flexible space between elements in
```
rrandallcainc: Cool, didn't know this!

src/buildHTML.js
kevinbarabash: Good call.

src/functions/href.js
kevinbarabash: I was confused at first, but this makes sense given your comments in `getTypeOfDomTree`.

src/buildHTML.js
kevinbarabash: I ran into some issues with `span`s so I'm going to leave them out for now.

kevinbarabash:
Thanks everyone for reviewing this.  This should make #687 much more tractable now.

rrandallcainc:
@kevinbarabash Thanks for implementing. Would extracting these: https://github.com/Khan/KaTeX/blob/master/static/katex.less#L217 follow a similar pattern?

kevinbarabash:
@rrandallcainc yup.  Anywhere where we create a span using one of those styles could be replaced with a call to `makeGlue`.

src/buildHTML.js
```diff
@@ -174,6 +200,21 @@ export const getTypeOfDomTree = function(node) {
     return null;
 };
 
+// If `node` is an atom return whether it's been assigned the mtight class.
+// If `node` is a document fragment, return the value of isLeftTight() for the
+// leftmost node in the fragment.
+// 'mtight' indicates that the node is script or scriptscript style.
+export const isLeftTight = function(node) {
+    if (node instanceof domTree.documentFragment) {
```
ylemkimon: Can the `node` be a `domTree.documentFragment`? Aren't they flattened at the start of `buildExpression`?

src/buildHTML.js
```diff
@@ -75,95 +83,113 @@ export const spliceSpaces = function(children, i) {
...
+    // spacing should go between atoms of different classes.
+    const nonSpaces =
+        rawGroups.filter(group => group && group.classes[0] !== "mspace");
+
+    // Before determining what spaces to insert, perform bin cancellation.
+    // Binary operators change to ordinary symbols in some contexts.
+    for (let i = 0; i < nonSpaces.length; i++) {
+        if (isBin(nonSpaces[i])) {
+            if (isBinLeftCanceller(nonSpaces[i - 1], isRealGroup)
+                    || isBinRightCanceller(nonSpaces[i + 1], isRealGroup)) {
```
ylemkimon: I think these functions should take `domTree.documentFragment` and `domTree.anchor` into account

src/functions/href.js
ylemkimon: Currently spacings around `\href` are duplicated as dummy spans are created.

src/buildHTML.js
kevinbarabash: Good catch.

src/buildHTML.js
kevinbarabash: `domTree.anchor` yes, but as noted in your previous comment, `documentFragment`s get flattened.

src/buildHTML.js
ylemkimon: @kevinbarabash It seems that `node`'s children can be a `domTree.documentFragment`, e.g., `\phantom{x}^2`.

src/buildHTML.js
```diff
@@ -75,95 +83,113 @@ export const spliceSpaces = function(children, i) {
...
+                            glueOptions = options.havingSize(
+                                expression[0].value.size);
+                        } else if (expression[0].type === "styling") {
+                            glueOptions = options.havingStyle(
+                                styleMap[expression[0].value.style]);
+                        }
+                    }
+
+                    const glue = buildCommon.makeGlue(
+                        spacings[left][right], glueOptions);
```
ylemkimon: Is using `spacings[left][right]` instead of `space` an intended behavior?

kevinbarabash: Yes.  Thanks for catching this.  It's obvious that we don't have any tests for the "tight" situation.

kevinbarabash: Tracking in #1102.

PR #1071, opened 2018-01-17
Move auto-render-spec.js to ./test

ylemkimon:
Fixed jest not running auto-render-spec.js
    + Changed `require` to `import`


kevinbarabash:
Thanks for spotting this and fixing it.  The `testMatch` in the jest config is pretty specific.  I'm sure this is fine, but I'm going to wait until travis is back up.

PR #1072, opened 2018-01-17
Export ES6 module in fontMetricsData.js

ylemkimon:
Use ES6 modules export(`export default`)


kevinbarabash:
@ylemkimon I assume you ran `make metrics` to generate fontMetricsData.js, is that right?

ylemkimon:
@kevinbarabash Yes 😄

PR #1073, opened 2018-01-17
\copyright, \textregistered, \textcircled support

edemaine:
This seems to work at regular size and `\small`, but not e.g. in subscript. Suggestions welcome.

Based on discussions with @kevinbarabash in https://github.com/KaTeX/katex-fonts/issues/9


kevinbarabash:
@edemaine can you post what it looks like for `\textstyle` and `\scriptstyle`?  Is it just the subscript that looks wrong, or is it also wrong in superscript?

edemaine:
Here's the current state for http://localhost:7936/?text=%5Cbegin%7Bmatrix%7D%0A%5Ccopyright%5Csmall%5Ccopyright%5Cfootnotesize%5Ccopyright%5Ctiny%5Ccopyright%5C%5C%0A%5Ccopyright_%7B%5Ccopyright_%7B%5Ccopyright%7D%7D%0A%5Cend%7Bmatrix%7D%0A :

image

It looks like `\llap` is resetting the font to normalsize.  This seems to be a bug in `\llap`, because LaTeX doesn't work that way.

edemaine:
I revisited this, implementing an actual `\textcircled` command as an accent, like in LaTeX.  This accent seems to behave differently from all other accents in that its width contributes to the width of the final symbol (cf. #1033), and it is never offset vertically.  I'm not sure what triggers this in (La)TeX, but I'm calling this a "full" accent barring a better name.  `\copyright` behaves very well now, as does `\textregistered`:

```latex
\copyright_{\copyright_{\copyright}} \small \copyright \tiny \copyright
```
image

This should be good to go, unless someone knows better terminology for this type of accent.

src/macros.js
```diff
@@ -117,6 +117,16 @@ defineMacro("\\rbrack", "]");
...
 
+// \DeclareTextCommandDefault{\textcopyright}{\textcircled{c}}
+// \DeclareTextCommandDefault{\textregistered}{\textcircled{%
+//      \check@mathfonts\fontsize\sf@size\z@\math@fontsfalse\selectfont R}}
+// \DeclareRobustCommand{\copyright}{%
+//    \ifmmode{\nfss@text{\textcopyright}}\else\textcopyright\fi}
+defineMacro("\\textcopyright", "\\textcircled{c}");
+defineMacro("\\copyright",
+    "\\TextOrMath{\\textcopyright}{\\text{\\textcopyright}}");
+defineMacro("\\textregistered", "\\textcircled{\\scriptsize R}");
```
kevinbarabash: yay for macros

src/katex.less
```diff
@@ -401,8 +401,13 @@
             text-align: center;
         }
 
-        .accent-body {
+        // Accents that are not of the accent-full class have zero width
+        // (do not contribute to the width of the final symbol).
+        .accent-body:not(.accent-full) {
             width: 0;
+        }
```
kevinbarabash: I wish we store more props directly on the `domTree` nodes we create so that we don't have all of this extra data hanging out in this file.  No action required.

src/symbols.js
```diff
@@ -684,6 +684,7 @@ defineSymbol(text, main, accent, "\u02da", "\\r"); // ring above
 defineSymbol(text, main, accent, "\u02c7", "\\v"); // caron
 defineSymbol(text, main, accent, "\u00a8", '\\"'); // diaresis
 defineSymbol(text, main, accent, "\u02dd", "\\H"); // double acute
+defineSymbol(text, main, accent, "\u25ef", "\\textcircled"); // \bigcirc glyph
```
kevinbarabash: I see, this is just an alias.

PR #1074, opened 2018-01-18
[PoC]Plugin architecture using webpack (HTML \class extension)

### LINKED ISSUE #90: Support for non-standard macros \class and \cssId

### LINKED ISSUE #762: define requirements for a plugin architecture

ylemkimon:
Depends on #1068 
Fixes #762 
Fixes #90, fixes #507 

See https://github.com/Khan/KaTeX/pull/1074/commits/5824510d7c56195d6254b23f70af6e5eb86d16b3.

This is a proof of concept of implementing a plugin architecture using webpack and implements non-standard `\class{name}{math}` extension. Using `CommonsChunkPlugin`, it is possible to build codes using KaTeX internal functions, without exposing them to the public API(although it exposes `webpackJsonp__name_` to the global) or duplicating their codes.

**Things to consider:**
- [ ] Custom lexer/parser
- [ ] Plugin registry
- [ ] 3rd party plugin
- [ ] Backward/Forward compatibility
* Use NamedModulesPlugin and HashedModuleIdsPlugin
* Version range check
* Used features list
* Runtime tests
* Wrap codes in try-catch, and unregister the plugin if error is caught


static/index.html
```diff
@@ -4,6 +4,7 @@
         <meta charset="UTF-8">
         <title>KaTeX Test</title>
         <script src="katex.js" type="text/javascript"></script>
+        <script src="contrib/html.js" type="text/javascript"></script>
```
ylemkimon: The user can use the plugin like this.

ylemkimon: The file looks like:
<details>

```
(function webpackUniversalModuleDefinition(root, factory) {
	if(typeof exports === 'object' && typeof module === 'object')
		module.exports = factory();
	else if(typeof define === 'function' && define.amd)
		define([], factory);
	else if(typeof exports === 'object')
		exports["contrib/html"] = factory();
	else
		root["katex"] = root["katex"] || {}, root["katex"]["contrib/html"] = factory();
})(this, function() {
return webpackJsonpkatex__name_([0],{

/***/ 59:
/***/ (function(module, __webpack_exports__, __webpack_require__) {

"use strict";
Object.defineProperty(__webpack_exports__, "__esModule", { value: true });
/* harmony import */ var __WEBPACK_IMPORTED_MODULE_0_babel_runtime_helpers_toConsumableArray__ = __webpack_require__(30);
/* harmony import */ var __WEBPACK_IMPORTED_MODULE_0_babel_runtime_helpers_toConsumableArray___default = __webpack_require__.n(__WEBPACK_IMPORTED_MODULE_0_babel_runtime_helpers_toConsumableArray__);
/* harmony import */ var __WEBPACK_IMPORTED_MODULE_1__src_defineFunction__ = __webpack_require__(2);
/* harmony import */ var __WEBPACK_IMPORTED_MODULE_2__src_buildCommon__ = __webpack_require__(0);
/* harmony import */ var __WEBPACK_IMPORTED_MODULE_3__src_mathMLTree__ = __webpack_require__(1);
/* harmony import */ var __WEBPACK_IMPORTED_MODULE_4__src_domTree__ = __webpack_require__(12);
/* harmony import */ var __WEBPACK_IMPORTED_MODULE_5__src_buildHTML__ = __webpack_require__(3);
/* harmony import */ var __WEBPACK_IMPORTED_MODULE_6__src_buildMathML__ = __webpack_require__(4);









Object(__WEBPACK_IMPORTED_MODULE_1__src_defineFunction__["b" /* default */])({
    type: "class",
    names: ["\\class"],
    props: {
        numArgs: 2,
        allowedInText: true,
        greediness: 3,
        argTypes: ["string", "original"]
    },
    handler: function handler(context, args) {
        var classes = args[0];
        var body = args[1];
        return {
            type: "class",
            classes: classes.value,
            value: Object(__WEBPACK_IMPORTED_MODULE_1__src_defineFunction__["c" /* ordargument */])(body)
        };
    },
    htmlBuilder: function htmlBuilder(group, options) {
        var elements = Object(__WEBPACK_IMPORTED_MODULE_5__src_buildHTML__["a" /* buildExpression */])(group.value.value, options, false);
        var classes = group.value.classes.trim().split(/\s+/);
        var fragment = new __WEBPACK_IMPORTED_MODULE_2__src_buildCommon__["a" /* default */].makeFragment(elements);

        fragment.children.forEach(function (children) {
            if (!(children instanceof __WEBPACK_IMPORTED_MODULE_4__src_domTree__["a" /* default */].svgNode)) {
                var _children$classes;

                (_children$classes = children.classes).push.apply(_children$classes, __WEBPACK_IMPORTED_MODULE_0_babel_runtime_helpers_toConsumableArray___default()(classes));
            }
        });
        return fragment;
    },
    mathmlBuilder: function mathmlBuilder(group, options) {
        var inner = Object(__WEBPACK_IMPORTED_MODULE_6__src_buildMathML__["a" /* buildExpression */])(group.value.value, options);
        var node = new __WEBPACK_IMPORTED_MODULE_3__src_mathMLTree__["a" /* default */].MathNode("mstyle", inner);
        node.setAttribute("class", group.value.classes);
        return node;
    }
});

/***/ })

},[59])["default"];
});
```
</details>

static/index.html
```diff
@@ -4,7 +4,7 @@
         <meta charset="UTF-8">
         <title>KaTeX Test</title>
         <script src="katex.js" type="text/javascript"></script>
-        <link href="katex.css" rel="stylesheet" type="text/css">
+        <script src="contrib/html.js" type="text/javascript"></script>
```
ylemkimon: The user can use the plugin like this.

ylemkimon: The file looks like:
<details>

```
(function webpackUniversalModuleDefinition(root, factory) {
	if(typeof exports === 'object' && typeof module === 'object')
		module.exports = factory();
	else if(typeof define === 'function' && define.amd)
		define([], factory);
	else if(typeof exports === 'object')
		exports["contrib/html"] = factory();
	else
		root["contrib/html"] = factory();
})(this, function() {
return webpackJsonp_name_([0],{

/***/ 59:
/***/ (function(module, __webpack_exports__, __webpack_require__) {

"use strict";
Object.defineProperty(__webpack_exports__, "__esModule", { value: true });
/* harmony import */ var __WEBPACK_IMPORTED_MODULE_0_babel_runtime_helpers_toConsumableArray__ = __webpack_require__(30);
/* harmony import */ var __WEBPACK_IMPORTED_MODULE_0_babel_runtime_helpers_toConsumableArray___default = __webpack_require__.n(__WEBPACK_IMPORTED_MODULE_0_babel_runtime_helpers_toConsumableArray__);
/* harmony import */ var __WEBPACK_IMPORTED_MODULE_1__src_defineFunction__ = __webpack_require__(2);
/* harmony import */ var __WEBPACK_IMPORTED_MODULE_2__src_buildCommon__ = __webpack_require__(0);
/* harmony import */ var __WEBPACK_IMPORTED_MODULE_3__src_mathMLTree__ = __webpack_require__(1);
/* harmony import */ var __WEBPACK_IMPORTED_MODULE_4__src_domTree__ = __webpack_require__(12);
/* harmony import */ var __WEBPACK_IMPORTED_MODULE_5__src_buildHTML__ = __webpack_require__(3);
/* harmony import */ var __WEBPACK_IMPORTED_MODULE_6__src_buildMathML__ = __webpack_require__(4);









Object(__WEBPACK_IMPORTED_MODULE_1__src_defineFunction__["b" /* default */])({
    type: "class",
    names: ["\\class"],
    props: {
        numArgs: 2,
        allowedInText: true,
        greediness: 3,
        argTypes: ["string", "original"]
    },
    handler: function handler(context, args) {
        var classes = args[0];
        var body = args[1];
        return {
            type: "class",
            classes: classes.value,
            value: Object(__WEBPACK_IMPORTED_MODULE_1__src_defineFunction__["c" /* ordargument */])(body)
        };
    },
    htmlBuilder: function htmlBuilder(group, options) {
        var elements = Object(__WEBPACK_IMPORTED_MODULE_5__src_buildHTML__["a" /* buildExpression */])(group.value.value, options, false);
        var classes = group.value.classes.trim().split(/\s+/);
        var fragment = new __WEBPACK_IMPORTED_MODULE_2__src_buildCommon__["a" /* default */].makeFragment(elements);

        fragment.children.forEach(function (children) {
            if (!(children instanceof __WEBPACK_IMPORTED_MODULE_4__src_domTree__["a" /* default */].svgNode)) {
                var _children$classes;

                (_children$classes = children.classes).push.apply(_children$classes, __WEBPACK_IMPORTED_MODULE_0_babel_runtime_helpers_toConsumableArray___default()(classes));
            }
        });
        return fragment;
    },
    mathmlBuilder: function mathmlBuilder(group, options) {
        var inner = Object(__WEBPACK_IMPORTED_MODULE_6__src_buildMathML__["a" /* buildExpression */])(group.value.value, options);
        var node = new __WEBPACK_IMPORTED_MODULE_3__src_mathMLTree__["a" /* default */].MathNode("mstyle", inner);
        node.setAttribute("class", group.value.classes);
        return node;
    }
});

/***/ })

},[59])["default"];
});
```
</details>

rrandallcainc:
This is really exciting to see. I'd like to take a deeper look at this soon. Would applying the non-standard `\id{name}{latex}` follow a similar pattern?

Makefile
```diff
@@ -1,5 +1,5 @@
-.PHONY: build dist lint setup copy serve clean metrics test coverage zip contrib flow
-build: test build/katex.min.js build/katex.min.css contrib zip compress
+.PHONY: build dist lint setup webpack serve clean metrics test coverage zip flow
```
rrandallcainc: Assuming most of the infrastructure related changes go away once the webpack PR is merged? 

What changes are specific to this PR?

ylemkimon: @rrandallcainc Currently only the last commit(https://github.com/Khan/KaTeX/pull/1074/commits/f3b64be8f5a6d2db03387a29e688e514c0a87373) is specific to this PR.

ylemkimon:
@rrandallcainc As an element ID should be unique, I think it should be applied to the enclosing span.

Makefile
rrandallcainc: Doh, should have read your full comment :)

rrandallcainc:
Just want to make sure I understand this plugin system:

Since there are two separate entry points (contrib/html, and katex) which both share common KaTeX code, you use the CommonsChunkPlugin to extract the KaTeX specific code into it's own chunk, which these two entry points can share. This ensures that there's no code duplication. Is that correct?

Might need clarification from @kevinbarabash but was the intention to inject plugin functionality at run-time, not build-time? In other words, I think this change is dependent on the webpack config. If an outside user wants to plugin their own custom function, would they need to create a webpack config to allow them to do so?

contrib/html/html.js
```diff
@@ -0,0 +1,46 @@
+// @flow
+import defineFunction, {ordargument} from "../../src/defineFunction";
+import buildCommon from "../../src/buildCommon";
+import mathMLTree from "../../src/mathMLTree";
+import domTree from "../../src/domTree";
+
+import {buildExpression as htmlBuildExpression} from "../../src/buildHTML";
+import {buildExpression as mmlBuildExpression} from "../../src/buildMathML";
+
+defineFunction({
```
rrandallcainc: It feels a little odd that some of this logic lies outside of the source (like this), but other logic needs to lie within the source (like Parser/types). That being said, I'm not sure there's a better way to implement it at this time without a major change.

ylemkimon:
@rrandallcainc 
> Since there are two separate entry points (contrib/html, and katex) which both share common KaTeX code, you use the CommonsChunkPlugin to extract the KaTeX specific code into its own chunk, which these two entry points can share.

The common codes(KaTeX specific codes) are extracted to the `katex` module, so KaTeX codes leave in the `katex` module.

> This ensures that there's no code duplication, and contrib/html can leverage common katex code. Is that correct?

Yes.

> In other words, I think this change is dependent on the webpack config. If an outside user wants to plugin their own custom function, would they need to create a webpack config to allow them to do so?

Yes, the developer has to have KaTeX sources and add an entry to the webpack config to build a plugin. It's like a SDK.

> It feels a little odd that some of this logic lies outside of the source (like this), but other logic needs to lie within the source (like Parser/types). That being said, I'm not sure there's a better way to implement it at this time without a major change.

I'm thinking of `defineGroupParser` function, which add custom group parsers to the Parser's `parseGroupOfType`. I think with `defineGroupParser`, `defineFunction`, `defineMacro`, `defineEnvironment`, most of features can be implemented without changing KaTeX's sources.

src/Parser.js
```diff
@@ -92,6 +92,12 @@ function assertFuncOrArg(parsed: ParsedFuncOrArgOrDollar): ParsedFuncOrArg {
     return parsed;
 }
 
+type GroupParser = (optional: boolean) => ?ParsedFuncOrArgOrDollar;
+const GROUP_PARSER_EXTENSIONS: { [type: string]: GroupParser } = {};
+export function defineGroupParser(type: string, parser: GroupParser) {
+    GROUP_PARSER_EXTENSIONS[type] = parser;
```
ylemkimon: Like this

kevinbarabash:
My original thinking of a plugin system is that it would be more of a runtime system thing.  I hadn't considered leveraging the fact that we're moving to webpack to build KaTeX.  It feels a little weird imposing plugin makers use a particular build system.  I need to think about it a bit more.

A runtime plugin system would require us to expose more methods for generating HTML and MathML.  I think if/when we moved to HAST for our internal representation of HTML and MathML then we wouldn't have to expose as much.  If/when we move to an intermediate representation we'd still want to expose methods for building that IR.

ylemkimon:
@kevinbarabash Done.

contrib/html/html.js
```diff
@@ -0,0 +1,59 @@
...
+
+defineGroupParser("string", function(optional: boolean): ?ParsedArg {
+    const res = this.parseStringGroupWithBalancedBraces("string", optional);
+    if (!res) {
+        return null;
+    }
+    return newArgument(new ParseNode("string", res.text, "text"), res);
+});
+
+// $FlowFixMe
```
kevinbarabash: What's the error?

src/Parser.js
kevinbarabash: Interesting idea.  This could also be used for defining a custom parser for `\ce` from the `mhchem` package.

src/Parser.js
```diff
@@ -758,6 +764,12 @@ export default class Parser {
             return this.parseUrlGroup(optional);
         }
```
kevinbarabash: Maybe `defineGroupParser` can be used for `color`, `size`, `url`, and even `original`.

contrib/html/html.js
```diff
@@ -0,0 +1,59 @@
...
+    htmlBuilder: (group, options) => {
+        const elements = htmlBuildExpression(group.value.value, options, false);
+        const classes = group.value.classes.trim().split(/\s+/);
+        const fragment = new buildCommon.makeFragment(elements);
+
+        fragment.children.forEach(children => {
+            if (!(children instanceof domTree.svgNode)) {
+                children.classes.push(...classes);
+            }
+        });
```
kevinbarabash: I would've expected all of the elements to be wrapped in a single `span` with the given class applied.  This applies the class to each element.  Is that what MathJax does?

contrib/html/html.js
```diff
@@ -0,0 +1,59 @@
...
+        return fragment;
+    },
+    mathmlBuilder: (group, options) => {
+        const inner = mmlBuildExpression(group.value.value, options);
+        const node = new mathMLTree.MathNode("mstyle", inner);
+        node.setAttribute("class", group.value.classes);
+        return node;
```
kevinbarabash: The behavior of the MathML builder is quite different from the HTML builder.  They should probably have similar semantics, e.g. apply the class to each individual element or wrapping all of the in a parent element.

webpack.common.js
```diff
@@ -7,7 +7,7 @@ const webpack = require('webpack');
 /*::
 type Target = {|
     name: string, // the name of output JS/CSS
-    entry: string, // the path to the entry point
+    entry: string | Object, // the path to the entry point
```
kevinbarabash: It's too bad there aren't any existing flow type definitions for webpack.

webpack.common.js
```diff
@@ -99,6 +102,10 @@ function createConfig(target /*: Target */, dev /*: boolean */,
             new webpack.EnvironmentPlugin({
                 NODE_ENV: dev ? 'development' : 'production',
             }),
+            target.name === 'katex' && new webpack.optimize.CommonsChunkPlugin({
+                name: 'katex',
+                minChunks: Infinity,
+            }),
```
kevinbarabash: So if any modules from the `katex` chunk also appear in any other chunk (e.g. the `contrib/html`) remove them from the other chunk.

ylemkimon: @kevinbarabash Yes 😄

contrib/html/html.js
ylemkimon: @kevinbarabash No, MathJax does, what you said:
>all of the elements to be wrapped in a single `span` with the given class applied.

Though, I'm not sure which is more suitable. For warraping in a `span`, I think we can do in a similar way to `\href`:
https://github.com/Khan/KaTeX/blob/2a2742453b86fba6fd5184efb3c3255400796fb7/src/functions/href.js#L25-L66

contrib/html/html.js
ylemkimon: @kevinbarabash `props.argTypes` expects an array of `ArgType`, so it doesn't recognize custom `string` group type. I couldn't find a way to extend `ArgType` from `defineGroupParser`.

webpack.common.js
ylemkimon: @kevinbarabash There seems to be an open issue in https://github.com/flowtype/flow-typed/issues/165. Also webpack has JSON Schema([schemas/webpackOptionsSchema.json](https://github.com/webpack/webpack/blob/master/schemas/webpackOptionsSchema.json)) and maybe we can convert it using `json-schema-to-flow-type`. I wonder there is any way to import JSON schema to flow directly.

rrandallcainc:
I think if we go with this way for now, it would be good to add some documentation around how users can leverage this plugin system.

ylemkimon:
@rrandallcainc I think we can provide a boilerplate plugin.

contrib/html/html.js
ylemkimon: Blocked on #1070

kevinbarabash: #1070 has been merged, but I think we should probably do what MathJax does since that's probably the only other TeX renderer which supports `\class`.

webpack.common.js
kevinbarabash: I wouldn't worry about it for this PR.

kevinbarabash:
> Custom lexer/parser

Looks like you have at least the parser half done.

> Plugin registry

We need some plugins first.  :P
 
> 3rd party plugin

I suppose that someone could make one use the technique demonstrated in this PR.  The problem though is that they'd need to require a number of files directly from the source.  Pretty much every JS library I've seen that supports plugins allows to create a plugin either without an dependencies or by requiring some utilities to help work with the data structures the library uses.

I think the code in this PR is useful.  It expands the number of extension points we have by introducing `defineGroupParser`.  It also provides a way for contrib extensions to access internals without duplicating code that already exists in the katex bundle.  I definitely want to merge this PR, but I would hesitate to call it a plugin architecture, mainly b/c I think people have certain expectations of JS plugin systems.

This could use some documentation.  Could you add an `Extending KaTeX` section to README.md and describe when you would want to create an extension using this technique and what are the main steps in doing so?

> Backward/Forward compatibility

We'd probably want some sort of versioning system.  Even before that though there are some key changes that I'd like to make first:
- simplify our parse tree (flatten and make it more regular?)
- swap out domTree for HAST
- introduce and intermediate representation

I think the last one may take a while so maybe that's a v2 thing and maybe we can move KaTeX to v1 after we finish the other two.

contrib/html/html.js
kevinbarabash: I think it's fine to just change it in types.js.  Eventually though, we may say that
```
type ArgType = string;
```
and add other programmatic checks as necessary.

rrandallcainc:
Agreed with `\class` being a single span. I assume that it would be easy to have `\id` follow the same pattern?

ylemkimon:
@rrandallcainc Yes, if we use single span, we can just set the id of it.

---
Blocked on #1103.

kevinbarabash:
@ylemkimon I wonder if you had any thoughts about how we might allow plugins/extensions to add fonts and font metrics.  I looking into generating Cyrillic fonts.  If we add support for more scripts with fonts and metrics it would be nice to allow people to load only those which they think they might need.

kevinbarabash:
I believe this is no longer blocked by other PRs.

ylemkimon:
@kevinbarabash Sorry, I was on a trip 😄 This is still blocked by spacing issues like #1108 and #1154.

dvergeylen:
Hi,

May I ask why this PR has been closed? Had it been included in other commits of some kind?

I would be very interested by adding custom classes.

#1108 and #1154 are both closed by now, is there anything else/reason blocking this PR?

kevinbarabash:
Custom classes (and ids) are being added in #1437.

PR #1076, opened 2018-01-19
Support more scripts in \text{} environments.

davidflanagan:
This diff is a follow-up to PR #1060 which added support for Indic scripts.
In order to support Czech, Turkish and Hungarian text (at least) inside
\text{} environments, we need to recognize the Latin Extended A and B
Unicode blocks. The patch also adds support for Georgian, and enhances
support for Cyrillic by defining the entire Cyrillic unicode block instead
of defining symbols for a subset of Cyrillic letters as we did previously.


davidflanagan:
The BoldSymbol screenshot failure has to do with the rendering of the (turkish) dotless i. In the original it is a curvy script i, but with this patch it becomes a straight sans-serif i. That seems like a regression, not an improvement!

The screenshot tests are hard, though: running them locally, I get a number of failures even on the master branch where I haven't changed anything.

davidflanagan:
Somehow with this patch the \boldsymbol\imath is rendered with css classes "mord boldsymbol latin_fallback" instead of "mord mathbf" as it is on master.  The latin_fallback is expected and does nothing, but boldsymbol instead of mathbf is a problem: it changes the font from Katex_Main to Katex_Math, and adds font-style:italic

kevinbarabash:
> The screenshot tests are hard, though: running them locally, I get a number of failures even on the master branch where I haven't changed anything.

Which command are you running?  `make screenshots` generates new screenshots, but `dockers/Screenshotter/screenshotter.sh --verify` is what's used to actually verify them.  The latter reruns the test multiple times on failure to avoid issues with fonts not loading right away.

src/fontMetrics.js
```diff
@@ -191,6 +191,7 @@ export type CharacterMetrics = {
 const getCharacterMetrics = function(
     character: string,
     font: string,
+    mode: string,
```
kevinbarabash: Use `Mode` from types.js.

test/unicode-spec.js
```diff
@@ -110,38 +110,63 @@ describe("unicode", function() {
...
             expect(supportedCodepoint(codepoint)).toBe(
-                allRE.test(String.fromCharCode(codepoint))
+                allRegExp.test(String.fromCharCode(codepoint))
             );
         }
     });
 
     it("scriptFromCodepoint() should return correct values", () => {
-        for (let codepoint = 0; codepoint <= 0xffff; codepoint++) {
+        outer: for (let codepoint = 0; codepoint <= 0xffff; codepoint++) {
```
kevinbarabash: I was just reading a blogpost today about little known JS features today and labels was one of the things in the blogpost.

PR #1087, opened 2018-01-24
A WIP patch toward making characters like å display correctly in text mode

davidflanagan:
See Issue #1066 for more details.

Note that this patch breaks some tests that will need to be fixed.

### DIFF
--- a/src/Parser.js
+++ b/src/Parser.js
@@ -1033,8 +1033,11 @@ export default class Parser {
         }
         // At this point, we should have a symbol, possibly with accents.
         // First expand any accented base symbol according to unicodeSymbols.
+        // But note that we only do this in math mode and for codepoints
+        // that we do not support in text mode
         if (unicodeSymbols.hasOwnProperty(text[0]) &&
-            !symbols[this.mode][text[0]]) {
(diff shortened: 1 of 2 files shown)


davidflanagan:
This is what I had in mind for fixing issue #1066

davidflanagan:
@kevinbarabash  and @edemaine what do you think? This approach is somewhat at odds with https://github.com/Khan/KaTeX/pull/1069 which supports the decomposed approach to accented forms in text mode. 

I'm not sure what LaTeX approach to Unicode is: do we embrace codepoints like å and allow them to display as single glyphs while also retaining support for pre-Unicode things like \aa?

kevinbarabash:
@davidflanagan this is what we used to do.  Unfortunately, it results in the wrong font being used for latin characters with accents.  Even though the font contains the base character and the accent, the browser doesn't know to render those together.  All it knows is that the glyph for `å` doesn't exist in the font and so it falls back to another font which does have the glyph.  The results in the following rendering for `\text{aå}` which is not what we want:
screen shot 2018-01-24 at 10 48 34 am

kevinbarabash:
Closing in preference of #1094.

Issue #1088, opened 2018-01-24
SVG-based functions like \sqrt don't render reliably inside other SVG images.


davidflanagan:
At Khan Academy we sometimes use KaTeX strings inside of SVG images. But now that \sqrt{} and \vec{} are themselves rendered as SVGs we've discovered a problem. Our SVG images set fill-opacity to 0, so when we try to render \sqrt{2} inside of one of these images, the square root sign is fully transparent and does not show up at all.

I've discussed this with @kevinbarabash. We're thinking that we should add some kind of `svg-reset` class name to the toplevel HTML element that we output. And then define this class to reset SVG css properties (like fill-opacity and stroke-width) to their default values.

The idea is that we don't want SVG glyphs to be distinguishable from glyphs actually rendered by a real font, so we shouldn't allow anything in the CSS environment to affect SVG glyphs if it would not also affect font glyphs. So presumably we should inherit the CSS color property, but not the SVG-specific CSS stroke property.

davidflanagan:
It turns out that this is an issue even without the svg-inside-svg. We need to prevent our svg glyphs from inheriting styles that should not affect them:

<img width="958" alt="screen shot 2018-01-24 at 2 24 49 pm" src="https://user-images.githubusercontent.com/516914/35360423-81393aae-0112-11e8-90e3-7833a79d6f4d.png">

ronkok:
I am fully in agreement with your proposed solution for styles.

Since the topic is nested SVGs, I wonder has anyone tried zooming them on Safari? Issue #883 revealed a problem.

kevinbarabash:
Closing as #1089 has be merged.  @ronkok I tried zooming (in and out) on Safari and things seem fine now.

PR #1089, opened 2018-01-25
Don't inherit SVG style properties from the environment.

davidflanagan:
We don't want to inherit styles that affect SVG painting unless those
styles also affect regular text rendering. This patch adds styles
to the stylesheet so that every svg element in KaTeX output starts
off with a known set of basic fill and stroke attributes.

This should resolve issue #1088


davidflanagan:
Okay, still need to make tests pass, but I'm hoping that the fix can be basically as simple as this.

kevinbarabash:
@davidflanagan maybe add a screenshot test for this.  It should be possible using the `pre` and `post` properties in `ss_data.yaml` to construct a container with styles that would affect the rendering without your changes.  Something like:
```
SvgReset:
    pre: <div style="fill-opacity: 0.5; stroke-width: 3, ...">
    tex: \sqrt{x+1}
    post: </div>
```

davidflanagan:
That's a great idea, Kevin. I'll do it.

davidflanagan:
Actually, the pre and post attributes on the screenshot tests set the content of spans before and after the math, but they're both siblings of the math element, not ancestors, so I can't use them to set styles that would affect the math...

PR #1094, opened 2018-01-26
update fonts and metrics so that accents are positioned correctly

kevinbarabash:
This depends on https://github.com/KaTeX/katex-fonts/pull/5 which changed almost every combining glyph in the fonts to be non-combining.  The ones that are left don't appear to be used.  I'd like to remove unused glyphs from the fonts eventually but in a separate PR.


src/symbols.js
```diff
@@ -656,8 +656,8 @@ defineSymbol(text, main, textord, "\u00f8", "\\o", true);
 defineSymbol(text, main, textord, "\u00c6", "\\AE", true);
 defineSymbol(text, main, textord, "\u0152", "\\OE", true);
 defineSymbol(text, main, textord, "\u00d8", "\\O", true);
-defineSymbol(text, main, accent, "\u02ca", "\\'"); // acute
-defineSymbol(text, main, accent, "\u02cb", "\\`"); // grave
+defineSymbol(text, main, accent, "\u00b4", "\\'"); // acute
+defineSymbol(text, main, accent, "\u0060", "\\`"); // grave
```
kevinbarabash: It would be nice if all of these were in the `\u02c0` to `\u02cf` range.

src/symbols.js
```diff
@@ -666,7 +666,7 @@ defineSymbol(text, main, accent, "\u02d9", "\\."); // dot above
 defineSymbol(text, main, accent, "\u02da", "\\r"); // ring above
 defineSymbol(text, main, accent, "\u02c7", "\\v"); // caron
 defineSymbol(text, main, accent, "\u00a8", '\\"'); // diaresis
```
kevinbarabash: Same.

kevinbarabash:
It looks like I broke `\degree` a bit.  I'll try to fix it after lunch.

test/screenshotter/ss_data.yaml
```diff
@@ -16,8 +16,10 @@
 Accents: \vec{A}\vec{x}\vec x^2\vec{x}_2^2\vec{A}^2\vec{xA}^2
 AccentsText: |
     \begin{array}{l}
-    \text{\H{e}} & \text{\'\i} \\
-    \text{\H{X}} & \text{\"\i} \\
+    \text{\'\i} & \text{\.\i} & \text{\`\i} & \text{\"\i} & \text{\H\i} & \text{\r\i} \\
+    \text{\'\j} & \text{\.\j} & \text{\`\j} & \text{\"\j} & \text{\H\j} & \text{\r\j} \\
+    \text{\'a} & \text{\.a} & \text{\`a} & \text{\"a} & \text{\H{a}} & \text{\r{a}} \\
+    \text{\'A} & \text{\.A} & \text{\`A} & \text{\"A} & \text{\H{A}} & \text{\r{A}} \\
```
davidflanagan: KA has specific bug reports from translators about e with a double acute accent and turkish dotless i and dotted captial I, so if you're willing to add those here as well that would be great.

test/__snapshots__/mathml-spec.js.snap
```diff
@@ -42,7 +42,7 @@ exports[`A MathML builder accents turn into <mover accent="true"> in MathML 1`]
           e
         </mi>
         <mo>
-          ´
```
kevinbarabash: This changed b/c we switched some of the accents to be in the "Spacing Modifier Characters" Unicode block.  See https://en.wikipedia.org/wiki/Spacing_Modifier_Letters.

src/fontMetricsData.js
```diff
@@ -484,13 +483,18 @@ export default {
         "126": [0.35, 0.34444, 0, 0, 0.575],
         "168": [0, 0.69444, 0, 0, 0.575],
         "172": [0, 0.44444, 0, 0, 0.76666],
-        "175": [0, 0.59611, 0, 0, 0.575],
         "176": [0, 0.69444, 0, 0, 0.86944],
         "177": [0.13333, 0.63333, 0, 0, 0.89444],
-        "180": [0, 0.69444, 0, 0, 0.575],
+        "198": [0, 0.68611, 0, 0, 1.04166],
```
davidflanagan: feature request for the future: it would be nice if the script that generates this file included comments up at the top saying that it is an autogenerated file and explaining how to regenerate it.

test/screenshotter/ss_data.yaml
```diff
@@ -16,8 +16,11 @@
...
+    \text{\.I} & \text{\H e} & \text{\i ı}
```
kevinbarabash: @davidflanagan would you like me to change `\text{\H e}` to `\text{\H ee̋}`?

davidflanagan: That would be great, if you don't mind. And maybe add an explicit dotted capital I as well to make the whole row the same. Thanks!

src/fontMetricsData.js
kevinbarabash: Good call, I've opened an issue for this.

kevinbarabash:
> Maybe file an issue to write some documentation about how this is all handled though.

I've opened an issue for this.

kevinbarabash:
> Also, src/domTree.js has some special cases for forms of lowercase letter i. I'm not sure why those are there, but they also use the combining form of the accents rather than the non-combining form. Please check that code still makes sense with this change before merging.

I had a look at the PR for that and it sounds like I wasn't doing the right thing at the time anyways.  Also, we don't allow any other unicode chars outside of text mode.  I'm okay with removing that behavior. We'll want to mention it in the release notes as a breaking change.

ronkok:
@kevinbarabash What does this do regarding `\vec`? Is there now a non-combining character available to use for `\vec`, or should we keep the current SVG arrangement?

kevinbarabash:
@ronkok I haven't fixed the `\vec` glyphs yet.  Probably best to do that in a separate PR b/c we'd want to strip out the SVG code for `\vec`.

ronkok:
@kevinbarabash Thanks for the update. I will take no action.

Issue #1093, opened 2018-01-26
Non-portable use of String method `.includes`


ccorn:
KaTeX v0.9.0-beta chokes on `\mathbb{C}` (and `\mathbb{R}` so it's probably the macro), telling me "undefined not callable".
KaTeX v0.8.3 and v0.9.0-alpha2 render that without problems.
Found using `katex.renderToString` with Duktape.

kevinbarabash:
Hmm... `\mathbb{C}` works in the test page on `master`, but that uses `katex.render`.  I wonder if there's a difference between that and `katex.renderToString`.

kevinbarabash:
I tried `katex.renderToString("\\mathbb{C}")` on `master` and that works as well.

kevinbarabash:
I checked `v0.9.0-beta` as well.  What's Duktape?  Could you post a more complete code example?

ccorn:
Hmm, when I try it with less intervening Ruby layers, it works. Sorry, that's not your bug, even though it manifests at the Javascript execution level. Have to track it down myself.

ccorn:
[Use of the string method `includes`](https://github.com/Khan/KaTeX/search?utf8=%E2%9C%93&q=is%3Afile+fontFamily.includes&type=) seems to be the culprit. Using the polyfill from [Mozilla's documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/includes) makes the error go away.

ccorn:
I'd propose to use `.indexOf(...) !== -1` instead, for compatibility.

kevinbarabash:
Nice find!

kevinbarabash:
We'll definitely want to fix this before the next release.

kevinbarabash:
@ccorn if you have time, would definitely accept a PR to fix this.

ccorn:
I'd like to do that, but I am not a Javascript expert, and I am not familiar with the build system around KaTeX. I can fork the repo, do a minimal `.indexOf` replacement, but then I would not know how to test that change. I could launch a PR and hope that some CI kicks in and that everything turns out well, but if not, then anyone familiar with KaTeX development should better take over. Also, compatibility issues such as this one are bound to happen again, and the preferred strategy might be different from fixing the instances as they are discovered.

ccorn:
Hmm, there is a Makefile. Known territory. OK, I'll ssh into a server that has Node available, and then try and clone and mess around a bit.

PR #1097, opened 2018-01-27
Remove non-portable use of String method .includes

ccorn:
This is a minimal fix for #1093 which apparently has been introduced with #1009.

Notes:
* I have checked (with `grep`) that there are no other uses of `.includes`.
* A `>= 0` would probably look more natural, but [Mozilla's polyfill](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/includes#Polyfill) uses `!== -1`, and I just follow their lead.
* No tests included and no countermeasures taken against future incidents like this one. I am not familiar enough with this project yet.

### DIFF
--- a/src/buildCommon.js
+++ b/src/buildCommon.js
@@ -244,7 +244,7 @@ const makeOrd = function(
             const fontData = mathit(value, mode, options, classes);
             fontName = fontData.fontName;
             fontClasses = [fontData.fontClass];
-        } else if (fontFamily.includes("math") || mode === "math") {
+        } else if (fontFamily.indexOf("math") !== -1 || mode === "math") {
             // To support old font functions (i.e. \rm \sf etc.) or math mode.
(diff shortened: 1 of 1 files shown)


ccorn:
[clabot:check]

kevinbarabash:
> No tests included and no countermeasures taken against future incidents like this one. I am not familiar enough with this project yet.

I think that's fair enough.  My plan was to use https://github.com/kevinbarabash/eslint-plugin-luddite to lint against things that we don't want to use I just never got around to it.

kevinbarabash:
@ccorn thanks for the PR.

ccorn:
Indeed, [that](https://github.com/kevinbarabash/eslint-plugin-luddite/blob/master/lib/rules/no-modern-string-methods.js#L17) looks like it could have caught the issue.

PR #1100, opened 2018-01-27
Enable hot module replacement(HMR)

### LINKED ISSUE #929: webpackDevServer does not reload CSS or main.js

ylemkimon:
Fixes #929 and an issue mentioned in #805.

Enables hot module replacement(HMR) in webpack-dev-server(`npm start`).

It updates the codes(js and css) and re-renders the math without refreshing the page and reloading the whole module.


webpack.dev.js
```diff
@@ -3,10 +3,13 @@ const { targets, createConfig } = require('./webpack.common');
 const path = require('path');
 const PORT = 7936;
 
+const katexTarget = targets.shift();
+katexTarget.entry = './static/main.js';
+
```
ylemkimon: It would be nice to have a single entry point, but it would increase the unminified distribution file(`katex.js`) by 2-3 KB.

static/main.js
```diff
@@ -67,13 +60,21 @@ function init() {
...
             }
         }
     }
+
+    if (module.hot) {
+        module.hot.accept('../katex.webpack.js', reprocess);
+    }
+}
+
+if (/^\/(?:index.html)?$/.test(location.pathname)) {
```
ylemkimon: I wonder there is a better way to load this code and enable HMR only in the main page(`static/index.html`).

kevinbarabash:
> It updates the codes(js and css) and re-renders the math without refreshing the page and reloading the whole module.

Awesome!  There have been a number of times during testing where I had changed the default test string without clicking the `permalink` button and then changed code only to have the test string revert to the default.

webpack.dev.js
ylemkimon: @kevinbarabash Yes, but then webpack will rebuild both bundles when KaTeX sources have been changed. I'm not sure how it will impact the performance.

kevinbarabash: `main.js` is pretty small so I can't see there being a huge perf impact when it comes to HMR.

webpack.common.js
```diff
@@ -99,6 +99,7 @@ function createConfig(target /*: Target */, dev /*: boolean */,
             new webpack.EnvironmentPlugin({
                 NODE_ENV: dev ? 'development' : 'production',
             }),
+            dev && new webpack.NamedModulesPlugin(),
```
kevinbarabash: cool

static/main.js
```diff
@@ -67,13 +65,19 @@ function init() {
...
     }
+
+    if (module.hot) {
+        module.hot.accept('../katex.webpack.js', reprocess);
+    }
 }
 
 init();
+
+export default katex;
```
kevinbarabash: Now that we have separate entry points we should be able to get rid of this.

ylemkimon:
@kevinbarabash 
It seems that I cannot get `CommonsChunkPlugin` to work with HMR.

So the separate entry point/bundle has to include the KaTeX module, which increases build time <1s and memory usage around 10 MBs.

Another option is to call `reprocess` in the KaTeX entry point(`katex.webpack.js`) and it only would increase unminified distribution file(`dist/katex.js`) by few bytes, but we have to have `reprocess` as global and it will not support HMR of `main.js` and `main.css`.

kevinbarabash:
Having the the separate entry point/bundle is what we had in `7554153`, right?  I'm okay with that.  It's probably worth documenting somewhere why we had to do it that way.

kevinbarabash:
> It seems that I cannot get CommonsChunkPlugin to work with HMR.

For some reason I though `CommonsChunkPlugin` was still in this PR.  I see now that it isn't.

static/main.js
kevinbarabash: @ylemkimon I guess we still need this given the reasons you gave in the main thread.

ylemkimon: @kevinbarabash Yes, in order to export KaTeX and make it accessible from console and other scripts.

ylemkimon:
@kevinbarabash Thank you for the review and merge.

I’ll try to find a way to avoid creating two big chunks.

PR #1101, opened 2018-01-27
add Main-BoldItalic font to allow nesting of \textit and \textbf

kevinbarabash:


kevinbarabash:
> Glad it was pretty straight-forward to implement

Me too.  You had already done all the heavy lifting. :)

PR #1103, opened 2018-01-27
Improve JS spacing

### LINKED ISSUE #1102: Fix spacing code for "tight" styles
And add tests.

ylemkimon:
* Remove dummy spans for spacing around `\href`: Spacing around `\href`(`anchor`) is handled in the `html.buildExpression`, and currently spacings around `\href` are duplicated
* Make bin cancellation aware of children of fragment and anchor
* ~Fix tight spacing not applied: #1102~ Fixed by #1106 
* Add surrounding argument to html.buildExpression: it is an array consisting type of nodes that will be added to the left and the right, and if given, will be used to determine bin cancellation and spacings of outermost nodes. See `delimsizing.js` for its usage.

**TODO:**
- [ ] spacings inside `\href`: #1108
- [ ] add/update tests and screenshots


src/functions/delimsizing.js
```diff
@@ -161,7 +159,8 @@ defineFunction({
     },
     htmlBuilder: (group, options) => {
         // Build the inner expression
-        const inner = html.buildExpression(group.value.body, options, true);
+        const inner = html.buildExpression(group.value.body, options, true,
+            [null, "mclose"]);
```
kevinbarabash: This seems to be the on place where the `surrounding` param is used.  Do you for other situations where we'll need the first value instead of just the second?

src/buildHTML.js
```diff
@@ -79,14 +78,25 @@ export const buildExpression = function(expression, options, isRealGroup) {
     const nonSpaces =
         rawGroups.filter(group => group && group.classes[0] !== "mspace");
 
+    // add dummy spans for determining spacings between surrounding atoms
+    nonSpaces.unshift(surrounding[0] &&
+        new makeSpan([surrounding[0]], [], options));
+    nonSpaces.push(surrounding[1] &&
+        new makeSpan([surrounding[1]], [], options));
```
kevinbarabash: Is it necessary to create the spans if `surrounding[0]` or `surrounding[1]` is `null` or `undefined`?

src/functions/delimsizing.js
```diff
@@ -210,18 +209,6 @@ defineFunction({
...
-        const lastChildType = html.getTypeOfDomTree(inner[inner.length - 1]);
-        const activeSpacings = options.style.isTight() ? tightSpacings : spacings;
-
-        if (lastChildType && activeSpacings[lastChildType]["mclose"]) {
-            const glue =
-                buildCommon.makeSpan(["mord", "rule"], [], options);
-            const dimension =
-                calculateSize(activeSpacings[lastChildType]["mclose"], options);
-            glue.style.marginRight = `${dimension}em`;
-            inner.push(glue);
```
kevinbarabash: Now this isn't included in `inner`/`children`.  Won't that affect the results of `getOutermostNode`?

src/functions/href.js
```diff
@@ -31,40 +31,7 @@ defineFunction({
...
-                const anc = buildCommon.makeAnchor(href, [], elements, options);
-                return new buildCommon.makeFragment([
-                    new buildCommon.makeSpan([first], [], options),
-                    anc,
-                    new buildCommon.makeSpan([last], [], options),
-                ]);
-            }
-        }
-        return new buildCommon.makeAnchor(href, classes, elements, options);
+        return new buildCommon.makeAnchor(href, [], elements, options);
```
kevinbarabash: Nice!

src/buildHTML.js
```diff
@@ -79,14 +78,25 @@ export const buildExpression = function(expression, options, isRealGroup) {
...
+        const left = getOutermostNode(nonSpaces[i], "left");
+        if (left.classes[0] === "mbin" &&
+                isBinLeftCanceller(nonSpaces[i - 1], isRealGroup)) {
+            left.classes[0] = "mord";
+        }
+
+        const right = getOutermostNode(nonSpaces[i], "right");
+        if (right.classes[0] === "mbin" &&
+                isBinRightCanceller(nonSpaces[i + 1], isRealGroup)) {
+            right.classes[0] = "mord";
```
kevinbarabash: Nice!

src/buildHTML.js
ylemkimon: @kevinbarabash It's for easy manipulation of `nonSpaces`. This way, we can safely assume that `nonSpaces[0]` and `nonSpaces[nonSpaces.length -1]` are dummy spans.

src/functions/delimsizing.js
ylemkimon: @kevinbarabash Not yet, but I think it makes sense to have for both sides.

src/functions/delimsizing.js
ylemkimon: @kevinbarabash The `html.buildExpression` inserts the appropriate spacing between the last child and `mclose`.

kevinbarabash:
@ylemkimon are you still planning to do the TODOs in the top comment?  Did you want me to comment on the code as is before adding tests/updating screenshots?

kevinbarabash:
whoops... wrong PR

ylemkimon:
@kevinbarabash Yes, and if you don’t mind, yes, please. Though, I’m not sure what is best way to add spacings recursively for \href and \class in #1074.

src/buildHTML.js
kevinbarabash: I would rewrite this as:
```
const nonSpaces = [
   surrounding[0] && makeSpan([surrounding[0]], [], options)),
   rawGroups.filter(group => group && group.classes[0] !== "mspace"),
   surrounding[1] && makeSpan([surrounding[1]], [], options)),
];
```
This avoids the `unshift` which could be costly depending on the length of the array and better shows that the `surrounding` array is being used to surround something.

Note: `new` isn't required for `makeSpan`.

kevinbarabash: I can't wait until we have everything converted to `flow` so that we can better enforce some of these assumptions.

src/buildHTML.js
```diff
@@ -58,9 +55,11 @@ const styleMap = {
  * nodes. documentFragments are flattened into their contents, so the
  * returned list contains no fragments. `isRealGroup` is true if `expression`
  * is a real group (no atoms will be added on either side), as opposed to
- * a partial group (e.g. one created by \color).
+ * a partial group (e.g. one created by \color). `surrounding` is an array
+ * consisting type of nodes that will be added to the left and right.
  */
-export const buildExpression = function(expression, options, isRealGroup) {
+export const buildExpression = function(expression, options, isRealGroup,
+        surrounding = []) {
```
kevinbarabash: Maybe make this `surrounding = [null, null]`.  I think eventually we'll want to enforce that this is a 2-tuple with `flow`.

src/functions/delimsizing.js
kevinbarabash: That's pretty cool.  Glad you were able to figure out a way to get all the spacing code in one place.

kevinbarabash:
I tried checking the checkboxes but apparently you can't do that to checkboxes inside review summaries.

kevinbarabash:
@ylemkimon thanks for the PR.

PR #1106, opened 2018-01-28
use correct spacing with tight styles

### LINKED ISSUE #1102: Fix spacing code for "tight" styles

kevinbarabash:
Fixes #1102.  I removed some dead code at the same time.


kevinbarabash:
spacing

The bottom rows are off by quite a bit b/c TeX uses fonts that are stretched horizontally for script and scriptscript styles, but the relative spacing looks correct.  KaTeX is red and LaTeX is in green.

kevinbarabash:
@ylemkimon thanks for reviewing.

Issue #1038, opened 2017-12-26
Nested Math with Non Default Text results in an Error


rrandallcainc:
Example: `\textsf{Hello $a+b$}` results in an error.

Expected result: `\textsf{Hello $a+b$}` should render the `a+b` with the default Math font.

kevinbarabash:
@rrandallcainc it should use the default math style.  pdflatex renders the following:
<img width="453" alt="screen shot 2017-12-26 at 5 24 15 pm" src="https://user-images.githubusercontent.com/1044413/34366999-ba4cc9f4-ea61-11e7-9a1e-789b57922d59.png">

The error shows up on `master`, but not on the demo page.  The error is:
```
TypeError: undefined is not an object (evaluating 'fontMap[fontFamily].fontName')
```

kevinbarabash:
I swapped out `\textbf` for `\textsf` in the screenshot above.

rrandallcainc:
Okay. I see the issue. I can implement a fix, but I'd be curious if there's a better way. Ideally I'd like to reset all font options when entering math mode. Could you think of a place where that could be implemented?

kevinbarabash:
@rrandallcainc you could probably add a `withMathMode` method to `Options` that resets everything.

edemaine:
Nice find. Be sure to add a test for this too!

rrandallcainc:
Just got back from a trip. I'll take a look at this soon.

rrandallcainc:
@kevinbarabash Starting to look @ this now. Agreed that `withMathMode()` is probably the most elegant way to go.

Trying to think...where's the best entry point for when we enter math mode?

My first thought is around: https://github.com/Khan/KaTeX/blob/master/src/Parser.js#L486. I'm guessing I could leverage `this.settings.options.withMathMode()`. But would that be considered hacky?

kevinbarabash:
@rrandallcainc maybe we could set a prop on the `styling` node that's created which could indicate to the HTML and MathML builders to call reset the fonts using `withMathMode()`.

PR #1111, opened 2018-01-28
Nested Math in Non-Default Text Fix

rrandallcainc:
Resolves: 

1. https://github.com/Khan/KaTeX/issues/1038
2. https://github.com/Khan/KaTeX/issues/1112


rrandallcainc:
One difference with math vs text font commands, is text commands change one aspect of the font (weight, shape, or family). Math commands change all aspects of the font (weight, shape, and family). I'm wondering, probably as a cleanup item, we'd want to leverage `options.withFont()` for math, and `options.withFontFamily/Shape/Weight` for text. Currently `options.withFontFamily` is being used for math fonts.

src/Options.js
```diff
@@ -192,6 +192,15 @@ class Options {
         });
     }
 
+    /**
+     * Create a new options object when switching from text to math mode.
+     */
+    withMathMode(): Options {
+        return this.extend({
+            fontFamily: 'mathit',
+        });
```
kevinbarabash: Should this also reset the fontShape and fontWeight?

src/buildMathML.js
```diff
@@ -259,7 +259,9 @@ groupTypes.styling = function(group, options) {
     };
 
     const newStyle = styleMap[group.value.style];
-    const newOptions = options.havingStyle(newStyle);
+    const newOptions = group.value.mathStart ?
+                       options.havingStyle(newStyle).withMathMode() :
```
kevinbarabash: I would've expected `options.withMathMode()` to be sufficient for this case.

src/Options.js
rrandallcainc: Good question. I actually thought the same thing until I broke the tests.

```
it("should render \\textsf{\\textbf{$\\mathrm{\\textsf{A}}$}} with the correct font", function() {
        const markup = katex.renderToString("\\textsf{\\textbf{$\\mathrm{\\textsf{A}}$}}");
        expect(markup).toContain("<span class=\"mord textsf textbf\">A</span>");
});
```

Basically when going from text mode, to math mode, then back to text mode, it should retain the font weight/shape from the first time it was in text mode. Hope that makes sense.

src/buildMathML.js
rrandallcainc: Not sure I follow this. `options.withMathMode()` instead of `options.havingStyle(newStyle)`?

src/Options.js
kevinbarabash: Should it not also retain the `fontFamily`?  What happens in this case?
```
\textsf{\textbf{$\textbf{A}$}}
```
It feels like we need to:
- store font styles separate for `math` and `text` modes
- store which mode we're currently in

rrandallcainc: Great catch. This means I'll probably resolve  #1112 in the process.

src/Options.js
```diff
@@ -64,7 +64,9 @@ class Options {
     size: number;
     textSize: number;
     phantom: boolean;
-    fontFamily: string | void;
+    font: string;
+    oldTextFont: boolean;
+    fontFamily: string;
     fontWeight: string;
     fontShape: string;
```
kevinbarabash: I was thinking we'd have properties like:
```
mathFont: string;
textFontFamily: string;
textFontWeight: string;
textFontShape: string;
mode: "text" | "math";
```
and then the API, would look something like:
```
.withMathFont(font: string);
.withTextFontFamily(family: string);
.withTextFontWeight(weight: string);
.withTextFontShape(shape: string);
```

rrandallcainc: Hmm, how would the old text fonts be handled? I.e. `\rm \it`

kevinbarabash: I forgot about `\rm \it`, but I think we could modify the suggested API in the following way:
```
.withTextFontFamily(family: string, isOldCommand: boolean = false);
.withTextFontWeight(weight: string, isOldCommand: boolean = false);
.withTextFontShape(shape: string, isOldCommand: boolean = false);
```

rrandallcainc: I don't think `withTextFontFamily/Weight/Shape` would make sense with old font commands. Old font commands effectively change all 3, not just one. Additionally you'd lose the state when you exit the old font mode (i.e. `\textbf{{\rm hello} world}`). They behave much more similar to the math commands, which is why I opted to leave the naming conventions generic.

kevinbarabash: `withTextFontFamily` would look something like:
```
withTextFontFamily(family: string, isOldCommand: boolean) {
   if (isOldCommand) {
       this.extend({
           fontFamily: fontFamily || this.fontFontFamily,
           fontWeight: "",
           fontStyle: "",
      });
   } else {
        this.extend({
            fontFamily: fontFamily || this.fontFamily,
        });
   }
};
```
As for the question of losing state, `{\rm hello}` creates a new `options` object when processing that group which shouldn't affect the `options` object for the `\textbf` group.

rrandallcainc: Hmm, I'll give this more thought either later today or tomorrow. Thanks for the quick review.

edemaine:
#1073 points out some inconsistencies (compared to LaTeX) in the font size when using `\llap`. Will this PR help with this? Or perhaps unrelated.

rrandallcainc:
@edemaine I don't believe this is related. This PR isn't concerned with the font sizing.

kevinbarabash:
@rrandallcainc friendly ping to see where things are at with this PR.

rrandallcainc:
@kevinbarabash I expect to get back to this early this week. Thanks for checking in.

rrandallcainc:
@kevinbarabash I apologize for the delay on this one. This should be ready for another look.

test/__snapshots__/katex-spec.js.snap
```diff
@@ -88,7 +88,7 @@ exports[`An implicit group parser within optional groups should work with \\colo
 ]
 `;
 
-exports[`An implicit group parser within optional groups should work with sizing commands: \\sqrt[\\small 3]{x} 1`] = `
+exports[`An implicit group parser within optional groups should work with old font functions: \\sqrt[\\tt 3]{x} 1`] = `
```
kevinbarabash: It looks like the order of these changed, but the parse trees look the same.

test/screenshotter/ss_data.yaml
```diff
@@ -309,7 +309,12 @@ TextStacked:
       \textsf{\textrm{\textbf{abc123}} \textbf{abc123} \textit{abc123}}\\
       \textit{abc123 \textbf{abc123} \textsf{abc123}}\\
   \end{matrix}
-TextWithMath: \text{for $a < b$ and $ c < d $}.
+TextWithMath: 
+  \begin{matrix}
+      \text{for $a < b$ and $ c < d $}. \\
+      \textsf{for $a < b$ and $ c < d $}. \\
+      \textsf{for $a < b \textbf{ and } c < d $}
```
kevinbarabash: It would be good to have some tests for the behavior of old fonts.

src/Options.js
```diff
@@ -193,29 +195,42 @@ class Options {
...
     /**
-     * Create a new options objects with the give font.
+     * Creates a new options object with the given math font or old text font.
+     * @type {[type]}
      */
-    withFontFamily(fontFamily: ?string): Options {
+    withFont(font: string): Options {
         return this.extend({
-            fontFamily: fontFamily || this.fontFamily,
+            font,
```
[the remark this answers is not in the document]
rrandallcainc: Correct. Also, we can't safely reset the values of fontFamily/weight/etc. 

Consider the following: `\textbf{$\mathsf{\textsf{a}$}`. This should render as a bold, sans-serif "a", but if we reset, we would get a regular sans-serif "a"

src/buildCommon.js
```diff
@@ -33,7 +33,7 @@ const mainitLetters = [
 const lookupSymbol = function(
     value: string,
     // TODO(#963): Use a union type for this.
-    fontFamily: string,
+    fontName: string,
```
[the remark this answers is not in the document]
rrandallcainc: A `fontFamily` applies to a group of fonts. For example, the SansSerif family can be: regular, bold, or italic. In order to properly lookup the symbol, we need a specific font, not a font family.

src/Options.js
```diff
@@ -42,9 +42,8 @@ export type OptionsData = {
     size?: number;
     textSize?: number;
     phantom?: boolean;
-    // TODO(#1009): Keep consistent with fontFamily/fontWeight. Ensure this has a
-    // string value.
-    fontFamily?: string | void;
+    font?: string;
+    fontFamily?: string;
```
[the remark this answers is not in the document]
rrandallcainc: Font families are a group of fonts, not a specific font. I'll add a comment.

test/__snapshots__/katex-spec.js.snap
rrandallcainc: I fixed a typo, assuming that's why?

test/screenshotter/ss_data.yaml
rrandallcainc: I'll update.

test/__snapshots__/katex-spec.js.snap
kevinbarabash: Sorry, I didn't mean to suggest that anything needed changing here.

rrandallcainc:
@kevinbarabash Updated.

Definitely agree that the `makeOrd` implementation could be cleaned up. I think it's current state was a side effect of the assumption that math/text fonts behaved similar. That being said, it looks like a lot of solid cleanup has been done around fonts recently (i.e. pulling out the oldFonts from the parser, separating the text functions, etc.). I don't think the cleanup for `makeOrd` should be too bad.

kevinbarabash:
@rrandallcainc glad to see this PR finally merged.  Thanks for seeing it through.

PR #1117, opened 2018-01-30
unicodeTextInMathMode setting

### LINKED ISSUE #1046: Unicode text characters in math mode
#992 added support for e.g. `é` in both math and text mode, mapping to `\acute{e}` and `\'e` respectively.   This felt natural to me... but I haven't been able to reproduce this behavior (specifically, the math-mode behavior) in LaTeX.  I'm compiling the following with `xelatex`, and only the text line is rendering:

```latex
\documentclass{article}
\usepackage{amsmath}
\usepackage{unicode-math}
\setmathfont{xits-math.otf}
\usepackage{xeCJK}
\begin{document}
é私はバナナです。

$é私はバナナです。$
\end{document}
````

I may be missing some packages though.  If this is really the state of XeLaTeX, should we remove support in KaTeX?  Alternatively, we could try to add support via the mythical `katex.sty`.

Related, I am wondering whether to add support for CJK in math mode, but rendering as text symbols.  This is how `\ae` works in LaTeX (with a warning), but doesn't seem to be how it works with CJK...

edemaine:
Fixes #1046
* When `unicodeTextInMathMode` is `true`, accented letters from
  `unicodeSymbols.js`, and CJK and other supported languages,
  get added support in math mode (as requested in #895).
* When `unicodeTextInMathMode` is `false`, all of these stop working in
  math mode, and are only supported in text mode (matching XeTeX behavior).
  Note that this is a backwards incompatibility with some 0.9.0 alpha/betas, but it brings us back into alignment with LaTeX/XeTeX, so seems worth doing (?).

I'm open to better/shorter names for the option name, or for an argument why the default should be true instead of false (more permissive?).


kevinbarabash:
@edemaine I thought that unicode chars currently don't work in math mode.

ylemkimon:
https://github.com/Khan/KaTeX/blob/4a87f38e1a9f74883723a0aebce2fb91f1464863/src/symbols.js#L722-L736
Should they be allowed in the math mode when `unicodeTextInMathMode == false`?

edemaine:
@kevinbarabash `é` and other accented Unicode characters currently work in math mode, and this PR breaks that (to match XeTeX behavior).

@ylemkimon Hmm, probably not... Really, they shouldn't be supported at all right now, because they're not in a font. But maybe we can try to predict the future. I don't think we'll ever support `Åå` in math mode, for example, because math mode just doesn't have that accent. So maybe we should just drop support for them now?

`Çç` and the other characters ought to be supported (by the font) sometime soon, but is blocked on https://github.com/KaTeX/katex-fonts/issues/2 and https://github.com/KaTeX/katex-fonts/issues/25 .  I tested in XeTeX, and `$ÇÐÞçðþ$` renders ð for some reason but none of the others.  So we could keep these in the text symbol list, but drop them from the math symbol list. I'd personally be a fan of just dropping them altogether from math support.  Or we could manually somehow add math support when `unicodeTextInMathMode === true`, I guess.  Votes?

kevinbarabash:
@edemaine got it.  I forgot whether we had already made that change or not.  I guess not.

ylemkimon:
I think `extraLatinMath` can be removed as it's in `unicodeSymbols.js` and #1125 was merged.

kevinbarabash:
I want to get https://github.com/Khan/KaTeX/pull/1143 into to 0.9.0 final first.  Then we can start 0.10.0-alpha with this.

edemaine:
@kevinbarabash That's fine, though personally I'd lean toward reducing the number of backward-incompatible changes (with non-alpha/beta changes).  I'll revise this PR today for consideration, but also fine to wait on it.

Personally I actually prefer the current KaTeX behavior of "just working" with Unicode.  It's a shame that LaTeX can't do that without a lot of work.  Maybe the default `unicodeTextInMathMode` should be `true`?  Or maybe we can change it to that once we can provide a `katex.sty` file that does accent mapping for text characters...

edemaine:
I removed the special handling of `Å` and `å`.  (This should really have been done in #1125. Sorry!)

I also realized (thanks to some additional tests) that ð was being handled twice -- once in the `\eth` definition and once in `extraLatin`.  I split this out (we still need an "extra" definition for ð for text mode, for now).

I think [the commit](https://github.com/Khan/KaTeX/pull/1117/commits/09a026774c5c739c6980e243786be6acb23ce6ef) for the above changes should be merged before 0.9.0 final, if possible, as they are bug fixes.  I can split this into a separate PR if you prefer.

This PR is also ready to go.  It now checks for `extraLatin` symbols and rejects them if they appear in math mode when `unicodeTextInMathMode` is `false`.  When we eventually remove `extraLatin`, we can remove this check.

kevinbarabash:
> I think the commit for the above changes should be merged before 0.9.0 final, if possible, as they are bug fixes. I can split this into a separate PR if you prefer.

Splitting it out in a separate PR would be the easiest b/c of the other breaking changes in this diff.

kevinbarabash:
@edemaine love the "next version" tag. ^_^

kevinbarabash:
I've release v0.9.0 so we'll be able to merge this after review.

PR #1119, opened 2018-01-31
extract font functions and implicit color function

kevinbarabash:


src/Parser.js
```diff
@@ -515,34 +504,6 @@ export default class Parser {
...
-        } else if (func in Parser.oldFontFuncs) {
-            const style = Parser.oldFontFuncs[func];
-            // If we see an old font function, parse out the implicit body
-            this.consumeSpaces();
-            const body = this.parseExpression(true, breakOnTokenText);
-            if (style.slice(0, 4) === 'text') {
-                return new ParseNode("text", {
-                    style: style,
-                    body: new ParseNode("ordgroup", body, this.mode),
-                }, this.mode);
```
kevinbarabash: This block was never executed b/c there were no values that started with `text` in `oldFontFuncs`.

src/functions/color.js
```diff
@@ -86,3 +87,33 @@ defineFunction({
...
+    props: {
+        numArgs: 1,
+        allowedInText: true,
+        greediness: 3,
+        argTypes: ["color"],
+    },
+    handler(context, args) {
+        const {parser, breakOnTokenText} = context;
+
+        const color = args[0];
```
kevinbarabash: This was unexpectedly nice, the parser has already parsed the arguments for us, yay!

src/functions/font.js
```diff
@@ -0,0 +1,92 @@
...
+    "\\it": "mathit",
+};
+
+// Old font changing functions
+defineFunction({
+    type: "font",
+    names: Object.keys(oldFontFuncsMap),
+    props: {
+        numArgs: 0,
+        allowedInText: true,
```
kevinbarabash: It feels like all implicit functions are allowed in text mode.  Does anyone know if that's actually the case?

ylemkimon: AFAIK, they are macros of `\fam = XX \tenXX`, and `\fam` has no effect in the text mode but allowed in it, so they are allowed in the text mode.

kevinbarabash: `\text{abc \bf def}` causes `def` to be bold which would seem to indicate that `\fam` does have an effect if `\bf` is defined using it.

PR #1120, opened 2018-01-31
move katex.less into src

kevinbarabash:


ylemkimon:
Following paths need to be updated:
https://github.com/Khan/KaTeX/blob/95d5c59d31a77c8d43d85fbc98bd26d4914ca5d3/package.json#L58
https://github.com/Khan/KaTeX/blob/95d5c59d31a77c8d43d85fbc98bd26d4914ca5d3/Makefile#L34

kevinbarabash:
@ylemkimon thank you for catching those paths again. :)

PR #1122, opened 2018-01-31
Set webpack-dev-server host to 0.0.0.0

### LINKED ISSUE #911: webpack server doesn't support remote connections
I just tried out the new `webpackDevServer.js` (via `npm start`) and there's been a regression in functionality compared to the old `server.js`: I can no longer connect to the server from the outside (via `http://hostname:7936`).  I can only connect from the same host.  This is annoying for me, as I like to develop on a remote Linux server and test from a web browser on my local Windows machine.

Is this intentional?  Is there some security hole that this would otherwise open up?  Or can we restore functionality?  [This post](https://medium.com/webpack/webpack-dev-server-middleware-security-issues-1489d950874a) seems relevant.  We can't reasonably provide the host name (via `--public`), as it depends on the individual user.  Is `disableHostCheck` safe enough for our usage?  We don't need to worry about leaking the source code...

@rrandallcainc

ylemkimon:
Fixes #911.

webpack-dev-server CLI has a default value of localhost, which only
allows accessing locally. Setting to 0.0.0.0 allows to access the server
from the external. However, the open option will open 0.0.0.0:7936,
which is an invalid address, so it has been removed.

CC @edemaine


kevinbarabash:
@edemaine go for it.

PR #1123, opened 2018-01-31
add support for unicode angle brackets, single vertical bar, double vertical bar

### LINKED ISSUE #243: Symbol unicode replacement doesn’t work

kevinbarabash:
This PR also allows these symbols to be used with left/right delimiters.  This should allow us to finally close #243.

TODO: write some tests


src/functions/delimsizing.js
```diff
@@ -38,8 +38,8 @@ const delimiters = [
     "(", ")", "[", "\\lbrack", "]", "\\rbrack",
     "\\{", "\\lbrace", "\\}", "\\rbrace",
     "\\lfloor", "\\rfloor", "\\lceil", "\\rceil",
-    "<", ">", "\\langle", "\\rangle", "\\lt", "\\gt",
-    "\\lvert", "\\rvert", "\\lVert", "\\rVert",
+    "<", ">", "\\langle", "\u27e8", "\\rangle", "\u27e9", "\\lt", "\\gt",
+    "\\lvert", "\\rvert", "\u2223", "\\lVert", "\\rVert", "\u2225",
```
ylemkimon: Just out of curiosity, is it valid to have right delimiters, e.g., `\rangle`, as a left delimiter?

kevinbarabash: I checked with quicklatex.com and it appears so although I'm not sure in what context you'd actually want to do that.

ronkok:
PR #993 and `unicode-math` both map U+2223 to `\mid`, not to `\lVert` or `\rVert`.  `\mid` is an operator, not a delimiter, useful for set builder notation.

The single character analogue to `\vert` is ASCII "|", U+007C. That's been true since the *TeXbook*, p156.

kevinbarabash:
> PR #993 and unicode-math both map U+2223 to \mid, not to \lVert or \rVert. \mid is an operator, not a delimiter, useful for set builder notation.

@ronkok thanks for catching this.  I'll remove the `\lvert`/`\rvert` changes from this diff.

ronkok:
Similarly, `unicode-math` maps U+2225 to `\parallel`, which is also an operator, not a delimiter.

test/katex-spec.js
```diff
@@ -1212,6 +1258,19 @@ describe("A left/right parser", function() {
     });
 });
 
+describe("left/right builder", () => {
+    const cases = [
+        ['\\left\\langle \\right\\rangle', '\\left< \\right>'],
+        ['\\left\\langle \\right\\rangle', '\\left\u27e8 \\right\u27e9'],
+    ];
```
kevinbarabash: I think this could be extend to something like:
```
const cases = {
    'shouldParse': [ ... ],
    'shouldBuild': [ ... ],
    'shouldNotParse': [ ... ],
    'shouldNotBuild': [ ... ],
    'shouldParseLike': [ ... ],
    'shouldBuildLike': [ ... ],
};
```
I'll open an issue for this.

Issue #1099, opened 2018-01-27
accents on characters inside \textit{} are incorrectly positioned


kevinbarabash:
<img width="1088" alt="screen shot 2018-01-27 at 12 59 58 am" src="https://user-images.githubusercontent.com/1044413/35469155-91448f4c-02fd-11e8-99ca-53960714ceef.png">

We need to take into account the skew (as well as the height of the character).

edemaine:
I've tracked the bug down to [these lines](https://github.com/Khan/KaTeX/blob/e4fe1c89d48585252e96b0f7d414b37d883aed70/src/buildCommon.js#L146-L160) of `buildCommon.mathDefault`

edemaine:
The bug seems to stem from the choice of font. `$A$` renders an A in font family KaTeX_Math style italic, while `$\textit{A}$` renders an A in font family KaTeX_Main style italic.  KaTeX correctly detects this, and uses the *differing* metrics already in `fontMetricsData.js`:

```js
    "Math-Italic": {
        "47": [0.19444, 0.69444, 0, 0, 0],
        "65": [0, 0.68333, 0, 0.13889, 0.75],
        ...
    }, ...
    "Main-Italic": {
        ...
        "65": [0, 0.68333, 0, 0, 0.74333],
        ...
    },
`

0.13889 is the correct skew.  0 is not.

I assume the choice of fonts is "correct" (though we could of course use the math font for a letter A if we wanted...), so the "real" problem is the font metrics.  @kevinbarabash, any idea why the font metrics for Main-Italic would be different from the font metrics for Math-Italic, when the symbols look the same?

kevinbarabash:
@edemaine not sure.  I'd have to see what we're doing in katex-fonts.  There is post-processing we do to some of the glyphs after extracting the paths from the source fonts.

PR #1125, opened 2018-02-01
Add support for \mathring

### LINKED ISSUE #1124: Support for \mathring
\mathring is a math mode accent defined in [amssymb](http://milde.users.sourceforge.net/LUCR/Math/mathpackages/amssymb-symbols.pdf).

It's basically the math mode equivalent of the text mode accent \r. I just needed it for a single letter (`$\mathring g$` for the round metric) so I tried to hack it up with `$\text{\r \textit{g}}$` but it doesn't look great:

screen shot 2018-01-31 at 12 17 27 am

compared to

screen shot 2018-01-31 at 12 19 52 am

I wonder if there's a better way to fake it?

edemaine:
Fix #1124. `\mathring` is the math-mode equivalent to the `\r` accent (thanks to @akalin for pointing it out!).

Support follows from a rather trivial symbol aliasing, plus (for Unicode support) listing it in `unicodeAccents.js`.

Screenshot of
```latex
\mathring A\mathring B\textit{\r A\r B}\text{\r A\r B}
```
shows that skew correction is fine, except somehow for `\textit` (as reported in #1099):
image

### DIFF
--- a/src/unicodeAccents.js
+++ b/src/unicodeAccents.js
@@ -13,6 +13,6 @@ module.exports = {
     '\u030c': {text: '\\v', math: '\\check'},
     '\u0302': {text: '\\^', math: '\\hat'},
     '\u0307': {text: '\\.', math: '\\dot'},
-    '\u030a': {text: '\\r'},
+    '\u030a': {text: '\\r', math: '\\mathring'},
     '\u030b': {text: '\\H'},
 };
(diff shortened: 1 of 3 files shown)


src/unicodeAccents.js
```diff
@@ -13,6 +13,6 @@ module.exports = {
     '\u030c': {text: '\\v', math: '\\check'},
     '\u0302': {text: '\\^', math: '\\hat'},
     '\u0307': {text: '\\.', math: '\\dot'},
-    '\u030a': {text: '\\r'},
+    '\u030a': {text: '\\r', math: '\\mathring'},
```
kevinbarabash: Cool!

Issue #1126, opened 2018-02-01
better organize and simplify tests


kevinbarabash:
Currently, test require writing verbose `it("should ...", () => { ... }` along with multiple `expect(...).toSomethingSomething(...)`.  Also, we separate parser tests from builder tests.  I would like suggest group all tests for a particular command together in the same place.  This will make it easier for people to get the whole picture of the behavior of particular commands.

I would also like to suggest a test case helper that accepts test cases in the following format:
```
const cases = {
   shouldParse: string[],
   shouldBuild: string[],
   shouldNotParse: string[],
   shouldNotBuild: string[],
   shouldParseLike: [string, string][],
   shouldBuildLike: [string, string][],
};

const generateTests(suiteName, cases) { ... }  // would generate `describe` and `it` calls
```
The output would be equivalent to
```
describe(suiteName, () => {
   it("should parse ...", () => expect(...).toParse());
   ...
   it("should build ...", () => expect(...).toBuild());
   ...
   it("should not parse ...", () => expect(...).toNotParse());
   ...
   // and so on
});
```

edemaine:
Related to #1031 (which should maybe be done first? or in any case we should build that general module).

danielhollas:
I'd like to give this a try. Is this still wanted?

I made a first working prototype but I have one question before the PR. I noticed in `katex-spec.js` that there are a lot of "custom" test that do not fit these patterns. If the aim is to bundle all tests for a given function together, I wonder whether something like this would not work better:

```js
describe("A function", function() {
   const cases = {...as above...}
   const generateTests(cases); // would generate it calls
   it("custom test1", ...)
   it("custom test2", ...)
});
```

kevinbarabash:
@danielhollas it's probably worth having some discussion about.  Given that you've put together a working prototype, maybe create a PR using that to replace a few existing tests and then people can see what it looks like in action but without you having to convert everything.  As for tests that don't follow this pattern, I think there's a category of tests that verify certain things about the structure of the output.  Are there others?

PR #1127, opened 2018-02-01
add cedilla

kevinbarabash:
screen shot 2018-01-31 at 9 58 45 pm

TODO:
- [x] update `makeUnicode.js` to prevent duplicate keys
- [x] update `Unicode` screenshot


metrics/mapping.pl
```diff
@@ -1189,50 +1197,6 @@
   ],
 };
 
-$map{cmss10} = {
```
kevinbarabash: I just noticed that this appears twice so I removed the one that didn't match makeFF.

src/unicodeMake.js
```diff
@@ -47,10 +52,14 @@ for (const letter of letters) {
...
-                    `    ${encode(normalized2)}: ${encode(combined2)},`
-                    + `  // ${normalized2} = ${accents[accent].text}`
-                    + `${accents[accent2].text}{${letter}}`);
+                const key = encode(normalized2);
+                if (!keys.has(key)) {
+                    keys.add(key);
+                    console.log(
+                        `    ${encode(normalized2)}: ${encode(combined2)},`
+                        + `  // ${normalized2} = ${accents[accent].text}`
+                        + `${accents[accent2].text}{${letter}}`);
```
kevinbarabash: The reason why there are duplicates is b/c `\c\'{c}` and `\'\c{c}` map to the same character.  Same goes for the uppercase version.

src/functions/accent.js
```diff
@@ -63,9 +63,11 @@ const htmlBuilder = (group, options) => {
...
 
     // calculate the amount of space between the body and the accent
-    const clearance = Math.min(
-        body.height,
-        options.fontMetrics().xHeight);
+    let clearance = group.value.label === "\\c"
+        ? body.height + body.depth
+        : Math.min(
+            body.height,
+            options.fontMetrics().xHeight);
```
edemaine: Is there a reason that this needs to be special-cased for `\c`?  Can we come up with a general rule?  Ah, I see, there's a fundamental difference according to whether the accent is above or below...  Maybe there should be an `const accentBelow = (group.value.label === "\\c")` which we'll later add more cases to?  Alternatively, `const accentBelow = (accent.depth > 0)` might work...

src/functions/accent.js
```diff
@@ -87,6 +89,9 @@ const htmlBuilder = (group, options) => {
             // shift the accent over to a place we don't want.
             accent.italic = 0;
             width = accent.width;
+            if (group.value.label === "\\c") {
+                clearance += accent.depth;
+            }
```
edemaine: I'm guessing `accent.depth` is zero for the other accents?  Still, might make sense to keep this wrapped in an `if` (possibly `if (accentBelow)`).

src/unicodeMake.js
edemaine: Thanks for fixing this!

I checked that there are still no 3-accent combos that resolve to a single character.

```js
            for (const accent3 of Object.getOwnPropertyNames(accents)) {
                if (accent === accent3 || accent2 === accent3) {
                    continue;
                }
                const combined3 = combined2 + accent3;
                const normalized3 = combined2.normalize('NFC');
                if (normalized3.length === 1) {
                    const key = encode(normalized3);
                    if (!keys.has(key)) {
                        keys.add(key);
                        console.log(
                            `    ${encode(normalized3)}: ${encode(combined3)},`
                            + `  // ${normalized3} = ${accents[accent].text}`
                            + `${accents[accent2].text}{${letter}}`
                            + `${accents[accent3].text}{${letter}}`);
                    }
                }
            }
```

src/unicodeMake.js
```diff
@@ -31,14 +31,19 @@ console.log("export default {");
...
+            const key = encode(normalized);
+            if (!keys.has(key)) {
+                keys.add(key);
+                console.log(
+                    `    ${encode(normalized)}: ${encode(combined)},`
+                    + `  // ${normalized} = ${accents[accent].text}{${letter}}`);
+            }
         }
         for (const accent2 of Object.getOwnPropertyNames(accents)) {
             if (accent === accent2) {
```
edemaine: I wonder if we should remove this `if (accent === accent2) { continue }` block.  Maybe there is an accent that can be doubled into a single character.  (I checked, and there isn't, but maybe better to have the check in the code.)

src/unicodeSymbols.js
```diff
@@ -35,23 +38,27 @@ export default {
...
     "\u0117": "\u0065\u0307",  // ė = \.{e}
+    "\u0229": "\u0065\u0327",  // ȩ = \c{e}
     "\u1e1f": "\u0066\u0307",  // ḟ = \.{f}
     "\u01f5": "\u0067\u0301",  // ǵ = \'{g}
     "\u1e21": "\u0067\u0304",  // ḡ = \={g}
     "\u011f": "\u0067\u0306",  // ğ = \u{g}
     "\u01e7": "\u0067\u030c",  // ǧ = \v{g}
     "\u011d": "\u0067\u0302",  // ĝ = \^{g}
     "\u0121": "\u0067\u0307",  // ġ = \.{g}
+    "\u0123": "\u0067\u0327",  // ģ = \c{g}
```
edemaine: This one is not going to render correctly (I assume); see @ylemkimon's [recent comment](https://github.com/KaTeX/katex-fonts/pull/26#issuecomment-362148045).  Perhaps we should fix this in another PR, though.  (It'll need another tweak to the font, I suspect.)

kevinbarabash:
@edemaine thanks for the review.  I'll try to make these changes over the weekend.

src/functions/accent.js
kevinbarabash: Good call.

src/unicodeSymbols.js
kevinbarabash: I'm not sure what the resolution is for `\c{g}`.  Let's handle it separately.

src/unicodeMake.js
kevinbarabash: Let's leave it since it isn't hurting anything.

kevinbarabash:
@edemaine this is ready for review again.

edemaine:
Sorry, missed the request to review this a second time. Can you resolve the conflicts first?

kevinbarabash:
I should've done that after merging that other diff.  Thanks.

src/fontMetrics.js
```diff
@@ -104,7 +104,6 @@ const extraCharacterMap = {
     'Ð': 'D',
     'Þ': 'o',
     'å': 'a',
-    'ç': 'c',
```
kevinbarabash: I should remove the uppercase one as well.

kevinbarabash: Done.

kevinbarabash:
@edemaine this is ready for review again.

test/unicode-spec.js
```diff
@@ -64,6 +77,32 @@ describe("unicode", function() {
...
+                        result.message = "'" + actual + "' failed to " +
+                            "build with error: " + e.message;
+                    } else {
+                        result.message = "'" + actual + "' failed " +
+                            "building with unknown error: " + e.message;
+                    }
+                }
+
+                return result;
+            },
```
edemaine: Sad to see all this code duplication. I guess more motivation to work on #1126! 😄

test/unicode-spec.js
```diff
@@ -72,9 +111,14 @@ describe("unicode", function() {
...
 
+    it("should build Latin-1 inside \\text{}", function() {
+        expect('\\text{ÀÁÂÃÄÅÈÉÊËÌÍÎÏÑÒÓÔÕÖÙÚÛÜÝàáâãäåèéêëìíîïñòóôõöùúûüýÿ' +
+            'ÆÇÐØÞßæçðøþ}').toBuild();
+    });
+
     it("should parse Latin-1 outside \\text{}", function() {
         expect('ÀÁÂÃÄÅÈÉÊËÌÍÎÏÑÒÓÔÕÖÙÚÛÜÝàáâãäåèéêëìíîïñòóôõöùúûüýÿ' +
-            'ÇÐÞçðþ').toParse();
+            'ÐÞðþ').toParse();
```
edemaine: This is backwards incompatibility: we used to (poorly) support `çÇ` in math mode. I'm not sure what's best here. At the least, removing support would mean a new major version. But we don't have good support to begin with, so maybe that's a good thing.

An alternative would be to define a macro like `\mathcedilla` (by analogy to `\mathring`) for math-mode cedilla and then we could support it in math mode. But I couldn't find any such macro in LaTeX.  We could provide one in `katex.sty` (#793) though...

test/unicode-spec.js
```diff
@@ -72,9 +111,14 @@ describe("unicode", function() {
             'ÆÇÐØÞßæçðøþ}').toParse();
     });
 
+    it("should build Latin-1 inside \\text{}", function() {
+        expect('\\text{ÀÁÂÃÄÅÈÉÊËÌÍÎÏÑÒÓÔÕÖÙÚÛÜÝàáâãäåèéêëìíîïñòóôõöùúûüýÿ' +
+            'ÆÇÐØÞßæçðøþ}').toBuild();
+    });
+
```
edemaine: This seems to be a duplicate of the above -- perhaps a merge typo?

test/unicode-spec.js
kevinbarabash: We're dropping support for `çÇ` in math mode in #1117.  I rebase after it gets merged.

test/unicode-spec.js
kevinbarabash: This is uses `.toBuild()` whereas the one above uses `.toParse()`.  https://github.com/Khan/KaTeX/issues/1126 should address this.

test/unicode-spec.js
```diff
@@ -81,7 +124,7 @@ describe("unicode", function() {
 
     it("should parse Latin-1 outside \\text{}", function() {
         expect('ÀÁÂÃÄÅÈÉÊËÌÍÎÏÑÒÓÔÕÖÙÚÛÜÝàáâãäåèéêëìíîïñòóôõöùúûüýÿ' +
-            'ÇÐÞçðþ').toParse({unicodeTextInMathMode: true});
+            'ÆÇÐØÞßæçðøþ).toParse({unicodeTextInMathMode: true});
```
kevinbarabash: bad merge :\

test/unicode-spec.js
```diff
@@ -81,7 +125,7 @@ describe("unicode", function() {
 
     it("should parse Latin-1 outside \\text{}", function() {
         expect('ÀÁÂÃÄÅÈÉÊËÌÍÎÏÑÒÓÔÕÖÙÚÛÜÝàáâãäåèéêëìíîïñòóôõöùúûüýÿ' +
-            'ÇÐÞçðþ').toParse({unicodeTextInMathMode: true});
+            'ÐÞðþ').toParse({unicodeTextInMathMode: true});
```
kevinbarabash: This test should pass with `ÆÇÐØÞßæçðøþ`.  I'm going to see if I can fix that.

src/symbols.js
```diff
@@ -739,7 +740,8 @@ for (let i = 0; i < letters.length; i++) {
 // but they are not actually in the font, nor are they supported by the
 // Unicode accent mechanism, so they fall back to Times font and look ugly.
 // TODO(edemaine): Fix this.
-export const extraLatin = "ÇÐÞçþ";
+export const extraLatin = "ÐÞþ";
+
```
edemaine: You'll need to add the cedilla c's to the math symbol table similar to the "ð" below, unless we add a math version of the accent.

edemaine: Somehow I thought we merged this already, but it looks like the above is still an outstanding issue.  We have two options:
1. Add a `\mathcedilla` or other nonstandard operator for a math version of the cedilla accent, and add support for `ç` etc. in math mode.
2. Add the following lines next to the similar lines for "ð":
```js
 defineSymbol(text, main, mathord, "ç", "ç");
 defineSymbol(text, main, mathord, "Ç", "Ç");
```

The former approach will give the proper glyphs; the latter approach will maintain backward compatibility by forcing çÇ to still fall back to system fonts.

kevinbarabash:
I'm going to start a fresh PR for this.

Issue #1121, opened 2018-01-31
remove Makefile


kevinbarabash:
Many of the existing targets have already been migrated to package.json.  The release.sh script will probably need a little tweaking.

edemaine:
We can change `make unicode` (I think the only Makefile rule that I added) into `npm run unicode` (i.e. a script) or something similar.

PR #1134, opened 2018-02-02
move metrics to katex-fonts submodule

kevinbarabash:
It makes more sense to have metrics generation in the same repo as font generation as both things must be done at the same time.  See https://github.com/KaTeX/katex-fonts/pull/27.


ylemkimon:
I think we can remove `make metrics`(and eventually remove Makefile, #1121) or call `buildMetrics.sh` in:
https://github.com/Khan/KaTeX/blob/7b635f79cfc28c46dcbf814d47f6977e6712797b/Makefile#L84-L85

kevinbarabash:
#1135 will take care of the `metrics` make target.

kevinbarabash:
This actually needs updating b/c I closed https://github.com/KaTeX/katex-fonts/pull/27 in preference of https://github.com/KaTeX/katex-fonts/pull/30.

PR #1139, opened 2018-02-03
Enable spacing functions in text mode

ronkok:
Spacing functions such as `\;` or `\quad` should be accepted in either math mode or text mode. This PR will also fix a bug in `\llap` and `\rlap` and enable them to accept spacing functions.


ronkok:
Done.

PR #1136, opened 2018-02-02
Add 0.08em padding above \sqrt

ronkok:
This way, any browser imprecision on the parent span.height will not encroach on the ink of the \sqrt top line.

PR #1131, opened 2018-02-01
Change frac-line from SVG line to SVG path

ronkok:
SVG paths have given us more predictable behavior than SVG lines.

PR #1143, opened 2018-02-05
Padding over \sqrt and Paths for frac-line

ronkok:
This replaces two earlier PRs, #1131 Change frac-line from SVG line to SVG path, and #1136 Add 0.08em padding above \sqrt line.

It was easier to create a substitute branch than it was to repair my messed up history.


src/stretchy.js
```diff
@@ -146,16 +143,6 @@ const katexImagesData: {
...
-
-    // The next three arrows are from the mhchem package.
-    // In mhchem.sty, min-length is 2.0em. But these arrows might appear in the
-    // document as \xrightarrow or \xrightleftharpoons. Those have
-    // min-length = 1.75em, so we set min-length on these next three to match.
-    xrightleftarrows: [["baraboveleftarrow", "rightarrowabovebar"], 1.75, 667],
-    xrightequilibrium: [["baraboveshortleftharpoon",
-        "rightharpoonaboveshortbar"], 1.75, 716],
-    xleftequilibrium: [["shortbaraboveleftharpoon",
-        "shortrightharpoonabovebar"], 1.75, 716],
```
kevinbarabash: I don't think these should've been removed.  It breaks the `ReactionArrows` test.

src/stretchy.js
```diff
@@ -47,9 +47,6 @@ const stretchyCodePoint: {[string]: string} = {
     xtwoheadrightarrow: "\u21a0",
     xlongequal: "=",
     xtofrom: "\u21c4",
-    xrightleftarrows: "\u21c4",
-    xrightequilibrium: "\u21cc",  // Not a perfect match.
-    xleftequilibrium: "\u21cb",   // None better available.
```
kevinbarabash: These probably shouldn't be removed.

ronkok:
@kevinbarabash Thank you. In my frustration, I got hasty.

kevinbarabash:
I've regenerate the screenshots in https://github.com/ronkok/KaTeX/pull/1.  It's incredible how many of our screenshots involve fractions.  Many of them probably don't need fractions to test what they're testing.

ronkok:
>  It's incredible how many of our screenshots involve fractions. 

I know that I contributed to that. When I was testing stretchy arrows, fractions were a cheap way to test vertical alignment and vertical kerns used for clearances.

kevinbarabash:
> When I was testing stretchy arrows, fractions were a cheap way to test vertical alignment and vertical kerns used for clearances.

I guess you need some way to test clearance on top.  It's either `\frac` or `\sqrt` I guess.

src/delimiter.js
```diff
@@ -354,38 +354,48 @@ const makeSqrtImage = function(height, options) {
...
+    // We create viewBoxes with 80 units of "padding" above each surd.
+    // Then browser rounding error on the parent span height will not
+    // encroach on the ink of the viniculum. But that padding is not
+    // included in the TeX-like `height` used for calculation of
+    // vertical alignment. So texHeight = span.height < span.style.height.
 
     if (delim.type === "small") {
         // Get an SVG that is derived from glyph U+221A in font KaTeX-Main.
-        viewBoxHeight = 1000;  // from font
+        viewBoxHeight = 1080;  // 1000 unit glyph height + 80 unit padding.
```
kevinbarabash: I think it be good to extract some of these values into constants, e.g.
```
const emBoxHeight = 1000;
const delimiterPadding = 80;

File: src/delimiter.js [hunk 4]
```diff
@@ -326,10 +326,10 @@ const sqrtSvg = function(sqrtName, height, viewBoxHeight, options) {
         // to a point near (14 units) the bottom of the surd. The viniculum
         // is 40 units thick. So the length of the line in question is:
         const vertSegment = viewBoxHeight - 54;
-        alternate = `M702 0H400000v40H742v${vertSegment}l-4 4-4 4c-.667.667
+        alternate = `M702 80H400000v40H742v${vertSegment}l-4 4-4 4c-.667.667
 -2 1.5-4 2.5s-4.167 1.833-6.5 2.5-5.5 1-9.5 1h-12l-28-84c-16.667-52-96.667
 -294.333-240-727l-212 -643 -85 170c-4-3.333-8.333-7.667-13 -13l-13-13l77-155
- 77-156c66 199.333 139 419.667 219 661 l218 661zM702 0H400000v40H742z`;
+ 77-156c66 199.333 139 419.667 219 661 l218 661zM702 80H400000v40H742z`;
```
kevinbarabash: If we have `const delimiterPadding = 80;` replace the `80` in the this string with `${delimiterPadding}`... or at the very least add a component stating what the 80 is so that if people need to change the padding in the future it's easy to find all the places that need updating.

src/delimiter.js
```diff
@@ -354,38 +354,48 @@ const makeSqrtImage = function(height, options) {
...
 
     if (delim.type === "small") {
         // Get an SVG that is derived from glyph U+221A in font KaTeX-Main.
-        viewBoxHeight = 1000;  // from font
+        viewBoxHeight = 1080;  // 1000 unit glyph height + 80 unit padding.
         const newOptions = options.havingBaseStyle(delim.style);
         sizeMultiplier = newOptions.sizeMultiplier / options.sizeMultiplier;
-        spanHeight = 1 * sizeMultiplier;
+        spanHeight = 1.08 * sizeMultiplier;
+        texHeight = 1.00 * sizeMultiplier;
```
kevinbarabash: It might be useful to extract a constant for `0.08` as well.

src/stretchy.js
```diff
@@ -342,50 +342,43 @@ const encloseSpan = function(
...
-        }
+    let parentClass = "stretchy";  // default
 
-        svgNode = new domTree.svgNode(lines, {
-            "width": "0.25em",
-            "height": "100%",
-            "viewBox": "0 0 10 10",
-            "preserveAspectRatio": "none",
+    if (className === "vertical-separator") {
+        path = new domTree.pathNode("vertSeparator");
```
kevinbarabash: Cool.  Thanks for updating vertical separators to be consistent with the horizontal ones.

src/stretchy.js
```diff
@@ -342,50 +342,43 @@ const encloseSpan = function(
...
+        // take up the middle fifth of the viewBox and span. If the context is
+        // normalsize/textstyle then the line will be 0.04em and the usual
+        // 1000:1 ratio holds. But if the context is scriptstyle, then
+        // lineThickness > 0.04em and we have a ratio somewhat different than
+        // 1000:1.
+
+        path = new domTree.pathNode("stdHorizRule");
+        svgNode = new domTree.svgNode([path], {
+            "width": "400em",
             "height": 5 * lineThickness + "em",
```
kevinbarabash: It's interesting that the line thickness changes for horizontal rules but not for vertical separators.  Is that just a LaTeX quirk?

src/katex.less
```diff
@@ -475,6 +477,8 @@
     .stretchy {
         width: 100%;
         display: block;
+        position: relative;
+        overflow: hidden;
```
kevinbarabash: Is this also b/c of a Chrome rendering issue?

src/svgGeometry.js
```diff
@@ -5,45 +5,57 @@
  */
 
 const path: {[string]: string} = {
+    // stdHorizRule is used for frac-lines, underlines, and overlines.
+    // It is 0.04em thick if the line comes from normalsize/textstyle.
+    // It is in a viewBox that is 5x as tall as the line, so that the
+    // full line will be rendered even if the browser rounds down
+    // the enclosing span size.
+    stdHorizRule: "M0 120 V80 H400000 v40 H0 z M0 120 V80 H400000 v40 H0 z",
```
kevinbarabash: I assume all uses of `80` here for padding similar to what's in delimiter.js.  Maybe we should have a constant for these so that it's easy to update if necessary.

src/delimiter.js
ronkok: Will do.

src/stretchy.js
ronkok: The vertical line constant value is what I found in the code when I changed it from a span border to an SVG path. I did not go back to underlying sources to check if that was correct.

src/katex.less
ronkok: No. Those two new lines are only necessary for IE. In earlier work, I had put those two lines of CSS into class `hide-tail` and then nested `hide-tail` inside a `stretchy` span. I found here that I could save one nesting level by writing those two new lines into `stretchy`.  I could possibly refactor some other code and save a nesting level elsewhere, but I've got other priorities right now.

src/svgGeometry.js
ronkok: Yup.

ronkok:
Well, I'm out of date with regard to a submodule. I tried running `git submodule update --init --recursive` but have gotten no relief. I'm stuck.

kevinbarabash:
@ronkok I'll check out your branch tomorrow and see if I can resolve the conflict.

ronkok:
@kevinbarabash I am not able to view the Travis log, but I believe that I have this PR to the point where it passes all tests except screenshotter.

kevinbarabash:
@ronkok thanks for figuring out the merge.  I've been slammed at work.  Will try to regenerate the screenshots tonight.

kevinbarabash:
Having a coverage target for patches is silly.  I'm going to try to get rid of that check.

PR #1144, opened 2018-02-05
Enable trailing functions

ronkok:
In general, TeX functions should not parse if an argument is omitted. So `{\sqrt}` should throw an error. But there seems to be an exception. That is when the function is written at the end of an expression and the parser encounters EOF instead of the argument.

Don’t believe me? Well, here are some examples rendered from http://www.quicklatex.com.

 1.  `\( \sqrt \)` renders as sqrt
 2.  `\( \frac{a}{a} \frac a \)` renders as frac
3.  `\( \xrightarrow \)` renders as xrightarrow

This PR captures that behavior and edits several parse tests to accommodate the change.

A side benefit will occur at http://utensil-site.github.io/available-in-katex/, where four extensible arrows will flip from red to green.

### DIFF
--- a/src/Parser.js
+++ b/src/Parser.js
@@ -628,7 +628,7 @@ export default class Parser {
                     arg = newArgument(this.handleUnsupportedCmd(), nextToken);
                 } else if (this.nextToken.text === "EOF") {
                     args.push(new ParseNode("text", {body: [], type: "text",
-                        }, this.mode));
+                    }, this.mode));
                     continue;
                 } else {
                     throw new ParseError(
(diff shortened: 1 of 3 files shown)


edemaine:
I don't believe you. 😄 I just tried the first example, in actual LaTeX, and both `$\sqrt$` and `\(\sqrt\)` generate errors:

```
! Missing { inserted.
<to be read again>
                   $
l.8 $\sqrt$

?
! Missing } inserted.
<inserted text>
                }
l.8 $\sqrt$

?
! Missing { inserted.
<to be read again>
                   $
l.10 \(\sqrt\)

?
! Missing } inserted.
<inserted text>
                }
l.10 \(\sqrt\)

?
```

Of course, TeX does its best to fill these in, and adds a blank argument (so it becomes `\sqrt{}`), but I don't think KaTeX should support this.  I'm pretty sure that's what's happening in the `\frac` example too -- TeX is just doing its best and adds another blank argument.

I haven't tried the other examples, though, so feel free to argue for more specific examples.

ronkok:
Then we'll forget it.

PR #1145, opened 2018-02-06
Add \nobreakspace

ronkok:


ronkok:
> Do we want a Unicode mapping for one of these

Maybe not. One can already get a non-breaking space using `~`. Also, `unicode-math-table.txt` contains no alias for U+00A0.

edemaine:
Sounds good; thanks for checking!

PR #1135, opened 2018-02-02
Remove Makefile and migrate to NPM scripts

ylemkimon:
Fixes #1121.

PR #1140, opened 2018-02-04
set up code coverage

kevinbarabash:
In #1139 @ronkok added unit tests for `lap` commands.  I didn't realize that we didn't have before.  I decided to setup of codecov reports to help identify changes in coverage with pull requests but also indicate places where we might want to add some more tests.

Fixes #748.

PR #1146, opened 2018-02-06
Update NPM scripts and documentation

ylemkimon:
This is a follow-up of #1135.

* Group more test npm scripts into `test:`
* Update documentations on the development workflow

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,5 +15,5 @@ install:
   - npm install -g codecov
 script:
   - npm test
-  - npm run coverage && codecov
-  - dockers/Screenshotter/screenshotter.sh --verify
+  - npm run test:jest:coverage && codecov
+  - npm run test:screenshots
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,8 @@ before_script:
   - docker pull selenium/standalone-chrome:2.48.2
   - docker images --no-trunc
 install:
+  - npm install
+  - npm ls || true
   - npm install -g codecov
 script:
   - npm test
--- a/dockers/Screenshotter/README.md
+++ b/dockers/Screenshotter/README.md
@@ -1,15 +1,14 @@
 # How to generate screenshotter images
 
-## Automatic generation of screen shots
+## Automatic generation of screenshots
 
 Now you too can generate screenshots from your own computer, and (hopefully)
 have them look mostly the same as the current ones! Make sure you have docker
 installed and running.
 If all you want is (re)create
-all the snapshots for all the browsers, then you can do so by running the
-`screenshotter.sh` script:
+all the snapshots for all the browsers, then you can do so by:
 
-    dockers/Screenshotter/screenshotter.sh
+    npm run test:screenshots:update
 
 It will fetch all required selenium docker images, and use them to
 take screenshots.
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -60,21 +60,21 @@
...
+    "test:screenshots:update": "npm run prestart && dockers/Screenshotter/screenshotter.sh",
+    "test:perf": "npm run prestart && NODE_ENV=test node test/perf-test.js",
     "clean": "rm -rf build/* node_modules/",
     "clean-install": "npm run clean && npm i",
-    "screenshots": "npm run prestart && dockers/Screenshotter/screenshotter.sh",
-    "verify-screenshots": "npm run screenshots -- --verify",
     "prestart": "node check-node-version.js && check-dependencies && cd src && node unicodeMake.js >unicodeSymbols.js",
     "start": "webpack-dev-server --hot --config webpack.dev.js",
     "build": "npm run prestart && rimraf build/* && webpack",
+    "watch": "npm run build -- --watch",
```
ylemkimon: I added `watch` in #1068, but with I'm not sure there is a use case.

package.json
```diff
@@ -60,21 +60,21 @@
...
     "test:jest": "jest",
-    "jest-update": "jest --updateSnapshot",
-    "coverage": "jest --coverage",
+    "test:jest:update": "jest --updateSnapshot",
+    "test:jest:coverage": "jest --coverage",
+    "test:screenshots": "npm run test:screenshots:update -- --verify",
+    "test:screenshots:update": "npm run prestart && dockers/Screenshotter/screenshotter.sh",
+    "test:perf": "npm run prestart && NODE_ENV=test node test/perf-test.js",
     "clean": "rm -rf build/* node_modules/",
     "clean-install": "npm run clean && npm i",
```
ylemkimon: These were added in #902, but I'm not sure there is a use case. Also, I think it may be confused with only removing `build` folder.

.travis.yml
```diff
@@ -12,8 +12,10 @@ before_script:
   - docker pull selenium/standalone-chrome:2.48.2
   - docker images --no-trunc
 install:
+  - npm install
+  - npm ls
```
ylemkimon: It seems that `npm install` was no longer run, as `install` was overridden in #1140, but tests passed due to cached packages.

.travis.yml
```diff
@@ -12,8 +12,10 @@ before_script:
   - docker pull selenium/standalone-chrome:2.48.2
   - docker images --no-trunc
 install:
+  - npm install
+  - npm ls || true
```
ylemkimon: It seems that `npm install` was no longer run, as `install` was overridden in #1140, but tests passed due to cached packages.

kevinbarabash: Interesting.  I guess `check-dependencies` would've caught any difference and just failed the build.  Thanks for fixing this.

dockers/Screenshotter/README.md
```diff
@@ -61,3 +60,12 @@ Examples:
 
     node dockers/Screenshotter/screenshotter.js -i Sqrt,SqrtRoot
     dockers/Screenshotter/screenshotter.sh --exclude=GreekLetters
+
+# How to run screenshotter tests
+
+You can verify screenshots by running:
```
kevinbarabash: Thanks for adding this.

package.json
kevinbarabash: Not sure, but it's fine to keep it.

edemaine:
I'm late to this party, and just discovered that I can no longer `npm run jest` to run the Jest test.  I can see the elegance of the new naming scheme, but I find them a bit tedious to type...  Thoughts on adding aliases of `jest` -> `test:jest` etc.?

PR #1151, opened 2018-02-09
Support Unicode colon equals

ronkok:
This PR adds support for three Unicode characters: `≔` `≕` `⩴`

### DIFF
--- a/src/macros.js
+++ b/src/macros.js
@@ -400,7 +400,7 @@ defineMacro("\\Colonsim", "\\dblcolon\\mathrel{\\mkern-1.2mu}\\sim");
 // Some Unicode characters are implemented with macros to mathtools functions.
 defineMacro("\u2254", "\\coloneqq");  // :=
 defineMacro("\u2255", "\\eqqcolon");  // =:
-defineMacro("\u2A74", "\\Eqqcolon");  // ::=
+defineMacro("\u2A74", "\\Coloneqq");  // ::=
 
 //////////////////////////////////////////////////////////////////////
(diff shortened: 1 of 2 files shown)


ronkok:
I withheld these characters from PR #933 because three packages address these characters and they all disagree on the alias.

| **Character** | **Code Point** | **mathtools and KaTeX** | **colonequals and KaTeX** | **unicode-math** |
|:--------------|:---------------------|:------------------------|:--------------------------|:-----------------|
| ≔             | U+2254               | \\coloneqq              | \\colonequals             | \\coloneq        |
| ≕             | U+2255               | \\eqqcolon              | \\equalscolon             | \\eqcolon        |
| ⩴             | U+2A74               | \\Eqqcolon              | \\equalscoloncolon        | \\Coloneq        |

Not only that, but the Unicode math aliases are used by mathtools to render a different character.

|   Alias   | unicode-math    | mathtools and KaTeX |
|:----------|:--------------------:|:-----------------------------------------------:|
| \\coloneq | ≔                    | :−                                              |
| \\eqcolon | ≕                    | −:                                              |
| \\Coloneq | ⩴                    | ::−                                             |

I have said in earlier PRs that for Unicode support, the internal alias is only a means, not an end. What matters is that the author must get the results desired from the input of a specific Unicode character. That is still true, so I have implemented these Unicode characters by writing macros that call the `mathtools` aliases.

It’s a good thing that those internal aliases don’t really matter. This could have been ugly.

ronkok:
Visual confirmation:
colonequals

PR #1156, opened 2018-02-13
Rewrite spacing commands as macros

### LINKED ISSUE #1036: Use different width space characters for different size spaces in MathML
http://jkorpela.fi/chars/spaces.html contains a list of difference size spaces.  We'll want to figure out which map to `\,`, `\;`, `\:`, etc.

### LINKED ISSUE #1130: define \! \, \: \; \enspace \qquad and \quad as macros using \kern
This would allow us to make the `replace` parameter in `defineSymbol` mandatory.

edemaine:
Fix #1130 by defining `\!`, `\negthinspace`, `\,`, `\thinspace`, `\:`, `\medspace`, `\;`, `\thickspace`, `\negmedspace`, `\negthickspace`, `\enspace`, `\enskip`, `\qquad`, `\quad` as macros.  Some of these are new.  Also, we now use the AMS definitions of spaces, which are slightly different (e.g., using 1mu = 1/18em instead of .05555em, and working in text mode instead of just math mode).

Fix #1036 by defining a new `SpaceNode` in mathMLTree that recognizes all special space amounts from https://www.w3.org/TR/2000/WD-MathML2-20000328/chapter6.html


src/mathMLTree.js
```diff
@@ -143,7 +134,83 @@ class TextNode {
...
+            return node;
+        }
+    }
+
+    /**
+     * Converts the math node into an HTML markup string.
+     */
+    toMarkup(): string {
+        if (this.character) {
+            return this.character;
```
edemaine: I'm a little confused about the correct markup here. Perhaps it should be \``<mtext>${this.character}</mtext>`\`, and then we need to remove this `<mtext>` wrapper if it gets included in an `<mo>` or other token element?

test/__snapshots__/mathml-spec.js.snap
```diff
@@ -142,7 +141,7 @@ exports[`A MathML builder should output \\limsup_{x \\rightarrow \\infty} correc
       <msub>
         <mo>
           <mi mathvariant="normal">
-            lim sup
+            lim&amp;ThinSpace;sup
```
edemaine: This extra escaping is a bug, caused by `operatorname`'s `mathmlBuilder` creating a `mathMLTree.TextNode`, which escapes its text.  I'm not quite sure how to fix it, other than switching to the Unicode representations...

kevinbarabash: We're definitely doing something wrong here.  In `toNode` we call `document.createTextNode` which expects to be passed an escaped string, yet we're passing `this.text`.  I feel like we should maybe be doing the escaping within the constructor and then we could add an `escape` as an optional second parameter with the default being `true`.

src/mathMLTree.js
kevinbarabash: I'm not sure how to do this either.  In the past I've look at what MathJax outputs, but `<mspace width="thinmathspace" />` doesn't seem ideal.  Is it bad if we include `<mtext>` nodes inside `<mo>` or other element types?

kevinbarabash:
@edemaine looks like some of the screenshots changed.  Could you add updated screenshots when you have some time?

edemaine:
@kevinbarabash Done. This has revealed more bugs, which I'll take a look at later.

kevinbarabash:
@edemaine what kind of bugs did this reveal?

edemaine:
@kevinbarabash I meant bugs in this PR.  I just merged with the many changes on master, and will take another look at this.  CircleCI's screenshot diffs will help too!

edemaine:
I finally took a look at these.  It turns out that the big changes are actually bug fixes!  `NegativeSpaceBetweenRel` was particularly broken -- the negative space wasn't rendering!  Not sure how we missed that (perhaps a side-effect of the line-breaking code?).  But I'm glad this fixes it.  And `Dots` was similarly missing a negative space (`\!`).  Here are `texcmp` outputs against the new behavior, confirming they are correct:

negativespacebetweenrel
dots

`ArrayMode` has minor changes, and the `texcmp` (below) is so far to not be very revealing.  But given that the other changes are substantial improvements, I'd say we should merge.

arraymode

kevinbarabash:
@edemaine there are some `flow` errors that need resolving before we can merge.

edemaine:
@kevinbarabash Oops, fixed!

kevinbarabash:
The `Mod` screenshots changed quite a bit.  Also the CJK rows in `Unicode-chrome.png` have disappeared.

edemaine:
@kevinbarabash Thanks for spotting those -- not sure how I missed Mod in particular.

https://github.com/Khan/KaTeX/pull/1156/commits/c6f3a7aec38e72b0af746769094060e0a07297d8 recompiles the `Unicode-chrome` screenshot.  This was caused some unfortunate nondeterminism in the screenshotter loading the supporting fonts, as evidenced by the screenshot tests actually passing with the wrong image!  But now it's the correct image, phew.

The problem with the `Mod` test is that the various mod operators created manual elements of class `mspace` with various subclasses to do the spacing.  It seemed quite a bit easier to start over completely, and reimplement the mod operators (amsmath versions) via macros.  It was quite easy, and cleans up the code further.  This is https://github.com/Khan/KaTeX/pull/1156/commits/2bb10f724e92f8209887e868d741779c8fdc98e5

edemaine:
Fixed `ModScript`; I just had to remember what `\nonscript` meant again. 😄 So glad we have `\mathchoice` now!  But I might try to implement `\nonscript` in a future PR.

src/functions.js
```diff
@@ -24,7 +24,6 @@ import "./functions/lap";
 import "./functions/math";
 import "./functions/mathchoice";
 import "./functions/mclass";
-import "./functions/mod";
```
kevinbarabash: Awesome!

src/mathMLTree.js
```diff
@@ -116,34 +109,114 @@ export class MathNode {
...
+    constructor(width: number) {
+        this.width = width;
+        // See https://www.w3.org/TR/2000/WD-MathML2-20000328/chapter6.html
+        // for a table of space-like characters.  We consistently use the
+        // &LongNames; because Unicode does not have single characters for
+        // &ThickSpace; (\u2005\u200a) and all negative spaces.
+        if (width >= 0.05555 && width <= 0.05556) {
+            this.character = "&VeryThinSpace;";  // \u200a
+        } else if (width >= 0.1666 && width <= 0.1667) {
+            this.character = "&ThinSpace;";      // \u2009
```
kevinbarabash: Cool!

src/mathMLTree.js
```diff
@@ -116,34 +109,114 @@ export class MathNode {
...
+
+    /**
+     * Create a Space node with width given in CSS ems.
+     */
+    constructor(width: number) {
+        this.width = width;
+        // See https://www.w3.org/TR/2000/WD-MathML2-20000328/chapter6.html
+        // for a table of space-like characters.  We consistently use the
+        // &LongNames; because Unicode does not have single characters for
+        // &ThickSpace; (\u2005\u200a) and all negative spaces.
```
kevinbarabash: I don't don't understand this comment.  I don't see `$LongNames;` anywhere.

edemaine: I meant `&ThickSpace;` is an example of a long name between a `&` and `;`, as opposed to the Unicode equivalent.  See the table in Section 6.1.4.  I could change to `&CharacterNames;` or just "character names" if that would be clearer.

edemaine:
@kevinbarabash Thanks for the review!

PR #1159, opened 2018-02-14
Support \underline in text mode

### LINKED ISSUE #1158: Supporting \underline in Text Mode
As a user I would expect `\text{\underline{hi}}` to work. Currently this works as expected in Math mode.

Actual Results: Error is thrown, and text is not rendered with an underline. 

Not sure if this is considered a bug, or just not yet implemented, but tagging as a bug for now.

edemaine:
Fix #1158 with a one-line change.  I could not find any other "long accents" that are supposed to work in text mode, but let me know if you find any!

Screenshot:
image

Comparison to LaTeX:
image


rrandallcainc:
@edemaine Meant to mention prior to approving. Perhaps we'd want to add a test for this? Might be able to add to the existing "Text" screenshot.

edemaine:
I added a simple test to the LowerAccent one which currently tests `\underline`. Let me know what you think!

Issue #1160, opened 2018-02-15
[CRITICAL BUG] Inject any html/css/js by using KaTeX


ArStah:
Please contact me to arsen.stahanov@gmail.com

kevinbarabash:
@ArStah would you be willing to submit this using hackerone.com?

ArStah:
@kevinbarabash i can't find your organization there

kevinbarabash:
@ArStah it should be listed as Khan Academy.

ArStah:
Look [here](https://gitlab.com/gitlab-org/gitter/webapp-temp-issues-for-move/issues/108) what happens

kevinbarabash:
This is appears to be a gitter issue.  It looks like they aren't properly sanitizing the text inside `$$ ... $$` blocks before adding them to the page.

edemaine:
Hard to tell without more information, but I imagine if there's any input to `katex.render`/`katex.renderToString` that can output a `<script>` tag, that'd be a bug in KaTeX. (Of course, gitter also shouldn't be relying on this, and should be sanitizing the output of KaTeX too... I certainly do this in my own project, though it'd be helpful to have guidelines for what to whitelist for KaTeX to work.) @kevinbarabash Are you suggesting that's not the case, and it's some other workaround that avoids calling KaTeX?

pascalwhoop:
Yes I also believe it shouldn't be positioned in KaTeX although it could be "good" to help people who add this library to help themselves by sanitizing against any html *inside* of KaTeX. Or is there any use case where people have html>KaTeX>html?

kevinbarabash:
@pascalwhoop could you expand on what you mean by `html>KaTeX>html`?

pascalwhoop:
Ah sorry. Usually you nest KaTeX code inside of html right? So let's say we use
```
<p>$$ f(x) = y^2 $$</p>
```
is there any reason why we should have it be something like
```
<p>$$ f(x) = y^2 <em>\text{chickens}</em> $$</p>
```

If not, then forbidding any html inside the string that is passed to `katex.render` could help developers help themselves. A bit.

kevinbarabash:
We already "forbid" HTML inside the string we pass to `katex.render` b/c it's not valid LaTeX`.

pascalwhoop:
Okay then how exactly does it occur, that the HTML breaks out? I guess you forbid it, throw an error and the string is output to the UI of the client, which then is in fact our HTML that got injected?

[...]

PR #1161, opened 2018-02-15
Add a paragraph about injection safety

edemaine:
Based on discussion in #1160. Comments welcome!


pascalwhoop:
Looks good to me. Maybe note that this is specifically important if the latex source code can come from untrusted third parties.

edemaine:
That's true, although even for non-malicious inputs, failing to escape can lead to incorrect HTML.  So it's useful more broadly too.  Maybe I should revise in this direction...

edemaine:
OK, I split the text into a "Security" subsection (with positive claims) and a "Handling errors" subsection (which should apply whether you're worried about security or not).  Let me know whether you like this better or worse.

Issue #1153, opened 2018-02-12
\mathrlap bounding box does not expand for tall argument


yeruoforever:
when input this:
$$T+\rlap{\overbrace{\phantom{a_0+a_1+a_2+\cdots+a_k}}^m}a_0+\underbrace{a_1+a_2+\cdots+a_k}_n+f$$
output is :
![2018-02-12 21-33-40](https://user-images.githubusercontent.com/19868036/36099239-d167dfa6-103c-11e8-9282-0f5102d15416.png)
some times output is like this：
![2018-02-12 21-40-10](https://user-images.githubusercontent.com/19868036/36099476-ca06b4c0-103d-11e8-8ca0-1c7f59d3923b.png)
$$T+\rlap{\overbrace{\phantom{a_0+a_1+a_2+\cdots+a_k}}^m}a_0+\underbrace{a_1+a_2+\cdots+a_k}_n+f$
$$

edemaine:
FWIW, that code should be
```latex
T+\mathrlap{\overbrace{\phantom{a_0+a_1+a_2+\cdots+a_k}}^m}a_0+\underbrace{a_1+a_2+\cdots+a_k}_n+f
```
I confirm that I see the same as your first image, which is buggy.  (I couldn't get the example with `$` to work, even with `\$`.  But LaTeX renders it like your second image.)  My guess is that the height of the `\phantom` is somehow not getting incorporated into the height computation of the `\overbrace`... but this will require some investigation.

ronkok:
There is a bug there and I'll look into it. In the meantime, @yeruoforever , is this the sort of thing you were seeking?
![overbrace](https://user-images.githubusercontent.com/16403058/36110663-467b551c-0fd8-11e8-8030-76231879462f.PNG)


If so, then you can use:
```
$$ T+\overbrace{a_0+\underbrace{a_1+a_2+\cdots+a_k}_n}^m + f $$
```

ronkok:
I believe that I see the source of the bug. The problem is in `\rmathrlap`, not `\overbrace` or `\phantom`.

 KaTeX `rlap` and `llap` have always worked by just nesting a span with absolute positioning inside a span that has relative positioning. Then it is easy to manipulate the horizontal alignment by setting `left: 0;`  or `right:0;`.  

But that doesn't address the vertical alignment issues when tall math structures come into play. We need to incorporate a `makeVList` into the `\mathrlap` HTML handler.  

I can work something up in the next couple of days.

[...]

PR #1162, opened 2018-02-16
Fix lap vertical alignment.

ronkok:
Fixes issue #1153.


ylemkimon:
**Firefox**
lap-firefox-diff
[Lap-firefox.png](https://30-108306065-gh.circle-artifacts.com/0/image/Lap-firefox.png)

**Chrome**
lap-chrome-diff
[Lap-chrome.png](https://29-108306065-gh.circle-artifacts.com/0/image/Lap-chrome.png)

ronkok:
@ylemkimon, I'm guessing that one is LaTeX and the other is KaTeX. Which is which?

ylemkimon:
It’s both KaTeX, where red is master branch and green is this PR.

kevinbarabash:
> It’s both KaTeX, where red is master branch and green is this PR.

@ylemkimon with such a big change to the screenshot, it's probably more useful to compare KaTeX output vs LaTeX output.

kevinbarabash:
@ronkok I'll generate screenshots over the weekend.

ylemkimon:
@kevinbarabash Yes, it's just an output of the screenshotter.

ronkok:
The most recent commit wraps the span with a `mord` span and thereby fixes the horizontal alignment relative to an adjacent span.

ylemkimon:
It seems that it is correctly rendered in the [diff](https://121-11195700-gh.circle-artifacts.com/0/diff/Lap-firefox-diff.png), but not in the [generated screenshot](https://121-11195700-gh.circle-artifacts.com/0/image/Lap-firefox.png).

kevinbarabash:
@ronkok do the screenshots need regenerating?

ronkok:
I expect so. I've added screenshotter tests to cover the spacing problem turned up in issue #1234.

ronkok:
@kevinbarabash  I believe that this PR, #1162, needs nothing more than screenshots. Since it fixes two rendering bugs, it also would be a good one to merge before the next release.

edemaine:
Screenshots updated.

ronkok:
Screenshot test "Lap" failed. When I run the same code locally, it looks fine to me.

@edemaine Can I ask for another favor and request a regeneration of screenshotter test "Lap"?

kevinbarabash:
@edemaine weird.  I'll give it a try.

kevinbarabash:
Here's the diff between LaTeX (green) and KaTeX (red):
lap

kevinbarabash:
@edemaine the first time I ran the screenshot I got a image where the overbrace was overlapping the expression underneath it.  I ran the screenshot again and got the proper output.

kevinbarabash:
Note: I had to add `\usepackage{mathtools}` to test.tex in order to get `texcmp.sh` to work.

src/functions/lap.js
```diff
@@ -36,8 +36,28 @@ defineFunction({
...
+        // At this point, we have correctly set horizontal alignment of the
+        // two items involved in the lap.
+        // Next, use struts to set the height of the HTML bounding box.
+        // Otherwise, a tall argument may be misplaced.
+        const topStrut = buildCommon.makeSpan(["strut"]);
+        const bottomStrut = buildCommon.makeSpan(["strut", "bottom"]);
+        topStrut.style.height = node.height + "em";
+        bottomStrut.style.height = (node.height + node.depth) + "em";
+        bottomStrut.style.verticalAlign = -node.depth + "em";
+        node = buildCommon.makeSpan([], [topStrut, bottomStrut, node]);
```
kevinbarabash: Suggest going with a single strut similar to the change https://github.com/Khan/KaTeX/pull/1307.

kevinbarabash:
Hmm... the Firefox failure might have to do with:
```
        if (key === "Lap" && opts.browser === "firefox" &&
            img.buf[0x32] === 0xf8) {
            /* There is some strange non-determinism with this case,
             * causing slight vertical shifts.  The first difference
             * is at offset 0x32, where one file has byte 0xf8 and
             * the other has something else.  By using a different
             * output file name for one of these cases, we accept both.
             */
            key += "_alt";
            file = path.join(dstDir, key + "-" + opts.browser + ".png");
            if (loadExpected) {
                loadExpected = promisify(fs.readFile, file);
            }
        }
```
I'm not sure how the generate `Lap_alt.png` though.

src/functions/lap.js
ronkok: Working on it.

ronkok: Done.

ronkok:
What I'm about to say is just a conjecture, but it may be possible to resolve the Firefox vertical alignment issue by using the current version of Firefox to generate the screenshots. Apparently, Firefox 57 [fixed a bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1379306) that took place within tables with border collapse. Since `makeVList` uses `display: table-cell`, that may have some bearing on this problem.

If this sounds like I'm grasping at straws, that's because I am.

kevinbarabash:
@edemaine @ronkok maybe we can grab the image that circle-ci is producing for Firefox and use that as the _alt version of the image.

edemaine:
I did confirm that updating Firefox changed the image (but haven't checked whether it affects determinism). Might be worth updating anyway, which may affect several screenshots, but that's for a different PR.

The annoying thing is that the Chrome screenshot has the same issue. Updating Chrome did not affect that screenshot. So we may need an alt image for that...

ronkok:
I'm stuck.

I want to get this PR merged so that I can then go clean up my repository before I resume work on anything else. But it seems that no one knows how to get screenshotter to properly handle the test for Lap. 

If we have to choose between that test and this PR's code, we have to take the code. It fixes two rendering bugs.

So the latest commit deletes the screenshotter test. I'm sorry to do that, but I can think of no better course of action.

kevinbarabash:
@ronkok can you open an issue to update the screenshotter test?  We can treat figuring out how to update the Lap screenshot tests as a separate task.

ronkok:
Will do.

test/screenshotter/ss_data.yaml
```diff
@@ -132,7 +132,6 @@ KaTeX:
 Kern:
     tex: \frac{a\kern{1em}b}{c}a\kern{1em}b\kern{1ex}c\kern{-0.25em}d
     nolatex: LaTeX fails to typeset this, “Missing number, treated as zero.”
-Lap: ab\mathllap{f}cd\mathrlap{g}hij\mathclap{k}lm \; ab\llap{f}cd\rlap{g}hij\clap{k}lm
```
kevinbarabash: Instead of removing the test completely, could you revert to the updated test you made and then comment it out so that it's easy for people pick up the work?

ronkok: Done

ronkok:
Yay!

PR #1164, opened 2018-02-17
add .codecov.yml and disable patch coverage

kevinbarabash:
It doesn't make sense to have a target for patch coverage b/c you may be editing a file that has really low coverage to begin with and it shouldn't be your responsibility to bring it up to the target as part of the PR.


kevinbarabash:
Only project-level coverage is being computed now.

PR #1167, opened 2018-02-17
Support \raise and \lower

ronkok:
This PR adds support for unbraced `\raise` and `\lower`. It implements the MathJax versions of these functions, which are more liberal than LaTeX in the kind of arguments they accept. That will provoke some discussion, so I’ll start by describing the differences.

Both versions take two arguments, as in `\raise<distance><subject>`. Then:

|                 | **LaTeX accepts**                 | **MathJax accepts**                                                               |
|:----------------|:----------------------------------|:----------------------------------------------------------------------------------|
| First argument  | Unbraced distances only           | Braced or unbraced distances                                                      |
| Second argument | TeX boxes only                    | Any valid LaTeX                                                                   |
| Example (text)  | `\raise5pt \hbox{some text}`      |` \raise5pt{some text}`<br><br>`\raise{5pt}{some text}` <br>`\raise5pt \hbox{some text}`        |
| Example (math)  | `\raise5pt \hbox{\$\frac a b\$}` | `\raise5pt{\frac a b}`<br><br>`\raise{5pt}{\frac a b}` <br>`\raise5pt \hbox{\$\frac a b\$}` |

LaTeX box functions, such as `\hbox`, switch out of math mode into text mode/normalsize. In order to raise math in LaTeX, one has to explicitly switch back into math mode and reset the style inside the box, as in `{\$\scriptstyle a/b\$}`. Since MathJax does not require a box, it doesn’t need a double mode switch.

I prefer the MathJax approach here. It provides

1. A superior, simpler syntax
2. Faster math rendering
3. Backwards compatibility with either LaTeX or MathJax

To provide backwards compatibility with LaTeX, this PR also provides support for `\hbox`.


ylemkimon:
**Firefox**
raise-firefox
[Raise-firefox.png](https://32-108306065-gh.circle-artifacts.com/0/image/Raise-firefox.png)

**Chrome**
raise-chrome
[Raise-chrome.png](https://31-108306065-gh.circle-artifacts.com/0/image/Raise-chrome.png)

kevinbarabash:
It's weird that the circleci build isn't showing up even though `master` has been merged in.  It's also weird that the Chrome screenshot has a scrollbar at the bottom.  @ronkok I'll generate some screenshots this weekend.  Hopefully we can figure out what's up with circleci later so that you can grab screenshots from the generated artifacts.

kevinbarabash:
@ronkok can you add tests to katex-spec.js to check that `\lower` and `\raise` parse and build?

ronkok:
Will do.

ylemkimon:
@kevinbarabash CircleCI has to be manually enabled for PRs from forked repo in the `Build settings`:
circleci

ronkok:
@kevinbarabash  It turns out that I had already written some parse tests. My latest commit adds a few build tests.

kevinbarabash:
@ylemkimon I've changed that setting in the project settings on circleci.  I'm guessing it only takes effect on new PRs.

src/functions/raise.js
```diff
@@ -0,0 +1,131 @@
...
+            body: body,
+            funcName: context.funcName,
+        };
+    },
+    htmlBuilder,
+    mathmlBuilder,
+});
+
+// We provide \hbox for backwards compatibility with LaTeX \raise.
+// This enables an author to write \raise6pt \hbox{text}.
```
kevinbarabash: Is it possible to use `\hbox` without `\raise` or `\lower` in front of it?

ronkok: Yes, `\hbox` is a stand alone function and can be used without `\raise`. I'm not sure why anyone would want to do so, at least so long as KaTeX does not enable line breaks.

ronkok:
This PR introduces a non-standard, that is, non-LaTeX syntax for `\raise`. That sort of thing has been discouraged by KaTeX ownership in the past. Naturally, I think there are excellent reasons for the syntax in this PR, but it would be good to get an explicit approval. 

@kevinbarabash Would you mind giving a thumbs up to the idea of `\raise` that works on something other than a LaTeX box?

edemaine:
@ronkok Not a bad idea to double check.  Personally, I'm fine with it, given that it does in particular support "correct" syntax.  I agree we're not entirely consistent on this point throughout KaTeX, though -- the `unicodeTextInMathMode` flag comes to mind (and I still feel like that should maybe default to `true` for convenience to the poor user...).

Maybe we could add some kind of `strict` mode that either warns or errors (perhaps `strict: 'warn'` or `strict: 'error'`?) when using not-actually-LaTeX behavior, like non-boxes in `\raise`/`\lower`, or Unicode text characters in math mode.  That wouldn't be for this PR, but could help get around these repeated issues: allow slightly more flexible syntax for convenience, but allow someone to tie down to actual LaTeX as desired.

edemaine:
@ronkok What about adding a warning when this feature is used beyond what LaTeX supports, similar to how `\hskip` and `\mskip` behave?  Then the detection is done, and in a future PR we can easily adapt this to error or silent behavior depending on @kevinbarabash's input and decisions about #1226. Also the warning should make it clear that this might not be kosher in a future version for those tracking `master` branch directly.

ronkok:
@edemaine Let me think a little about that warning idea. It sounds reasonable. I'd like to try out an implementation before I sign on entirely.

kevinbarabash:
I've responded to #1226.  I think we should move ahead with a `strict` option.  I've marked this PR as blocked b/c it would be nice to have that option in place first and then update this PR to use that setting.

ronkok:
I'm closing this PR temporarily. I'll re-open after PR #1278 has landed and I have edited this PR as per the new strict mode.

nhpf:
Hi @ronkok, as of 2022, can we still expect future KaTeX support for `\raise` and `\lower`?

Issue #475, opened 2016-05-17
throwOnError confusing


johanneswilm:
the name "throwOnError" sounds like it would cover all types of errors. It is unclear whether it is supposed to cover all types of errors or only those based on yet unknown latex commands.

This throws a parse error:

```
katex.render('2+x{',document.body,{throwOnError:false})
```

lolmaus:
Also ran into this issue. It looks like a bug to me.

shuttle1987:
From by discussions in this PR https://github.com/Khan/KaTeX/pull/540 I now realize that `throwOnError` only changes how KaTeX handles the behavior of unsupported TeX commands. It doesn't currently deal with other parsing related issues, those will still throw exceptions.

edemaine:
FWIW, I too was confused by this.  Effectively, the option is "throw on use of incorrect \command" -- not what I expected.

edemaine:
Fixed by #1169.

Issue #538, opened 2016-09-20
Improve error message on unsupported command


gagern:
This is a spin-off from https://github.com/Khan/KaTeX/issues/526#issuecomment-246549854.

If input contains some unsupported command nested in some group or environment, the parser will complain about the fact that the group resp. environment isn't ended with one of the expected symbols. The most likely scenario, however, is that the user didn't really mean to end the environment, but did expect the function in question to exist. So it would be better to report that appropriately.

Source locations: if [this check](https://github.com/Khan/KaTeX/blob/a16ae7a5ebc97cfb610a2143254393918fd0650d/src/Parser.js#L783) finds no matching function, [`null` will get returned](https://github.com/Khan/KaTeX/blob/a16ae7a5ebc97cfb610a2143254393918fd0650d/src/Parser.js#L804) which propagates up the tree and out of [the `parseExpression` invocation](https://github.com/Khan/KaTeX/blob/a16ae7a5ebc97cfb610a2143254393918fd0650d/src/environments.js#L18) where [it triggers the observed error message](https://github.com/Khan/KaTeX/blob/a16ae7a5ebc97cfb610a2143254393918fd0650d/src/environments.js#L31-L32). [The `!throwOnError` handling](https://github.com/Khan/KaTeX/blob/a16ae7a5ebc97cfb610a2143254393918fd0650d/src/Parser.js#L159-L163) might be the right place to generate a more useful error message, too. Doing so may make behavior with or without `throwOnError` more consistent as well.

PR #1169, opened 2018-02-18
Stop throwing ParseError when throwOnError is false

edemaine:
This PR is motivated by the recent security issue reported in #1160. I would argue that almost anyone who uses KaTeX on a user-specified string doesn't want an exception to happen, but rather wants KaTeX to do its best job at rendering the given LaTeX.  `throwOnError` is the obvious "solution", but it actually only fixes invalid commands, but any other type of `ParseError` still gets thrown.  This is nice because you get partial renderings, but not nice because it doesn't fix the original problem of not throwing exceptions.  As a result, I believe many KaTeX users (myself included) write our own `try ... catch ...` block to render something else when KaTeX throws an error.  And if done poorly, this leads to security holes like #1160.

This PR makes `throwOnError` more useful by rendering HTML for all types of `ParseError`.  If it's just an unknown command, you get a partial render as before.  But if you have some other type of error, you get a `<span class="katex-error">` whose text is the LaTeX source code, its color is `options.errorColor` (defaulting to red as usual), and it has a hover title that displays the actual error message.  This is what I did in [my own KaTeX error handler](https://github.com/edemaine/coauthor/blob/8a2f9c5f19591d04a83e1fa26d74c5b1fc0c0e71/lib/formats.coffee#L437-L453).  Crucially, the output HTML is properly escaped (thanks to our existing domTree infrastructure -- see the test snapshot).

There's an obvious alternative: show the error message directly, and hover to see the LaTeX. Or some other combination.  But IMHO this PR is probably closest to the intent of the author: if you can't render nice LaTeX, at least render the source code.

Another question is naming.  I took over `throwOnError` because it is the natural name for this feature (and probably what most people expect).  But it breaks backwards compatibility, so would need to be a new version.  We could use another name to keep backward compatibility.  We could also provide the old behavior, via something like `throwOnBadCommand`.  I'm not sure of a great use for this...  If one wants to reformat the error style, they could use CSS, or (in the worst case) parse the resulting `<span class="katex-error">` and restructure.

Along the way:
* Use new `katex.__renderToDomTree` in katex-spec's `_getBuilt`.  (This was necessary to get the new error handling in `_getBuilt`.)
* Fix jest results which must always be functions, not strings.


ylemkimon:
How about using error level and deprecating `throwOnError`:
0 - suppress all errors, instead print LaTeX source code, this PR's behavior when `!throwOnError`
1 - suppress all errors except unsupported command error, current behavior when `!throwOnError`
2 - throw all errors, current/PR behavior when `throwOnError`

kevinbarabash:
@edemaine thanks for the PR.  This is definitely a pain point for users.  I will review this weekend.

test/katex-spec.js
```diff
@@ -98,10 +100,10 @@ const parseAndSetResult = function(expr, result, settings) {
     } catch (e) {
         result.pass = false;
         if (e instanceof ParseError) {
-            result.message = "'" + expr + "' failed " +
+            result.message = () => "'" + expr + "' failed " +
                 "parsing with error: " + e.message;
```
kevinbarabash: Why did `result.message` change to a callback?

test/__snapshots__/katex-spec.js.snap
```diff
@@ -1,5 +1,39 @@
...
+  "depth": 0,
+  "height": 0,
+  "maxFontSize": 0,
+  "style": {
+  }
+}
+`;
+
+exports[`A parser that does not throw on unsupported commands should properly escape LaTeX in errors 1`] = `"<span class=\\"katex-error\\" title=\\"ParseError: KaTeX parse error: Expected group after &#x27;^&#x27; at position 2: 2^̲&amp;&quot;&lt;&gt;\\" style=\\"color:#933\\">2^&amp;&quot;&lt;&gt;</span>"`;
+
```
kevinbarabash: nice

static/main.js
```diff
@@ -21,7 +21,7 @@ function init() {
     const macros = {};
     // TODO: Add toggle for displayMode.
     // https://github.com/Khan/KaTeX/issues/1035
-    const options = {displayMode: true};
+    const options = {displayMode: true, throwOnError: false};
```
kevinbarabash: Cool.

katex.js
```diff
@@ -68,6 +70,25 @@ const generateParseTree = function(
...
+ * renders the invalid LaTeX as a span with hover title giving the KaTeX
+ * error message.  Otherwise, simply throws the error.
+ */
+const renderError = function(
+    error,
+    expression: string,
+    options: Settings,
+) {
+    if (options.throwOnError || !(error instanceof ParseError)) {
+        throw error;
```
kevinbarabash: Does re-throwing maintain the call stack?

katex.js
```diff
@@ -68,6 +70,25 @@ const generateParseTree = function(
     return parseTree(expression, settings);
 };
 
+/**
+ * If the given error is a KaTeX ParseError and options.throwOnError is false,
+ * renders the invalid LaTeX as a span with hover title giving the KaTeX
+ * error message.  Otherwise, simply throws the error.
+ */
+const renderError = function(
```
kevinbarabash: This is a really slick solution.

katex.js
ylemkimon: @kevinbarabash Yes 😄

kevinbarabash:
> How about using error levels:
0 - suppress all errors, instead print LaTeX source code, this PR's behavior when !throwOnError
1 - suppress unsupported command error, current behavior when !throwOnError
2 - throw all errors, current/PR behavior when throwOnError

I can't think of a situation where people would actually want level 1.  They'd still need to use a try/catch to avoid errors other than unsupported command errors.

ylemkimon:
@kevinbarabash A user may want to implement custom error handler, but in the custom error handler, it cannot access KaTeX internals, i.e., if it's unsupported command error, print the plain command name, though as you said, I'm not sure there are people want it.

edemaine:
I could see wanting to build a custom handler for custom formatting in the error case.  An alternative would be to provide a callback to call instead of an error handler.  Maybe `throwOnError` could receive a function argument to handle this case?

test/katex-spec.js
edemaine: This is necessary since some update to Jest: Jest requires all messages to be callable.  Obviously not directly related to this PR, but I ran into this (presumably long-standing) bug when my tests were failing.

ylemkimon:
This will also fix #475.

Related: #538

ylemkimon:
With #1226, I think we should have an error callback and handle errors differently in there.

kevinbarabash:
@edemaine looks like there's some lint:
```
/home/travis/build/Khan/KaTeX/test/katex-spec.js
  3120:16  error  'buildTree' is not defined  no-undef
```

edemaine:
Thanks for the review, @ylemkimon. Do you think I should add an error callback function for custom error handling/rendering? Or leave that for consideration in another PR?

ylemkimon:
Now I think currently there is no way to access KaTeX internals, and custom error handler probably should be another issue/PR. Thank you for the PR!

PR #1177, opened 2018-02-25
Port delimiter.js to @flow.

marcianx:


src/delimiter.js
```diff
@@ -115,12 +155,16 @@ const makeLargeDelim = function(delim, size, center, options, mode, classes) {
  * Make an inner span with the given offset and in the given font. This is used
  * in `makeStackedDelim` to make the stacking pieces for the delimiter.
  */
-const makeInner = function(symbol, font, mode) {
+const makeInner = function(
+    symbol: string,
+    font: "Size1-Regular" | "Size4-Regular",
```
kevinbarabash: I'm surprised we only call this with `Size1` or `Size4` since there are sizes in between.

src/delimiter.js
```diff
@@ -323,7 +373,12 @@ const makeStackedDelim = function(delim, heightTotal, center, options, mode,
 const vbPad = 80;   // padding above the surd, measured inside the viewBox.
 const emPad = 0.08; // padding, in ems, measured in the document.
 
-const sqrtSvg = function(sqrtName, height, viewBoxHeight, options) {
+const sqrtSvg = function(
+    sqrtName: string,
```
kevinbarabash: We should be able to scope this to specific strings, e.g. "sqrtTall", etc.

src/delimiter.js
```diff
@@ -510,22 +585,30 @@ const stackLargeDelimiterSequence = [
 
 /**
  * Get the font used in a delimiter based on what kind of delimiter it is.
+ * TODO(#963) Use more specific font family return type once that is introduced.
  */
-const delimTypeToFont = function(type) {
+const delimTypeToFont = function(type: Delimiter): string {
```
kevinbarabash: Maybe we should limit the return type `"Main-Regular" | "Size1-Regular" | ... | "Size4-Regular"`, wdyt?

kevinbarabash: Nevermind, `flow` will probably have an issue actually verifying the return type.

src/functions/delimsizing.js
```diff
@@ -275,7 +275,13 @@ defineFunction({
...
                 group.value.value, 1, options,
                 group.mode, []);
-            middleDelim.isMiddle = {value: group.value.value, options: options};
+
+            // Property `isMiddle` not defined on `span`. It is only used in
+            // this file above. Fixing this requires refactoring the htmlBuilder
+            // return type to support passing additional data (harder) or adding
+            // `isMiddle` to `span` just for this case (easier).
+            // $FlowFixMe
+            middleDelim.isMiddle = {value: group.value.value, options};
```
kevinbarabash: Adding `isMiddle` to `span` doesn't seem like the right way to go b/c it's so specific to this particular situation.  The "harder" solution is probably the way to go.  Thanks for leaving a comment with suggestions o how to fix the `$FlowFixMe`.

src/delimiter.js
ronkok: `sqrtName` can take one of five values:, "sqrtMain", "sqrtSize1", "sqrtSize2", "sqrtSize3", "sqrtSize4", "sqrtTall"

marcianx: It's called with `sqrtSvg("sqrtSize" + delim.size, ...)` which doesn't automatically satisfy `"sqrtSize1" | "sqrtSize2" | "sqrtSize3"`. Since this is a local function only called within this file, I'm not sure it's worth having to override the type-checker to accept this code.

src/functions/delimsizing.js
marcianx: You're right. I updated it to identify which option is more ideal.

src/delimiter.js
ronkok: > a local function only called within this file, 

I agree that not much is gained by another restriction.

src/delimiter.js
marcianx: Yea, similar to the previous comment, I'd have to override the compiler. Since this is only used locally, I thought it to not be worth it, but I am open to either change if preferred.

kevinbarabash:
@marcianx thanks. :)

Issue #1178, opened 2018-02-25
Subscript/height issue within contenteditable


alexeymrkn:
The issue is very well seen on these pictures. I'm using the latest version of KaTeX, the issue is reproducible in Chrome and Opera on macOS as well as Chrome for Android. Haven't had a chance to test it on other platforms yet.

**Chrome**
<img src="https://user-images.githubusercontent.com/6084153/36646076-9aa69eb6-1a83-11e8-9ba4-31afd538fb93.jpg" width=350>

**Safari**
<img src="https://user-images.githubusercontent.com/6084153/36646051-4f031ebc-1a83-11e8-8838-5d4eb49c5724.jpg" width=350>

ronkok:
I wonder if the KaTeX CSS is insufficiently isolated from external CSS.

@alexeymrkn Could you point to a test page where this occurs?

alexeymrkn:
@ronkok this code is a part of a project that I maintain locally. After your request I've tried to reproduce the bug on a remote server (simplifying the code to ensure clearness) but failed to do so. It took me quite a time to realize that this bug appears only when inline math is rendered inside a `contenteditable` block. 

So here is a [jsFiddle playground](https://jsfiddle.net/91sc0vru/5/) demonstrating the issue.

ronkok:
Among the  many clever steps in `makeVList`, it inserts a span with class `vlist` inside a span with class `vlist-r`, [here](https://github.com/Khan/KaTeX/blob/master/src/buildCommon.js#L560). The `vlist` span has no contents and it has a CSS rule of `display: table-cell;`. So it should render with the small height that the code defines for it.

Apparently Chrome will, in  `contenteditable` mode only, treat the cell as if it contains some text content, which then defines a `min-height` for the cell.  This can be evaded by placing yet another an empty span into the `vlist` span.

kevinbarabash:
@ronkok thanks for investigating this.  I wish that we didn't have to content with `contenteditable`.  It's so poorly defined and each browser does its own thing. :\

ronkok:
@alexeymrkn Thank you for determining that this problem takes place within a `contenteditable` div. That tip was the key to fixing the bug. A fix has now been merged into KaTeX master.

alexeymrkn:
@ronkok It was my pleasure to make this nano contribution to this awesome project.

PR #1179, opened 2018-02-26
Fix contenteditable mode

ronkok:
`makeVList` sets depth with an empty span for which it defines a span height. But Chrome, in `contenteditable` mode only, treats that span as if it contains some text.  The resulting `min-height` can over-ride our desired depth.

So we insert an empty span inside the `depthStrut` span. This fixes issue #1178.

I don't know how to write a test for this. The condition only occurs inside an external `contenteditable` div.

### DIFF
--- a/src/buildCommon.js
+++ b/src/buildCommon.js
@@ -551,12 +551,12 @@ const makeVList = function(params: VListParam, options: Options): domTree.span {
     let rows;
     if (minPos < 0) {
         // We will define depth in an empty span with display: table-cell.
-		// It should render with the height that we define. But Chrome, in
-		// contenteditable mode only, treats that span as if it contains some
-		// text content. And that min-height over-rides our desired height.
-		// So we put another empty span inside the depth strut span.
-		const emptySpan = makeSpan([], []);
-		const depthStrut = makeSpan(["vlist"], [emptySpan]);
+        // It should render with the height that we define. But Chrome, in
+        // contenteditable mode only, treats that span as if it contains some
+        // text content. And that min-height over-rides our desired height.
+        // So we put another empty span inside the depth strut span.
+        const emptySpan = makeSpan([], []);
+        const depthStrut = makeSpan(["vlist"], [emptySpan]);
         depthStrut.style.height = -minPos + "em";
 
         // Safari wants the first row to have inline content; otherwise it
--- a/src/buildCommon.js
+++ b/src/buildCommon.js
@@ -550,7 +550,13 @@ const makeVList = function(params: VListParam, options: Options): domTree.span {
     // A second row is used if necessary to represent the vlist's depth.
     let rows;
     if (minPos < 0) {
-        const depthStrut = makeSpan(["vlist"], []);
+        // We will define depth in an empty span with display: table-cell.
+		// It should render with the height that we define. But Chrome, in
+		// contenteditable mode only, treats that span as if it contains some
+		// text content. And that min-height over-rides our desired height.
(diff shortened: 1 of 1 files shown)


ronkok:
Oops. I forgot to create a feature branch. Now I've made a pull request from my local master branch. Sorry.

This PR should get reviewed before my other pending PRs. Until this one lands I won't be able to help much with the others.

kevinbarabash:
@ronkok for testing, I think a screenshot test is probably appropriate.  In https://github.com/Khan/KaTeX/commit/53203ed92cc3a2a89b0de178bf540f34a3f81f4f @davidflanagan added a option to the screenshot tests to include a `style` property.  We could also add a `content editable` property in a similar fashion.

davidflanagan:
Since you mentioned me and screenshot tests, I had an idea this weekend that I'll just drop here: I don't know how the screenshots are being generated in headless mode inside those docker containers, but I wonder if there is an option to specify the dpi resolution of the virtual screen that is used. For some of @ronkok's recent work on horizontal lines it seems like testing high resolution and low resolution screenshots might be helpful

ronkok:
@kevinbarabash I don't think the usual screenshot infrastructure will capture the issue here. Everything within a KaTeX span renders fine. Below, I show two fractions, one with the current master and one with this PR. Both renderings are fine because vertical alignment is set by `makeVList`.
contenteditable_1

But the following image shows how KaTeX interacts with surrounding HTML. The image on the left is the currrent master and on the right is the proposed PR.
contenteditable_2

So I'm still stuck. I'm sure that I've fixed the problem, but I don't know how to make screenshotter demonstrate it.

edemaine:
In `static/main.js` (used in the test server via `npm start`), there's an option called `after=...` that lets you put text (maybe arbitrary HTML?) after the rendered KaTeX stuff.  (Though it also only supports display math at the moment...)  Anyway, it's plausible we could at a YAML field `after:` that lets us put some HTML after the LaTeX part in the screenshot test.  But this might be for another issue/PR?

edemaine:
Actually, it turns out that `test/screenshotter/test.html` already supports "pre" and "post" queries to put text before and after the rendered KaTeX.  It would be easy to add support for `pre:` and `post:` fields in the test YAML to set these.  This would allow us to build a test that illustrates bug #836, for example.

But we'll also need a "contenteditable" flag.  I will give it a try.

edemaine:
My bad.  `test/screenshotter/ss_data.js` already imports the `pre` and `post` strings, so that already works (and is in one test).  So I just need to add a `contenteditable` flag.  At least now I know how!

edemaine:
Unfortunately, I've been unable to reproduce this issue in the screenshotter.  Maybe a side-effect of headless mode that it ignores contentEditable?  I've tried even this hard-coded HTML (in `test/screenshotter/test.html`):

```html
    <div id="container" contentEditable="true">
      TEXT TEXT TEXT TEXT TEXT TEXT TEXT TEXT TEXT TEXT TEXT TEXT TEXT TEXT
      <span id="math"></span>
      TEXT TEXT TEXT TEXT TEXT TEXT TEXT TEXT TEXT TEXT TEXT TEXT TEXT TEXT
    </div>
```

The result (in Chrome) is this:

image

Am I missing something?  Otherwise, I'm not sure we can make a screenshot test of `contentEditable`.

ronkok:
All this PR does is put one empty span inside another empty span. Maybe we could do without a test?

kevinbarabash:
@edemaine I like change to test.html, it's too bad it doesn't reproduce the issue. :(

PR #1180, opened 2018-02-27
Update to 0.9.0

### LINKED ISSUE #1171: Update demo page to use v0.9.0.
This change will need to be made on the `gh-pages` branch.

edemaine:
Fix #1171


edemaine:
I realize that I didn't actually test this locally. Probably should before merging..

edemaine:
Just tested (via `python -m SimpleHTTPServer 8001` and checking on Chrome that e.g. `\boldsymbol` works).  So should be good for review (though apparently no review is necessary on this branch).

edemaine:
@ronkok I just noticed that [function-support.html](https://github.com/Khan/KaTeX/blob/gh-pages/function-support.html) already uses `unpkg.com` (and would need to be updated to remove the `-beta`).  We should probably unify to use only one method of inclusion...

ronkok:
@edemaine, When I most recently updated the function support page, I was too impatient to wait the 2 or 3 days that the cloudflare CDN takes to update. So I wrote it to load from `unpkg.com`.  I would be happy to use cloudflare for 0.9.0 and then change back to unpkg for future releases, if that turns out to be the work flow.

I have a updated HTML page prepared for 0.9.0, but I usually don't submit a PR to the `gh-pages` branch until all my pending `master` PRs have landed. That way, I avoid any risk that my rudimentary Git skills will mess up my repository.

edemaine:
I guess the question is why `index.html` uses

```html
    <link href="bower_components/katex/dist/katex.min.css" type="text/css" rel="stylesheet">
    <script src="bower_components/katex/dist/katex.min.js" type="text/javascript"></script>
    <link href="bower_components/normalize.css/normalize.css" type="text/css" rel="stylesheet">
    <link href="bower_components/font-awesome/css/font-awesome.min.css" type="text/css" rel="stylesheet">
```

while `function-support.html` uses

```html
        <link rel="stylesheet" href="https://unpkg.com/katex@0.9.0/dist/katex.min.css">
        <script src="https://unpkg.com/katex@0.9.0/dist/katex.min.js"></script>
        <script src="https://unpkg.com/katex@0.9.0/dist/contrib/auto-render.min.js"></script>
```

I think one should change to match the other.

ronkok:
@edemaine, Once or twice, the function support page has been a release ahead of the demo page, at times when the demo page was not updated at release time.  So for the function support page, I got into the habit of using either the cloudflare CDN or unpkg.

edemaine:
@ronkok Changing both to unpkg (with possibly different versions) would seem reasonable then, if they are indeed that fast.

edemaine:
~~Unfortunately, https://unpkg.com/katex@0.9.0/dist/katex.min.js seems to fail to load, even though it's listed on https://unpkg.com/katex@0.9.0/dist/ (and looks fine on an npm install).~~  Hmm, now it works...

edemaine:
An alternative to this PR is #1183.

kevinbarabash:
Closing in preference of #1183.

PR #1182, opened 2018-02-27
Add unicode-fonts as submodule

ylemkimon:
\+ Removed checkout process from screenshotter.sh.

I think it's better to have them as a submodule, since git and GitHub handle it very well and we already have one submodule. Also, it's cross-platform and no extra manual step is needed when running `screenshotter.js` directly.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,6 @@
 [submodule "submodules/katex-fonts"]
 	path = submodules/katex-fonts
 	url = https://github.com/KaTeX/katex-fonts
+[submodule "submodules/unicode-fonts"]
+	path = submodules/unicode-fonts
+	url = https://github.com/Khan/KaTeX-test-fonts.git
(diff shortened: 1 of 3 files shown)


.gitmodules
```diff
@@ -1,3 +1,6 @@
 [submodule "submodules/katex-fonts"]
 	path = submodules/katex-fonts
 	url = https://github.com/KaTeX/katex-fonts
+[submodule "submodules/unicode-fonts"]
+	path = submodules/unicode-fonts
+	url = https://github.com/Khan/KaTeX-test-fonts.git
```
kevinbarabash: I can move this repo to the `KaTeX` organization.

kevinbarabash: It probably makes it sense to have it at `KaTeX/test-fonts` or `KaTeX/unicode-test-fonts`, wdyt?

edemaine: I think either is fine.

ylemkimon: As its current name is `KaTeX-test-font` and in the future maybe we can add other ASCII fonts support, I move for the former.

kevinbarabash:
@ylemkimon I've move the test font repo to https://github.com/KaTeX/katex-test-fonts.  Can you change the submodule folder to `katex-test-fonts` when you update the url?

PR #1183, opened 2018-02-27
Switch from bower to unpkg version of KaTeX

edemaine:
This PR is an alternative to #1180.  It would make updating in the future, but relies on unpkg.com.  (Of course, we could always switch back if we have trouble with unpkg.com in the future.)


edemaine:
One advantage of this PR is that it makes it transparent which version of KaTeX we're using in the demo site, which has been a point of confusion in the past.

kevinbarabash:
Let's give unpkg.com a try.

PR #1193, opened 2018-03-03
Add display/displayMode toggle on test page

### LINKED ISSUE #1035: Add a toggle to the test page to toggle displayMode option

edemaine:
* Fix #1035
* Refactor query parsing code to promote re-use
* Also add support for (unescaped) single-character macros


static/main.js
```diff
@@ -5,6 +5,14 @@
 import katex from '../katex.webpack.js';
 import './main.css';
 
+function argumentRegex(keyRegex) {
+    return new RegExp(`(?:^\\?|&)(${keyRegex})=([^&]*)`, "g");
+}
+
+function argument(keyRegex) {
+    return argumentRegex(keyRegex).exec(window.location.search);
+}
```
ylemkimon: Would it be possible to document functions?

static/main.js
```diff
@@ -13,32 +21,38 @@ function init() {
...
     while ((match = macroRegex.exec(window.location.search)) !== null) {
         options.macros = macros;
-        macros["\\" + match[1]] = decodeURIComponent(match[2]);
+        macros[match[1].replace(/%5[cC]/, "\\")] = decodeURIComponent(match[2]);
         macroString += "&" + match[0].substr(1);
     }
 
+    // Use `display=0` or `displayMode=0` to turn off displayMode.
+    if ((match = argument("display|displayMode"))) {
+        if (match[2] === "false" || match[2] === "0" || match[2] === "no") {
```
ylemkimon: I think `n` and maybe `f` are more common.

static/main.js
```diff
@@ -13,32 +21,38 @@ function init() {
...
+        input.value = decodeURIComponent(match[2]);
     }
 
     const macros = {};
-    // TODO: Add toggle for displayMode.
-    // https://github.com/Khan/KaTeX/issues/1035
     const options = {displayMode: true};
-    const macroRegex = /(?:^\?|&)(?:\\|%5[Cc])([A-Za-z]+)=([^&]*)/g;
+    const macroRegex = argumentRegex("(?:\\\\|%5[Cc])[A-Za-z]+|.");
     let macroString = "";
```
ylemkimon: The name `macroString` seems to be misleading.

kevinbarabash: Maybe `optionsString` may be more appropriate.

static/main.js
kevinbarabash: I think these are fine.  TBH I would've been okay with only checking for `displayMode=0`.

static/main.js
```diff
@@ -13,32 +21,38 @@ function init() {
...
+    if ((match = argument("text"))) {
+        input.value = decodeURIComponent(match[2]);
     }
 
     const macros = {};
-    // TODO: Add toggle for displayMode.
-    // https://github.com/Khan/KaTeX/issues/1035
     const options = {displayMode: true};
-    const macroRegex = /(?:^\?|&)(?:\\|%5[Cc])([A-Za-z]+)=([^&]*)/g;
+    const macroRegex = argumentRegex("(?:\\\\|%5[Cc])[A-Za-z]+|.");
```
kevinbarabash: I didn't realize that we supported macros as URL params, that's pretty cool.

static/main.js
kevinbarabash: I would probably refactor things to parse all the URL params to an object using a library from npm.  Then we could have simpler regexes for just the keys to check if there are any macros and the rest become simple key lookups.

edemaine:
Good comments. Most of these are inherited from the old code, but this is probably a fine time to clean it up.

static/main.js
edemaine: I added support for anything starting with 0, n, or f.  Hopefully less guesswork in the future.

static/main.js
edemaine: This has turned out to simplify the code a lot.  Good idea, @kevinbarabash!

static/main.js
edemaine: I've rewritten the code to avoid this recopying of the search string.

edemaine:
I implemented the comments, which cleaned up the code a lot. I also added some small features:
* `pre` and `post are aliases for `before` and `after`. This matches the option names in the screenshotter, so seems nice.
* Macros with escaped characters should now work, thanks to the `query-string` package.

I have not implemented a GUI for these options yet; maybe best for another PR?

kevinbarabash:
> I have not implemented a GUI for these options yet; maybe best for another PR?

SGTM

ylemkimon:
The file permissions of `package.json` and `package-lock.json` seems to be changed.

edemaine:
@ylemkimon Thanks for spotting that! (darn Windows)  Should be fixed now.

kevinbarabash:
@ylemkimon thanks for reviewing.  Feel free to merge PRs.

PR #1194, opened 2018-03-03
Work around negative space bug in Chrome

### LINKED ISSUE #1191: Inline \KaTeX is too wide in 0.9.0
From https://beta.observablehq.com/@mbostock/katex-0-9-0-spacing-bug:

image

Here I’m using KaTeX to render its logo (`\KaTeX`). In 0.8.3, the resulting span was 56.28px wide. In 0.9.0, it’s 72.31px wide. It seems that 0.9.0 makes use of negative right margins, and unfortunately it seems that the negative right margins of descendants of the root span do not accumulate as expected, at least in Chrome and Safari. In Firefox the inline element has the expected width.

edemaine:
Fix #984 and fix #1191 by working around [this Chromium bug](https://bugs.chromium.org/p/chromium/issues/detail?id=818440).

There seem to be two effects of the Chrome negative spacing bug:
* Within a KaTeX equation, we get spans that are too wide (e.g. in the box).  ~~This is fixed by changing `.mspace` from `display:inline-block` to `display:inline`. Alas, while this fixes negative spaces, it breaks (destroys) positive spaces.~~
* A non-display KaTeX equation leaves extra horizontal space on the right, affecting surrounding text.  ~~This is fixed by setting `.base` to have `width: min-content` as suggested by @ronkok in #984.~~

Both are fixed by `.katex` to have `width: min-content` as suggested by @ronkok in #984.  This is automatically inherited by all elements that don't define an explicit width like `width: 100%`.

The included screenshot test is still somewhat defective, showing a space before the period.  This seems limited to the screenshotter, however, as it works fine in the tester via http://localhost:7936/?text=%5Cfbox%7B%5C%241%2C%5C!000%2C%5C!000%7D%5CKaTeX&display=0&after=. (with #1193 merged in for no-display-mode support).  I tried upgrading the screenshotter (selenium/standalone-chrome from 2.48.2 to latest), but I could still not get the screenshotter to align with actual Chrome's rendering.

Here's an illustration of how it works with the tester:
image

Old screenshot test for comparison:
negativespace-chrome-bad

You can also see the problem in two places on the KaTeX front page (note space before colon!):

image

Manually changing the CSS in Chrome results in a good rendering:
image


kevinbarabash:
I'm not sure why the circleci tests are failing but it doesn't matter b/c they aren't required checks.

kevinbarabash:
@edemaine thanks for the PR.

PR #1195, opened 2018-03-05
Port spacingData and unicodeSymbols to @flow.

marcianx:


src/spacingData.js
```diff
@@ -1,22 +1,41 @@
+// @flow
 /**
  * Describes spaces between different classes of atoms.
  */
 
-const thinspace = {
+export type Space = {| number: number, unit: "mu" |};
```
ylemkimon: I think we can use `type {Measurement} from "./units";`.

src/spacingData.js
```diff
@@ -1,22 +1,41 @@
...
+    mrel: Space,
+    mopen: Space,
+    mclose: Space,
+    mpunct: Space,
+    minner: Space,
+}>;
+
 // Spacing relationships for display and text styles
-export const spacings = {
+export const spacings: {[string]: Spacings} = {
```
ylemkimon: Is it possible to flowtype the keys of `spacings`?

src/spacingData.js
```diff
@@ -1,22 +1,41 @@
...
+const thickspace: Space = {
     number: 5,
     unit: "mu",
 };
 
+// Making the type below exact with all optional fields doesn't work due to
+// - https://github.com/facebook/flow/issues/4582
+// - https://github.com/facebook/flow/issues/5688
+// However, since *all* fields are optional, $Shape<> works as suggested in 5688
+// above.
```
kevinbarabash: Interesting.  Thanks for adding links back to the issues.

src/spacingData.js
kevinbarabash: Could make this:
```
export const spacings: {$Keys<Spacings>: Spacings}
```

src/spacingData.js
```diff
@@ -69,7 +88,7 @@ export const spacings = {
 };
 
 // Spacing relationships for script and scriptscript styles
-export const tightSpacings = {
+export const tightSpacings: {[string]: Spacings} = {
```
kevinbarabash: same

src/spacingData.js
marcianx: Ooooo! Nice!

src/spacingData.js
marcianx: Done.

src/spacingData.js
marcianx: I want to add one thing before this can be submitted. Per the discussion at https://github.com/facebook/flow/issues/4582, I need to add a `& {}` to the `$Shape<>` type since `$Shape<>` seems to currently allow `null|undefined`. Feel free to do this yourself before merging (if everything else is LG). Otherwise, I'll do it tonight,

marcianx: Done fixed. It's ready now.

marcianx:
Good point! Made it exact.

src/spacingData.js
```diff
@@ -1,22 +1,40 @@
+// @flow
 /**
  * Describes spaces between different classes of atoms.
  */
+import type {Measurement} from "./units";
 
-const thinspace = {
+const thinspace: Measurement = {
```
kevinbarabash: I guess there's no reason to restrict these to `unit: "mu"`.  Thanks for switching to `Measurement`.

src/units.js
```diff
@@ -37,7 +37,7 @@ const relativeUnit = {
     "mu": true,
 };
 
-export type Measurement = {number: number, unit: string};
+export type Measurement = {| number: number, unit: string |};
```
kevinbarabash: I love exact types.  They catch way more issues.

PR #1197, opened 2018-03-08
Webpack Bundle Analysis

rrandallcainc:
Since webpack has access to the dependency tree, there are some nice plugins that can help visualize the data. This can be especially useful when trying to decrease the size of the bundle, or look for any red flags. 

I figured it would be good to open this up to see if others would find it useful.

The below image represents it's output. You can hover over items, and click to drill down.

screen shot 2018-03-08 at 2 45 47 pm


kevinbarabash:
@rrandallcainc thanks for the PR.

edemaine:
Cool!

kevinbarabash:
It's a good thing we split things into separate files in the `functions` folder otherwise we'd just see a couple of big blobs.

PR #1198, opened 2018-03-08
Documenting String.raw

rrandallcainc:
`String.raw` is a nice little gem that allows users to input latex in string without having to double backslack. Since it's not obvious, thought it would be good to document.


ylemkimon:
I think it should be noted that it's ES6 feature, which may not be supported in IE and old versions, and it may not include `${`, `\u`, and `\x`.

README.md
```diff
@@ -29,6 +29,11 @@ Call `katex.render` with a TeX expression and a DOM element to render into:
 katex.render("c = \\pm\\sqrt{a^2 + b^2}", element);
 ```
 
+Note: If you don't want to double backslack to escape, you can use `String.raw`
```
edemaine: Shouldn't that be "double backslash"? Or is this a word I don't know?

rrandallcainc: If it is, I don't know it either. Apparently I'm trying to add subliminal messaging for slack.

edemaine: Industrial sponsorship! 😄

rrandallcainc:
@ylemkimon These are good points. I just happened to find `String.raw` useful when working with latex in my projects. Perhaps we could link to the mdn docs, and users can leverage based on their use case. There is also a babel polyfill that users can leverage if they need to support older browsers. 

I could be convinced that this isn't worth documenting either. Perhaps this is common knowledge to folks, or not as useful as I think.

edemaine:
I think this is a good idea to document, as LaTeX is so heavy with `\`s.  LInking to [the MDN docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/raw) seems good.

The edge cases @ylemkimon mentioned would also be good to mention, as they would mess with things like `\underline`, and groups in nested math `${`.  I find that `String.raw`\u1234`.length === 6` in browser console, but apparently it doesn't work in ` .js` file. ☹️

kevinbarabash:
The 2018 revision of template tags address `\u` and `\x`.  It's unfortunate about `${}`.

rrandallcainc:
Hmm, I'll look to see if there's a better way to do this, without suffering the edge cases. If not, I can update the docs.

edemaine:
@rrandallcainc Barring some magic solution, perhaps you can document the edge cases and we can get this merged?

rrandallcainc:
Sorry, I really have not had time to work on this project. Going to close this now to reduce noise, and will make a note to re-open when I have time. Thanks for checking in.

edemaine:
No worries.  I'll take a crack at revising them, then.

edemaine:
OK, this is ready for review.

PR #1200, opened 2018-03-09
Fix \nobreakspace, refactor "regular space" checking

### LINKED ISSUE #1199: \nobreakspace seems to be broken
I'm getting error when trying to use it:
```
TypeError: Cannot read property 'className' of undefined
```
On https://khan.github.io/KaTeX/ there is an error too (but different one).

edemaine:
Fix #1199 by adding `\nobreakspace` to the list of commands/symbols that behave like a regular space character.  Refactor to put that list in one place, and use an object so it checks instantly instead of sequentially.

I also added a test (making an existing `\nobreakspace` test check for `toBuild`, which is stronger than `toParse`) that I confirmed didn't work before this PR.


src/buildCommon.js
```diff
@@ -682,6 +682,16 @@ const spacingFunctions: {[string]: {| size: string, className: string |}} = {
     },
 };
 
+// A lookup table to determine whether a spacing function/symbol should be
+// treated like a regular space character.
+const regularSpace: {[string]: boolean} = {
```
rrandallcainc: This is a much cleaner approach 👍

test/katex-spec.js
```diff
@@ -2999,11 +2999,11 @@ describe("The \\mathchoice function", function() {
 
 describe("Symbols", function() {
     it("should parse \\text{\\i\\j}", () => {
-        expect("\\text{\\i\\j}").toParse();
+        expect("\\text{\\i\\j}").toBuild();
```
rrandallcainc: Just curious, why did these change?

edemaine: No reason; it would be fine as is, too.  But `toBuild` also checks `toParse`, so this seemed like a better test.

kevinbarabash:
@edemaine thanks for the PR.  @rrandallcainc thanks for reviewing.

PR #1201, opened 2018-03-09
Fix \nobreakspace

ronkok:
Fixes issue #1199


edemaine:
Oops, duplicate of #1200.  We both should have claimed the issue by self-assigning...

This PR has an additional test (though `toParse` should be `toBuild` -- `toParse` wouldn't have failed before), which I could copy over to #1199.  #1199 has a nicer table.

ronkok:
@edemaine No problem. You can copy the test over if you like. I'll close this PR.

edemaine:
Copied, thanks!

PR #1205, opened 2018-03-11
Support Unicode delimiters

ronkok:
This PR maps these delimiter characters to exactly the same functions as does `unicode-math-table.txt`.


edemaine:
@ronkok Would you mind posting a screenshot here on Github demoing the new behavior on the test server?  I feel like the main thing to double-check here is the correspondence between Unicode characters and the correct LaTeX, and this would make it particularly easy.

ylemkimon:
Is it possible to use datas from `symbols` from `src/symbols.js`?

ronkok:
@edemaine Good call.

delimiters

ronkok:
> Is it possible to use datas from `symbols` from `src/symbols.js`?

I don't understand the question. What sort of data do you seek?

kevinbarabash:
@ronkok I think what @ylemkimon may have been getting at was that in symbols.js the call to `defineSymbol` links a latex command and its unicode character.  It would be nice if we could use that information in delimiter.js instead of having to repeat the unicode character there as well.

Aside: I thought I added support for some of the other unicode delimiters like the left/right triangular brackets.  I can't seem to find the PR or the code anywhere. :(

ronkok:
`delimiter.js` is called only by the functions that resize delimiters. So delimiters like `ulcorner` and `urcorner`that do not change size do not  appear in `delimiter.js`.  But the ones that do change size need a little more definition applied for the size change.

ronkok:
I'm closing this PR so that I can clean up the master branch in my repository. I'll reopen another delimiter PR when that's done.

ylemkimon:
> I think what @ylemkimon may have been getting at was that in symbols.js the call to defineSymbol links a latex command and its unicode character. It would be nice if we could use that information in delimiter.js instead of having to repeat the unicode character there as well.

Yes, that's what I meant 😄 Sorry for being unclear.

ronkok:
Okay, I understand the question now. But you know, I'm probably not the person you want to have refactoring your code. I'll follow existing patterns and support new functions. I'll chase down rendering bugs. But a code clean up operation should be done be someone with a stronger claim to expertise.

PR #1208, opened 2018-03-12
Add \yen support to text mode

### LINKED ISSUE #1204: Currency symbols not allowed inside \text
For example \yen works, but not \text{¥}

edemaine:
Fix #1204 by adding `\yen` support to text mode.

More generally, I checked all the text+mode symbols supported by [amsfonts](http://mirrors.ctan.org/fonts/amsfonts/doc/amsfonts.pdf) (see Section 6), and found that `\yen` is the only one lacking text-mode support.

Here's a screenshot of the tester confirming support:

image

### DIFF
--- a/test/katex-spec.js
+++ b/test/katex-spec.js
@@ -2612,6 +2612,16 @@ describe("The symbol table integrity", function() {
     });
 });
 
+describe("Symbols", function() {
+    it("should support AMS symbols in both text and math mode", function() {
+        // These text+math symbols are from Section 6 of
+        // http://mirrors.ctan.org/fonts/amsfonts/doc/amsfonts.pdf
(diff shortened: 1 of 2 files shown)


test/katex-spec.js
```diff
@@ -2612,6 +2612,16 @@ describe("The symbol table integrity", function() {
     });
 });
 
+describe("Symbols", function() {
+    it("should support AMS symbols in both text and math mode", function() {
+        // These text+math symbols are from Section 6 of
+        // http://mirrors.ctan.org/fonts/amsfonts/doc/amsfonts.pdf
+        const symbols = "\\yen\\checkmark\\circledR\\maltese";
+        expect(symbols).toBuild();
+        expect(`\\text{${symbols}}`).toBuild();
```
kevinbarabash: cool

Issue #1027, opened 2017-12-22
Single-symbol macro arguments


edemaine:
PR #1024 added support for detecting text or math mode (e.g. via `\TextOrMath`) within a macro, not just in the parser.  There is one edge case that doesn't work yet: in a function's argument that has a specified argument type (`math` or `text`) and when that argument is given by a single-symbol macro.

For example, `test/katex-spec.js` has a commented-out test that should work but doesn't:
```js
    it("\\TextOrMath should work in a macro passed to \\text", function() {
        compareParseTree("\\text\\mode", "\\text{text}",
            {"\\mode": "\\TextOrMath{text}{math}"});
    });
```

This test has two issues. Because of #924, only the first letter of `\mode` gets passed to the `\text` function. Because of this issue, that letter is `m` instead of `t`, because the mode gets set right after `\mode` gets expanded by the `MacroExpander`.  The offender is the `else` case of `parseGroup` (when it gets called by `parseGroupOfType`, but to fix this, we'll have to avoid having already read the symbol by the time `parseGroup` gets called.

I'm hoping that both issues will be fixed by one idea: when the `MacroExpander` encounters a function (like `\text`), it should grab the arguments *at the same time* (and in the right mode), and return everything to the parser, instead of just the function token.  This is probably roughly what TeX does too, for built-in commands.

PR #1213, opened 2018-03-13
Support \( and fix instant mode switching for $, \(, \text

### LINKED ISSUE #1212: \text accepts nested $...$ fragments but not \(...\)
In LaTeX and MathJax, this works fine:
```
\[
  x = z \quad \text{because $x = y$ and \(y = z\)}
\]
```
KaTeX doesn't like the nested `\(...\)` fragment, though.

This is [documented](https://khan.github.io/KaTeX/function-support.html) and probably well-known, but I didn't find an issue on it.

This could probably be generalized to accepting only the delimiters passed in to `renderMathInElement` with `display: false`, but I don't know if the structure of the code makes that easy.

edemaine:
* Fix #1212 by supporting `\(...\)` in addition to `$...$` inline math nested inside `\text`. This turned out to be harder than one would think because of the way `$` was being handled specially, while `\(` needed to be handled as a function, which immediately consumed the `\(` token, which meant that the first following token got consumed in text mode instead of math mode (unlike `$` where we could switch the mode before consumption).
* Added new `consumeMode` setting for functions which tells the parser to switch modes just for the consumption of the command token.
* Now that we're working with functions, move all the `$` handling code into `functions/math.js`.  Somewhat bizarrely, we can define `$` as a function (nearly identical to `\(`) and it all works (just like we can have single-character macros).  This move removed a lot of messy stuff from `Parser.js`: `ParsedDollar`, `ParsedFuncOrArgOrDollar`, `newDollar`, `assertFuncOrArg`, and at least three explicit checks for `if (... === "$")`.
* Moved the `consume()` for the command token from `parseSymbol` to `parseGivenFunction`.  This ended up not being strictly necessary, but seems conceptually cleaner.
* Partially address #1027 by setting `consumeMode: "text"` for `\text` commands.  As a result, `\text` now instantly switches the mode to `"text"`, so even an unbraced macro argument will properly detect text mode.  I added a test for this, which is a slightly weaker form of #1027.

### DIFF
--- a/src/Parser.js
+++ b/src/Parser.js
@@ -136,7 +136,9 @@ export default class Parser {
      * and fetches the one after that as the new look ahead.
      */
     consume() {
+        console.log('before consume', this.nextToken);
         this.nextToken = this.gullet.expandNextToken();
+        console.log('after consume', this.nextToken);
     }
 
     /**
@@ -455,18 +457,22 @@ export default class Parser {
 
         const func = start.result;
 
-        if (func === "$") {
+        if (func === "$" || func === "\\(") {
+            console.log(func);
             if (this.mode === "math") {
-                throw new ParseError("$ within math mode");
+                throw new ParseError(`${func} within math mode`);
             }
             const outerMode = this.mode;
             this.switchMode("math");
             // Expand next symbol now that we're in math mode.
             this.consume();
-            const body = this.parseExpression(false, "$");
-            // We can't expand the next symbol after the $ until after
+            const close = (func === "\\(" ? "\\)" : func);
+            console.log('parsing until', close);
+            const body = this.parseExpression(false, close);
+            console.log(body);
+            // We can't expand the next symbol after the closing $ until after
             // switching modes back.  So don't consume within expect.
-            this.expect("$", false);
+            this.expect(close, false);
             this.switchMode(outerMode);
             this.consume();
             return new ParseNode("styling", {
@@ -940,7 +946,9 @@ export default class Parser {
         const nucleus = this.nextToken;
         let text = nucleus.text;
 
-        if (functions[text]) {
+        if (text === "$" || text === "\(") {
+            return newDollar(nucleus);
+        } else if (functions[text]) {
             this.consume();
             // If there exists a function with this name, we return the function and
             // say that it is a function.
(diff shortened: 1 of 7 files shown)


test/katex-spec.js
```diff
@@ -872,7 +871,40 @@ describe("A text parser", function() {
...
+    });
+
+    it("should detect unbalanced $", function() {
+        expect("$").toNotParse();
+        expect("\\text{$}").toNotParse();
+    });
+
+    it("should not mix $ and \\(..\\)", function() {
+        expect("$x\\)").toNotParse();
+        expect("\\(x$").toNotParse();
```
ylemkimon: Shouldn't be these two in the text mode like:
```js
        expect("\\text{$x\\)}").toNotParse();
        expect("\\text{\\(x$}").toNotParse();
```
?

src/Parser.js
```diff
@@ -964,8 +937,6 @@ export default class Parser {
                     body: arg,
                     star: star,
                 }, "text"), nucleus);
-        } else if (text === "$") {
-            return newDollar(nucleus);
```
ylemkimon: ❤️ that they are gone and became so simple!

src/functions/math.js
```diff
@@ -0,0 +1,44 @@
+// @flow
+import defineFunction from "../defineFunction";
+import ParseError from "../ParseError";
+
+// Switching from text mode back to math mode
+defineFunction({
+    names: ["\\(", "$"],
```
ylemkimon: 👍

test/katex-spec.js
edemaine: Yes, thanks for catching this!  (Otherwise we were just testing use of `$`/`\(` outside math mode again.)  Fixed.

edemaine:
@ylemkimon Thanks for the quick review! Should be good to go now.

src/functions/math.js
```diff
@@ -0,0 +1,44 @@
...
+        // switching modes back.  So don't consume within expect.
+        parser.expect(close, false);
+        parser.switchMode(outerMode);
+        parser.consume();
+        return {
+            type: "styling",
+            style: "text",
+            value: body,
+        };
+    },
```
kevinbarabash: This isn't too bad and seems to follow the same pattern as `\left`/`\right` handling.

Issue #1075, opened 2018-01-18
 \begin{bmatrix} does not work in displaymode=false


kiliansinger:
Using bmatrix in displaymode=false does not render 
example:
<!DOCTYPE html>
<html>
  <head>
    <meta charset="UTF-8">
    <title>Auto-render test</title>
<link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/KaTeX/0.9.0-beta/katex.min.css" integrity="sha384-L/SNYu0HM7XECWBeshTGLluQO9uVI1tvkCtunuoUbCHHoTH76cDyXty69Bb9I0qZ" crossorigin="anonymous">
<script src="https://cdnjs.cloudflare.com/ajax/libs/KaTeX/0.9.0-beta/katex.min.js" integrity="sha384-ad+n9lzhJjYgO67lARKETJH6WuQVDDlRfj81AJJSswMyMkXTD49wBj5EP004WOY6" crossorigin="anonymous"></script>
<script src="https://cdnjs.cloudflare.com/ajax/libs/KaTeX/0.9.0-beta/contrib/auto-render.min.js" integrity="sha384-EkJr57fExjeMKAZnlVBuoBoX0EJ4BiDPiAd/JyTzIA65ORu4hna7V6aaq4zsUvJ2" crossorigin="anonymous"></script>
  
  </head>
  <body>
    <div id="test">
\[ \begin{bmatrix} -(D_1+D_2) &amp; D_2 &amp; 0\\[0.3em] D_2        &amp; -(D_2+D_3) &amp; D_3 \\[0.3em] 0 &amp; D_3 &amp; -(D_3+D_4) \end{bmatrix} \]
we have ...  \( \begin{bmatrix} -(D_1+D_2) &amp; D_2 &amp; 0\\[0.3em] D_2        &amp; -(D_2+D_3) &amp; D_3 \\[0.3em] 0 &amp; D_3 &amp; -(D_3+D_4) \end{bmatrix} \)

    </div>
    <script>
      renderMathInElement(
          document.getElementById("test"),
          {
              delimiters: [
                  {left: "$$", right: "$$", display: true},
                  {left: "\\[", right: "\\]", display: true},
                  {left: "$", right: "$", display: false},
                  {left: "\\(", right: "\\)", display: false}
              ]
          }
      );
    </script>
  </body>
</html>

ronkok:
Well, that's not good. The problem seems to come from the row spacing code.

[...]

PR #1214, opened 2018-03-14
Update function support page for release v0.9.0

ronkok:


ylemkimon:
The console warns:
```
katex.min.js:1 No character metrics for '¶' in style 'Main-Regular'
katex.min.js:1 No character metrics for '§' in style 'Main-Regular'
```

ylemkimon:
Font metrics don't seem to be updated after https://github.com/KaTeX/katex-fonts/pull/23.

CC @kevinbarabash

edemaine:
@ylemkimon Uh oh. Nice find. Confirmed that this problem happens on the demo page too, as well as in the `npm start` test server. Perhaps should move this to a separate issue, but it's clearly a bug we need to fix. I've lost track: are metrics now in the katex or katex-fonts repo?

ylemkimon:
@edemaine It's in the `katex-fonts` 😄

function-support.html
```diff
@@ -5,12 +5,12 @@
...
+	<link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/KaTeX/0.9.0/katex.min.css" integrity="sha384-TEMocfGvRuD1rIAacqrknm5BQZ7W7uWitoih+jMNFXQIbNl16bO8OZmylH/Vi/Ei" crossorigin="anonymous">
+	<script src="https://cdnjs.cloudflare.com/ajax/libs/KaTeX/0.9.0/katex.min.js" integrity="sha384-jmxIlussZWB7qCuB+PgKG1uLjjxbVVIayPJwi6cG6Zb4YKq0JIw+OMnkkEC7kYCq" crossorigin="anonymous"></script>
+	<script src="https://cdnjs.cloudflare.com/ajax/libs/KaTeX/0.9.0/contrib/auto-render.min.js" integrity="sha384-IiI65aU9ZYub2MY9zhtKd1H2ps7xxf+eb2YFG9lX6uRqpXCvBTOidPRCXCrQ++Uc" crossorigin="anonymous"></script>
 
-	<link rel="stylesheet" href="https://unpkg.com/katex@0.9.0-beta/dist/katex.min.css">
-	<script src="https://unpkg.com/katex@0.9.0-beta/dist/katex.min.js"></script>
-	<script src="https://unpkg.com/katex@0.9.0-beta/dist/contrib/auto-render.min.js"></script>
-
-<style type="text/css">
+<style>
```
edemaine: Is there a reason that you dropped `type="text/css"` here?  I'm not sure it matters.

function-support.html
```diff
@@ -5,12 +5,12 @@
     <meta name="viewport" content="initial-scale=1">
     <title>Function Support in KaTeX</title>
     <link rel="icon" href="http://khan.github.io/favicon.ico">
+	
+	<link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/KaTeX/0.9.0/katex.min.css" integrity="sha384-TEMocfGvRuD1rIAacqrknm5BQZ7W7uWitoih+jMNFXQIbNl16bO8OZmylH/Vi/Ei" crossorigin="anonymous">
```
edemaine: We switched to `unpkg.com` for the main page, so presumably we should continue to follow suit here?

function-support.html
```diff
@@ -579,7 +591,7 @@ <h2 id="environments">Environments</h2>
     <div class="indented">
 		<p>KaTeX also supports <code>{darray}</code> and <code>{dcases}</code>.</p>
 	
-        <p>Acceptable line separators include: <code>\\</code>, <code>\cr</code>, and <code>\\[<em>distance</em>]</code>. <em>Distance</em> can be written with any of the <a href="#units">KaTeX units</a>.</p>
+        <p>Acceptable line separators include: <code>\\</code>, <code>\cr</code>, and <code>\cr[<em>distance</em>]</code>. <em>Distance</em> can be written with any of the <a href="#units">KaTeX units</a>.</p>
```
edemaine: Mentioning `\cr[distance]` seems good, but `\\[distance]` is valid too, right?

function-support.html
```diff
@@ -980,8 +1070,9 @@ <h3 id="spacing">Spacing</h3>
     </table>
 	<table>
 		<tr>
-			<td>Notes:</td><td>{<em>distance</em>} will accept any of the <a href="#units">KaTeX units</a>.<br>
-			<code>\mkern</code> and <code>\mskip</code> will not work in text mode and both will write a console warning for any unit except <em>mu</em>.</td>
+			<td style="vertical-align: top;">Notes:</td><td><p>{<em>distance</em>} will accept any of the <a href="#units">KaTeX units</a>.</p>
```
edemaine: This would look better (top aligned) if the first `<p>` gets a `style="margin_top: 0"`, or an analogous CSS rule to remove the top margin from the first `<p>` within a `<td>`.

function-support.html
```diff
@@ -1848,7 +1939,8 @@ <h2 id="relations">Relations</h2>
 	<table>
 		<tr>
 			<td>Direct Input:</td>
-			<td style="font-family: KaTeX_Main, KaTeX_AMS;">= &lt; &gt; : ∈ ∋ ∝ ∼ ∽ ≂ ≃ ≅ ≈ ≊ ≍ ≎ ≏ ≐ ≑ ≒ ≓ ≖ ≗ ≜ ≡ ≤ ≥ ≦ ≧ ≫ ≬ ≳ ≷ ≺ ≻ ≼ ≽ ≾ ≿ ⊂ ⊃ ⊆ ⊇ ⊏ ⊐ ⊑ ⊒ ⊢ ⊣ ⊩ ⊪ ⊸ ⋈ ⋍ ⋐ ⋑ ⋔ ⋙ ⋛ ⋞ ⋟ ⌢ ⌣ ⩾ ⪆ ⪌ ⪕ ⪖ ⪯ ⪰ ⪷ ⪸ ⫅ ⫆</td>
+			<td><span style="font-family: KaTeX_Main, KaTeX_AMS;">= &lt; &gt; : ∈ ∋ ∝ ∼ ∽ ≂ ≃ ≅ ≈ ≊ ≍ ≎ ≏ ≐ ≑ ≒ ≓ ≖ ≗ ≜ ≡ ≤ ≥ ≦ ≧ ≫ ≬ ≳ ≷ ≺ ≻ ≼ ≽ ≾ ≿ ⊂ ⊃ ⊆ ⊇ ⊏ ⊐ ⊑ ⊒ ⊢ ⊣ ⊩ ⊪ ⊸ ⋈ ⋍ ⋐ ⋑ ⋔ ⋙ ⋛ ⋞ ⋟ ⌢ ⌣ ⩾ ⪆ ⪌ ⪕ ⪖ ⪯ ⪰ ⪷ ⪸ ⫅ ⫆
+			</span>≔ ≕ ⩴</td>
```
edemaine: Shouldn't these be in the `<span>`?

edemaine: Similarly, `ℂ ℍ ℕ ℙ ℚ ℝ ℤ` above seems to be outside the span, which causes them to render in the wrong font.

function-support.html
ronkok: I don't have a strong preference. Our READ.ME says to use the Cloudflare CDN, so I used it here to match. There have been a couple times when I have updated the function support page within one day of the release. I then use `unpkg.com` because the Cloudflare CDN can take 2 or 3 three days to propagate the files.

If you feel strongly about `unpkg.com`, I'll change it this time, too.

function-support.html
ronkok: Issue #1075 occurs when the auto-render extension encounters a `\\[0.3em]` row separator inside a pair of `\[...\]` math delimiters. Until we fix that bug, I thought I would nudge people toward `\cr[0.3em]`.

function-support.html
ronkok: Yep. I'll change it.

function-support.html
ronkok: I disagree.

I expect that some authors will copy-and-paste directly from the function support page into their text editor. So when I say that characters are available for "Direct Input", I need to supply characters at the correct Unicode code point. Otherwise, someone will copy what they think is a double-struck C, but will be a normal C when it lands in their text editor.

So I want to keep these characters as is. I'm sorry about the font, but I want to minimize author confusion.

function-support.html
edemaine: Ah, interesting.  [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/style) confirms that this is the default value, so definitely unnecessary.  Cool.

function-support.html
edemaine: I think it makes sense to match `index.html`, and that really does need quick updates (as it will coincide with release), so I'd strongly lean toward using `unpkg` here.

function-support.html
edemaine: Ah, good point.  Sounds good, as long as you'll remember to add it back once #1075 gets solved.

function-support.html
edemaine: Ah, I see.  That makes sense.  Sorry for missing the subtlety here (especially if it's been discussed before).

Issue #1215, opened 2018-03-14
𝔽 character in \text{} and elsewhere


yorrd:
it seems like this unicode character does not get interpreted and throws an error (unexpected token, the usual). 

General question: we're currently switching over from MathJax (performance...), and therefore have a LOT of formulas that used to work on MathJax but don't anymore (most notably the `\mathbb` stuff). I'm just curious: why don't you support the syntax and just resolve those symbols to unicode? That's what we're doing now with `\mathbb{R}`, unfortunately we have to use a standard *F* for the time being until this works. 

The same issue also applies to the double struck I and K and probably a lot of other chars.

edemaine:
I'll leave the Unicode double-struck characters for @ronkok to comment on, but it sounds like we should support those. One related question: does your formula work on LaTeX? That is our standard reference, not MathJax.

You can add support for them now by adding a macro mapping e.g. `"𝔽": "{\mathbb F}"`.

KaTeX has generally shied away from supporting characters incorrectly, but I agree that it would be nice to have a "just render it somehow" option. I can try to add one.

yorrd:
It does work in LaTeX. I would absolutely love that, would solve pretty much all our issues :)

ronkok:
@yorrd We've been adding Unicode characters in small steps. There were a couple of PRs two years ago that tried to add all characters at once. The discussion about those PRs got bogged down because a few characters have ambiguous mappings. Should ⊥ be a relation or a textord? KaTeX wanted to avoid getting any of those calls wrong and have to do breaking changes later.

So we've been adding characters in smaller, more digestible chunks of the Unicode table. And we haven't yet gotten to anything outside the Unicode basic multi-lingual plane.  That's why 𝔽 is not yet accepted. 

@edemaine There are nearly a thousand characters in the Unicode table for [Mathematical Alphanumeric Symbols](http://www.unicode.org/charts/PDF/U1D400.pdf). Do you have any ideas for doing that compactly? Or do we write 1000 macros?

ronkok:
>  Or do we write 1000 macros?

I don't mind writing 1000 macros. I can automate part of it. I just wonder if there is some elegant way to support those characters that will render faster and/or require us to ship less code.

edemaine:
@ronkok Thinking about the mathbb characters as a starting point (and maybe that's a good next Unicode PR to work on), I could see doing a for loop over 0..25 and constructing the Unicode and ASCII character codes in the loop (and still doing defineMacro in the loop).

[...]

PR #1217, opened 2018-03-14
Allow all Unicode symbols in nonstrict mode

### LINKED ISSUE #380: Don't generate an exception when an arbitrary symbol is used inside a formula
Hi Kevin and thanks for your work first;

current behaviour of KaTeX is such that most of the symols aside latin letters, numbers and special symbols cause `ParseError`. Couple of examples:

1. a formula with dimensions of quantities, in say, Russian, like `{1 Н\over 2 м^2} = 0,5 Па` (in English that would be: `{1 N\over m^2} = 0.5 Pa`)
2. a formula with an "already parsed" symbol like `A = ρgh` (this would be nice for TiddlyWiki as the editable content is this case is more readable)

It would be nice if those symbols are displayed "as is" (except for the font size). In fact, #379 is about this, too.

edemaine:
Fix #380.

Now that the `strict` is merged, I changed symbol parsing to allow all Unicode symbols when the appropriate strict allows it.  By default, this allows everything, but the user will get (probably two) warnings about them.  This is a big change to KaTeX (so might be good to get into 0.10.0), but I think it will be a welcome change to many users.

Old description:

For users who aren't worried about exact layout, but want to render all sorts of Unicode, the new `allowAllSymbols: true` setting allows users to get KaTeX to render something instead of failing (though it still should report a warning about missing character metrics).  Obviously this feature should be used with care, but it can help users get things working while we work to add more Unicode character support.

Based on discussions in #1215. @yorrd said "I would absolutely love that, would solve pretty much all our issues :)" I think this is a useful feature and should let @yorrd switch to KaTeX earlier, but please also create issues for any symbols you spot that we don't support but should (like the double-stroke characters of #1215), i.e., when you get the character metric warnings.


src/Parser.js
```diff
@@ -996,8 +996,11 @@ export default class Parser {
             }
             symbol = new ParseNode(symbols[this.mode][text].group,
                             text, this.mode, nucleus);
-        } else if (supportedCodepoint(text.charCodeAt(0)) &&
-            (this.mode === "text" || this.settings.unicodeTextInMathMode)) {
+        } else if ((this.settings.allowAllSymbols &&
+                    text.charCodeAt(0) >= 0x80) ||
```
edemaine: The `>= 0x80` check is to prevent things like `^` from getting interpreted as a character.  I'm open to other suggestions here.  (I thought about having a table of special symbols...)  But it seems unlikely that there are other interesting characters in the `0`-`0x7f` range that we don't already support.

src/Parser.js
```diff
@@ -996,8 +996,11 @@ export default class Parser {
...
+                   (supportedCodepoint(text.charCodeAt(0)) &&
+                    (this.mode === "text" ||
+                     this.settings.unicodeTextInMathMode))) {
```
ylemkimon: Should `allowAllSymbols` override `unicodeTextInMathMode`, i.e., supported unicode characters should be allowed in math mode even if `!unicodeTextInMathMode` when `allowAllSymbols`?

edemaine: I believe that is how it works now. Do you think it's counterintuitive? Another alternative would be to have `allowAllSymbols` take an argument of `"math"` or `"text"`, and `true` for both. Whatever we decide, the behavior should be documented.

ylemkimon:
This will fix original issue in #380.

src/Parser.js
ylemkimon: @edemaine I'm not sure. But as it's more counterintuitive to disallow _supported_ characters in math mode while allowing _unsupported_ characters, I think this PR's behavior is OK.

edemaine: I added documentation that this is the behavior.  Should be good to go now.

I think this issue would be cleaner if we moved to a default `true` value of `unicodeTextInMathMode`, or default "nonstrict" from #1226.

kevinbarabash:
At some point in the future we may want to provide a callback to allow users to provide metrics about characters they care about.

src/Parser.js
kevinbarabash: We should probably only have one setting so that people don't get confused about which setting to use or what the interaction between these two settings should be.  I think we should probably just introduce the `strict` setting in this diff in order to minimize the number of breaking changes.  We can transition the other settings to use the `strict` setting in a subsequent PR.

edemaine:
Now that the `strict` is merged, I rejiggered this PR to use it.  This is a big change to KaTeX: now all Unicode symbols will be allowed by default, but the user will get (probably two) warnings about them.  But I think it will be a welcome change to many users.

ronkok:
The next update to the function support page will be challenging. So far, that page has shown a lot but explained very little. I hope that I can explain `strict` and `allowAllSymbols` in a concise, understandable way.

edemaine:
@ronkok Perhaps you can just say something like "Symbols not listed on this page are officially unsupported by KaTeX, and will by default generate a warning. You can turn off the warning, or turn the warning into an error, by adjusting the `strict` setting as described in the README."

ronkok:
Yes, something along those lines is called for.  I'd appreciate it if you would look over the page when I turn in the next version.

README.md
```diff
@@ -77,6 +77,9 @@ You can provide an object of options as the last argument to `katex.render` and
...
 - `maxSize`: `number`. If non-zero, all user-specified sizes, e.g. in `\rule{500em}{500em}`, will be capped to `maxSize` ems. Otherwise, users can make elements and spaces arbitrarily large (the default behavior).
 - `strict`: `boolean` or `string` or `function` (default: `"warn"`). If `false` or `"ignore`", allow features that make writing LaTeX convenient but are not actually supported by (Xe)LaTeX (similar to MathJax). If `true` or `"error"` (LaTeX faithfulness mode), throw an error for any such transgressions. If `"warn"` (the default), warn about such behavior via `console.warn`. Provide a custom function `handler(errorCode, errorMsg, token)` to customize behavior depending on the type of transgression (summarized by the string code `errorCode` and detailed in `errorMsg`); this function can also return `"ignore"`, `"error"`, or `"warn"` to use a built-in behavior.  A list of such features and their `errorCode`s:
+  - `"unknownSymbol"`: `boolean`. Use of unknown Unicode symbol, which will
```
ylemkimon: Redundant `boolean`.

test/unicode-spec.js
```diff
@@ -137,6 +137,18 @@ describe("unicode", function() {
...
 
+    it("should not allow emoji", function() {
+        expect('✌').toNotParse(strictSettings);
+        expect('\\text{✌}').toNotParse(strictSettings);
+        expect('✌').toWarn();
+        expect('\\text{✌}').toWarn();
+    });
+
+    it("should allow emoji with allowAllSymbols", function() {
+        const settings = {allowAllSymbols: true};
```
ylemkimon: Test needs to be updated.

edemaine:
@ylemkimon Thanks for spotting that old code.  All fixed now!  I confirmed that there are now no references to `allowAllSymbols`.

PR #1218, opened 2018-03-15
Fix cli.js: add nomnom dependency, add -u option

### LINKED ISSUE #1216: CLI errors; nomnom belongs in dependencies rather than devDependencies
Since cli.js depends on nomnom but is listed in devDependencies, katex installed in the usual manner (ex. `yarn global add katex`) fails with an error about a missing module.

```bash
$ katex --help
module.js:559
    throw err;
    ^
Error: Cannot find module 'nomnom'
    at Function.Module._resolveFilename (module.js:557:15)
    at Function.Module._load (module.js:484:25)
    at Module.require (module.js:606:17)
    at require (internal/module.js:11:18)
    at Object.<anonymous> (/home/delta/.config/yarn/global/node_modules/katex/cli.js:9:17)
    at Module._compile (module.js:662:30)
    at Object.Module._extensions..js (module.js:673:10)
    at Module.load (module.js:575:32)
    at tryModuleLoad (module.js:515:12)
    at Function.Module._load (module.js:507:3)
```

edemaine:
* Add `nomnom` dependency (not dev dependency): fix #1216
* Add -u option for new `unicodeTextInMathMode` option
* Cleanup another doc string


cli.js
```diff
@@ -47,7 +47,14 @@ const options = require("nomnom")
         abbr: "b",
         flag: true,
         default: false,
-        help: "Restores the old behavior of (pre-0.8.0) KaTeX.",
+        help: "Restores the old (pre-0.8.0) behavior of \\color.",
```
ylemkimon: I think describing its behavior, i.e., `\color` will work like LaTeX's `\textcolor` is better.

edemaine: Sounds good; pushed a revision.

package.json
```diff
@@ -40,7 +40,6 @@
     "less": "~2.7.1",
```
ylemkimon: The permission(mode) of this file seems to be changed. Note that the executable bit differences can be ignored by `git config core.fileMode false`:
```
core.fileMode
       If false, the executable bit differences between the index and the
       working copy are ignored; useful on broken filesystems like FAT.
       See git-update-index(1). True by default.
```
(from `git-config(1)`)

cli.js
```diff
@@ -47,7 +47,14 @@ const options = require("nomnom")
         abbr: "b",
         flag: true,
         default: false,
-        help: "Restores the old behavior of (pre-0.8.0) KaTeX.",
+        help: "Makes \\color behave like LaTeX's 2-argument \\textcolor, instead of LaTeX's one-argument \\color mode change.",
```
ylemkimon: The line exceeds 84 letters.

package.json
edemaine: Thanks for the tip!  I'll do `git config --global core.fileMode false` on my Windows machines.

Both comments should be addressed now.

ylemkimon:
It's in 😄

PR #1220, opened 2018-03-15
Add regenerate option to the screenshotter

ylemkimon:
Sometimes it is useful to generate new screenshots and verify screenshots/generate diffs _at the same time_.

By passing `--regenerate` with `--verify`, it will generate new screenshots of those that has failed in `test/screenshotter/new`.


dockers/Screenshotter/screenshotter.js
```diff
@@ -3,6 +3,7 @@
 
 const childProcess = require("child_process");
 const fs = require("fs");
+const mkdirp = require("mkdirp");
```
ylemkimon: We already have `mkdirp`.

kevinbarabash:
I like this but won't have time to review this for a while.  If someone else could review this that would be much appreciated.

dockers/Screenshotter/screenshotter.js
```diff
@@ -80,6 +83,10 @@ const opts = require("nomnom")
         flag: true,
         help: "With `--verify`, produce image diffs when match fails",
     })
+    .option("regenerate", {
+        flag: true,
+        help: "With `--verify`, generate new screenshots when match fails",
+    })
```
edemaine: I wonder if this option would be better named `keepnew`/`keepNew` or `savenew`/`saveNew` or something similar.  I would have guessed `regenerate` would overwrite the existing image, like the default behavior, but it instead saves it in a `new` directory.

dockers/Screenshotter/screenshotter.js
```diff
@@ -494,10 +505,14 @@ function takeScreenshot(key) {
                               // corners
                     diffFile, // output file name
                 ]);
-            })
-            .then(function() {
+            });
+        }
+        if (!opts.regenerate) {
+            promise = promise.then(function() {
                 return promisify(fs.unlink, bufFile);
```
edemaine: I thought this would end up deleting the `new` file when `--regenerate` isn't passed, which would have been a bit unexpected, but it just deletes it within the docker.  Nice!

dockers/Screenshotter/screenshotter.js
ylemkimon: @edemaine I think these tools should not overwrite files without a prompt or user's explicit informed approval. For its name, I'm open to suggestions 😄 Maybe we can print out where diffs and/or new screenshots have been generated.

edemaine: Agreed, I'm happy it doesn't overwrite!

What do you think about a `--new` flag, given that the directory name is called `new`, just like `--diff` and `diff`?  But maybe it's also ambiguous.  My favorite other name is probably `--savenew`...

I like the idea of printing out the directories, if possible.  It would also be nice to mention the directories in the option docstrings.

edemaine:
@ylemkimon I think we also need to add `/test/screenshotter/new` to `.gitignore`, matching `/test/screenshotter/diff`.

dockers/Screenshotter/screenshotter.js
ylemkimon: @edemaine Renamed to `new`.

ylemkimon:
@edemaine Done 😄

PR #1227, opened 2018-03-18
Support Unicode Mathematical Alphanumeric Symbols

ronkok:
This PR adds support for Unicode range U+1D400 to U+1D7FF, [Mathematical Alphanumeric Symbols](https://www.unicode.org/charts/PDF/U1D400.pdf).

### DIFF
--- a/src/MacroExpander.js
+++ b/src/MacroExpander.js
@@ -10,6 +10,7 @@ import builtinMacros from "./macros";
 import type {Mode} from "./types";
 import ParseError from "./ParseError";
 import objectAssign from "object-assign";
+import {expandWideChar} from "./wide-character";
 
 import type {MacroContextInterface, MacroMap, MacroExpansion} from "./macros";
 
@@ -143,12 +144,12 @@ export default class MacroExpander implements MacroContextInterface {
     expandOnce(): Token | Token[] {
         const topToken = this.popToken();
         const name = topToken.text;
-        const isMacro = (name.charAt(0) === "\\");
+        const isMacro = (name.charAt(0) === "\\" || name.charAt(0) === "\uD835");
         if (isMacro && controlWordRegex.test(name)) {
             // Consume all spaces after \macro (but not \\, \', etc.)
             this.consumeSpaces();
         }
-        if (!this.macros.hasOwnProperty(name)) {
+        if (!(this.macros.hasOwnProperty(name) || name.charAt(0) === "\uD835")) {
             // Fully expanded
             this.pushToken(topToken);
             return topToken;
@@ -226,9 +227,16 @@ export default class MacroExpander implements MacroContextInterface {
      * Caches macro expansions for those that were defined simple TeX strings.
      */
     _getExpansion(name: string): MacroExpansion {
-        const definition = this.macros[name];
-        const expansion =
-            typeof definition === "function" ? definition(this) : definition;
+        let expansion;
+        if (name.charAt(0) === "\uD835") {
+            // Surrogate pair. We support Unicode range 1D400–1D7FF,
+            // Mathematical Alphanumeric Symbols.
+            expansion = expandWideChar(name, this.mode);
+        } else {
+            const definition = this.macros[name];
+            expansion =
+                typeof definition === "function" ? definition(this) : definition;
+        }
         if (typeof expansion === "string") {
             let numArgs = 0;
             if (expansion.indexOf("#") !== -1) {
@@ -257,4 +265,3 @@ export default class MacroExpander implements MacroContextInterface {
         return expansion;
     }
 }
-
--- a/src/MacroExpander.js
+++ b/src/MacroExpander.js
@@ -228,12 +228,13 @@ export default class MacroExpander implements MacroContextInterface {
      */
     _getExpansion(name: string): MacroExpansion {
         let expansion;
+        let definition;
         if (name.charAt(0) === "\uD835") {
             // Surrogate pair. We support Unicode range 1D400–1D7FF,
             // Mathematical Alphanumeric Symbols.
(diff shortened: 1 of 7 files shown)


ronkok:
This PR works by writing macros on the fly that invoke the appropriate KaTeX font function and base character. This macro method worked well for most characters. But not all. This PR falls back to system fonts in the cases where:

1. A KaTeX font was unavailable, e.g., script `a-z`
2. No KaTeX function exists to reach the font, e.g., bold Fraktur
3. I was too lazy to build out the table, i.e., Greek letters

The table below shows the details.

| **Item**                     | **math mode** | **text mode** | **Issue**                              |
|:-----------------------------|:--------------|:--------------|:---------------------------------------|
| A-Z bold upright             | ✓             | ✓             |                                        |
| a-z bold upright             | ✓             | ✓             |                                        |
| A-Z italic                   | ✓             | ✓             |                                        |
| a-z italic                   | ✓             | ✓             |                                        |
| A-Z bold italic              | ✓             | ✓             |                                        |
| a-z bold italic              | ✓             | ✓             |                                        |
| A-Z script                   | ✓             | system        | No function                            |
| a-z script                   | system        | system        | No font                                |
| A-Z bold script              | system        | system        | textcal{\\textbf{…}} doesn’t work    |
| a-z bold script              | system        | system        | No font                                |
| A-Z Fraktur                  | ✓             | system        | No function                            |
| a-z Fraktur                  | ✓             | system        | No function                            |
| A-Z double-struck            | ✓             | system        | No function                            |
| a-z double-struck            | system        | system        | No font                                |
| A-Z bold Fraktur             | system        | system        | \\textfrak{\\textbf{…}} doesn’t work |
| a-z bold Fraktur             | system        | system        | ditto                                  |
| A-Z sans-serif               | ✓             | ✓             |                                        |
| a-z sans-serif               | ✓             | ✓             |                                        |
| A-Z bold sans-serif          | ✓             | ✓             |                                        |
| a-z bold sans-serif          | ✓             | ✓             |                                        |
| A-Z italic sans-serif        | ✓             | ✓             |                                        |
| a-z italic sans-serif        | ✓             | ✓             |                                        |
| A-Z bold italic sans         | system        | system        | No font                                |
| a-z bold italic sans         | system        | system        | No font                                |
| A-Z monospace                | ✓             | ✓             |                                        |
| a-z monospace                | ✓             | ✓             |                                        |
| dotless i italic             | ✓             | ✓             |                                        |
| dotless j italic             | ✓             | ✓             |                                        |
| bold Greek                   | system        | system        | Sloth                                  |
| italic Greek                 | system        | system        | Sloth                                  |
| bold italic Greek            | system        | system        | Sloth                                  |
| bold sans-serif Greek        | system        | system        | Sloth                                  |
| bold italic sans-serif Greek | system        | system        | Sloth                                  |
| 0–9 bold                     | ✓             | ✓             |                                        |
| 0–9 double-struck            | ✓             | ✓             |                                        |
| 0–9 sans-serif               | ✓             | ✓             |                                        |
| 0–9 bold sans-serif          | ✓             | ✓             |                                        |
| 0–9 monospace                | ✓             | ✓             |                                        |

ronkok:
This PR is heavily dependent on work done earlier by @kevinbarabash and @rrandallcainc to enable nested font functions. Thanks! The nested functions work splendidly for bold, italic, and sans-serif.

ronkok:
Here's visual confirmation of the work.

math-alpha

ronkok:
I was unable to write a macro that would result in bold Fractur. The KaTeX font exists, but I can find no function that invokes that font. Can anyone suggest a bold Fractur function that I can write into this PR?

Similarly, I was unable to write a macro that would work in text mode for script font, Fraktur, or blackboard bold. Is there some trick I am missing? Can such a macro be written?

ronkok:
In this work, I wrote myself into an infinite loop. The current code has a work-around, but it is quite contrived. I'd like to do better.

@edemaine Is there a LaTeX macro function that will act something like: `\passArgThruWithoutExpandingIt{#1}` ?

If such a thing existed, I could remove file `sysfont.js` from this PR.

edemaine:
@ronkok I haven't grokked all the code here yet -- there's a lot going on.  Could you elaborate on what you'd like this macro to do?  Maybe explain the purpose of `\sysfont`?

ronkok:
General operation:

The entry point is in `MacroExpanders.js`. It traps a wide character at line 232 and calls function `expandWideCharacter(string, mode)`. Say the character is a bold 𝐀, U+1D40. In that case `expandWideCharacter` will return `\mathbf{A}`. That works well.

The difficult cases are when KaTeX does not have the appropriate font available, say for bold script A (𝔄), U+1D4D0. For the fall back, I want KaTeX to render the character in system font. It would be convenient if   `expandWideCharacter` could just return the original character, 𝔄. But if I do that, MacroExpander sends it right back to `expandWideCharacter` and I have an infinite loop.

My current, work around is instead to return `\sysfont{lowSurrogate}` which avoids an infinite loop. But this seems like a odd way to work. I'd like some more elegant way to avoid that infinite loop.

I suppose I could avoid macros altogether if I was willing to write 990 more symbols into `symbols.js`. I'm considering that.

ronkok:
I should add that `\sysfont` knows how to reconstruct the original string character from the low surrogate. But it wouldn't exist at all if I had a more elegant way to avoid an infinite loop.

ronkok:
I'm going to look a completely different way to do this. Instead of an entry point in `MacroExpander`, maybe I can write a entry point in `buildCommon.lookupSymbol`. If it works, it would be cleaner.

ronkok:
I'm closing this PR while I consider another way to accomplish this.

Issue #260, opened 2015-06-20
Accept unicode from mathematical alphabets on input


gagern:
While #243 is discussing ways of accepting unicode characters as input for those situations where the output is using the same unicode symbol, we don't use special symbols on output for the fonts #132 is woking on, but instead use regular latin letters with a special font. Nevertheless it would be nice if we could still accept these special unicode symbols on input.
#51 already gave some translation tables, but in my eyes it essentially boils down to these sets:
- 𝔸𝔹ℂ𝔻𝔼𝔽𝔾ℍ𝕀𝕁𝕂𝕃𝕄ℕ𝕆ℙℚℝ𝕊𝕋𝕌𝕍𝕎𝕏𝕐ℤ𝕜
- 𝔄𝔅ℭ𝔇𝔈𝔉𝔊ℌℑ𝔍𝔎𝔏𝔐𝔑𝔒𝔓𝔔ℜ𝔖𝔗𝔘𝔙𝔚𝔛𝔜ℨ𝔞𝔟𝔠𝔡𝔢𝔣𝔤𝔥𝔦𝔧𝔨𝔩𝔪𝔫𝔬𝔭𝔮𝔯𝔰𝔱𝔲𝔳𝔴𝔵𝔶𝔷
- 𝒜ℬ𝒞𝒟ℰℱ𝒢ℋℐ𝒥𝒦ℒℳ𝒩𝒪𝒫𝒬ℛ𝒮𝒯𝒰𝒱𝒲𝒳𝒴𝒵
- 𝓐𝓑𝓒𝓓𝓔𝓕𝓖𝓗𝓘𝓙𝓚𝓛𝓜𝓝𝓞𝓟𝓠𝓡𝓢𝓣𝓤𝓥𝓦𝓧𝓨𝓩

The first two of these sets should be fairly non-controversial. For the latter two, I'm less sure. Unicode doesn't distinguish between script and caligraphic, so we could map the third to either `\mathscr` or `\mathcal`. The last set is obviously bold, and we only have a bold version of KaTeX_Caligraphic, so perhaps we should map everything to `\mathcal` and add some boldification (see #259) to the latter.

When working with string literals for these beasts, keep in mind that many of these letters are outside of the BMP, so they are encoded using two UTF-16 code units in JavaScript. A simple `charAt` won't always suffice for these cases. I believe that a conforming JavaScript RegExp engine should be able to deal with non-BMP codepoints.

gagern:
[...]

PR #1232, opened 2018-03-20
Support some  Unicode Mathematical Alphanumeric Symbols

ronkok:
This PR adds support for some of the characters in Unicode range U+1D400 to U+1D7FF, [Mathematical Alphanumeric Symbols](https://www.unicode.org/charts/PDF/U1D400.pdf).  Specifically, it adds support for:
* A-Z a-z bold
* A-Z a-z bold italic
* A-Z a-z Fractur
* A-Z a-z sans-serif
* A-Z double struck
* A-Z script

Addresses issue #1215 and parts of #260.


ronkok:
@yorrd, This PR supports these characters in math mode. Do you need them in text mode, as well?

ronkok:
In issue #260, @gagern requested support for characters in the range U+1D4D0 to U+1D4E9. Unicode lists that range as “MATHEMATICAL BOLD SCRIPT”, but the KaTeX fonts do not include glyphs for bold script. 

Most math fonts (STIX, Asana) show glyphs in that range that resemble Computer Modern Script font. The exception is Microsoft’s Cambria Math font, which maps those code points to calligraphic glyphs. (It does the same thing for the Unicode A-Z script section.) Murray Sargent [has written](https://blogs.msdn.microsoft.com/murrays/2016/02/05/unicode-math-calligraphic-alphabets/) that he would make that choice differently today, but it’s too late to change now.

I’m reluctant to write a fall back that uses a system font to render bold script characters. The system may very well use Cambria Math and display calligraphic characters instead. Then documents would break at some future date when KaTeX has a bold script font.

We need a KaTeX font for bold script.

ronkok:
Visual confirmation:
wide-characters

src/symbols.js
```diff
@@ -716,36 +716,76 @@ defineSymbol(text, main, spacing, "\u00a0", " ");
 defineSymbol(text, main, spacing, "\u00a0", "~");
 
 // There are lots of symbols which are the same, so we add them in afterwards.
+let i = 0;
+let ch = "";
```
ylemkimon: Why reuse variables?

ylemkimon:
> Visual confirmation:
Are missing characters intentional?

src/buildCommon.js
```diff
@@ -183,6 +184,11 @@ const mathit = function(
             fontName: "Main-Italic",
             fontClass: "mainit",
         };
+    } else if (value.charCodeAt(0) === 0xD835) {
+        return {
+            fontName: "Math-Italic",
```
ylemkimon: Are metrics the same as `Math-Italic`?

src/symbols.js
ronkok: Why not?

src/buildCommon.js
ronkok: > Are missing characters intentional?

The missing sans-serif "a" is not intentional. I don't see any other missing characters. This is a PR for characters in the Unicode range from U+1D4D0 to U+1D4E9. A letter like "ℂ" does not not belong in this PR because its code point is U+2102, in the Unicode basic multilingual plane.

> Are metrics identical with Math-Italic?

No, but they are much closer to correct than the just using the metrics for the letter "M", as we do elsewhere for some Unicode characters. The biggest discrepancy is in the Fraktur lower case letters. Some of them, such as 𝔷, have descenders that are not present in Math-Italic.

src/symbols.js
ronkok: Also, the style guide in the KaTeX Contributing page instructs us to: "declare variables in the outermost scope that they are used".

edemaine: This kind of re-use of variables generally makes a compiler's job harder, because it becomes harder to tell the scope of the variable that actually matters and thus to figure out the dependencies.  I'm not familiar enough with JS JIT compilers and whether this would actually trip them up, or whether multiple variable declarations would actually cost more (without optimization, it will do more stack work).  But I share @ylemkimon's feeling that the variable `i`s and `ch`s are really different and should be declared separately...

ronkok: Okay. Revision pending.

src/buildCommon.js
edemaine: I'm a little confused why you'd need to be hacking metrics like this.  Is this PR using fallback fonts?  I think that's something that KaTeX generally tries to avoid, instead using characters from its own fonts, and adding to the fonts if needed.  Could you clarify which characters in this PR are supported by KaTeX fonts vs. being hacked in?

ronkok: Error on my part. I wasn't aware that we had font metrics for fonts other than Main and AMS. Now, having actually looked into file `fontMetricsData`, I see that I don't need to use a proxy for the metrics.

I'll revise the PR to use the correct metrics. No proxy is necessary. This PR calls only on KaTeX fonts and does not have a system font fall back.

src/symbols.js
```diff
@@ -716,36 +716,76 @@ defineSymbol(text, main, spacing, "\u00a0", " ");
...
+    if (i < 26) {
+        // KaTeX fonts have only capital letters for blackboard bold and script.
+        wideChar = String.fromCharCode(0xD835, 0xDD38 + i); // A-Z double struck
+        defineSymbol(math, main, mathord, ch, wideChar);
+
+        wideChar = String.fromCharCode(0xD835, 0xDC9C + i); // A-Z script
+        defineSymbol(math, main, mathord, ch, wideChar);
+    }
+
+    // TO DO: Add bold script when it is supported by a KaTeX font.
```
ylemkimon: Could you use `TODO`(without the space), as some IDEs track them and one may do `git grep TODO`?

ronkok: Will do.

ronkok:
@edemaine @ylemkimon So far, this PR supports these characters in math mode only. Do you think we should support them in text mode as well? It would make the `symbols` object substantially larger, perhaps slightly slower.

src/buildCommon.js
```diff
@@ -183,6 +184,12 @@ const mathit = function(
             fontName: "Main-Italic",
             fontClass: "mainit",
         };
+    } else if (value.charCodeAt(0) === 0xD835) {
+        const [wideFontName, wideFontClass] = wideCharacterFont(value, mode);
+        return {
+            fontName: wideFontName,
+            fontClass: wideFontClass,
+        };
```
ylemkimon: I think these logics should be in `mathDefault` or `makeOrd`. Currently `\mathit{𝔄}` produces `𝔄` with metrics for `A`.

ronkok: Well, that's a problem. Thank you for pointing this out. I will fix it, but maybe not right away. I have a busy few days ahead.

ronkok: @ylemkimon I took a look and things are in this PR fine as they are. To test, I did a `console.log(metrics);` on a few letters. A clear example is the letter "z".  

*  In default Main-Italic, the depth of letter"z" should be zero and that is exactly what the console reports. 
* In Fraktur, the depth of letter "𝔷", per `fontMetrics.js`, should be 0.18906. Again, that's exactly what the console reports when I input "𝔷".
* When I input `frak{z}`, the console again reports a depth of 0.18906

Remember that the data in fontMetrics.js` does not go up to an index value of U+1D400. The letters in the KaTeX fonts are placed at their ASCII char codes, not their Unicode code points. The data is indexed the same way. So it's correct to ask for metrics of the letter "z" in my example, or "A" in your example.

I don't know how you got metrics for `\mathit{𝔄}`. When I tried to input `\mathit{𝔄}`, KaTeX gave me a parse error. If it would parse, one could argue that the proper metrics would indeed be for "A". According to @rrandallcainc in PR #1111, math commands, such as `\mathit` "change all aspects of the font (weight, shape, and family)".

I think this PR, as written, is getting the proper metrics.

ylemkimon: @ronkok I checked out your branch and using `npm start` it renders `\mathit{𝔄}` without `ParseError`:

image

Regarding the metrics, I was looking at the old source code and got wrong, sorry. It uses metrics of 𝔄 and produces 𝔄.

The question is, should math font commands alter them? Currently, `mathit` renders as-is, e.g., `\mathit{𝔄}` gives 𝔄, whereas other font commands change its font, e.g., `\mathrm{𝔄}` gives A.

ylemkimon: Also, it seems that the `\boldsymbol` doesn't bolden given character, instead it renders in bold italic.

ylemkimon: Regarding their behavior, I think it should follow #1217, unless LaTeX commands exist to typeset them in the text mode.

ronkok: > Also, it seems that the `\boldsymbol` doesn't bolden given character, instead it renders in bold italic.

A bold italic is exactly what `\boldsymbol` should do. It's part of the AMS package. Here's an rendering of `\boldsymbol{ABD}` from quicklatex.com:  

Should `\mathit{𝔄}` override the Fraktur font and render an *A*? That's an interesting question. I'll do some research into how XeTeX behaves. If the answer turns out to be yes, that will take me some time. I'm booked up for the next few days.

ylemkimon: So `\boldsymbol{𝖠}` should give 𝑨 rather than 𝝖, right?

ronkok: So `\boldsymbol{𝖠}` should give 𝑨 rather than 𝝖, right?

Yep.

edemaine:
Maybe also worth testing whether XeLaTeX supports these in text mode.

ronkok:
It turns out that in MathJax, `\mathit{𝐀𝐁𝐂𝐃}` does **not** override the font inherent in the wide characters 𝐀𝐁𝐂𝐃.  `\mathit{𝐀𝐁𝐂𝐃}` renders just like 𝐀𝐁𝐂𝐃.  I think that's relevant since we're doing this at the prompting of someone who wants to migrate from MathJax to KaTeX. 

Also, MathJax does support wide characters, such as 𝐀𝐁𝐂𝐃, in both math mode and text mode.

I haven't yet gotten wide characters to work in XeLaTeX. I don't know if I have a preamble issue or if the editor at ShareLaTeX.com can't deal with wide characters. I'll try a bit more.

ronkok:
Results are in. XeLaTeX (with the `unicode-math` package) and MathJax act alike in that they both:

1.  do support wide characters, such as 𝐀𝐁𝐂𝐃, in both math mode and text mode.
2.  fail to override font of a wide character with `\mathit`. `\mathit{𝔸 𝔹 𝔻}` renders just like 𝔸 𝔹 𝔻.

I'm going to modify this PR to include text mode. Otherwise, I'm going to leave it alone.

ronkok:
The most recent commit supports text mode. It also deletes the sans-serif characters. No one has asked for them yet, and it will be easier to add characters later than to remove them.  The characters now supported by this PR, in both math mode and text mode are:

* A-Z a-z bold
* A-Z a-z bold italic
* A-Z a-z Fractur
* A-Z double struck
* A-Z script

ronkok:
Below is visual confirmation for math mode:
mathmode

Next is visual confirmation for text mode:
textmode

ylemkimon:
> It also deletes the sans-serif characters. No one has asked for them yet, and it will be easier to add characters later than to remove them.

I think there is no reason not to add sans-serif.

> Otherwise, I'm going to leave it alone.

Should font commands other than `\mathit` change its font, e.g., `\mathrm{𝔄}` gives A?

ronkok:
`\mathrm{𝔄}` renders just like 𝔄 in both XeLaTeX and MathJax. 

On the sans-serif question, I have no strong feelings either way. Just to be clear, I'll point out that the following characters would be easy to code in at this time:

* A-Z a-z 0-9 sans serif
* A-Z a-z 0-9 sans serif bold
* A-Z a-z sans serif italic
* A-Z a-z 0-9 monospace
* 0-9 bold

After @edemaine and @kevinbarabash  have weighed in, I would be happy to follow group consensus and include any or all of those characters in this PR.  Other Latin letters or numerals in the Unicode range U+1D400 to U+1D7FF are missing fonts or font metrics, so the list above is an upper limit.

I could also add some Greek letters, but the code writing there is a little more tedious. One has to watch out for special treatment of phi, theta, kappa, `\nabla`, and `\partial`.

kevinbarabash:
I'm confused as to why the unicode equivalent of `\mathscr{B}` is missing.  `\mathscr{B}` itself renders just fine.

kevinbarabash:
> Script B, ℬ, is at Unicode code point U+212C

Interesting.  

> I can write another PR to cover the other missing characters
  
Sounds good.  I think there's enough going on in this PR as is though.

ylemkimon:
Currently, `\mathrm{𝔄}` seems to be rendered to upright roman `A` in the PR, since its logics are in `mathit`.

ronkok:
@ylemkimon Thank you for pointing that out. You are correct. That behavior is contrary to XeLaTex, so I'll look for a way to fix it.

ronkok:
The most recent commit corrects the bug found by @ylemkimon. `\mathrm` now renders as it does in XeLaTeX.  Visual confirmation below:

mathrm

I also added a screenshotter test.

kevinbarabash:
The problem seems to be that the surrogate pairs screenshot test is too wide.  @ronkok could you change that test to render things in an array?  Also, are you able to access the screenshots from circleci?  If not, I can copy/paste them into this PR after the tests run.

surrogate-pairs

kevinbarabash:
Another option is to make the font smaller.

kevinbarabash:
surrogatepairs-chrome
surrogatepairs-firefox

kevinbarabash:
@ronkok I've posted the screenshots for the SurrogateParis test above.  If you update your PR with those this should be good to go.

ronkok:
> If you update your PR with those [screenshots] this should be good to go.

I'm sorry, but I don't know how to do that.

kevinbarabash:
@ronkok add the image files I posted to the `test/screenshotter/images` folder and commit and push that to this PR.

ronkok:
Done.

src/wide-character.js
```diff
@@ -0,0 +1,100 @@
...
+ * Data below is from https://www.unicode.org/charts/PDF/U1D400.pdf
+ * That document sorts characters into groups by font type, say bold or italic.
+ *
+ * In the arrays below, each subarray consists three elements:
+ *      * The CSS class of that group when in math mode.
+ *      * The CSS class of that group when in text mode.
+ *      * The font name, so that KaTeX can get font metrics.
+ */
+
+const wideLatinLetterData: Array<[string, string, string]> = [
```
kevinbarabash: It would be good to introduce some types like:
```
type MathCssClass = "mathbf" | "mathit" | ...;
type TextCssClass = "textbf" | "textit" | ...;
type FontName = "Main-Bold" | "MainItalic" | ...;
```
This will make it easier to pick up errors as people continue to work on this code.

src/buildCommon.js
```diff
@@ -234,7 +235,11 @@ const makeOrd = function(
     // Math mode or Old font (i.e. \rm)
     const isFont = mode === "math" || (mode === "text" && options.font);
     const fontOrFamily = isFont ? options.font : options.fontFamily;
-    if (fontOrFamily) {
+    if (value.charCodeAt(0) === 0xD835) {
+        // surrogate pairs get special treatment
+        const [wideFontName, wideFontClass] = wideCharacterFont(value, mode);
```
kevinbarabash: In the future it would be nice if we had a function that handled the non-wide character situation in a similar way to how `wideCharacterFont` does it.

src/wide-character.js
```diff
@@ -0,0 +1,100 @@
...
+    ["mathbf", "textbf", "Main-Bold"],                // A-Z bold upright
+    ["mathbf", "textbf", "Main-Bold"],                // a-z bold upright
+    ["mathit", "textit", "Main-Italic"],              // A-Z italic
+    ["mathit", "textit", "Main-Italic"],              // a-z italic
```
kevinbarabash: It looks like `mathit` actually uses `KaTex_Math-Italic` instead of `KaTeX_Main-Italic`

src/symbols.js
```diff
@@ -732,13 +731,53 @@ for (let i = 0; i < textSymbols.length; i++) {
...
+    const ch = letters.charAt(i);
+
+    // The hex numbers in the next line are a surrogate pair.
+    // 0xD835 is the high surrogate for all letters in the range we support.
+    // 0xDC00 is the low surrogate for bold A.
+    wideChar = String.fromCharCode(0xD835, 0xDC00 + i);  // A-Z a-z bold
+    defineSymbol(math, main, mathord, ch, wideChar);
+    defineSymbol(text, main, textord, ch, wideChar);
+
+    // We omit italic. Math mode A-Za-z are italic by default.
```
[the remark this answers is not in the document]
ronkok: Correct. Any characters from that range will be unsupported. I couldn't think of a reason why anyone would want to use characters from that range, since the default is already italic.

But perhaps it's just that my imagination isn't broad enough. Let me know if you want to include that range in this PR.

src/symbols.js
```diff
@@ -732,13 +731,53 @@ for (let i = 0; i < textSymbols.length; i++) {
...
+    defineSymbol(math, main, mathord, ch, wideChar);
+    defineSymbol(text, main, textord, ch, wideChar);
+
+    wideChar = String.fromCharCode(0xD835, 0xDD04 + i);  // A-Z a-z Fractur
+    defineSymbol(math, main, mathord, ch, wideChar);
+    defineSymbol(text, main, textord, ch, wideChar);
+
+    if (i < 26) {
+        // KaTeX fonts have only capital letters for blackboard bold and script.
+        wideChar = String.fromCharCode(0xD835, 0xDD38 + i); // A-Z double struck
```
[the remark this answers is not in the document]
ronkok: Good point. I'll adjust the comment.

src/wide-character.js
ronkok: I'll work on it.

src/symbols.js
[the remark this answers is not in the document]
kevinbarabash: People might be copy/pasting from other software where this range is being used for italics.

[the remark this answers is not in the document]
ronkok: I went ahead and added those characters in the most recent commit.

src/wide-character.js
ronkok: I tried adding the types that you suggested. The result was an enormous number of flow errors. My flow skills are not strong.

So I have reverted to specifying a string in the most recent commit.

edemaine:
@kevinbarabash Did you want to try to fix the Flow issue, or accept this as is? It looks like it's probably good to go.

kevinbarabash:
> @kevinbarabash Did you want to try to fix the Flow issue, or accept this as is? It looks like it's probably good to go.

This is fine as is.  We can add `flow` in later.  That being said, as soon as we get to 100% flow we should probably be more strict.

PR #1235, opened 2018-03-21
Add test job to CircleCI

ylemkimon:
* Added `test` job to circleCI: takes about 2 minutes.
* Built KaTeX can be accessed through build artifacts


ylemkimon:
@kevinbarabash It has problems if CircleCI is enabled on the forked repo. It runs the build on its owner, and reports only one job. I'm not sure it's problem of CircleCI 2.0 or workflow feature.

kevinbarabash:
I would like replace travis with circle, but this issue seems like a deal breaker.  Can we try using CircleCI 2.0?

ylemkimon:
@kevinbarabash What do you think about failing the build when it's on the forked repository, and documenting about it on CONTRIBUTING.md?

kevinbarabash:
> @kevinbarabash What do you think about failing the build when it's on the forked repository, and documenting about it on CONTRIBUTING.md?

That's definitely an option.  How would we do that?

.circleci/config.yml
```diff
@@ -1,23 +1,31 @@
-defaults: &defaults
+checkout_submodule: &checkout_submodule
+  run:
+    name: Checkout submodule
+    command: |
+        git submodule sync
+        git submodule update --init --recursive
+
+npm_cache_key: &npm_cache_key
+  npm-deps-v1-{{ checksum "package.json" }}
```
kevinbarabash: Cool!  Why the `-v1-`?

.circleci/config.yml
```diff
@@ -1,23 +1,31 @@
...
+
+restore_npm_cache: &restore_npm_cache
+  restore_cache:
+    keys:
+      - *npm_cache_key
+      - npm-deps-v1-
```
kevinbarabash: Isn't this already include in `*npm_cache_key`?

.circleci/config.yml
```diff
@@ -38,20 +46,57 @@ defaults: &defaults
 
 version: 2
```
kevinbarabash: This means we're using `circleci v2`?

ylemkimon: @kevinbarabash Yes, we're using CircleCI 2.0.

.circleci/config.yml
ylemkimon: @kevinbarabash `npm_cache_key` includes the checksum of the file `package.json`, so if `package.json` changes, it will not match old caches.

.circleci/config.yml
ylemkimon: @kevinbarabash It's just for internal versioning of the cache, allowing to invalidate the cache.

ylemkimon:
Now I think, if we don't need #1244, I think going with option #1 and removing screenshot tests from Travis may be better.

kevinbarabash:
@ylemkimon what made you change your mind?  FWIW, I think using travis for the unit tests only is okay since they run pretty fast.  This PR implements option 1, correct?

kevinbarabash:
@ylemkimon I spent some time migrating another project completely off of travis-ci and can't be happier running everything on circleci.  I find travis-ci is so buggy.  Some times the builds take really long or hang completely.  Then there's #1484 where there's no to travis-ci build from the GitHub UI.  This is not the first time that's happened either.  I think we should go with option 2.

ylemkimon:
@kevinbarabash Travis CI not running on #1484 was probably because I've put `[skip ci]` to the commit message. But I agree Travis can be unreliable.

Regarding CircleCI, https://github.com/Khan/KaTeX/pull/1484#issuecomment-406838051:

> There were couple of instances of CircleCI Chrome build failing, that I'd to rerun manually. As the problems are limited to Chrome builds, I think it's not a CircleCI issue, but a Selenium issue and upgrading its version might fix the problems.

ylemkimon:
@kevinbarabash What changes/revisions does this require?

ylemkimon:
@kevinbarabash Should I remove Travis CI?

kevinbarabash:
I'm going to update the required checks as well.

PR #1236, opened 2018-03-21
List react-katex in Libraries

### LINKED ISSUE #1155: Is there a React / ES6 implementation?
Hi there,

I'm trying to set up KaTeX in react-16 using ES6. However I can't get it working. Also the react-katex repo is poorly maintained, so I'm currently kind of stuck. How to integrate katex rendering into react-16?

dudko:
Hello Khaners,

as one of the contributors of `react-katex`, React components that use KaTeX to typeset math expressions, I would like to you to list `react-katex` in your Libraries listing.

Thank you!

(fix #1155)

### DIFF
--- a/README.md
+++ b/README.md
@@ -138,6 +138,10 @@ will appear larger than 1cm in browser units.
 
 - [katex-ruby](https://github.com/glebm/katex-ruby) Provides server-side rendering and integration with popular Ruby web frameworks (Rails, Hanami, and anything that uses Sprockets).
 
+### React
+- [react-katex](https://github.com/talyssonoc/react-katex) React components that use KaTeX to typeset math expressions
+
+
 ## Contributing
 
 See [CONTRIBUTING.md](CONTRIBUTING.md)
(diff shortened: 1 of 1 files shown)


dudko:
[clabot:check]

edemaine:
@dudko Thanks for the PR!  Have you read our issue #1155?  There were some issues with react-katex reported there, though perhaps they've all been addressed by now.  Also it sounds like react-latex is a competitor that might also be worth mentioning.  What do you think is best?

dudko:
@edemaine , thank you for your quick response. I've encountered mentioned issues with `react-katex` in past. Code base got outdated and incompatible with new react releases. I did contribute to the project recently and hopefully resolved all those issues.

Regardless of `react-latex`, it does pretty much the same job as `react-katex`. Both packages are definitely worth mentioning. It's only on users to decide which one is better. :)

edemaine:
Awesome! My only suggestion would be to put React above Ruby, for capitalization and prominence. But also fine as is.

Issue #785, opened 2017-08-12
add a CHANGELOG.md


kevinbarabash:
- backfill with PR titles from previous releases
- require new PRs add an entry to CHANGELOG.md (add something in CONTRIBUTING.md)

mbourne:
(Not sure where to put this comment...)

On this page, https://github.com/Khan/KaTeX/releases/tag/v0.8.1 the link in your comment correctly points to v0.8.0, but the anchor text says v0.8.1. It would reduce confusion if fixed.

kevinbarabash:
@mbourne thanks.  I've fixed it now.

kevinbarabash:
I'm going to close this since the release notes in GitHub already provide this info.

Issue #1132, opened 2018-02-02
Interpolating radical symbols


kevinbarabash:
Now that we have radical symbols defined inside our JavaScript, it's possible we could interpolate between them.  One of the things that's bugged me about LaTeX's `\sqrt` is that the depth can vary depending on the contents even if the contents' depths are the same.

<img width="443" alt="screen shot 2018-02-01 at 8 51 46 pm" src="https://user-images.githubusercontent.com/1044413/35712817-08cbb58e-0792-11e8-9562-26e6618e3c2b.png">

Ideally the depth of the `\sqrt` symbol in the first two expressions would match.  Same goes for the second two expressions.

Of course since this is not something LaTeX does it wouldn't be something we'd turn on by default.

PR #1237, opened 2018-03-22
Add CHANGELOG

Kully:
Re: https://github.com/Khan/KaTeX/issues/785


ylemkimon:
I'm personally against changelogs auto-generated from commit logs and/or merged pull requests.

From [keepachangelog.com](keepachangelog.com):
> Can changelogs be bad?
Yes. Here are a few ways they can be less than useful.
Commit log diffs
Using commit log diffs as changelogs is a bad idea: they're full of noise. Things like merge commits, commits with obscure titles, documentation changes, etc.
The purpose of a commit is to document a step in the evolution of the source code. Some projects clean up commits, some don't.
The purpose of a changelog entry is to document the noteworthy difference, often across multiple commits, to communicate them clearly to end users.

Kully:
> I'm personally against changelogs auto-generated from commit logs and/or merged pull requests.

I did not autogenerate this, but I agree with the implication that the current CHANGELOG is too cluttered/needs to be more readable by a user.

Kully:
closing for now

kevinbarabash:
A CHANGELOG that simply aggregates the list of bug fixes, new features, and misc from each release in https://github.com/khan/KaTeX/releases might address the concerns about a cluttered change log.

edemaine:
Agreed. I'm unclear on best practices here. Most often I see a CHANGELOG but no Github release notes. Given that we have the latter, do we need the former?

kevinbarabash:
> Given that we have the latter, do we need the former?

The nice thing about the former is that you can see them all together whereas the releases are paginated which isn't that bad imo.

Kully:
> The nice thing about the former is that you can see them all together whereas the releases are paginated which isn't that bad imo.

Plotly's Changelog has sections in each release that work well: Added, Fixed, Updated https://github.com/plotly/plotly.py/blob/master/CHANGELOG.md

@kevinbarabash 
An alternative to just one-paging the release info is picking the relevant notes and categorizing based on these 3 sections, or to just organize them all under these sections. WDYT?

ylemkimon:
They seem to follow [keepachangelog.com](https://keepachangelog.com/en/1.0.0/):
> `Added` for new features.
`Changed` for changes in existing functionality.
`Deprecated` for soon-to-be removed features.
`Removed` for now removed features.
`Fixed` for any bug fixes.
`Security` in case of vulnerabilities.

kevinbarabash:
@Kully the CHANGELOG.md file you linked to looks alright.  I like the categories that @ylemkimon posted to.  Currently the release notes combine `Added` and `Changed` into a single `New Features`.  I kind of like separate them out though.  I guess every once in a while we do have `Breaking Changes` which is kind of like `Changed`.  I'm not sure what would fit into `Changed` but not `Breaking Changes`.  🤔

Kully:
> @Kully the CHANGELOG.md file you linked to looks alright. Should I continue with it then? I deleted the branch because you closed the original issue. I can make the CHANGELOG and split `Nee Features` into `Added` and `Changed` as best I can.

And regarding `Breaking Changes`, I imagine its any change to the code in which functions do not work/raise errors, eg. if `\sqrt{}` no longer worked. This seems like it would be a very small category. And `Changes` would be things like interpolating the sqrt symbol: https://github.com/Khan/KaTeX/issues/1132

What do you think?

kevinbarabash:
I see, so changes might be something like "we made this existing feature better in some way"?

Kully:
@kevinbarabash Yeah exactly. At Plotly that's how we interpret it

PR #1238, opened 2018-03-24
Replace CDN links

LukasDrgon:
I replaced the unpkg links with jsDelivr links. [jsDelivr](https://www.jsdelivr.com/package/npm/katex) works in the same way, but was built specifically for production usage. This should solve problems described in https://github.com/unpkg/unpkg/issues/88.

### DIFF
--- a/function-support.html
+++ b/function-support.html
@@ -6,9 +6,9 @@
     <title>Function Support in KaTeX</title>
     <link rel="icon" href="http://khan.github.io/favicon.ico">
 	
-	<link rel="stylesheet" href="https://unpkg.com/katex@0.9.0/dist/katex.min.css">
-	<script src="https://unpkg.com/katex@0.9.0/dist/katex.min.js"></script>
-	<script src="https://unpkg.com/katex@0.9.0/dist/contrib/auto-render.min.js"></script>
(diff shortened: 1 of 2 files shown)


LukasDrgon:
[clabot:check]

ylemkimon:
Is jsDelivr better than cdnjs?

MartinKolarik:
cdnjs uses CloudFlare, jsDelivr uses a combination of four providers, including one in China.

edemaine:
@ronkok I think you were the first to suggest unpkg. OK with this change? Sounds good to me.

ronkok:
Quick CDN propagation is unimportant for `function-support.html`. For it, I'm content with nearly any CDN.

Quick CDN propagation will be *very* important for `index.html` if we wish to make its revision part of the standard release script. 

When jsDelivr is used, how quickly do users see the updated files?

MartinKolarik:
@ronkok it's < 60 seconds, in most cases there's no delay at all.

ronkok:
@MartinKolarik Fantastic. I think this PR is a great idea.

kevinbarabash:
@LukasDrgon cool!  Thanks for the PR.

PR #1239, opened 2018-03-25
Treat svg-wrapping spans differently from DOM-wrapping spans.

marcianx:
In preparation for the conversion of buildHTML to @flow, separate the two
different uses of domTree.span -- one to wrap a single SVG and one to wrap
arbitrary DOM nodes. This separation is to help avoid type issues that pop up
due to buildHTML methods accessing the .children field of spans as if they are
DOM elements and not SVG ones.

### DIFF
--- a/src/domTree.js
+++ b/src/domTree.js
@@ -29,18 +29,22 @@ const createClass = function(classes: string[]): string {
 };
 
 // To ensure that all nodes have compatible signatures for these methods.
-interface VirtualDomNode {
+interface VirtualNodeInterface {
     toNode(): Node;
     toMarkup(): string;
 }
 
-export interface CombinableDomNode extends VirtualDomNode {
-    tryCombine(sibling: CombinableDomNode): boolean;
+interface HtmlDomInterface extends VirtualNodeInterface {
+    classes: string[];
+    height: number;
+    depth: number;
+    maxFontSize: number;
+
+    tryCombine(sibling: HtmlDomNode): boolean;
 }
 
 /**
- * All `HtmlDomNode`s MUST have `height`, `depth`, and `maxFontSize` numeric
- * fields.
+ * All `HtmlDomNode`s must implement HtmlDomInterface.
  *
  * `HtmlDomNode` is not defined as an interface since `documentFragment` also
  * has these fields but should not be considered a `HtmlDomNode`.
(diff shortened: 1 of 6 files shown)


src/domTree.js
```diff
@@ -29,23 +29,31 @@ const createClass = function(classes: string[]): string {
...
 
-export interface CombinableDomNode extends VirtualDomNode {
-    tryCombine(sibling: CombinableDomNode): boolean;
+interface HtmlDomInterface extends VirtualNodeInterface {
+    classes: string[];
+    height: number;
+    depth: number;
+    maxFontSize: number;
+
+    tryCombine(sibling: HtmlDomNode): boolean;
```
ylemkimon: I don't know about flowtype naming convention, but should the name of an interface contain the word `Interface`?

marcianx: I didn't see any guidelines regarding that.

I originally didn't using `*Interface`. But this file defines quite a few new types, a number of which are closely related to these interfaces; so it helped to make the interfaces stand out. Note that none of the interfaces are actually exported from this file, so changing the convention here in the future should be very easy as well.

src/stretchy.js
```diff
@@ -166,10 +167,10 @@ const groupLength = function(arg: ParseNode): number {
     }
 };
 
-const svgSpan = function(group: ParseNode, options: Options): domTree.span {
+const svgSpan = function(group: ParseNode, options: Options): DomSpan | SvgSpan {
```
ylemkimon: The type `SvgSpan` may be confused with the function `svgSpan`.

marcianx: `svgSpan` is actually a constructor for an `SvgSpan` or a list of `SvgSpan`s within a `DomSpan`. So it doesn't feel out of place there. Currently, we have (in increasing order of dependency):

- `SvgSpan` - the fundamental type
- `buildCommon.makeSvgSpan` - `SvgSpan` constructor with `ChildType` in `span<ChildType>` specialized to `svgNode` (aids type inference).
- `stretchy.svgSpan` - to convert a `ParseNode` (that needs SVGs) to a span wrapping one or more SVGs for output.

I'm not sure what to name the last one, but currently, all uses of `stretchy.svgSpan` include the word `stretchy` making the context obvious, so I don't feel as concerned about confusion between the method and the type.

marcianx: Just wanted to check -- are you okay with this? I don't want to merge if you still have any unanswered concerns.

ylemkimon: @marcianx Yes, I think it's good to go 😄

kevinbarabash: Maybe we should change the factory methods to be something like `createSvgSpan` or `makeSvgSpan`.

PR #1243, opened 2018-03-27
Change CDN links

LukasDrgon:
Given the advantages of jsDelivr discussed in #1238 I think it would be a good idea to use it in the README as well. What do you think?

### DIFF
--- a/update-sri.js
+++ b/update-sri.js
@@ -26,7 +26,7 @@ Promise.all(process.argv.slice(3).map(file =>
         // 5 - integrity opening quote: "
         // 6 - old hash: sha384-…
         // 7 - integrity hash algorithm: sha384
-        const re = /((["'])https?:\/\/cdnjs.cloudflare.com\/ajax\/libs\/KaTeX\/)[^\/"']+(\/([^"']+)\2(?:\s+integrity=(["'])(([^-]+)-[^"']+)\5)?)/g;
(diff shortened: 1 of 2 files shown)


ylemkimon:
I'm +1.

ronkok:
Me too. Consistency is good of course. But more importantly, the existing README points to a CDN that takes 2 or 3 days to fully propagate the files.  It's common during those days to get a question on Gitter asking why the release doesn't work.

edemaine:
Sounds good to me too. Thanks for the PR! 

Before merging, we might wait for @kevinbarabash to return, as he usually runs the release scripts. I wouldn't want to mess those up (though this PR seems to address them).

update-sri.js
```diff
@@ -35,7 +35,7 @@ Promise.all(process.argv.slice(3).map(file =>
             return pre + version + post;
         });
         return Promise.all(Object.keys(hashes).map(hash =>
-            read(path.join("dist", hashes[hash].file), null)
+            read(hashes[hash].file, null)
```
kevinbarabash: why did this change?

MartinKolarik: jsDelivr links already include `dist`, see above in readme

kevinbarabash: Cool.  I double checked the regex on regex101.com and it checks out.  The regex could probably be simplified, but your changes are minimal so it's okay.

PR #1247, opened 2018-04-02
Type correction to allow documentFragment to be children of span.

marcianx:
This is necessary because documentFragments are put in spans in some cases --
e.g. src/functions/sqrt.js. Not sure why this wasn't caught by the type system.
It's possible that it might be happening in buildHTML as well -- it is not yet
been typed and thus violations of the existing types haven't yet had a change to
be caught.


marcianx:
Given that this function is only added for the sake of the type system -- any suggests as to what to do about the codecov report?

kevinbarabash:
@marcianx the codecov check is non-blocking.

kevinbarabash:
> This is necessary because documentFragments are put in spans in some cases --
e.g. src/functions/sqrt.js.

The code changes look fine, but could you expand on this before merging.  Is it b/c `children` for a `span` can only be `DomChildNode` which doesn't include `documentFragment`?

marcianx:
Yes. The original assumption when I was putting in the types was that documentFragments hold children but are never children themselves. Then I found that my assumption was false in my ongoing efforts to put types in buildHTML.

src/domTree.js
```diff
@@ -43,13 +43,9 @@ interface HtmlDomInterface extends VirtualNodeInterface {
...
-/**
- * All `HtmlDomNode`s must implement HtmlDomInterface.
- *
- * `HtmlDomNode` is not defined as an interface since `documentFragment` also
- * has these fields but should not be considered a `HtmlDomNode`.
- */
-export type HtmlDomNode = DomSpan | SvgSpan | anchor | symbolNode;
+/** All `HtmlDomNode`s must implement HtmlDomInterface. */
+export type HtmlDomNode =
+    DomSpan | SvgSpan | anchor | symbolNode | documentFragment;
```
ylemkimon: If all `HtmlDomNode`s implement `HtmlDomInterface`, couldn't we use `HtmlDomInterface` in place of `HtmlDomNode`?

ylemkimon:
Possibly related: https://github.com/facebook/flow/issues/2723, https://github.com/facebook/flow/issues/2282

src/domTree.js
marcianx: I'd considered that previously and now that you mention it again, I'm reconsidering my decision.

To elaborate on my original intention: the type sum is technically stricter and also allows one to clearly see which discrete possibilities apply (e.g. `span<SomethingElse>` is not allowed even through it implements the interface). My original concern was that `interface`s are strictly structural -- even if you don't explicitly say `implements HtmlDominterface`, something can accidentally implement it and allow us to pass in a value of that type.

However, given that `HtmlDomInterface` currently includes `tryCombine` and we have explicit constructors for `DomSpan` and `SvgSpan`, there's pretty much no chance of something accidentally satisfying the interface. So you're right, we could probably just replace it with the interface now.

Done. We can also revert it if ever needed as the change is pretty trivial.

PR #1248, opened 2018-04-02
Added ≝ symbol

youqad:


youqad:
[clabot:check]

src/symbols.js
```diff
@@ -98,6 +98,8 @@ defineSymbol(math, main, rel, "\u221d", "\\propto", true);
 defineSymbol(math, main, rel, "\u22a2", "\\vdash", true);
 defineSymbol(math, main, rel, "\u22a3", "\\dashv", true);
 defineSymbol(math, main, rel, "\u220b", "\\owns");
+defineSymbol(math, main, rel, "\u225d", "\\overset{\\mathrm{def}}{=}");
```
edemaine: Did you test this? I'm pretty sure `defineSymbol` only works with a single `\command`, not an entire expression like this.  Instead, you should use something like `\equalsWithDef` here (ideally an existing LaTeX command that reproduces this) and then `defineMacro` that to the expression.

youqad: I'm sorry, I'm a complete beginner when it comes to KaTeX. I didn't test it (I'm not even sure how to do it, to be honest). Where is defined your `defineMacro` function? (because I don't think there's an already existing LaTeX command reproducing this "equal by definition")
Thanks a lot for your help, and sorry for the inconvenience!

edemaine: No problem.  `defineMacro` goes in `macros.js`.  If there's no LaTeX equivalent, then I'd suggest putting this directly as a `defineMacro`, without any symbol definition.  (You can define a single character as a "macro".)

To test, run `npm start` and connect to the URL printed.  Maybe post a screenshot of it working. You might also check out https://github.com/Khan/KaTeX/blob/master/CONTRIBUTING.md

@ronkok Thoughts on this Unicode support? How does it compare to XeLaTeX etc.?

ronkok:
@youqad Great idea to add this symbol. You beat me to it. In fact, my personal To Do list includes characters: ≘ ≙ ≚ ≛ ≝ ≞ ≟. LaTex package `unicode-math` supports all of those characters, so XeLaTeX supports all of them.  So does MathJax.

I have some suggestions that will improve the rendering. And by "improve", I am aiming at a rendering similar to what XeLaTeX would provide. It uses font glyphs from mathematical fonts like STIX or Cambria Math. To get a KaTeX macro to render like those, I think we should:
1. Use `\stackrel` instead of `\overset`. Then we get relation spacing instead of mathord spacing.
2. Reduce the size of the upper element.
3. Reduce the gap between the "=" and the upper element.

We can get the first two items with these macros:
```
// Unicode stacked relations
defineMacro("\u2258", "\\stackrel{\\tiny{\\frown}}{=}");
defineMacro("\u2259", "\\stackrel{\\tiny{\\wedge}}{=}");
defineMacro("\u225A", "\\stackrel{\\tiny{\\vee}}{=}");
defineMacro("\u225B", "\\stackrel{\\scriptsize{\\star}}{=}");
defineMacro("\u225D", "\\stackrel{\\tiny{\\mathrm{def}}}{=}");
defineMacro("\u225E", "\\stackrel{\\tiny{\\mathrm{m}}}{=}");
defineMacro("\u225F", "\\stackrel{\\tiny{?}}{=}");
```

That gives a result like this one:
stackrel

That result shows something pretty close to the font glyphs used by XeLaTeX.  I was planning to use the `\lower` function in PR #1167 to tighten up the gaps in ≘ ≙ ≚ ≛. That PR is pending review. 

But I personally am ok with the spacing shown above for ≝ ≞ ≟. If others agree, those characters could go into this PR and proceed independently of `\lower`.

edemaine:
@ronkok This is great, thanks. Definitely looks like the way we should go. One small thing: can't you use existing `\raisebox` support? I thought #1167 was about providing `\raise`/`\lower` aliases for `\raisebox`, but I haven't looked at it yet so I'm not sure.

ronkok:
@edemaine Yes, one could use `\raisebox`. But some of these characters don't qualify as text, so the macro would have reset to math mode inside the `\raisebox` argument.  For example: `\raisebox{$\star$}`. I was hoping that `\raise` and `\lower` could work without that extra complexity.

edemaine:
@ronkok OK, good point. Essentially we want `\mraisebox` like we have `\mllap`.

How does XeLaTeX implement these? Can we copy their macro definitions directly?

ronkok:
In XeLaTeX, one first loads a math font, say `\setromanfont{STIX}`. So we can't use precisely the XeLaTeX methods until the KaTeX fonts include glyphs for ≘ ≙ ≚ ≛ ≝ ≞ ≟. I think that the macros above, when adjusted for vertical kern distance, are about as good as we do for now.

edemaine:
Replaced by #1274.

Issue #1173, opened 2018-02-19
Thick grey bar on some fractions


mbourne:
In v0.9.0, I see some good fraction lines and some overly thick grey ones (Chrome on Windows)

![image](https://user-images.githubusercontent.com/7541940/36362873-3d30d298-1573-11e8-9c2c-a4c846167fcd.png)

Firefox is a bit better in that it is less obvious (at least on my monitor setup). In the fractions on the first line, it appears to have a darker bottom 1px line and a lighter "shadow" above it, the middle line is good, and the last line is mostly thick grey:

![image](https://user-images.githubusercontent.com/7541940/36362984-36465d1c-1574-11e8-851a-c86a7a5e3972.png)

So unfortunately it doesn't appear that using path in the svg has completely fixed it.

I'm sticking with v0.9.0-alpha for my production site for now.

ronkok:
Well, that’s disappointing. I hoped that we had fixed things. @kevinbarabash, I’ll recap below the various ways that I have attempted to help.

| **Version**  | **Ink Source** | **Ink Height** | **Span Height**       | **Results**                                                                                                          |
|:-------------|:---------------|:---------------|:----------------------|:---------------------------------------------------------------------------------------------------------------------|
| 0.9.0-alpha  | span<br>border    | 0.04em         | 0.04em                | Base case. Many reports of disappearing fraction lines.                                                                  |
| 0.9.0-alpha1 | span<br>border    | 0.04em         | 0.04em<br>min-height:1px | Improvement, but not complete success. Reports come in that there are still a few disappearing lines.                |
| 0.9.0-alpha2 | SVG<br>path       | Big            | 0.04em                | Some lines are rendered thinner than standard. This teaches us that browsers handle `span.height` imprecisely.       |
| 0.9.0-beta   | SVG<br>line       | 0.04em         | 0.2em                 | Spans no longer pinch down on line ink. @akalin reports a complete fix.                                              |
| 0.9.0-beta1  | SVG<br>line       | 0.04em         | 0.2em                 | No change in the code, but reports come in that some lines are too thick.                                            |
| 0.9.0        | SVG<br>path       | 0.04em         | 0.2em                 | @mbourne reports that some lines have fuzzy edges and some are gray. This is due to anti-aliasing on the paths. |

v0.9.0 looks terrific on my Microsoft Surface and I suspect that the anti-aliasing is a problem that diminishes with higher pixel density. So it may not be a future problem.

[...]

PR #1249, opened 2018-04-02
Revert frac-lines to borders

ronkok:
This PR reverts the rendering of frac-lines from SVG paths back to span borders. This solves the thick grey bar reported by @mbourne in issue #1173.

The result is a frac-line similar to KaTeX v0.9.0-alpha1. The frac-lines are span borders and the CSS contains a `min-height: 1px;` rule to keep those borders visible.

There is one difference between this PR and v0.9.0-alpha1. The earlier work contained a second CSS rule:
```
@media screen and (-webkit-min-device-pixel-ratio: 2), screen and (min-resolution: 192dpi) {
      .katex .mfrac .fracline {
            min-height:0.5px;
    }
}
```
That second CSS rule did not help much. Instead, it caused some lines to disappear.  So this PR omits that second CSS rule.

In the thread to issue #1173, people tested the rendering from the proposed approach and reported that frac-lines disappeared only at sizes that were already too small to read.


src/environments/array.js
```diff
@@ -202,8 +201,9 @@ const htmlBuilder = function(group, options) {
             }
 
             if (colDescr.separator === "|") {
-                const separator = stretchy.ruleSpan("vertical-separator", 0.05,
-                    options);
+                const separator = buildCommon.makeSpan(
+                    ["vertical-separator"], [], options
+            );
```
ylemkimon: The indentation of this line seems to be wrong.

ronkok: Indentation fixed. Thanks for the note.

kevinbarabash:
@ronkok sorry for taking so long to look at this.  I'm generating the screenshots right now and will send you PR shortly.  I'd like to get this in before the next release.  Thank you for reaching out to the community to test this.

kevinbarabash:
@ylemkimon thanks for updating that last Chrome screenshot.

PR #1251, opened 2018-04-04
Fix lap vertical alignment?

### LINKED ISSUE #1153: \mathrlap bounding box does not expand for tall argument

### LINKED ISSUE #1234: \llap is misaligned after \frac
\frac{1}{2}\llap| in Katex renders
screenshot_20180321_140526

However in Mathjax it renders
screenshot_20180321_140453

edemaine:
This is an alternative to #1162 to fix #1234, based on a comment I made in the latter: https://github.com/Khan/KaTeX/issues/1234#issuecomment-374934462

I do not fully understand why this one-line change works, `position: absolute` always being a little mysterious, but it seems to work... at least, the examples in #1234 all seem to work correctly (I think):
```latex
\frac{a}{b}\llap| \sin\llap|1 \sqrt{\mathclap|a\mathclap|}\mathclap|
```
image
vs. current release:
image

I post this here for further consideration, in particular by @ronkok, to decide whether this is actually good and could potentially simplify #1162.  He likely understands the layout better than me and can better judge.  #1162 is still necessary to fix issue #1153.

### DIFF
--- a/src/katex.less
+++ b/src/katex.less
@@ -274,6 +274,7 @@
 
         > .inner {
             position: absolute;
+            top: 0;
         }
 
         > .fix {


ronkok:
This PR does indeed fix issue #1234. That would be enough if issue #1234 were the only problem.

But this PR does nothing to fix issue #1153. That issue is mis-named; it's really a vertical alignment problem with `\rlap` when it must deal with tall elements.  So PR #1162 adds some struts that will make the HTML bounding box tall enough. When those struts are added, then the CSS `top: 0;` tactic of this PR is no longer enough to fix the vertical alignment. So PR #1162 applies `makeVList` to get it all aligned properly.

The remaining Chrome horizontal alignment bug in PR #1162 is an odd one. The gap shrinks when there is more math written on the same line.  So the result of  `\frac{a}{b}\llap| \sin\llap|1 \sqrt{\mathclap|a\mathclap|}\mathclap|` doesn't look so bad in PR #1162.

llap

So maybe I could overcome the Chrome bug by wrapping inside another span?  I dunno. I'll try.

edemaine:
OK, cool. I don't think this PR is useful then.

PR #1252, opened 2018-04-04
Specify height for SVG elements

ian-r-rose:
I ran into an issue where a low-specificity CSS rule elsewhere on the page was overriding the KaTeX SVG height, rendering fraction lines invisible. Does this seem like a reasonable addition to the style sheet?

### DIFF
--- a/src/katex.less
+++ b/src/katex.less
@@ -445,7 +445,7 @@
         display: block;
         position: absolute; // absolute relative to parent
         width: 100%;
-        height: 100%;
+        height: inherit;
 
         // We want to inherit colors from our environment
         fill: currentColor;
--- a/src/katex.less
+++ b/src/katex.less
@@ -445,6 +445,7 @@
         display: block;
         position: absolute; // absolute relative to parent
         width: 100%;
+        height: 100%;
 
         // We want to inherit colors from our environment
         fill: currentColor;


ian-r-rose:
Ah, I see this is also related to #1249 and #1173. Feel free to close if this is not the direction you want to go.

ronkok:
I don't see how this PR could hurt. SVG `height` is normally set in each instance of an inline SVG. The exception, frac-lines, will no longer use SVGs if/when PR #1249 is merged. So I think this PR will have no effect. But again, it doesn't hurt.

ian-r-rose:
Here is a small example demonstrating the problem: a low-specificity, seemingly innocuous CSS rule will interfere with fraction line rendering:
```html
<!DOCTYPE html>
<meta charset="utf-8">
<link rel="stylesheet" href="http://cdnjs.cloudflare.com/ajax/libs/KaTeX/0.9.0/katex.min.css">
<script src="http://cdnjs.cloudflare.com/ajax/libs/KaTeX/0.9.0/katex.min.js"></script>
<style>
  svg {
    height: auto;
  }
  .example {
    font-size: 30px;
    text-align: center;
    margin: 30px 0;
  }
</style>
<body>

  <div class="example tex" data-expr="\frac{a}{b}">
  </div>

<script>
  window.onload = function() {
      var tex = document.getElementsByClassName("tex");
      Array.prototype.forEach.call(tex, function(el) {
          katex.render(el.getAttribute("data-expr"), el);
      });
  };
</script>
```

ronkok:
I can confirm both the diagnosis and the proposed solution. And it applies to more than just frac-lines. Other SVGs, like `\sqrt`, have the same behavior. 

So I retract my earlier comment and I very much endorse this PR.  Thank you, @ian-r-rose.

PR #1253, opened 2018-04-06
README: Add size badge

daniellimws:
Add a size badge showing size of gzipped katex.min.js

It looks like the release script already handles changing the version so I shall not touch that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,6 +2,7 @@
 [![Build Status](https://travis-ci.org/Khan/KaTeX.svg?branch=master)](https://travis-ci.org/Khan/KaTeX)
 [![codecov](https://codecov.io/gh/Khan/KaTeX/branch/master/graph/badge.svg)](https://codecov.io/gh/Khan/KaTeX)
 [![Join the chat at https://gitter.im/Khan/KaTeX](https://badges.gitter.im/Join%20Chat.svg)](https://gitter.im/Khan/KaTeX?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge)
+![](http://img.badgesize.io/Khan/KaTeX/v0.9.0/dist/katex.min.js?compression=gzip)
 
 KaTeX is a fast, easy-to-use JavaScript library for TeX math rendering on the web.
 


edemaine:
Do we want to show the minified-but-not-gzipped size as well?  (I'm not really sure -- do most http servers send files gzipped now? If so, probably not important.)

ylemkimon:
The release script only handles `package.json` and CDN links.

daniellimws:
But I see something like [this](https://github.com/Khan/KaTeX/commit/cf6743b2a8e0d8694f48d1c70afde03d47a73b94) being committed in the release script. Or was it done manually?

README.md
```diff
@@ -2,6 +2,7 @@
 [![Build Status](https://travis-ci.org/Khan/KaTeX.svg?branch=master)](https://travis-ci.org/Khan/KaTeX)
 [![codecov](https://codecov.io/gh/Khan/KaTeX/branch/master/graph/badge.svg)](https://codecov.io/gh/Khan/KaTeX)
 [![Join the chat at https://gitter.im/Khan/KaTeX](https://badges.gitter.im/Join%20Chat.svg)](https://gitter.im/Khan/KaTeX?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge)
+![](https://img.badgesize.io/Khan/KaTeX/v0.9.0/dist/katex.min.js?compression=gzip)
```
ylemkimon: It seems that `badgesize` doesn't support https.

ylemkimon: There seems to be two alternative:
(i) use `shields.io` file size badge: https://img.shields.io/github/size/Khan/KaTeX/dist/katex.min.js%3Fref%3Dv0.9.0.svg (Note: the `ref` part is not an official API)
(ii) calculate file sizes in release script and create custom badge: e.g., https://img.shields.io/badge/size-234%20kb-brightgreen.svg

daniellimws: I think we can use the one from shields.io

ylemkimon:
Unfortunately, it seems it doesn't support gzipped size. Also, should we show sizes of `katex.css` and fonts?

kevinbarabash:
I looked at the shields.io page and saw at least one badge for a minified zipped file.  I don't understand their API at all though.  If we give it the URL for the gzipped file does it just magically do the right thing?

daniellimws:
I think there's one that checks the npm package.

daniellimws:
Should I change it back to badgesize.io?

ylemkimon:
@daniellimws Yes, please.

Issue #1254, opened 2018-04-07
xleftarrow/xrightarrow superscript incorrectly positioned


pzinn:
The superscript of xleftarrow/xrightarrow seems to have a fixed position, that's independent of its size. The result is, when the superscript is large, it will be too low and possibly overlap with the arrow.
example:
```
R \xrightarrow{\begin{pmatrix}a\\b\\c\end{pmatrix}} S
```

edemaine:
Fixed by #1256.

bhushan-ix:
This is fixed for HTML but for mathML its still not positioned correctly.
<img width="1430" alt="Screenshot 2024-10-22 at 7 42 47 PM" src="https://github.com/user-attachments/assets/1730e896-19b8-488b-a4c9-ab1ba7d5bcd8">

ronkok:
If you are interested in rendering MathML, I suggest the [Temml library](https://temml.org/). KaTeX's MathML will render your example correctly in Firefox, but Chromium renders MathML per the MathML Core specification, not MathML 3.0. MathML Core does not recognize the `<mpadded>` padding attributes that were in MathML 3.0, so the rendering is incorrect in Chromium.

There are lots of differences like that one and Temml attempts to write MathML that renders correctly in both browsers. For instance, your example is shown at [this link](https://temml.org/#%5Cxrightarrow%7B%5Ctext%7Btesting%20text%7D%7D).

By the way, if you use Temml to render extensible arrows, you'd better ship the Latin Modern font. The MathML core spec has what I consider to be [a bug](https://github.com/w3c/mathml-core/issues/121#issuecomment-1686729281) regarding extensible arrows and only the Latin Modern renders these arrows really well.

PR #1256, opened 2018-04-08
Fix extensible arrow sup vertical alignment

ronkok:
Fixes issue #1254


ronkok:
@ylemkimon You make a good point. I should be more nuanced. So the most recent commit adds the upper element depth to the shift only if: (1) upper element depth > 0.25 em, or (2) the arrow type is a left equilibrium. These are the two cases where we get interference between upper element and arrow.

For a visual comparison, the first image is from this PR.  The second is from AMS-LaTeX. In both images, the "g" vertical alignment is not adjusted, but the big array is adjusted.

revisedarrow

latexarrow

PR #1258, opened 2018-04-08
Remove special \verb space handling now that space is in Typewriter-Regular

edemaine:
Simplifies some old `\verb` code.  I've tested that this works once https://github.com/KaTeX/katex-fonts/pull/41 gets merged and pointed to by the submodule. (We can't merge until then.)

Example of working behavior together with fixed `\texttt` behavior:
image


edemaine:
@ylemkimon Thanks for the review. I had to update the screenshots as they differ at the subpixel level. (I tried a screenshot diff, but I couldn't see any red or green, as the differences are too small.) Are you still OK with merging?

ylemkimon:
@edemaine Sorry, I was busy and missed the notification of the comment.

PR #1263, opened 2018-04-19
[plugin system] Add defineSymbol to the main katex object

HosseinAgha:
Discussed in #762   
I did not add anything to contribution guidelines as this is still in beta but I think it is ready to be merged.   
I will also add a plugin template to the project when the API is finalized.  
Please check out the `PluginManager` class as it contains the main logic. Then you can check out a KaTeX plugin in action: [persian-katex-plugin](https://github.com/HosseinAgha/persian-katex-plugin).   
Here are some things that I was not sure about or I needed your guidance to implement:  
- [x] KaTeX uses closures to store its state (functions, symbols, etc). This makes the plugins changes  permanent (you cannot reset the library to default). A good approach that I can think of is to use classes to keep these data structures but this will break the main KaTeX API as users have to instantiate the KaTeX class in order to use it. So I did not touch the old system.
- [ ] I could not find a clean way to expose the KaTeX `flow types` to plugins. Do you have any suggestions?
- [ ] For plugins to dynamically inject their new css classes I could think of 2 solutions:
    - Migrate KaTeX styles to a css in js solution like [aphrodite](https://github.com/Khan/aphrodite)
    - Ask user to add plugins css in addition to `KaTeX.css`. 
      I used this approach as I wasn't sure if it is OK to change the old behaviour.
- [ ] I think migrating KaTeX development server and github.io website to a `React App` like the one used in [persian-katex-plugin](https://github.com/HosseinAgha/persian-katex-plugin) opens up whole new possibilities e.g. we can have buttons to apply different plugins to the library and check out their effect. I'm ready to implement that if you are OK.
- [x] I also added `yarn.lock` as there is no harm for it to be there plus it provides faster and safer installation for people who use [yarn](https://yarnpkg.com/en/) and most known libraries usually have it in addition to `package-lock.json`: _just removed yarn.lock from this PR_
- [x] [persian-katex-plugin](https://github.com/HosseinAgha/persian-katex-plugin) adds new symbols, fonts and fontMetrics so it only uses `defineSymbol` and `addFontMetrics`. I'm not sure if we need to expose anything more than `defineFunction` to other plugins as I haven't used it yet: _we will not expose defineFunction in this PR_

All the current tests are passing.   
I will also add more tests for the plugin system (and squash the commits) when we finalize the API.


ylemkimon:
Thank you for the pull request! Could you separate code style fixes to another PR?

HosseinAgha:
@ylemkimon I dropped the cleanup commit

src/PluginManager.js
```diff
@@ -0,0 +1,29 @@
...
+ * plugins can extend or change library's behaviour at the runtime.
+ */
+import { defineSymbol } from './symbols';
+import { defineFunction } from './functions';
+import { addFontMetrics } from './fontMetrics';
+
+const compiler = {
+    defineSymbol,
+    defineFunction,
+    addFontMetrics,
```
kevinbarabash: Why not just expose these functions directly on the `katex` object?  There isn't really a concept of plugins internally.  I think `defineSymbol` is relatively stable, but `defineFunction` requires access to a lot of internal classes to be useful.

HosseinAgha: These functions are basically internal utilities of KaTeX.  
I did not think exposing them to normal users was a good idea on the other hand advanced users can use them without defining a new javascript class.

HosseinAgha:
@kevinbarabash 
- I reverted all remaining cleanups 
- Added defineSymbol to main katex object only in this PR
- removed the `yarn.lock` from the PR
- created a new PR for changes related to addFontMetric: [katex #1269](https://github.com/Khan/KaTeX/pull/1269)
- rebased and squashed all the commits

src/PluginManager.js
kevinbarabash: We can expose them with a `__` underscore prefix to indicate that they should be used with caution.

HosseinAgha: Yes I have done this.   
Please checkout my next comment.

edemaine:
I'm not sure if this is the PR for it, but it occurs to me that exposing defineMacro (for global macros, as an alternative to passing them in) could also be handy.

kevinbarabash:
Exposing `defineMacro` as `__defineMacro` could be a separate PR.

src/symbols.js
```diff
@@ -19,7 +19,7 @@
 
 import type {Mode} from "./types";
 
-type Font = "main" | "ams";
+type Font = string; // can be "main" | "ams" | any other injected font by plugins
```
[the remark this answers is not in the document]
HosseinAgha: It will not work until user defines a `font-face` and adds appropriate font metrics to use a new font.   
I will revert this as we have to deal with this issue in #1269. These types are not much useful until we add katex libdef to [flow-typed](https://github.com/flowtype/flow-typed) project or use something like [.js.flow](https://github.com/flowtype/flow-typed/wiki/FAQs#what-is-the-preferred-way-for-a-module-to-provide-its-own-libdef) files when they got stable.
After deleting this I don't think we need any new tests in this PR.

src/symbols.js
```diff
@@ -19,7 +19,7 @@
 
 import type {Mode} from "./types";
 
-type Font = "main" | "ams";
+type Font = "main" | "ams"
```
kevinbarabash: I'm surprised that this passed the linter.

kevinbarabash:
> I could not find a clean way to expose the KaTeX flow types to plugins. Do you have any suggestions?

I need to read up on this more, but it sounds like you can define an index.js.flow file.

>  For plugins to dynamically inject their new css classes I could think of 2 solutions:
> - Migrate KaTeX styles to a css in js solution like aphrodite
> - Ask user to add plugins css in addition to KaTeX.css.  I used this approach as I wasn't sure if it is OK to change the old behaviour.

My hope is that we can reduce the need for CSS in general in the future.  It limits our ability to support other rendering backends in the future like canvas or SVG.  Obviously with the case of adding support for additional fonts some CSS has to be defined somewhere in order to define new font faces.  I think what you did is fine.

> I think migrating KaTeX development server and github.io website to a React App like the one used in persian-katex-plugin opens up whole new possibilities e.g. we can have buttons to apply different plugins to the library and check out their effect. I'm ready to implement that if you are OK.

Being able to try out different plugins would be awesome.  I'd try making that work first in the dev server environment.  One thing I'd like to investigate before introducing React for the test app is look at restructuring this repo to be a mono-repo.  I'm going to open new issue that describes the benefits of such an approach.

Issue #905, opened 2017-09-26
add "Differences from MathJax" section


kevinbarabash:
Here's an incomplete list of differences:
- fractions have less vertical space by default
- effect of color commands continue until the end of the enclosing group or another color command
- `\text{x % y}` errors because the `\text` command isn't closed

edemaine:
Is there a way we can automatically test features supported by each?  It'd be cool to have a side-by-side table of checkboxes (and motivate future PR work).

kevinbarabash:
Here's the list of functions that MathJax supports: http://docs.mathjax.org/en/latest/tex.html.  We could maybe write a scraper for that page and then autogenerate our own list which is what #103 was suggesting.

ronkok:
@kevinbarabash, I began a list with similar motivations at the end of [wiki's function support page](https://github.com/Khan/KaTeX/wiki/Function-Support-in-KaTeX). I haven't maintained that page since I started the [HTML function support page](https://khan.github.io/KaTeX/function-support.html). So the wiki page is now out-of-date. Which brings up some questions:

1. Do you want me to recreate in the [HTML page](https://khan.github.io/KaTeX/function-support.html) the material that is now in the bottom of the [wiki page](https://github.com/Khan/KaTeX/wiki/Function-Support-in-KaTeX)?
2. Do you want that page to identify the functions that are from non-AMS packages?
3. Is it possible to make the wiki link redirect to the HTML page, so that the HTML page seems to be part of the wiki?
4. If not, do we want to keep (and maintain) both the wiki page and the HTML page?

kevinbarabash:
@ronkok I think we probably want to deprecate that wiki page as it's out of date.  I was thinking that we'd have a section within README.md on the differences so that it's more prominent. 

I don't think it's possible to make the wiki page redirect, but I think we could updating the wiki page to have just a link to the HTML documentation.  If we still have a link to that wiki page from the README.md then we should probably change it to point to the HTML page.

By non-AMS, do you mean things like `\color` which are from the `color` package or things like `KaTeX` which aren't in any standard TeX package?

ronkok:
Re non-AMS: I do mean exactly packages like `color` and `mathtools` that have several functions now incorporated into KaTeX. I guess this is the same issue as #793. I'm just considering whether such information belongs on the function support page.

Re wiki page: I am thinking that I might erase all the material on the wiki page and leave behind a link to the HTML page. Any objections?

ronkok:
[...]

PR #1264, opened 2018-04-20
Mathjax Scraper

Kully:
For https://github.com/Khan/KaTeX/issues/905

I wrote a Python Scraper to grab all the functions from  http://docs.mathjax.org/en/latest/tex.html and inject them into the README. WDYT?

It's still a WIP.

To Do:
-  [ ] automate script excecution (i.e. `! python mathjax_scraper.py`)
-  [ ] inject the current KaTeX functions in the right column (https://github.com/Khan/KaTeX/issues/103) 

cc. @kevinbarabash @ronkok


kevinbarabash:
@Kully cool.  Injecting the KaTeX functions should be too hard b/c you can load functions.js, symbols.js, environments.js, and macros.js and then read the various tables which store those things.

Kully:
> b/c you can load functions.js, symbols.js, environments.js, and macros.js and then read the various tables which store those things.

How would I do that?

Also, what do you think about a bash script that runs the script each time a commit is merged or something like that? So that we are automating the README updating

ronkok:
This is an excellent idea. It can take the place of [this page](http://utensil-site.github.io/available-in-katex/), which seems to be un-maintained these days.

I wonder, though, if the README page is the best place. README already has a lot of information. Maybe a page in the [wiki](https://github.com/khan/katex/wiki) would be better?

kevinbarabash:
> I wonder, though, if the README page is the best place. README already has a lot of information. Maybe a page in the wiki would be better?

I agree that it feels like too much info for the README.  I'm not sure how difficult it would be to update the wiki with a script.  I guess we could manually copy/paste it in there.  Maybe a `FEATURES.md` or something like that.

Kully:
> This is an excellent idea. It can take the place of this page

Do either of you know why that page isn't rendering properly? It was from an old PR iirc

ronkok:
That page took its KaTeX from the `bower` distribution. These days, `bower` is deprecated and KaTeX now uses `jsdelivr` for its CDN. So its link to KaTeX is a dead link. I've made a PR to that repository that would update the link. No response.

kevinbarabash:
> > b/c you can load functions.js, symbols.js, environments.js, and macros.js and then read the various tables which store those things.

> How would I do that?

@Kully I was thinking something along the lines:
```
import symbols from "./src/symbols.js";
import functions from "./src/functions.js";
import environments from "./src/environments.js";
import macros from "./src/macros.js";

const commands = [
   ...Object.keys(symbols.math),
   ...Object.keys(symbols.text),
   ...Object.keys(functions),
   ...Object.keys(environments),
   ...Object.keys(macros),
].sort();

// inject commands in .md file
```

kevinbarabash:
I think the difficult it how to make the JS code work with your python code.  I would probably rewrite the python code in JS and then you'd be able to easily access the KaTeX and MathJax list of commands at the same time.

kevinbarabash:
@Kully will you be able to update this PR?

Kully:
> @Kully will you be able to update this PR?

Yes I will, I am just on vacation for the month of July.

ylemkimon:
Closing in favor of #1536. Thank you for your contribution!

PR #1267, opened 2018-04-22
update fonts to use \not character without left side-bearing correction, fixes #1189

### LINKED ISSUE #1189: \not broken with new spacing code
0.9.0 seems to have introduced a bug in this simple example: `\not E`

0.9.0 renders as follows:
image
```html
<span class="katex-html" aria-hidden="true">
  <span class="strut" style="height:0.69444em;"></span>
  <span class="strut bottom" style="height:0.8888799999999999em;vertical-align:-0.19444em;"></span>
  <span class="base">
    <span class="mrel latin_fallback" style="position:absolute;padding-left:0.8em;"≯</span>
    <span class="mord rule" style="margin-right:0.2777777777777778em;"></span>
    <span class="mord mathit" style="margin-right:0.05764em;">E</span>
  </span>
</span>
```

0.9.0-alpha2 (as on the current demo site) renders correctly:
image

```html
<span class="katex-html" aria-hidden="true">
  <span class="strut" style="height:0.69444em;"></span>
  <span class="strut bottom" style="height:0.8888799999999999em;vertical-align:-0.19444em;"></span>
  <span class="base">
    <span class="mord mathit" style="padding-left:0.277771em;position:relative;">
      <span class="mainrm" style="position:absolute;right:0;"≯</span>
      <span class="mord mathit" style="margin-right:0.05764em;">E</span>
    </span>
  </span>
</span>
```

kevinbarabash:
Fix #1189


kevinbarabash:
Safari
screen shot 2018-04-21 at 8 31 45 pm

Chrome
screen shot 2018-04-21 at 8 32 17 pm

ylemkimon:
> I can't see visual diffs of the screenshot images in git anymore. Anyone know why that might be?

GitHub has changed default diff for all file types to the `source diff`. You can view the image diff by clicking `Display the rich diff`:
image

kevinbarabash:
@edemaine this is ready for review again.

src/buildHTML.js
```diff
@@ -150,11 +150,6 @@ export const buildExpression = function(expression, options, isRealGroup,
     for (let i = 0; i < groups.length; i++) {
         if (groups[i].value === "\u0338") {
             groups[i].style.position = "absolute";
```
edemaine: Is any of this loop still needed?  Can you test without it?

kevinbarabash: It is still necessary.  The `position = "absolute"` causes the `\u0338` to overlay the the following group/atom.

edemaine: Is this equivalent to `\rlap`? An alternative would be to macro `\not` to `\rlap{\not@}` and rename the symbol to `\not@`.  I'm not quite sure what the intended behavior is, as LaTeX's source code is pretty opaque:
```latex
\DeclareMathSymbol{\not}{\mathrel}{symbols}{"36}
```

edemaine:
FWIW, I checked that the screenshot behavior (still) matches LaTeX (as you've probably done before):

not

src/buildHTML.js
kevinbarabash: I'll give the `\rlap` approach a try.  I'm not sure what to make sure of that chunk of LaTeX's source though.

kevinbarabash:
Yeah, I'm pretty happy with the output.  The implementation is a little hacky.  Hopefully using `\rlap` will work. 🤞

kevinbarabash:
screen shot 2018-05-29 at 12 04 22 am

I ended up doing `defineMacro("\\not", "\\mathrel{\\mathrlap{\\not@}}");`  Almost perfect expect that the width of the `mord` is 2px when it should be 0px.  I'm not sure where the width is coming from.  The children all have widths, but the widths are marked as implicit when the browser is displaying the computed styles.

edemaine:
Hmm, I've been unable to reproduce this, on Chrome for Windows.  I am observing that the `\mathrlap` causes a `vlist` (for some reason?), and some of those `vlist` elements have width 2px but a margin-left of -2px, which should cancel out by design.  (Something about zero-width elements not rendering the same.)  So my guess is that they're actually positioned correctly.  Did you check whether the screenshots changed?

kevinbarabash:
I was testing on Safari.  I did not check whether the screenshots changed.  I'll run the screenshot tests locally and see what happens.

src/buildHTML.js
```diff
@@ -149,12 +149,8 @@ export const buildExpression = function(expression, options, isRealGroup,
     // Process \\not commands within the group.
     for (let i = 0; i < groups.length; i++) {
         if (groups[i].value === "\u0338") {
+            // Results in a solidus being overlaid over the following group/atom.
             groups[i].style.position = "absolute";
```
edemaine: Back to this approach, I wonder if this could be done when building the symbol, instead of a `for` loop after the fact?

ronkok:
>  I am observing that the `\mathrlap` causes a `vlist` (for some reason?)

The reason is to keep the `\mathrlap`ed item attached to the baseline. Consider an item in a `\mathrlap` that is located on the same line of text with something tall. The `\mathrlap` item would align its top to the top of the tall item.

By wrapping it in a `vlist`, I was able to keep it aligned to the baseline. There may be other, better ways to do that. I used `vlist` because it was easy.

edemaine:
P.S. @kevinbarabash If you'd like to leave these alternatives to another PR, this one is also fine as is -- it's an improvement over the existing code!

kevinbarabash:
We should address the issue with glyphs that use `0x338` that was raised in https://github.com/KaTeX/katex-fonts/issues/47 before merging this.

PR #1268, opened 2018-04-22
add prettier along with pre-commit hook to autoformat code, fixes #1266

### LINKED ISSUE #1266: Add prettier to update formatting automatically as a pre-commit hook
The PR that adds this should also submit all the formatting changes.

kevinbarabash:


.eslintrc
```diff
@@ -20,7 +20,6 @@
         "eqeqeq": [2, "allow-null"],
         "guard-for-in": 2,
         "indent": "off",
-        "indent-legacy": [2, 4, {"SwitchCase": 1}],
```
kevinbarabash: This was conflicting with some of the formatting that prettier outputs.

.prettierrc.json
```diff
@@ -0,0 +1,10 @@
+{
+    "printWidth": 84,
+    "parser": "flow",
+    "tabWidth": 4,
+    "semi": true,
+    "singleQuote": false,
+    "trailingComma": "all",
+    "bracketSpacing": false,
+    "arrowParens": "always"
```
kevinbarabash: Many of these are prettier defaults, but I thought being explicit about what the settings were was better.

src/Lexer.js
```diff
@@ -37,20 +37,21 @@ const commentRegexString = "%[^\n]*[\n]";
...
+    "([ \r\n\t]+)|" + // whitespace
+    `(${commentRegexString}` + // comments
+    "|[!-\\[\\]-\u2027\u202A-\uD7FF\uF900-\uFFFF]" + // single codepoint
+    `${combiningDiacriticalMarkString}*` + // ...plus accents
+    "|[\uD800-\uDBFF][\uDC00-\uDFFF]" + // surrogate pair
+    `${combiningDiacriticalMarkString}*` + // ...plus accents
+    "|\\\\verb\\*([^]).*?\\3" + // \verb*
+    "|\\\\verb([^*a-zA-Z]).*?\\4" + // \verb unstarred
+    `|${controlWordRegexString}` + // \macroName
+    `|${controlSymbolRegexString}` + // \\, \', etc.
```
kevinbarabash: One downside of prettier is that it doesn't maintain custom comments.  I would suggest moving to putting comments on their own lines instead of after code if we want nice formatting.

kevinbarabash:
I'm going to update this after everything has been moved into /functions.

edemaine:
Does prettier implement anything useful that lint doesn't check for? Personally, I'm finding our lint precommit hook annoying lately, because lint is really slow (on my computer, maybe 5-10 seconds) and I usually commit after a full `npm test` (though there are exceptions). So now I'm in the habit of `git commit -n` to avoid the preconmit hook. Adding more to the hook would be inconvenient... Also, I feel a little weird about having code modify my code from under me.

kevinbarabash:
> Does prettier implement anything useful that lint doesn't check for? Personally, I'm finding our lint precommit hook annoying lately, because lint is really slow (on my computer, maybe 5-10 seconds) and I usually commit after a full npm test (though there are exceptions). So now I'm in the habit of git commit -n to avoid the preconmit hook. Adding more to the hook would be inconvenient... Also, I feel a little weird about having code modify my code from under me.

@edemaine lint takes about the same time for.  Most of what we use lint for can be replaced with prettier.  Removing the rules that prettier covers might speed up a lint a little bit, but probably not by much.  The nice thing about prettier is that it'll fix any issues automatically.  It feels a little weird at first having something muck with your code.  It's changed how I code though.  I no longer waste time thinking about line length or indentation or trailing commas or whatever.  I let prettier do it's thing.

One thing we might look at to speed up lint is https://github.com/alansouzati/eslint-parallel.

kevinbarabash:
I tried `eslint-parallel` and there was no appreciable speed up.

kevinbarabash:
Maybe instead of adding prettier we can turn on a couple of more eslint rules.  The main thing that I wanted was some consistency about how we do `import`s with destructuring.  In some places we have `import { foo } from "./foo"` and in other places we have `import {bar} from "./bar"`.

edemaine:
This sounds preferable to me.  Thanks!

Issue #729, opened 2017-06-11
Add support for Persian & Arabic alphabet in text and eastern-arabic numerals + Proposal Commit


HosseinAgha:
Currently countries with more than 700 million people combined use Persian, Arabic and Urdu alphabets worldwide.  
These countries also have a huge number of students. Iran itself has 18 million students for example.  
Because of this we see more and more educational startups that are targeting these countries.  
I tried to add support for symbols of these languages to KaTeX with as minimal footprint as possible.     
I think the results are pretty good. I haven't added any screenshotter tests yet but by testing with `make serve` and by eye I think I got a fairly precise result.  
We face 2 challenges when we try to add these symbols:  
1. They use [Eastern Arabic Numerals](https://en.wikipedia.org/wiki/Eastern_Arabic_numerals) as opposed to [(Western) Arabic Numerals](https://en.wikipedia.org/wiki/Arabic_numerals) in their schools.
2. In English alphabet for example each character code has a single glyph with a specified size but The characters of these alphabets are cursive so they change shape and size.  

- Adding the symbols was easy because like CJK characters, all these alphabets are in a uniform `[0x0600 - 0x06FF]` unicode range.  
- I couldn't use KaTeX default fonts because they don't have this character set and the glyphs have serious problems. So I added a new (open) font called [Vazir-Code](https://github.com/rastikerdar/vazir-code-font). The biggest advantage of this font is that it is monospace so cursive characters have a roughly same height and width everywhere.  
- I calculated font metrics using awesome [fontkit](https://github.com/devongovett/fontkit) package.
- Then I added the Eastern Arabic numerals as `textords` in `math` mode and alphabet as `textords` in `text` mode.  
- finally I added a new font beside `main` and `math` font.  
- in `buildCommon` I add a new css class that only sets the font-family to Vazir-Code.   

IMHO we can have 2 different approaches from here:  

1. If the changes generally looks good to you, I can add the font to the main `fonts` folder of KaTeX, calculate metrics using the python script and finally add some test and screenshotter test.  
2. We also can add a small plugin system to KaTeX so everyone can add their own fonts and css to the system an library should consume the plugins symbols.  

IMHO first approach is better solution here because Latin (I mean every latin based alphabet), CJK and Persian-Arabic alphabets constitute majority of the alphabets in use.  

### You can find my changes in this commit: https://github.com/HosseinAgha/KaTeX/commit/8696318b7509681a75dfa213009111cc818a66d0  

[...]

PR #1269, opened 2018-04-22
[plugin system] Add a utility function (setFontMetrics) to extend builtin fontMetrics

HosseinAgha:
Store current font metrics object inside a `metricMap` variable.  
Also export a function called `addFontMetrics` from `fontMetrics.js` and expose it on the main `katex` object so plugins can add their own font metrics to library in run time.  
Checkout #762 and #729


src/fontMetrics.js
```diff
@@ -183,14 +183,28 @@ export type CharacterMetrics = {
...
+export type MetricMap = {
+    [string]: { [string]: number[] }
+}
+let metricMap = {};
+
+/**
+ * This function extends metricMap object with a new metric information
+ */
+export function addFontMetrics(newMetrics: MetricMap) {
+    metricMap = Object.assign(metricMap, newMetrics);
```
kevinbarabash: I'm not sure how `Object.assign` works with nested objects.  Will is merge sub objects or overwrite them?

test/katex-spec.js
```diff
@@ -98,10 +98,12 @@ const parseAndSetResult = function(expr, result, settings) {
     } catch (e) {
         result.pass = false;
         if (e instanceof ParseError) {
-            result.message = "'" + expr + "' failed " +
+            result.message = () =>
```
kevinbarabash: These changes seem unrelated to `addFontMetrics`.  Why the switch to a callback?

src/fontMetrics.js
HosseinAgha: It shallowly merges the source object into destination object so sub objects will be overwritten take a look at [this link](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign#Cloning_an_object).   
It was a intended behaviour to prevent users from partially changing a font metrics. as fonts come as a whole file so their metrics should also come as whole.

test/katex-spec.js
HosseinAgha: I initially made a mistake in my code and some of tests failed but the error I got was `result.message is not a function`. I looked at the documentation of jest and found out that message should be a function.   
It seems like tests for these parts of `katex-spec` always passed so we never got any error that covered these error types so we never found out.

HosseinAgha:
Please take a look at [persian-katex-plugin](https://github.com/HosseinAgha/persian-katex-plugin) this is a complete implementation of Persian formulas for KaTeX using this branch.   
You can also find out how I generated the metrics using [generateDefaultMetrics.js](https://github.com/HosseinAgha/persian-katex-plugin/blob/master/generateDefaultMetrics.js) and ultimately [extractMetricsFromTTF.js](https://github.com/HosseinAgha/persian-katex-plugin/blob/master/utils/extractMetricsFromTTF.js).  
I'm not sure how to add proper tests because these test need to add new fonts and metrics to the KaTeX object.   
I also would be very happy if I could add the [generateMetrcis.js](https://github.com/HosseinAgha/persian-katex-plugin/blob/master/utils/generateMetrics.js) as a utility function to the main library.

test/katex-spec.js
edemaine: I've encountered these bugs throughout as well. They are sometimes in my PRs, though I guess we should do a thorough cleanup.

kevinbarabash:
@HosseinAgha I had a look at your plugin.  I'm still a little confused about which classes are being set on the spans containing characters from the Vazir font.  In katex.less we define classes like `.texttt` that result in the `font-family` being set to `KaTeX_Typewriter;`.  How does this work with your plugin?

kevinbarabash:
Answering my own question.  I checked out your repo and ran the test app.  It looks like `Vazir-Regular` is being set as one of the classes on the span containing the number `۹`.  This is super cool!  I'm happy to see it was too much work.  I'll spend some more time reviewing this PR this evening and get back to you regarding your question about how to test this.

src/fontMetrics.js
kevinbarabash: I'm a little concerned both about the behavior and the naming of this function.  This function can actually override metrics depending on the keys in `newMetrics`.  Also, it could be a partial overwrite.  To make things more predictable, I'd like to suggest changing this to:
```
function setFontMetrics(fontName, metrics) {
   metricMap[fontName] = metrics;
}
```
The problem with this is we have a function called `getFontMetrics` which returns a dictionary of sigma and xi values which are used for positioning things.  The issue is that the names are symmetric, but the values they deal with are completely different which is unexpected.

The thing is, is that `getFontMetrics` actual returns things that look like font metrics.  Font metrics usually refer to values that apply to a font as a whole, e.g. units per em, max ascender, max descender and the like.

Can you change the name of `getFontMetrics` to `getGlobalMetrics`?  Then we can define `setFontMetrics` as described above without there being confusion between the two.

src/fontMetrics.js
```diff
@@ -183,14 +183,28 @@ export type CharacterMetrics = {
...
+}
+// add the default font metrics
+addFontMetrics(defaultMetricMap);
```
kevinbarabash: I think we can just `import metricMap from "../submodules/katex-fonts/fontMetricsData";` above and not bother calling `addFontMetrics` here.

test/katex-spec.js
kevinbarabash: I think I've run into the same issue now that you mention it.  We should add `// @flow` to the test files at some point to (not now).

src/buildCommon.js
```diff
@@ -636,8 +645,8 @@ const retrieveBaseFontName = function(font: string): string {
         case "texttt":
             baseFontName = "Typewriter";
             break;
-        default:
-            throw new Error(`Invalid font provided: ${font}`);
+        default: // probably a font added by a plugin
+            baseFontName = font;
```
kevinbarabash: The comment should probably be more definitive, e.g.
```
default:
   baseFontName = font;  // use by fonts added by a plugin
```
I like that things line up in your plugin.  I wish we didn't have so many mappings for font related things that we have to maintain.

src/buildCommon.js
```diff
@@ -151,12 +151,19 @@ const mathDefault = function(
             return makeSymbol(
                 value, fontName, mode, options,
                 classes.concat("amsrm", options.fontWeight, options.fontShape));
-        } else { // if (font === "main") {
+        } else if (font === "main" || !font) {
```
kevinbarabash: Is the `!font` necessary?  It looks like all symbols defined in `symbol.js` have a font of either `"ams"` or `"main"`.

HosseinAgha: It is the same line that prevented the failing test mentioned above.  
In some special cases font was undefined. It looked strange for me but I assumed this is needed in somewhere else so I did not change the old behaviour.

HosseinAgha:
Thank you @kevinbarabash for your review.  
I'll start fixing the issues tonight.   
As there is a significant time zone difference between us I think you will get them tomorrow :)

src/fontMetrics.js
HosseinAgha: Done

src/buildCommon.js
HosseinAgha: Done

HosseinAgha:
@kevinbarabash Is it OK if we warn the user if one of the default KaTeX metrics was being overwritten?  
something like (in setFontMetrics): 
```javascript
if (metricMap.hasOwnProperty(fontName)) {
  console.warn(`metrics for ${fontName} is being overwritten by a plugin`)
}
```

src/buildCommon.js
kevinbarabash: My guess is that we don't define a font for a lot of the unicode characters that we all now.  @ronkok is that correct?

src/buildCommon.js
```diff
@@ -632,7 +639,7 @@ const retrieveTextFontName = function(
             baseFontName = "Typewriter";
             break;
         default:
-            throw new Error(`Invalid font provided: ${fontFamily}`);
+            baseFontName = fontFamily; // use fonts added by a plugin
```
kevinbarabash: It would be nice if we could check when we get an invalid font.  Right now there's no list of valid fontFamilies.  Ideally we should have one and setFontMetrics should update that list with the font that's being added.  Also, this makes me wonder if we should change `__setFontMetrics` to be `__defineFont` and have it take something a little more structured, e.g.:
```
__defineFont(
  baseFontName, 
  fontFamily, 
  metrics: { regular: {...}, bold: {...}, italic: {...}, "italic-bold": {...} },
);
```
That's a rather big change.  Would you be able to tackle this in a follow up PR?

src/buildCommon.js
ronkok: > don't define a font for a lot of the unicode characters that we [allow].

@kevinbarabash  I think our Unicode work as falling into two groups. 

One group began with your PR #410 that added support for Greek characters already in KaTeX fonts. My work has followed in that path, adding support for as many Unicode characters as I could, but always characters that are in KaTeX fonts. In this string of PRs, we have always specified a font.

The other group of PRs began with your PR #508 for CJK and was extended, especially by @davidflanagan and lately by @edemaine's PRs #1217 and #1278 that allow, in non-strict mode, all Unicode code points. This string of PRs does not, to my knowledge, specify a font, but rather lets the browser pick a system font. And I believe that font metrics are assumed to match the Latin letter "M".

kevinbarabash:
@HosseinAgha thanks picking this up again.  There are some changes described in my comments above that I think would be helpful in making this more robust and easier for other people to use it.  I approved this PR though b/c I want to get this on to `master` so that we can continue to iterate on it.

HosseinAgha:
I added some tests as you requested and fix the conflicts I think we are ready to merge.  
@kevinbarabash Thank you for your review.

src/buildCommon.js
HosseinAgha: Yes, good idea.  
What is the difference between baseFontName and fontFamily here?

kevinbarabash:
@HosseinAgha thanks for adding tests.

PR #1276, opened 2018-04-28
Make ParseNode `value` payload and defineFunction handler functions type-safe

marcianx:
I can't think of any practical way of porting `buildHTML` and `buildMathML` over without avoiding a lot of `$FlowFixMe`s before making `ParseNode` typesafe. So this has to be done first. Learned a nice type system trick along the way thanks to some internet searches (see `ParseNode`).
Making the defineFunction handler typesafe had to be done as well, since those handlers generate the vast majority of the ParseNode payloads and porting them ensured that the `ParseNode` payload types were actually accurate.

There will be a (hopefully easier) follow-up that will make the arguments of the `htmlBuilder` and `mathBuilder` parameters of `defineFunction` type safe.

I also saw an opportunity to do a minor refactor of `styleMap` and `StyleStr`, but I'm going to leave that to another PR. I'd rather get this (rather pervasive) PR in sooner rather than later.


src/ParseNode.js
```diff
@@ -10,15 +13,15 @@ import type {Mode} from "./types";
  * For details on the corresponding properties see `Token` constructor.
  * Providing such information can lead to better error reporting.
  */
-export default class ParseNode {
-    type: *;
-    value: *;
+export default class ParseNode<TYPE: NodeType> {
```
kevinbarabash: `ParseNode<T: NodeType>` is more common when defining parameterized types.

src/ParseNode.js
```diff
@@ -31,3 +34,239 @@ export default class ParseNode {
         this.loc = SourceLocation.range(first, last);
     }
 }
+
+export type NodeType = $Keys<ParseNodeTypes>;
```
kevinbarabash: Cool!

src/ParseNode.js
```diff
@@ -31,3 +34,239 @@ export default class ParseNode {
...
+        text: string,
+        math?: string,
+    },
+    isStretchy: boolean,
+    isShifty: boolean,
+    base: ParseNode<*>,
+};
+
+// Map from `type` field value to corresponding `value` type.
+type ParseNodeTypes = {
```
kevinbarabash: Wow, this is great!

src/ParseNode.js
```diff
@@ -31,3 +34,239 @@ export default class ParseNode {
         this.loc = SourceLocation.range(first, last);
     }
 }
+
+export type NodeType = $Keys<ParseNodeTypes>;
+export type NodeValue<TYPE: NodeType> = $ElementType<ParseNodeTypes, TYPE>;
+
+export type AccentStructType = {
```
kevinbarabash: Why is this not an exact type?

src/ParseNode.js
```diff
@@ -31,3 +34,239 @@ export default class ParseNode {
...
+    // ("accent" and "op" handled above to capture additional variants.)
+    "bin": string,
+    "close": string,
+    "inner": string,
+    "mathord": string,
+    "open": string,
+    "punct": string,
+    "rel": string,
+    "spacing": string,
+    "textord": string,
```
kevinbarabash: All of the groups typed as a `string` stick out a like a sore thumb.  No action required on this, but something we'll probably want to refactor in the future.

src/Parser.js
```diff
@@ -554,10 +557,10 @@ export default class Parser {
      */
     parseArguments(
         func: string,   // Should look like "\name" or "\begin{name}".
-        funcData: FunctionSpec | EnvSpec,
+        funcData: FunctionSpec<*> | EnvSpec,
```
kevinbarabash: Will `EnvSpec` need to be parameterized eventually as well?

src/Parser.js
```diff
@@ -977,11 +980,12 @@ export default class Parser {
...
         if (match) {
             for (let i = 0; i < match[0].length; i++) {
-                const accent = match[0][i];
+                const accent: string = match[0][i];
                 if (!unicodeAccents[accent]) {
                     throw new ParseError(`Unknown accent ' ${accent}'`, nucleus);
                 }
-                const command = unicodeAccents[accent][this.mode];
+                const command: ?{text: string, math?: string} =
+                    unicodeAccents[accent][this.mode];
```
kevinbarabash: Is `unicodeAccents` not typed in such a way that the proper type for `command` would be inferred?

src/defineFunction.js
```diff
@@ -92,23 +93,22 @@ type FunctionDefSpec = {|
...
     // Any other data can be added to the object, which will be passed
     // in to the function in buildHTML/buildMathML as `group.value`.
-    handler: ?FunctionHandler,
+    handler: ?FunctionHandler<NODETYPE>,
 
     // This function returns an object representing the DOM structure to be
     // created when rendering the defined LaTeX function.
-    // TODO: Port buildHTML to flow and make the group and return types explicit.
+    // TODO: Change `group` to ParseNode<NODETYPE> and make return type explicit.
     htmlBuilder?: (group: *, options: Options) => *,
```
kevinbarabash: What's preventing us from doing this TODO now?

src/defineFunction.js
```diff
@@ -119,7 +119,8 @@ type FunctionDefSpec = {|
  * 2. requires all arguments except argTypes.
  * It is generated by `defineFunction()` below.
  */
-export type FunctionSpec = {|
+export type FunctionSpec<NODETYPE: NodeType> = {|
```
kevinbarabash: I would change this to `export type FunctionSpec<T: NodeType>`.

src/defineFunction.js
```diff
@@ -128,27 +129,44 @@ export type FunctionSpec = {|
     numOptionalArgs: number,
     infix: boolean,
     consumeMode: ?Mode,
+
+    // FLOW TYPE NOTES: Doing either one of the following two
+    //
+    // - removing the NOTETYPE type parameter in FunctionSpec above;
+    // - using ?FunctionHandler<NODETYPE> below;
```
kevinbarabash: I would've expected `?FunctionHandler<NODETYPE>` to work. :\

src/functions/styling.js
```diff
@@ -30,11 +30,15 @@ defineFunction({
         parser.consumeSpaces();
         const body = parser.parseExpression(true, breakOnTokenText);
 
+        // TODO: Refactor to avoid duplicating styleMap in multiple places (e.g.
+        // here and in buildHTML and de-dupe the enumeration of all the styles).
+        // $FlowFixMe: The names above exactly match the styles.
+        const style: StyleStr = funcName.slice(1, funcName.length - 5);
```
kevinbarabash: Can you explain in more detail what the flow error was here?

src/stretchy.js
```diff
@@ -107,7 +107,7 @@ const mathMLnode = function(label: string): mathMLTree.MathNode {
 // corresponds to 0.522 em inside the document.
 
 const katexImagesData: {
-    [string]: ([string[], number, number] | [string[], number, number, string])
+    [string]: ([string[], number, number] | [[string], number, number, string])
```
kevinbarabash: Why did we change from an array of strings to a one-tuple in the second option?

src/stretchy.js
```diff
@@ -159,27 +159,35 @@ const katexImagesData: {
...
         span: DomSpan | SvgSpan,
         minWidth: number,
         height: number,
     } {
         let viewBoxWidth = 400000;  // default
-        const label = group.value.label.substr(1);
+        // The "accent" ParseNode value could be a string since it's one of the
+        // groups enumerated in symbols.js. But the logic here assumes that it's
+        // a struct.
+        // $FlowFixMe
```
kevinbarabash: Why is this a flow error though?  You've typed it as a struct below and the above comment makes it sound like is already a struct.

kevinbarabash:
It looks like one of the changes introduced a parsing error in the Smash screenshot test.  I'm surprised all of the unit tests pass.  It might be worth adding a unit test that fails before fixing the issue.

src/ParseNode.js
marcianx: I'm happy to change it if you insist. I just wanted to make sure here and in other places where I've used a parametrized type that it emphasizes that this is the `type` and not the `value` of the parse node. WDYT?

src/defineFunction.js
marcianx: Me too! This error is very confusing. ;(
I didn't yet try to do a minimally-reproduceable example to file a bug on flow.

src/Parser.js
marcianx: Unsure. I'll see once I try to port buildHTML over.

src/Parser.js
marcianx: There are no flow types on `unicodeAccents` it since it's also used by `unicodeMake`. So I unfortunately had to do it this way.

src/defineFunction.js
marcianx: I tried it, but there were type failures. This requires handling more type assertions in the various  `htmlBuilder`s and `mathmlBuilder`s. This is the kind of stuff within `buildHTML` and `buildMathML` as well.

I figured that since this is already a lot and is in a good spot and catches a lot of errors already, I'd do the rest in a separate PR.  ;)

src/Parser.js
```diff
@@ -402,7 +402,9 @@ export default class Parser {
             }
         }
 
-        if (superscript || subscript) {
+        // Base must be set if superscript or subscript are set per logic above,
+        // but need to check here for type check to pass.
+        if ((superscript || subscript) && base) {
```
marcianx: +kevinb
This is the failure
```
\begin{matrix}
    ^3+[-1][1-1]1=1(=1)\lvert a\rvert~b \\
    \scriptstyle{^3+[-1][1-1]1=1(=1)\lvert a\rvert~b} \\
    \scriptscriptstyle{^3+[-1][1-1]1=1(=1)\lvert a\rvert~b} \\
\end{matrix}
```
and I can see that would fail with this change in the code. This examples has a superscript `^3` without a base! Isn't that supposed to be a parse failure in LaTeX? Why do we support it?

marcianx:
I can see what would cause it. Could you possibly look at the place in the code where I just left a comment? I need your feedback on this.

src/Parser.js
marcianx: Interestingly, there are already tests for this and this is apparently a feature. I'll check to see why they didn't fail.

marcianx: Fixed and pushed.

src/defineFunction.js
marcianx: This is definitely non-trivial and definitely belongs in a separate PR. There would need to be more refinement of the `ParseNode` types. It's better explained by sharing the next PR with you once it's done. ;)

marcianx:
I just pushed another commit to this PR to make the `ParseNode` payload more type-safe. This requires renaming two of the symbol `Group`s -- "accent" -> "accent-token" and "op" -> "op-token" -- since these group names are _also_ used as `ParseNode` types. Before this, the "accent" `ParseNode` had value type
`string | AccentStructType`. Now, that's been split into:

- "accent": AccentStructType
- "accent-token": string

This leads to better type safety and fewer type assertions when porting the `htmlBuilder` and `mathmlBuilder` to flow.

edemaine:
I haven't looked at the PR, but wanted to mention that I'm not a fan of the way that functions (provided by `defineFunction`) get their contents wrapped inside `value`, often resulting in `value: {value:...}`. This may not be the time to fix it though.

marcianx:
@edemaine We want to get to the point where we can do that refactor. It would be far easier to catch errors once `buildHTML` and `buildMathML` are ported to flow, which is what this PR is toward. The goal is to eventually end up with a more flattened `ParseNode` format described here
https://github.com/Khan/KaTeX/issues/892
at the end of a fruitful brainstorm.

src/ParseNode.js
kevinbarabash: I hadn't thought about that.  I guess we can leave it as is.

marcianx:
@kevinbarabash I just want to confirm: Your review is not waiting on me for anything, correct? I just want to make sure I didn't overlook anyrthing.

src/Parser.js
marcianx: This was wrong. I no longer needed to add the type. I'd misread the original double-nested map lookup. I ran into this while working on the second PR.

src/Parser.js
marcianx: The answer is yes. This is small enough that I can do it fully (both the `hander` and the `htmlBuilder` part) in one fell swoop. It requires a helper from the next PR, so I'm going to keep it as part of the next PR for now.

src/defineFunction.js
kevinbarabash: Well, at least we have a detailed comment here that we can dig into later.

src/functions/styling.js
kevinbarabash: Gotcha.  It would be nice to limit the slicing to remove leading `\` as much as possible.  Not sure how much that would help with this flow error though.

src/stretchy.js
kevinbarabash: This is such a weird typing.  Hopefully we can making things a bit more regular in the future.  No action required.

src/stretchy.js
kevinbarabash: You can check if `typeof(group.value) !== "string"`, e.g. https://flow.org/try/#0PTAEAEDMBsHsHcBQiAuBPADgU1AZxQE4DyBACrAJYB2KoAvHodQOagA+oA3gB4BcoVAK4BbAEZYCAGlBp+QsRIC+AbmTQstDJRr98xMttoMe-AKzTZoAIwAGFcgqRQACnTZYkZ1uooAlKABCOgYAIj0WEP9ORFBQAGNYKlxYdQA6OGYvQ1TuX1VFRCA

src/symbols.js
```diff
@@ -20,9 +20,13 @@
 import type {Mode} from "./types";
 
 type Font = "main" | "ams"
+// Some of these have a "-token" suffix since these are also used as `ParseNode`
+// types for raw text tokens, and we want to avoid conflicts with higher-level
+// `ParseNode` types. These `ParseNode`s are constructed within `Parser` by
+// looking up the `symbols` map.
```
kevinbarabash: Does using the same string actually cause an issue?  It seems like they're different types so `flow` should keep them straight for us.  Am I missing something?

marcianx: Yes, `Group` here is a different type, but _every_ string in `Group` is also used as a `ParseNode` type due to [this line in `Parser.js`](https://github.com/Khan/KaTeX/blob/ba24d2073ba44664a3ba82d0d22c309834b60385/src/Parser.js#L968-L969), resulting in this unfortunate name conflict. All of these `Group` values are listed under the second "section" of types within `ParseNode.js` where the value type is `string`. See the comment in this PR in `ParseNode.js` that says

    // From symbol groups, constructed in Parser.js via `symbols` lookup.
    // (Some of these have "-token" suffix to distinguish them from existing
    // `ParseNode` types.)

src/Parser.js
kevinbarabash: Sorry for the delay response.  I tested the above code on quicklatex.com and it works.  Thanks for fixing the issue.

src/symbols.js
kevinbarabash: I see.  Yeah, reusing those strings feels bad news.  Thanks for highlighting this issue.

Issue #1275, opened 2018-04-28
\underset and exponent


pzinn:
in KaTeX, 
```
\underset{*}{x}^2
```
results in the exponent 2 being right above x (instead of above to the right in LaTeX).
is that a known/expected behavior?

kevinbarabash:
We must be assuming that any character inside `\underset` should be typeset using the limits structure.  Looks like a bug.  Thanks for reporting.

ronkok:
@kevinbarabash I can confirm that `\underset` and `\overset` both use the limits structure. However, `\stackrel` is done differently and apparently in a more robust manner. I'll see if I can extend the `\stackrel` method to `\underset` and `\overset`.

ronkok:
@pzinn I just turned in PR #1277, which will fix this issue.

edemaine:
Fixed by #1277. Thanks for the report!

PR #1277, opened 2018-04-28
Fix \underset

ronkok:
This PR modifies `\underset` and `\overset` so that they will behave like LaTeX when an exponent is applied after the `\underset`. Fixes issue #1275.

I also changed the spacing from KaTeX `mop` to KaTeX `mrel`. The result seems to be a better match to LaTeX.


ronkok:
Visual confirmation:
underset

edemaine:
It looks like a few screenshots need to be updated.

ronkok:
I have never mastered Docker and so am unable to update screenshotter tests. In the past, @kevinbarabash  has been kind enough to update them for the PRs that I submit. In the most recent example, he ran the test; then pasted the two relevant screenshot files into a comment on the PR's comment thread.  I took the png files and copied them into the screenshotter folder in that branch.

That could work again if anyone would care to help me out.

edemaine:
I updated the screenshots for you (pushing directly to your branch).  Check out the https://github.com/Khan/KaTeX/pull/1277/files?short_path=da83f1e#diff-da83f1efcd086fe5f05c338688da3c18 to make sure it looks like the right new behavior.

ronkok:
@edemaine  The new screenshots look perfect to me. Feel free to proceed with final approval and a merge.

PR #1278, opened 2018-04-30
Implement strict mode (replacing unicodeTextInMathMode)

### LINKED ISSUE #1226: Strict mode proposal
* We currently have the (wordy) `unicodeTextInMathMode` option to switch between what you might call "strict LaTeX behavior" (no Unicode text symbols in math mode) and "flexible/permissive behavior" (which aligns with MathJax and is probably more convenient to use, especially if you don't want to export to regular LaTeX/XeTeX later).
* Another example of this is #1167, which wants to allow non-boxes as arguments to `\raise`/`\lower`.  Again we have a natural "strict LaTeX compatibility" vs. "flexible/convenient behavior".  I imagine a recurring theme here.
* `src/functions/kern.js` warns about `mu` units used by a spacing command not starting with `m`, and warns about non-`mu` units used by a spacing command starting with `m`.  This seemed like a reasonable compromise between strict LaTeX compatibility (let users know that they're breaking the rules) and flexibility (annoying to keep track of LaTeX's many different spacing functions).

I propose that we define a global setting `strict` which controls this dichotomy.  `strict: true` or `strict: "error"` (allow one or both forms?) can force strict LaTeX behavior as much as possible, and raise an error upon violation.  `strict: "warn"` can check for LaTeX compatibility, but just print a warning upon misuse.  Maybe `strict: (msg) => ...` can provide a callback for custom handling / warning.  `strict: false` can be the default permissive mode where violations are ignored.  We could implement this with a helper, maybe a `Parser.checkStrict` method where `Settings` are readily available, that implements the appropriate behavior.

Let me know your thoughts!  I think this will help us avoid making repeatedly difficult (and possibly inconsistent) decisions about this dichotomy, though there are probably multiple ways to resolve it.

edemaine:
Add new `"strict"` setting (default value false) that can take a boolean (whether to throw an error or silently ignore), string (`"ignore"`, `"warn"`, or `"error"`), or a function possibly returning such a value. This enables a variety of ways of handling or ignoring transgressions from "true" LaTeX behavior, making KaTeX easier to use while still providing the ability for strict LaTeX adherance.

Resolve #1226, implementing that spec, for two existing transgressions from regular LaTeX:

* `src/functions/kern.js` had some errors and warnings about use of (units in) math vs. text mode commands.
* The former setting `unicodeTextInMathMode` (not in any released version) needed to be set to true to enable Unicode text symbols in math mode.

Now these are controlled by the `strict` setting.  By default, KaTeX is now very permissive, but if desired, the user can request warnings or errors.


README.md
```diff
@@ -72,8 +72,10 @@ You can provide an object of options as the last argument to `katex.render` and
...
-- `unicodeTextInMathMode`: `boolean`. If `true`, supported unicode text characters like `é` and `試` will also work in math mode. (They always work in text mode.) The default is `false`, which matches XeTeX behavior; `true` emulates MathJax behavior.
 - `maxSize`: `number`. If non-zero, all user-specified sizes, e.g. in `\rule{500em}{500em}`, will be capped to `maxSize` ems. Otherwise, users can make elements and spaces arbitrarily large (the default behavior).
+- `strict`: `boolean` or `string` or `function` (default: `false`). If `false` or `"ignore`", ignore features that make writing LaTeX convenient but are not actually supported by (Xe)LaTeX (similar to MathJax). If `true` or `"error"` (LaTeX compatibility mode), throw an error for any such transgressions. If `"warn"`, warn about such behavior via `console.warn`. Provide a custom function `handler(errorCode, errorMsg, token)` to default behavior depending on the type of transgression (summarized by the string code `errorCode` and detailed in `errorMsg`).  A list of such features and their `errorCode`s:
```
ylemkimon: Doesn't `compatibility mode` usually mean non-strict mode?

edemaine: Agreed, "compatibility" maybe isn't the right term.  Perhaps "adherence" or "faithfulness"?

edemaine: I've pushed a rewording of this paragraph that in particular uses "faithfulness" instead of "compatibility".

src/Settings.js
```diff
@@ -45,10 +53,37 @@ class Settings {
...
-            utils.deflt(options.unicodeTextInMathMode, false);
+        this.strict = utils.deflt(options.strict, false);
         this.maxSize = Math.max(0, utils.deflt(options.maxSize, Infinity));
     }
+
+    /**
+     * Report nonstrict (non-LaTeX-compatible) input.
+     * Can safely not be called if `this.strict` is false in JavaScript.
+     */
+    nonstrict(errorCode: string, errorMsg: string, token?: Token | ParseNode) {
```
kevinbarabash: I like this.  Maybe rename `token` to `tokenOrParseNode`.

src/Settings.js
```diff
@@ -45,10 +53,37 @@ class Settings {
...
+            strict = strict(errorCode, errorMsg, token);
+        }
+        if (!strict || strict === "ignore") {
+            return;
+        } else if (strict === true || strict === "error") {
+            throw new ParseError(
+                "LaTeX-incompatible input and strict mode is set to 'error': " +
+                `${errorMsg} [${errorCode}]`, token);
+        } else if (strict === "warn") {
+            typeof console !== "undefined" && console.warn(
```
kevinbarabash: When is `console` not defined?

static/main.js
```diff
@@ -28,6 +28,17 @@ function init() {
...
     }
 
+    // Use `strict=warn` for warning strict mode or `strict=error`
+    // (or `=1`/`=t`/`=true`/`=y`/`=yes`)
+    // to turn off displayMode (which is on by default).
+    if (query.strict) {
+        if (query.strict.match(/^(1|t|y|e)/)) {
+            options.strict = "error";
+        } if (query.strict && query.strict.match(/^(w)/)) {
+            options.strict = "warn";
```
kevinbarabash: I like the simplicity of this.

src/Settings.js
```diff
@@ -1,20 +1,28 @@
...
 import utils from "./utils";
+import ParseError from "./ParseError.js";
+import ParseNode from "./ParseNode";
+import {Token} from "./Token";
 
 import type { MacroMap } from "./macros";
 
+export type StrictFunction =
+    (errorCode: string, errorMsg: string, token?: Token | ParseNode) =>
+    ?(boolean | string);
```
kevinbarabash: Exposing this level of customizability to users seems like overkill.

src/Settings.js
```diff
@@ -1,20 +1,28 @@
...
+
 export type SettingsOptions = {
     displayMode?: boolean;
     throwOnError?: boolean;
     errorColor?: string;
     macros?: MacroMap;
     colorIsTextColor?: boolean;
-    unicodeTextInMathMode?: boolean;
+    strict?: boolean | string | StrictFunction;
```
kevinbarabash: I think it's sufficient if `strict?: boolean | "ignore" | "warn" | "error"`.

src/Settings.js
edemaine: I was matching `ParseError`'s constructor.  If I change one, I should probably change both...  I could also see something like `where` or `context` being a good name.

src/Settings.js
edemaine: I can't think of such an environment, but presumably they might exist (or the global console could be overwritten?).  In any case, I'm following existing code from the KaTeX codebase, in particular:

```js
        typeof console !== "undefined" && console.warn(
            "No character metrics for '" + value + "' in style '" +
                fontName + "'");
```

It might be good to define a `katex.warn` or something that can be overwritten by the user, but defaults to `console.warn`.

src/Settings.js
edemaine: This level of customizability lets users support some strict behavior and not others.  @ylemkimon at least seemed to think this would be useful, and it's not exactly hard to support.  Thoughts?

src/Settings.js
edemaine: I replaced `string` with `"ignore" | "warn" | "error"` but left `StrictFunction` for now.

src/Settings.js
edemaine: Another point for allowing a function is that it enables custom warning functionality, if `console.warn` isn't the desired warning output.  For example, this would make our own testing easier, as `jest` doesn't really handle warnings well.

src/Settings.js
kevinbarabash: Oh wow, I didn't realize we were doing that elsewhere.

src/Settings.js
ylemkimon: I think it's nice to have customizability as long as we have simpler options to set them. We may add lots of features to the non-strict mode, and `strict` option may become too generic.

src/Parser.js
```diff
@@ -960,15 +963,20 @@ export default class Parser {
...
             symbol = new ParseNode(symbols[this.mode][text].group,
                             text, this.mode, nucleus);
-        } else if (supportedCodepoint(text.charCodeAt(0)) &&
-            (this.mode === "text" || this.settings.unicodeTextInMathMode)) {
+        } else if (supportedCodepoint(text.charCodeAt(0))) {
+            if (this.settings.strict && this.mode === 'math') {
+                this.settings.nonstrict("unicodeTextInMathMode",
+                    `Unicode text character "${text[0]}" used in math mode`,
+                    nucleus);
+            }
```
ylemkimon: How about allowing all unicodes(non-ASCII) on non-strict mode? (#1217, `allowAllSymbols`)

edemaine: I was wondering about that, but planning to deal with it when we return to discussing #1217. `allowAllSymbols` feels like more of a hack, not to be encouraged -- the rendering will probably be poor in general, and it's really a sign that we need to add symbols to the fonts.  So I was thinking of keeping that a separate option... but I'm open to suggestions.  (And maybe it is worth discussing here, in case we want to tweak the `strict` functionality to deal with it...)

ylemkimon: I think that's where the customizability should come in. Normal users would intend used symbols to be shown, but advanced users or who worries about the layout would want to know unsupported symbol has been used.

edemaine: OK, I agree.  I think this is more motivation to set the default `strict` value to `"warn"`, so at least people get a sense of why the rendering is funny in these cases.

edemaine:
I've gone ahead and changed the default setting for `strict` to `"warn"`.

I also added functionality to check for warnings in Jest, turning them into a special type of `Error`.  These changes are also a first step toward #1126.

edemaine:
@ylemkimon Thanks for the review!
@kevinbarabash Are you OK with this being merged, with the following decisions?
* Default `strict` setting is `"warn"`
* Allow functions to be specified in `strict` argument

edemaine:
OK, I propose merging this first, then I can add nonstrict behavior to `\\`. @kevinbarabash , are you OK with leaving function support in?

edemaine:
The changes to make strict mode control `\newline` behavior are extensive, so I'm going to put them in a separate PR.

PR #1279, opened 2018-04-30
Add version to katex object

### LINKED ISSUE #1190: katex.version
For debugging purposes, it would be handy if `katex.version` would return the version string for the loaded version of KaTeX.

EthanLuisMcDonough:
Fix #1190

### DIFF
--- a/katex.js
+++ b/katex.js
@@ -20,6 +20,8 @@ import type ParseNode from "./src/ParseNode";
 
 import { defineSymbol } from './src/symbols';
 
+import { version } from "./package.json";
+
 /**
  * Parse and build an expression, and place that expression in the DOM node
  * given.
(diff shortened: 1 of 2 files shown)


EthanLuisMcDonough:
Thanks for the feedback.  Changes have been made.

edemaine:
@EthanLuisMcDonough Just checking: did you test that indeed `katex.version` returns the right version in a built version of `katex.js` / `katex.min.js` (e.g. using the test server run via `npm start`)?

EthanLuisMcDonough:
Yes, I tested.  `katex.version` returns `"0.10.0-pre"`.

edemaine:
I further confirmed that, after `npm build`, `build/katex.js` has a line starting
```js
module.exports = {"name":"katex","version":"0.10.0-pre",...
```
We'll want to make sure that this gets the actual build number when using the release script, but looks like it should work.

PR #1283, opened 2018-05-02
Greek caps

### LINKED ISSUE #1282: support for non-LaTeX greek symbols
Currently, greek letters can be input either as TeX (e.g., \alpha) or directly as unicode (\u03b1), but in both cases *only if they're defined in TeX*. Is there any reason to forbid greek letters that TeX didn't define, e.g., capital Eta? They could still be input as e.g. \u0397. The change to symbols.js would be minimal.

ronkok:
Supports Greek letters ΑΒΕΖΗΙΚΜΝΟΡΤΧ.

Resolves #1282.


ronkok:
Visual confirmation:
greekcaps

PR #1284, opened 2018-05-03
Nonitalic Greek capital letters

ronkok:
This PR fixes my error in PR #1283 and makes Greek capital letters non-italic.


ronkok:
Visual confirmation:
nonitalgreek

ronkok:
Well, I'm looking for something that works in both math mode and text mode. `\mathrm` doesn't work in text mode.  We could use `textrm` in place of `\rm`. I don't have a strong preference between those two. 

Which would you prefer?

edemaine:
Ah right. Then instead you should add braces around each. (Currently, text after the Greek letter will also be nonitalic!)

ronkok:
Done. Thank you for pointing that out.

edemaine:
My next worry is that these won't behave the same as other Greek letters do within e.g. `\mathit`, `\mathbf`, `\boldsymbol`. Would it be better to add them as symbols and modify the Greek letter list in `Parser.js`?

src/macros.js
```diff
@@ -127,19 +127,19 @@ defineMacro("\u211D", "\\mathbb{R}");
...
+defineMacro("\u0396", "{\\rm{Z}}");
+defineMacro("\u0397", "{\\rm{H}}");
+defineMacro("\u0399", "{\\rm{I}}");
+defineMacro("\u039A", "{\\rm{K}}");
+defineMacro("\u039C", "{\\rm{M}}");
+defineMacro("\u039D", "{\\rm{N}}");
+defineMacro("\u039F", "{\\rm{O}}");
+defineMacro("\u03A1", "{\\rm{P}}");
+defineMacro("\u03A4", "{\\rm{T}}");
+defineMacro("\u03A7", "{\\rm{X}}");
```
edemaine: Sorry, could you also remove the braces around the letter? We just need `{\\rm A}` for example.

ronkok:
> Would it be better to add them as symbols

Yes, it would be better to add them as symbols. But in order for that to work, KaTeX font glyphs and font metrics are needed.  Otherwise, the code in `buildCommon.js` will not work.

Perhaps it would be better to close this PR and wait until someone adds the glyphs to the fonts. I don't personally think that these particular characters are the most urgent issue we have right now.

edemaine:
Given that I already merged #1283 (oops), we may want to still merge this... could use a second opinion, though.

ronkok:
> Given that I already merged #1283 

I would be happy to turn in a PR that reverts #1283. I agree that it would be good to get another opinion first.

PR #1285, opened 2018-05-03
Alternate approach to capital Greek letters

edemaine:
Alternative to #1283 and #1284 for solving #1282, implementing Latin-looking capital Greek letters via **symbols** instead of macros, which interacts with fonts more correctly.

Visual confirmation: (the first 5 letters are Latin, and remain unchanged, while the Β's to the right are capital betas)

image

This ends up making a mapping from the capital Greek letters to the Latin equivalents in the symbols table. I think it's better than macros because of the correct interaction with fonts. In the future, we could decide to populate the font with Greek capitals for better copy/paste behavior (see #1283 discussion), but the code would be nearly identical, so this is maybe good for now.


ronkok:
@edemaine You beat me to this PR by about ten minutes. I realized this was possible about an hour ago. 

My apologies to  you and @pzinn for wasting your time on a detour. Your approach here is much better.

edemaine:
Cool, glad you agree! I wouldn't have thought of this approach without going through all the ensuing discussion, so it's not time wasted.

Do you want to review this PR?

ronkok:
@edemaine  I don't believe that I can do a final review. I never did accept @kevinbarabash's invitation to collaborate in this repository, out of fear that my poor Git skills would cause some catastrophe. And it looks as though that invitation has now expired. So I remain a contributor.

For what it is worth, I've looked at your code and it looks perfect to me.

PR #1286, opened 2018-05-04
Tilde \textasciitilde and fix ~ in \verb

### LINKED ISSUE #1262: \verb renders ~ as space
I just noticed by accident that `\verb|~|` renders as a space (like `\verb| |`) instead of like a tilde as it should (does in LaTeX).

edemaine:
* Add support for `\textasciitilde` which reproduces tilde character in text mode.
* Make `~` render as tilde instead of space within `\verb`, by manually mapping that character to `\textasciitilde`. This seems to be the only character that needs such manual mapping to reproduce LaTeX's `\verb` behavior. Fix #1262.
* Need to merge #1258 first. (Only the [last commit](https://github.com/Khan/KaTeX/pull/1286/commits/1a9ebcba111a5f9e2517cb3518aa1238a08a88f1) is new.)


kevinbarabash:
@edemaine this is no longer blocked, but there's some conflicts that need resolving.

edemaine:
@kevinbarabash Thanks for all the reviews and merges! I resolved the conflicts here.

Issue #327, opened 2015-08-10
Break/wrap formula if too wide


scaljeri:
Is there a way to break formula's if they are to wide (with respect to their parent) ?

spicyj:
Not currently, but I was talking to @xymostech and the rules sound simple (breaking is allowed after any top-level rel or bin) so it might not be too difficult to add.

scaljeri:
+1 I really need this feature. Is `responsiveness` also included in the equation ?

spicyj:
I don't know what you mean by that, sorry?

scaljeri:
For example, if you switch from landscape to portrait, the `formula` adjusts to the space it has (responsive)

invkrh:
I think responsiveness is a big problem here.
Formules viewed from mobile break all responsive design.
They are too long, the view point is auto scaled to adapt these formulas.
I can not shorten the width of the formulas by setting width to a smaller value.
Really want to know any workaround.

mbourne:
@invkrh: Some possibilities until responsiveness becomes a feature in KaTeX:

(1) Give the enclosing element a style of "`overflow:auto`". Then your math will be full size and users can scroll the equation left and right to see it all.

For me, this has serious drawbacks, since on most mobile devices there is no indication that the element can be scrolled horizontally (no horizontal scrollbar appears like you get on a desktop browser, until you actually start scrolling the element. So if the page edge comes at an awkward part of the equation, it's possible the user has no idea that scrolling is possible and will never see the hidden part. It's possible to add a horizontal scroll bar with javascript of course).

Another option is display:table
http://netdna.webdesignerdepot.com/uploads7/how-to-create-horizontal-scrolling-using-display-table-cell/table-layout-horizontal/

(2) What I do is

(a) Manually break the equation into pieces at logical places (like equal signs, plus signs, minus signs, etc) so it will wrap logically

(b) For cases where splitting the equation is not possible, (it's a fraction, say), I nudge the font-size down with javascript until it does fit. I get away with that most of the time with the content on my site.

You can see both these things going on in this page (including in the drop-down answer sections):

http://m.intmath.com/fourier-series/2-full-range-fourier-series.php

The elements that cannot be split and that I know will not fit into a 300 px width are given the class "fitToWidth". The function "fitWidth" reduces the font-size if required.

I know it's clunky and could be done better (including the addition of an onRotate handler), but that will need to wait until the next revamp.

[...]

Issue #1233, opened 2018-03-21
Line breaks sometimes happen immediately after an inline KaTeX block


akalin:
This is undesirable if e.g. a comma follow an inline equation.

This seems to happen to me most often on Chrome for Linux (Ubuntu 17.10), but I'm not 100% sure I haven't seen it on Chrome for Mac.

Example [gist](https://gist.github.com/akalin/15067e7ec3f7fbdd07d4fbb11b192918) with [display](https://rawgit.com/akalin/15067e7ec3f7fbdd07d4fbb11b192918/raw/330f631b36b60251ff2586908c8204a99732e9c6/test-0.9.0.html). If you adjust the right border of the browser window, you can get it to break between "A = (ε_0 + H)/2" and its following comma.

I'm not sure if KaTeX has much control over this behavior, but I figure I'd report it in case anyone has any ideas.

Screenshot:

![screenshot from 2018-03-20 23-36-23](https://user-images.githubusercontent.com/425576/37697132-495ca600-2c98-11e8-97ff-a1a2220f3cc4.png)

akalin:
Writing the equation as "A = (ε_0 + H)/2\text{,}" does work in this case, but it would be nice not to have to do that. (Also, I'm not 100% sure that it sometimes doesn't break between the two anyway.)

edemaine:
I've encountered this too, and tweaked [my own KaTeX preprocessor](https://github.com/edemaine/coauthor/blob/master/lib/formats.coffee#L430) to put following punctuation inside any math blocks.  This definitely works, as KaTeX doesn't support any line breaks within a math block. 

It'd be nice if there'd be a pure CSS workaround...  Alternatively, auto-render and a solution to #604 (which I'm imagining roughly following the code linked above) could implement this hacky solution.

akalin:
Ah, cool, glad to hear this is a known thing. I was considering modifying auto-render, too, but I don't quite understand what #604 has to do with it. My idea would be as follows:

1) After autorender splits up a piece of text into a list of text and math blocks, but before it renders the math...
2) For each text block following an inline math block, if some option is set with a string of punctuation, or a regexp or something...
3) If the beginning of the text block matches the punctuation string/regexp, wrap it in a `\text{...}` fragment and append it to the previous inline math block. Looks like you use a nobr span; that might work, too.

Would this be something that would be useful to have in auto-render, or should I expect to just have it for myself?

kevinbarabash:
[...]

PR #1287, opened 2018-05-04
Line breaks for inline formulas

edemaine:
Allow for line breaks in inline formulas according to TeX's rules.  Display formulas are not affected.  Part of #327, specifically the plan suggested in https://github.com/Khan/KaTeX/issues/327#issuecomment-352278017

The TeXBook [[p.173](http://www.ctex.org/documents/shredder/src/texbook.pdf#page=184)] says "A formula will be broken only after a relation symbol like $=$ or $<$ or $\rightarrow$, or after a binary operation symbol like $+$ or $-$ or $\times$, where the relation or binary operation is on the ``outer level'' of the formula (i.e., not enclosed in {...} and not part of an \over construction)."

In inline math, we break the outer `katex-html` into multiple `base` spans, splitting wherever a line break can be (in particular, right after top-level `mop` or `mbin`).  Each `base` span gets its own strut, which handles vertical spacing between lines properly.  This was easier than I thought it'd be!

In addition, I added support for `\allowbreak` and `\nobreak` which modify this behavior.  By also marking them as (zero-width) spaces, I believe they don't affect anything else about layout.

Examples from the test server with `display=0` set:

image

image

image

image

image

image


src/katex.less
```diff
@@ -32,13 +36,6 @@
     // mode, while still allowing background/foreground setting on root .katex
     * { -ms-high-contrast-adjust: none !important; }
 
-    .katex-html {
-        // Making .katex inline-block allows line breaks before and after,
-        // which is undesireable ("to $x$,"). Instead, adjust the .katex-html
```
sophiebits: Are you able to verify that in `to $x$,` it's impossible to have a break between the `x` and comma? In a narrow container it should break before `x` instead.

edemaine: No, in fact this is #1233. I've left this comment as is, though it clearly is incorrect, and I can edit it. This PR has no effect on this issue.

ronkok:
Very, very cool stuff.  I do have some questions. 

Will `\\` be available as a mandatory line break?

In LaTeX, there are some nuances about where line breaks can occur. Some examples:

| **Item**                                                     | **Line break allowed** |
|:-------------------------------------------------------------|:-----------------------|
| skips                                                        | yes                    |
| kerns                                                        | no                     |
| `\,` or `~`                                                  | no                     |
| Inside boxes, including `\hbox`, `\raisebox`, or `\colorbox` | no                     |

Do you have any plan in mind for incorporating these nuances? Is it a matter of going through the existing code and setting an `allowbreak` flag or `nobreak` flag?

src/katex.less
sophiebits: Huh. I swear this used to work.

edemaine:
@ronkok Actually, I don't think those rules apply here.  We're dealing with line breaks *within math mode*. I just did some tests, and none of `\ `, `\quad`, `\hspace`, `\hskip` enable line breaks within math mode.  For example:

```latex
$this\ is\ some\ long\ text\quad
 this\ is\ some\ long\ text\quad
 this\ is\ some\ long\ text\quad
 ...
$
```

ronkok:
@edemaine You're right. I also did some testing and I can confirm that line breaks are not prevented in math mode by `\,`, `\ `, `\space`, or by kerns. That's terrific. It makes your work even better.

`~` does prevent a line break in math mode. So does `\nobreakspace`.

Also, boxes, e.g. `\raisebox` prevent any line breaks. Inside a box, even `\\` will fail to produce a line break. With boxes, can we do something with CSS? Something like a `nobreak !important` rule?

edemaine:
@ronkok I hope you mean "line breaks are forbidden at whitespace". The example I gave above does not get line breaks in LaTeX (or in this PR).  Line breaks only seem to be allowed at `\allowbreak` and after bins and ops.  In (inline) math, `~` and `\ ` are treated identically, so far as I can see.

This PR doesn't allow line breaks in any nested objects, which includes boxes.  It only looks for breakable things at the top level.  Same as TeX.  `\\` could similarly be dealt with at the top level only, presumably via a `<br>` tag.

ronkok:
That's odd. I get no line breaks when I try:
```
\(\aaaaaaaaaaaaaaaaaa+~aaaaaaaaaaaaaaaaaaaaaa+~aaaaaaaaaaaaaaaaaaaa+~aaaaaaaaaaaaaaaaaaaaaaa+~aaaaaaaaaaaaaa \)
```
It seems that the `~` directly after a `bin` prevents the line break that would otherwise occur.

I'm glad to hear about the boxes.

edemaine:
Oh, I see our confusion. I meant that spaces/glue by themselves do not permit line breaks. You're saying that a nonbreaking space *immediately after a valid line break place* prevents that line break.  Got it!  That should be easy to fix.

ronkok:
@edemaine In your most recent commit, you ensured that glue with stay with the operator. Could you make an exception for `\hspace*`? It exists to provide a space at the start of a line after a line break.

edemaine:
@ronkok Good idea.  I'd like to resolve #1289 first though; it's pretty confusing.

In the latest commit, I changed `~` and `\nobreakspace` to also have nonbreaking behavior.

edemaine:
@ronkok I read the source code for `\hspace*` and was able to implement it almost exactly like LaTeX (!).  Our macro system has come a long way...  LaTeX uses a `\rule` to protect the space from the line-breaking algorithm, so I did the same.

Demo:

hspace*

hspace

I also fixed a bug that allowed line breaks where I didn't want them (needed to set `white-space: nowrap` at *some* level, namely, `.base`).

ronkok:
Cool! This is really coming together well.

edemaine:
I've successfully implemented `\\`, at least for inline math (where LaTeX supports it), on top of this.  For simplicity of review, I've split it off into a separate PR, #1298.  That can be reviewed after this PR.

src/buildCommon.js
```diff
@@ -692,16 +692,28 @@ const spacingFunctions: {[string]: {| size: string, className: string |}} = {
...
 
 // A lookup table to determine whether a spacing function/symbol should be
 // treated like a regular space character.
-const regularSpace: {[string]: boolean} = {
-    " ": true,
-    "\\ ": true,
-    "~": true,
-    "\\space": true,
-    "\\nobreakspace": true,
+const regularSpace: {[string]: { className?: string }} = {
```
ylemkimon: Can `{ className?: string }` be `{| className?: string |}`?

src/buildCommon.js
```diff
@@ -692,16 +692,28 @@ const spacingFunctions: {[string]: {| size: string, className: string |}} = {
...
+        className: "nobreak",
+    },
+    "\\allowbreak": {
+        size: "0em",
+        className: "allowbreak",
+    },
 };
 
 // A lookup table to determine whether a spacing function/symbol should be
 // treated like a regular space character.
```
ylemkimon: Comments need to be updated

src/buildCommon.js
edemaine: This seemed to require `className` to be specified, i.e., flow then ignored the question mark. But perhaps a flow expert could comment...

edemaine: It seems to be this bug: https://github.com/facebook/flow/issues/2977

ylemkimon: I thought I saw it somewhere, and there it is 😄

src/buildHTML.js
```diff
@@ -638,25 +641,63 @@ export default function buildHTML(tree, options) {
...
+    // Create one base node for each chunk between potential line breaks.
+    // The TeXBook [p.173] says "A formula will be broken only after a
+    // relation symbol like $=$ or $<$ or $\rightarrow$, or after a binary
+    // operation symbol like $+$ or $-$ or $\times$, where the relation or
+    // binary operation is on the ``outer level'' of the formula (i.e., not
+    // enclosed in {...} and not part of an \over construction)."
+
+    let parts = [];
+
+    function newBody() {
```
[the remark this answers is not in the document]
edemaine: I like not having to repeat `htmlNode.children.push(body);` and `parts = []` but that may be just me...  If anyone has a third opinion, I'm happy to go with the majority!

src/buildCommon.js
edemaine: Glad you spotted this!  Updated.

edemaine:
I updated the screenshots, and there are lots of tiny changes in Chrome only, I believe because of Chrome's rounding-to-integer-pixel issues we've encountered in other contexts (e.g. sqrt line width).

Along the way, I realized I needed to move the `width: min-content` Chrome hack into `.base` instead of `.katex`, for negative spaces to work properly (or, at least as well as they did before -- the `NegativeSpaces` test is even slightly better on Chrome now).

There are still some more significant changes in the screenshots that I'm not sure about:

oplimits-chrome-diff

raisebox-chrome-diff

svgreset-chrome-diff

loweraccent-chrome-diff

horizontalbraces-chrome-diff

deepfontsizing-chrome-diff

arraymode-chrome-diff

src/buildHTML.js
```diff
@@ -638,25 +641,63 @@ export default function buildHTML(tree, options) {
...
+    htmlNode.setAttribute("aria-hidden", "true");
 
-    // Wrap the struts and body together
-    const htmlNode = makeSpan(["katex-html"], [topStrut, bottomStrut, body]);
+    // Create one base node for each chunk between potential line breaks.
+    // The TeXBook [p.173] says "A formula will be broken only after a
+    // relation symbol like $=$ or $<$ or $\rightarrow$, or after a binary
+    // operation symbol like $+$ or $-$ or $\times$, where the relation or
+    // binary operation is on the ``outer level'' of the formula (i.e., not
+    // enclosed in {...} and not part of an \over construction)."
```
kevinbarabash: Nice comment, thanks or linking to the TeXBook.

src/buildHTML.js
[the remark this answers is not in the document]
kevinbarabash: I think we should pull out this function and pass in parts and have it pass back a body.  It's a few extra lines, but it makes things easier to reason about b/c the function won't have any side effects once it's extracted.   Also, I think we should change the `body` to `line` and `newBody(parts)` to `buildLine(expressions)`.

src/buildHTML.js
```diff
@@ -638,25 +641,63 @@ export default function buildHTML(tree, options) {
...
+        parts = [];
+    }
 
-    htmlNode.setAttribute("aria-hidden", "true");
+    for (let i = 0; i < expression.length; i++) {
+        parts.push(expression[i]);
+        if ((utils.contains(expression[i].classes, "mbin") ||
+             utils.contains(expression[i].classes, "mrel") ||
+             utils.contains(expression[i].classes, "allowbreak"))) {
+            // Put any post-operator glue on same line as operator.
```
kevinbarabash: nice

src/buildHTML.js
```diff
@@ -638,25 +641,63 @@ export default function buildHTML(tree, options) {
...
+            while (i < expression.length - 1 &&
+                   utils.contains(expression[i + 1].classes, "mspace")) {
+                i++;
+                parts.push(expression[i]);
+            }
+            // Don't break at all if \nobreak immediately following
+            // (after glue).
+            if (!(i < expression.length - 1 &&
+                  utils.contains(expression[i + 1].classes, "nobreak"))) {
```
kevinbarabash: This `if` is confusing.  Maybe rewrite as:
```
const isNobreakNext = i < expression.length - 1 && utils.contains(expression[i + 1].classes, "nobreak");
if (!isNobreakNext) {
   newBody();
}
```

src/buildHTML.js
```diff
@@ -638,25 +641,63 @@ export default function buildHTML(tree, options) {
...
+        htmlNode.children.push(body);
+        parts = [];
+    }
 
-    htmlNode.setAttribute("aria-hidden", "true");
+    for (let i = 0; i < expression.length; i++) {
+        parts.push(expression[i]);
+        if ((utils.contains(expression[i].classes, "mbin") ||
+             utils.contains(expression[i].classes, "mrel") ||
+             utils.contains(expression[i].classes, "allowbreak"))) {
```
kevinbarabash: I'd create a helper called `hasClass` that can check if node in build tree has a specific class.

test/katex-spec.js
```diff
@@ -50,8 +50,13 @@ const _getBuilt = function(expr, settings) {
     // grab the root node of the HTML rendering
     const builtHTML = rootNode.children[1];
 
-    // Remove the outer .katex and .katex-inner layers
-    return builtHTML.children[2].children;
+    // combine the non-strut children of all base spans
```
kevinbarabash: Why?

edemaine:
I tracked down the bulk of the screenshot differences to this CSS rule:
```css
    .vlist-s {
        // This cell solves Safari rendering problems. It has text content, so
        // its baseline is used for the table. A very small font avoids line-box
        // issues; absolute units prevent user font-size overrides from breaking
        // rendering. Safari refuses to make the box zero-width, so we give it
        // a known width and compensate with negative right margin on the
        // inline-table.
        width: 2px;
```
The placement of `width: min-content` on the `.base` element instead of the `.katex-html` element ends up removing this space for some reason (in inspect mode, the `.vlist-s` element has a width of 0 instead of 2).  Adding a rule of `min-width: 2px` fixes the problem.

Now all screenshot diffs (against master) appear entirely black, meaning the differences are all subpixel, as expected, except for `ArrayMode` which still has a global horizontal shift as above.  I'll keep investigating that one.

edemaine:
What luck!  The horizontal offset in display math actually seems to be an improvement over past behavior.

Here's `ArrayMode` in master, with Chrome inspecting the `.base` element:

image

And here's the same test on this PR:

image

The new version has the correct bounding box on the `.base` element, resulting in correct centering.  The old centering was shifted too far to the right.  This is also pretty evident if you look at the screenshot images (e.g. in diff mode): [old screenshot](https://raw.githubusercontent.com/Khan/KaTeX/c2ae9e0da6e5dab1662444bae11b08d92842865d/test/screenshotter/images/ArrayMode-chrome.png) vs. [new screenshot](https://raw.githubusercontent.com/Khan/KaTeX/05adb3d39a8789346fbb8ad5bda7cb60f54b380e/test/screenshotter/images/ArrayMode-chrome.png).

So I'm happy with all the screenshots now.  I'll clean up the code according to the comments next.

test/katex-spec.js
edemaine: This is for backwards compatibility. Several tests delve into the contents of the children returned from `getBuilt`, for example:

```js
    it("should create ords when at the beginning of lists", function() {
        const built = getBuilt("+ x");

        expect(built[0].classes).toContain("mord");
        expect(built[0].classes).not.toContain("mbin");
    });
```

These tests want to ignore the `.base` elements, but now that there are multiple such elements, their children need to be concatenated together.  The `.strut`s used to be outside the `.base` elements, but now they're inside, so I also strip them away for these tests.

It occurs to me, though, that this simplification could be done in `getBuilt`, not `_getBuilt`.  This would be better.

edemaine:
@kevinbarabash I implemented all your comments on the code.  This should be good to go now, except for probably we should add some multiline line-breaking tests.

kevinbarabash:
@edemaine thanks.  This looks great.  I'm so happy you were able to sort out the horizontal spacing differences.  I'll do another pass this evening.

edemaine:
Added a screenshot test, which should test everything, including `\nobreak`, `\allowbreak`, and `\hspace` vs. `\hspace*`.  While testing, I realized that I broke `\nobreak` when spaces became `.mspace`.  Should be all set now.

test/katex-spec.js
kevinbarabash: Let's leave things as they are now.  Eventually I'd like to replace those tests with snapshot tests.  We still might want to do some simplification depending on what the snapshots look like.

src/buildHTML.js
```diff
@@ -205,7 +205,7 @@ export const getTypeOfDomTree = function(node, side = "right") {
 // 'mtight' indicates that the node is script or scriptscript style.
 export const isLeftTight = function(node) {
     node = getOutermostNode(node, "left");
-    return utils.contains(node.classes, "mtight");
+    return node.hasClass("mtight");
```
kevinbarabash: yay!

src/buildHTML.js
```diff
@@ -629,21 +632,18 @@ export const buildGroup = function(group, options, baseOptions) {
 };
 
 /**
- * Take an entire parse tree, and build it into an appropriate set of HTML
- * nodes.
+ * Combine an array of HTML DOM nodes (e.g., the output of `buildExpression`)
+ * into an unbreakable HTML node of class .base, with proper struts to
+ * guarantee correct vertical extent.  `buildHTML` calls this repeatedly to
+ * make up the entire expression as a sequence of unbreakable units.
```
kevinbarabash: Nice comment.

src/buildHTML.js
```diff
@@ -654,10 +654,60 @@ export default function buildHTML(tree, options) {
     // normal place) so we use an absolute value for vertical-align instead
     bottomStrut.style.verticalAlign = -body.depth + "em";
```
kevinbarabash: I didn't realize that you could pass `verticalAlign` a length.  It makes me wonder why the `topStrut` is even necessary since the `bottomStrut`'s height is already `body.height + body.depth`.

src/buildHTML.js
```diff
@@ -654,10 +654,60 @@ export default function buildHTML(tree, options) {
...
+}
+
+/**
+ * Take an entire parse tree, and build it into an appropriate set of HTML
+ * nodes.
+ */
+export default function buildHTML(tree, options) {
+    // buildExpression is destructive, so we need to make a clone
+    // of the incoming tree so that it isn't accidentally changed
+    tree = JSON.parse(JSON.stringify(tree));
```
kevinbarabash: I didn't realize this.  I wonder how we're mutating `tree`.  🤔

src/buildHTML.js
```diff
@@ -654,10 +654,60 @@ export default function buildHTML(tree, options) {
...
+            expression[i].hasClass("allowbreak")) {
+            // Put any post-operator glue on same line as operator.
+            // Watch for \nobreak along the way.
+            let nobreak = false;
+            while (i < expression.length - 1 &&
+                   expression[i + 1].hasClass("mspace")) {
+                i++;
+                parts.push(expression[i]);
+                if (expression[i].hasClass("nobreak")) {
+                    nobreak = true;
```
kevinbarabash: This is a lot cleaner.

kevinbarabash:
Is anyone able able to view side-by-side images in the files tab on GitHub?  All I see is `Binary file not shown.`.  It's not just this diff either. :(

kevinbarabash:
I had a look at the screenshots.  They look good.

edemaine:
@kevinbarabash Check out https://github.com/Khan/KaTeX/pull/1267#issuecomment-387722597 . I was playing with the various "rich diff" options and they're pretty cool! You can even overlay the two imaged in various ways (though not as good as our red/green diffs).

Thanks for the review! I excited to use this feature.

src/buildHTML.js
edemaine: I'm curious too! I haven't seen any, but this comment has been in the code for a long time.

src/buildHTML.js
edemaine: Hmm, good question! I should experiment with removing the topStrut.

Issue #1288, opened 2018-05-05
vertical layout of \underset and \overset


pzinn:
Thanks for fixing #1275. This may be nitpicking, but I'm still not quite happy with \underset and \overset.
Compare the output of
```
\overset{a}{b}b\underset{a}{b}b
```
in LaTeX and KaTeX. In LaTeX, all b's are at exactly the same height, whereas in KaTeX the b's wobble.
Is this a known issue/limitation? Only reference to underset/overset I found is #1081 which I think is something different.
I understand that's not top priority, but I thought I'd mention.

edemaine:
KaTeX:
![image](https://user-images.githubusercontent.com/2218736/39663825-ff4f5552-5047-11e8-8f45-2bde7eeeea57.png)

LaTeX (via quicklatex):

![image](https://user-images.githubusercontent.com/2218736/39663828-13c18960-5048-11e8-8be6-851025491f93.png)

The wiggles definitely seem wrong, but so does the horizontal spacing between elements? Ideas, @ronkok?

ronkok:
My first guess is that the base element is being shifted from the baseline to a point where it is centered on the math axis. And my second guess is that we can fix that by changing line 75 in `op.js` from 
```
if (base instanceof domTree.symbolNode) {
```
to 
```
if (group.value.symbol) {
```

In fact, that line actually did read at one time exactly the way I am proposing. Some one had to have changed it and I don't know why. So before I charge in and change it back, I want to think a bit about why they may have done that. Give me a day or two.

edemaine:
@ronkok I traced that particular line back, and it's been there for at least 3 years. Maybe there was another place where you changed it which you're thinking of?

ronkok:
Maybe it was a different line that changed. I still think that line is the most likely way to fix this problem. I'll do some testing.

ronkok:
Fixed by #1290. Thanks for the bug report!

PR #1290, opened 2018-05-05
Adjust \underset alignment and spacing

ronkok:
Fixes issue #1288.

1. Fixes vertical alignment by keeping the base element on the baseline.
2. Assigns horizontal spacing to \underset in a manner that aligns with LaTeX `\binrel`.


ronkok:
Results:
adjustunderset

ronkok:
I'll be curious to look at all the other screenshotter tests. The change I made in `op.js` should not change anything but `\underset`. We'll see if that's true.

edemaine:
I updated the `OverUnderset` screenshot.  Thanks for adding the wiggle test to avoid this breaking in the future!

ronkok:
@edemaine The screenshot test looks good to me. Thank you.

src/functions.js
```diff
@@ -66,26 +66,39 @@ defineFunction([
...
-    const xAxisOp = new ParseNode("op", {
+    let mclass = "mrel";
+    if (context.funcName !== "\\stackrel") {
+        // LaTeX applies \binrel spacing to \overset and \underset.
+        // \binrel spacing varies with (bin|rel|ord) of the atom in the argument.
+        // We'll do the same.
+        if (baseArg.type === "ordgroup") {
+            mclass = "m" + baseArg.value[0].type.replace("math", "");
+        } else {
+            mclass = "m" + baseArg.type.replace("math", "");
```
edemaine: Is `baseArg.type` / `baseArg.value[0].type` guaranteed to be one of `mathbin`/`mathrel`,`mathord`?  What if there's an `ordgroup` in an `ordgroup`, for example?...  I'm not suggesting going any deeper, but I worry that we need additional checking here.

ronkok: > What if there's an `ordgroup` in an `ordgroup`?

Then the result will have a CSS class `mordgroup`, that is, a class with no style rules attached. Clearly that is not ideal, but I didn't really want to do a recursive search.

In the current KaTeX release, `\overset` and `\underset` both assign `\mathop` spacing regardless of the argument type. Would it be okay if I searched for an a (bin|rel|ord) type in the top two levels, use it if found, but use `\mathop` for a default if the `ordgroup`s go deeper?

ronkok: @edemaine I just made a commit that tries to resolve nested `ordgroups`. Thank you for the nudge. What I had before was sloppy.

Mind you, what I have now does seem particularly clean, but nothing better came to mind.

ronkok:
Well, the arguments passed to this function are ParseNodes, not DomNodes, so `getTypeOfDomTree` is not helpful without a lot of pre-processing. That seems kind of wasteful. Let me suggest that the base of a `\overset` can fall into three categories:

| Category                           | Example                           | We want mclass to be: |
|:----------------------------|:----------------------------|:--------------------------|
| single symbol                    | \\overset ? =                   | baseArg.type              |
| group containing a single symbol | \\overset {?}{=}   | baseArg\[0].type          |
| everything else                  | \\overset ? {{a+b} \\div c} | `mord`                     |

If that's true, then we can get the needed information with:
```
		let atomType = "";
		if (baseArg.type === "ordgroup") {
			atomType = baseArg.value[0].type;
		} else {
			atomType = baseArg.type;
		}
		if (/bin|rel/.test(atomType)) {
			mclass = "m" + atomType;
		} else {
			mclass = "mord";
		}
```

Yes, what I've written is a bit of hack, but after looking at that LaTeX code, I don't feel too bad about that. And this code stays local and fast, without excursions into `buildHTML`. 

What do you think?

edemaine:
@ronkok This seems almost perfect.  However, I believe there's some confusion about the cases, because `\overset{?}{=}` and `\overset ? =` should parse identically (and do -- I just checked).  So the arguments are *always* wrapped in `ordgroup`.

That said, I think the code you posted is exactly what we want.  This way we can handle a non-ordgroup value, just in case.  We don't want to go any deeper, though, so that `\overset{{?}}{{=}}` behaves as it does in LaTeX (I think) -- treating the arguments like an ord group.

ronkok:
> some confusion about the cases

> the code you posted is exactly what we want.

Better lucky than good, eh?

I'll make the change and commit.

edemaine:
Hmm, it occurs to me that, because this code is looking at the parse tree, it won't properly handle other parse tree nodes that end up acting like a `bin` or `rel`.  For example, an `\overset` within an `\overset` argument won't behave correctly with this code, but ought to with LaTeX.

One alternative would be to wait until HTML/MathML build time, and use the computed class.  This would be closer to LaTeX's behavior.  I'm not sure whether it's worth the added complexity... maybe at least a comment?

ronkok:
That's reasonable. I added a comment.

edemaine:
I'm good with this. @kevinbarabash did you want to look at it, or should we just merge?

kevinbarabash:
@ronkok thanks for fixing this.  @edemaine thanks for reviewing.   The changes look reasonable.  Let's merge it!

src/functions.js
```diff
@@ -70,26 +70,48 @@ defineFunction("mclass", [
...
+    const baseOp = new ParseNode("op", {
         type: "op",
         limits: true,
         alwaysHandleSupSub: true,
         symbol: false,
-        value: ordargument(mathAxisArg),
-    }, mathAxisArg.mode);
+        suppressBaseShift: context.funcName !== "\\stackrel",
+        value: ordargument(baseArg),
+    }, baseArg.mode);
```
ylemkimon: `ParseNode`s are flow-typed by #1276:
https://github.com/Khan/KaTeX/blob/5a4aedd8827776acd737c993b9416c36302465f7/src/ParseNode.js#L66-L73

ronkok: @ylemkimon Thank you for adding `suppressBaseShift` to the flow type enumeration. I was trying to think of a way to make `\underset` work without doing that. I was getting nowhere.

Issue #1080, opened 2018-01-21
bounding box is wrong for \cancel groups


kevinbarabash:
![strikethrough](https://user-images.githubusercontent.com/1044413/35199882-c294528c-fed2-11e7-85ae-0451a5f93c89.png)

I added
```
\usepackage{cancel}
\usepackage{ulem}
```
to test.tex in order to generate the image above.  The spacing around binary operators is better than it used to be.  It seems like the addition of the `\cancel` bar doesn't affect the bounding box of `\cancel` in the LaTeX rendering (green).  The bounding box of `\cancel` groups is essentially the same as the bounding box of the contents.  This is probably most noticeably in `{\cancel B}_1^2`.

ronkok:
@kevinbarabash The `cancel` package documentation states that:

> By default, none of these commands affects the horizontal spacing, so they might over-
print  neighboring  parts  of  the  formula  (or  text).

I tried to accomplish that with CSS adjacency rules. It looks like I missed some of the possible combinations. I don't know if PR #1070 makes this easier or harder.

kevinbarabash:
> I don't know if PR #1070 makes this easier or harder.

Hopefully easier.  🤞

kevinbarabash:
I'm not sure why pdflatex rendered `\sout` as underlines.

edemaine:
For reference, here's a texdiff on the #1295 branch.  I'm not convinced it's much better... but hopefully easier to fix?

![strikethrough](https://user-images.githubusercontent.com/2218736/39774128-f8f3858e-52c7-11e8-882d-32b472535294.png)

kevinbarabash:
It's looks like there are two things going on:
- LaTeX's `\cancel`s are generally steeper, especially for single lowercase letters
- LaTeX has less spacing around the `\cancel`led element.

PR #1295, opened 2018-05-08
Switch makeGlue from .mord .rule to .mspace

### LINKED ISSUE #1289: Why does makeGlue use .mord .rule instead of .mspace?
`makeGlue`, which is what `\hskip` uses to build a space in HTML, looks like this:

```js
const makeGlue = (measurement: Measurement, options: Options): DomSpan => {
    // Make an empty span for the rule
    const rule = makeSpan(["mord", "rule"], [], options);
    const size = calculateSize(measurement, options);
    rule.style.marginRight = `${size}em`;
    return rule;
};
```

Does anyone know why this space has classes `.mord .rule` instead of the more obvious `.mspace`?

For example, `buildHTML`'s `buildExpression` looks at the first class of groups to see whether it is `mspace`, to ignore spaces.  It seems like `\hskip`s ought to be treated as spaces, but they aren't currently?...

edemaine:
Fix #1289 

I haven't run screenshot tests yet; I'm curious to see whether this messes with anything.


edemaine:
Amazingly, only two screenshots were affected.  Here's the diff:

StrikeThrough:
strikethrough-chrome-diff

StrikeThroughColor:
strikethroughcolor-chrome-diff

The result of this PR is in green/yellow, with more spaces.  Somewhat easier to see in the [Github diff](https://github.com/Khan/KaTeX/pull/1295/files?short_path=4f6a6b3#diff-4f6a6b3baa51f7f57c8d0ded46b28fc7).  I think this shows that the change fixes some spacing bugs, properly treating the `+` as a binary operator.  One example of the LaTeX source code is `\bcancel 5 +\bcancel{5ay}`.

kevinbarabash:
@edemaine could you post diff images between the new render and the LaTeX rendering?

edemaine:
I posted it on #1080 where you can compare to the old diff.  I'm afraid it's not much better... but hopefully a step in the right direction.

strikethrough

PR #1298, opened 2018-05-08
Support for top-level \newline and \\ in inline math

### LINKED ISSUE #171: Add support for line breaks
Add support for line breaks. `\\`, `\newline`...

Thanks.

edemaine:
Built on top of #1287, so blocked on that.  Fix #171.

This was a little tricky because `\\` was defined as an `endOfExpression`.  Instead made `\\` a termination specific to an array environment.

Outside an array environment, buildHTML handles the `cr` object, resulting in a `.newline` class.

Currently this turns into a `display: block` (with appropriate vertical spacing) only in inline math, matching LaTeX behavior.  But we should consider always turning into a `display: block`, maybe only at the top level, at least in nonstrict mode...

Demo:

image

With optional argument:

image


kevinbarabash:
@edemaine consider pushing your branches to `Khan/KaTeX`, then you could make this PR based on `inlinebreak` instead of `master` and it'd be easier to see the changes that you added in this PR.

edemaine:
Sorry about that (and probably a good idea for this sort of thing). I didn't expect anyone to look at it until #1287 lands, but meanwhile, you can check out the diff here:

https://github.com/edemaine/KaTeX/compare/inlinebreak...edemaine:newline

edemaine:
I resolved the conflicts, fixed a bug found via the new tests, and added tests.

One issue I wonder about: do we want `\\` to do anything (at the top level) in display mode?  It doesn't do anything in LaTeX (acts as a null space), and that's what I've implemented here.  But I imagine a lot of people might want it, in nonstrict mode at least.  The tricky part is that the two behaviors require different CSS...

One solution: I've added documentation to the README to document both this difference and how to tweak it for those who want non-LaTeX behavior.

src/functions/cr.js
```diff
@@ -0,0 +1,56 @@
...
+        numOptionalArgs: 1,
+        argTypes: ["size"],
+        allowedInText: true,
+    },
+
+    handler: (context, args, optArgs) => {
+        return {
+            type: "cr",
+            // \\ and \cr both end the row in a tabular environment
+            newRow: context.funcName !== "\\newline",
```
ylemkimon: Where is `newRow` used?

edemaine: It's never used. I expected to want it in `environments/array.js`, but that explicitly checks for `\\` or `\cr` without thinking of it as a function, so this isn't needed there.  I could remove it.  Preference?

ylemkimon: I think we can comment it out, as it can be used in the future.

edemaine: Sounds good. Done.

kevinbarabash:
@edemaine I think it would be nice to have `\\` split lines in non-strict mode in display mode.  Thanks for adding a note in the README about how to enable this behavior.

kevinbarabash:
It's weird that the travis build failed but circleci was okay.  It looks like there's an issue with the flow types in cr.js.

kevinbarabash:
I see, circleci isn't running the tests yet.

ylemkimon:
@kevinbarabash CircleCI currently doesn't run tests, including `flow`, but it will added by PR #1211.

edemaine:
I fixed the flow error (caused by commenting out `newRow`). Should be ready to merge now.

src/functions/cr.js
```diff
@@ -0,0 +1,57 @@
...
+    },
+
+    handler: (context, args, optArgs) => {
+        return {
+            type: "cr",
+            // \\ and \cr both end the row in a tabular environment
+            // This flag isn't currently needed by environments/array.js
+            //newRow: context.funcName !== "\\newline",
+            // \\ and \newline both end the line in an inline math environment
+            newLine: context.funcName !== "\\cr",
```
ylemkimon: I think we can do something like:
```
            newLine: context.funcName !== "\\cr" && (!context.parser.settings.displayMode || nonStrict),
```

edemaine: In LaTeX, `\\` "works" (does nothing) in display mode, same as the current code.  Your change would cause an error in strict mode.  But that's true, we could drop the "cr" class in strict mode; good idea.

edemaine:
This is an interesting test of the `strict` setting... `"error"` might cause an error when using `\\` in display mode which isn't *terrible*, but LaTeX just silently ignores the `\\`...  Do we want to add "fall-back" capabilities to the `nonstrict` method?...

kevinbarabash:
The intent of `strict` mode is twofold: 
- complain when people use things that aren't supported in LaTeX, e.g. unicode in math mode
- implement the same behavior as LaTeX

Since we're doing the latter already in this case there's no need to warn or throw and error.

kevinbarabash:
It should still do nothing, b/c it's valid LaTeX to use `//` in display mode.  If `//` caused an error in LaTeX then it should also cause an error in KaTeX in strict.  But since it doesn't, we shouldn't raise an error either.

edemaine:
I agree, no error ever, but I'm a little confused about when the `\\` should do nothing and when it should break the line.

kevinbarabash:
My understanding is that `\\` should always break lines when not in display mode.  In display mode it should only break lines if `strict` is set to `false` or `ignore`.  `strict: "warn"` and `strict: "error"` seem like two variants of strict mode so in either of those situations `\\` shouldn't do anything in display mode.

edemaine:
Change of plans. I'm going to merge this as is, then play with strict behavior in #1278.

PR #1304, opened 2018-05-10
Fix underbrace

### LINKED ISSUE #1303: \underbrace vertical offset when subscript has depth
Correct behavior: `x=\underbrace{\frac{x}{y}}_h` (frac line is aligned with base line)

image

Incorrect behavior: `x=\underbrace{\frac{x}{y}}_g` (frac line is above base line)

image

@ronkok, any ideas why this might be happening?

ronkok:
Fix the vertical alignment of `\underbrace` so that it correctly takes into account the descender of the lower element.

Fixes #1303.


ronkok:
Visual confirmation:
underbrace

ronkok:
I'll need some help to generate new screenshots for HorizontalBraces. The bug is visible right there in the old screenshot.

edemaine:
That was fast! Screenshots pushed.

PR #1305, opened 2018-05-10
Cleanup domTree.js to re-use code

edemaine:
* Refactor `span` and `anchor` into common `HtmlDomContainer` superclass, without changing internal data structure (e.g. not storing `tagName`).
* Makes it easy to add more similar tags
* Number of lines falls from 676 to 587

My motivation for doing this is that I'd like to try reimplementing `array` in terms of `table`, `tr`, `td`, which requires several new tags similar to `span` but with a different `tagName`.  This will make that work much easier.


kevinbarabash:
@edemaine it's like you read my mind.  I was thinking of this exact thing when reading through #1287.

ylemkimon:
I should've checked how much polyfill this generates. I think this PR need to be reverted.

edemaine:
@ylemkimon Is it a lot? Maybe we can avoid it somehow?

ylemkimon:
@edemaine It it self is about 50KB. I think we should avoid using class inheritance until our target browsers natively support it. I've created a PR to use helper functions in #1469.

kevinbarabash:
babel has a `loose` mode setting for some transforms.  We could also investigate that.

ylemkimon:
@kevinbarabash #1469 enables the loose mode. However, Babel's loose inheritance is still expensive.

PR #1306, opened 2018-05-11
Support \hline

### LINKED ISSUE #267: Add horizontal and vertical line support in environments

ronkok:
Support `\hline`. Fix #267, part of #269.


ronkok:
Here's how they look:
hlinepic

edemaine:
@ylemkimon I don't think it's actually a problem.  @ronkok's last few PRs have been this way, and the old commits get swept up in the squash.  At some point it'd be good to start new PRs from the master branch, though.  (For me it's `git checkout upstream/master; git checkout -b newbranch`.)

edemaine:
@ronkok Wow, it's exciting to see how this is relatively easy thanks to stretchy `\overline`!

There are two natural features that aren't yet in this PR; is it worth adding them now?
* `\hline` can be repeated to make multiple horizontal lines between rows. (I think it also interrupts the vertical lines... that may be harder to do, but could be skipped for now.) This would be good for "full `\hline` support".
* `\cline{3-5}` does `\hline` only for columns 3 through 5. I'm guessing you have to render the horizontal line for each column separately anyway (because tables are rendered column-by-column), so this isn't too hard...  But also not as important as `\hline` itself.

ronkok:
@ylemkimon GitH(Hub) has been acting that way since I did manual edit on a merge conflict in the `\ll` PR. As soon as this PR and PR #1162 are merged, I'll dive in and fix my repository. Until then, I don't want to make any rash moves.

ronkok:
@edemaine Later today, I'll add the double `\hline`. As you suggest, today's effort will not interrupt any vertical lines.  I may also add a dotted `\hline`.

No promises on `\cline`. My currrent `\hline`s are *not* done column by column. I just let the browser stretch each one the full width of the array. In order to implement `\cline`, I would have to work my way deeper into the array parser. That may not happen in time for the next release.

src/environments/array.js
```diff
@@ -76,6 +93,19 @@ function parseArray(
...
             }
             rowGaps.push(cr.value.size);
+
+            possibleHLine = parser.nextToken.text;
+            while (/\s/.test(possibleHLine)) {
+                parser.consume();
+                possibleHLine = parser.nextToken.text;
+            }
+            if (possibleHLine === "\\hline") {
+                hlines.push(true);
```
ylemkimon: Can `group.value.addHLines = true;` be done here?

ronkok: @ylemkimon Stand by. I've got a code cleanup commit coming in about an hour. It will change this section of the code.

src/environments/array.js
```diff
@@ -76,6 +93,19 @@ function parseArray(
                 throw new ParseError(`Failed to parse function after ${next}`);
             }
             rowGaps.push(cr.value.size);
+
+            possibleHLine = parser.nextToken.text;
+            while (/\s/.test(possibleHLine)) {
+                parser.consume();
+                possibleHLine = parser.nextToken.text;
+            }
```
ylemkimon: Can this be done by `parser.consumeSpaces()`? (same for above)

edemaine:
The screenshot renders as 1246x752, but needs to be at most 1024x768.  Can you make it a little less wide, e.g., reducing the `\quad`s?

ronkok:
There's no point in cutting it close. One of the arrays was redundant and I took it out.

src/environments/array.js
```diff
@@ -30,8 +30,22 @@ export type ArrayEnvNodeData = {
...
     rowGaps?: number[],
+    numHLinesBeforeRow?: number[],
 };
 
+function nextTokenIsHLine(parser: Parser): boolean {
+    let next = parser.nextToken.text;
+    while (/^\s$/.test(next)) {
+        parser.consume();
+        next = parser.nextToken.text;
+    }
```
ylemkimon: Could this be done using `parser.consumeSpaces()`?

src/environments/array.js
```diff
@@ -46,6 +60,18 @@ function parseArray(
...
     const rowGaps = [];
+    const numHLinesBeforeRow = [];
+
+    // Test for \hline at the top of the array.
+    if (nextTokenIsHLine(parser)) {
+        numHLinesBeforeRow.push(1);
+        if (nextTokenIsHLine(parser)) {
+            numHLinesBeforeRow[0] = 2;
+        }
+    } else {
```
ylemkimon: I think the following is more readable (same for below), but it's matter of personal preference:
```js
    if (nextTokenIsHLine(parser)) {
        if (nextTokenIsHLine(parser)) {
            numHLinesBeforeRow.push(2);
        } else {
            numHLinesBeforeRow.push(1);
        }
    } else {
```

src/environments/array.js
edemaine: Yes. Please use that instead.

src/environments/array.js
edemaine: Is there something magic about 2? Why not a `while` loop and counter? Does LaTeX not support more than 2 `\hline`s?

src/environments/array.js
ronkok: Will do.

src/environments/array.js
ronkok: Until just now, I had never tried more than two `\hline`s. `while` loop coming right up.

ronkok:
And now, with multiple `\hline`s:
hlinepic

src/environments/array.js
```diff
@@ -130,6 +157,16 @@ const htmlBuilder = function(group, options) {
...
     let totalHeight = 0;
+
+    // Set a position for \hline(s) at the top of the array, if any.
+    if (numHLinesBeforeRow[0] > 0) {
+        hlinePos.push(0);
+    }
+    for (let i = 2; i <= numHLinesBeforeRow[0]; i++) {
+        totalHeight += 0.25;
+        hlinePos.push(totalHeight);
+    }
```
ylemkimon: This can be simplified to (same for below):
```js
    if (numHLinesBeforeRow[0] > 0) {
        totalHeight += 0.25 * (numHLinesBeforeRow[0] - 1);
        hlinePos.push(totalHeight);
    }
```

src/environments/array.js
```diff
@@ -30,8 +30,18 @@ export type ArrayEnvNodeData = {
...
 };
 
+function nextTokenIsHLine(parser: Parser): boolean {
+    parser.consumeSpaces();
+    if (parser.nextToken.text === "\\hline") {
+        parser.consume();
+        return true;
+    }
+    return false;
+}
```
ylemkimon: How about:
```js
function getNumHLines(parser: Parser): number {
    let n = 0;
    parser.consumeSpaces();
    while (parser.nextToken.text === "\\hline") {
        parser.consume();
        n++;
        parser.consumeSpaces();
    }
    return n;
}
```
(or `for` loop if you prefer) and

src/environments/array.js
```diff
@@ -46,6 +56,13 @@ function parseArray(
     let row = [];
     const body = [row];
     const rowGaps = [];
+    const numHLinesBeforeRow = [0];
+
+    // Test for \hline at the top of the array.
+    while (nextTokenIsHLine(parser)) {
+        numHLinesBeforeRow[0] += 1;
+    }
```
ylemkimon:
```js
    const numHLinesBeforeRow = [];

    numHLinesBeforeRow.push(getNumHLines(parser));
```
(same for below)?

src/environments/array.js
ronkok: @ylemkimon We need a loop in order to handle multiple `\hline`s. So I think you mean that the code could read something like:
```
    for (let i = 1; i <= numHLinesBeforeRow[0]; i++) {
        if (i >  1) {
            totalHeight += 0.25;
        }
        hlinePos.push(totalHeight);
    }
```

That would be simpler for a human to read. But force of habit makes me try to avoid putting conditional statements inside a loop. Maybe this is not the place for that kind of thing. After all, we're not likely to get very many `\hline`s. 

Would you prefer my suggested revision to the current code?

src/environments/array.js
ronkok: I like it.

src/environments/array.js
ronkok: That's an outstanding suggestion. I'll do it.

src/environments/array.js
ylemkimon: @ronkok I thought `hlinePos` is added only once, sorry. I think we can leave as-is.

ronkok: > I thought `hlinePos` is added only once

Which goes to show that code I suggested above would indeed be easier for a human to read. The speed loss with so few `\hline`s is negligible. I'll make the change.

ronkok:
Once again, I'll need help to generate a screenshot for HLine. Sorry to be a burden.

ronkok:
Thanks, @edemaine!

PR #1309, opened 2018-05-12
Support \tag and \gdef

### LINKED ISSUE #899: support \tag
Suggested as a precursor to full equation numbering support in #350, it has AFAICT the following behavior:
- add a `\tag` command will force the math to typeset in display mode but respects the current style
- only one `\tag` command is allowed per math block
- the content of the `\tag` is typeset in `text` mode (not sure if entering math mode is allowed or not, my assumption would be not)
- the content of the `\tag` is right aligned and typset within parens, e.g. `\tag{3.1c}` yields `(3.1c)`

edemaine:
The primary goal here was to implement `\tag` and thereby fix #899.  As discussed in that issue, I've tried a variety of ways to lay out the tag.  My recent experience with struts opened up some new possibilities, though, and I found a nice layout that (a) centers the equation as before and (b) properly aligns the tag.  The solution is to `position: absolute; right: 0px` the tag, but add a strut matching the equation to make it vertically aligned correctly. However, it's possible for the tag to overlap the equation; no floating behavior.  (I don't think it's possible to get the proper centering and have float behavior, without client-side JavaScript.)  But it was always possible for the equation to be too big for the line anyway.

`\tag` has a delayed effect (the argument gets rendered at the end of the display formula), which is implemented in LaTeX via the macro system.  So I implemented some basic TeX-side macros (as requested in #250): you can now define a macro via `\gdef\foo{...}` or `\gdef\foo#1#2{...}`.  It's a bit restrictive, but essentially opens up the existing macro functionality from the TeX input side, which is cool in its own right.  This is a good step toward `\def` and `\setlength` -- I have a plan for those.

Demos:

image

image

### DIFF
--- a/src/MacroExpander.js
+++ b/src/MacroExpander.js
@@ -9,7 +9,6 @@ import {Token} from "./Token";
 import builtinMacros from "./macros";
 import type {Mode} from "./types";
 import ParseError from "./ParseError";
-import objectAssign from "object-assign";
 
 import type {MacroContextInterface, MacroMap, MacroExpansion} from "./macros";
 
@@ -21,7 +20,7 @@ export default class MacroExpander implements MacroContextInterface {
 
     constructor(input: string, macros: MacroMap, mode: Mode) {
         this.lexer = new Lexer(input);
-        this.macros = objectAssign({}, builtinMacros, macros);
+        this.macros = Object.assign({}, builtinMacros, macros);
         this.mode = mode;
         this.stack = []; // contains tokens in REVERSE order
     }
--- a/src/buildHTML.js
+++ b/src/buildHTML.js
@@ -668,8 +668,7 @@ export default function buildHTML(tree, options) {
     // Build the expression contained in the tree
     const expression = buildExpression(tree, options, true);
 
-    const htmlNode = makeSpan(["katex-html"], []);
-    htmlNode.setAttribute("aria-hidden", "true");
+    const children = [];
 
     // Create one base node for each chunk between potential line breaks.
     // The TeXBook [p.173] says "A formula will be broken only after a
@@ -697,14 +696,25 @@ export default function buildHTML(tree, options) {
             }
             // Don't allow break if \nobreak among the post-operator glue.
             if (!nobreak) {
-                htmlNode.children.push(buildHTMLUnbreakable(parts, options));
+                children.push(buildHTMLUnbreakable(parts, options));
                 parts = [];
             }
         }
     }
     if (parts.length > 0) {
-        htmlNode.children.push(buildHTMLUnbreakable(parts, options));
+        children.push(buildHTMLUnbreakable(parts, options));
     }
+
+    const htmlNode = makeSpan(["katex-html"], children);
+    htmlNode.setAttribute("aria-hidden", "true");
+
+    const strut = makeSpan(["strut"]);
+    strut.style.height = (htmlNode.height + htmlNode.depth) + "em";
+    strut.style.verticalAlign = (-htmlNode.depth) + "em";
+    htmlNode.children.push(makeSpan(["tag"], [
+        strut,
+        buildCommon.mathsym("*", "math"),
+    ]));
 
     return htmlNode;
 }
--- a/src/buildHTML.js
+++ b/src/buildHTML.js
@@ -708,13 +708,24 @@ export default function buildHTML(tree, options) {
     const htmlNode = makeSpan(["katex-html"], children);
     htmlNode.setAttribute("aria-hidden", "true");
 
-    const strut = makeSpan(["strut"]);
-    strut.style.height = (htmlNode.height + htmlNode.depth) + "em";
-    strut.style.verticalAlign = (-htmlNode.depth) + "em";
-    htmlNode.children.push(makeSpan(["tag"], [
-        strut,
-        buildCommon.mathsym("*", "math"),
-    ]));
+    // \tag output is guaranteed to be the last node of the last unbreakable.
+    // Bring it out to the top level.
+    if (children && children.length > 0) {
+        const lastChild = children[children.length - 1];
+        if (lastChild && lastChild.children.length > 0) {
+            const lastNode = lastChild.children[lastChild.children.length - 1];
+            if (lastNode.hasClass("tag")) {
+                lastChild.children.pop();
+                const strut = makeSpan(["strut"]);
+                strut.style.height = (htmlNode.height + htmlNode.depth) + "em";
+                strut.style.verticalAlign = (-htmlNode.depth) + "em";
+                htmlNode.children.push(makeSpan(["tag"], [
+                    strut,
+                    lastNode,
+                ]));
+            }
+        }
+    }
 
     return htmlNode;
 }
--- a/src/macros.js
+++ b/src/macros.js
@@ -118,7 +118,7 @@ defineMacro("\\gdef", function(context) {
     while (arg.length === 1 && arg[0].text === "#") {
         arg = context.consumeArgs(1)[0];
         if (arg.length !== 1) {
-            throw new ParseError(`Invalid argument number "${arg}"`);
+            throw new ParseError(`Invalid argument number length "${arg.length}"`);
         }
         if (!(/^[1-9]$/.test(arg[0].text))) {
             throw new ParseError(`Invalid argument number "${arg[0].text}"`);
--- a/src/macros.js
+++ b/src/macros.js
@@ -396,6 +396,17 @@ defineMacro("\\thinspace", "\\,");    //   \let\thinspace\,
 defineMacro("\\medspace", "\\:");     //   \let\medspace\:
 defineMacro("\\thickspace", "\\;");   //   \let\thickspace\;
 
+// \tag@in@display form of \tag
+defineMacro("\\tag", "\\@ifstar\\tag@literal\\tag@paren");
+defineMacro("\\tag@paren", "\\tag@literal{(#1)}");
+defineMacro("\\tag@literal", (context) => {
+    if (context.macros["\\df@tag"] && context.macros["\\df@tag"].tokens &&
+        context.macros["\\df@tag"].tokens.length > 0) {
+        throw new ParseError("Multiple \\tag");
+    }
+    return "\\gdef\\df@tag{\\tag@{#1}}";
+});
+
 //////////////////////////////////////////////////////////////////////
 // LaTeX source2e
 
--- a/src/macros.js
+++ b/src/macros.js
@@ -8,12 +8,18 @@ import fontMetricsData from "../submodules/katex-fonts/fontMetricsData";
 import symbols from "./symbols";
 import utils from "./utils";
 import {Token} from "./Token";
+import ParseError from "./ParseError";
 
 /**
  * Provides context to macros defined by functions. Implemented by
  * MacroExpander.
  */
 export interface MacroContextInterface {
+    /**
+     * Object mapping macros to their expansions.
+     */
+    macros: MacroMap;
+
     /**
      * Returns the topmost token on the stack, without expanding it.
      * Similar in behavior to TeX's `\futurelet`.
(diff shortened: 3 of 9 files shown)


ylemkimon:
The screenshots seem to be empty.

edemaine:
@ylemkimon Oops, thanks for catching that.  There was a typo in the screenshot test.

src/macros.js
```diff
@@ -95,6 +101,42 @@ defineMacro("\\TextOrMath", function(context) {
...
+        numArgs++;
+        if (parseInt(arg[0].text) !== numArgs) {
+            throw new ParseError(`Argument number "${arg[0].text}" out of order`);
+        }
+        arg = context.consumeArgs(1)[0];
+    }
+    // Final arg is the expansion of the macro
+    context.macros[name] = {
+        tokens: arg,
+        numArgs,
```
kevinbarabash: I forgot that each `arg` is an array of tokens.  This makes sense.

src/macros.js
```diff
@@ -95,6 +101,42 @@ defineMacro("\\TextOrMath", function(context) {
...
+    let numArgs = 0;
+    arg = context.consumeArgs(1)[0];
+    while (arg.length === 1 && arg[0].text === "#") {
+        arg = context.consumeArgs(1)[0];
+        if (arg.length !== 1) {
+            throw new ParseError(`Invalid argument number length "${arg.length}"`);
+        }
+        if (!(/^[1-9]$/.test(arg[0].text))) {
+            throw new ParseError(`Invalid argument number "${arg[0].text}"`);
+        }
```
kevinbarabash: Is LaTeX limited in the number of args it can have?

src/macros.js
```diff
@@ -95,6 +101,42 @@ defineMacro("\\TextOrMath", function(context) {
...
+        numArgs++;
+        if (parseInt(arg[0].text) !== numArgs) {
+            throw new ParseError(`Argument number "${arg[0].text}" out of order`);
+        }
```
kevinbarabash: Even though this check is sufficient, having both allows for more nuanced error report.  Nice!

test/__snapshots__/mathml-spec.js.snap
```diff
@@ -505,3 +505,42 @@ exports[`A MathML builder should use <munderover> for large operators 1`] = `
...
+                2
+              </mn>
+            </msup>
+          </mrow>
+        </mtd>
+      </mlabeledtr>
+    </mtable>
+    <annotation encoding="application/x-tex">
+      \\tag{hi} x+y^2
+    </annotation>
```
kevinbarabash: Cool!

test/katex-spec.js
```diff
@@ -2898,6 +2898,24 @@ describe("A macro expander", function() {
...
+        compareParseTree("\\gdef{\\foo}{x^2}\\foo+\\foo", "x^2+x^2");
+        compareParseTree("\\gdef\\foo{hi}\\foo+\\text{\\foo}", "hi+\\text{hi}");
+        compareParseTree("\\gdef\\foo#1{hi #1}\\text{\\foo{Alice}, \\foo{Bob}}",
+            "\\text{hi Alice, hi Bob}");
+        compareParseTree("\\gdef\\foo#1#2{(#1,#2)}\\foo 1 2+\\foo 3 4",
+            "(1,2)+(3,4)");
+        expect("\\gdef\\foo#2{}").toNotParse();
+        expect("\\gdef\\foo#1#3{}").toNotParse();
+        expect("\\gdef\\foo#1#2#3#4#5#6#7#8#9{}").toParse();
+        expect("\\gdef\\foo#1#2#3#4#5#6#7#8#9#10{}").toNotParse();
```
kevinbarabash: Too bad it doesn't go to `#11`. ;)

test/katex-spec.js
```diff
@@ -2898,6 +2898,24 @@ describe("A macro expander", function() {
...
+        expect("\\gdef\\foo#{}").toNotParse();
```
kevinbarabash: It would be cool if the matcher took an optional param to verify the error message.

test/katex-spec.js
```diff
@@ -2898,6 +2898,24 @@ describe("A macro expander", function() {
...
     //});
 
+    it("\\gdef defines macros", function() {
+        compareParseTree("\\gdef\\foo{x^2}\\foo+\\foo", "x^2+x^2");
+        compareParseTree("\\gdef{\\foo}{x^2}\\foo+\\foo", "x^2+x^2");
+        compareParseTree("\\gdef\\foo{hi}\\foo+\\text{\\foo}", "hi+\\text{hi}");
+        compareParseTree("\\gdef\\foo#1{hi #1}\\text{\\foo{Alice}, \\foo{Bob}}",
+            "\\text{hi Alice, hi Bob}");
+        compareParseTree("\\gdef\\foo#1#2{(#1,#2)}\\foo 1 2+\\foo 3 4",
+            "(1,2)+(3,4)");
```
kevinbarabash: fancy

test/katex-spec.js
```diff
@@ -2898,6 +2898,24 @@ describe("A macro expander", function() {
...
+        expect("\\gdef\\foo#2{}").toNotParse();
+        expect("\\gdef\\foo#1#3{}").toNotParse();
+        expect("\\gdef\\foo#1#2#3#4#5#6#7#8#9{}").toParse();
+        expect("\\gdef\\foo#1#2#3#4#5#6#7#8#9#10{}").toNotParse();
+        expect("\\gdef\\foo#{}").toNotParse();
+        expect("\\gdef\\foo\\bar").toParse();
+        expect("\\gdef{\\foo\\bar}{}").toNotParse();
+        expect("\\gdef{}{}").toNotParse();
+    });
```
kevinbarabash: Would the following parse?
```
\gdef{\foo#1#2{#1 + #2}}
\gdef{\bar#1#2{#1 - #2}}
\gdef{\foobar#1#2{(\foo{#1}{#2})(\bar{#1}{#2})}}
\foobar{x}{y}
```
I would expect the result to be:
```
(x + y)(x - y)
```

test/katex-spec.js
```diff
@@ -2914,6 +2932,16 @@ describe("A macro expander", function() {
         expect("\\liminf")
             .toParseLike("\\mathop{\\operatorname{lim\\,inf}}\\limits");
     });
+
+    it("\\tag support", () => {
+        const displayMode = new Settings({displayMode: true});
+        expect("\\tag{1}\\tag{2}x+y").toNotParse();
+        expect("\\tag{1}\\tag{2}x+y").toNotParse(displayMode);
```
kevinbarabash: Because of strict mode, but we may want to support this in the future.

kevinbarabash:
@edemaine it looks like one of the screenshot tests is failing, either `FractionTest` or the one right after it.

src/MacroExpander.js
```diff
@@ -21,7 +20,7 @@ export default class MacroExpander implements MacroContextInterface {
 
     constructor(input: string, macros: MacroMap, mode: Mode) {
         this.lexer = new Lexer(input);
-        this.macros = objectAssign({}, builtinMacros, macros);
+        this.macros = Object.assign({}, builtinMacros, macros);
```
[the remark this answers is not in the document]
edemaine: It's used in `src/domTree.js` (in what's now `HtmlDomContainer`, but it predates that).  But Webpack already compiles it to
```js
this.style = __WEBPACK_IMPORTED_MODULE_3_babel_runtime_core_js_object_assign___default()({}, style);
```
along with ~90 lines implementing `Object.assign`, with a fallback to the native `Object.assign`.  I imagine this is about as good as what `objectAssign` was doing, and seems simpler.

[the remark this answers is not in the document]
kevinbarabash: I didn't know any of that.  Thanks for the explanation.  Sounds good to me.

test/katex-spec.js
kevinbarabash: Also, what happens with the following:
```
\gdef{\foo#1{#1 + \foo}} \foo{1}
```
My assumption is that this will cause an infinite loop.  Is example valid LaTeX?  Does it also cause an infinite loop there as well?

src/buildHTML.js
```diff
@@ -712,7 +711,29 @@ export default function buildHTML(tree, options) {
...
+                lastChild.children.pop();
+                const strut = makeSpan(["strut"]);
+                strut.style.height = (htmlNode.height + htmlNode.depth) + "em";
+                strut.style.verticalAlign = (-htmlNode.depth) + "em";
+                htmlNode.children.push(makeSpan(["tag"], [
+                    strut,
+                    lastNode,
+                ]));
+            }
+        }
```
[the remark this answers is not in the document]
edemaine: I have a solution to this messiness that I'm working on.  It involves wrapping the parse tree in a different sort of `tag` node, making it easy to peel off on the outside.  Coming soon!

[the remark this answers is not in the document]
kevinbarabash: Looking forward to see what you come up with.

test/katex-spec.js
edemaine: Here's a demo of the first example:
image

Yes, `\gdef\foo{\foo}\foo` causes an infinite loop.  I confirmed that LaTeX does the same; I didn't even see a stack overflow or other resource error, it just hung forever.  It might be good to put a limit on the MacroExpander to prevent a DOS attack.

test/__snapshots__/mathml-spec.js.snap
```diff
@@ -209,32 +209,30 @@ exports[`A MathML builder should render boldsymbol with the correct mathvariants
 <math>
   <semantics>
     <mrow>
-      <mrow>
```
kevinbarabash: Not sure what cause this change, but I'm happy to see it.

test/__snapshots__/mathml-spec.js.snap
kevinbarabash: I forgot that we were adding annotations to everything.  I see the magic is up above.

test/__snapshots__/mathml-spec.js.snap
```diff
@@ -505,3 +499,55 @@ exports[`A MathML builder should use <munderover> for large operators 1`] = `
 </math>
 
 `;
+
+exports[`A MathML builder tags use <mlabeledtr> 1`] = `
+
+<math>
+  <semantics>
+    <mtable side="right">
```
kevinbarabash: I assume that this causes the first `<mtd>` to be layed out to the right of the second.

test/katex-spec.js
```diff
@@ -2681,6 +2699,30 @@ describe("A macro expander", function() {
...
+    it("should build", () => {
+        expect("\\tag{hi}x+y").toBuild(displayMode);
+    });
+
+    it("should ignore location of \\tag", () => {
+        expect("\\tag{hi}x+y").toParseLike("x+y\\tag{hi}", displayMode);
+    });
+
+    it("should handle \\tag* like \\tag", () => {
+        expect("\\tag{hi}x+y").toParseLike("\\tag*{({hi})}x+y", displayMode);
```
kevinbarabash: Nice test

src/buildHTML.js
```diff
@@ -659,11 +659,17 @@ export default function buildHTML(tree, options) {
     // of the incoming tree so that it isn't accidentally changed
     tree = JSON.parse(JSON.stringify(tree));
 
+    // Strip off outer tag wrapper for processing below.
+    let tag = null;
+    if (tree.length === 1 && tree[0].type === "tag") {
+        tag = tree[0].value.tag;
+        tree = tree[0].value.body;
```
kevinbarabash: This is much better.

src/MacroExpander.js
```diff
@@ -149,6 +159,13 @@ export default class MacroExpander implements MacroContextInterface {
             return topToken;
         }
         const {tokens, numArgs} = this._getExpansion(name);
+        if (this.maxExpand !== Infinity) {
+            this.maxExpand--;
+            if (this.maxExpand < 0) {
```
kevinbarabash: Thanks for adding this check + config option.

src/buildHTML.js
```diff
@@ -691,22 +697,43 @@ export default function buildHTML(tree, options) {
...
+
+    const htmlNode = makeSpan(["katex-html"], children);
+    htmlNode.setAttribute("aria-hidden", "true");
+
+    // Adjust the strut of the tag to be the maximum height of all children
+    // (the height of the enclosing htmlNode) for proper vertical alignment.
+    if (tag) {
+        const strut = tagChild.children[0];
+        strut.style.height = (htmlNode.height + htmlNode.depth) + "em";
+        strut.style.verticalAlign = (-htmlNode.depth) + "em";
```
kevinbarabash: I feel like we do similar things elsewhere.  At some point we may want to create a helper for this.

test/__snapshots__/mathml-spec.js.snap
edemaine: I added a change to remove the outer mrow if all it contained was an mrow or mtable. This is especially nice (I think) for tagged equations, but thought I'd add it for mrow too.

test/__snapshots__/mathml-spec.js.snap
edemaine: Yes. The order of the columns is strict: the first one should be the label, and the second one should be the formula, and then this controls the visual layout. See [mlabeledtr docs](https://developer.mozilla.org/en-US/docs/Web/MathML/Element/mlabeledtr). (This is all new to me so I hope I got it all right!)

src/buildHTML.js
edemaine: Ah yes, I copy/pasted from `buildHTMLUnbreakable`. We could factor it out.

edemaine:
Thanks for the quick review! Glad you liked all the changes. Exciting to have this finally done.

test/__snapshots__/mathml-spec.js.snap
kevinbarabash: I had a look at the docs.  Looks reasonable to me.

kevinbarabash:
@edemaine we'll want to eventually move `tag` into `/functions` like everything else so we'll need to figure out a way to implement it using `defnineFunction`.  I wonder if we can split up `buildHTML` and export the chunk that `tag` uses so that the `tag` implementation could be extracted to a separate file.

ibayer:
@edemaine 
Thanks for adding the `\tag` support!

Can this be used to reference a tagged equation?
I also observed that the tag doesn't seem to respect the document layout (see tag way out on the left
in screen shot below). I have compared this with Mathjax line numbering which stays in the text column.
Is this the expected behavior?

image

I'm using KaTex to render equation in markdeep.

doraven:
@ibayer 
I had the same issue as you.
image
It's extremly right of the explore window.
or this one looking better
image

looking inside the dom structure , i found that the upper one did not set div width.
image
It's better now

edemaine:
@doraven What did you set the div width to in order to fix the problem? Was it specific to your layout, or something could be applied generically?

edemaine:
#1721 should fix this issue.

PR #1314, opened 2018-05-13
Strict setting controls \newline display-mode behavior; fix MacroExpander space handling

edemaine:
Two commits:
1. https://github.com/Khan/KaTeX/commit/c1c5e7ae115f7f606d13ffcee6f1dd4a4e54bda4 uses `strict` to control the behavior of `\\` and `\newline` in display mode.
    * New `"newLineInDisplayMode"` error code.
    * `\\` is now a macro, bound either to `\newline` (at top level) or `\cr` (within array environment). This is closer to how LaTeX behaves, and it lets the `cr` handler know what `\\` is supposed to represent (by not having to handle `\\`).
    * When `strict` is `"warn"` or `"ignore"` or `false` or a function that returns any of those, `\\` causes a newline in display mode.  When `strict` is `"error"` or `true` or a function that returns one of those or a function that raises an error, `\\` doesn't display anything.  This is controlled via a new `strictBehavior` helper.
2. https://github.com/Khan/KaTeX/pull/1314/commits/b50c0502b8c22fd8cc21f0bbd38816c6fa24175d fixes a bug that making `\\` into a macro revealed: when `\\` expands to `\cr`, it caused the MacroExpander to consume spaces after the `\\` (because it should consume spaces after `\cr) even though that's not how LaTeX works.  I fixed this by putting the space-after-control-word consumption into the lexer, which is where it belongs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,6 +79,9 @@ You can provide an object of options as the last argument to `katex.render` and 
 - `strict`: `boolean` or `string` or `function` (default: `"warn"`). If `false` or `"ignore`", allow features that make writing LaTeX convenient but are not actually supported by (Xe)LaTeX (similar to MathJax). If `true` or `"error"` (LaTeX faithfulness mode), throw an error for any such transgressions. If `"warn"` (the default), warn about such behavior via `console.warn`. Provide a custom function `handler(errorCode, errorMsg, token)` to customize behavior depending on the type of transgression (summarized by the string code `errorCode` and detailed in `errorMsg`); this function can also return `"ignore"`, `"error"`, or `"warn"` to use a built-in behavior.  A list of such features and their `errorCode`s:
   - `"unicodeTextInMathMode"`: Use of Unicode text characters in math mode.
   - `"mathVsTextUnits"`: Mismatch of math vs. text commands and units/mode.
+  - `"newLineInDisplayMode"`: Use of `\\` or `\newline` in display mode.
+    This never throws an error, but in strict mode, no line break results,
+    as in LaTeX.
 
 For example:
 
@@ -131,13 +134,6 @@ will appear larger than 1cm in browser units.
 - MathJax defines `\color` to be like `\textcolor` by default; set KaTeX's
   `colorIsTextColor` option to `true` for this behavior.  KaTeX's default
   behavior matches MathJax with its `color.js` extension enabled.
-- KaTeX breaks lines with `\\` and `\newline` in inline math, but ignores them
-  in display math (matching LaTeX's behavior, but not MathJax's behavior).
-  To allow `\\` and `\newline` to break lines in display mode,
-  add the following CSS rule:
-  ```css
-  .katex-display > .katex > .katex-html > .newline { display: block !important; }
-  ```
 
 ## Libraries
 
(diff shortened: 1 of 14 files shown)


src/Settings.js
```diff
@@ -84,6 +84,44 @@ class Settings {
...
+                "LaTeX-incompatible input and strict mode is set to 'warn': " +
+                `${errorMsg} [${errorCode}]`);
+            return false;
+        } else {  // won't happen in type-safe code
+            typeof console !== "undefined" && console.warn(
+                "LaTeX-incompatible input and strict mode is set to " +
+                `unrecognized '${strict}': ${errorMsg} [${errorCode}]`);
+            return false;
+        }
+    }
```
ylemkimon: I think the name `nonstrict` and `strictBehavior` are confusing and it should start with verb, as they are methods.

README.md
```diff
@@ -79,6 +79,9 @@ You can provide an object of options as the last argument to `katex.render` and
 - `strict`: `boolean` or `string` or `function` (default: `"warn"`). If `false` or `"ignore`", allow features that make writing LaTeX convenient but are not actually supported by (Xe)LaTeX (similar to MathJax). If `true` or `"error"` (LaTeX faithfulness mode), throw an error for any such transgressions. If `"warn"` (the default), warn about such behavior via `console.warn`. Provide a custom function `handler(errorCode, errorMsg, token)` to customize behavior depending on the type of transgression (summarized by the string code `errorCode` and detailed in `errorMsg`); this function can also return `"ignore"`, `"error"`, or `"warn"` to use a built-in behavior.  A list of such features and their `errorCode`s:
   - `"unicodeTextInMathMode"`: Use of Unicode text characters in math mode.
   - `"mathVsTextUnits"`: Mismatch of math vs. text commands and units/mode.
+  - `"newLineInDisplayMode"`: Use of `\\` or `\newline` in display mode.
+    This never throws an error, but in strict mode, no line break results,
```
ylemkimon: I think we can have two categories.

src/Parser.js
```diff
@@ -967,15 +967,15 @@ export default class Parser {
...
                 extraLatin.indexOf(text) >= 0) {
-                this.settings.nonstrict("unicodeTextInMathMode",
+                this.settings.reportNonstrict("unicodeTextInMathMode",
                     `Latin-1/Unicode text character "${text[0]}" used in ` +
                     `math mode`, nucleus);
             }
             symbol = new ParseNode(symbols[this.mode][text].group,
                             text, this.mode, nucleus);
         } else if (supportedCodepoint(text.charCodeAt(0))) {
             if (this.settings.strict && this.mode === 'math') {
```
ylemkimon: Not part of this PR, but `this.settings.strict` check seems not necessary. (same for above)

src/Settings.js
ylemkimon: I think they're okay, but I'm not a native English speaker.

src/Parser.js
edemaine: It's not necessary; it is an optimization, so the string didn't have to be generated unless strict mode was set to something that needed it (see the comment for `reportNonstrict`). This was probably more useful when the default for `strict` was `false`, though, making this a common case.

README.md
edemaine: Split into a second category.

PR #1318, opened 2018-05-16
Move test helpers into common modules

### LINKED ISSUE #1031: test helpers
As discussed previously with @kevinbarabash, we should move the helper functions in `katex-spec.js` like `toParse`, `toNotParse`, and `toParseLike` into a common module shared by all test modules.

I'd also like to suggest a `toBuildLike` which compares the HTML outputs.  I think we do this in various places by manually calling `getBuilt` twice and comparing.

edemaine:
Fix #1031 
* `helpers.js` gets all the helper functions
* `setup.js` gets the common Jest setup (serializer, expect extensions)

This PR is basically just refactoring; I did very little to clean up the code itself.  But the hope is that, by putting all the shared code (e.g. `expect` extensions) into a common module, it'll be easier to do test code cleanup.


edemaine:
@kevinbarabash Alternatively, we could add tests to test the test framework. 😄 
Shall I change that in this PR?

kevinbarabash:
> Shall I change that in this PR?

May as well.  Thanks. :)

test/helpers.js
```diff
@@ -0,0 +1,97 @@
...
+    try {
+        return _getBuilt(expr, settings || defaultSettings);
+    } catch (e) {
+        result.pass = false;
+        if (e instanceof ParseError) {
+            result.message = () => "'" + expr + "' failed " +
+                "parsing with error: " + e.message;
+        } else {
+            result.message = () => "'" + expr + "' failed " +
+                "parsing with unknown error: " + e.message;
```
ylemkimon: `parsing` -> `building` (same for above)

test/setup.js
```diff
@@ -2,18 +2,239 @@
...
+        } catch (e) {
+            result.pass = false;
+            if (e instanceof ParseError) {
+                result.message = () => "'" + actual + "' failed to " +
+                    "build with error: " + e.message;
+            } else {
+                result.message = () => "'" + actual + "' failed " +
+                    "building with unknown error: " + e.message;
+            }
+        }
```
ylemkimon: Why not use `buildAndSetResult`?

test/helpers.js
```diff
@@ -0,0 +1,97 @@
...
+import parseTree from "../src/parseTree";
+import Settings from "../src/Settings";
+
+export const defaultSettings = new Settings({
+    strict: false, // deal with warnings only when desired
+});
+export const strictSettings = new Settings({strict: true});
+
+export const _getBuilt = function(expr, settings) {
+    const usedSettings = settings ? settings : defaultSettings;
```
ylemkimon: I think we can use `||` and overwrite parameter, i.e., `settings = settings || defaultSetting`.

kevinbarabash:
I think it's probably best to DRY the code up in a separate PR.  I find reviewing changes were stuff is moved and refactored at the same time hard to follow.

package.json
```diff
@@ -87,6 +87,10 @@
     "nomnom": "^1.8.1"
   },
   "jest": {
+    "collectCoverageFrom": [
+      "src/**/*.js",
+      "contrib/**/*.js"
```
kevinbarabash: Thanks!

kevinbarabash:
@edemaine there's some lint and a couple of failing tests.

ylemkimon:
@edemaine I'll do it in another PR 😄

PR #1324, opened 2018-05-18
Specify flow output types for htmlBuilder and mathmlBuilder.

marcianx:
This is extremely trivial (no logic change or new assertions), and could actually have been done as part of the previous PR. It's useful to have this in before moving things into functions/* as the new additions are more strongly type-checked by this.

### DIFF
--- a/src/defineFunction.js
+++ b/src/defineFunction.js
@@ -2,13 +2,15 @@
 import {groupTypes as htmlGroupTypes} from "./buildHTML";
 import {groupTypes as mathmlGroupTypes} from "./buildMathML";
 import {checkNodeType} from "./ParseNode";
+import domTree from "./domTree";
 
 import type Parser from "./Parser";
-import type ParseNode from "./ParseNode";
-import type {NodeType, NodeValue} from "./ParseNode";
+import type ParseNode, {NodeType, NodeValue} from "./ParseNode";
 import type Options from "./Options";
 import type {ArgType, BreakToken, Mode} from "./types";
+import type {HtmlDomNode} from "./domTree";
 import type {Token} from "./Token";
+import type {MathNode, TextNode} from "./mathMLTree";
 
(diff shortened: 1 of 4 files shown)


src/defineFunction.js
```diff
@@ -104,13 +106,14 @@ type FunctionDefSpec<NODETYPE: NodeType> = {|
...
+    htmlBuilder?: (group: ParseNode<NODETYPE>, options: Options) => HtmlDomNode,
 
     // This function returns an object representing the MathML structure to be
     // created when rendering the defined LaTeX function.
-    // TODO: Make return type explicit.
-    mathmlBuilder?: (group: ParseNode<NODETYPE>, options: Options) => *,
+    mathmlBuilder?: (
+        group: ParseNode<NODETYPE>,
+        options: Options,
+    ) => MathNode | TextNode | domTree.documentFragment,
```
ylemkimon: Can it return a `documentFragment`?

marcianx: Surprisingly, yes:
https://github.com/Khan/KaTeX/blob/7b22eeb64ad7acf0090a393ba339f044e5886c15/src/functions/op.js#L219

ylemkimon: If it's used only once, I think we should refactor it or create separate method for MathML as `documentFragment` may have HTML-specific features in the future.

marcianx: Currently, I'm focused primarily on putting a type system in place to capture existing invariants and avoiding modifying any core logic (except where the type system catches actual errors). I've added TODO comments for this refactor separately, and I think it would be easier to do once all of KaTeX is strongly and statically typed.

edemaine: I also think it makes some sense to be able to allow a `documentFragment` return value... they are certainly helpful!  But anyway, we can leave this for future discussion.

PR #1325, opened 2018-05-18
Move "mclass" node logic to functions/mclass.

marcianx:
Cut and paste.

### DIFF
--- a/src/functions/mclass.js
+++ b/src/functions/mclass.js
@@ -1,0 +1,101 @@
+// @flow
+import defineFunction, {ordargument} from "../defineFunction";
+import buildCommon from "../buildCommon";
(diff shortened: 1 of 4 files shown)


src/functions/mclass.js
```diff
@@ -0,0 +1,101 @@
...
+import ParseNode from "../ParseNode";
+
+import * as html from "../buildHTML";
+import * as mml from "../buildMathML";
+
+const makeSpan = buildCommon.makeSpan;
+
+function htmlBuilder(group, options) {
+    const elements = html.buildExpression(group.value.value, options, true);
+
```
ylemkimon: I think blank line here is unnecessary.

marcianx: Removed.

PR #1327, opened 2018-05-19
 Move "xArrow" into functions/arrow.js and add flow types.

marcianx:
1) Copy-paste.
2) Add flow types.

### DIFF
--- a/src/stretchy.js
+++ b/src/stretchy.js
@@ -168,7 +168,7 @@ const groupLength = function(arg: ParseNode<*>): number {
 };
 
 const svgSpan = function(
-    group: ParseNode<"accent"> | ParseNode<"accentUnder">,
+    group: ParseNode<"accent"> | ParseNode<"accentUnder"> | ParseNode<"xArrow">,
     options: Options,
 ): DomSpan | SvgSpan {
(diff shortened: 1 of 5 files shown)


src/stretchy.js
```diff
@@ -180,9 +180,11 @@ const svgSpan = function(
         let viewBoxWidth = 400000;  // default
         const label = group.value.label.substr(1);
         if (utils.contains(["widehat", "widetilde", "utilde"], label)) {
+            // $FlowFixMe: implied by the labels above.
+            const grp: ParseNode<"accent"> | ParseNode<"accentUnder"> = group;
```
kevinbarabash: Should this also include `ParseNode<"xArrow">`?

marcianx: No, since that doesn't have `grp.value.base` which I access below. `"widehat", "widetilde", "utilde"` (from the `if` condition) all correspond to one of these two types, which is what I mean by "implied by the labels above".

kevinbarabash: I see.  Can you add those details to the `$FlowFixMe` comment?

marcianx: Done. Please see if it's clear enough.

PR #1332, opened 2018-05-20
Horiz brace

marcianx:
1) Copy-paste.
2) Add flow types.

Man, I keep learning new things about this code base (and the inadequacy of some of the types, which I'm going to have to deal with later) as I keep adding strict type checking.

### DIFF
--- a/src/functions.js
+++ b/src/functions.js
@@ -1,30 +1,11 @@
 // @flow
 /** Include this to ensure that all functions are defined. */
-import {
-    default as _defineFunction,
-    _functions,
-} from "./defineFunction";
-
-import type {FunctionPropSpec, FunctionHandler} from "./defineFunction";
-import type {NodeType} from "./ParseNode";
+import {_functions} from "./defineFunction";
 
 // WARNING: New functions should be added to src/functions and imported here.
 
 const functions = _functions;
 export default functions;
-
-// Define a convenience function that mimcs the old semantics of defineFunction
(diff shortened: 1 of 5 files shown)


src/functions.js
```diff
@@ -73,19 +54,7 @@ import "./functions/font";
...
-}, function(context, args) {
-    const base = args[0];
-    return {
-        type: "horizBrace",
-        label: context.funcName,
-        isOver: /^\\over/.test(context.funcName),
-        base: base,
-    };
-});
+import "./functions/horizBrace";
```
ylemkimon: 👍
I think we can remove new lines between `import` statements as they are single line.

src/functions.js
```diff
@@ -1,31 +1,12 @@
...
-import {
-    default as _defineFunction,
-    _functions,
-} from "./defineFunction";
-
-import type {FunctionPropSpec, FunctionHandler} from "./defineFunction";
-import type {NodeType} from "./ParseNode";
+import {_functions} from "./defineFunction";
 
 // WARNING: New functions should be added to src/functions and imported here.
```
ylemkimon: I think we can remove the warning, too.

marcianx: Done.

src/functions.js
marcianx: Done.
Also, as an additional commit at the end, I remove intermediate comments (these appear inconsistently only on an arbitrary subset) and alphabetized the whole list of functions/ imports. I can undo this last part if you prefer.

PR #1335, opened 2018-05-21
Spacing

marcianx:


src/domTree.js
```diff
@@ -15,17 +15,10 @@ import type Options from "./Options";
...
 const createClass = function(classes: string[]): string {
-    classes = classes.slice();
-    for (let i = classes.length - 1; i >= 0; i--) {
-        if (!classes[i]) {
-            classes.splice(i, 1);
-        }
-    }
-
-    return classes.join(" ");
+    return classes.filter(cls => cls).join(" ");
```
kevinbarabash: Nice!

src/buildCommon.js
```diff
@@ -222,10 +222,10 @@ const boldsymbol = function(
 /**
  * Makes either a mathord or textord in the correct font and color.
  */
-const makeOrd = function<NODETYPE: "textord" | "mathord">(
+const makeOrd = function<NODETYPE: "spacing" | "textord" | "mathord">(
     group: ParseNode<NODETYPE>,
     options: Options,
-    type: NODETYPE,
+    type: NODETYPE | "textord",
```
kevinbarabash: Isn't `"textord"` already included in `NODETYPE`?

marcianx: You're right, that's confusing. I changed `type` to `"mathord" | "textord"` instead. Previous PR dealt with the case where `ParseNode<"mathord">` was called type `"mathord"` and similarly for `"textord"` so the original type made sense. Now, I've got `ParseNode<"spacing">` is called with type `"textord"` which means that I have to break the tight-knit relationship between the `ParseNode` type and `type`.

src/domTree.js
ylemkimon: I'm not sure why, but `.filter(Boolean)` seems to be more commonly used.

marcianx: Fascinating. Never seen that myself. Would you prefer it in this case?

kevinbarabash: This is fine as is.

PR #1336, opened 2018-05-21
Refactor test helpers

ylemkimon:
* Combine common codes from `parse` and `build`, and dispatch using `Mode`
* Remove `toNotBuild` and `toNotParse` and use `expect.not`
* Support tagging literals: https://github.com/Khan/KaTeX/pull/1336#issuecomment-392411213
* Improve failed test `message`:
  - Use color
  - Print stack traces
  - Print diff
  - Follow jest matcher output style

expect | tests | success | ParseError | Warning | other errors
------ | ---- | ------- | ---------- | ------- | ------------
expect.toParse | parse | O | X | X | X
expect.not.toParse | parse | X | O | O | X
expect.toBuild | build | O | X | X | X
expect.not.toBuild | build | X | O | O | X
expect.toFailWithParseError | parse | X | O* | X | X
expect.not.toFailWithParseError | parse | O | X** | O | X
expect.toWarn | build | X | X | O | X
expect.not.toWarn | build | O | O | X | X

(O = pass, X = fail)
    * only if matches given string, if given
** only if doesn't match given string, if given


test/katex-spec.js
```diff
@@ -2416,9 +2413,7 @@ describe("A parser that does not throw on unsupported commands", function() {
     });
 
     it("should build katex-error span for other type of KaTeX error", function() {
-        // Use _getBuilt instead of getBuilt to avoid calling expect...toParse
-        // and thus throwing parse error
-        const built = _getBuilt("2^2^2", noThrowSettings);
+        const built = getBuilt("2^2^2", noThrowSettings);
```
ylemkimon: `_getBuilt` never threw an error due to `noThrowSettings`.

test/katex-spec.js
```diff
@@ -218,24 +218,21 @@ describe("A subscript and superscript parser", function() {
     });
 
     it("should produce the same thing regardless of order", function() {
-        const parseA = stripPositions(getParsed("x^2_3"));
-        const parseB = stripPositions(getParsed("x_3^2"));
-
-        expect(parseA).toEqual(parseB);
+        expect("x^2_3").toParseLike("x_3^2");
```
ylemkimon: This is only non-trivial change of test cases. Other changes are just `toNotXXX` to `not.toXXX`.

edemaine:
The advantage of the previous global warning -> error conversion is that we could make sure no code produced warnings when it wasn't supposed to.  In fact, I was going to suggest changing the default `strict` setting to the actual default, instead of `strict: false`, to make sure no warnings were generated when they weren't supposed to.  I guess such warnings would be printed to console, but it wouldn't prevent a PR from merging, for example.  What do you think?

ylemkimon:
@edemaine I think console outputs are noticeable enough. If we want to test that it doesn't produce warnings, we should use `expect(expr).not.toWarn()`.

kevinbarabash:
@ylemkimon are you planning changes or should I review?

ylemkimon:
@kevinbarabash I'm planning to change functions to support tagging literals, e.g.:
```js
const x = r`\KaTeX`;
```
```js
expect`\def\foo{1}`.toParseLike`1`;
```

edemaine:
@ylemkimon I'm no longer in agreement on the warnings. Running from the command-line, I'd see warnings, but if they don't cause tests to fail, Travis will still allow the merge, which it shouldn't.  Real-life example: In #1385 discussion, we spotted a bug through the existing warning -> error mechanism that we may not have noticed otherwise.

kevinbarabash:
@edemaine is there an action here for @ylemkimon?  I'm not sure what the next step is for this PR.

edemaine:
1. I would turn warnings back into errors, but that's open to discussion.
2. Someone needs to actually review the code; I haven't yet. 
3. Resolve conflicts (sorry!). 
4. @ylemkimon wanted to add some literal magic.

kevinbarabash:
Since it sounds like there's still some changes pending I'm going to mark this as "needs-revision".

ylemkimon:
I have reverted `console.warn()` to throw an error.

ylemkimon:
This is ready for review.

kevinbarabash:
I'm going to try to look at this this weekend.  I should've suggested saving the tagged template change for a separate PR.

ylemkimon:
@kevinbarabash I agree. For review:
- **Test helper improvement:**
  - https://github.com/Khan/KaTeX/pull/1336/commits/b3dcf40b9025b25041d7561f4b1a6d8f07c42c0f (except around `L221`)
  - https://github.com/Khan/KaTeX/pull/1336/commits/4e84e9397bcee78618b704a9bfa7d645e789be4b
  - https://github.com/Khan/KaTeX/pull/1336/commits/4e387b896cfaf00320ae131aebe4e2409f9efb59 (minor)
- **Test cases update:**
  - https://github.com/Khan/KaTeX/pull/1336/commits/b3dcf40b9025b25041d7561f4b1a6d8f07c42c0f#diff-ee8cb4a5b33e0ca875448106e9247be1L221
  - https://github.com/Khan/KaTeX/pull/1336/commits/4250d060888a85d7cc9a7be1d905001887f93375
  - https://github.com/Khan/KaTeX/pull/1336/commits/721f62f9e307d66aec4a2eaa7e0c96e3c0e9ab2e
  - https://github.com/Khan/KaTeX/pull/1336/commits/96cdfa0b4075255f39fad4ac95d7962026921a14
  - https://github.com/Khan/KaTeX/pull/1336/commits/4ad8dfdf73989bad4bdeb2f7646132c331c04196
- **Tagged literals:**
  - https://github.com/Khan/KaTeX/pull/1336/commits/17ca3840d2820fd1501a5b4d493f1a0771446a66
  - https://github.com/Khan/KaTeX/pull/1336/commits/a8c91ebb6fb28a846e951484d08dd4932f15f134
- **Minor fixes(no semantic change):**
  - https://github.com/Khan/KaTeX/pull/1336/commits/6cdff3c854d3f1a0235e6083e9f7fe417fe1c68a
  - https://github.com/Khan/KaTeX/pull/1336/commits/aeb5350fc2fa63835b7f3c659a5c8f6c00d3c38a

test/katex-spec.js
```diff
@@ -219,105 +212,102 @@ describe("A subscript and superscript parser", function() {
     });
 
     it("should produce the same thing regardless of order", function() {
-        const parseA = stripPositions(getParsed("x^2_3"));
-        const parseB = stripPositions(getParsed("x_3^2"));
-
-        expect(parseA).toEqual(parseB);
+        expect`x^2_3`.toParseLike`x_3^2`;
```
ylemkimon: Changed to use `toParseLike`.

test/katex-spec.js
```diff
@@ -1498,102 +1483,100 @@ describe("A comment parser", function() {
...
+        expect`x%y`.not.toParse();
     });
 
     it("should not produce or consume space", () => {
-        expect("\text{hello% comment 1\nworld}")
-            .toParseLike("\text{helloworld}");
-        expect("\text{hello% comment\n\nworld}")
-            .toParseLike("\text{hello world}");
+        expect("\\text{hello% comment 1\nworld}").toParseLike`\text{helloworld}`;
+        expect("\\text{hello% comment\n\nworld}").toParseLike`\text{hello world}`;
```
ylemkimon: There were some cases of using single backslash with `t` resulting in a horizontal tab(`\t`).

test/errors-spec.js
```diff
@@ -6,76 +6,76 @@ describe("Parser:", function() {
 
     describe("#handleInfixNodes", function() {
         it("rejects repeated infix operators", function() {
-            expect("1\\over 2\\over 3").toFailWithParseError(
+            expect`1\over 2\over 3`.toFailWithParseError(
```
kevinbarabash: This looks a little strange to me b/c I'm not used to template strings returning things that aren't strings.  Maybe we could introduce a `tex` template literal function.  It would look like:
```
expect(tex`1\over 2\over 3`).toFailWithParseError(...
```
The `tex` function could be exported for use by other people.  Also, it would be easier to understand what's going on.  I'm still not sure how `expect` is acting as a template function without us modifying/replacing it with our own implementation.

test/helpers.js
```diff
@@ -5,10 +5,59 @@ import ParseError from "../src/ParseError";
...
+    Error.captureStackTrace(this, global.console.warn);
+    this.name = this.constructor.name;
+    this.message = message;
+}
+Object.setPrototypeOf(ConsoleWarning.prototype, Error.prototype);
+
+/**
+ * Return the first raw string if x is tagged literal. Otherwise return x.
+ */
+export const r = x => x != null && x.hasOwnProperty('raw') ? x.raw[0] : x;
```
kevinbarabash: This function needs a better name.

test/helpers.js
```diff
@@ -67,34 +110,87 @@ export const stripPositions = function(expr) {
...
+        noun: 'building',
+        Verb: 'Build',
+    },
+};
+
+const getTree = (expr, settings, mode) => {
+    const result = expectKaTeX(expr, settings, mode);
     try {
-        return parseTree(expr, settings);
+        expect(result).toHavePassed();
```
kevinbarabash: I would expect `getTree` to call `katex.render()` and return the tree or clean up the stack trace and rethrow as necessary.  Then `expectKaTeX` can call `getTree` and we don't have introduce a special matcher just to make things work in this order.  Is there something I'm missing that necessitates this setup?

test/helpers.js
```diff
@@ -67,34 +110,87 @@ export const stripPositions = function(expr) {
...
-        }
+        Error.captureStackTrace(e, mode.get); // remove helpers in stack entries
+        throw e;
     }
+    return result._tree;
 };
 
-export const buildAndSetResult = function(expr, result,
-                                          settings = new Settings()) {
+export const expectKaTeX = (expr, settings = new Settings(), mode, isNot,
```
kevinbarabash: Does the default `settings` work when `undefined` is passed as the second arg?

kevinbarabash: I really like that we're supporting `.not` now, but this helper function has complicated the implementation quite a bit.  I'd rather have a few small helper functions that each of the `toParse()`, `toBuild()`, etc. call then a single function that tries to do it all.  Maybe we could add something format the message in addition to the function you added to clean up the stack trace.

test/helpers.js
ylemkimon: @kevinbarabash I borrowed it from python's raw string prefix. Maybe `tex`, as you suggested above?

test/helpers.js
ylemkimon: @kevinbarabash 
> Does the default settings work when undefined is passed as the second arg?

Yes. Default function parameters allow formal parameters to be initialized with default values if no value or `undefined` is passed.

test/helpers.js
kevinbarabash: I wonder if people will want to use it like
```
tex`\frac{${num}}{${den}}`
```
That could be a separate PR though if people think that's worth supporting.

ylemkimon:
@kevinbarabash 
> I would expect `getTree` to call `katex.render()` and return the tree or clean up the stack trace and rethrow as necessary.  Then `expectKaTeX` can call `getTree` and we don't have introduce a special matcher just to make things work in this order.

Yes, I agree. My original intention was to reuse the error message from `expectKaTeX` and clean-up stack traces. However, now I think, they are not needed, they made codes unnecessarily complicated, and even  made errors hard to debug.

I've removed `getTree` in f783eb8.

test/errors-spec.js
ylemkimon: @kevinbarabash I agree that it can be unfamiliar since it's new syntax. However, it's new way to call the function with arguments: an array of strings with `raw` property, an array of raw strings, and expressions in `${}`. As `expect` accepts any object, it just passes the first argument through and we can handle them in the matchers. Maybe we can document them in the `CONTRIBUTING.md`?

edemaine: If `expect` passes through only one argument, does that mean this won't work with `${interpolated expressions}`?

I don't think the new code *looks* much more confusing than the old code -- the weirdest part is that `expect` just passes on its argument, and the action is in the chained methods afterward -- but that's a general Jest thing.  However, until I read the [template literal description](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals) I didn't really understand what the code was doing.

What about `expectTex`1 \over 2 \over 3` which is equivalent to `expect(tex`1 \over 2 \over 3`)`, i.e., the string gets processed before calling `expect`, instead of requiring all matches to process it?  I think the weirdest part is overloading `expect` in this unintended way (unless Jest suggests doing this?).

ylemkimon: It's not related, but Jest suggests a way of [creating a table using tagged template literal](https://jestjs.io/docs/en/api#2-describeeach-table-name-fn):

```js
describe.each`
  a    | b    | expected
  ${1} | ${1} | ${2}
  ${1} | ${2} | ${3}
  ${2} | ${1} | ${3}
`('$a + $b', ({a, b, expected}) => {
  test(`returns ${expected}`, () => {
    expect(a + b).toBe(expected);
  });

  test(`returned value not be greater than ${expected}`, () => {
    expect(a + b).not.toBeGreaterThan(expected);
  });

  test(`returned value not be less than ${expected}`, () => {
    expect(a + b).not.toBeLessThan(expected);
  });
});
```

test/helpers.js
ylemkimon: @kevinbarabash Wouldn't it be same as `String.raw`?

test/errors-spec.js
kevinbarabash: Interesting.  I was unaware of jest's use of tagged template literals.  `expectTex` seems like a good compromise.

kevinbarabash: @ylemkimon apologies for not raising my concerns about `expect` as a tagged template earlier.

test/helpers.js
kevinbarabash: I supposed it would be in which case I guess people can just use `String.raw`. 😅

kevinbarabash:
> With removing getTree, I think expectKaTeX became simple enough. Maybe we can extract `expected` string construction into separate function, like `printActualErrorMessage`?

Sounds good.

test/errors-spec.js
ylemkimon: @kevinbarabash @edemaine 

> @ylemkimon apologies for not raising my concerns about `expect` as a tagged template earlier.

I don't mind.

> What about ```expectTex`1 \over 2 \over 3` ``` which is equivalent to ```expect(tex`1 \over 2 \over 3`)```, i.e., the string gets processed before calling `expect`, instead of requiring all matches to process it?  I think the weirdest part is overloading `expect` in this unintended way (unless Jest suggests doing this?).

> Interesting.  I was unaware of jest's use of tagged template literals.  `expectTex` seems like a good compromise.

I still disagree. The purpose of using tagged template literal is to avoid using escape backslashes and simplify test cases. I think using `expectTeX` defeats this purpose as there is no difference between processing before passing to `expect` and processing in the matchers.

ylemkimon:
>> With removing getTree, I think expectKaTeX became simple enough. Maybe we can extract `expected` string construction into separate function, like `printActualErrorMessage`?

> Sounds good.

Done in fc47230.

test/errors-spec.js
edemaine: @ylemkimon There would be a difference if we needed to use Jest's matchers.  But I can't actually think of any situation where we'd want to use those on a TeX string.

PR #1338, opened 2018-05-21
Cleanup MathML <mrow>, <mtext>, <mn>

### LINKED ISSUE #203: MathML for multi-digit numbers is wrong
From [this page on trig functions](https://www.khanacademy.org/math/trigonometry/less-basic-trigonometry/trig-symmetry-periodicity/e/symmetry-and-periodicity-of-trig-functions), this is the MathML for the first expr:  sin(α)=0.34

``` html
<math><semantics>
<mrow>
  <mo>sin</mo>
   <mo>(</mo>
   <mi>α</mi>
   <mo>)</mo>
   <mo>=</mo>
   <mn>0</mn><mi mathvariant="normal">.</mi><mn>3</mn><mn>4</mn>
</mrow>
<annotation encoding="application/x-tex">\sin (\alpha) = 0.34</annotation>
</semantics></math>
```

Notice that what should be `<mn>0.34</mn>` has turned into three `<mn>`s and one `<mi>`.

The "sin" should also be in a `<mi>` -- that's [#202](https://github.com/Khan/KaTeX/issues/202)

### LINKED ISSUE #880: Remove unnecessary <mrow>s from MathML output
Any `<mrow>` with a single child and no attributes probably doesn't need to be in the tree.  We should probably have some sort of post-processing step on the MathML tree before rendering it.

edemaine:
* Avoid unnecessary `<mrow>` wrapping (fix #880)
* buildMathML gains two helpers:
    * `makeRow` helper wraps an array of nodes in `<mrow>`, unless the array has length 1, in which case no wrapping is necessary.
    * `buildExpressionRow` for common case of `makeRow(buildExpression(...))`
* Combine adjacent `<mtext>`s in all cases, not just immediately within `\text`.
   * No more need for `makeTextRow` helper or anything fancy in text MathML handler.
* Concatenate `<mn>`s and decimal point into single `<mn>` (fix #203)

### DIFF
--- a/src/buildMathML.js
+++ b/src/buildMathML.js
@@ -162,18 +162,24 @@ groupTypes.tag = function(group, options) {
  */
 export const buildExpression = function(expression, options) {
     const groups = [];
-    let currentText = null;
+    let lastGroup;
     for (let i = 0; i < expression.length; i++) {
         const group = buildGroup(expression[i], options);
-        if (group.type === 'mtext' && currentText !== null) {
-            currentText.children.push(...group.children);
+        // Concatenate adjacent <mtext>s
+        if (group.type === 'mtext' && lastGroup && lastGroup.type === 'mtext') {
+            lastGroup.children.push(...group.children);
+        // Concatenate adjacent <mn>s
+        } else if (group.type === 'mn' &&
+                   lastGroup && lastGroup.type === 'mn') {
+            lastGroup.children.push(...group.children);
+        // Concatenate <mn>...</mn> followed by <mi>.</mi>
+        } else if (group.type === 'mi' && group.children.length === 1 &&
+                   group.children[0].text === '.' &&
+                   lastGroup && lastGroup.type === 'mn') {
+            lastGroup.children.push(...group.children);
         } else {
             groups.push(group);
-            if (group.type === 'mtext') {
-                currentText = group;
-            } else {
-                currentText = null;
-            }
+            lastGroup = group;
         }
     }
 
--- a/src/buildMathML.js
+++ b/src/buildMathML.js
@@ -40,29 +40,6 @@ export const makeRow = function(body) {
     }
 };
 
-export const makeTextRow = function(body, options) {
-    // Convert each element of the body into MathML, and combine consecutive
-    // <mtext> outputs into a single <mtext> tag.  In this way, we don't
-    // nest non-text items (e.g., $nested-math$) within an <mtext>.
-    const inner = [];
-    let currentText = null;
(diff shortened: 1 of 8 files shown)


test/__snapshots__/mathml-spec.js.snap
```diff
@@ -507,22 +526,9 @@ exports[`A MathML builder tags use <mlabeledtr> 1`] = `
...
-                i
-              </mtext>
-            </mrow>
-            <mtext>
-              )
-            </mtext>
-          </mrow>
+          <mtext>
+            (hi)
+          </mtext>
```
kevinbarabash: Awesome!

test/__snapshots__/mathml-spec.js.snap
```diff
@@ -112,6 +110,35 @@ exports[`A MathML builder should generate the right types of nodes 1`] = `
...
+
+<math>
+  <semantics>
+    <mrow>
+      <mi>
+        sin
+      </mi>
+      <mo>
+        ⁡
+      </mo>
```
kevinbarabash: Is this `<mo>` supposed to contain a `&#x2061;`?

src/buildMathML.js
```diff
@@ -181,40 +157,57 @@ groupTypes.tag = function(group, options) {
...
+            lastGroup.children.push(...group.children);
+        // Concatenate adjacent <mn>s
+        } else if (group.type === 'mn' &&
+                   lastGroup && lastGroup.type === 'mn') {
+            lastGroup.children.push(...group.children);
+        // Concatenate <mn>...</mn> followed by <mi>.</mi>
+        } else if (group.type === 'mi' && group.children.length === 1 &&
+                   group.children[0].text === '.' &&
+                   lastGroup && lastGroup.type === 'mn') {
+            lastGroup.children.push(...group.children);
```
kevinbarabash: Nice, I just wrote a comment about it, but I see you fixed it.  Thanks!

kevinbarabash: I see you added a test for it, yay!

kevinbarabash:
I looks like you have regenerate the snapshots.

test/__snapshots__/mathml-spec.js.snap
edemaine: I believe it already is.  Shame that we can't see it in the preview, but I checked there's something there.

edemaine:
Oops, fixed!

kevinbarabash:
@edemaine thanks for addressing these longstanding issues with our MathML output and getting us started on our 1.0 milestone (#203 was the first issue closed in the milestone).

PR #1339, opened 2018-05-22
Move "ordgroup" and "tag" builders into functions/*.

marcianx:


edemaine:
FYI, alas, this will conflict a bunch with #1338.  Whoever goes second should take care to port the changes over.

marcianx:
@edemaine: Do your thing. I can wait.

PR #1348, opened 2018-05-23
\def support (and \gdef and \global\def)

edemaine:
Solve most of #250.  Following [discussion there](https://github.com/Khan/KaTeX/issues/250#issuecomment-390637002), I implemented `\def` (and reimplemented `\gdef`) to behave as in LaTeX.
* Re-organize macros object as a new `Namespace` object.  TeX would call these "groups", but we use group all over the place in the parser, so I thought this would be too confusing.
* There is one `builtinNamespace` that loads the built-in macros once.  Unlike current code, this never gets copied!  So it should speed up repeated calls to KaTeX.
* Each namespace has a `parent` pointer, ultimately ending in `builtinNamespace`.  `getMacro` walks up this hierarchy as necessary.
* `setMacro` sets in the local namespace, or if `global` is specified, in *all* namespaces up the hierarchy.  I didn't realize how TeX behaves, but it is.

An alternative design would be to store the namespaces in an array (stack), so local `setMacro` would set within `namespaces[namespaces.length-1]`, while global `setMacro` would be a for loop over the namespaces array.  The advantage I see of the current approach is that the complexity can go into the `Namespace` class.  But I'm happy to rewrite to this alternative approach, which would put that complexity into `MacroExpander`.

One important note (also in the README) is that the `macros` object passed into `katex.render` etc. will be modified if the LaTeX code defines its own macros via `\def` or `\gdef`.  This is backwards-incompatible, and could be an issue if the same macros object gets passed in many times.  (I added `katex.__defineMacro` to define extra built-in macros, to provide a better way to provide the same macros many times.)  But I think it's also an important feature: it enables consecutive calls to KaTeX to share state.  For example, one math blob can define some macros, and later math blobs can use those macros.  This is extremely useful.  I believe MathJax behaves the same way.  It will also help with automatic equation numbering.

This PR paves the way for finishing #250, as well as length support (#687).


kevinbarabash:
Looking forward to reviewing this this evening. :)

src/MacroExpander.js
```diff
@@ -153,38 +172,38 @@ export default class MacroExpander implements MacroContextInterface {
...
                 throw new ParseError("Too many expansions: infinite loop or " +
                     "need to increase maxExpand setting");
             }
         }
-        let expansion = tokens;
-        if (numArgs) {
-            const args = this.consumeArgs(numArgs);
+        let tokens = expansion.tokens;
+        if (expansion.numArgs) {
+            const args = this.consumeArgs(expansion.numArgs);
```
kevinbarabash: It looks like this renaming doesn't actually change any of the logic, is that correct?

src/MacroExpander.js
```diff
@@ -234,11 +253,13 @@ export default class MacroExpander implements MacroContextInterface {
...
     /**
      * Returns the expanded macro as a reversed array of tokens and a macro
-     * argument count.
-     * Caches macro expansions for those that were defined simple TeX strings.
+     * argument count.  Or returns `null` if no such macro.
      */
-    _getExpansion(name: string): MacroExpansion {
-        const definition = this.macros[name];
+    _getExpansion(name: string): ?MacroExpansion {
+        const definition = this.namespace.getMacro(name);
```
kevinbarabash: It would be nice to have a convenience function `getMacro` directly on `MacroExpander` since we do `.namespace.getMacro` in a couple of places.

README.md
```diff
@@ -81,7 +81,7 @@ You can provide an object of options as the last argument to `katex.render` and
...
 - `errorColor`: `string`. A color string given in the format `"#XXX"` or `"#XXXXXX"`. This option determines the color that unsupported commands and invalid LaTeX are rendered in when `throwOnError` is set to `false`. (default: `#cc0000`)
-- `macros`: `object`. A collection of custom macros. Each macro is a property with a name like `\name` (written `"\\name"` in JavaScript) which maps to a string that describes the expansion of the macro. Single-character keys can also be included in which case the character will be redefined as the given macro (similar to TeX active characters).
+- `macros`: `object`. A collection of custom macros. Each macro is a property with a name like `\name` (written `"\\name"` in JavaScript) which maps to a string that describes the expansion of the macro. Single-character keys can also be included in which case the character will be redefined as the given macro (similar to TeX active characters). *This object will be modified* if the LaTeX code defines its own macros via `\def` or `\gdef`, which enables consecutive calls to KaTeX to share state.
```
kevinbarabash: Nice warning.  I wonder if we should make a copy so we don't have to give this warning.

src/Namespace.js
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    setMacro(name: string, expansion: MacroExpansion, global: Boolean = false) {
+        // Global set is equivalent to setting in all namespaces up the stack.
+        if (global) {
+            let namespace = this;
+            while (namespace) {
+                namespace.macros[name] = expansion;
+                namespace = namespace.parent;
+            }
```
ylemkimon: What’s the difference between defining the macro at the top-level and defining at all levels?

src/Namespace.js
```diff
@@ -0,0 +1,44 @@
...
+    constructor(parent: ?Namespace, macros: MacroMap = {}) {
+        this.parent = parent;
+        this.macros = macros;
+    }
+
+    getMacro(name: string): ?MacroExpansion {
+        let ns = this;
+        while (ns && !ns.macros.hasOwnProperty(name)) {
+            ns = ns.parent;
+        }
```
ylemkimon: Wouldn’t this increase the complexity to O(k), where k is number of namespaces(groups)? Though, I’m not sure how much it will affect the performance.

src/Namespace.js
kevinbarabash: If we only define it at the top level as soon as we exit the current group then it will no longer be defined, or, if it was previously defined using `\def` in that parent group, then the macro would use that definition.

src/Namespace.js
kevinbarabash: I'm a little concerned about performance.  We could do some benchmarking and/or profiling to see how much this affects things.  We could maybe save a little bit of look up time by reusing the namespace from the parent group and only creating a new one if we run into a `\def` statement.  If we also reuse the `builtinNamespace` in the same way then this should have almost zero impact on users who aren't using `\def` b/c each lookup will be O(1).

src/Namespace.js
ylemkimon: I meant the bottom of the stack 😄 But as you said, there is a precedence, so how about deleting the definition if exists and defining only at the bottom?

kevinbarabash: The thing is there could be a definition at any level (or all levels) so `\global\def` has to clobber them all.

ylemkimon: I was thinking something like:
```js
let namespace = this;
while (namespace.parent) {
    delete namespace.macros[name];
    namespace = namespace.parent;
}
namespace.macros[name] = expansion;
```

test/katex-spec.js
```diff
@@ -2681,6 +2681,36 @@ describe("A macro expander", function() {
         expect("\\gdef{}{}").toNotParse();
     });
```
ylemkimon: It seems `\def`s expect a control sequence right after its command, so `\def{\foo}{1}` is invalid LaTeX.

test/katex-spec.js
```diff
@@ -2956,6 +2986,12 @@ describe("Newlines via \\\\ and \\newline", function() {
     it("should not allow \\cr at top level", () => {
         expect("hello \\cr world").toNotBuild();
     });
+
+    it("array redefines and resets \\\\", () => {
+        expect("a\\\\b\\begin{matrix}x&y\\\\z&w\\end{matrix}\\\\c")
+            .toParseLike("a\\newline b\\begin{matrix}x&y\\cr z&w\\end{matrix}" +
+                         "\\newline c");
+    });
 });
```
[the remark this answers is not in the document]
ylemkimon: @kevinbarabash It works in LaTeX.

src/Namespace.js
kevinbarabash: I see, but then lookup is always expensive for global macros b/c you have to iterate to the root every time.

test/katex-spec.js
```diff
@@ -2956,6 +2986,12 @@ describe("Newlines via \\\\ and \\newline", function() {
     it("should not allow \\cr at top level", () => {
         expect("hello \\cr world").toNotBuild();
     });
+
+    it("array redefines and resets \\\\", () => {
+        expect("a\\\\b\\begin{matrix}x&y\\\\z&w\\end{matrix}\\\\c")
+            .toParseLike("a\\newline b\\begin{matrix}x&y\\cr z&w\\end{matrix}" +
+                         "\\newline c");
```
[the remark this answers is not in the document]
ylemkimon: With tagged templates, I think we can do something like:
```js
const expectToParse = str => expect(str.raw[0]).toParse();
expectToParse`\foo`;
```
I'll add them to #1336.

src/Parser.js
```diff
@@ -844,6 +844,10 @@ export default class Parser {
             if (mode) {
                 this.switchMode(mode);
             }
+            // Start a new namespace
+            if (!optional) {
+                this.gullet.pushNamespace();
```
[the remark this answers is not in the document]
ylemkimon: It seems they are allowed in optional groups and their scope is limited to the group, e.g. `\sqrt[\def\foo{3}\foo]{x}` works but `\sqrt[\def\foo{3}]{x\foo}\foo` doesn't.

src/Namespace.js
ylemkimon: Also, I think `\gdef` should modify `builtinNamespace`.

ylemkimon:
How about:

Level | Scope | Macro definition in JS | in LaTeX
----- | ----- | ---------------------- | --------
**0** | global(document, KaTeX instance) | `builtinNamespace`,<br>`__defineMacro` | `\gdef`
**1** | expression(function call) | `settings.macro` | `\def` at the top level
**>2** | group(lazily instantiated) | N/A | `\def`

?

test/katex-spec.js
edemaine: Also `\def\foo 1` doesn't work in TeX (the `1` is a pattern match, not an expansion). Unfortunately, I don't know how to detect either case. Our current MacroExpander doesn't distinguish between `\foo` and `{\foo}`, I believe (both return `[{text: '\foo'}]`).

edemaine:
1. I have a plan for making lookups and local defs constant time, though gdefs will still take O(k). Essentially, the idea is to have an undo stack. This is inspired by how TeX does it (based on a rough reading of the source code, though I still don't fully understand it), which I believe achieves the same performance.

2. I really don't like the idea of built-ins being modifiable by LaTeX code. This makes it difficult or impossible to isolate multiple LaTeX documents with the same KaTeX module. This would be an issue in any system with multiple LaTeX authors. We don't want macros defined by one document to adversely affect the other documents. The current system puts the caller in control of when to share state and when to reset.

test/katex-spec.js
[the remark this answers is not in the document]
edemaine: Yes, that will work too. However, as another commented-out test mentions, `\global\mydef...` doesn't work yet (but should), because I'm not grabbing the argument to `\global` correctly.

ylemkimon:
> I really don't like the idea of built-ins being modifiable by LaTeX code. This makes it difficult or impossible to isolate multiple LaTeX documents with the same KaTeX module. This would be an issue in any system with multiple LaTeX authors.

Now I think, it isn't good idea. I didn't thought of server-side rendering, which may share single KaTeX instance, or single page containing multiple documents.

kevinbarabash:
@edemaine do you want to save the optimization for a separate PR?

edemaine:
I'm happy to work on it first, unless you think this is high priority. It will probably involve rewriting much of the code.

kevinbarabash:
I thought it might be more of an incremental change, but if it's a rewrite let's wait.

edemaine:
OK, I did the constant-time lookup rewrite.

Now we use an "undo" stack similar to TeX, so get and set are constant-time operations.  `get()` still has to check two objects: one with all current settings, and then the built-ins.  Local `set()` sets the current value and (when appropriate) adds an undo operation to the undo stack.  Global `set()` still takes time linear in the number of groups, possibly changing the undo operation at every level.

`Namespace` now refers to a space of nameable things like macros or lengths.  This is a totally different meaning from before, but still a reasonable name for the thing, I think, and still encapsulates all the complexity into the `Namespace` class.

src/Parser.js
```diff
@@ -89,7 +89,7 @@ export default class Parser {
         // Use old \color behavior (same as LaTeX's \textcolor) if requested.
         // We do this after the macros object has been copied by MacroExpander.
         if (settings.colorIsTextColor) {
-            this.gullet.macros["\\color"] = "\\textcolor";
```
ylemkimon: Wouldn't this modify `settings.macro`?

src/Parser.js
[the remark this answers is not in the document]
ylemkimon: Doesn't a optional group start new namespace?

src/parseTree.js
```diff
@@ -18,11 +18,13 @@ const parseTree = function(toParse: string, settings: Settings): ParseNode<*>[]
         throw new TypeError('KaTeX can only parse string typed expression');
     }
     const parser = new Parser(toParse, settings);
+    // Blank out any \df@tag to avoid spurious "Duplicate \tag" errors
+    delete parser.gullet.macros.current["\\df@tag"];
```
ylemkimon: I think `current` should not be accessed directly, and it'd be nice to have expression-scope namespace as we may add other state-macros like `\df@tag`.

src/Parser.js
```diff
@@ -89,7 +89,7 @@ export default class Parser {
...
+            this.gullet.macros.set("\\color", "\\textcolor");
```
ylemkimon: Wouldn't this modify `settings.macro`?

src/parseTree.js
kevinbarabash: How does LaTeX handle this?

static/main.js
```diff
@@ -78,6 +79,8 @@ function init() {
     }
 
     function reprocess() {
+        // Ignore changes to global macros caused by the expression
+        options.macros = Object.assign({}, macros);
```
kevinbarabash: Good call.

test/katex-spec.js
kevinbarabash: I'm pretty sure changing this would break a whole bunch of things, but maybe we should look at doing so sometime in the future.  We can add some kind of helper function to unwrap groups with a single child.

test/katex-spec.js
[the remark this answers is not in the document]
kevinbarabash: @edemaine did you changes address this?

src/Parser.js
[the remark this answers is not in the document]
edemaine: Sorry, I forgot to fix this according to @ylemkimon's experiments. (I had assumed optional groups didn't make their own groups because they didn't use `{}` notation.) I'll remove the `if (!optional)`.

src/MacroExpander.js
edemaine: Correct.  There was a need for `expansion` being the thing containing the tokens, so I did the rename.  Sorry it's confusing as a diff.

src/MacroExpander.js
edemaine: The new `.macros.get(...)` is probably good now.

README.md
edemaine: As stated in the root post, this is important for functionality sake, so you can share macros across calls (including e.g. the current equation number).  But it is also confusing and can lead to bugs if you don't realize what's happening... so I'm not sure what's best.  A toggle?

src/Namespace.js
```diff
@@ -0,0 +1,73 @@
...
+                this.current[undef] = undefs[undef];
+            }
+        }
+    }
+
+    get(name: string): ?Value {
+        if (this.current.hasOwnProperty(name)) {
+            return this.current[name];
+        } else {
+            return this.builtins[name];
```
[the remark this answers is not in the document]
edemaine: No, it can't.  (I assume you mean `||`, as `|` doesn't short circuit.  I used to have it this way, but it fails tests.)  We need to distinguish a macro set to the empty string (which evaluates to `false`) vs. an `undefined` value.

src/Namespace.js
```diff
@@ -0,0 +1,73 @@
...
+    current: Mapping<Value>;
+    builtins: Mapping<Value>;
+    undefStack: Mapping<Value>[];
+
+    /**
+     * Both arguments are optional.  The first argument is an object of built-in
+     * mappings which never change.  The second argument is an object of initial
+     * mappings, which will constantly change according to any `set`s done.
+     */
+    constructor(builtins: Mapping<Value> = {}, topGroup: Mapping<Value> = {}) {
```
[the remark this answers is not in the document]
edemaine: I thought `topGroup` made it clearer that the macro set might be modified.  But happy to change to something else.

test/katex-spec.js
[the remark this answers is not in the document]
edemaine: No.  I'll add some TODOs.

test/katex-spec.js
edemaine: Adding a TODO in `test/katex-spec.js`.

src/parseTree.js
edemaine: @ylemkimon is right!  I tested the following with LaTeX:
```latex
\def\x{1}\x
\[\def\x{2}\]
\x
$\def\x{3}$
\x
$$\def\x{4}$$
\x
```
and all of the `\x`s (outside math mode where they're set) evaluate to `1`.  So TeX does make a group for each math-mode expression, justifying adding such a thing.

src/Parser.js
edemaine: Indeed, good catch!  I've tweaked this to no longer happen, thanks to the new expression-level group, and added a corresponding test.

edemaine:
I think I got all the comments, but let me know if I missed anything.

src/MacroExpander.js
kevinbarabash: It's fine.  The refactor was pretty isolated.  I just wanted to be sure I wasn't missing something here.

ronkok:
I'm working on a test page with version 0.10.0-beta, and I haven't yet gotten `\gdef` to work as I think it is intended. Here's my HTML:
```
<p>\( \gdef\foo{\sum} \foo  \)</p>
<p>\( \foo \)</p>
```

The first paragraph works as expected. The second paragraph returns a parse error: "Expected 'EOF', got '\foo' at position 2".

Should the second paragraph render a `\sum`?

edemaine:
`auto-render` doesn't preserve macros from math expression to math expression by default.  Perhaps it should?  But you should be able to make it do so by replacing
```js
renderMathInElement(document.body);
```
with
```js
var macros = {}; // preserve macros between multiple expressions
renderMathInElement(document.body, {macros: macros});
```

ronkok:
@edemaine Yes, that works! We should revise the documentation for auto-render.
 
By the way, this whole global macro thing is awesome.

I also tried `\newcommand`, and it does not preserve macros from expression to expression.  I take it that is the intent?

ronkok:
One last thing while I'm testing 0.10.0-beta:  The KaTeX README page, the auto-render README, and the KaTeX demo page all currently point to 0.10.0-alpha, not beta. Which is weird because at least two them pointed at -beta yesterday.

edemaine:
Yes, `\newcommand` works like `\def`. Sadly there seems to be no LaTeX equivalent to `\gdef`.

I think it'd make sense to modify auto-render to at least share macros within the same call to `renderMathInElement`... but agreed, should clarify in docs.

I wouldn't have expected the demo page to point to beta, but I thought the release scripts would have handled the READMEs.  They are correct in the release tag (https://github.com/Khan/KaTeX/commit/8624d4ccffbf8543ec87e75b24ed9551a7c16714) but that has other things which we don't want to merge in.  I guess we have to copy/paste manually?

ronkok:
Okay, thanks for the explanation of `\newcommand`. It will help me to get the function support page correct.

qnxor:
> auto-render doesn't preserve macros from math expression to math expression by default. Perhaps it should?

@edemaine sorry for resurrecting this old thread, but could we make that the default? Persisting macros is, after all, expected Latex behavior.

edemaine:
> @edemaine sorry for resurrecting this old thread, but could we make that the default? Persisting macros is, after all, expected Latex behavior.

Things have changed since this old thread. auto-render does now share a `macros` object across all calls to KaTeX. See the last paragraph of https://katex.org/docs/autorender.html

> An entry in the docs about achieving persistent macros with/without auto-render would be useful

The relevant option is `macros`, as defined here: https://katex.org/docs/options.html

We recently added a docs section about this in fact: https://katex.org/docs/api.html#persistent-macros
If you have ideas for further docs improvement, I'm all ears!
(but please open a new issue or discussion for it)

qnxor:
> Things have changed since this old thread. auto-render does now share a macros object across all calls to KaTeX. See the last paragraph of https://katex.org/docs/autorender.html

My point was that it should be enabled by default. It's expected Latex behavior. Mathjax behaves correctly in that sense, whereas Katex requires manual work to enable it.

> We recently added a docs section about this in fact: https://katex.org/docs/api.html#persistent-macros

I missed that one, thanks for pointing it out. Seems sufficient for developers. I just [patched](https://github.com/discourse/discourse-math/pull/51/commits/f18f0b456b72888fadb6096f0e641f97438e4079) Discourse to enable it (but it should really be enabled by default in Katex in my view).

edemaine:
It is enabled by default in auto-render. Are you suggesting that KaTeX's `render(ToString)` should use a global variable with all current macros that gets re-used every call? That seems like both a security risk and prone to error; it would disable a lot of applications that want to run many isolated LaTeX "instances". For example, in an online message forum, I wouldn't want macros defined in one poster's message to bleed into another poster's message; a malicious poster could redefine all LaTeX commands and cause all other uses of KaTeX to fail, and even a non-malicious poster could easily break other people's messages by accident. By isolating the macro state to each user message, or whatever each system using KaTeX decides to do, we can easily avoid such problems.  `macros` is essentially the interpreter state.

Again, please start another issue/discussion if you'd like to continue discussing.

PR #1349, opened 2018-05-24
Export htmlBuilder from accent, horizBrace, and op for supsub.

marcianx:


src/functions/accent.js
```diff
@@ -153,25 +162,25 @@ const htmlBuilder = (group, options) => {
...
 
-        return supsubGroup;
+        return supSubGroup;
     } else {
         return accentWrap;
     }
 };
 
-const mathmlBuilder = (group, options) => {
+const mathmlBuilder: MathMLBuilder<"accent"> = (group, options) => {
```
marcianx: FYI: The reason for adding an explicit type in each of the `mathmlBuilder`s is to allow the type inference to succeed on `defineFunction` called later in this file.

Recall that `defineFunction` ties together the `type` argument with the output of `handler` and the input of `htmlBuilder` and `mathmlBuilder`. In this file, `htmlBuilder` is strictly more generic in this file (`HtmlBuilderSupSub` is a subtype of `HtmlBuilder` per the variance rules), but flow requires us to either *explicitly* cast it down from `HtmlBuilderSupSub` to `HtmlBuilder` to prevent type inference from failing, or, as it turns out, adding an explicit type for the `mathmlHandler` seems to be enough to satisfy it. IMO, none of this should matter since `defineFunction` explicitly specifies `type` which should be enough, to determine what the expected types of `handler`, `htmlBuilder`, and `mathmlBuilder` are, but it's still beyond flow's comfort zone.

src/ParseNode.js
```diff
@@ -290,13 +290,15 @@ type ParseNodeTypes = {
...
+    // The union allows either ParseNode<*> or the union of two specific nodes.
+    node: ?ParseNode<*> | ParseNode<*>,
     type: NODETYPE,
 ): ParseNode<NODETYPE> {
     const typedNode = checkNodeType(node, type);
     if (!typedNode) {
         throw new Error(
-            `Expected node of type ${type}, but got node of type ${node.type}`);
+            `Expected node of type ${type}, but got ` +
+            (node ? `node of type ${node.type}` : String(node)));
```
ylemkimon: `String(node)` can be shortend to `${node}`.

src/functions/accent.js
ylemkimon: Isn't `HtmlBuilder` a subtype of `HtmlBuilderSupSub`?

src/ParseNode.js
```diff
@@ -290,13 +290,15 @@ type ParseNodeTypes = {
  * typing. Throws if the node's type does not match.
  */
 export function assertNodeType<NODETYPE: NodeType>(
-    node: ParseNode<*>,
+    // The union allows either ParseNode<*> or the union of two specific nodes.
+    node: ?ParseNode<*> | ParseNode<*>,
```
ylemkimon: Shouldn't `ParseNode<A> | ParseNode<B>` be subtype of `ParseNode<*>`?

src/ParseNode.js
marcianx: That's what I started with, but the flow type system disallows it.

src/functions/accent.js
marcianx: Functions are contravariant on their input types. Floor actually had some nice documentation on it: https://flow.org/en/docs/lang/variance/

src/ParseNode.js
marcianx: That's what I originally thought, but then I remembered that `*` doesn't mean "anything". It means "something specific that I'd like the inference to figure out". The constraint in `ParseNode`'s type parameter is that it's one of the strings in `NodeType`. So it shouldn't be able to correspond to two possible values.

marcianx: More info here in the Existential type is here:
https://flow.org/en/docs/types/utilities/#toc-existential-type
Note that sometimes the type inference can't figure out what the Existential type should be since it does *global* type inference and in my experience kinda just constraints the possibilities that there type can have. That's why we were able to get away a lot more with just using `*` for the input if `htmlBuilder` and `mathmlBuilder` before we ported everything to flow. So I generally recommend always specifying explicit types whenever possible -- I've caught more type errors due to the increased strictness that way.

marcianx: More info here in the Existential type is here:
https://flow.org/en/docs/types/utilities/#toc-existential-type
Note that sometimes the type inference can't figure out what the Existential type should be since it does *global* type inference and in my experience kinda just constraints the possibilities that there type can have. That's why we were able to get away a lot more with just using `*` for the input if `htmlBuilder` and `mathmlBuilder` before we ported everything to flow. So I generally recommend always specifying explicit types whenever possible -- I've caught more type errors due to the increased strictness that way.

src/functions/accent.js
ylemkimon: Note to self: as `HtmlBuilderSupSub` accepts less specific input, it's subtype of `HtmlBuilder`.

src/ParseNode.js
ylemkimon: Note to self: `flow` doesn't allow implicit type coercion.

ylemkimon:
Thank you for clarifying a few things 😄 

What do you think about having null-ParseNode, i.e., `ParseNode<"null">` and returning it from `Parser.parseFunction()`, so that we can tighten maybe `ParseNode` types and replace null check with NodeType check? I think allowing `null` or `undefined` may be bug-prone.

src/functions/horizBrace.js
```diff
@@ -130,13 +135,7 @@ defineFunction({
...
     htmlBuilder,
-    mathmlBuilder(group, options) {
-        const accentNode = stretchy.mathMLnode(group.value.label);
-        return new mathMLTree.MathNode(
-            (group.value.isOver ? "mover" : "munder"),
-            [mml.buildGroup(group.value.base, options), accentNode]
-        );
-    },
+    mathmlBuilder,
 });
```
ylemkimon: Too many newlines.

marcianx: Removed.

marcianx:
@ylemkimon 
> What do you think about having null-ParseNode, i.e., ParseNode<"null"> and returning it from Parser.parseFunction(), so that we can tighten maybe-ParseNode types and replace null check with NodeType check (at least for base of supsub)? I think allowing null or undefined may be bug-prone.

Maybe-`ParseNode` types are useful and actually offer us *more* safety (see further below). To address your specific concern, we can't erroneously pass in a variable of type `?string` or `?number` when `?ParseNode<*>` is expected. The compiler will still complain. The only case we might erroneously pass in `null` and have the compiler accept it is if it were a literal `null` or `undefined`, or it was a variable of type `void`. I don't think these are issues we would run into.

Whereas, the `?Foo` notation has many benefits:

1.  `?Foo` captures the intent quite directly and is the canonical way in flow (and js) to indicate an optional type.
2. Related to 1, the type system can _more_ easily be used to distinguish cases when we are guaranteed to have a non-nullary `ParseNode` (`ParseNode<*>`), vs only optionally having it (`?ParseNode<*>`). If we added `ParseNode<"null">`, we wouldn't be able to do this as straightforwardly.
3. The compiler won't allow us to dereference `?ParseNode<*>` at all, thereby forcing a null check, whereas it would allow us to dereference `ParseNode<null>`.

ylemkimon:
Thank you for the detailed explanation!

src/ParseNode.js
kevinbarabash: Sorry for the late comment.  Isn't
```
node: ?ParseNode<*> | ParseNode<*>
```
equivalent to
```
node: ?ParseNode<*>
```
?

ylemkimon: @kevinbarabash As @marcianx said above, from what I understand, `*` doesn't mean "anything", rather it is inferred to one (at the compile-time). So `?ParseNode<*>` cannot accept `ParseNode<A> | ParseNode<B>` as `*` can be either A or B but it cannot be determined at the compile-time.

kevinbarabash:
```
node: ?ParseNode<*>
``` 
should be equivalent to 
```
node: ParseNode<*> | null | undefined
```
I would expect:
```
node: ?ParseNode<*> | ParseNode<*>
```
to be equivalent to:
```
node: ParseNode<*> | null | undefined | ParseNode<*>
```
but since there's two `ParseNode<*>` in the same union we should be able to drop one result in:
```
node: ParseNode<*> | null | undefined
```
which is equivalent to
```
node: ?ParseNode<*>
```

marcianx: To elaborate on ylemkimon@'s comment, if `*` meant "anything", then yes, `ParseNode<*>` is equivalent to `ParseNode<*> | ParseNode<*>` because each `ParseNode<*>` there is the *same*. But in flow, each `*` is an *existential* type, which means that *each* `*` refers to something in particular (in our case, it's constrainted to be a value from `NodeType`) that the flow type system has to infer. The two `*`s may be inferred to be *different* things as they are independent and not connected in any way, and thus the union is strictly more generic.

More concretely, let's say you are the flow type inference system and you see a parameter of type `ParseNode<*>`. If a caller passes in something of type `ParseNode<"accent"> | ParseNode<"supsub">`, what specific value is `*` inferred to be?
However, if the parameter has type `ParseNode<*> | ParseNode<*>`, then the first `*` can be inferred to be "accent", and the second one can be inferred to be "supsub".

PR #1357, opened 2018-05-25
Update function support page for 0.10.0-alpha

ronkok:


function-support.html
```diff
@@ -6,9 +6,9 @@
     <title>Function Support in KaTeX</title>
     <link rel="icon" href="http://khan.github.io/favicon.ico">
 	
-	<link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/katex@0.9.0/dist/katex.min.css">
-	<script src="https://cdn.jsdelivr.net/npm/katex@0.9.0/dist/katex.min.js"></script>
-	<script src="https://cdn.jsdelivr.net/npm/katex@0.9.0/dist/contrib/auto-render.min.js"></script>
+    <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/katex@0.10.0-alpha/dist/katex.min.css" integrity="sha384-BTL0nVi8DnMrNdMQZG1Ww6yasK9ZGnUxL1ZWukXQ7fygA1py52yPp9W4wrR00VML" crossorigin="anonymous">
+    <script src="https://cdn.jsdelivr.net/npm/katex@0.10.0-alpha/dist/katex.min.js" integrity="sha384-y6SGsNt7yZECc4Pf86XmQhC4hG2wxL6Upkt9N1efhFxfh6wlxBH0mJiTE8XYclC1" crossorigin="anonymous"></script>
+	<script src="https://cdnjs.cloudflare.com/ajax/libs/KaTeX/0.9.0/contrib/auto-render.min.js" integrity="sha384-IiI65aU9ZYub2MY9zhtKd1H2ps7xxf+eb2YFG9lX6uRqpXCvBTOidPRCXCrQ++Uc" crossorigin="anonymous"></script>
```
edemaine: Why did auto-render switch to cloudflare? We should probably use one source; I'd say `jsdelivr`.

function-support.html
```diff
@@ -507,10 +525,11 @@ <h2 id="environments">Environments</h2>
    c &amp; d
 \end{matrix}</code></pre>
             </td>
-            <td>\(\begin{array}{c|c} a & b \\ c & d \end{array}\)</td>
+            <td>\(\begin{array}{c|c} g & b \\ \hline c & d \end{array}\)</td>
             <td>
                 <pre><code>\begin{array}{c|c}
-   a &amp; b \\
+   g &amp; b \\
+   \hline
```
edemaine: Awesome.

function-support.html
```diff
@@ -593,7 +612,7 @@ <h2 id="environments">Environments</h2>
 	
         <p>Acceptable line separators include: <code>\\</code>, <code>\cr</code>, and <code>\cr[<em>distance</em>]</code>. <em>Distance</em> can be written with any of the <a href="#units">KaTeX units</a>.</p>
```
edemaine: I know you like `\cr` because it's less likely to cause issues, but it still seems worth mentioning `\\[distance]` too.

function-support.html
```diff
@@ -813,105 +832,70 @@ <h3 id="other-letters">Other Letters</h3>
...
+				<td>Monospace</td>
+				<td><span class="katex"><span class="mathtt">A-Z a-z 0-9</span></span></td>
 			</tr>
 		</tbody>
 	</table>
-	
+
+	<h4 id="letters-in-text">Letters inside \text{}</h4>
+		<p><code>\text{…}</code> will accept functions: <code>\aa</code> <code>\AA</code> <code>\ae</code> <code>\AE</code> <code>\oe</code> <code>\OE</code> <code>\o</code> <code>\O</code> <code>\ss</code> <code>\i</code> <code>\j</code><br>
+		… and return them as: \(\text{\aa\: \AA\: \ae\: \AE\: \oe\: \OE\: \o\: \O\: \ss\: \i\: \j}\)</p>
```
edemaine: I would have expected this to be laid out like "Other Letters" is (compare `\imath` with its text equivalent, `\i`).

function-support.html
```diff
@@ -813,105 +832,70 @@ <h3 id="other-letters">Other Letters</h3>
...
+
+	<h3 id="unicode">Unicode</h3>
+		<p>The letters listed above will render in any KaTeX rendering mode.</p>
```
edemaine: Maybe add
```, and should match XeLaTeX behavior with `\usepackage{unicode-math}` ```?  (or "approximately match")

File: function-support.html [hunk 6]
```diff
@@ -813,105 +832,70 @@ <h3 id="other-letters">Other Letters</h3>
...
+		
+		<p>If the KaTeX rendering mode is set to <code>strict: false</code> or <code>strict:"warn"</code> (default), then KaTeX will accept all Unicode letters.</p>
+		
+		<p>If a letter is not listed above, it may cause small vertical alignment issues. KaTeX has detailed metrics for glyphs in Latin, Greek, and Cyrillic, but other glyphs are treated as if they are each as tall as the letter <q>M</q>.</p>
```
edemaine: I would merge the above two paragraphs.  And maybe also stress that unlisted symbols won't be in the fonts, so they fall back to an unknown system font, and won't "look right", in addition to the metrics issue.

function-support.html
```diff
@@ -947,6 +931,21 @@ <h2 id="annotation">Annotation</h2>
 		</tr>
 	</table>
 	
+	<div class="indented">
+		<p><code>\tag{hi} x+y^{2x}</code></p>
+		<p>\[\tag{hi} x+y^{2x} \]</p>
+		<p><code>\tag*{hi} x+y^{2x}</code></p>
+		<p>\[\tag*{hi} x+y^{2x} \]</p>
+		<p><code>\tag</code> is not yet supported inside the <code>{gather}</code> environment.</p>
```
edemaine: `{gather}` is a top-level environment, which KaTeX doesn't support.  So maybe we should say `{gathered}`.  And should probably also list `{aligned}` and `{alignedat}`; I believe `amsmath` supports `\tag` in `equation`, `gather`, `multiline`, `split`, `align`, `alignat`, `ftalign`, but among those, we only have `gathered`, `aligned`, `alignedat`.

function-support.html
```diff
@@ -947,6 +931,21 @@ <h2 id="annotation">Annotation</h2>
...
+	<div class="indented">
+		<p><code>\tag{hi} x+y^{2x}</code></p>
+		<p>\[\tag{hi} x+y^{2x} \]</p>
+		<p><code>\tag*{hi} x+y^{2x}</code></p>
+		<p>\[\tag*{hi} x+y^{2x} \]</p>
+		<p><code>\tag</code> is not yet supported inside the <code>{gather}</code> environment.</p>
+	</div>	
+	
+<h3 id="line-breaks">Line Breaks</h3>
+	<p>KaTeX 0.10.0+ will insert automatic line breaks in inline math. These can be suppressed by placing math inside a pair of braces, as in <code>{F=ma}</code>.</p>
```
edemaine: Maybe add to the first sentence "after binary operators (like `+` and `\cdot`) and relations (like `=` and `<`) at the top level, like LaTeX".

function-support.html
```diff
@@ -947,6 +931,21 @@ <h2 id="annotation">Annotation</h2>
...
+	
+	<p>Hard line breaks are <code>\\</code> and <code>\newline</code>.</p>
+	
+	<p>In display math, KaTeX does not insert automatic line breaks. It ignores display math hard line breaks when rendering option <code>strict: true</code>.</p>
```
edemaine: Maybe add "to match LaTeX"?  Maybe this is getting too redundant though.  Feel free to ignore this comment in particular.

function-support.html
```diff
@@ -1195,7 +1195,7 @@ <h2 id="logic">Logic and Set Theory</h2>
 
 <h2 id="macros">Macros</h2>
 	<div class="indented">
-	<p>Before macros can be used, they must be defined in the KaTeX <a href="https://github.com/Khan/KaTeX#rendering-options">rendering options</a>. Available functions include:</p>
+	<p>Macros can be defined either via  a KaTeX <a href="https://github.com/Khan/KaTeX#rendering-options">rendering options</a> or by <code>\gdef</code>. Available functions include:</p>
```
edemaine: "a KaTeX rendering options" needs to be "KaTeX rendering options" or "a KaTeX rendering option".

function-support.html
```diff
@@ -1195,7 +1195,7 @@ <h2 id="logic">Logic and Set Theory</h2>
 
 <h2 id="macros">Macros</h2>
 	<div class="indented">
-	<p>Before macros can be used, they must be defined in the KaTeX <a href="https://github.com/Khan/KaTeX#rendering-options">rendering options</a>. Available functions include:</p>
+	<p>Macros can be defined either via  a KaTeX <a href="https://github.com/Khan/KaTeX#rendering-options">rendering options</a> or by <code>\gdef</code>. Available functions include:</p>
 		<div class="indented">
 		<p><code>\mathchoice \TextOrMath \@ifstar \@ifnextchar \@firstoftwo \@secondoftwo \relax</code></p>
 		<p><code>@</code> is a valid character for commands, as if <code>\makeatletter</code> were in effect.</p>
```
edemaine: The second paragraph (about `@`) is inside the `<div class="indented">` -- is this intentional?  Looked a little funny to me.

function-support.html
```diff
@@ -2692,13 +2692,13 @@ <h2 id="symbols">Symbols and Punctuation</h2>
 			<td>\(\triangleright\)</td>
```
edemaine: I think `--` and `---` need to be wrapped in `\text` too.  (In math mode, you get multiple minus signs.)

edemaine:
One other thing: we added support for nested `\(...\)` in addition to `$...$`, which could be mentioned in the Style section when mentioning nested `$...$`.

ronkok:
@edemaine Thank you for the review. Every time you've reviewed one of these, it has made the document better. I'll pick up the comments this evening, but first I want to ask you:

1.  I wrote `\gdef` into the macro section. Is that appropriate at this time or is it premature?
2.  Do you agree with the way I addressed the Arabic/Persian plugin?

edemaine:
`\gdef` did make it into the alpha, so it seems fine to mention as you did.

The Persian/Arabic plugin is pretty clear on it being a WIP.  But I think it might need #1269 merged before it would actually work with the current KaTeX, so maybe it's premature to list it here (maybe by beta time?).  @HosseinAgha could comment further.

function-support.html
ronkok: In my first draft, I had this sentence written in almost the exact words that you suggest. I cut it back because I'm trying to avoid explanation that just duplicates LaTeX documentation. When KaTeX behavior differs in subtle ways from LaTeX or MathJax, then I have to explain things, but otherwise this is a document that shows examples but does not explain. That way, I can fit the document onto a single HTML page.

Maybe I'll think of some very concise way to mention where the breaks will come. But if not, I'm inclined to let people learn that by observation or by looking up the LaTeX docs.

edemaine: Maybe just add "like LaTeX" then? I don't want them to think it adds them willy-nilly.  But I'm not sure most people know the rule, so it also wouldn't hurt.  (I certainly didn't!)

function-support.html
ronkok: > Why did auto-render switch to cloudflare?

Because cloudflare is what we still have written in the auto-render README, and I copied and pasted it without looking at it. You're right though, we should use `jsdelivr` and we should change that README.

function-support.html
ronkok: I promise that I'll put `\\[distance]` back in as soon as we fix auto-render.  I'm pretty sure that we'll make the next release.

ronkok:
@edemaine One more question. Previous versions of the function support page contained: 


Function | Produces 
-- | -- 
~ | non-breaking space 
\space | non-breaking space 
\nobreakspace | non-breaking space 
\space | non-breaking space

In your new line break code, I recall that you made `~` and `\nobreakspace` into no-break spaces. What about `\space` and `\<space>`? Do I need to change those?

ronkok:
Answering my own question:  I've done some testing. Neither LaTeX nor KaTeX's new line break code treats `\space` or `\<space>` as a no-break space. I had incorrect information in previous versions of the function support page.  I'll change it.

ronkok:
Don't merge this yet. I'm investigating something.

ronkok:
I have found this odd thing, and I'm not sure if I should raise a top-level issue.  When I first wrote the function support page's CSS, I wrote a rule:
```
.katex-html {margin-top: 0.3em; margin-bottom: 0.3em;}
```

By that means, I got some extra row spacing in the tables, but only where KaTeX had run a script, which is exactly what I wanted. In this release, that margin was ineffective and the rows were squashed together. The new line breaking code had removed a katex.less rule that previously said `.katex-html {display: inline-block;}` and instead made the first child of `.katex-html` into an `inline-block`. And that made my margin ineffective.

In my latest commit, I've augmented the rule in the function support page. It now reads:
```
.katex-html {
  display: inline-block;
  margin-top: 0.3em;
  margin-bottom: 0.3em;
}
```

You see, I reinstated the ` display: inline-block;` at the `.katex-html` level. And the row spacing is back as it was.

So I've found a breaking change, but it affects only people that have done something sneaky and underhanded. I think probably that the `.katex-html {display: inline-block;}` rule should be reinstated. It doesn't interfere with any line breaking.

How about it? Should I create a top level issue?

HosseinAgha:
@ronkok I think it would be better if we mention in the docs after the #1269 merge.   
Also I only support Persian formulas in the plugin. Arabic support is much more complicated and should be handled in another plugin as the formulas are right-to-left.

ronkok:
@HosseinAgha Thanks. I've commented out the sentence on your plug-in. We'll include it in a later release.

function-support.html
edemaine: Agreed, this is a significant bug -- they're also at 0.9.0.  I'll fix it in a separate PR.

ronkok:
This should now be ready to go.

function-support.html
```diff
@@ -8,7 +8,7 @@
 	
     <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/katex@0.10.0-alpha/dist/katex.min.css" integrity="sha384-BTL0nVi8DnMrNdMQZG1Ww6yasK9ZGnUxL1ZWukXQ7fygA1py52yPp9W4wrR00VML" crossorigin="anonymous">
     <script src="https://cdn.jsdelivr.net/npm/katex@0.10.0-alpha/dist/katex.min.js" integrity="sha384-y6SGsNt7yZECc4Pf86XmQhC4hG2wxL6Upkt9N1efhFxfh6wlxBH0mJiTE8XYclC1" crossorigin="anonymous"></script>
-	<script src="https://cdnjs.cloudflare.com/ajax/libs/KaTeX/0.9.0/contrib/auto-render.min.js" integrity="sha384-IiI65aU9ZYub2MY9zhtKd1H2ps7xxf+eb2YFG9lX6uRqpXCvBTOidPRCXCrQ++Uc" crossorigin="anonymous"></script>
+	<script src="https://cdn.jsdelivr.net/npm/katex@0.9.0/dist/contrib/auto-render.min.js"></script>
```
edemaine: Can you update to 0.10.0? You can now copy/paste from the auto-render README.

function-support.html
```diff
@@ -883,18 +903,14 @@ <h4>Unicode Mathematical Alphanumeric Symbols</h4>
...
-		… and return them as: \(\text{\aa\: \AA\: \ae\: \AE\: \oe\: \OE\: \o\: \O\: \ss\: \i\: \j}\)</p>
 
 	<h3 id="unicode">Unicode</h3>
 		<p>The letters listed above will render in any KaTeX rendering mode.</p>
 		
-		<p>If the KaTeX rendering mode is set to <code>strict: false</code> or <code>strict:"warn"</code> (default), then KaTeX will accept all Unicode letters.</p>
-		
-		<p>If a letter is not listed above, it may cause small vertical alignment issues. KaTeX has detailed metrics for glyphs in Latin, Greek, and Cyrillic, but other glyphs are treated as if they are each as tall as the letter <q>M</q>.</p>
+		<p>If the KaTeX rendering mode is set to <code>strict: false</code> or <code>strict:"warn"</code> (default), then KaTeX will accept all Unicode letters. The letters not listed above will be rendered from system fonts, not KaTeX-supplied fonts, so their typograhpy may clash. They may also cause small vertical alignment issues. KaTeX has detailed metrics for glyphs in Latin, Greek, and Cyrillic, but other glyphs are treated as if they are each as tall as the letter <q>M</q>.</p>
```
edemaine: Typo: typograhpy -> typography

function-support.html
```diff
@@ -1195,12 +1211,12 @@ <h2 id="logic">Logic and Set Theory</h2>
...
 <h2 id="macros">Macros</h2>
 	<div class="indented">
-	<p>Macros can be defined either via  a KaTeX <a href="https://github.com/Khan/KaTeX#rendering-options">rendering options</a> or by <code>\gdef</code>. Available functions include:</p>
+	<p>Macros can be defined either via  a KaTeX <a href="https://github.com/Khan/KaTeX#rendering-options">rendering option</a> or by <code>\gdef</code>. Available functions include:</p>
 		<div class="indented">
 		<p><code>\mathchoice \TextOrMath \@ifstar \@ifnextchar \@firstoftwo \@secondoftwo \relax</code></p>
-		<p><code>@</code> is a valid character for commands, as if <code>\makeatletter</code> were in effect.</p>
 		</div>
 	</div>
+	<p><code>@</code> is a valid character for commands, as if <code>\makeatletter</code> were in effect.</p>
```
edemaine: I think you moved it too far out -- now it doesn't have enough indent. Should be in one level but not two.

ronkok:
Done. Thanks for the comments. This should be good to go.

PR #1359, opened 2018-05-26
Move "supsub" handers into functions/supsub.js and add flow types.

marcianx:
Also simplified the return type of `utils.getBaseElem()`. See the comment on the last commit for why it can now be done.


src/functions/supsub.js
```diff
@@ -0,0 +1,239 @@
...
+            minSupShift = metrics.sup2;
+        }
+
+        // scriptspace is a font-size-independent size, so scale it
+        // appropriately for use as the marginRight.
+        const multiplier = options.sizeMultiplier;
+        const marginRight = (0.5 / metrics.ptPerEm) / multiplier + "em";
+
+        let supsub;
+        if (supm && subm) {
```
marcianx: The original code structure was:
```
if (!group.value.sup) {
  ...
} else if (!group.value.sub) {
  ...
} else {
  ...
}
```
As in many other refactors, this was reordered to make positive/existence checks (which requires start with the original `else` case where both are non-nullable) instead of negative/non-existence checks so that non-nullability of the `supm` and `subm` are immediately obvious to the type system.

src/buildMathML.js
```diff
@@ -10,7 +10,6 @@ import buildCommon from "./buildCommon";
 import fontMetrics from "./fontMetrics";
```
ylemkimon: I think we can remove above warning.

src/buildHTML.js
```diff
@@ -206,166 +206,13 @@ export const isLeftTight = function(node) {
     return node.hasClass("mtight");
```
ylemkimon: I think we can remove above warning, too.

src/buildHTML.js
```diff
@@ -206,166 +206,13 @@ export const isLeftTight = function(node) {
...
-
-        // Wrap the supsub vlist in a span.msupsub to reset text-align.
-        const mclass = getTypeOfDomTree(base) || "mord";
-        return makeSpan([mclass],
-            [base, makeSpan(["msupsub"], [supsub])],
-            options);
-    },
-};
+/** This is a map of group types to the function used to handle that type. */
+export const groupTypes = {};
```
ylemkimon: Nice!
What do you think about moving them to `src/defineFunction.js` like `_functions`?

marcianx: Honestly, not sure at all. Currently, `groupTypes` are used by both the `build*.js` files and the `defineFunction` and `defineEnvironment`. I honestly can't tell the difference/advantage in moving one place or the other. The only minor consideration that comes to mind is that if we move both `groupTypes` into `defineFunction`, we're going to have to come up with two different names for them since they would be in the same file, whereas now they are accessed as `mml.groupTypes` and `html.groupTypes` where they are used, which is kinda nice.

They really should be imported nowhere else, though I see that `functions/raisebox` requires it -- I can clean that up by having it require `size`'s `buildBuilder` directly. I'd do it in a separate PR, but it can also be in this one if you prefer.

src/buildHTML.js
marcianx: Done.

src/buildMathML.js
marcianx: Done.

src/functions/supsub.js
```diff
@@ -0,0 +1,239 @@
...
+        let supm;
+        let subm;
+
+        const metrics = options.fontMetrics();
+
+        // Rule 18a
+        let supShift = 0;
+        let subShift = 0;
+
+        const isCharacterBox = valueBase && !utils.isCharacterBox(valueBase);
```
ylemkimon: `!utils.isCharacterBox(valueBase)` -> `utils.isCharacterBox(valueBase)`

marcianx: Whoops! Fixed.

src/buildHTML.js
ylemkimon: Ah, it's also used in `defineEnvironment`. I though they are only defined in `defineFunction`.

ylemkimon: @marcianx Too late but what do you think about renaming them to `groupBuilders`?

marcianx: You know what, I'm definitely empathizing with this now (see the `raisebox` PR I'll send soon). All the `functions/*` generally import *all* the exports within `buildHTML` and `buildMathML`, but they really should not be accessing `groupTypes` at all. So you're right. I'll move this into `src/defineFunction` and not import them anywhere else.

PR #1361, opened 2018-05-26
 Correct (type-wise) raisebox's usage of sizing's buildHtml. 

marcianx:


src/functions/raisebox.js
```diff
@@ -28,16 +28,17 @@ defineFunction({
...
+            body: group.value.value,
+            font: "mathrm", // simulate \textrm
+        }, group.mode);
+        const sizedText = new ParseNode("sizing", {
+            type: "sizing",
+            value: [text],
             size: 6,                // simulate \normalsize
-        }}, options);
+        }, group.mode);
+        const body = sizing.htmlBuilder(sizedText, options);
```
ylemkimon: How about importing `sizingGroup` and
```js
const newOptions = options.havingSize(6); // simulate \normalsize
const body = sizingGroup([text], newOptions, options);
```
?

marcianx: That's true. The reason I did this was so that there's only one place the canonical implementation lives. That way, if it's ever updated in `sizing.js` we wouldn't have to remember to update it here as well. What do you think?

ylemkimon: @marcianx That seems right 😄

src/buildHTML.js
```diff
@@ -1,8 +1,8 @@
 /**
  * This file does the main work of building a domTree structure from a parse
  * tree. The entry point is the `buildHTML` function, which takes a parse tree.
- * Then, the buildExpression, buildGroup, and various groupTypes functions are
- * called, to produce a final HTML tree.
+ * Then, the buildExpression, buildGroup, and various groupBuilders functions
```
kevinbarabash: Remaining `groupTypes` to `groupBuilders` is an improvement.

src/functions/raisebox.js
```diff
@@ -28,16 +28,17 @@ defineFunction({
...
+            type: "text",
+            body: group.value.value,
+            font: "mathrm", // simulate \textrm
+        }, group.mode);
+        const sizedText = new ParseNode("sizing", {
+            type: "sizing",
+            value: [text],
             size: 6,                // simulate \normalsize
-        }}, options);
+        }, group.mode);
```
kevinbarabash: Nice use of `ParseNode` here.

src/buildHTML.js
marcianx: Thoroughly agree! That was @ylemkimon's idea in the last PR.

src/functions/raisebox.js
marcianx: It's required by the type system. ;)

PR #1362, opened 2018-05-26
Update contrib READMEs to jsdelivr v0.10.0-alpha

edemaine:
This is particularly important as update-sri.js only updates jsdelivr links.

I was able to get `update-sri.js` to do the heavy lifting (computing checksums) after updating the links in general, and downloading v0.10.0-alpha from CDN into `dist`.


kevinbarabash:
Hmm... these should've been updated as part of the release script.  Sorry for not checking them.

PR #1369, opened 2018-05-27
Update flow version and address type errors.

marcianx:


marcianx:
1. Yes, I'd like to have time to investigate possibilities for handling this better and remove the `$FlowFixMe`s.
2. These `$FlowFixMe`s do not negate the advantages! Once we assign to `_environment` and `_function`, we're deliberately making the type loose (all possible `NodeType`s need to be stored in each of these maps). All the type-safety we get is due to the types of `defineFunction` and `defineEnvironment`. Specifically, if you can call them successfully, then you're passing in handlers/builders of compatible types, and that is retained!

marcianx:
BTW, any of you all who like playing with the types, I'd appreciate any suggestions or workarounds on https://github.com/facebook/flow/issues/6369 to resolve these `$FlowFixMe`s.

marcianx:
Huh that's a good point. We would still need the `TypeValueMap` to be able to make `defineEnvironment` and `defineFunction` typesafe. But it would become a `TypeNodeMap` and we can simply use `$Values(TypeNodeMap)`. We would have to figure out something for `_environment` and `_function`, but we need to lose typesafety there anyway, so we could hack something.

Issue #1360, opened 2018-05-26
Spacing on wide characters is wrong.


ronkok:
Their span needs to be given a `mord`  class and that isn't happening. I'll take care of it.

![badspacing](https://user-images.githubusercontent.com/16403058/40578718-bfc75920-60cd-11e8-82f3-de7721fdd3a0.PNG)

edemaine:
Thanks @ronkok! Perhaps this will also fix #1154.

PR #1371, opened 2018-05-27
Fix wide character spacing

ronkok:
Fixes issue #1360.

Visual confirmation:
widespacing

### DIFF
--- a/src/buildCommon.js
+++ b/src/buildCommon.js
@@ -239,7 +239,7 @@ const makeOrd = function<NODETYPE: "spacing" | "mathord" | "textord">(
         // surrogate pairs get special treatment
         const [wideFontName, wideFontClass] = wideCharacterFont(value, mode);
         return makeSymbol(value, wideFontName, mode, options,
-			["mord", wideFontClass]);
+            classes.concat(wideFontClass));
     } else if (fontOrFamily) {
         let fontName;
         let fontClasses;
--- a/src/buildCommon.js
+++ b/src/buildCommon.js
@@ -238,7 +238,8 @@ const makeOrd = function<NODETYPE: "spacing" | "mathord" | "textord">(
     if (value.charCodeAt(0) === 0xD835) {
         // surrogate pairs get special treatment
         const [wideFontName, wideFontClass] = wideCharacterFont(value, mode);
-        return makeSymbol(value, wideFontName, mode, options, [wideFontClass]);
+        return makeSymbol(value, wideFontName, mode, options,
+			["mord", wideFontClass]);
     } else if (fontOrFamily) {
         let fontName;
         let fontClasses;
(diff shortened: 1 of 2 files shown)


test/katex-spec.js
```diff
@@ -1548,6 +1548,11 @@ describe("An HTML font tree-builder", function() {
         expect(markup).toContain(span);
     });
 
+    it("should render wide characters with mord and with the correct font", function() {
+        const markup = katex.renderToString(String.fromCharCode(0xD835, 0xDC00));
+        expect(markup).toContain("<span class=\"mord mathbf\">A</span>");
+    });
```
kevinbarabash: We might also want a test like
```
expect("M = a + b").toBuildLike("\textbf{M = a + b}");
```
where "M = a + b" is the bolded versions.  Not sure the best way to store a string with bolded characters.

ronkok: I just tried something equivalent to:
```
expect("𝐀 = 𝐚").toBuildLike("\\mathbf{A} = \\mathbf{a}");
```

It turns out that they don't build in an equivalent way. `\mathbf{A}` returns a `mord` span wrapped inside a `mord mathbf` span. "𝐀" returns a single `mord mathbf` span. They render in an equivalent way, but the test doesn't like it.

ronkok:
> This should fix the italic, script, etc. unicode chars right?

Yep. Anything with a high surrogate `\uD835`.

test/katex-spec.js
edemaine: Can you test
```js
expect("𝐀 = 𝐚").toBuildLike("\\mathbf A = \\mathbf a");
```
?

ronkok: @edemaine That one works! Cool!

src/buildCommon.js
```diff
@@ -238,7 +238,8 @@ const makeOrd = function<NODETYPE: "spacing" | "mathord" | "textord">(
     if (value.charCodeAt(0) === 0xD835) {
         // surrogate pairs get special treatment
         const [wideFontName, wideFontClass] = wideCharacterFont(value, mode);
-        return makeSymbol(value, wideFontName, mode, options, [wideFontClass]);
+        return makeSymbol(value, wideFontName, mode, options,
+			["mord", wideFontClass]);
```
ylemkimon: How about `classes.concat(wideFontClass)`?

ronkok: @ylemkimon Yes, that will work. Is that faster?

kevinbarabash: The benefit is that if someone adds something to `classes` they don't have to also update this line.

ronkok: Ah, that makes sense. Thank you for pointing it out.

PR #1373, opened 2018-05-28
Make ParseNodeTypes more regular

kevinbarabash:
I added a `type` field to all of the object types cleaned up some other issues.  I'm going to try updating things like `bin` and `close` which are typed as `string` to be typed as 
```
{|
   type: "bin",
   value: string,
|}
```
This should help to eventually make `type ParseNode` eventually be one big disjoint union.

### DIFF
--- a/src/ParseNode.js
+++ b/src/ParseNode.js
@@ -38,16 +38,8 @@ export default class ParseNode<TYPE: NodeType> {
 export type NodeType = $Keys<ParseNodeTypes>;
 export type NodeValue<TYPE: NodeType> = $ElementType<ParseNodeTypes, TYPE>;
 
-export type AccentStructType = {|
-    type: "accent" | "accentUnder",
-    label: string,
-    isStretchy?: boolean,
-    isShifty?: boolean,
-    base: ParseNode<*>,
-|};
-
 export type LeftRightDelimType = {|
-    type?: "leftright",
+    type: "leftright",
     body: ParseNode<*>[],
     left: string,
     right: string,
@@ -62,14 +54,6 @@ type ParseNodeTypes = {
         value: ParseNode<*>[],
     |},
     "color-token": string,
-    "leftright": {|
-        body: [{|
-            type: "array",
-            hskipBeforeAndAfter: boolean,
-        |} | ParseNode<*>],
-        left: string,
-        right: string,
-    |},
     // To avoid requiring run-time type assertions, this more carefully captures
     // the requirements on the fields per the op.js htmlBuilder logic:
     // - `body` and `value` are NEVER set simultanouesly.
@@ -99,11 +83,13 @@ type ParseNodeTypes = {
         value: ParseNode<*>[],
     |},
     "supsub": {|
+        type: "supsub",
         base: ?ParseNode<*>,
         sup?: ?ParseNode<*>,
         sub?: ?ParseNode<*>,
     |},
     "tag": {|
+        type: "tag",
         body: ParseNode<*>[],
         tag: ParseNode<*>[],
     |},
@@ -114,6 +100,7 @@ type ParseNodeTypes = {
     |},
     "url": string,
     "verb": {|
+        type: "verb",
         body: string,
         star: boolean,
     |},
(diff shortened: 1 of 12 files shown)


src/ParseNode.js
```diff
@@ -62,14 +54,6 @@ type ParseNodeTypes = {
         value: ParseNode<*>[],
     |},
     "color-token": string,
-    "leftright": {|
```
kevinbarabash: It appears twice in the type.  The first one here was being ignored.

src/functions/kern.js
```diff
@@ -39,7 +39,7 @@ defineFunction({
         }
         return {
             type: "kern",
-            dimension: args[0].value,
+            dimension: args[0].value.value,
```
kevinbarabash: The types for `args` weren't enough to pick that this needed to change.

test/katex-spec.js
```diff
@@ -2823,7 +2824,7 @@ describe("Unicode", function() {
     });
 
     it("should parse more relations", function() {
-        expect("⊂⊃⊆⊇⊏⊐⊑⊒⊢⊣⊩⊪⊸⋈⋍⋐⋑⋔⋛⋞⋟⌢⌣⩾⪆⪌⪕⪖⪯⪰⪷⪸⫅⫆≘≙≚≛≝≞≟").toBuild();
```
ylemkimon: The test case seems to be mistakenly changed.

kevinbarabash:
I started converting `"bin"`, `"rel"`, etc. but there were too many errors.  We should add `// @flow` to buildHTML.js and buildMathML.js first.

src/functions/kern.js
marcianx: As the first line in this function, could you do: `const size = assertNodeType(args[0], "size");` and use that instead of `args[0]`? That will capture any type errors in its usage very precisely.

marcianx:
I've got in in-progress PR to convert `buildHTML` and `buildMathML`. I'd gotten distracted by the `AnyParseNode` thing (union of all `ParseNode` types), though.

kevinbarabash:
Looking at the way we handle `bin`, `close`, `inner`, etc. I think we should roll these into a single group type called "atom".  The `ParseNodeType` would then look like:
```
...
"atom": {|
   type: "atom",
   kind: "ord" | "op" | "bin" | "rel" | "open" | "close" | "punt" | "inner" | "over" | "under" | "acc" | "rad",
   value: string,
|},
...
```
The list of atom kinds comes from page 158 of the TeXBook.

Also on the page is the follow prose:
> An atom’s nucleus, superscript, and subscript are called its fields, and there are four possibilities for each of these fields. A field can be
> - empty;
> - a math symbol (specified by family and position number); a box; or
> - a math list.

Eventually we might have an `atom` type that looks like:
```
type Atom = {|
   kind: "ord" | "op" | "bin" | "rel" | "open" | "close" | "punt" | "inner" | "over" | "under" | "acc" | "rad",
   nucleus: null | Symbol | Box | Atom[],
   sup: null | Symbol | Box | Atom[],
   sub: null | Symbol | Box | Atom[],
|},
```
This is a larger change but would bring our internals close to TeX internals.

marcianx:
I've been trying to replace `ParseNode<*>` with an `AnyParseNode` (full type union) as that seems to be a necessary step to typing this properly and I've run into a case where the *first* `"atom"` refactor you're suggesting would make things quite a bit easier. I think I'm going to `$FlowFixMe` and get on with it.

src/functions/rule.js
```diff
@@ -20,8 +20,8 @@ defineFunction({
         return {
             type: "rule",
             shift: shift && shift.value,
-            width: width.value,
-            height: height.value,
+            width: width.value.value,
+            height: height.value.value,
```
marcianx: @kevinbarabash Sorry I missed this, but this is actually incorrect!
In replacing the use of the `ParseNode<*>` type with (the actually correct) `AnyParseNode`, the compiler caught this!
PR will come soon-ish...with various flow type fixes.

kevinbarabash: Whoops.  Thanks for spotting that.

marcianx: And actually, I confused myself. While these two lines were correct, shift.value should also have had a `.value` added on to it.

PR #1374, opened 2018-05-28
Enable greenkeeper-lockfile

ylemkimon:
* Enabled `greenkeeper-lockfile`

@kevinbarabash Could you create a dedicated account(or use @khanbot, we can use user's token but it  grants access to entire account's repo), grant it a push access to `Khan/KaTeX`, create a [GitHub access token](https://github.com/settings/tokens) with `public_repo` permission, and add it to [TravisCI environment variable settings](https://travis-ci.org/Khan/KaTeX/settings) ([TravisCI doc](https://docs.travis-ci.com/user/environment-variables/#Defining-Variables-in-Repository-Settings)) as `GH_TOKEN`? Note that it will not be visible in build logs and it will not be available to pull request builds.


kevinbarabash:
I'd prefer not to use user tokens.  Is there some way to scope the token to a specific repo?

ylemkimon:
@kevinbarabash Unfortunately, there is no way to limit personal access token’s access, but one can create a bot-dedicated account, which only have push access to one repository.

There is [SSH deploy key](https://developer.github.com/v3/guides/managing-deploy-keys/#deploy-keys), which grants access to single repo. However, since private key cannot be stored in the environment variable, [file encryption and decryption](https://oncletom.io/2016/travis-ssh-deploy/) is required. Also `greenkeeper-lockfile` doesn’t officially support SSH keys, but it’ll probably work.

kevinbarabash:
@ylemkimon I see.  The SSH deploy key looks complicated.  I think I'll create a separate account just for KaTeX repos.

edemaine:
FWIW, it'd be great to get the new `package-lock.json` merged, as `npm install` is always insistent on changing it instead of respecting it. (Makes me wonder why we check it in at all...)

ylemkimon:
@edemaine I've created a separate PR for updating `package-lock.json`: #1378.

kevinbarabash:
@ylemkimon sorry for not creating a token yet.  Should we do something similar for the yarn.lock file?

ylemkimon:
@kevinbarabash Yes, I think so. FYI, Circle CI provides an easy way to use SSH keys, i.e., use the deploy key instead of account token.

PR #1375, opened 2018-05-28
Fix auto-render \\[…] bug

ronkok:
Fixes issue #1075.

Currently, a row separator such as `\\[0.3em]` will cause the auto-render extension to fail if the row separator is placed between `\(…\)` math delimiters.

This PR re-orders the array of left delimiters that get searched for, putting `\[` last in the list.  Then, `\(` is found before `\\[0.3em]` and the row delimiter is no longer mistaken for a math delimiter.


ronkok:
Update. Now I have done a local build from the source code and tested the result. It works fine and it does in fact resolve the `\\[` bug when inside `\(  \)` delimiters.

PR #1377, opened 2018-05-28
Support Unicode ⟂ ¬ ⊨ ‼ ∌

ronkok:
This is the fifteenth and possibly the last PR in my Unicode series. Today’s work includes five characters.

  

| **Character** | **Code Point** | **Unicode name**           | **Maps to**                  | **atom type** | **agrees with unicode math** | **Comment**                                                                              |
|:--------------|:---------------|:---------------------------|:-----------------------------|:--------------|:-----------------------------|:-----------------------------------------------------------------------------------------|
| ⟂             | U+27C2         | PERPENDICULAR              | \\perp                       | rel           | ✓                            | See below                                                                                |
| ¬             | U+00Ac         | NOT                        | \\neg                        | ord           | ✓                            |                                                                                          |
| ⊨             | U+22A8         | TRUE                       | \\vDash                      | rel           | ✓                            | KaTeX fonts have no glyph at ⊧, U+22A7, MODELS                                           |
| ‼             | U+203C         | DOUBLE EXCLAMATION MARK    | \\mathclose{!\\mkern-0.8mu!} | close         | No                           | unicode math got this one wrong. They called it an ord. It should be a [double factorial](https://www.johndcook.com/blog/2010/09/21/variations-on-factorial/). |
| ∌             | U+220C         | DOES NOT CONTAIN AS MEMBER | \\notni                      | rel           | almost                       | unicode math maps to \\nni                                                               |

  

Unicode character ⟂, U+27C2, PERPENDICULAR, has a clear mapping to `\perp`. Unicode math uses that mapping, which makes sense.

But there is also Unicode character ⊥, U+ U+22A5, UP TACK, and now things get awkward. KaTeX has a glyph at U+22A5 and uses it to render `\perp`. But unicode math maps that character to `\bot`, which is a ord, not a relation. I’ve elected to include the character with the unambiguous mapping and not to include U+22A5 among the characters in this PR.


src/macros.js
```diff
@@ -233,6 +233,11 @@ defineMacro("\u225D", "\\stackrel{\\tiny\\mathrm{def}}{=}");
 defineMacro("\u225E", "\\stackrel{\\tiny\\mathrm{m}}{=}");
 defineMacro("\u225F", "\\stackrel{\\tiny?}{=}");
 
+// Misc Unicode
+defineMacro("\u27C2", "\\perp");
+defineMacro("\u27C2", "\\mathclose{!\\mkern-0.8mu!}");
+defineMacro("\u27C2", "\\notni");
```
ylemkimon: All codepoints point to `27C2`

ronkok: Well that's embarrassing. @ylemkimon thank you for catching that.  Fixed now.

ronkok: And for corroboration, here's a snapshot.
u1

test/katex-spec.js
```diff
@@ -2893,7 +2893,7 @@ describe("Unicode", function() {
     });
 
     it("should parse symbols", function() {
-        expect("£¥ðℂℍℑℓℕ℘ℙℚℜℝℤℲℵℶℷℸ⅁∀∁∂∃∇∞∠∡∢♠♡♢♣♭♮♯✓°\u00b7").toParse();
+        expect("£¥ðℂℍℑℓℕ℘ℙℚℜℝℤℲℵℶℷℸ⅁∀∁∂∃∇∞∠∡∢♠♡♢♣♭♮♯✓°¬‼\u00b7").toParse();
```
edemaine: It occurs to me that these tests should be done in strict mode.

ronkok: @edemaine  The same thing occurred to me.

edemaine: @ronkok Do you want to take a crack at this?  There are probably many of these tests.  You should just need to replace `toParse()` or `toBuild()` with `toParse(strictSettings)` or `toBuild(strictSettings)`.

ronkok: Sure. I can get to it later this week.

PR #1378, opened 2018-05-29
Update package-lock.json

ylemkimon:


kevinbarabash:
It's possible that in some cases that changes to package-lock.json file weren't being checked in.  Perhaps we should fail the build if `package-lock.json` changes after travis does `npm install`. 🤔

edemaine:
Is there some way to `npm install` according to `package-lock.json` instead of `package.json`?

kevinbarabash:
After reading https://docs.npmjs.com/files/package-lock.json#description it sounds like package-lock.json is only useful for intermediate deps as opposed to the deps that appear in package.json.  We could try pinning deps in package.json and see if that helps.

PR #1379, opened 2018-05-29
Fix ligatures vs. \tt

### LINKED ISSUE #1367: Incorrect rendering of multiple - in \texttt
Try
```
\texttt{a---b}
```
In LaTeX the --- will appear as 3 individual "-". In KaTeX they will get merged into a single dash, which shouldn't happen for such a monospace font.

### LINKED ISSUE #1380: \texttt MathML doesn't mention monospace
`\texttt{...}` doesn't generate `mathvariant="monospace"`.  ~~I probably broke this with #1338.~~

```js
katex.renderToString('\\text{\\tt t tt}\\mathtt{mtt}\\texttt{ttt}')
```
produces
```html
...
<span class="katex-mathml">
  <math>
    <semantics>
      <mrow>
        <mtext mathvariant="monospace">t tt</mtext>
        <mrow>
          <mi mathvariant="monospace">m</mi><mi mathvariant="monospace">t</mi><mi mathvariant="monospace">t</mi>
        </mrow>
        <mtext>ttt</mtext>
      </mrow>
    ...```

Only the last one is incorrect.

### PR DESCRIPTION
**Pull request by edemaine**
1. Fix #1367: ligatures shouldn't happen in `\texttt{...}` and `\text{\tt ...}`.  Because we don't know the font at parse time, we leave the ligatures in, and then turn them back into their textual contents if we discover the right font and that the symbol is in a new `symbols` table of `ligatures`.  Alas, this has to be done in both HTML and MathML, but it's relatively straightforward.
2. Fix ligatures not always triggering, by putting them in `buildExpression` instead of `buildGroup`.  For example, `\text{\rm --}` did not get a ligature.
3. Fix #1380: convert `\text...` font commands into `<mtext mathvariant="..."` as best as we can.  Also fix `<mtext>` combining code to require matching `mathvariant`s.

### CODE REVIEW

File: src/buildCommon.js [hunk 1]
```diff
@@ -260,9 +260,19 @@ const makeOrd = function<NODETYPE: "spacing" | "mathord" | "textord">(
                                             options.fontShape);
             fontClasses = [fontOrFamily, options.fontWeight, options.fontShape];
         }
+
         if (lookupSymbol(value, fontName, mode).metrics) {
             return makeSymbol(value, fontName, mode, options,
                 classes.concat(fontClasses));
+        } else if (ligatures.hasOwnProperty(value) &&
+                   fontName.substr(0, 10) === "Typewriter") {
+            // Deconstruct ligatures in monospace fonts (\texttt, \tt).
```
kevinbarabash: The reason why we deconstruct them instead of not constructing them in the first place is that we don't actually know what the current font is.  I wonder if we could form the ligatures in ordgroup.js instead.  That would avoid having to undo work.

edemaine: I thought about doing it in `buildExpression` and it might be possible, but it's worrying to me. For example, `-{}-` does not make a ligature, like `{-}-`. In the current parse tree, these can be distinguished from `--`, but with any optimization (such as what we do in MathML's buildExpression), they might not. It makes sense for ligatures to be at the lexing level, and doing it in the parser approximates this. I also don't imagine people use texttt too often, so it shouldn't cost much time. But I could also try it the other way.

edemaine: I read the TeXbook some about this.  One point is that `\lq\lq` makes a ligature, so it's good we do this after macro expansion, not in the lexer.

I believe one "right" way to do this would be if the parser already knew what font we were in.  This is difficult with the current separation between macros and functions, but maybe it could be done.  Definitely more effort though.

I think it'd be possible to do in `buildExpression`, but it'd have to be done twice, once for MathML and once for HTML.  Should I explore this further?

Issue #1320, opened 2018-05-16
Allow configuration of line height (\baselinestretch)


clarfon:
Right now, the line height of generated data is hard-coded to 1.2. Unfortunately, the `lh` unit is far off, so, it would likely have to be a setting on the `katex.render` function.

It would be desirable to have this so that the line height of equations could match the line height of their surrounding text. This already happens for inline equations, but doesn't if the equations span multiple lines.

edemaine:
Line height is (I believe) controlled by CSS. Did you try defining a custom CSS rule to override tue built-in one, like
```css
.katex { line-height: 1.5 !important; }
```
?

I agree that it would be nice to make this more naturally configurable. This is essentially asking for `\baselineskip`/`\baselinestretch` as part of #687.

clarcharr:
@edemaine You can change the line height for `katex` and it will actually affect the text, if it takes up only one line. Otherwise, text which takes up multiple lines will always have lines `1.2em` tall, as that's factored into the spacing that KaTeX does normally.

So, to clarify, setting line-height for `katex` does work, and setting font-size works too because everything is in terms of `em`. But because everything is in terms of em, there's actually no way of configuring line-height for multiline content via `katex-display`.

edemaine:
Ah I think I see. Can you provide an example of something that ought to be (i.e. in LateX) is more vertically spaced when the line height is larger?

ronkok:
@clarcharr  We don't currently have an an equivalent to LaTeX `\renewcommand{\arraystretch}{1.5}`, which would change the line spacing of an entire environment. But you can set extra environment line spacing on a line by line basis, using `\cr[0.3em]` as your line separator.

edemaine:
By the way, `\gdef\arraystretch{1.5}` will be pretty easy to support once we have #1309.

infusion:
I would like to see this feature as well, especially with auto-render extracting the line height from the parent DOM node to apply it to multi-line blocks like `array` or `equation`. @edemaine I changed the line-height attribute of `.katex` in the web developer tools on site and it appears to leave exponentials and subscripts of symbols behind, even if it principally works.
Implementing arraystretch could make sense in a second step but in contrast to LaTex, in the web world the renderer should obey the main document. 
Adding line height as a config option as suggested by the OP plus making it set by the auto-renderer should be the better option in my opinion.

clarfon:
So, what exactly is the current progress on this?

kevinbarabash:
[...]

PR #1381, opened 2018-05-29
Support \arraystretch as a macro definition

edemaine:
With this PR, `\arraystretch` can be set via `settings.macros` or using `\def` or `\gdef`.  This was pretty easy now that we have `\def` and `\gdef`, as `arraystretch` was already supported internally.

This partially addresses #1320 (which is really asking for `\baselinestretch`).  `\arraystretch` was also mentioned in #687, though that issue is more about actual lengths, whereas `\arraystretch` is a macro not a length.

I did need to write two helpers: `expandMacro` and `expandMacroAsText` in `MacroExpander`.  These let a function or environment like `{array}` get the current value of `\arraystretch`, fully expanded, and use it to compute a number.  A cool example:

image


src/MacroExpander.js
```diff
@@ -249,6 +249,40 @@ export default class MacroExpander implements MacroContextInterface {
         throw new Error(); // eslint-disable-line no-unreachable
     }
 
+    /**
+     * Fully expand the given macro name and return the resulting list of
+     * tokens, or return `undefined` if no such macro is defined.
+     */
+    expandMacro(name: string): Token[] | void {
```
ylemkimon: Return type can be `?Token[]` (same for below)

edemaine: They are technically different: `?Token[]` includes `null` and `undefined`, while `Token[] | void` includes just `undefined`.  (I learned the `| void` trick looking elsewhere at this codebase.)  I agree the former is simpler looking, while the latter is more precise...  Do you still think I should change it? (Also happy to hear any other votes.)

kevinbarabash: I think it's fine as is.  I find `?Token[]` confusing b/c it's hard to tell whether it means `?(Token[])` or `(?Token)[]`.

PR #1382, opened 2018-05-29
\newcommand, \renewcommand, \providecommand

### LINKED ISSUE #250: add support for macros via \def and \newcommand

edemaine:
Basic support for `\newcommand` (fix #250), `\renewcommand`, and `\providecommand` (which doesn't care whether the command is defined):
```latex
\newcommand{\macro}{definition}
\newcommand\macro{definition}
\newcommand\macro[args]{definition}
% and the \renewcommand and \providecommand variants
```
Definition in this PR: A command is "defined" in KaTeX if it's a macro or a function or (added) a symbol.

Still to do for the future is to define optional arguments via `\newcommand`'s "default" argument: `\newcommand{\macro}[args][default]{definition}`.

I also had to revamp the testing infrastructure a bit to generate a new `Settings` object; otherwise, `settings.macros` gets contaminated between tests.  I removed `defaultSettings` but instead created `nonstrictSettings`.  This will impact #1336 in hopefully a minor way.


kevinbarabash:
@edemaine awesome stuff!

> Definition in this PR: A command is "defined" in KaTeX if it's either a macro or a function.

Why not check in `symbols` as well?

ylemkimon:
> I also had to revamp the testing infrastructure a bit to generate a new Settings object; otherwise, settings.macros gets contaminated between tests.

How about using `Object.freeze`? It prevents new properties from being added to it and prevents existing properties from being removed or changed, without an exception in non-strict mode. But I'm not sure how much overhead reusing `Settings` object will reduce.

edemaine:
@kevinbarabash Good idea! I should check symbols too.

@ylemkimon I'm not sure that would really work... The macros object needs to be changable in all tests involving `\def`s. We could plausibly just create new `Settings` objects for those tests, but my sense is that this is a pretty minor overhead.

ylemkimon:
Does this work with `\limits`, `\nolimits`, and `\hline`?

edemaine:
@ylemkimon What do you mean? (maybe a specific example?) This is just an interface to the macro system, which should handle that all very well...

ylemkimon:
I meant as they are directly handled by `Parser.js` and `environment.array.js`, they are probably not in `functions` and it may not expand macros before parsing them.(I haven’t checked) So I was wondering, does  redefining them or using them in the macro definition work.

edemaine:
@ylemkimon Ah, I see! Yes, that's an issue.  I added a function for `\hline` that throws an exception, similar to other failing functions we have elsewhere in the codebase, but this didn't work for `^`, `_`, `\limits`, and `\nolimits` which already have error handling within `Parser.js`.  So I added a list of `implicitCommands` (name up for debate) to catch them.

ylemkimon:
There seems to be also `\relax` in `MacroExpander.js`

src/macros.js
```diff
@@ -157,6 +175,57 @@ defineMacro("\\global", (context) => {
...
+    if (arg.length !== 1) {
+        throw new ParseError(
+            "\\newcommand's first argument must be a macro name");
+    }
+    const name = arg[0].text;
+
+    // A command is "defined" in KaTeX if it's a macro, function, or symbol.
+    const exists = context.macros.get(name) || functions.hasOwnProperty(name) ||
+        symbols.math.hasOwnProperty(name) || symbols.text.hasOwnProperty(name)
+        || implicitCommands.hasOwnProperty(name);
```
ylemkimon: I believe this will be useful in many ways, even as a public API! Could you move this to another class or a separate module?

edemaine: Good idea!  It's difficult to make it a public API in `katex` because it depends on the current state of the macros (i.e. the current group).  So I've made it a method of the `MacroExpander` for now, so at least it's available to all macro definitions (defined by JavaScript functions).

src/MacroExpander.js
```diff
@@ -15,6 +17,16 @@ import type {MacroContextInterface, MacroDefinition, MacroExpansion}
...
 import type Settings from "./Settings";
 
+// List of commands that act like macros but aren't defined as a macro,
+// function, or symbol.  Used in `isDefined`.
+export const implicitCommands = {
+    "\\relax": true,     // MacroExpander.js
+    "^": true,           // Parser.js
+    "_": true,           // Parser.js
+    "\\limits": true,    // Parser.js
+    "\\nolimits": true,  // Parser.js
```
kevinbarabash: Thanks for commenting where these are defined.

src/MacroExpander.js
```diff
@@ -282,5 +294,19 @@ export default class MacroExpander implements MacroContextInterface {
...
+     * functionality), meaning that it's a macro (in the current group),
+     * a function, a symbol, or one of the special commands listed in
+     * `implicitCommands`.
+     */
+    isDefined(name: string): boolean {
+        return this.macros.get(name) != null ||
+            functions.hasOwnProperty(name) ||
+            symbols.math.hasOwnProperty(name) ||
+            symbols.text.hasOwnProperty(name) ||
+            implicitCommands.hasOwnProperty(name);
```
kevinbarabash: nice

src/Namespace.js
```diff
@@ -56,7 +56,10 @@ export default class Namespace<Value> {
     }
 
     /**
-     * Get the current value of a name.
+     * Get the current value of a name, or `undefined` if there is no value.
+     * Note: Do not use `if (namespace.get(...))` to detect whether a macro
+     * is defined, as the definition may be the empty string which evaluates
+     * to `false` in JavaScript.  Use `if (namespace.get(...) != null)`.
```
kevinbarabash: Can you add a `namespace.has(name)` method that encapsulates this logic?

src/environments/array.js
```diff
@@ -617,3 +618,17 @@ defineEnvironment({
     htmlBuilder,
     mathmlBuilder,
 });
+
+// Catch \hline outside array envirnoment
+defineFunction({
+    type: "text", // Doesn't matter what this is.
```
kevinbarabash: Why doesn't it matter?

src/macros.js
```diff
@@ -157,6 +172,54 @@ defineMacro("\\global", (context) => {
...
+    let arg = context.consumeArgs(1)[0];
+    if (arg.length !== 1) {
+        throw new ParseError(
+            "\\newcommand's first argument must be a macro name");
+    }
+    const name = arg[0].text;
+
+    const exists = context.isDefined(name);
+    if (exists && !existsOK) {
+        throw new ParseError(`Command ${name} already defined in \\newcommand`);
```
kevinbarabash: This error message is a little confusing, maybe:
```
\\newcommand{${name}}{...} is trying to redefine ${name} which is already defined.
```

src/macros.js
```diff
@@ -157,6 +172,54 @@ defineMacro("\\global", (context) => {
...
+    }
+    if (!exists && !nonexistsOK) {
+        throw new ParseError(
+            `Command ${name} not already defined in \\renewcommand`);
```
kevinbarabash: This error message is a little confusing as well, maybe:
```
\\renewcommand{${name}}{...} is trying to redefine a command which doesn't exist.
```
Feel free to wordsmith these a bit more.  The main confusion for me was the "in" in "defined in \\(re)?newcommand".

test/katex-spec.js
```diff
@@ -2757,6 +2761,57 @@ describe("A macro expander", function() {
...
+
+    it("\\newcommand is local", () => {
+        expect("\\newcommand\\foo{1}\\foo{\\renewcommand\\foo{2}\\foo}\\foo")
+            .toParseLike("1{2}1");
+    });
+
+    it("\\newcommand accepts number of arguments", () => {
+        compareParseTree("\\newcommand\\foo[1]{#1^2}\\foo x+\\foo{y}",
+                         "x^2+y^2");
+        compareParseTree("\\newcommand\\foo[10]{#1^2}\\foo 0123456789", "0^2");
```
kevinbarabash: I took me a while to grok what's going on here.  `\foo` is eating all 10 of the args but is only using the first one, right?

test/katex-spec.js
```diff
@@ -2829,7 +2884,7 @@ describe("Unicode accents", function() {
             "\\tilde n" +
             "\\grave o\\acute o\\hat o\\tilde o\\ddot o" +
             "\\grave u\\acute u\\hat u\\ddot u" +
-            "\\acute y\\ddot y");
+            "\\acute y\\ddot y", nonstrictSettings);
```
kevinbarabash: I noticed you added `nonstrictSettings` in a couple of places.  Does this modify the behaviour fo the test?  How is it related to the `\newcommand`, `\renewcommand`, etc. part of this PR?

src/Namespace.js
edemaine: Added, and used in `isDefined`, but I'd hesitate to use it in e.g. `_getExpansion` because it'd double the number of lookups.

src/environments/array.js
edemaine: Because there's no `handler` defined, so the type isn't used.  I copy/pasted this comment from `functions/math.js`.

test/katex-spec.js
edemaine: Before, we used `{strict: false}` in all tests.  Now we use the default settings `{}` in all tests by default, which is more natural.  A few tests needed to have `{strict: false}` specified, though, and I added `nonstrictSettings` to these.

src/macros.js
edemaine: Edited both.

edemaine:
@kevinbarabash and @ylemkimon, thanks for all of the comments and reviewing!

twistedcubic:
Thanks for adding macros support! Are there plans to support macros definitions persistence across multiple `renderToString` calls? This would be useful to parse math across a document, without appending the macros definition each time.

edemaine:
@twistedcubic This is already supported for `\gdef` (but not `\newcommand` etc.). Check out the README documentation for the `macros` setting.  If you pass in an object, global macro definitions via `\gdef` will modify that object, so if you pass in the same object to future `renderToString`s they will share macros.  Some examples of use:

* https://github.com/edemaine/coauthor/blob/master/lib/formats.coffee#L461-L474
* https://github.com/Khan/KaTeX/pull/1425

Feel free to suggest ways in which this could be tweaked!

twistedcubic:
Thanks @edemaine! I was hoping for `\newcommand` persistence, because `\newcommand` has become much more widely used than `\def` or `\gdef`, but I could see if it's a design decision to keep `\newcommand` local, and resort to `\gdef` for everything global. I will do string substitution on the macros in the meanwhile.

I tried `\gdef` with `console.log(katex.renderToString("\\gdef\\a{1}"));
console.log(katex.renderToString("\\a"));`, but got the "EOF expected" error (for the second line) that happens when KaTeX doesn't recognize a symbol. Whereas `console.log(katex.renderToString("\\gdef\\a{1} \\a"));` produces the expected `1`. Can you tell me what I am missing?

colinfang:
@twistedcubic  I think @edemaine  means that you need to do 

```
let macros = {}
console.log(katex.renderToString("\\gdef\\a{1}", {macros: macros}));
console.log(katex.renderToString("\\\a", {macros: macros}));
```

twistedcubic:
Thanks! Using`\newcommand` does work for defining macros now, but these definitions don't persist across different calls to `renderToString` like `gdef` macros do. 

Are there plans to make macros defined with `newcommand` global, maybe with an option to do so?

twistedcubic:
I re-read Erik's answer from June, which I thought was before this PR's merge, but now it looks to be after. So I think the answer to whether `newcommand` macros can be made global, even as an optional operation, is no.

edemaine:
@twistedcubic Nope, all this discussion has been after the merge.  Probably should have been a seperate discussion issue...

I find the current behavior rather asymmetric (`\gdef` vs. no `\newcommand` equivalent), but this is merely a reflection of LaTeX.  If you can find an existing LaTeX macro that behaves like `\newcommand` but is global (`\globalcommand` or something), then I'd be happy to add it (and it would be very easy to do so).  Also, if we someday supporting parsing at the top level instead of within a math environment, perhaps as an option to `renderToString`, then that would provide a way to do what you want.

twistedcubic:
Thanks for your comment @edemaine. I also find it asymmetric. I do see your point on it being a reflection of LaTex, but since `\newcommand` is so frequently nowadays, I think it would be very useful to allow a `\newcommand` macro to persist across calls to `renderToString`.

twistedcubic:
Do you have any tips on how I can write a layer on top of katex.js, to remember the macros parsed from `\newcommand` commands in some macros list, and then pass this macros object to other calls of renderToString? Basically what you said should be very easy to do above.

My problem is that I can't concatenate the TeX strings together into one call of `renderToString`, especially since I want to continue parsing even after one string contains illegal syntax.

I would be happy to open a separate discussion topic, if this is unnecessarily tagging onto the work that you already finished. Thanks!

cirosantilli:
I made a dedicated issue for the `\global \newcommand` feature request at: https://github.com/KaTeX/KaTeX/issues/2070

PR #1385, opened 2018-05-30
Make letter tests strict

ronkok:
The recently added rendering mode `strict:"warn"` will allow any Unicode characters to be accepted. We expect `strict: true` to be pickier than that.

This PR sets `strict: true` on tests in `katex-spec.js` that should enforce that pickiness. Those mostly involve Unicode characters.

### DIFF
--- a/test/katex-spec.js
+++ b/test/katex-spec.js
@@ -14,6 +14,7 @@ import {
     defaultSettings,
     _getBuilt, getBuilt, getParsed, stripPositions,
 } from "./helpers";
+import strictSettings from "./helpers";
 
 const defaultOptions = new Options({
     style: Style.TEXT,
@@ -23,15 +24,15 @@ const defaultOptions = new Options({
 
 describe("A parser", function() {
     it("should not fail on an empty string", function() {
-        expect("").toParse();
+        expect("").toParse(strictSettings);
     });
 
     it("should ignore whitespace", function() {
-        expect("    x    y    ").toParseLike("xy");
+        expect("    x    y    ").toParseLike("xy", strictSettings);
     });
 
     it("should ignore whitespace in atom", function() {
-        expect("    x   ^ y    ").toParseLike("x^y");
+        expect("    x   ^ y    ").toParseLike("x^y", strictSettings);
     });
 });
 
@@ -102,7 +103,7 @@ describe("A punct parser", function() {
     const expression = ",;\\colon";
 
     it("should not fail", function() {
-        expect(expression).toParse();
+        expect(expression).toParse(strictSettings);
     });
 
     it("should build a list of puncts", function() {
@@ -643,15 +644,15 @@ describe("A text parser", function() {
     });
 
     it("should parse math within text group", function() {
-        expect("\\text{graph: $y = mx + b$}").toParse();
-        expect("\\text{graph: \\(y = mx + b\\)}").toParse();
+        expect("\\text{graph: $y = mx + b$}").toParse(strictSettings);
+        expect("\\text{graph: \\(y = mx + b\\)}").toParse(strictSettings);
     });
 
     it("should parse math within text within math within text", function() {
-        expect("\\text{hello $x + \\text{world $y$} + z$}").toParse();
-        expect("\\text{hello \\(x + \\text{world $y$} + z\\)}").toParse();
-        expect("\\text{hello $x + \\text{world \\(y\\)} + z$}").toParse();
-        expect("\\text{hello \\(x + \\text{world \\(y\\)} + z\\)}").toParse();
+        expect("\\text{hello $x + \\text{world $y$} + z$}").toParse(strictSettings);
+        expect("\\text{hello \\(x + \\text{world $y$} + z\\)}").toParse(strictSettings);
+        expect("\\text{hello $x + \\text{world \\(y\\)} + z$}").toParse(strictSettings);
+        expect("\\text{hello \\(x + \\text{world \\(y\\)} + z\\)}").toParse(strictSettings);
     });
 
     it("should forbid \\( within math mode", function() {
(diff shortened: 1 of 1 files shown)


test/katex-spec.js
```diff
@@ -2849,68 +2850,68 @@ describe("Unicode accents", function() {
...
-        expect("\\Aa").toNotParse(new Settings({strict: true}));
+        expect("\\text{\\aa\\AA}").toParseLike("\\text{\\r a\\r A}", strictSettings);
+        expect("\\aa").toNotParse(strictSettings);
+        expect("\\Aa").toNotParse(strictSettings);
     });
 
     it("should parse combining characters", function() {
-        expect("A\u0301C\u0301").toParseLike("Á\\acute C");
-        expect("\\text{A\u0301C\u0301}").toParseLike("\\text{Á\\'C}");
+        expect("A\u0301C\u0301").toParseLike("Á\\acute C", strictSettings);
```
edemaine: I'm confused -- this test should not pass under strict settings.

test/katex-spec.js
```diff
@@ -2849,68 +2850,68 @@ describe("Unicode accents", function() {
...
+        expect("\\text{A\u0301C\u0301}").toParseLike("\\text{Á\\'C}", strictSettings);
     });
 
     it("should parse multi-accented characters", function() {
-        expect("ấā́ắ\\text{ấā́ắ}").toParse();
+        expect("ấā́ắ\\text{ấā́ắ}").toParse(strictSettings);
```
edemaine: This test should also be nonstrict (or separate the `\text` part in strict and the non-`\text` part in nonstrict).

test/katex-spec.js
```diff
@@ -2849,68 +2850,68 @@ describe("Unicode accents", function() {
...
         // Doesn't parse quite the same as
         // "\\text{\\'{\\^a}\\'{\\=a}\\'{\\u a}}" because of the ordgroups.
     });
 
     it("should parse accented i's and j's", function() {
-        expect("íȷ́").toParseLike("\\acute ı\\acute ȷ");
-        expect("ấā́ắ\\text{ấā́ắ}").toParse();
+        expect("íȷ́").toParseLike("\\acute ı\\acute ȷ", strictSettings);
+        expect("ấā́ắ\\text{ấā́ắ}").toParse(strictSettings);
```
edemaine: Same here.

test/katex-spec.js
```diff
@@ -2849,68 +2850,68 @@ describe("Unicode accents", function() {
...
     });
 });
 
 describe("Unicode", function() {
     it("should parse negated relations", function() {
-        expect("∉∤∦≁≆≠≨≩≮≯≰≱⊀⊁⊈⊉⊊⊋⊬⊭⊮⊯⋠⋡⋦⋧⋨⋩⋬⋭⪇⪈⪉⪊⪵⪶⪹⪺⫋⫌").toParse();
+        expect("∉∤∦≁≆≠≨≩≮≯≰≱⊀⊁⊈⊉⊊⊋⊬⊭⊮⊯⋠⋡⋦⋧⋨⋩⋬⋭⪇⪈⪉⪊⪵⪶⪹⪺⫋⫌").toParse(strictSettings);
```
edemaine: Nice to see `toBuild` in the examples below!  Should this one also be `toBuild`?

test/katex-spec.js
```diff
@@ -3042,16 +3043,16 @@ describe("Newlines via \\\\ and \\newline", function() {
...
     it("should parse spacing functions in math or text mode", () => {
-        expect("A\\;B\\,C\\nobreakspace \\text{A\\;B\\,C\\nobreakspace}").toBuild();
+        expect("A\\;B\\,C\\nobreakspace \\text{A\\;B\\,C\\nobreakspace}").toBuild(strictSettings);
     });
 
     it("should render ligature commands like their unicode characters", () => {
-        const commands = getBuilt("\\text{\\ae\\AE\\oe\\OE\\o\\O\\ss}");
-        const unicode = getBuilt("\\text{æÆœŒøØß}");
+        const commands = getBuilt("\\text{\\ae\\AE\\oe\\OE\\o\\O\\ss}, strictSettings");
+        const unicode = getBuilt("\\text{æÆœŒøØß}, strictSettings");
```
edemaine: The quotes are in the wrong place here.

test/katex-spec.js
ronkok: The result we're getting on this test is not just an artifact of the testing system. I've put `ÁĆ  Á\acute C` into a strict mode test page and verified that it does render without an error message.

Since this PR is just about applying `strict` to tests, I'm not going to try to overcome the underlying issue. On this particular test, I will omit the `strict`.

test/katex-spec.js
ronkok: On a test page, this string does raise a strict-mode error. I don't know why jest did not flag it. I will omit `strict` from this line.

test/katex-spec.js
ronkok: Ditto.

test/katex-spec.js
ronkok: The `∉` entry is done via macro, not from `symbols.js`. It renders just fine. But KaTeX sends a console message that says, "No character metrics for '∉' in style 'Main-Regular'", which jest flags as an error in the build.

Is that a bug? Should it render without the console error?

edemaine: `\notin` produces the same error.  So that's a bug, though not necessarily for this PR.

ronkok: I was mistaken earlier when I said it was a macro. It is in fact in `symbols.js`.  Here's the line:
```
defineSymbol(math, main, rel, "\u2209", "\\notin", true);
```

I don't see anything wrong with that line of code. Maybe there is something wonky in metrics?

edemaine: Yep, 8713 is missing from `fontMetricsData.js`.  I've added an issue here: https://github.com/KaTeX/katex-fonts/issues/47

ronkok:
@edemaine That's great. Thanks for the help.

PR #1386, opened 2018-05-31
Separate parsing of \left and \right for improved type safety.

marcianx:


kevinbarabash:
@marcianx awesome!  I wonder if eventually we might just move the `handlers` for `\left` and `\right` into Parser.js.  I think there were in Parser.js to begin with and at the time I was striving for uniformity with the `defineFunction` API.  The truth is, tracking the left/right depth works out is more of a parser thing to do.  Also, if this is in the parser we may be able to get rid of having to create the extra parse node for the right delimiter.

marcianx:
I wouldn't mind that. This does feel a little out of place. The parser already recognizes `"\\right"` as an expression ender, so it's not like part of that knowledge isn't already there.

PR #1387, opened 2018-05-31
Replace ParseNode<*> with a more accurate AnyParseNode and fix flow errors.

marcianx:
This is the correct type and is needed for (1) full type-safety, and (2) as a requisite to add types to `buildHTML` and `buildMathML`.

### DIFF
--- a/src/ParseError.js
+++ b/src/ParseError.js
@@ -1,6 +1,7 @@
 // @flow
-import ParseNode from "./ParseNode";
 import {Token} from "./Token";
+
+import type {AnyParseNode} from "./ParseNode";
 
 /**
  * This is the ParseError class, which is the main error thrown by KaTeX
@@ -15,8 +16,8 @@ class ParseError {
         // Error position based on passed-in Token or ParseNode.
 
     constructor(
-        message: string,              // The error message
-        token?: Token | ParseNode<*>, // An object providing position information
+        message: string,               // The error message
+        token?: ?Token | AnyParseNode, // An object providing position information
     ) {
         let error = "KaTeX parse error: " + message;
         let start;
--- a/src/ParseNode.js
+++ b/src/ParseNode.js
@@ -1,9 +1,10 @@
 // @flow
 import SourceLocation from "./SourceLocation";
-import type {ArrayEnvNodeData} from "./environments/array.js";
+import {GROUPS} from "./symbols";
+import type {ArrayEnvNodeData} from "./environments/array";
 import type {Mode, StyleStr} from "./types";
-import type {Token} from "./Token.js";
-import type {Measurement} from "./units.js";
+import type {Token} from "./Token";
+import type {Measurement} from "./units";
 
 /**
  * The resulting parse tree nodes of the parse tree.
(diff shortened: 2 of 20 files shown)


src/ParseError.js
```diff
@@ -15,8 +16,8 @@ class ParseError {
         // Error position based on passed-in Token or ParseNode.
 
     constructor(
-        message: string,              // The error message
-        token?: Token | ParseNode<*>, // An object providing position information
+        message: string,               // The error message
+        token?: ?Token | AnyParseNode, // An object providing position information
```
ylemkimon: Is `?` in front of `Token` necessary?

marcianx: Yes, so that a nullable value can be explicitly provided.

kevinbarabash:
I'll review this weekend.

src/ParseNode.js
```diff
@@ -40,18 +41,83 @@ export type NodeValue<TYPE: NodeType> = $ElementType<ParseNodeTypes, TYPE>;
...
 export type LeftRightDelimType = {|
     type: "leftright",
-    body: ParseNode<*>[],
+    body: AnyParseNode[],
     left: string,
     right: string,
 |};
 
+// ParseNode's corresponding to Symbol `Group`s in symbols.js.
+export type SymbolParseNode =
```
kevinbarabash: This is definitely a step in the right direction, but not it's final form.  That can wait for another PR.

src/environments/array.js
```diff
@@ -450,10 +452,17 @@ defineEnvironment({
...
-        let colalign = args[0];
-        colalign = colalign.value.map ? colalign.value : [colalign];
-        const cols = colalign.map(function(node) {
+    handler(context, args) {
+        // Since no types are specified above, the two possibilities are
+        // - The argument is wrapped in {} or [], in which case Parser's
+        //   parseGroup() returns an "ordgroup" wrapping combinations of
+        //   "mathord"s ("l", "c") or "textord"s ("|").
+        // - The argument is a bare "mathord" or "textord".
+        const ordNode = checkNodeTypes2(args[0], "mathord", "textord");
```
kevinbarabash: This whole mathord/textord thing is a little weird.  We should probably have:
```
{
  type: "ord",
  mode: "math" | "text",
  nucleus: ...
}
```
This change can happen in a future PR though.

src/Parser.js
```diff
@@ -352,14 +352,15 @@ export default class Parser {
 
             if (lex.text === "\\limits" || lex.text === "\\nolimits") {
                 // We got a limit control
-                if (!base || base.type !== "op") {
+                const opNode = checkNodeType(base, "op");
+                if (opNode) {
```
kevinbarabash: Thanks for making things more consistent.

src/Parser.js
```diff
@@ -463,10 +464,12 @@ export default class Parser {
             const result = env.handler(context, args, optArgs);
             this.expect("\\end", false);
             const endNameToken = this.nextToken;
-            const end = assertNodeType(this.parseFunction(), "environment");
+            let end = this.parseFunction();
             if (!end) {
                 throw new ParseError("failed to parse function after \\end");
-            } else if (end.value.name !== envName) {
+            }
+            end = assertNodeType(end, "environment");
```
kevinbarabash: Good call.

src/environments/array.js
```diff
@@ -85,8 +88,8 @@ function parseArray(
         }
     }
 
-    let row = [];
-    const body = [row];
+    let lastRow = [];
```
kevinbarabash: Why the name change?

src/environments/array.js
```diff
@@ -194,8 +194,8 @@ const htmlBuilder = function(group, options) {
         hlinePos.push(totalHeight);
     }
 
-    for (r = 0; r < groupValue.body.length; ++r) {
-        const inrow = groupValue.body[r];
+    for (r = 0; r < group.value.body.length; ++r) {
+        const inrow = group.value.body[r];
```
kevinbarabash: These are the ones I missed in my PR?

src/environments/array.js
```diff
@@ -403,7 +404,8 @@ const alignedHandler = function(context, args) {
     res.value.body.forEach(function(row) {
         for (let i = 1; i < row.length; i += 2) {
             // Modify ordgroup node within styling node
-            const ordgroup = row[i].value.value[0];
+            const styling = assertNodeType(row[i], "styling");
+            const ordgroup = assertNodeType(styling.value.value[0], "ordgroup");
```
kevinbarabash: Nice!

src/functions/color.js
```diff
@@ -99,10 +98,7 @@ defineFunction({
         argTypes: ["color"],
     },
     handler({parser, breakOnTokenText}, args) {
-        const color = args[0];
-        if (!color) {
-            throw new ParseError("\\color not followed by color");
-        }
+        const color = assertNodeType(args[0], "color-token");
```
kevinbarabash: We could do a `check` here instead and then we'd be able to keep the error message from before.

src/ParseNode.js
```diff
@@ -315,11 +380,71 @@ export function assertNodeType<NODETYPE: NodeType>(
...
+    }
+    return typedNode;
+}
+
+/**
+ * Returns the node more strictly typed iff it is of the given type. Otherwise,
+ * returns null.
+ */
+export function checkSymbolNodeType(node: ?AnyParseNode): ?SymbolParseNode {
+    if (node && GROUPS.hasOwnProperty(node.type)) {
```
kevinbarabash: We'll be able to get rid of this weirdness once we have a separate `atom` parse node type.

src/environments/array.js
marcianx: That's a good idea. I agree.

src/environments/array.js
marcianx: I removed the local variable `lastRow` below as it is always the same as this one (whose type the inference can easily figure out). I renamed this to `lastRow` since it's relevant in that context to make the semantics more obviously correct. I can renamed it back if you prefer.

src/environments/array.js
marcianx: No. This is a consequence of me removing
```
const groupValue = assertNodeType(group, "array").value;
```
since it should not be needed. `htmlBuilder`'s `group` parameter is already strictly typed as `ParseNode<"array">`.

src/functions/color.js
marcianx: This check should never fail here due to the `argTypes: ["color"],` above. It would fail someplace in `ParseNode` instead.

src/ParseNode.js
marcianx: Yes! This is exactly what I was referring to when you originally suggested the `atom` idea.

src/environments/array.js
marcianx: Actually, this should just be `atom` now instead of `ord`. When I rebased on latest head, some change resulted in a `rel` type object here. So I removed `assert/checkNoteTypes2` and used `assert/checkSymbolNodeType` instead.

kevinbarabash: yes, you're right, maybe:
```
type: "atom", 
kind: "ord" | "bin" | "rel" | ..., 
mode: "math" | "text",
nucleus: ...
```

src/environments/array.js
kevinbarabash: Reusing this variable in the `else if (next === "\\end")` block makes sense.  Since there can be multiple rows the variable doesn't hold strictly the last row I think that calling it `row` would be more accurate.

marcianx: Doesn't it always hold the last row? Below, whenever it's assigned a new value, it's also pushed to the end of `body`. I don't see any other pushes into `body`.

kevinbarabash: Inside the `while` loop it's gets push to `body` whenever there's a `\\cr` and a new one is created at the same time.

marcianx: Ah, I think I understand our difference in interpretations. Just to be on the same page: my interpretation of
```
lastRow = [];
body.push(lastRow);
```
is that at after these two lines, `lastRow` still points to what's *currently* the last row in `body` (as opposed to final last row). It made its role in the rest of the logic a bit clearer to me, but I can understand the confusion due to a different interpretation.
Renamed back to `row`.

PR #1388, opened 2018-05-31
Automatic mrel/mbin spacing for \boldsymbol

### LINKED ISSUE #1154: \boldsymbol removes spacing in expressions
Rendering: `a\boldsymbol{+}c` will not contain spacing between the appropriate terms. 

screen shot 2018-02-12 at 4 08 39 pm

edemaine:
Fix #1154. This code ended up being very similar to code @ronkok wrote for `\stackrel`.  At the same time, I simplified that code (using `?:`) and corrected a bug (handling an empty ordgroup, which test now checks for).


src/functions/font.js
```diff
@@ -53,13 +53,44 @@ defineFunction({
...
+        const body = args[0];
+        // amsbsy.sty's \boldsymbol inherits the argument's bin|rel|ord status
+        // (similar to \stackrel in functions/mclass.js)
+        let mclass = "mord";
+        const atomType = (body.type === "ordgroup" ?
+            body.value.length && body.value[0].type : body.type);
+        if (/^(bin|rel)$/.test(atomType)) {
+            mclass = "m" + atomType;
+        }
+        return {
```
kevinbarabash: After rebasing you'll need to return a `ParseNode` from here.

test/__snapshots__/katex-spec.js.snap
```diff
@@ -1,5 +1,435 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
+exports[`A font parser \\boldsymbol should inherit mbin/mrel from argument 1`] = `
```
kevinbarabash: This snapshot is surprising hard to read.  It would be cool if we could choose to filter the props to the tree to only show the `value` and `classes` on some tests.

edemaine: Sounds like a good idea.  Maybe for after #1336?

kevinbarabash: The snapshot should be easier to read in future PRs if we end of having to tweak something.

kevinbarabash:
I feel like this might benefit from a screenshot test.  Maybe render some math using both `\boldsymbol` and `\bm`.

src/functions/font.js
```diff
@@ -53,13 +53,44 @@ defineFunction({
...
+        greediness: 2,
+    },
+    handler: ({parser}, args) => {
+        const body = args[0];
+        // amsbsy.sty's \boldsymbol inherits the argument's bin|rel|ord status
+        // (similar to \stackrel in functions/mclass.js)
+        let mclass = "mord";
+        const atomType = (body.type === "ordgroup" ?
+            body.value.length && body.value[0].type : body.type);
+        if (/^(bin|rel)$/.test(atomType)) {
```
ronkok: Should `body.value.length &&` be placed at the beginning of the ternary?

edemaine:
I've added a screenshot test, and fixed @ronkok's point.  This is ready for review again.

edemaine:
@kevinbarabash Thanks for the reviews!

PR #1391, opened 2018-06-01
Automatic submodule updating via husky tool

### LINKED ISSUE #1133: Improve dev docs around submodules in CONTRIBUTING.md
Whenever switching branches it's best to also do a `git submodule update` b/c for some reason `git` doesn't do this by default.

edemaine:
As suggested by @ylemkimon in #1133, this PR adds Git hooks to call `git submodule update --init --recursive` (something I've now memorized, but still find a bit tedious, and difficult to explain) after checkout and merge.

I switched from the `pre-commit` NPM package to a more general [`husky` package](https://github.com/typicode/husky) (used by jQuery, babel, etc.).  It seems nice, and preserves the old `pre-commit` behavior.

These scripts get installed during `npm install`, so we still have a problem when people initially check out the package and expect things (`npm start`, `cli.js`) to just work.  Should we add this `git` command to the `npm install` script?

I also updated the documentation, which should hopefully fix #1133.


kevinbarabash:
> Should we add this git command to the npm install script?

Won't that affect people doing `npm install katex`?

edemaine:
I hadn't fully formed the thought... but I guess I mean something like the `prestart` script in `package.json`, which is triggered by `build`, `test`, etc.  For example, `check-node-version.js` could see if the submodules aren't checked out, and if so, check them out.  (I don't think we want to update submodules every time we e.g. `npm test` -- just trying to catch the "initial checkout" case.)

edemaine:
@kevinbarabash Thanks for rhe reviews!

PR #1392, opened 2018-06-01
Support \cfrac

### LINKED ISSUE #85: Add support for \cfrac

ronkok:
Much of this comes directly from #135, so credit to @kevinbarabash.

This was an easier PR to write than #135 since KaTeX now has well established methods for pt-to-em conversions and nulldelimiter.

As in #135, this PR does not support the LaTeX optional argument for numerator justification.  It supports a `\cfrac #1 #2 ` syntax.

Fixes #85.


ronkok:
The screenshot below compares a `\frac` to a `\cfrac`.
cfrac

kevinbarabash:
@ronkok here are the updated screenshots:

Chrome
fractiontest-chrome

Firefox
fractiontest-firefox

src/functions/genfrac.js
```diff
@@ -197,7 +209,10 @@ defineFunction({
                 group.value.leftDelim, delimSize, true,
                 options.havingStyle(style), group.mode, ["mopen"]);
         }
-        if (group.value.rightDelim == null) {
+
+        if (group.value.continued) {
+            rightDelim = buildCommon.makeSpan([]); // zero width for \cfrac
```
kevinbarabash: I thought this might cause `\left`/`\right` delimiters to not render, but it doesn't.  I guess the left/right delimeters for `genfrac` are for `a \choose b`.  Do we need a similar `if (group.value.continued)` for `leftDim` up above?

ronkok: I don't think so. In LaTeX, they write a `\kern-\nulldelimiterspace`, but only on the right side of the fraction, not on the left side. We get the same behavior by omitting the space taken by our `\nulldelimiter`.

edemaine:
Happy to see this work, which dates back to 2014, come to fruition! Thanks @ronkok and @kevinbarabash!

kevinbarabash:
I can't believe it's been that long.

edemaine:
@OrkhanAlikhanov Great! Let us know if there are other sticking points you know of, or discover.

OrkhanAlikhanov:
@edemaine Hey! I've removed MathJax and integrated KaTeX. Everything seems okay so far. Rendering is now super fast! Especially in android and ios apps. Thank you all!

edemaine:
@OrkhanAlikhanov Glad to hear it!!

PR #1393, opened 2018-06-02
update function handlers to return a ParseNode

kevinbarabash:
It was kind of weird having the function handlers return data which the parser would use to create `ParseNode`s from.  Everywhere else we just create a `ParseNode`.  This change makes things more consistent by having the function handlers also just create a `ParseNode`.

### DIFF
--- a/src/Parser.js
+++ b/src/Parser.js
@@ -9,7 +9,7 @@ import {validUnit} from "./units";
 import {supportedCodepoint} from "./unicodeScripts";
 import unicodeAccents from "./unicodeAccents";
 import unicodeSymbols from "./unicodeSymbols";
-import ParseNode from "./ParseNode";
+import ParseNode, {assertNodeType} from "./ParseNode";
 import ParseError from "./ParseError";
 import {combiningDiacriticalMarksEndRegex} from "./Lexer.js";
 import Settings from "./Settings";
@@ -442,12 +442,12 @@ export default class Parser {
 
         if (func === "\\begin") {
             // begin...end is similar to left...right
-            const begin = this.parseGivenFunction(start);
-            // $FlowFixMe
+            const begin =
+                assertNodeType(this.parseGivenFunction(start), "environment");
+
             const envName = begin.value.name;
             if (!environments.hasOwnProperty(envName)) {
                 throw new ParseError(
-                    // $FlowFixMe
                     "No such environment: " + envName, begin.value.nameGroup);
             }
             // Build the environment object. Arguments and other information will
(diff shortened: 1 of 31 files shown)


kevinbarabash:
Since `ParseNode` is just a data class we should be able to get rid of it in the future once all of the data in the `ParseNode` lives at the same level.  Then it'll be easy to enforce what should appear in each node using flow obviating the need for `ParseNode` completely.

src/Parser.js
```diff
@@ -443,9 +443,11 @@ export default class Parser {
         if (func === "\\begin") {
             // begin...end is similar to left...right
             const begin = this.parseGivenFunction(start);
+            // $FlowFixMe
```
kevinbarabash: @marcianx we should be able to use `checkNodeType` or `assertNodeType` here to clean up these `$FlowFixMe`s right?

src/Parser.js
```diff
@@ -244,8 +244,8 @@ export default class Parser {
                 denomNode = new ParseNode("ordgroup", denomBody, this.mode);
             }
 
-            const value = this.callFunction(funcName, [numerNode, denomNode], []);
-            return [new ParseNode(value.type, value, this.mode)];
```
marcianx: It's a pity: we never passed in the source location when constructing these nodes, and now we're spreading that behavior everywhere.

marcianx:
Thanks! This was definitely one of the necessary steps for the final refactor!

ylemkimon:
Will it be possible to remove either `ParseNode.type` or `ParseNode.value.type`? It seems duplicated.

src/Parser.js
```diff
@@ -461,7 +463,7 @@ export default class Parser {
             const result = env.handler(context, args, optArgs);
             this.expect("\\end", false);
             const endNameToken = this.nextToken;
-            const end = this.parseFunction();
+            const end = assertNodeType(this.parseFunction(), "environment");
```
marcianx: Ah, it was not until merging with my PR that I realized. I'd done this `assertNodeType` after the `if (!end)` below so that we could give a more logical error message. What do you think? I can leave that in in my current PR. (We can discuss it on my PR, I suppose.)

marcianx:
@ylemkimon Yes, that's the eventual goal -- `ParseNode` would become a type union of objects, each of which would have a single `type` field. (Kevin already started some of the work in making sure all the value types have a `.type` field on them.) We want to eventually get to the state Kevin suggested here: https://github.com/Khan/KaTeX/issues/892

PR #1395, opened 2018-06-02
Added ability to create a dashed vertical line in arrays, using ':'

zacharyrs:
Not really sure what else there is to say, just needed this for a class.

Open to suggestions if there are any?


zacharyrs:
[clabot:check]

kevinbarabash:
@zacharyrs thanks for the PR.  Can you update the `Arrays` test in `ss_data.yaml` so that one of the separators is the separator you added?  The tests will fail, but we'll be able to grab the updated screenshots from circleci and update them.  Also, can you update the the `should accept vertical separators` test in katex-spec.js to use the new separator?

zacharyrs:
Think that should do it, now to see if it looks okay...

edemaine:
@zacharyrs I didn't know this was possible in LaTeX -- cool.  Are you implementing the same dash length as [arydshln.sty](https://ctan.org/pkg/arydshln?lang=en)?  Or is there another simpler package that offers this feature?  (Maybe we should add a TODO to support `;{dash/gap}` notation, and the horizontal dashed commands.)

ronkok:
@zacharyrs I agree. This is seriously cool. 

@edemaine I took a look at the `arydshln` package this morning. Its dashed lines have a `\dashsegment` and a `\dashgap` and the default is 4pt for both segment and gap. Both are longer than what Chrome provides for CSS `border-style: dashed`, so I tried to reproduce the `arydshln` line style another way. I was not happy with my results.

The CSS `border-style: dashed` does not provide for custom dash segments or gaps. In order to get customization, I tried `border-image`. I was successful in getting segment lengths and gap lengths that match `arydshln`, but at a cost. It's an image, and its a thin one, so it is prone to the same Chrome disappearing line problems that we have struggled to overcome with fraction lines.

Personally, I'd rather have sharp, reliable CSS `border-style: dashed` than have an exact match to `arydshln` segment lengths. I think this PR should proceed as is. We can always change the line type later if we find a way to do it in a reliable way.

edemaine:
@ronkok Thanks for the investigation.  I hadn't realized `border-style` is so restrictive, and it's probably a bit much (and difficult) to go to SVG for this.  Agreed that this approach is probably all we want.  Maybe the main TODO would be similar horizontal dashed lines.

edemaine:
@zacharyrs You'll need to update the screenshot images manually as part of the PR.  See [CONTRIBUTING.md](https://github.com/Khan/KaTeX/blob/master/CONTRIBUTING.md#screenshot-tests) and [the linked screenshotter documentation](https://github.com/Khan/KaTeX/tree/master/dockers/Screenshotter).  Or let us know if you need any help.

ronkok:
> Maybe the main TODO would be similar horizontal dashed lines.

Agreed. `arydshln` has a `\hdashline` function. I'll see if I can add that to the `\hline` work that we did recently.

ronkok:
> it's probably a bit much (and difficult) to go to SVG for this.

I was to use `border-image: repeating-linear-gradient`, so it was only two lines of code. It's not difficult to do. The problem is that I just don't trust browsers with an image this thin. Span borders are the way to go.

kevinbarabash:
Here's the screenshot from Chrome:

arrays-chrome

The rendering is unfortunately too wide.  The Firefox web driver just bails when it tries to take a screenshot of something wider than the browser window.  Chrome's web driver is a little more forgiving.  @zacharyrs can you remove the column you added?  We already have a number of `|` separators, we can change one of those to `:`.

zacharyrs:
@kevinbarabash @edemaine 

Thanks for the explanations on how to grab screenshots, I assumed I just didn't have access. 

Also, woops regarding it being too wide, I wasn't sure so I just went overboard with it, I'll tidy it up in a sec. Possibly worth changing the font size of that section so you can have longer output?

Regarding where this is from, I just needed it for a uni class where they said to use a dashed line, and Boostnote, which uses KaTeX, couldn't. It's basically the arydshln package, but I only implemented ':', as ';' seemed a bit pointless till there's a way to configure the length of the strokes.

After a quick google, I found this https://stackoverflow.com/a/31315911/9440025, which does give a few different ways we could add this with more functionality.


Also, on a bit of a tangent, and more just a question regarding LaTeX/KaTeX, why use matrix when there's no support for separators, or more specifically, why not add such support to the various matrix types?
It'd be tangent to the existing implementations, but shouldn't be a dealbreaker?

zacharyrs:
@kevinbarabash @edemaine 

I can't seem to grab the screenshots, not sure where they are...

edemaine:
@zacharyrs Can you access https://circleci.com/gh/Khan/KaTeX/1203#artifacts/containers/0 ?

ronkok:
@zacharyrs  You have to open an account with CircleCI before they will show you the artefacts.

test/screenshotter/ss_data.yaml
```diff
@@ -33,7 +33,7 @@ Alignedat: |
         3a &= 3 & 17b &= 34 &\quad 400c &= 1200
     \end{alignedat}
 Arrays: |
-    \left(\begin{array}{|rl|c||}
+    \left(\begin{array}{|rl:c}
```
edemaine: I think we still want to test (the space between) a double vertical bar, if we can fit it in the test...

edemaine:
You should be able to log in with your Github account, I think.

ronkok:
> why not add [separators] to the various matrix types?

We did recently add `\hline` and it works in all the environments. As to vertical lines, we don't put them into the various matrix types because AMS-LaTeX doesn't put them there.  You can insert vertical lines with the array environment, as in:

bmatrix

edemaine:
> why use matrix when there's no support for separators, or more specifically, why not add such support to the various matrix types?

Most mathematics I've seen doesn't need separators. Hence the `...matrix` shorthands. But `array` is the main environment with the most options; it's easy enough to wrap in `\left`/`\right`.

test/screenshotter/ss_data.yaml
zacharyrs: Woops, okay yeah I'll put that back.

zacharyrs:
Okay, lines are back up.

screen shot 2018-06-03 at 1 19 40 pm


And regarding the not adding support, it's something that I needed for uni, so I was curious - it's a real pain to wrap everything in `\left` and `\right`, plus I'm lazy...

From what I gather, there's no real support to create/modify environments locally, but I can create some macros instead. I'll take a look into adding that to BoostNote I guess.

kevinbarabash:
@zacharyrs you'll need to update the screenshots.

Chrome:
arrays-chrome

Firefox:
arrays-firefox

zacharyrs:
@kevinbarabash update the screenshots?

kevinbarabash:
@zacharyrs the tests are failing before the screenshots that are produced during the screenshot tests don't match what existing on your branch.  Copy the images I posted into test/screenshotter/images and push that change and you should be good.

zacharyrs:
Woops that makes sense. Done now.

zacharyrs:
When I have some more time, I'll add support for customising the dashes, probably via an SVG.

kevinbarabash:
@zacharyrs sounds good.  Thanks again for the PR.

ronkok:
@zacharyrs, Congratulations on this PR. It is seriously cool.

Next, has everybody noticed that since the release of v0.10.0-alpha, no one has opened an issue to say that fraction lines are disappearing? No one has complained that lines are gray with fuzzy edges instead of black and sharp? It looks like we have finally overcome browser rendering problems with these thin lines.

That’s great news! We should celebrate that!

And so I want to want to advise that we avoid images, including SVGs, for separator lines. We tried SVGs, configured three different ways, in release 0.9.0-alpha2, release 0.9.0-beta, and release 0.9.0. We got user complaints on every one of them, on both fractions lines and vertical separator lines. SVG lines are better on retina screens than on low-pixel-density screens, but there are a lot of low-density screens still out there.

@zacharyrs, I think your PR is awesome just the way it is. I’d like to save you from wasting the time that I wasted on SVGs for separator lines and fraction lines. Someday, when everybody has a high-density screen, we can revisit this decision. Until then, let’s allow your awesome work to go on being awesome. Let’s use the lines as they are now.

zacharyrs:
@ronkok thanks for the feedback, perhaps gradients could be used instead?
https://stackoverflow.com/a/31315911/9440025 suggests three methods to implement a dashed border that's customisable, the second being border gradients.

With them, the major con would be browser support, but https://caniuse.com/#feat=css-gradients shows that's fine now.
The minor con would be border radius would not work, but that wouldn't be used on the separator anyway.

zacharyrs:
@ronkok thanks for the explanation!

I'm currently cramming for uni exams, but after these I'll revisit it.
I did play around with a very thin background image, and can see that if less than 0.5px fills any pixel, it doesn't go black, but otherwise doesn't seem to go grey either. Also, with regards to borders, they do vanish when less than 0.5px coverage occurs. You can see this by just offsetting a 0.5px border by a 0.5px margin.

I think that with the background image, if a min-width of 1px was set, then at least one pixel would go dark, which should prevent the issue. By setting the background width to a huge number, the lines will span the whole element, which we can then just ensure is at least 1px wide.

ronkok:
> Also, with regards to borders, they do vanish when less than 0.5px coverage occurs. 

KaTeX now has a `min-width: 1px` CSS rule on the spans that have these borders. That rule has prevented vanishing at text sizes big enough to read. I hope that continues.

zacharyrs:
Yes, I'd use that same concept on the gradient border.

Using the macOS Color Meter tool, locked onto a pixel, and then fiddling with the margin, it appears to completely black out whichever pixel the background covers more of on Chrome. This also holds true on Firefox.

ronkok:
> a very thin background image

Another lesson learned from sad experience: At least some browser default settings for printing a page will not print background images. So we've changed all our background images to inline SVGs. 

I believe that `border-image: repeating-linear-gradient` will print, but should be checked before it goes into production.

zacharyrs:
> browser default settings for printing a page will not print background images

Can mitigate that on chrome, safari, and firefox with the following:
```css
* {
	-webkit-print-color-adjust: exact !important; /* Chrome, Safari */
	color-adjust: exact !important; /* Firefox */
}
```

IE is still a pain though.

PR #1400, opened 2018-06-02
Implement \neq and \notin via macros like LaTeX's

edemaine:
As discussed in https://github.com/KaTeX/katex-fonts/issues/47#issuecomment-394103578, I simplified LaTeX's definitions of `\neq` and `\notin` into KaTeX macros.

They seem visually very close to LaTeX's; here's a texcmp for the tweaked `Not` test (only third line is new):

image

I assume this will remain true after #1267 but it would be good to test.

This PR removes the need for the `\neq` and `\notin` symbols from the fonts, so it would fix https://github.com/KaTeX/katex-fonts/issues/47

The new `\neq` should be essentially the same as the font-based one.  Here is a comparison of the new `\notin` symbol (top) vs. the old `\notin` symbol (bottom):

image


kevinbarabash:
I think this is a move in the right direction.  I'm confused though why the slant of the slash is different in `\notin` in the third line especially if `\notin` is now defined as `\not\in` now.

src/macros.js
```diff
@@ -223,6 +223,17 @@ defineMacro("\\llap", "\\mathllap{\\textrm{#1}}");
...
 
+// Negated symbols from base/fontmath.ltx:
+// \def\neq{\not=} \let\ne=\neq
+// \DeclareRobustCommand
+//   \notin{\mathrel{\m@th\mathpalette\c@ncel\in}}
+// \def\c@ncel#1#2{\m@th\ooalign{$\hfil#1\mkern1mu/\hfil$\crcr$#1#2$}}
+defineMacro("\\neq", "\\not=");
+defineMacro("\\ne", "\\neq");
+defineMacro("\u2260", "\\neq");
+defineMacro("\\notin", "\\mathrel{{\\in}\\mathllap{/\\mskip1mu}}");
```
kevinbarabash: Shouldn't the `/` be a `\not`?  It seems weird to use a different glyph for `\notin`.

ronkok: @kevinbarabash If I remember correctly, `\notin` was created shortly before `\not` was available.

edemaine: This is what LaTeX does; see the definition in comments. I believe it's a conscious decision because the `\in` glyph is relatively narrow. So I think we should follow suit.

kevinbarabash: > They seem visually very close to LaTeX's; here's a texcmp for the tweaked Not test (only third line is new):

SGTM.

edemaine: ...unless there's another LaTeX package that redefines `\notin` differently? I know AMS doesn't redefine it. I'm on a phone so can't easily check symbols-a4.

edemaine: There is a rather large variety of `\notin` symbols listed in [symbols-a4](http://tug.ctan.org/info/symbols/comprehensive/symbols-a4.pdf), from packages `fdsymbol`, `stix`, `txfonts`, `mathabx`, `MnSymbol`.

image

Not sure what to make of all this.  I guess I lean toward matching LaTeX's default behavior, but I'm also happy to design a `\not`-based macro if that's the preference.

ronkok:
For what it's worth, here is the MathML from (1) font-based `\notin` and (2) macro-based `\notin`.

```
<mrow>
  <mo>∉</mo>
</mrow>

<mrow>
  <mstyle>
    <mrow>
      <mo>∈</mo>
    </mrow>
    <mpadded lspace="-1width" width="0px">
      <mrow>
        <mi mathvariant="normal">/</mi>
        <mspace width="0.05555555555555555em"></mspace>
      </mrow>
    </mpadded>
  </mstyle>
</mrow>
```

edemaine:
@ronkok Yes, this is a downside to the macro approach in general.  I was thinking we could fix all of these with a `\mathml{LaTeX code used for HTML}{MathML equivalent}` command.

ronkok:
Now that would be cool.

src/macros.js
kevinbarabash: I'd lean towards `LaTeX`'s default behaviour as well.

Issue #894, opened 2017-09-18
cli.js doesn't run within the repo


kevinbarabash:
Steps to reproduce:
- git clone https://github.com/Khan/KaTeX
- ./cli.js

Results in the following output:
```
module.js:491
    throw err;
    ^

Error: Cannot find module './'
    at Function.Module._resolveFilename (module.js:489:15)
    at Function.Module._load (module.js:439:25)
    at Module.require (module.js:517:17)
    at require (internal/module.js:11:18)
    at Object.<anonymous> (/Users/kevinbarbash/khan/KaTeX/cli.js:6:15)
    at Module._compile (module.js:573:30)
    at Object.Module._extensions..js (module.js:584:10)
    at Module.load (module.js:507:32)
    at tryModuleLoad (module.js:470:12)
    at Function.Module._load (module.js:462:3)
```

Whatever the fix for this is, we'll want it to also work when running `npm install katex`.

raoulb:
Running `cli.js` from a node installation (there called `katex`) works fine atm.

raoulb:
If I change the line 6 to:
```
const katex = require("./katex");
```
I get
```
import ParseError from "./src/ParseError";
^^^^^^

SyntaxError: Unexpected token import
    at createScript (vm.js:56:10)
    at Object.runInThisContext (vm.js:97:10)
    at Module._compile (module.js:542:28)
    at Object.Module._extensions..js (module.js:579:10)
    at Module.load (module.js:487:32)
    at tryModuleLoad (module.js:446:12)
    at Function.Module._load (module.js:438:3)
    at Module.require (module.js:497:17)
    at require (internal/module.js:20:19)
    at Object.<anonymous> (/home/raoul/KaTeX/cli.js:6:15)
```
Now afaict this is something with new ES6 things.

edemaine:
@raoulb Are you using Node 8? I think `import` is only available there.

To fix this, we might want to build a version of `cli.js` via Babel.

raoulb:
No, not on 8 yet:
```
nodejs --version
v6.11.3
```

edemaine:
FWIW, here's how `cli.js` used to work:

```
git clone https://github.com/Khan/KaTeX
cd KaTeX
git submodule update --recursive --init
npm run dist          (this takes a long time)
./cli.js
```

However, recently, this stopped working: now I get

```
.../KaTeX/dist/katex.js:10
})(window, function() {
   ^

ReferenceError: window is not defined
[...]

PR #1401, opened 2018-06-02
Enable output modules to be used in browser or Node

edemaine:
This makes cli.js work again from the Node command line; see https://github.com/Khan/KaTeX/issues/894#issuecomment-393962739.

Workaround for bug reported in https://github.com/webpack/webpack/issues/6522

### DIFF
--- a/webpack.common.js
+++ b/webpack.common.js
@@ -57,6 +57,9 @@ function createConfig(target /*: Target */, dev /*: boolean */,
             library: target.library,
             libraryTarget: 'umd',
             libraryExport: 'default',
+            // Enable output modules to be used in browser or Node.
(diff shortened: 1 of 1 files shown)


kevinbarabash:
Good find.

nebrelbug:
How did you implement the workaround? I'm trying to author a library :)

edemaine:
@nebrelbug If you select the "Files Changed" tab in this pull request, you'll see the relevant changes. I believe we changed this line in a future PR as well, so you might also check the master branch for this file. If I recall, the linked webpack issue also describes the workaround.

nebrelbug:
@edemaine thanks! The webpack issue was a little unclear, but I understand now.

Issue #872, opened 2017-09-11
Avoid bundling overhead


kevinbarabash:
Webpack adds a little bit of overhead to each module whereas Rollup does fancy variable rewriting to avoid this.  As we continue to split up functions.js, buildHTML.js, and buildMathML.js into smaller modules, the amount of overhead will continue to grow.

obedm503:
this would be really cool. right now the dist build has repeated babel utilities which would be avoided if using rollup's fancy scope hoisting

kevinbarabash:
@obedm503 I know about rollup, but haven't used it.  How much work would it be to switch over?  Does it work with hot reloading?

obedm503:
didn't know you were using webpack. I guess the same thing could be achieved using the [ModuleConcatenationPlugin](https://webpack.js.org/plugins/module-concatenation-plugin/). that would remove the duplicated utilities

kevinbarabash:
Cool.  @obedm503 would definitely be interested in a PR for that change if have time.

obedm503:
looking into this. can you explain the build system? from what I gather, you're using webpack during dev but browserify for production?

kevinbarabash:
@obedm503 I moved fonts into a submodule recently and thought everything would work correctly for the publish.  It appears that isn't the case.  I'll look into the fonts issue tomorrow.

kevinbarabash:
@obedm503 can you check if there's anything in `submodules/katex-fonts`?  If there isn't please try running:
```
git submodule sync --recursive && git submodule update --init --recursive
```
Also, are you on Windows?  There's an issue with symlinks on that platform.

obedm503:
`prepublishOnly` now works! and no I'm not on windows.
with that out of the way, can you explain the build system?

kevinbarabash:
Hmm... it looks like we're using browserify + babelify for building distribution files.  That being said, webpack.config.js looks like it would build at least the unminified versions as well.  I feel like we should probably move off of browserify to webpack.

kevinbarabash:
@obedm503 I tried adding `ModuleConcatenationPlugin` but it isn't working b/c there are some files that aren't ES6 modules in the bundle, e.g.
```
        [40] ./submodules/katex-fonts/fonts/KaTeX_Main-BoldItalic.woff 78 bytes {0} [built]
             ModuleConcatenation bailout: Module is not an ECMAScript module
             ModuleConcatenation bailout: Module is not an ECMAScript module
```
and 
```
[...]

PR #1403, opened 2018-06-03
experimenting with rollup

kevinbarabash:
The minified bundle size using rollup is 240K vs. 253K for webpack.  The bundle size saving is 5% which isn't a lot.  There's still lots of work that would need to be done:
- [ ] getting this working with the dev server
- [ ] hot-loading
- [ ] handling css/less files


webpack.common.js
```diff
@@ -98,6 +99,7 @@ function createConfig(target /*: Target */, dev /*: boolean */,
             !dev && new MiniCssExtractPlugin({
                 filename: minimize ? '[name].min.css' : '[name].css',
             }),
+            new webpack.optimize.ModuleConcatenationPlugin(),
```
ylemkimon: Does `ModuleConcatenationPlugin` reduce build size?

kevinbarabash: I tried this, but it bails out.  See my comments in https://github.com/Khan/KaTeX/issues/872.

kevinbarabash:
Closing since this was just an experiment.  The more probably direction to reduce module overhead in our bundled code is to upgrade to babel 7 and use `ModuleConcatenationPlugin`.

PR #1407, opened 2018-06-04
Support \hdashline

ronkok:
Support `\hdashline` from package `arydshln`. This PR acts as a complement to PR #1395.

Similarly to #1395, the dashed line is rendered as `border-bottom-style: dashed;`. That does not exactly match the `dashsegment` and `dashgap` lengths in `arydshln`, but it does render black lines with sharp edges.


kevinbarabash:
@ronkok thanks for including `\hline` as well as `\hdashline` in the screenshot:
hdashline

ronkok:
@kevinbarabash You're welcome. Since `\hline` and `\hdashline` are now in the array screenshot, maybe we could delete the `\hline` screenshot?

src/environments/array.js
```diff
@@ -45,18 +45,22 @@ type ArrayEnvNodeDataIncomplete = {|
...
+    // Return an array. The array length = number of hlines.
+    // Each element in the array tells if the line is dashed.
+    const hlineInfo = [];
     parser.consumeSpaces();
-    while (parser.nextToken.text === "\\hline") {
+    let nxt = parser.nextToken.text;
+    while (nxt === "\\hline" || nxt === "\\hdashline") {
         parser.consume();
-        n++;
+        hlineInfo.unshift(nxt === "\\hdashline");
```
edemaine: Why is this `unshift` instead of `push`? In particular, why is the order reversed?

Efficiency wise, it'd be better to `push` and then reverse e.g. while returning. Though I can't inagine anyone having more than a few consecutive `\hline`s.

src/environments/array.js
```diff
@@ -187,12 +194,17 @@ const htmlBuilder: HtmlBuilder<"array"> = function(group, options) {
...
-        if (i >  1) {              // The first \hline doesn't add to height.
-            totalHeight += 0.25;
+    function setHLinePos(hlinesInGap: boolean[]) {
+        let n = 0;
+        while (hlinesInGap.length > 0) {
+            n += 1;
+            if (n > 1) {
+                totalHeight += 0.25;
+            }
+            hlines.push({pos: totalHeight, isDashed: hlinesInGap.pop()});
```
edemaine: Ah, here you are processing in reversed order. I'd find it more intuitive to store them in order, then use a `for` loop or `map` here.

ronkok:
> somewhere saying that the boolean stores whether it's dashed.

That's on line 53 of `array.js`. I couldn't think of a more prominent place to put it.

I'm picking up rhe other comments in the upcoming commit.

PR #1408, opened 2018-06-05
Port buildHTML to @flow.

marcianx:
Looking through this PR, some of these changes to `buildHTML`...seem ancient, and I was surprised they were not in yet. Looking back, I just realized that the last 21 PRs I sent were just so that this PR could finally finish. O_o

`buildMathML` should come in soon as well.

### DIFF
--- a/src/buildHTML.js
+++ b/src/buildHTML.js
@@ -61,6 +61,18 @@ const styleMap: {[StyleString]: StyleInterface} = {
 
 type Side = "left" | "right";
 
+const DomEnum = {
+    mord: "mord",
+    mop: "mop",
+    mbin: "mbin",
+    mrel: "mrel",
+    mopen: "mopen",
+    mclose: "mclose",
+    mpunct: "mpunct",
+    minner: "minner",
+};
+export type DomType = $Keys<typeof DomEnum>;
+
 /**
  * Take a list of nodes, build them in order, and return a list of the built
  * nodes. documentFragments are flattened into their contents, so the
@@ -73,7 +85,7 @@ export const buildExpression = function(
     expression: *[],
     options: Options,
     isRealGroup: boolean,
-    surrounding: [?string, ?string] = [null, null],
+    surrounding: [?DomType, ?DomType] = [null, null],
 ): HtmlDomNode[] {
     // Parse expressions into `groups`.
     const rawGroups: HtmlDomNode[] = [];
@@ -208,18 +220,6 @@ const getOutermostNode = function(
     }
     return node;
 };
-
-const DomEnum = {
-    mord: "mord",
-    mop: "mop",
-    mbin: "mbin",
-    mrel: "mrel",
-    mopen: "mopen",
-    mclose: "mclose",
-    mpunct: "mpunct",
-    minner: "minner",
-};
-export type DomType = $Keys<typeof DomEnum>;
 
 // Return math atom class (mclass) of a domTree.
 export const getTypeOfDomTree = function(
--- a/src/buildHTML.js
+++ b/src/buildHTML.js
@@ -10,14 +10,14 @@ import ParseError from "./ParseError";
 import Style from "./Style";
 import buildCommon from "./buildCommon";
 import domTree from "./domTree";
-import utils from "./utils";
+import utils, {assert} from "./utils";
 import {checkNodeType} from "./ParseNode";
 import {spacings, tightSpacings} from "./spacingData";
 import {_htmlGroupBuilders as groupBuilders} from "./defineFunction";
 
 import type Options from "./Options";
+import type {AnyParseNode} from "./ParseNode";
 import type {HtmlDomNode, DomSpan} from "./domTree";
-import type {StyleInterface} from "./Style";
 
 const makeSpan = buildCommon.makeSpan;
 
@@ -51,8 +51,7 @@ const isBinRightCanceller = function(
     }
 };
 
-type StyleString = "display" | "text" | "script" | "scriptscript";
-const styleMap: {[StyleString]: StyleInterface} = {
+const styleMap = {
     "display": Style.DISPLAY,
     "text": Style.TEXT,
     "script": Style.SCRIPT,
@@ -82,7 +81,7 @@ export type DomType = $Keys<typeof DomEnum>;
  * consisting type of nodes that will be added to the left and right.
  */
 export const buildExpression = function(
-    expression: *[],
+    expression: AnyParseNode[],
     options: Options,
     isRealGroup: boolean,
     surrounding: [?DomType, ?DomType] = [null, null],
@@ -92,7 +91,6 @@ export const buildExpression = function(
     for (let i = 0; i < expression.length; i++) {
         const output = buildGroup(expression[i], options);
         if (output instanceof domTree.documentFragment) {
-            // $FlowFixMe: should consist entirely of `symbolNode`s and `span`s.
             const children: HtmlDomNode[] = output.children;
             rawGroups.push(...children);
         } else {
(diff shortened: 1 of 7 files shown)


src/buildHTML.js
```diff
@@ -55,14 +69,20 @@ const styleMap = {
  * a partial group (e.g. one created by \color). `surrounding` is an array
  * consisting type of nodes that will be added to the left and right.
  */
-export const buildExpression = function(expression, options, isRealGroup,
-        surrounding = [null, null]) {
+export const buildExpression = function(
+    expression: *[],
+    options: Options,
+    isRealGroup: boolean,
+    surrounding: [?string, ?string] = [null, null],
```
ylemkimon: `surrounding` is one of `DomType`s.

marcianx: Ah, thanks! Fixed.

src/buildHTML.js
```diff
@@ -55,14 +81,20 @@ const styleMap = {
  * a partial group (e.g. one created by \color). `surrounding` is an array
  * consisting type of nodes that will be added to the left and right.
  */
-export const buildExpression = function(expression, options, isRealGroup,
-        surrounding = [null, null]) {
+export const buildExpression = function(
+    expression: *[],
```
ylemkimon: Is it currently impossible to type `expression` and

src/buildHTML.js
```diff
@@ -215,14 +258,19 @@ export const makeNullDelimiter = function(options, classes) {
  * function for it. It also handles the interaction of size and style changes
  * between parents and children.
  */
-export const buildGroup = function(group, options, baseOptions) {
+export const buildGroup = function(
+    group: *,
```
ylemkimon: `group` and

src/buildHTML.js
```diff
@@ -273,7 +321,7 @@ function buildHTMLUnbreakable(children, options) {
  * Take an entire parse tree, and build it into an appropriate set of HTML
  * nodes.
  */
-export default function buildHTML(tree, options) {
+export default function buildHTML(tree: *, options: Options): DomSpan {
```
ylemkimon: `tree`?

src/functions/delimsizing.js
```diff
@@ -224,11 +230,15 @@ defineFunction({
         if (hadMiddle) {
             for (let i = 1; i < inner.length; i++) {
                 const middleDelim = inner[i];
-                if (middleDelim.isMiddle) {
+                // Property `isMiddle` not defined on `span`. See comment in
+                // "middle"'s htmlBuilder.
+                // $FlowFixMe
+                const isMiddle: IsMiddle = middleDelim.isMiddle;
```
kevinbarabash: I find it really weird that `isMiddle` is not a boolean... we should probably change this property to be called simply `middle`.  Maybe add a TODO to make that change separately.

src/functions/supsub.js
```diff
@@ -175,7 +175,7 @@ defineFunctionBuilders({
         }
 
         // Wrap the supsub vlist in a span.msupsub to reset text-align.
-        const mclass = html.getTypeOfDomTree(base) || "mord";
+        const mclass = html.getTypeOfDomTree(base, "right") || "mord";
```
kevinbarabash: This change isn't necessary after re-enabling the param default.

src/buildHTML.js
marcianx: Sorry, sloppy me. I didn't go back through the stuff I'd already added types for way back when. Fixed.

src/buildHTML.js
marcianx: Done

src/buildHTML.js
marcianx: Done

src/buildHTML.js
```diff
@@ -71,23 +103,25 @@ export const buildExpression = function(expression, options, isRealGroup,
...
-        surrounding[1] && makeSpan([surrounding[1]], [], options),
+        surrounding[1] ? makeSpan([surrounding[1]], [], options) : null,
     ];
 
     // Before determining what spaces to insert, perform bin cancellation.
     // Binary operators change to ordinary symbols in some contexts.
     for (let i = 1; i < nonSpaces.length - 1; i++) {
-        const left = getOutermostNode(nonSpaces[i], "left");
+        // $FlowFixMe: Only the first and last can be nullable.
+        const nonSpacesI: HtmlDomNode = nonSpaces[i];
```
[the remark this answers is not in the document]
marcianx: `$FlowFixMe` just allows the assignment; flow would still type `nonSpaces` as `HtmlDomNode` and assume its type as that in the subsequent lines. In any case, I added a general-purpose `assert()` method in utils since I've wanted it more than once before and removed the `$FlowFixMe`.

src/buildHTML.js
```diff
@@ -215,14 +258,19 @@ export const makeNullDelimiter = function(options, classes) {
...
+): HtmlDomNode {
     if (!group) {
         return makeSpan();
     }
 
     if (groupBuilders[group.type]) {
         // Call the groupBuilders function
-        let groupNode = groupBuilders[group.type](group, options);
+        // $FlowFixMe
+        let groupNode: HtmlDomNode = groupBuilders[group.type](group, options);
```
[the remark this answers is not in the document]
marcianx: `HtmlBuilder<NodeType>` isn't the type union of all possibilities since it's not variant w.r.t. `NodeType` (by virtue of us computing `NodeValue` from it). Really, it should be `{[NodeType]: AnyHtmlBuilder}`. But it's probably not worth the effort. Almost all the typesafety we get is from the definitions `defineFunction` and `definedEnvironment`. Almost nothing else touches `_htmlGroupBuilders`; the couple methods that do expose a very strongly-typed signature.

src/functions/accent.js
```diff
@@ -19,7 +20,8 @@ export const htmlBuilder: HtmlBuilderSupSub<"accent"> = (grp, options) => {
     let base: AnyParseNode;
     let group: ParseNode<"accent">;
 
-    const supSub = checkNodeType(grp, "supsub");
+    // $FlowFixMe: Maybe due to https://github.com/facebook/flow/issues/6379
+    const supSub: ?ParseNode<"supsub"> = checkNodeType(grp, "supsub");
```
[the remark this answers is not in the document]
marcianx: Done

src/functions/supsub.js
marcianx: I added it because it wasn't obvious to me when it's omitted that the default behavior is "right". Is there any *obvious* reason to make "right" the default over, say, "left"? If not, I think having it explicitly would aid readability.

src/buildHTML.js
```diff
@@ -71,23 +103,25 @@ export const buildExpression = function(expression, options, isRealGroup,
...
     // spacing should go between atoms of different classes, and add dummy
-    // spans for determining spacings between surrounding atoms
-    const nonSpaces = [
-        surrounding[0] && makeSpan([surrounding[0]], [], options),
+    // spans for determining spacings between surrounding atoms.
+    const nonSpaces: (?HtmlDomNode)[] = [
+        surrounding[0] ? makeSpan([surrounding[0]], [], options) : null,
         ...rawGroups.filter(group => group && group.classes[0] !== "mspace"),
-        surrounding[1] && makeSpan([surrounding[1]], [], options),
+        surrounding[1] ? makeSpan([surrounding[1]], [], options) : null,
```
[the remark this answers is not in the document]
kevinbarabash: No action necessary here.  Currently we only know the atom type for a group until after we build the HTML for it.  Eventually, it would be nice if we could do:
```
ParseNode tree => LayoutNode tree => HtmlNode tree
```
where
```
LayoutNode =
   | Box
   | Rule
   | Kern
   | Glue
   | Glyph
```
The benefit of this would be being able to have different rendering backends so that we could do any of the following:
```
LayoutNode tree => HtmlNode tree
LayoutNode tree => SvgNode tree
LayoutNode tree => canvas render
```
Having a tree of `LayoutNode`s would allow us to do any post-processing such as bin cancellation on the `LayoutNode` tree once instead of having to have separate copies of the code for each of the final trees.

kevinbarabash:
> Looking through this PR, some of these changes to buildHTML...seem ancient, and I was surprised they were not in yet. Looking back, I just realized that the last 21 PRs I sent were just so that this PR could finally finish. O_o

I appreciate how deliberate you've been about these changes.  I'm glad that we're getting a place where we're able to do larger refactorings safely.

> buildMathML should come in soon as well.

Looking forward to it.

PR #1410, opened 2018-06-05
Implement \colon as amsmath does

### LINKED ISSUE #152: Support for ties (~)
Since we don't support breaking at all yet I think this can just be equivalent to `\ `.

ronkok:
Fixes #152.

### DIFF
--- a/src/macros.js
+++ b/src/macros.js
@@ -289,6 +289,11 @@ defineMacro("\\varPhi", "\\mathit{\\Phi}");
 defineMacro("\\varPsi", "\\mathit{\\Psi}");
 defineMacro("\\varOmega", "\\mathit{\\Omega}");
 
+// \renewcommand{\colon}{\nobreak\mskip2mu\mathpunct{}\nonscript
+// \mkern-\thinmuskip{:}\mskip6muplus1mu\relax}
+defineMacro("\\colon", "{\\mkern2mu\\mathpunct{}" +
+    "\\mathchoice{\\mkern-3mu}{\\mkern-3mu}{}{}{:}\\mkern7mu}");
+
 // \newcommand{\boxed}[1]{\fbox{\m@th$\displaystyle#1$}}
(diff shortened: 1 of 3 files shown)


test/katex-spec.js
```diff
@@ -116,6 +116,12 @@ describe("A punct parser", function() {
     });
 });
 
+describe("A \\colon builder", function() {
+    it("should not fail", function() {
+        expect("a \\colon b").toBuild(strictSettings);
+    });
+});
+
```
ylemkimon: I'm not sure whether we should have seperate test for a macro.

edemaine: FWIW, the "macro expander" section of tests has a few tests along these lines. But probably the screenshot test is more important here.

src/macros.js
```diff
@@ -289,6 +289,11 @@ defineMacro("\\varPhi", "\\mathit{\\Phi}");
 defineMacro("\\varPsi", "\\mathit{\\Psi}");
 defineMacro("\\varOmega", "\\mathit{\\Omega}");
 
+// \renewcommand{\colon}{\nobreak\mskip2mu\mathpunct{}\nonscript
+// \mkern-\thinmuskip{:}\mskip6muplus1mu\relax}
+defineMacro("\\colon", "{\\mkern2mu\\mathpunct{}" +
+    "\\mathchoice{\\mkern-3mu}{\\mkern-3mu}{}{}{:}\\mkern7mu}");
+
```
edemaine: `7mu` should be `6mu`. (`plus` does not add, it states allowable growth when stretching.)

I think you should also add `\nobreak` at the beginning (which prevents a line break before the macro), and I also feel like the outer `{}` should be removed to match LaTeX... Could you test it in this form to make sure there are no bad breaks?

As a minor thing, I'd change `\mkern` to `\mskip` to match the macro definition as closely as possible, but they are of course identical at this point.

PR #1416, opened 2018-06-08
Remove redundant \lbrack and \rbrack

ronkok:
PR #1069 added macros for `\lbrack` and `\rbrack`. But KaTeX already supported `\lbrack` and `\rbrack` with code in `delimiter.js`. This PR removes the redundant macros.

### DIFF
--- a/src/delimiter.js
+++ b/src/delimiter.js
@@ -219,12 +219,12 @@ const makeStackedDelim = function(
         top = "\\Uparrow";
         repeat = "\u2016";
         bottom = "\\Downarrow";
-    } else if (delim === "[" || delim === "\\lbrack") {
+    } else if (delim === "[") {
         top = "\u23a1";
         repeat = "\u23a2";
         bottom = "\u23a3";
         font = "Size4-Regular";
-    } else if (delim === "]" || delim === "\\rbrack") {
+    } else if (delim === "]") {
         top = "\u23a4";
         repeat = "\u23a5";
         bottom = "\u23a6";
@@ -480,7 +480,7 @@ const makeSqrtImage = function(
 // There are three kinds of delimiters, delimiters that stack when they become
 // too large
 const stackLargeDelimiters = [
-    "(", ")", "[", "\\lbrack", "]", "\\rbrack",
+    "(", ")", "[", "]",
     "\\{", "\\lbrace", "\\}", "\\rbrace",
     "\\lfloor", "\\rfloor", "\u230a", "\u230b",
     "\\lceil", "\\rceil", "\u2308", "\u2309",
--- a/src/functions/delimsizing.js
+++ b/src/functions/delimsizing.js
@@ -35,7 +35,7 @@ const delimiterSizes = {
 };
 
 const delimiters = [
-    "(", ")", "[", "\\lbrack", "]", "\\rbrack",
+    "(", ")", "[", "]",
     "\\{", "\\lbrace", "\\}", "\\rbrace",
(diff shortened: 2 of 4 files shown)


edemaine:
Can you hold off on merging this until I can review? I'm a little worried about whether this works correctly with optional arguments.

edemaine:
OK, I can confirm this doesn't break anything.  For example:
```latex
\begin{array} ... \\ \lbrack 1ex \rbrack ... \end{array}
```
does *not* trigger an optional argument in LaTeX (because it uses `\futurelet` which doesn't do expansion).

I do wonder whether this is the right approach, though.  Why have multiple lines in `delimiter.js` and `functions/delimsizing.js` handling `\lbrack` and `\rbrack` when the macro would suffice?  That is, why not go the other way?

ronkok:
> when the macro would suffice?

There's a lot of merit to that argument. There's also something to be said for having a list of all the delimiters [in one place](https://github.com/Khan/KaTeX/blob/master/src/functions/delimsizing.js#L37). I don't have strong feeling either ways. I chose the redundant code that came last chronologically.

edemaine:
OK, I found a more legitimate argument: in your branch, `\lbrack` and `\rbrack` don't work in text mode (because there are no delimiters in text mode), but they should.

ronkok:
Okay, I'll put the macro back and remove the redundancy in `delimiter.js` and `delimsizing.js`. Are there any other delimiter rearrangements needed?

ronkok:
I've now reverted the macros and removed redundant code from `delimiter.js` and `delimsizing.js`. Here's confirmation that `[`, `]`, `\lbrack`, and `\rbrack` still work, unsize, with `\big`, and with `\left`:
brack

ylemkimon:
@edemaine @ronkok Too late, but we have `\lbrack` and `\rbrack` in `symbols.js`:
https://github.com/Khan/KaTeX/blob/1dc261298ed327f7451095fe5ccf4e3dcc3ec589/src/symbols.js#L619-L620

ronkok:
@ylemkimon Those two lines of code need to stay there, or else `\left [` and `\right ]` will not work. (I tested that.) The code in `delimsizing.js` is pretty tightly bound to `symbols.js`.

ylemkimon:
@ronkok But with symbol definition, isn't macro definition redundant?

ronkok:
> isn't macro definition redundant?

That was my belief at the start of this PR. But @edemaine has found a quirk involving text mode that makes the macros non-redundant. So the macros stay. I've removed as much code as I can without breaking something.

edemaine:
Oh, are there still symbol definitions for lbrack and rbrack? If so, those can go. Macros already take priority.

edemaine:
I guess we could have just added lbrack and rbrack as text symbols too...

ronkok:
No, the symbol definitions cannot go. If they are deleted, then `\left` and `\right` stop working for`[`, `]`, `\lbrack`, and `\rbrack`.

In `delimsizing.js`, the `\left` and `\right` functions each make a `checkDelimiter(...)` call, which goes through a long and winding path to verify type via those two lines of code.  The typing system doesn't understand that the call could be coming from a macro. It assumes that the call is coming from `symbols.js`.

ronkok:
> we could have just added lbrack and rbrack as text symbols too...

That would have avoided the sort of confusion we're going through right now. Maybe it's still worth doing to avoid future confusion by others.

edemaine:
Looking at the code more, I guess it would be consistent with the other symbols. Sorry for the hassle.

PR #1418, opened 2018-06-09
Support \pmb

ronkok:
`\pmb` is an AMS function that authors can use to simulate a bold font even when the KaTeX fonts do not contain a bold glyph. So it would be useful in, for example, `\pmb{\mu}`.

It works by typesetting three copies of the given argument with small offsets. So it's a hack. It isn't as good as a real font glyph. But it's better than nothing.


ronkok:
Here's how `\pmb` compares to regular math.

pmb

ylemkimon:
Too bad IE9 doesn't support CSS `text-shadow` property.

ronkok:
IE9 doesn't support CSS `flex` either, but @kevinbarabash has started to move in the `flex` direction. I expect that KaTeX will, in the not too distant future, deprecate IE9.

edemaine:
I think it's cool to reproduce this functionality in the way that (AMS)LaTeX does.  One high-level question: is there a reason this isn't easy to do with a macro and e.g. `\mathllap`/`\mathrlap`?  I haven't tried, but something like
```latex
\hspace{.4mu}\mathllap{#1\hspace{.4mu}}#1\mathrlap{\hspace{.4mu}#1}\hspace{.4mu}
```?

**Comment by ronkok** [comment]
`\pmb` is another function that applies `\binrel` spacing to the result. I don't know a good way to do that in a macro. If not for that, the rest would be easy to do in a macro.

**Comment by ronkok** [comment]
> the rest would be easy to do in a macro.

Again, with an exception for the MathML. The current PR provides a reasonable MathML. A macro with three copies would be pretty well obfuscated.

**Comment by edemaine** [comment]
Yeah. I still think a `\mathml{LaTeX code}{MathML equivalent}` macro would be a good way to go here, though I realize sanitization will be a bit of a challenge.  Anyway, maybe now is not the time.

**Comment by kevinbarabash** [comment]
@edemaine  how would `\mathml{LaTeX code}{MathML equivalent}` be used in the case of `\pmb`?

**Comment by edemaine** [comment]
Thinking more about it, I think I have finally figured out a proposal, and made it issue #1452, as this PR is probably not the place to discuss it.  (Though there is the question of whether this PR should block on it.)

**Comment by ronkok** [comment]
@ylemkimon Why is this PR labeled as "blocked"? I don't think it is blocked. The `\binrel` requirement makes this PR unfeasible as a macro regardless of the status of PR #1454.

**Comment by edemaine** [comment]
I could try to implement `\@binrel`...

**Comment by ylemkimon** [comment]
@ronkok Sorry, I've remove the label. (Anyway, #1454 is merged.) Would it be possible to control spacing at the `buildExpression` like `enclosing` class #1437?

**Comment by ronkok** [comment]
Okay, now `\pmb` is blocked by #1487. That was fast.

**Comment by edemaine** [comment]
Sorry @ronkok! I hope you don't mind this preference toward macros, and the possibility that it will cause this PR to be largely rewritten. In fact, this PR has served to make some very useful tools come into existence, which I hope will also make it easier to add features like this in the future. So, definitely not time wasted!

**Comment by ronkok** [comment]
@edemaine  Not a problem. `\pmb` will be a pretty simple macro to write given the new tools.

**Comment by kevinbarabash** [comment]
✨ "blocked" begone ✨

**Comment by ronkok** [comment]
The latest commit changes `\pmb` to a macro. A big thank you to @edemaine, who made macros capable of dealing with `\binrel` and improved MathML.

File: src/macros.js [hunk 1]
```diff
@@ -659,6 +659,12 @@ defineMacro("\\mod", "\\allowbreak" +
     "\\mathchoice{\\mkern18mu}{\\mkern12mu}{\\mkern12mu}{\\mkern12mu}" +
     "{\\rm mod}\\,\\,#1");
 
+// \pmb    --   A simulation of bold.
+// It works by typesetting three copies of the argument with small offsets.
+// Ref: a rather lengthy macro in ambsy.sty
+defineMacro("\\pmb", "\\html@mathml{\\@binrel{#1}{\\mathrlap{\\mathrlap{#1}" +
+    "{\\mkern0.4mu\\raisebox0.4mu{$#1$}}}{\\mkern0.8mu#1}}}{\\mathbf{#1}}");
```
edemaine: I think it's a fluke of KaTeX that `\raisebox` works with an unbraced argument `0.4mu`.  LaTeX doesn't work that way (while TeX's `\raise` does), and we should probably fix that eventually.  So I'd be happier if we replaced with `\raisebox{0.4mu}`.

edemaine: I also feel like some of the other braces on line 666 are unnecessary, at least the first one (because of the one within `\raisebox`).  Hmm, maybe the others are necessary in case of macros in the argument.  Maybe you can also add this test:

```js
expect("\\def\\x{1}\\pmb{\\x\\def\\x{2}}").toParseLike("\\pmb{1}")
```

ronkok: Done.

src/macros.js
```diff
@@ -659,6 +659,12 @@ defineMacro("\\mod", "\\allowbreak" +
     "\\mathchoice{\\mkern18mu}{\\mkern12mu}{\\mkern12mu}{\\mkern12mu}" +
     "{\\rm mod}\\,\\,#1");
 
+// \pmb    --   A simulation of bold.
+// It works by typesetting three copies of the argument with small offsets.
+// Ref: a rather lengthy macro in ambsy.sty
+defineMacro("\\pmb", "\\html@mathml{\\@binrel{#1}{\\mathrlap{\\mathrlap{#1}" +
+    "\\mkern0.4mu\\raisebox{0.4mu}{$#1$}}{\\mkern0.8mu#1}}}{\\mathbf{#1}}");
```
edemaine: This is much easier to read now, thanks!

I have one remaining question: why is the `\mathrlap{#1}` nested inside another `\mathrlap`?  I imagine it could also go at the very beginning, like `\mathrlap{#1}\mathrlap{\mkern0.4mu...}`.  Any particular reason to prefer one over the other?

ronkok:
I agree that the new screenshots look better. What's weird is that the Chrome screenshot failed but Firefox passed.

PR #1423, opened 2018-06-12
Update READMEs to v0.10.0-beta

edemaine:
Problem reported by @ronkok here: https://github.com/Khan/KaTeX/pull/1348#issuecomment-396676971

Upon analysis, two things happened:
1. I used the 1-argument form of `release.sh`, because I didn't need to change the next version.  But then the release script doesn't update the READMEs at all in the master branch. We might consider fixing this.
2. I don't have permission to push directly to the master branch, so I believe this part of the script can't work for me anyway (having just tried it manually).

As a stop-gap for this release, I made a PR grabbing the relevant files from
https://github.com/Khan/KaTeX/commit/8624d4ccffbf8543ec87e75b24ed9551a7c16714


kevinbarabash:
I didn't even realize the script supported `1-argument`. 😂

kevinbarabash:
> I don't have permission to push directly to the master branch, so I believe this part of the script can't work for me anyway (having just tried it manually).

@edemaine I'm confused since although `master` is "protected" there's no restrictions on who can push to `master` out of those people who have write access.  Also, if you didn't have push access how did https://github.com/Khan/KaTeX/commit/8624d4ccffbf8543ec87e75b24ed9551a7c16714 get on `master`?

edemaine:
@kevinbarabash https://github.com/Khan/KaTeX/commit/8624d4ccffbf8543ec87e75b24ed9551a7c16714 is a tag, not on master.

kevinbarabash:
I didn't even realize that was possible.

PR #1427, opened 2018-06-13
Use u2016 for Vert

bpiwowar:
See issue comment [https://github.com/Khan/KaTeX/issues/286#issuecomment-118419984](286) : \Vert is associated with ∥ (u2225), which can be slanted or not, but should be with ‖ (u2016).


ronkok:
@bpiwowar, You make a strong case that U+2016 should be the Unicode code point used for `\lVert`. Unfortunately, the KaTeX fonts do not contain a glyph at code point U+2016. So the change you have proposed to `symbols.js` is insufficient by itself. As submitted, this will just cause KaTeX to issue a console warning that font metrics are unavailable for "‖", and then it will render a glyph from some system font, not the KaTeX fonts.

For this to be complete, one would have to change `symbols.js`, as you have done, and also add a glyph to the KaTeX_Main-Regular font in the KaTex font [submodule](https://github.com/KaTeX/katex-fonts).

ylemkimon:
Closing this as a `won't fix`. Please feel free to reopen or create a new pull request. Thank you for your contribution!

PR #1430, opened 2018-06-14
Support \oiint and \oiiint

### LINKED ISSUE #392: Support for \oiint
Are there any plans to support \oiint ?

ronkok:
This PR supports \oiint and \oiiint by overlaying SVG ovals onto the glyphs from \iint and \iiint.

_**Edit by @ylemkimon:** This pull request fixes #392._


ronkok:
Here's screenshot:

oiint

ronkok:
I submitted this PR too soon. It doesn't position `sub` elements correctly. I'll fix it.

kevinbarabash:
Nice.  I was going to try to add glyphs to the fonts, but this is better b/c it should save space.

ronkok:
There. I've got the `sub` fixed and tests are in. This is ready for review.

ronkok:
> this is better b/c it should save space.

Yes, it will save some space. That comes at the cost of several exceptions cluttering up the code. I would be happy to see someone come along and replace this with font glyphs.

kevinbarabash:
> Yes, it will save some space. That comes at the cost of several exceptions cluttering up the code. I would be happy to see someone come along and replace this with font glyphs.

I guess everything has it's drawbacks.  I agree that fewer exceptions is probably the right call here.

kevinbarabash:
Here's the associated katex-fonts issues: https://github.com/KaTeX/katex-fonts/issues/20.  @ronkok I'm curious which package you got the geometry for the ellipses from.

ronkok:
I liked the look of the MnSymbol ovals. Nicely curved, not too pinched at the end. So I started with its (public domain) path geometry. Of course, MnSymbol has different font glyphs than KaTeX, so I had to stretch the Size2 ovals and shrink the Size1 ovals. I think they came out pretty well.

kevinbarabash:
@ronkok they look good!

kevinbarabash:
@ylemkimon thanks for reviewing.

PR #1434, opened 2018-06-16
use expect.toMatchObject

kevinbarabash:
`jest` has some pretty fancy matchers so I thought I'd try one out.  What do people think?

### DIFF
--- a/test/katex-spec.js
+++ b/test/katex-spec.js
@@ -183,9 +183,7 @@ describe("A subscript and superscript parser", function() {
     });
 
     it("should produce supsubs for superscript", function() {
-        const parse = getParsed("x^2")[0];
-
-        expect(parse).toMatchObject({
+        expect(getParsed("x^2")[0]).toMatchObject({
             type: "supsub",
             value: {
                 base: expect.any(Object),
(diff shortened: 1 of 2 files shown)


test/katex-spec.js
```diff
@@ -185,37 +185,61 @@ describe("A subscript and superscript parser", function() {
...
-        expect(parse.value.base).toBeDefined();
-        expect(parse.value.sup).toBeDefined();
-        expect(parse.value.sub).toBeUndefined();
+        expect(parse).toEqual(
+            expect.objectContaining({
+                type: "supsub",
+                value: expect.objectContaining({
+                    base: expect.any(Object),
+                    sub: undefined,
+                    sup: expect.any(Object),
```
kevinbarabash: TBH, I'm not sold on the syntax.

test/katex-spec.js
```diff
@@ -185,37 +185,53 @@ describe("A subscript and superscript parser", function() {
     it("should produce supsubs for superscript", function() {
         const parse = getParsed("x^2")[0];
 
-        expect(parse.type).toBe("supsub");
-        expect(parse.value.base).toBeDefined();
-        expect(parse.value.sup).toBeDefined();
-        expect(parse.value.sub).toBeUndefined();
+        expect(parse).toMatchObject({
```
kevinbarabash: If parse contains extra keys, they're ignored which is what we want in this test.

kevinbarabash:
I'm going to close this for now and will re-open once we've moved to `yarn`.

PR #1436, opened 2018-06-16
Add mhchem extension

ronkok:
~~This is a proposal more than it is a pull request. Or perhaps it’s a proof of concept. The proposed code allows KaTeX to interact with a proposed `mhchem` extension. Such a `mhchem` extension would be installed in a way similar to the [auto-render](https://github.com/Khan/KaTeX/tree/master/contrib/auto-render) extension or the [copy-tex](https://github.com/Khan/KaTeX/tree/master/contrib/copy-tex) extension. That is, web site administrators would list the `mhchem-for-katex.js` file in the `<head>` of their page.~~

~~I invite comments. I expect that @edemaine can write a better helper function to recreate the argument to be passed to the extension. I’m sure that @mhchem can improve on the edits that I will suggest to the `mhchem` code.~~

~~It’s pretty simple, really. `mhchem` expands a macro. So I’ve written a suggested wrapper around the mhchem code that exposes a object with a single method: `mhchem.expand(str: string, stateMachine: "ce"|"pu") : {expansion: string, errorMsg: string}`.~~

~~The code in this PR checks for existence of the object, then calls the function.~~

General revision (Nov 3): This PR adds support for the `\ce` and `\pu` functions from the `mhchem` package. However, this PR makes no revisions to any core KaTeX files. Instead, it exploits the fact that KaTeX exposes the `defineMacro` function. All the `mhchem` code is a separate file. So web page administrators will be able to choose whether or not to include `mhchem` as an extension.

As in the original version of this PR, the code is mostly a copy of the `mhchem` package for MathJax. I still have made only four modifications, as described just below.


kevinbarabash:
@ronkok this is super cool!  I'm glad that you were able to adapt the MathJax extension without too much work.  Could you post some example renderings?

ronkok:
Screenshots:
ce1
ce2
ce3
ce4
ce5
ce6

mhchem:
This is splendid!

What I see here, seems very reasonable to me. So, I would leave the creation of the first version up to you. You decide, where to place the files, how to include them, etc. After that, I'd like to have control over the mhchem core for fixes and updates. You decide, how you would like to have that. Maybe a file at `KaTeX/contrib/mhchem/mhchem.js` where I can file pull requests? Or a completely separate repository? Maybe with your standard minification, maybe with mine (that is specially tweaked). You decide. I don't think, I will need a veto.

I am looking forward to having mhchem support in KaTeX and working with you.

ronkok:
@mhchem That sound very reasonable to me. Let's give @kevinbarabash a chance to review this PR. If he approves, then I'll work up a PR to create an `mhchem` extension arranged in a manner similar to [copy-tex](https://github.com/Khan/KaTeX/tree/master/contrib/copy-tex). After I've written the first version of the extension, you can then maintain it.

ylemkimon:
As most of macro expansions don't need to access KaTeX internals and `defineMacro` is already exposed, I think we can easily add plugins like `mhchem`!

kevinbarabash:
@mhchem I think having things live at `KaTeX/contrib/mhchem/mhchem.js` makes sense.  @ronkok can you include the modified mhchem.js you created as part of this PR since the current code isn't much use without it.

src/macros.js
```diff
@@ -624,6 +624,50 @@ defineMacro("\\hspace", "\\@ifstar\\@hspacer\\@hspace");
...
+    }
+};
+
+defineMacro("\\ce", function(context) {
+    return mhChemExpansion(context, "ce");
+});
+
+defineMacro("\\pu", function(context) {
+    return mhChemExpansion(context, "pu");
+});
```
ylemkimon: I think macro definitions should also be in `contrib` too.

ronkok: > I think macro definitions should also be in `contrib` too.

I would like that, too, so let's discuss what would be required to make that happen. At present, KaTeX exposes `__defineMacro: defineMacro`, but that only enables one to define a macro in the KaTeX macro language. KaTeX does not support branching macro code such as `\if … \else … \fi`, so the complexity of `mhchem` is out of reach by that method.

I needed a way to (1) capture the functions `\ce` and `\pu`, (2) reconstruct the string in the argument to those functions, (3) pass the string to the external `mhchem` extension, and (4) receive the expanded macro or an error message. This PR does that.

Maybe I should have added an abstraction layer that would enable other similar extensions to do something similar. Such an abstraction layer would expose a `_defineComplexMacro` callback and would enable `complexMacro.expand(argString: string, functionName: string) : {expansion: string, errorMsg: string}`  Then I could have used that exposed capability to define `\ce` and `\pu` in the `contrib` folder

Maybe that's what's people mean when they talk of a plugin architecture. I don't know. I'm out of my depth here. There are different ways to do this and I'm probably not the best person to do decide which way is best. That's why I opened this PR by describing it as a proof of concept, not a finished product. I would welcome instruction. If someone wanted to step in and define that proposed callback function, I would be fine with that, too.

edemaine: `__defineMacro` works just like `defineMacro`.  Its second argument can be a function, as above.

ronkok: > `defineMacro`. Its second argument can be a function, as above. 

I might be able to work with that. Let me get back to you.

ronkok: Update: I can indeed pass a function in `_defineMacro`, just as @edemaine suggests.  In fact, no modifications are necessary to `macros.js` at all. It works fine.

Before, instructions to the web site administrator were to place the reference to `mhchem.js` in the `<head>` of the HTML page. As now proposed, they will have to do that, and also  write the following into their KaTeX rendering options:

```html
macros: {
  "\\ce": function(context) {return mhchem.expand(context.consumeArgs(1)[0], "ce")},
  "\\pu": function(context) {return mhchem.expand(context.consumeArgs(1)[0], "pu")},
  "\\tripledash": "\\vphantom{-}\\raisebox{2mu}{$\\mkern2mu\\tiny\\text{-}\\mkern1mu\\text{-}\\mkern1mu\\text{-}\\mkern2mu$}"
}
```

In exchange for that extra installation complexity, we get cleaner code in `macros.js`. Is that a good trade off?

mhchem: I don't like it. In particular, the `\\tripledash` part is part of internals a user should not never see. Also, the other definitions do not make me happy. What will happen when you will change the internals some time in the future?

ronkok: I don't like it much either. And now that I look back, I think that @ylemkimon wanted to suggest something to move `\ce` definition into `contrib`, not into a rendering option. But I don't know how to accomplish such a thing. If anyone can tell me how, I'll take a crack at it. 

Otherwise, I think we are better off to keep the code in `macros.js`.

ylemkimon: Is it impossible to do following in the extension:
```js
katex.__defineMacro("\\ce", function(context) {
    return mhChemExpansion(context, "ce");
});

katex.__defineMacro("\\pu", function(context) {
    return mhChemExpansion(context, "pu");
});
```
?

ronkok: @ylemkimon Thank you for the instruction!

ronkok:
The latest commit moves all the code from `macros.js` to `mhchem.js`.  Now this PR makes no changes whatsoever to core KaTeX. All the code is in the extension. Also, no rendering options need be defined by a web site administrator. All the macros are defined in `mhchem.js`. 

Thank you to @ylemkimon and @edemaine for instruction on how to better arrange things. It's a better PR now.

ronkok:
The KaTeX linter says that nearly every line of `mhchem.js` has a problem. Acturally, `mhchem.js` is written to style standards that are good, but different than the KaTeX standards. Is there a way to modify the `eslint` script that makes an exception for the mhchem extension?

edemaine:
A couple other quick comments:
* Why is `mhchem.min.js` in the repo?  Presumably it should instead be generated?
* It seems that mhchem is licensed under Apache 2.0, while KaTeX is licensed under MIT.  I think some care is needed here, though I'm not an expert on the Apache license.  Maybe @mhchem can comment.

ronkok:
> Why is `mhchem.min.js` in the repo? 

I'll go further. There is no particular reason for `mhchem.js` to be in the KaTeX repo.  @mhchem could take possession of the `mhchem.js` file, keep it somewhere in `https://github.com/mhchem/` and push it to a CDN. It would run, right now, with no revision needed to KaTeX.

In the KaTeX repo, there would still be a `mhchem` extension, but it would contain only a README page that pointed to `mhchem.min.js` on its CDN. I'd also add `\ce` and `\pu` to the KaTeX function support page.

In that arrangement, there are no lint issues, no licensing conflicts, and @mhchem gets to keep control over his code.

Comments?

edemaine:
I would lean more toward just the main mhchem module being "off site", and just the KaTeX-related tweaks being here.  I was wondering whether `mhchem.js` could be obtained from a Git submodule.  This would require the upstream copy to have a different or more tweakable loading strategy.

Personally, I serve KaTeX files from my own server, not CDN.  (Debatable whether that's a good idea.)  This would be one reason that it'd be more convenient to have things in the KaTeX distribution, but avoiding the main duplication would be nice.

ronkok:
> main mhchem module being "off site", and just the KaTeX-related tweaks being here.

The previously existing version of `mhchem.js` is a single file that contains a fair amount of MathJax-specific glue code. It turned out to be fairly easy to remove those parts and replace with code that ties into KaTeX's `defineMacro()`.  But there were no "main mhchem" files that contain non-MathJax material. 

> Personally, I serve KaTeX files from my own server. 

@mhchem makes the `mhchem.js` file available in the `https://github.com/mhchem/` repo. It's just as easy to download them from there as from here. And we could write a link into the KaTeX `contrib/mhchem` README page .

mhchem:
> But there were no "main mhchem" files that contain non-MathJax material.

Which could be created if need be.

To repeat myself: It's your decision where the files live. I don't have a preference. (As long as I can keep the behavior of mhchem for MathJax and mhchem for KaTeX aligned, I am happy.)

ronkok:
> It's your decision where the files live.

So I will confess a personal interest. We now have a `mhchem.js` that works nicely with KaTeX. We've completed the part of the job at which I am any good. As soon as we move away from Javascript or CSS and into infrastructure, then we move into areas in which I know very little. And so my personal interest is that any path is okay, so long as someone else does it.

kevinbarabash:
> > But there were no "main mhchem" files that contain non-MathJax material.
> Which could be created if need be.

That seems like best path forward.  If this module was published then it could be required by the MathJax plugin and the KaTeX extension.  @mhchem how much would is involved in this?

mhchem:
I could create it, but please help me unterstand what is actually needed. I think both, MathJax and KaTeX, want to have one single minified JavaScript file for the mhchem extension. So how would a "core mhchem" help in that case? Wouldn't it be copied into the extension file anyway? Then I don't see a reason for a separate core file. Having 3 files, would create the need of another abstraction layer, more network traffic, delays. So, if the result should be one JS file for mhchem for KaTeX, I would rather have that file embed the mhchem core. To sync the extension for MathJax and KaTeX would be a matter of using a visual diff tool and skipping the frameword-specific part.

(BTW, for mhchem for MathJax, I do the minification using a hand-crafted pre-minification step that leads to a much smaller minified file.)

edemaine:
What I imagine is a JavaScript module, ideally published via NPM, that exposes what KaTeX and MathJax extensions need of the API from mhchem.  The extensions include this module and add the relevant hooks to the host system (KaTeX or MathJax).  Yes, there are 3 source files, but the idea is to use a modern build system to compile them into single files.  I hope you're not making `mhchem.min.js` by hand 😄 so you're already using a build system, but you might need a more complicated one if you're also maintaining the MathJax extension.

For an example, we already do this with all our `contrib` extensions.  See for example [auto-render](https://github.com/Khan/KaTeX/tree/master/contrib/auto-render), which is split into two modules, but end up compiling to a single `auto-render.js` and `auto-render.min.js` (via webpack).  So it'd be a no-brainer on the KaTeX side; we already build `contrib` extensions by hand, so I'd imagine just `require('mhchem')` within `contrib/mhchem` to import the generic code, and add the KaTeX hooks.  This would get compiled to a single module that e.g. ends up on CDNs, but there's no copying or diffing of code going on.  @mhchem could release a new mhchem generic module, and we could update to build with that version after testing.

mhchem:
I understand. 
(BTW, I do neither build `min.js` by hand, nor do I have a build system. 😃)

But what do we do with the `texify` stage of the mhchem core? The part where MathJax needs `\rlap` and `\raise` and KaTeX needs `\mathrlap` and `\raisebox` (and the other 99% are identical)? How would you handle these cases? I would like to avoid a `isKaTeX` evaluation at runtime.

-----

On another page: Are you sure that `// Recreate the argument string from KaTeX's array of tokens` from the top of this page really works correctly?
```
        // Eliminate spaces between any function and {
        str = str.replace(/\s(?=\{)/g, "");
```
This looks like it would fail on things like `\ce{{marine water} - {water} -> {...}}`

ronkok:
> Are you sure that [Recreate the argument string from KaTeX's array of tokens] really works correctly?

You're right.  There's a bust there. `context.consumeArgs` eats spaces that `mhchem` treats as significant. 

I think we need to write  a new `context.argString(…)` method.  As a argument, it would take a number that indicates the number of macro arguments. It would return an array of strings, one for each macro argument. It would preserve the spaces.

@edemaine You've done most of the development lately in `MacroExpander.js`. Do you mind if I take a crack at this proposed method or do you wish to reserve that to yourself?

ylemkimon:
@ronkok Another way is to check tokens' `loc` and pad with spaces.

ronkok:
@ylemkimon Thanks for the tip! The latest commit does indeed check each token's `loc` and it works like a charm.

ylemkimon:
What version of `mhchem` are you working on? I'm currently looking at [`v3.3.0`](https://github.com/mhchem/MathJax-mhchem/blob/master/unpacked/mhchem.js) of `mhchem`, and it seems to be different.

ylemkimon:
How about:
(i) create a repo, `KaTeX-mhchem` in @mhchem 
(ii) grant push access to some of KaTeX contributors
(iii) set as a submodule under `contrib/mhchem` in the KaTeX repo
?

This way, @mhchem has full control of the code and we can easily apply KaTeX changes to the extension and incorporate it into our build/release system. And when we have stable API and plugin architecture, we can just delete the submodule.

ronkok:
The latest commit updates the `mhchem.js` file to `mhchem` v3.3.0. I have once again changed the wrapper code, changed `\rlap` to `\mathrlap`, edited the `\raise` instances, and edited the reaction arrows.

I also noticed some `\color` code in there. It is not documented in the mhchem [Manual](https://mhchem.github.io/MathJax-mhchem/). I think that @mhchem intends to remove it. I did not touch the `\color` code because I'm not quite sure which of them are input and which are output. If we ever intend for people to use color we'll have to change the output `\color` instances to `\textcolor`.

contrib/mhchem/index.html
```diff
@@ -0,0 +1,752 @@
...
+<!--To text this example, place mhchem.js and this file into the same folder
+    and run this file.
+  -->
+<html>
+  <head>
+    <meta charset="UTF-8">
+    <title>mhchem test</title>
+    <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/katex@0.10.0-beta/dist/katex.min.css" integrity="sha384-9tPv11A+glH/on/wEu99NVwDPwkMQESOocs/ZGXPoIiLE8MU/qkqUcZ3zzL+6DuH" crossorigin="anonymous">
+    <script src="https://cdn.jsdelivr.net/npm/katex@0.10.0-beta/dist/katex.min.js" integrity="sha384-U8Vrjwb8fuHMt6ewaCy8uqeUXv4oitYACKdB0VziCerzt011iQ/0TqlSlv8MReCm" crossorigin="anonymous"></script>
+    <script src="https://cdn.jsdelivr.net/npm/katex@0.10.0-beta/dist/contrib/auto-render.min.js" integrity="sha384-aGfk5kvhIq5x1x5YdvCp4upKZYnA8ckafviDpmWEKp4afOZEqOli7gqSnh8I6enH" crossorigin="anonymous"></script>
```
ylemkimon: If we decide to maintain the extension at this repo, we can do
```html
<script src="/contrib/auto-render.js" type="text/javascript"></script>
```
and access the page in `http://localhost:7936/contrib/mhchem/index.html` with `npm start`.

ronkok:
Now that I have learned how to use `katex.__defineMacro` to write macros in a external file, I can see other possibilities. For instance this method could easily emulate the `texvc` package, which defines some abbreviated function names used in mediawiki. Large parts, perhaps all, of the `physics` package could be emulated the same way.

So it would be good to see a decision made on how to coordinate packages like this one with KaTeX.

kevinbarabash:
> If no one has any preference, I think we can put them in this repo (KaTeX main repo) for a while.

I think that's probably the easiest path forward.  We can always change that in the future.

danielhollas:
Hi There. Can I ask whether you have some timeline for this feature? I heard that Khan Academy is planning to work on the chemistry content in the near future and I am sure this would come in handy. Moreover, it would also help i18n teams, because current chemistry content uses `\text` for chemical elements, which break the Translation Assistant.

contrib/mhchem/index.html
kevinbarabash: I think that make sense as this `index.html` is more about testing.  It would be nice to have something within `webpage` to demo this there.

I love that this contrib modules works with the auto-render module.

contrib/mhchem/README.md
```diff
@@ -0,0 +1,17 @@
+# mhchem extension
+
+This extension adds to KaTeX the `\ce` and `\pu` functions from the [mhchem](https://mhchem.github.io/MathJax-mhchem/) package.
+
+### Usage
+
+This extension isn't part of core KaTeX, so the script should be separately included. Write the following line into the page�s `<head>`. Place it *after* the line that points to `katex.js`.
+
+```html
+<script src="https://cdn.jsdelivr.net/npm/katex@0.10.0-beta/dist/contrib/mhchem.min.js"></script>
```
kevinbarabash: This will need to be updated and we'll need to do a publish for this to work.  Also, the webpack config files need to be updated to compile this file.

kevinbarabash:
I also think this could use some screenshot tests.  That's something that could be done in a separate PR as well.  

Here are some examples of the output (there's many more examples in the index.html):
screen shot 2018-11-03 at 3 34 00 pm

screen shot 2018-11-03 at 3 34 07 pm

screen shot 2018-11-03 at 3 35 05 pm

Most of the output looks very nice.  There seems to be a minor issue with how some of the bonds are rendering.

screen shot 2018-11-03 at 3 34 45 pm

kevinbarabash:
It's a little disheartening to see code coverage go down by 11%.  Could you exclude the mhchem folder from the coverage report?  We can add it back in once we have some tests for it.  I'll have a look at the rest of the changes later this week.

ronkok:
> Could you exclude the mhchem folder from the coverage report?

Sure. BTW, I've got more work to do. So far, the only comment I've addressed is the bond rendering.  It'll be a couple of days yet before this is ready for review.

ronkok:
Well, I've copied the mhchem Manual page and edited it slightly so it can reside in the KaTeX contrib folder. So the documentation is in better shape.

I've also edited the `webpack.common` file, but please be aware that I have zero experience with webpack. My work there needs careful review.

I've done what I can here. I hope its okay.

contrib/mhchem/mhchem-manual.html
```diff
@@ -0,0 +1,279 @@
...
+:::: ^0_-1n-
+It might be ambiguous whether a superscript belongs to the left or right element. There is automatic detection (digits only = mass number = belongs to right side), but to make sure you can type `{}` as a separator.
+:::: H{}^3HO
+:::: H^3HO
+
+## Reaction Arrows
+:::: A -> B
+:::: A <- B
+:::: A <-> B % not to be used according to IUPAC and ACS
+:::: A <--> B
```
ylemkimon: `rightleftarrows` seems little bit off on Chrome:
image

contrib/mhchem/mhchem-manual.html
```diff
@@ -0,0 +1,279 @@
...
+If you need more control than is offered here, take a look at the siunitx extension.
+
+## Compatibility with __LaTeX/mhchem__
+Most of these examples work identically for KaTeX/mhchem and LaTeX/mhchem. Exceptions are indicated. Also, some edge-cases may render differently. I will try to minimize the differences in the future. But even then, KaTeX/mhchem will always be more tolerant of sloppy input (e.g. a missing space) than LaTeX/mhchem.
+
+## Contact and Support
+If you have a question and cannot find an answer---neither here nor with a web search---then
+- mail me (:myaddress:) or
+- create an GitHub issue (https://github.com/mhchem/MathJax-mhchem/issues) or
+- ask at StackExchange (http://meta.chemistry.stackexchange.com/) with the mhchem tag.
```
ylemkimon: I think we should include KaTeX issue tracker.

contrib/mhchem/mhchem.js
```diff
@@ -0,0 +1,1687 @@
+﻿/* eslint-disable */
```
ylemkimon: Could you include the diff (patch) file between MathJax mhchem and this?

ronkok: @ylemkimon Where would you like to see the diff file? In the `contrib` folder? Here in this string of comments?

contrib/mhchem/mhchem.js
```diff
@@ -0,0 +1,1687 @@
...
+});
+
+//  Needed for \bond for the ~ forms
+//  Raise by 2.56mu, not 2mu. We're raising a hyphen-minus, U+002D, not 
+//  a mathematical minus, U+2212. So we need that extra 0.56.
+katex.__defineMacro("\\tripledash", "{\\vphantom{-}\\raisebox{2.56mu}{$\\mkern2mu"
++ "\\tiny\\text{-}\\mkern1mu\\text{-}\\mkern1mu\\text{-}\\mkern2mu$}}");
+
+
+var mhchem = (function () {
```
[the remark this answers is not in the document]
ronkok: @ylemkimon My purpose here is to prevent functions like `mhchemParser()` or `texify()` from becoming available globally. The current arrangement does this successfully. It there any compelling reason to change it?

contrib/mhchem/mhchem.js
```diff
@@ -0,0 +1,1687 @@
...
+ *  limitations under the License.
+ */
+
+//
+// Coding Style
+//   - use '' for identifiers that can by minified/uglified
+//   - use "" for strings that need to stay untouched
+
+// version: "3.3.0" for KaTeX
+
```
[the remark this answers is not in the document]
ronkok: What does this accomplish?

kevinbarabash:
@ylemkimon any idea as to why `circleci` is complaining about the `$CIRCLE_PROJECT_USERNAME != "Khan"` in the [test](https://circleci.com/gh/KaTeX/KaTeX/5727) job?

contrib/mhchem/mhchem.js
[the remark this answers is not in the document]
ronkok: Please be more specific. What should this line of code look like?

contrib/mhchem/mhchem.js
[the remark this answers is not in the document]
kevinbarabash: @ylemkimon how with this work if people are adding script tags to the page for `katex` and `mhchem` extension?

contrib/mhchem/mhchem.js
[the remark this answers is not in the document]
kevinbarabash: I think you'll want to get rid of this line and at the bottom of the file replace
```
    return {
     expand: chemParse,
   };
```
with
```
    module.exports = {
        expand: chemParse,
    };
```
and drop the trailing:
```
})();
```
I'm curious about why `chemParse` was being exported.  Is it for debugging purposes?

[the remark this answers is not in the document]
ronkok: > our other code doesn't do this.

Our other code is bundled by webpack into a single file. This file remains separate and is called from a reference in an HTML <head>

> and drop the trailing `})();`

Then functions `mhchemParser()`,  `texify()`, `assertNever()`, and `assertString()` become global. Why would we want to do that?

contrib/mhchem/mhchem-manual.html
ronkok: I've deleted this file entirely. I don't want to maintain future versions of it. Instead, I've added a link to the original mhchem Manual. Hopefully, @mhchem will merge a [PR](https://github.com/mhchem/MathJax-mhchem/pull/8) in which I've written KaTeX into the file.

contrib/mhchem/mhchem-manual.html
ronkok: Yep. There was a problem with the SVG. Fixed now.

contrib/mhchem/mhchem.js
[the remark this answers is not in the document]
ronkok: As usual @ylemkimon  is correct. It also turns out that `mhchem.js` does not need a `module.exports = { expand: chemParse, };` statement. `chemParse` is called locally. It's really the macro callbacks that are being exported.

I've modified the file accordingly.

contrib/mhchem/mhchem.js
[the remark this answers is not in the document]
ylemkimon: @ronkok It'd allow using this extension with module loaders, i.e., on Node(CJS) or AMD/ESD.

@kevinbarabash Webpack will try to detect which module loader is used and if there is no module loader is detected then `katex` from global variables is used.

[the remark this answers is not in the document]
ronkok: Done.

kevinbarabash:
@ronkok thanks for all of the changes.  I'm excited about support for mhchem and even more excited that you were able to do it as an extension.  Even though mhchem is a fairly standard LaTeX package I think this PR will serve as a good example how people might extend KaTeX is less standard ways via extensions.

ronkok:
@kevinbarabash You are very welcome.

This, I think, is a good pause point. I have submitted PR's that supported 140 TeX control words, supported 600 Unicode characters, fixed 9 rendering bugs (not counting the ones I created), and added working examples to the documentation. I think I've done most of what I came here to do. Let me know if you find bugs in the code I submitted. Beyond that, my involvement henceforward will be much less that it has been the past 18 months.

To you, to @edemaine  and @ylemkimon, I've appreciated the support and encouragement. You have a great library going here.

ylemkimon:
@ronkok Thank you for your great contributions! It was a pleasure to work with you 😄

ronkok:
@kevinbarabash 1½ years ago, you said you would like to hear if I used KaTeX in any future projects. Well, now I have. I put up https://hurmet.app/ a few days ago and I'm quite happy with it.

kevinbarabash:
@ronkok neat!  It reminds me of juypter notebooks. I like how it allows you to computations on taels in a very succinct way and over course it's nice to see nicely typeset math. 🙂

Issue #681, opened 2017-04-26
Feature request: \rotatebox


meygerjos:
I would like to make symbols upside down.  This is done in LaTeX, using the `graphicsx` package, with `\rotatebox[origin=c]{180}{\heartsuit}` (for an upsidedown `\heartsuit`).  It would be great if KaTeX could implement `\rotatebox`, or even just the special case `\rotatedbox[origin=c]{180}`!

[The suggestion that I make this feature request](https://gitter.im/Khan/KaTeX?at=590097d38bb56c2d11e4d027)

kevinbarabash:
I think this will be pretty easy to get working right for single glyphs or a simple horizontal run.  I worry about the center be wrong for larger layouts.

edemaine:
This is related to #587 (horizontal metrics). To support `\rotatebox` in general, we'd need an understanding of the size of a box.  I still think this is possible, once we have horizontal metric data from fonts (#632), though it's going to be an effort to accumulate the box size.  That said, we may need these box sizes for many things (`\multicolumn` in `tabular` is on my mind) so it'd be worth trying, once #632 is in.

At the HTML level, this should be implementable via [CSS3's rotate transform](https://developer.mozilla.org/en-US/docs/Web/CSS/transform-function/rotate).  The tricky part is to get the `transform-origin` correct, which is where we need the box size.

kevinbarabash:
If we apply the rotate transform to the <div> containing the box that we want to rotate it should work correctly for some cases because the default transform origin is 50%, 50%.  That being said, in looking at some of our layouts in the DOM inspector, sometimes there are elements that are partially outside the bounds of the visible contents of the layout which may or may not cause issues with where the origin actually appears.

edemaine:
@kevinbarabash Hmm, definitely sounds like it's worth an experiment.  If it works in many/useful cases, probably worth including an initial version.

kevinbarabash:
<img width="809" alt="screen shot 2017-04-30 at 11 03 54 am" src="https://cloud.githubusercontent.com/assets/1044413/25565453/c258d302-2d94-11e7-95d7-7daa1c5669b6.png">

I removed `overflow: auto` from the container that KaTeX was rendering in and added `display: inline-block; transform: rotate(310deg);`

meygerjos:
Here is a `javascript` workaround until this is implemented in KaTeX.  Posting because it might be useful to other people who want upside-down symbols and come across this thread.

```
[...]

PR #1437, opened 2018-06-17
Add \class and \cssId on non-strict mode

ylemkimon:
* Added `string` group type, which passes raw string argument


ronkok:
As @kevinbarabash once [reminded me](https://github.com/Khan/KaTeX/issues/681#issuecomment-329038770), some people may want to protect their web site from users adding classes. Maybe this (terrific) code should go behind its own rendering option `omitClass`, default = true.

src/functions/html.js
```diff
@@ -0,0 +1,53 @@
...
+import defineFunction, {ordargument} from "../defineFunction";
+import buildCommon from "../buildCommon";
+import ParseNode, {assertNodeType} from "../ParseNode";
+
+import * as html from "../buildHTML";
+import * as mml from "../buildMathML";
+
+defineFunction({
+    type: "html",
+    names: ["\\class", "\\cssId"],
```
rrandallcainc: While I know mathjax calls this "cssId", I think it would be nice to give it a more generic name (i.e. "id"). My only reasoning is, if multiple renderers end up being supported, css may have no meaning (i.e. in a canvas).

src/functions/html.js
```diff
@@ -0,0 +1,53 @@
...
+    props: {
+        numArgs: 2,
+        argTypes: ["string", "original"],
```
rrandallcainc: I'm assuming we want this allowed in text?

ylemkimon:
@ronkok I'm not sure how they can be abused with `class`es or `id`s, as styles are predefined.

src/functions/html.js
ylemkimon: @rrandallcainc Set `allowedInText` `true`.

ronkok:
> as styles are predefined.

Styles are predefined, that's true. But a malicious user, writing in an input box like the one I am writing in now, could mis-apply existing CSS style rules to places where they should not be. An administrator of a site like this one might want to restrict that. 

@ylemkimon I think this  PR is terrific, and I want to see it succeed. I only write to suggest a way that everybody can be happy with it.

rrandallcainc:
Curious how "reserved" classes/id's are handled. For example, if someone tried `\class{strut}{blah}` would it apply the strut class unintentionally? 

Also, do you plan on adding tests for this?

src/functions/html.js
kevinbarabash: @rrandallcainc were you thinking that `id` might be useful for hit testing in canvas based renderer.

kevinbarabash:
My guess is that `class` and `cssId` (or `id`) are commands that would be used by site/app developers and not something that really make sense for user submitted content so providing a way to disable these commands should be sufficient.

src/Parser.js
```diff
@@ -789,26 +758,6 @@ export default class Parser {
         return newArgument(new ParseNode("color-token", match[0], this.mode), res);
     }
 
-    /**
-     * Parses a url string.
-     */
-    parseUrlGroup(optional: boolean): ?ParsedArg {
```
kevinbarabash: This change conflicts with https://github.com/Khan/KaTeX/pull/1440.

src/ParseNode.js
```diff
@@ -256,6 +257,12 @@ export type ParseNodeTypes = {
         href: string,
         body: AnyParseNode[],
     |},
+    "html": {|
+        type: "html",
+        class?: string,
+        cssId?: string,
```
kevinbarabash: Maybe we could change this be:
```
type: "html"
attribute: "class" | "id",
value: string,
body: AnyParseNode[],
```

src/ParseNode.js
```diff
@@ -146,6 +147,10 @@ export type ParseNodeTypes = {
         type: "size",
         value: Measurement,
     |},
+    "string": {|
+        type: "string",
+        value: string,
+    |},
```
kevinbarabash: Is this only used to replace `url`?

src/functions/html.js
```diff
@@ -0,0 +1,53 @@
...
+        }
+        return span;
+    },
+    mathmlBuilder: (group, options) => {
+        const math = mml.buildExpressionRow(group.value.body, options);
+        if (group.value.class) {
+            math.setAttribute('class', group.value.class);
+        }
+        if (group.value.cssId) {
+            math.setAttribute('id', group.value.cssId);
```
kevinbarabash: I've found that browsers tend to get confused about things when there are multiple elements with the same `id`.

src/functions/html.js
ry-randall: @kevinbarabash Hit testing could be one use case. I was more thinking along the lines of being able to manipulate specific parts of the LaTeX separately. For example, let's say I had `\frac{\id{numerator}{a}}{\id{denominator}{b}}` and I wanted to switch the two. I see as id/class as general ways of targeting specific areas of the latex.

src/functions/html.js
ylemkimon: Does it apply to the `MathML` elements, too?

ronkok:
@ylemkimon I don't really have a strong personal opinion on this issue. I spoke up earlier to make you aware of a comment that @kevinbarabash had once made to me. Anything he likes on this issue will be okay with me, too.

edemaine:
Hmm, related to #1436 then I guess.  Is there a nice uniform solution?  It'd be cool to be able to specify `custom-*` or something as allowed classes, or explicitly list allowed classes, though the "safe" specifier might also be sufficient in most cases.

edemaine:
P.S. I do worry about defining an `id` that matches another `id` on the page, thereby hijacking some things tied to the page.  It's not like the KaTeX object is going to *do* anything with that replacement (until we support `\javascriptToRunWhenClicked{...}`), but it could easy break existing code that uses `document.getElementById`.  So some kind of limitation is needed.

kevinbarabash:
> So some kind of limitation is needed.

I think the limitation should be at the command level.  We should disable `\class` and `\id` by default and allow users to opt-in by enabling "unsafe" mode.  This should be relatively simple to do and allows users to move to the new version of KaTeX without opening a major attack vector.  Users whose sites don't accept user input and who need `\class` and/or `\id` can turn of "unsafe" mode.

ronkok:
> I think the limitation should be at the command level.

What if a web site administrator wants protection from a malicious user? Would the limitation be better if set as a KaTeX rendering option?

ylemkimon:
I think we can go with `RegExp | boolean`, with default `false`. I'm afraid this will clutter settings, with a feature only few people will use. I think we should categorize settings and/or have a naming system.

kevinbarabash:
@ylemkimon the `boolean` is whether or not `class` and `id` are allowed and the `RegExp` is what values they all?

kevinbarabash:
Agree that a naming strategy for options/settings is a good idea.

ry-randall:
@ylemkimon I took a look @ this, and it looks like this PR removes the spacing when it wraps around operators (i.e. `\class{part-1}{\frac{1}{2} + \frac{3}{2}} = \class{part-1}{2}`). 

I'm thinking it wouldn't hurt to add in screenshot tests as part of this PR as well.

ry-randall:
In regards to the broken spacing, I think you'll find this PR helpful: https://github.com/Khan/KaTeX/pull/1388

src/functions/html.js
ry-randall: Assuming you'll want to apply allowInText: true also to html, no?

kevinbarabash:
For this to move forward it sounds like we need:
- test cases involving `\class{part-1}{\frac{1}{2} + \frac{3}{2}} = 2`
- changes to the code that address the spacing issues and make those test cases pass

ylemkimon:
Blocked on #1706.

dvergeylen:
Hi @ylemkimon ,

As #1706 and #1711 are both merged by now, do you see anything blocking this?

Is #1771 blocking this or could a (disabled by default) option (e.g: `{allowClass: true, allowId: true}` ) be reasonable security setting?

I am currently needing this feature for a WYSIWYG editor, would be glad to give a hand!

edemaine:
@dvergeylen Yes, this is blocked on #1771, or more precisely, #1794. We don't want to add a bunch of options and then switch to a different mechanic. Your feedback on #1794 would be appreciated!

steveluc:
While we are waiting for #1794, is there a workaround to add styling to an element that corresponds to a substring of tex input?  Id is overkill for this and so is class.  For example, there could be something like \localId{id1}{x^2} which associates id1 with an element implementing the x^2 substring.  We could have an option passed into render that supplies a map or object to be populated with these ids.  Then whatever permission the page has to style elements, it can apply to these elements.  Thanks.

kevinbarabash:
I think this is still block, but by #1794 now.

kevinbarabash:
This is no longer blocked.

kevinbarabash:
Closing in preference of #2082.

PR #1440, opened 2018-06-19
Allow only allowed protocols in \href

ylemkimon:
* Added `allowedProtocols` settings


ylemkimon:
@edemaine Thank you for the review! I've added boolean argument to allow or disallow all protocols. I'm not sure about the function.

kevinbarabash:
One alternative solution might be to allow `[ "*" ]` to represent all protocols.  This is going to get more interesting if we ever support embedding images.

edemaine:
@kevinbarabash That might be considered more standard. I could go either way. @ylemkimon Thoughts?

ylemkimon:
I think `["*"]` is more consistent. Changed to `["*"]`.

edemaine:
Looks great!

ylemkimon:
I wonder this qualifies as a security vulnerability.

edemaine:
Maybe, given that `javascript:` was allowed before?  How do we indicate this?

kevinbarabash:
We can add a `security` section in the release notes of the next release.  This is technically a change in behaviour so we should probably go to 0.11.0.

Issue #1439, opened 2018-06-18
Accessing the style property on domTree throws an error


ry-randall:
I'm working on integrating the latest KaTeX version into my canvas rendering implementation.

I noticed that when trying to access the style property on a `domTree.documentFragment` node, it throws an error. 

Unfortunately, I have no way of distinguishing between a `documentFragment` node, and a `span` from the outside world (i.e. I cannot do `node instanceof documentFragment` since it's not exposed via `katex.js`

I'm wondering why it's necessary to have `documentFragment` throw an error as opposed to returning something like "null". If it is indeed necessary, can we expose `domTree` or `documentFragment` via `katex.js`?

edemaine:
I was surprised to see this code:
```js
    get style(): CssStyle {
        throw new Error('DocumentFragment does not support style.');
    }

    set style(_: CssStyle) {
        throw new Error('DocumentFragment does not support style.');
    }
```

I guess this is necessary for typing purposes, because the `HtmlDomNode` has `classes: string[]`.  I can see `set` raising an error, but I like the idea of `get` returning `[]`. Any objections?  The one scary thing might be someone doing `myDocumentFragment.classes.push('something')`, which will then act as a no-op...

@rrandallcainc For now, you could wrap `x.classes` in a `try` block.  Changing this behavior will actually make it harder to detect the difference between `documentFragment`s and other types. 😄 We could also easily export the DOM types -- that seems quite natural.

rrandallcainc:
Okay, great. I'll open a PR to export later today!

PR #1446, opened 2018-06-22
Expose dom tree

ry-randall:
A simple solution for type checking dom nodes. This is a followup to the discussion found here: https://github.com/Khan/KaTeX/issues/1439

### DIFF
--- a/katex.js
+++ b/katex.js
@@ -190,7 +190,7 @@ export default {
      */
     __defineMacro: defineMacro,
     /**
-     * Exposes the dom tree, which can be useful for type checking against nodes.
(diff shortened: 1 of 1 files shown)


katex.js
```diff
@@ -189,4 +189,12 @@ export default {
      * adds a new macro to builtin macro list
      */
     __defineMacro: defineMacro,
+    /**
+     * Exposes the dom tree, which can be useful for type checking against nodes.
```
edemaine: Maybe this should be rewritten from "dom tree" to something like "dom tree node types"?

PR #1447, opened 2018-06-23
lexer: Remove `match-at` dependency, use RegExp

ylemkimon:
Closes #1443.


package-lock.json
```diff
@@ -4154,9 +4154,9 @@
       "dev": true
     },
     "flow-bin": {
-      "version": "0.74.0",
-      "resolved": "https://registry.npmjs.org/flow-bin/-/flow-bin-0.74.0.tgz",
-      "integrity": "sha512-tIN9J5qg71S4UbofCu80tve8a+p7Hj7ytwUtu79cLg9KJVVTNnVVJXKgCghVzaZT1Rvl9SMHVPlDs9uYhPHEGQ==",
+      "version": "0.75.0",
+      "resolved": "https://registry.npmjs.org/flow-bin/-/flow-bin-0.75.0.tgz",
+      "integrity": "sha1-uW0e6Z07RGoyJr5mtAEyJM6d8mA=",
```
edemaine: I believe the latest NPM uses `sha512` hashes exclusively. Or does it normally use a mix? I've found I had to delete `node_modules` and clean the NPM cache, then `npm install`, to fix this.

src/Lexer.js
```diff
@@ -67,33 +62,30 @@ const commentRegex = new RegExp(`^${commentRegexString}`);
...
      */
     lex(): Token {
         const input = this.input;
-        const pos = this.pos;
+        const pos = this.tokenRegex.lastIndex;
         if (pos === input.length) {
             return new Token("EOF", new SourceLocation(this, pos, pos));
         }
-        const match = matchAt(tokenRegex, input, pos);
+        const match = this.tokenRegex.exec(input);
```
edemaine: I can think of one important difference here: if there's something that doesn't match the regex, `exec` will (I think) skip over it, while `matchAt` will fail.  Thoughts on how to fix this?

src/Lexer.js
```diff
@@ -67,33 +62,30 @@ const commentRegex = new RegExp(`^${commentRegexString}`);
...
 export default class Lexer implements LexerInterface {
     input: string;
-    pos: number;
+    tokenRegex: RegExp;
 
     constructor(input: string) {
         // Separate accents from characters
         this.input = input;
-        this.pos = 0;
+        this.tokenRegex = new RegExp(tokenRegexString, 'g');
```
edemaine: It's cool that this works even when `feed`ing another string, thanks to the semantics of `exec`.

package-lock.json
ylemkimon: @edemaine There are issues of `npm` mixing integrity hashes, https://github.com/npm/npm/issues/16938 and https://github.com/npm/npm/issues/17749, closed but people are still reporting problems.

src/Lexer.js
ylemkimon: @edemaine `matchAt` and `exec` both return `null` when match fails, and a `ParseError` will be thrown below.

kevinbarabash: If there's a match that occurs after part of the string which isn't matched by the regex, `exec` will happily jump to the next match and return that.  We could keep track of the current position and then verify that match that `exec` occurs at that position.

ylemkimon: @kevinbarabash Ah, I didn't know, hence the name match _At_. Added an assertion to check match index matches the last index. Also updated a test case to test regression.

package-lock.json
edemaine: Shouldn't we try to get sha512 hashes in there?  Again, I've found resetting NPM as listed above (thanks to these various issues) has fixed the problem.

src/Lexer.js
```diff
@@ -67,33 +62,30 @@ const commentRegex = new RegExp(`^${commentRegexString}`);
...
         const input = this.input;
-        const pos = this.pos;
+        const pos = this.tokenRegex.lastIndex;
         if (pos === input.length) {
             return new Token("EOF", new SourceLocation(this, pos, pos));
         }
-        const match = matchAt(tokenRegex, input, pos);
-        if (match === null) {
+        const match = this.tokenRegex.exec(input);
+        if (match === null || match.index !== pos) {
```
edemaine: Nice change.

src/Lexer.js
edemaine: Oops, I was wrong about this.  I just tried the following in Chrome:

```js
s='aaaaaaa'
re.exec(s)  // ["a", index: 0, input: "aaaaaaa", groups: undefined]
re.exec(s)  // ["a", index: 1, input: "aaaaaaa", groups: undefined]
re.exec(s)  // ["a", index: 2, input: "aaaaaaa", groups: undefined]
t='aaaa'
re.exec(t)  // ["a", index: 3, input: "aaaa", groups: undefined]
```

It seems we should set `this.tokenRegex.lastIndex = 0` in `feed`.

ylemkimon: @edemaine Is it possible to feed new string to the lexer? Currently, new `RegExp` instance is created when instantiating lexer.

kevinbarabash: I don't believe so.

test/errors-spec.js
```diff
@@ -285,9 +285,9 @@ describe("Lexer:", function() {
 
     describe("#_innerLex", function() {
         it("rejects lone surrogate char", function() {
-            expect("\udcba").toFailWithParseError(
+            expect("\udcba ").toFailWithParseError(
                    "Unexpected character: '\udcba' at position 1:" +
-                    " \udcba\u0332");
+                    " \udcba\u0332 ");
```
kevinbarabash: Is the addition o the space necessary in this test?

src/Lexer.js
edemaine: Oops, indeed, all is well. I thought `Lexer` had the `feed` method but it's `MacroExpander`, which creates a new `Lexer` each time.

edemaine:
I'm curious whether this speeds up KaTeX at all. Probably too marginal to tell.

PR #1453, opened 2018-06-27
Support \brace and \brack

ronkok:
This PR supports functions `\brace` and `\brack`, useful for [Stirling numbers](https://www.johndcook.com/blog/2018/06/07/stirling-numbers/). These functions originally come from the *TeXbook*, page 360.

Here's how they look:
brace


edemaine:
Cool, I didn't know about these features. Are they in base LaTeX too?

ronkok:
Yep. Per *The LaTeX Source 2ε Sources* PDF, 2016, page 259.

PR #1454, opened 2018-06-27
\char character escaping and nicer MathML via \html@mathml

### LINKED ISSUE #1452: Better MathML support in macros
We've generally moved toward adding features via macros when possible.  This is great because it's easier to add features, and is closer to how LaTeX does everything, so easier to guarantee compatibility.  But it often leads to messier MathML output, essentially because MathML is a bit more semantic than visual.  Two examples are #1400 and #974, though probably most macros should be reviewed on this point.

Here's a proposal for how to fix these, while still getting the macro benefits: add a **function** called `\htmlmathml` (or some other name) that takes two arguments.  The HTML builder for this node type calls the HTML builder on the first argument, while the MathML builder for this node type calls the MathML builder on the second argument.  So both arguments are LaTeX, and it gives you the ability to write the MathML "equivalent" of a fancy macro.  Some examples:

* `\notin` could be redefined to `\htmlmathml{...current definition...}{\mathop{∉}}` (for #1400)
* `\LaTeX` could be redefined to `\htmlmathml{...current definition...}{LaTeX}` (for #974)
* `\pmb` could be defined to `\htmlmathml{...how LaTeX does pmb...}{\mathbf{#1}}` (for #1418)

Originally I thought about how to do this via a macro, but I don't think that's really possible at that point, because macros need to have a single expansion.  Resolving this at function time seems like the right way.  Of course, it's still limited to producing MathML that can be produced by other existing functions, but I don't think we really want raw MathML (or any *ML) as arguments to LaTeX functions.

edemaine:
I implemented `\html@mathml`, which fixes #1452.  (I added `@` to the name to suggest it's for internal use.  Happy to hear other suggestions for a name.)

Unfortunately, it's harder to use than I anticipated in #1452, and some of the macro definitions contained herein are bad, because they lead to infinite loops.  For example, I wanted `\notin` to expand to the equivalent of `\mathbin{∉}` in MathML.  But `∉` expands to `\notin`.

Maybe we could support LaTeX's ``` `∉``` notation for literal characters, which prevents them from expanding?

`\TeX`, `\LaTeX`, `\KaTeX` work well.  `\copyright` works reasonably well, because `©` isn't defined (yet), but it generates a warning about using a character not in our fonts... which it ideally shouldn't when being used in a MathML context like this.


kevinbarabash:
> I added @ to the name to suggest it's for internal use. Happy to hear other suggestions for a name.

Is there any way to restrict a macro for internal use only?  I was thinking that can't disable this strict mode b/c now of the commands that exist in strict mode use this macro.  In practice though, this macro shouldn't have an impact on people's use of strict mode.  One benefit of restricting use of internal macros for internal use only is that it guarantees we can change there name without having to bump the version.

As for the name, I don't have any good suggestions.  I find that many commands have cryptic names.  `math@mathml` should be fine as long as it's documented.

kevinbarabash:
@edemaine did you forget to include a file?

edemaine:
@kevinbarabash Oops, indeed!  Sorry about that.

I don't see any particular reason to restrict this macro to internal use only.  It'd be fine for regular users to use it too, to fine-tune their MathML.

kevinbarabash:
> Maybe we could support LaTeX's `∉ notation for literal characters, which prevents them from expanding?

That sounds like a good idea.  In general I think the more things we can do to be more like LaTeX the better.  Would people be able to use this notation or is it an internal part of LaTeX?

kevinbarabash:
Reading about `\catcode` in the TeXBook.  It seems rather limited and only makes sense when your font is limited to 128 or 256 glyphs.  It would be nice to avoid level of fidelity if possible.

edemaine:
I think this will work!  I tested the following LaTeX:

```latex
% Make the letter k into an active character (definable as a macro)
\catcode`\k=13
% Define its expansion
\defk{hi}
% Now use it in three contexts
$k =
\string k
\char`k
$
```

The result is that `\string k` and ```\char`k``` produce the literal symbol `k`, while the first unprotected instance expands to `hi`.  So indeed, `\string` and ```\char` ``` can be used to escape symbols that are defined as macros.  This gives me a direction to pursue on this PR: implement `\string` and/or `\char`, and use them to avoid infinite recursion on the MathML side.

edemaine:
I implemented `\char` for producing literal characters, which resolved the infinite-loop issue.  Ready for review.

src/functions/char.js
```diff
@@ -0,0 +1,37 @@
...
+            throw new ParseError(
+                `\\@char has invalid argument type ${arg.type}`);
+        }
+        const group = arg.value;
+        let number = "";
+        for (let i = 0; i < group.length; i++) {
+            if (group[i].type !== "textord") {
+                throw new ParseError(
+                    `\\@char has invalid argument piece ${group[i].type}`);
+            }
```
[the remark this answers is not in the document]
edemaine: Thanks!  Used in two instances.

src/macros.js
```diff
@@ -131,6 +136,59 @@ defineMacro("\\TextOrMath", function(context) {
...
+        } else {
+            number = token.text.charCodeAt(0);
+        }
+    } else {
+        base = 10;
+    }
+    if (base) {
+        // Parse a number in the given base, starting with first `token`.
+        const regex = numberRegex[base];
+        if (!token.text.match(regex)) {
```
ylemkimon: You can check the digit using predefined map(`{0:0,1:1,2:2,3:3,4:4,5:5,6:6,7:7,8:8,9:9,A:10,a:10,B:11,b:11,C:12,c:12,D:13,d:13,E:14,e:14,F:15,f:15}` and checking `map[token.text] < base`. I'm not sure which is more readable.

src/functions/char.js
```diff
@@ -0,0 +1,37 @@
...
+            number += group[i].value;
+        }
+        const code = parseInt(number);
+        if (isNaN(code)) {
+            throw new ParseError(`\\@char has non-numeric argument ${number}`);
+        }
+        return new ParseNode("textord", String.fromCharCode(code), parser.mode);
```
ylemkimon: How does this prevent no font metrics error?

src/macros.js
edemaine: Hmm, maybe I could also just use `parseInt`.  `parseInt('9',8)` and `parseInt('g', 16)` both return `NaN`.  I'm not sure whether there are any other weird things allowed by `parseInt`, but it doesn't seem so.

src/functions/char.js
edemaine: It will still produce a "font metrics error" later if that character isn't in the symbols table.  `\char` wasn't intended to bypass that, but rather to avoid macro expansion on the character.  Indeed, `\char` is currently used just in a MathML context, where a missing character in the symbols table isn't an issue.  Maybe that's what you're asking: this is why we don't get such a warning.

src/macros.js
ylemkimon: @edemaine `parseInt` is probably slower than `regex`.

edemaine: Good point. I'm also worried about slightly different implementations of `parseInt` on different browsers.  I switched to your map approach, with no more call to `parseInt`.

ylemkimon:
Is this ready to merge?

edemaine:
Yes. Thanks for the reviews!

PR #1457, opened 2018-06-28
Support \atopwithdelims and \overwithdelims

ronkok:
This PR adds support for `\atopwithdelims` and `\overwithdelims`, two infix functions with custom delmiters. These are originally from the *TeXbook*, pages 152 or 292. They also appear in  MathJax, so they close the function coverage gap a little.

On the other hand, they're pretty obscure and they do nothing that cannot be done by other means. It's not obvious to me that these functions are worth the code space that must be shipped. If this PR were to be denied, I would not be offended.

The next fraction PR will contain three functions that give the author customized control over the fraction bar thickness, so even if this PR is denied, it was still worth the trouble to refactor `genfrac.js`.

Here's a screenshot, with one of the possible selections for delimiters:
fracdelims

If `\atopwithdelims` is merged, then it ~~might~~ will be possible to define `\brace`, `\brack`, and `\choose` each with a one line macro instead of using the code I turned in yesterday. In fact, that is exactly how LaTeX 2ε does define them. ~~I haven't tested it with the KaTeX macro builder, so I'm 100% sure it would work here yet.~~ Maybe we can circle back to that one later.


edemaine:
While we can't define new general inline macros per se, the following definitions (taken from latex.ltx) should work just fine in KaTeX:
```latex
\def\choose{\atopwithdelims()}
\def\brack{\atopwithdelims[]}
\def\brace{\atopwithdelims\{\}}
```

ronkok:
Yes, I just tested those three macros on this branch, and they do work. I guess I shouldn't have withdrawn my earlier comment.

src/functions/genfrac.js
```diff
@@ -321,3 +321,68 @@ defineFunction({
...
+    names: ["\\\\atopwithdelimsfrac", "\\\\overwithdelimsfrac"],
+    props: {
+        numArgs: 4,
+    },
+    handler: ({parser, funcName}, args) => {
+        const [numer, infixNode, denom] = args;
+
+        // Look into the parse node to get the desired delimiter.
+        // $FlowFixMe
+        const leftDelim = delimFromValue(infixNode.value.leftDelim.value);
```
kevinbarabash: There are a couple of issues with the typing:
- `leftDelim` may not be defined
- if it defined, its type (as defined in ParseNode.js) is `ParseNode` which makes the type of `leftDelim.value` `NodeValue` which could be a lot of things (at least the way things are typed right now).

kevinbarabash: I was able to figure out how to fix the `$FlowFixMe`s.  Here's the updated `handler`:
```
    handler: (context, args) => {
        const [numer, , denom] = args;

        const infixNode = assertNodeType(args[1], "infix");

        // Look into the parse node to get the desired delimiter.
        const leftDelim = infixNode.value.leftDelim
            ? delimFromValue(
                checkDelimiter(infixNode.value.leftDelim, context).value)
            : null;
        const rightDelim = infixNode.value.rightDelim
            ? delimFromValue(
                checkDelimiter(infixNode.value.rightDelim, context).value)
            : null;
        const hasBarLine = context.funcName === "\\\\overwithdelimsfrac";

        return new ParseNode("genfrac", {
            type: "genfrac",
            numer: numer,
            denom: denom,
            continued: false,
            hasBarLine: hasBarLine,
            leftDelim: leftDelim,
            rightDelim: rightDelim,
            size: "auto",
        }, context.parser.mode);
    },
```

kevinbarabash: I've also exported `checkDelimiter` from delimsizing.js and imported it into this file.

ronkok:
@kevinbarabash Thank you for the example of how to apply `assertNodeType`. I'll try to apply that in the future.

ronkok:
@kevinbarabash After reading your comments, I did a little testing in LaTeX and tried these functions with missing delimiters (one or both). It turns out that LaTeX does parse such a thing. So that `{a \overwithdelims ( b}` will render as 

To match that behavior, I would need more than just better typing; I think would need pretty significant changes in `Parser.js`. And I don't want to go there.

I had misgivings about this PR before. Now I have more of them. So here's what I propose to do: I'll close this PR and instead open a different one that adds support for AMS function `\genfrac` and *TeXbook* function `\above`. They both give an author the ability to customize fraction bar thickness. One follows the `\frac` syntax and the other is infix.

If you ask me to reopen this PR, of course I will do so. But my preference would be to wait until some user turns in a issue that requests support for `\atopwithdelims` and/or `\overwithdelims`. I bet that never happens.

PR #1458, opened 2018-06-29
Support \genfrac and \above

ronkok:
This PR adds support for `\genfrac` and `\above`, two functions that give the author control over the thickness of the fraction bar thickness. `\above` is an infix function, originally from the *TeXbook*, pages 143 or 292. `\genfrac` is an AMS function that also gives options for delimiters and style (textstyle, displaystyle, etc.).


src/functions/genfrac.js
```diff
@@ -321,3 +335,137 @@ defineFunction({
...
+        numArgs: 3,
+        argTypes: ["math", "size", "math"],
+    },
+    handler: ({parser, funcName}, args) => {
+        const numer = args[0];
+        const sizeNode = assertNodeType(args[1], "infix");
+        const denom = args[2];
+
+        // $FlowFixMe
+        const barSize = sizeNode.value.sizeNode.value.value;
```
kevinbarabash: The following should work:
```
const infixNode = assertNodeType(args[1], "infix");
const sizeNode = assertNodeType(infixNode.value.sizeNode);
const barSize = sizeNode.value.value;
```
@marcianx does that look right?

src/functions/genfrac.js
```diff
@@ -321,3 +335,137 @@ defineFunction({
...
+        // Find out if we want displaystyle, textstyle, etc.
+        let size = "auto";
+        let styl = checkNodeType(args[3], "ordgroup");
+        if (styl) {
+            if (styl.value.length > 0) {
+                size = stylArray[Number(styl.value[0].value)];
+            }
+        } else {
+            styl = assertNodeType(args[3], "textord");
+            size = stylArray[Number(styl.value)];
```
kevinbarabash: Is this really how TeX does this?   I guess `\genfrac` is pretty low-level but I would've expected some sort of identifier or something.

src/functions/genfrac.js
```diff
@@ -321,3 +335,137 @@ defineFunction({
...
+    return delim;
+};
+
+defineFunction({
+    type: "genfrac",
+    names: ["\\genfrac"],
+    props: {
+        numArgs: 6,
+        greediness: 6,
+        argTypes: ["math", "math", "size", "text", "math", "math"],
```
kevinbarabash: Apparently TeX accepts any math for the first two args, but if it's anything other than a delimiter the results end up being weird.

src/functions/genfrac.js
marcianx: Looks right, except you were missing the "size" in the second line:
```
const sizeNode = assertNodeType(infixNode.value.sizeNode, "size");
```
by any chance?

src/functions/genfrac.js
ronkok: `\genfrac` is an AMS function, so I'll quote from the *User’s Guide for the amsmath Package*, 

>the  fourth  argument is a mathstyle override:  integer values 0–3 select respectively `\displaystyle`, `\textstyle`, `\scriptstyle`, and `\scriptscriptstyle`.

The *Guide* goes on to suggest that user's should use `\genfrac` to define their own macro with a short name, rather than repeatedly use `\genfrac` in full.

src/functions/genfrac.js
ronkok: Done. Thank you for the help.

PR #1459, opened 2018-07-01
Support Unicode corners

ronkok:
In the KaTeX fonts, the `⌜` `⌝` `⌞` `⌟` characters, for the AMS functions `\ulcorner`, `\urcorner`, `\llcorner`, and `\lrcorner` respectively, are located at the Unicode code points for Box Drawings corners. In contrast, `unicode-math` maps those functions to Unicode code points U+231C - U+231F. The names of those code points are Top Left Corner, Top Right Corner, etc, in the Unicode Miscellaneous Technical block. `unicode-math` doesn't map the Box Drawing corners to anything at all.

This PR adds support for U+231C through U+231F.

### DIFF
--- a/src/macros.js
+++ b/src/macros.js
@@ -320,6 +320,10 @@ defineMacro("\u225F", "\\stackrel{\\tiny?}{=}");
 defineMacro("\u27C2", "\\perp");
 defineMacro("\u203C", "\\mathclose{!\\mkern-0.8mu!}");
 defineMacro("\u220C", "\\notni");
+defineMacro("\u231C", "\\ulcorner");
+defineMacro("\u231D", "\\urcorner");
+defineMacro("\u231E", "\\llcorner");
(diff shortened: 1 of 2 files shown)


src/macros.js
```diff
@@ -320,6 +320,10 @@ defineMacro("\u225F", "\\stackrel{\\tiny?}{=}");
 defineMacro("\u27C2", "\\perp");
 defineMacro("\u203C", "\\mathclose{!\\mkern-0.8mu!}");
 defineMacro("\u220C", "\\notni");
+defineMacro("\u231C", "\\ulcorner");
+defineMacro("\u231D", "\\urcorner");
+defineMacro("\u231E", "\\llcorner");
+defineMacro("\u231F", "\\urcorner");
```
ylemkimon: This should be `\lrcorner`.

ronkok:
@ylemkimon Good point. Fixed now. Thanks!

Issue #1079, opened 2018-01-21
\operatorname doesn't support all math functions


kevinbarabash:
```
\operatorname{x\pi\frac{1}{2}}
``` 
is valid LaTeX, but KaTeX throws the following exception:
```
undefined is not an object (evaluating 'n.replace')
```

leungbk:
With the current version of KaTeX, `\operatorname{x\pi\frac{1}{2}}` throws no errors for me besides the "no character metrics" error.

I noticed that `\operatorname{\sum}` doesn't work right now, though it is valid LaTeX. Setting `acceptUnicodeChar = true` for `\sum` in its `defineSymbol()` invocation at line 654 of `symbols.js`, and enforcing `math` mode (solely to accommodate `\sum` in this test case) in the `htmlBuilder` of `operatorname.js` allows `\operatorname{\sum}` to render. What is the reason for not setting `acceptUnicodeChar = true` by default?

ronkok:
> What is the reason for not setting `acceptUnicodeChar = true` [on `\sum`] by default?

@leungbk, When we were implementing Unicode support, I included the big operators by writing them in the `defineFunction` section of file `ops.js`. I felt that this gave someone reading the code a more direct way to see the connection to operator symbol re-sizing, `\limits`, etc. 

It may be that we could do both that and also set the `acceptUnicodeChar` flag in `symbols.js`.

kevinbarabash:
@ronkok #1461 closes this, right?

ronkok:
Yes, I closes all the use cases that I can think of.

PR #1461, opened 2018-07-02
Fix \operatorname

ronkok:
This PR addresses issue #1079 and enables `\operatorname` to support general math functions, not just text. ~~At least, that's true in the HTML part. The MathML part still works under the assumption that the input is all text characters. I'm not sure yet how to overcome that.~~

Part of the changes necessary to accommodate other functions is due to work done by @kevinbarabash  in PR #111. I've taken what he began and extended it. It turns out that the code is now simpler than the original. 

Here's `\operatorname` with a lot of odd stuff in the argument:
op


ronkok:
MathML is updated. I think this PR is now ready for review.

src/functions/operatorname.js
```diff
@@ -42,47 +38,74 @@ defineFunction({
...
+            }
+        }
+
+        if (isAllString) {
+            // Write a single TextNode instead of multiple nested tags.
+            let word = expression.map(node => node.toText()).join("");
+            // Per amsopn package,
+            // change minus to hyphen and \ast to asterisk
             word = word.replace(/\u2212/g, "-");
             word = word.replace(/\u2217/g, "*");
```
ylemkimon: What if `!isAllString`, but has these characters?

ronkok: Good point. I've made a change to pick that up.

ylemkimon: @ronkok If then, isn't `isAllString` unnecessary?

ronkok: `isAllString` still has a purpose. The preferred behavior is to put a string inside a single `<mo>...</mo>` tag. And that's what happens if `isAllString` remains true. When it is false, the code will write  nested tags, such as `<mo><mi>...</mi><mfrac>...</mfrac></mo>`.

ylemkimon: @ronkok I was thinking about these two lines.

ronkok: @ylemkimon You're right, as usual. I'll delete those two lines.

src/functions/operatorname.js
```diff
@@ -42,47 +38,78 @@ defineFunction({
...
-                    letter = letter.replace(/\u2212/, "-");   // minus => hyphen
-                    letter = letter.replace(/\u2217/, "*");
-
-                    // Use math mode for Greek letters
-                    mode = (/[\u0391-\u03D7]/.test(letter) ? "math" : "text");
-                    output.push(buildCommon.mathsym(letter, mode));
-                } else {
-                    output.push(child);
+                    let letter = child.value;
+                    if (/[\u2212\u2217]/.test(letter)) {
```
ylemkimon: Is this check necessary?

ronkok: That check is still necessary but the similar one above has no effect. So I removed the ineffective code.

ronkok:
I had a bad test string. It included "π", which gave a error message because KaTeX fonts do not contain any upright lower case Greek letters. The corrected test contains "Π" instead of  "π".

src/functions/operatorname.js
ylemkimon: @ronkok Doesn't `replace` do nothing if those letters don't exist?

ronkok: > Doesn't `replace` do nothing if those letters don't exist?

I don't think that's the relevant question. Those letters *do* exist.  Say that an author inputs `\operatorname{atan*}`.  During `html.buildExpression()` above, that asterisk, *, U+002A, is converted to Asterisk Operator, ∗, U+2217. Since we're emulating the amsopn package and it reverts that operator back to a text asterisk, we should do the same thing.

ylemkimon: @ronkok I meant, there won't be much difference, but can we just do:

```js
child.value = child.value.replace(/\u2212/, "-")
    .replace(/\u2217/, "*");
```

without `/[\u2212\u2217]/.test(letter)` check?

ronkok: Sure. Will do.

ylemkimon: @ronkok Could you change this one, too?

ronkok: Will do.

ylemkimon:
Thank you!

ronkok:
And thank you for the review.

Issue #1098, opened 2018-01-27
lint against modern JavaScript construct we don't want to polyfill


kevinbarabash:
Some polyfills are expensive, but we still want to maintain compatibility on older browsers so we should lint against those things we don't want to polyfilly, e.g. `.includes()` on strings.

Use https://github.com/kevinbarabash/eslint-plugin-luddite.

ylemkimon:
What do you think about enabling `loose` option on babel preset? (Related to #850)

kevinbarabash:
https://github.com/Khan/KaTeX/pull/1469 added a lint rule for this.

PR #1469, opened 2018-07-10
Remove codes that require expensive polyfill

### LINKED ISSUE #850: enable "loose" mode for transform-es2015-spread babel plugin
The "loose" mode assumes that all iterables are arrays which is a safe assumption to make in our codebase.  This will save use some space when using the spread operator.  See https://babeljs.io/docs/plugins/transform-es2015-spread/.

Unfortunately, changing settings to individual plugins within a babel preset is tricky.  Probably best to use https://github.com/developit/modify-babel-preset.

ylemkimon:
Related: #1098

Some polyfills are expensive in terms of size and performance, due to iterator, generator, or symbol polyfill and/or polyfill for old browsers(`< IE 9`).

- **class inheritance(`extends`)***: 50KB, #1305 
- **`for...of`***: 25KB
- ~**array destructuring***: 20KB~
- **`Object.assign()`**: 10KB, https://github.com/Khan/KaTeX/pull/1048#discussion_r163119802
- **`Object.keys()`**: 9KB, the code can be simpler without it

* **`Object.freeze()`**: 4KB, `>= IE 9` supports `Object.freeze`, but Babel polyfills it. Changed to `Object["freeze"]` to prevent polyfill. https://github.com/Khan/KaTeX/pull/904#discussion_r140668395

- **enabled loose mode**: 50KB, only those in **bold** affect current codes
  - **class methods are enumerable.**
  - **methods are defined on the class prototype with simple assignments instead of being defined.**
  - **all iterables are assumed to be arrays in spread/destructuring**: Fixes #850
  - tagged template literal objects aren't frozen.
  - computed property names use simple assignments instead of being defined.

\* Added eslint rule to disallow it

This PR reduces about 75KB(25KB minified). Tested on `IE 9`.


ronkok:
If array destructuring is expensive only in IE<9, can we stop using that polyfill? We don't support IE<9.

I mention this because I wrote the array destructuring into `stretchy.js` because a review comment suggested that I do so..

ylemkimon:
@ronkok FYI, it'll not be polyfilled in Babel 7 loose mode. So it can be reverted when we upgrade to Babel 7.

src/domTree.js
```diff
@@ -65,26 +176,11 @@ export class HtmlDomContainer<ChildType: VirtualNodeInterface>
...
-        if (options) {
-            if (options.style.isTight()) {
-                this.classes.push("mtight");
-            }
-            const color = options.getColor();
-            if (color) {
-                this.style.color = color;
-            }
-        }
+        initNode.call(this, classes, children, options, style);
```
kevinbarabash: Does this set the prototype correctly?

src/domTree.js
```diff
@@ -21,6 +24,106 @@ const createClass = function(classes: string[]): string {
...
 
+const initNode = function<ChildType>(
+    classes?: string[],
+    children?: ChildType[],
+    options?: Options,
+    style?: CssStyle,
+) {
+    this.classes = classes || [];
+    // $FlowFixMe
+    this.children = children || [];
```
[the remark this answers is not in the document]
kevinbarabash: You might try doing: `children?: ChildType[] = []` in the param list and then this would become `this.children = children`.

.eslintrc
```diff
@@ -35,6 +35,7 @@
         "no-new": 2,
         "no-new-func": 2,
         "no-new-object": 2,
+        "no-restricted-syntax": [2, "ForOfStatement", "ArrayPattern", "ClassDeclaration[superClass]", "ClassExpression[superClass]"],
```
kevinbarabash: Cool!

src/domTree.js
ylemkimon: @kevinbarabash `initNode` doesn't touch prototypes but just initializes properties.

src/domTree.js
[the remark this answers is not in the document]
ylemkimon: @kevinbarabash It still gives following errors:
```
Error --------------------------------------------------------------------------------------------- src/domTree.js:34:21

Cannot assign `children || []` to `this.children` because:
 - `ChildType` [1] is incompatible with `ChildType` [2] in array element.
 - `ChildType` [1] is incompatible with `VirtualNodeInterface` [2] in array element.
 - `ChildType` [1] is incompatible with `HtmlDomNode` [3] in array element.
 - property `classes` is missing in `ChildType` [1] but exists in `HtmlDomNode` [3] in array element.
 - property `depth` is missing in `ChildType` [1] but exists in `HtmlDomNode` [3] in array element.
 - property `hasClass` is missing in `ChildType` [1] but exists in `HtmlDomNode` [3] in array element.
 - property `height` is missing in `ChildType` [1] but exists in `HtmlDomNode` [3] in array element.
 - property `maxFontSize` is missing in `ChildType` [1] but exists in `HtmlDomNode` [3] in array element.
 - property `style` is missing in `ChildType` [1] but exists in `HtmlDomNode` [3] in array element.
 - property `toMarkup` is missing in `ChildType` [1] but exists in `VirtualNodeInterface` [2] in array element.
 - property `toMarkup` is missing in `ChildType` [1] but exists in `VirtualNodeInterface` [3] in array element.
 - property `toNode` is missing in `ChildType` [1] but exists in `VirtualNodeInterface` [2] in array element.
 - property `toNode` is missing in `ChildType` [1] but exists in `VirtualNodeInterface` [3] in array element.
 - property `tryCombine` is missing in `ChildType` [1] but exists in `HtmlDomNode` [3] in array element.
 - `ChildType` [2] is incompatible with `VirtualNodeInterface` [2] in array element.
 - `ChildType` [2] is incompatible with `HtmlDomNode` [3] in array element.

   src/domTree.js:34:21
    34|     this.children = children || [];
                            ^^^^^^^^^^^^^^

References:
   src/domTree.js:29:16
    29|     children?: ChildType[],
                       ^^^^^^^^^ [1]
   src/domTree.js:163:15
   163|     children: ChildType[];
                      ^^^^^^^^^ [2]
   src/domTree.js:217:15
   217|     children: HtmlDomNode[];
                      ^^^^^^^^^^^ [3]
```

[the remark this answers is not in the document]
ylemkimon: @marcianx Could you give any insights? Thank you!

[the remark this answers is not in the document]
marcianx: This earnestly feels broken, and I didn't find a way to cajole the type system to accept it without this override. (I wonder whether typescript would have been a better choice due to greater maturity.) That being said, a simple workaround is to move the `children` assignment into the original constructors:
```
initNode.call(this, classes, options);
```
and remove the type parameter to `initNode` since it's no longer necessary then.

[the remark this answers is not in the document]
ylemkimon: @marcianx That seems nice compromise 😄

src/buildCommon.js
```diff
@@ -246,9 +246,9 @@ const makeOrd = function<NODETYPE: "spacing" | "mathord" | "textord">(
     const fontOrFamily = isFont ? options.font : options.fontFamily;
     if (value.charCodeAt(0) === 0xD835) {
         // surrogate pairs get special treatment
-        const [wideFontName, wideFontClass] = wideCharacterFont(value, mode);
```
kevinbarabash: Babel's output for restructuring seems to be pretty short.  How much of a savings is this?

src/domTree.js
```diff
@@ -25,6 +28,103 @@ const createClass = function(classes: string[]): string {
...
+ */
+const toNode = function(tagName: string): HTMLElement {
+    const node = document.createElement(tagName);
+
+    // Apply the class
+    node.className = createClass(this.classes);
+
+    // Apply inline styles
+    for (const style in this.style) {
+        if (this.style.hasOwnProperty(style)) {
```
kevinbarabash: Maybe we should go with
```
Object.keys(this.style).forEach(
   key => node.style[key] = this.style[key]);
```
It's faster and short.

src/domTree.js
```diff
@@ -25,6 +28,103 @@ const createClass = function(classes: string[]): string {
...
+            // $FlowFixMe Flow doesn't seem to understand span.style's type.
+            node.style[style] = this.style[style];
```
kevinbarabash: Maybe if we type `this.style` as a `CSSStyleDeclaration` it will make `flow` happy.  See https://github.com/facebook/flow/blob/v0.75.0/lib/dom.js#L1418.  That way we only allow valid CSS properties into `this.style`.

src/SourceLocation.js
```diff
@@ -14,7 +14,8 @@ export default class SourceLocation {
         this.lexer = lexer;
         this.start = start;
         this.end = end;
-        Object.freeze(this); // Immutable to allow sharing in range().
+        // $FlowFixMe, do not polyfill
+        Object["freeze"](this); // Immutable to allow sharing in range().
```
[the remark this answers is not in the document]
ylemkimon: @kevinbarabash Yes, it just looks for the propety name. As computed property may depend on runtime, it doesn’t polyfill methods called using computed property.

src/buildCommon.js
ylemkimon: @kevinbarabash The last time I checked, it included Iterator and Symbol polyfill, but now I check it seems to be short. I’ll check again and revert if necessary.

src/buildCommon.js
```diff
@@ -315,7 +315,8 @@ const sizeElementFromChildren = function(
     let depth = 0;
     let maxFontSize = 0;
 
-    for (const child of elem.children) {
+    for (let i = 0; i < elem.children.length; i++) {
```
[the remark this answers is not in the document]
ylemkimon: @kevinbarabash But then, we should override the Babel preset but I’m not sure there is clear way to do this.

.eslintrc
```diff
@@ -70,6 +70,13 @@
         "valid-jsdoc": 0,
         "require-jsdoc": 0
     },
+    "overrides": [{
+        "files": ["katex.js", "src/**/*.js"],
+        "excludedFiles": "unicodeMake.js",
```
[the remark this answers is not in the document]
ylemkimon: @kevinbarabash I think we can have `lib` folder and put configuration files and scripts there.

src/buildCommon.js
ylemkimon: @kevinbarabash It seems something got mixed up at the last time. The loose mode assumes right hand sided to be an array. I reverted relevant commit. Thank you!

src/domTree.js
ylemkimon: @kevinbarabash I agree functional approach is the best. But in that case, it uses `Object.keys()` which this PR aims to avoid. I'll open another PR, changing loops to use map, every, filter, find, findIndex, reduce, some, etc.

src/domTree.js
ylemkimon: @kevinbarabash Changing to `CSSStyleDeclaration` seems to give a whole new errors. I'll try to make another PR and if I can't I'll open a new issue.

src/buildCommon.js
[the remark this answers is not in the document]
kevinbarabash: Yeah, that's definitely one of the shortcomings of Babel's configuration story.

src/domTree.js
kevinbarabash: Hmm... so typing a `{}` as a `CSSStyleDeclaration` won't work.  That's too bad.

PR #1470, opened 2018-07-11
Remove tree cloning before building HTML

ylemkimon:
Related #225.

The only place where tree is mutated was `accent`.

This also removes `JSON.stringify()` polyfill, which is about 2KB. (#1469)

### DIFF
--- a/src/buildHTML.js
+++ b/src/buildHTML.js
@@ -318,10 +318,6 @@ function buildHTMLUnbreakable(children, options) {
  * nodes.
  */
 export default function buildHTML(tree: AnyParseNode[], options: Options): DomSpan {
-    // buildExpression is destructive, so we need to make a clone
-    // of the incoming tree so that it isn't accidentally changed
-    tree = JSON.parse(JSON.stringify(tree));
-
     // Strip off outer tag wrapper for processing below.
     let tag = null;
     if (tree.length === 1 && tree[0].type === "tag") {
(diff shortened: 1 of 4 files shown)


ylemkimon:
@edemaine I scanned through `htmlBuilder`s to find where `ParseNode`(group) was mutated. There was one place where `ParseNode` was temporarily modified, `functions/op.js` added by #1430. Other than that, it seemed `ParseNode`(group) is not touched.

Also, I've added helper code to throw an error if tree was modified during `buildHTML` and run `jest` test suites.

edemaine:
Awesome.  Is it possible / a good idea to add that helper to the Jest test suite?  Or too messy / high overhead?

ylemkimon:
@edemaine As it needs to access intermediate tree before building HTML, I've edited the `buildTree` function directly. I'm not sure of a good way to add this to the test helpers.

edemaine:
@ylemkimon Hmm. If we're calling `parseTree` and `buildTree` separately in the tester, couldn't we duplicate the tree in between and then see if it changed during `buildTree`? Of course, #1336 should really be merged before tweaking any of this.

ylemkimon:
@edemaine Unfortunately, we're calling them together in the test helpers by calling `katex.__renderToDomTree`.

edemaine:
I guess an option would be to have an optional callback on `renderToDomTree` of what to do with the parse tree... but that may not be worth it.  I'm happy enough that you tested it for existing code.

ylemkimon:
@edemaine It'd be nice to have some kinds of hooks system 😄

kevinbarabash:
Something else we might look into to ensure that structures aren't mutated by accident is to use flow to mark objects as read-only.  See https://flow.org/en/docs/types/utilities/#toc-readonly for details.

PR #1471, opened 2018-07-11
Set `documentFragment.style` to empty object

### LINKED ISSUE #1439: Accessing the style property on domTree throws an error

ylemkimon:
Fixes #1439.

The `classes` property of `documentFragment` already returns an empty array. 

Removed `style` getter and setter in `documentFragment`, which raises an `Error`. Separately, this removes class property getter and setter polyfill, which is about 10KB. (#1469)

### DIFF
--- a/src/domTree.js
+++ b/src/domTree.js
@@ -246,6 +246,7 @@ class documentFragment implements HtmlDomNode {
     height: number;
     depth: number;
     maxFontSize: number;
+    style: CssStyle;          // Never used; needed for satisfying interface.
 
     constructor(children?: HtmlDomNode[]) {
         this.children = children || [];
@@ -253,6 +254,7 @@ class documentFragment implements HtmlDomNode {
         this.height = 0;
         this.depth = 0;
         this.maxFontSize = 0;
+        this.style = {};
     }
 
     hasClass(className: string): boolean {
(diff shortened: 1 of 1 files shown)


edemaine:
This seems fine, and simpler to me, but I'll reraise a question I asked in https://github.com/Khan/KaTeX/issues/1439#issuecomment-399452762:
> The one scary thing might be someone doing `myDocumentFragment.classes.push('something')`, which will then act as a no-op...
I think the real issue here is we're abusing the type interface.  But this was true before, too.  Perhaps the proper way would be to have two interfaces, one with style and one without (the former could subclass from the latter).  We could save this change for another PR if you prefer.

ylemkimon:
@edemaine I agree. But, to seperate `documentFragement` from `HtmlDomNode`, it seems that `htmlBuilder`s that build `documentFragment` should be separated. I think this needs refactoring the type system and should be another PR.

Issue #954, opened 2017-10-27
Generate a ES6 module npm package


justinfagnani:
See https://unpkg.com/katex@0.9.0-alpha1/src/MacroExpander.js which contains an `import type...` statement.

This prevents using the JS modules published to npm.

kevinbarabash:
@justinfagnani package.json's `main` field points to `"dist/katex.js"` which should not contain any type annotations.

justinfagnani:
`dist/katex.js` isn't a JavaScript module though, and `/katex.js` is. Since browsers are now supporting modules, I'd like to just load a real importable JS module from unpkg without going through any kind of bundling step.

I guess I was confused by the presence of what looked like modules because many of the files don't have Flow annotations. Can you just run them through the flow compiler and publish the raw output, without bundling with WebPack?

rrandallcainc:
@justinfagnani When you say modules, I'm assuming you're referring to ES6 modules? At the moment, the entry point for Katex.js is of CommonJS format.

This is required for backwards compatibility, and an ES6 entry point would be a breaking change.

That being said, perhaps there could be multiple entry points for katex.js, one of them being perhaps katex.es.js (as an ES6 module)?

Babel can be used to strip out the flow annotations, or this library should work: https://github.com/flowtype/flow-remove-types

That would probably be run on each of the files in the dependency tree.

kevinbarabash:
I've changed the title of this issue.  I'm not sure the best way to do this.  This came up in another issue as well.  It sounded like making an ES6 specific npm package wasn't necessary and that there were some package.json fields that might do the trick.

rrandallcainc:
Please refer to the last comment here: https://github.com/Khan/KaTeX/pull/902

Looks like pckg.esnext/pckg.module are options for the ES6 module's entry point. As far as stripping out the flow-type, I'm guessing that could be added to the build script?

justinfagnani:
Thanks a lot for looking into this!

@rrandallcainc yes, I'm referring to ES6 modules. And I'm not suggesting changing `dist/katex.js`* file, but the ones published in `src/`. Since you probably want to keep those around for sourcemaps you could compile them to a `lib/` directory, unbundled, that just mirrors `src/`. That's what I typically do for my TypeScript projects.

On a little tangent, is there an alternate file extension to use with Flow that doesn't lead people to think the files are standard JS?

*yet! I hope the JS ecosystem can move away from CJS as soon as possible :)

[...]

PR #1475, opened 2018-07-12
Inline version using `babel-plugin-version-inline`

ylemkimon:
Currently, `katex.version` is imported from `package.json`. This increases bundling overhead since `JSON` module doesn't support module concatenation and tree shaking.

Also, it makes impossible to use as ECMAScript modules(#954).


ylemkimon:

kevinbarabash:
A simple but useful plugin.  Nice find!

PR #1477, opened 2018-07-12
Make `unicodeAccents.js` ES6 module

ylemkimon:
* Use `babel-register` in `unicodeMake.js` to require ES6 module.

This reduces bundling overhead since non-ES6 module didn't support module concatenation and tree shaking.


kevinbarabash:
Another option is passing the `--harmony` flag to `node`, but I don't know which versions support ES6 modules with that flag.

ylemkimon:
@kevinbarabash Node 8.5+ supports ES modules with `experimental-modules` flag, but the file extension needs to be `.mjs`.

PR #1478, opened 2018-07-13
Refactor documentFragment and implement both HtmlDomNode and MathDomNode interfaces (stepping stone to port buildMathML to flow)

marcianx:


src/domTree.js
```diff
@@ -211,6 +205,33 @@ class span<ChildType: VirtualNodeInterface> extends HtmlDomContainer<ChildType>
     }
 }
 
+// NOTE: Defining both DomSpan and SvgSpan type directly as span<HtmlDomNode> and
+// span<svgNode> seems to confuse flow when dealing with the type DomSpan|SvgSpan.
+
+/** Span wrapping other DOM nodes. See NOTE above. */
+export class DomSpan extends span<HtmlDomNode> {
```
ylemkimon: The polyfill of class inheritance is expensive. #1469 is pending review, but I think it's best to avoid class inheritance.

src/domTree.js
```diff
@@ -235,21 +256,20 @@ class anchor extends HtmlDomContainer<HtmlDomNode> {
...
-    children: HtmlDomNode[];
-    classes: string[];         // Never used; needed for satisfying interface.
+export class documentFragment extends tree.documentFragment<HtmlDomNode>
+    implements HtmlDomNode {
+    // Never used; needed for satisfying interface.
+    classes: string[];
     height: number;
     depth: number;
     maxFontSize: number;
     style: CssStyle;          // Never used; needed for satisfying interface.
```
ylemkimon: Could you have a look at questions raised at https://github.com/Khan/KaTeX/issues/1439#issuecomment-399452762 and https://github.com/Khan/KaTeX/pull/1471#issuecomment-404193446? I've tried to make `classes` and `style` optional and/or separate `documentFragement` from `HtmlDomNode`, but it seems to be closely intertwined.

src/domTree.js
marcianx: Ok, after the documentFragment refactor per below, I was able to undo this without flow being confused again.

src/domTree.js
marcianx: I've coalesced `documentFragment` into a single class that implements both `HtmlDomNode` and `MathDomNode` interfaces. It required a `$FlowFixMe`.

I also ran into additional cases of this uninformative error that's most certainly a bug in flow:
```
Error: "documentFragment [1] is incompatible with MathNode [2]."
```
I've got 4 instances of it (one was "incompatible with symbolNode") that I had to suppress with `$FlowFixMe`. I'd appreciate any insights, or we could let it be.

src/buildCommon.js
```diff
@@ -233,7 +235,7 @@ const makeOrd = function<NODETYPE: "spacing" | "mathord" | "textord">(
     group: ParseNode<NODETYPE>,
     options: Options,
     type: "mathord" | "textord",
-): domTree.symbolNode | domTree.documentFragment {
+): domTree.symbolNode | HtmlDocumentFragment {
```
ylemkimon: Changing the order, i.e., `HtmlDocumentFragment | domTree.symbolNode` seems to work.

ylemkimon: This seems to fix other three `$FlowFixMe`s in `src/functions/*.js`.

marcianx: Great! Excellent catch! The flow error could definitely do with some improving.
Updated.

marcianx:
BTW, the PR to port `buildMathML` to flow is ready, but that sits on top of this PR, so I'll wait until this is submitted to send that PR.

kevinbarabash:
How does this interact with #1469?

PR #1479, opened 2018-07-13
Build ECMAScript modules

### LINKED ISSUE #954: Generate a ES6 module npm package

ylemkimon:
Fixes #954.

This makes it possible to use KaTeX directly with only minimal transpiling, removing flow typings and `transform-class-properties`.

You can test modules by running `npm run build` and then `npm start` in the root KaTeX directory and then visiting http://localhost:7936/module.html with a web browser which supports modules.

Also I've committed `build` files to `module-test` branch on `ylemkimon/KaTeX`, so it can be loaded from CDN:
```html
<link rel="stylesheet" type="text/css" href="https://cdn.rawgit.com/ylemkimon/KaTeX/module-test/build/katex.min.css">
<script type="module" type="text/javascript">
    import katex from 'https://cdn.rawgit.com/ylemkimon/KaTeX/module-test/build/katex.mjs';
</script>
```

Tested on
* Node with `esm`, with `--experimental-modules` flag
* webpack 4, rollup.js
* Chrome 67, Firefox 61, and Edge 17.


.babelrc
```diff
@@ -1,19 +1,44 @@
...
+            "plugins": [
+                "transform-class-properties",
+                "transform-runtime"
+            ]
+        },
+        "esm": {
+            "plugins": [
+                "transform-class-properties",
+                "./contrib/babel-plugin-esm/babel-plugin-esm.js"
+            ]
```
ylemkimon: Duplications are inevitable due to strange `.babelrc` `env` merging: https://github.com/babel/babel/issues/5276. This can be simplified in Babel 7 using `.babelrc.js`.

contrib/babel-plugin-esm/babel-plugin-esm.js
```diff
@@ -0,0 +1,25 @@
+module.exports = ({types: t}) => ({
```
ylemkimon: I've put this file here to prevent polluting the root.

ry-randall:
As a side note, this _should_ help enable tree-shaking as well for consumers using webpack. Cool!

package.json
```diff
@@ -44,6 +47,7 @@
     "mkdirp": "^0.5.1",
     "pako": "1.0.6",
     "query-string": "^6.1.0",
+    "renamer": "^0.7.3",
```
ylemkimon: These dependencies are required for cross-environment environment variable setting and renaming files. It'd be nice to do without them.

package.json
```diff
@@ -73,10 +77,11 @@
     "prestart": "node check-node-version.js && check-dependencies && node src/unicodeMake.js > src/unicodeSymbols.js",
     "start": "webpack-dev-server --hot --config webpack.dev.js",
     "analyze": "webpack --config webpack.analyze.js",
-    "build": "npm run prestart && rimraf build/* && webpack",
+    "build": "npm run prestart && rimraf build/* && webpack && cross-env BABEL_ENV=esm npm run build:esm",
+    "build:esm": "babel katex.js -o build/katex.mjs && babel src -d build/lib --ignore unicodeMake.js,types.js && babel submodules/katex-fonts/fontMetricsData.js -o build/lib/fontMetricsData.js && renamer -e -f \\.js$ -r .mjs build/lib/**",
     "watch": "npm run build -- --watch",
     "dist": "npm test && npm run build && npm run dist:copy && npm run dist:zip && npm run dist:dist",
-    "dist:copy": "cd build && mkdirp katex && cp -r katex.js katex.min.js katex.css katex.min.css contrib fonts ../README.md katex",
+    "dist:copy": "cd build && mkdirp katex && cp -r katex.js katex.min.js katex.css katex.min.css katex.mjs contrib fonts lib ../README.md katex",
```
ylemkimon: I'm not sure it's better to provide concatenated(bundled) module.

package.json
```diff
@@ -2,7 +2,8 @@
   "name": "katex",
   "version": "0.11.0-pre",
   "description": "Fast math typesetting for the web.",
-  "main": "dist/katex.js",
+  "main": "dist/katex",
+  "module": "dist/katex.mjs",
```
kevinbarabash: Does node use this when doing `import ... from "katex"` when the `--harmony` flag is enabled?

.eslintrc
```diff
@@ -75,5 +76,18 @@
         "node": true,
         "browser": true
     },
+    "overrides": [{
+        "files": ["katex.js", "src/**/*.js"],
```
kevinbarabash: I've never used `"overrides"` before.  TIL :)

.eslintrc
```diff
@@ -75,5 +76,18 @@
...
+        "rules": {
+            "import/no-unresolved": 2,
```
kevinbarabash: Which files wouldn't we want this enabled for?

.eslintrc
```diff
@@ -75,5 +76,18 @@
...
+            "import/extensions": [2, "never"]
+        },
+        "env": {
+            "node": false
```
kevinbarabash: This will warn if we try to use `process` inside of `src/**/*.js`?

.babelrc
```diff
@@ -1,20 +1,45 @@
...
+            ],
+        },
+        "esm": {
             "plugins": [
-                "transform-es2015-modules-commonjs"
+                "transform-class-properties",
+                "version-inline",
+                ["import-rename", {
+                    // source files and font metrics data are moved to lib/
+                    "^\\./src/": "./lib/",
```
kevinbarabash: Sneaky.

.babelrc
```diff
@@ -1,20 +1,45 @@
...
+                    "^\\.\\./submodules/katex-fonts/": "./",
+                    // file extensions are changed to .mjs
+                    // browsers are not able to resolve paths like Node
```
kevinbarabash: It's too bad that `webpack` doesn't support outputting ES6 bundles.  I can't see very many people using `./lib/katex.js` in the browser since that will result in a bunch of additional requests.

eslint-import-resolver.js
```diff
@@ -0,0 +1,20 @@
+const resolve = require('resolve');
+const path = require('path');
+
+module.exports = {
```
kevinbarabash: Why isn't this standard behaviour for `eslint-plugin-import`?

static/module.html
```diff
@@ -0,0 +1,42 @@
...
+    which supports modules: http://localhost:7936/module.html
+  -->
+<html>
+    <head>
+        <meta charset="UTF-8">
+        <title>KaTeX Test</title>
+        <link rel="stylesheet" type="text/css" href="/build/katex.css">
+        <link rel="stylesheet" type="text/css" href="/main.css">
+        <script type="module" type="text/javascript">
+            import katex from '/build/katex.mjs';
```
kevinbarabash: How is this file being built, I thought `webpack` didn't output ES6 modules?

package.json
ylemkimon: @kevinbarabash Node only uses `main` field, which is extensionless, and when modules are enabled(currently `--experimental-modules`), Node resolves it to `dist/katex.mjs`.

.babelrc
ylemkimon: @kevinbarabash I'm not sure which use cases will be more common, using individual modules or using the main module.

If latter is more common, we can distribute bundled module using rollup.

eslint-import-resolver.js
ylemkimon: @kevinbarabash This blocks importing core modules, e.g., `fs` and `path`, and using `.json` modules, which are supported by Node and standard behavior for `eslint-plugin-import`.

static/module.html
ylemkimon: @kevinbarabash The files are built using `babel` directly with plugins to change `import` paths and rename files.

If we're to distribute the bundled file, we can use `rollup` and simplify the process.

.eslintrc
ylemkimon: @kevinbarabash Files which run on Node and are not included in the build.

.babelrc
ylemkimon: @kevinbarabash Yes, but it uses AST and I didn't want to use another module bundler.

.eslintrc
ylemkimon: @kevinbarabash Yes.

static/module.html
kevinbarabash: I think distributing a single bundle for the ES module is probably the way to go.  That way people can use it in production.

ylemkimon: @kevinbarabash Switched to rollup to bundle the ES module. I'm surprised how light and fast rollup is 😄

ylemkimon:
Another question is should we provide minified `katex.mjs`?

kevinbarabash:
@ylemkimon I think that could be a separate PR, but only if people ask for it.  My take on it is that minification only seems useful in the browser and I've only been using ES6 modules in the browser for testing and development and minification isn't useful in those scenarios.

ylemkimon:
@kevinbarabash Then this is ready for review 😄

static/module.html
```diff
@@ -0,0 +1,42 @@
...
+            const math = document.getElementById("math");
+
+            input.addEventListener("input", reprocess, false);
+            reprocess();
+
+            function reprocess() {
+                try {
+                    katex.render(input.value, math, {displayMode: true, throwOnError: false, macros: {}});
+                } catch (e) {
+                    if (e.__proto__ === katex.ParseError.prototype) {
```
kevinbarabash: I know we do some funky stuff in `ParseError`, does it prevent `e instanceof katex.ParseError` from working?

ylemkimon: @kevinbarabash No, I don't think so. I just copied this from `static/index.html`.

src/Parser.js
```diff
@@ -11,7 +11,7 @@ import unicodeSymbols from "./unicodeSymbols";
 import utils from "./utils";
 import {assertNodeType, checkNodeType} from "./parseNode";
 import ParseError from "./ParseError";
-import {combiningDiacriticalMarksEndRegex, urlFunctionRegex} from "./Lexer.js";
```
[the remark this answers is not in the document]
ylemkimon: @kevinbarabash It no longer requires extension-less import statement and is just a style change.

Issue #1476, opened 2018-07-12
Sizing Commands spacing not right with roots


ry-randall:
Example: `\Huge \sqrt[22]{a}`

<img width="373" alt="screen shot 2018-07-12 at 11 15 31 am" src="https://user-images.githubusercontent.com/16921005/42642484-f6ff264a-85c4-11e8-924e-6e1cea989d38.png">

ronkok:
Well, it’s complicated. It turns out that KaTeX, LaTeX, and MathJax all have their own unique approach to large radicals. Below are three treatments of `\[ \sqrt a \Huge \sqrt a \sqrt[2] a \]`. Those are KaTeX, LaTeX, and MathJax, in that order.
![root](https://user-images.githubusercontent.com/16403058/42662938-59e6448c-85e8-11e8-91d1-add06b276d25.PNG)

There are three reasons why math can get taller or shorter. (1) Stacked elements such as fractions. (2) Style changes such as `\scriptstyle`. (3) Size changes such as `\Huge`. KaTeX deals consistently with tall arguments by choosing tall radical glyphs. LaTeX and MathJax do that for stacked elements but act differently when the source of the tallness is due to a size change such as `\Huge`.

LaTeX `\sqrt` ignores size changes altogether. LaTeX `\sqrt[<root>]` deals with `\Huge` not by picking a taller glyph but by applying a scale factor to the `\normalsize` radical. MathJax does the same thing whether or not the `\sqrt` has an optional root. That’s why MathJax roots fit next to the `\Huge` radical. There is more space there than there is next to a KaTeX unscaled but taller glyph.

I don’t see any reason to match LaTeX behavior in this case. It looks broken to me. Even when the glyph is taller, the root is in the wrong horizontal position. 

I can match the MathJax behavior and that will create enough space so that the root does not interfere with the radical. However, that will be a significant effort so I want some buy-in up front before I start work. Is the MathJax behavior what we want?

ronkok:
An alternate approach would be to let the KaTeX radicals remain as is, and instead vary the root's kern distance. There is currently a `\kern-10mu` on the right side of the root. That distance needs to vary with size if we keep the current radicals.

ry-randall:
Hmm, switching the kern distance based on the size seems like the easiest solution by far, but could likely cause issues in the future. 

For example, if we expect that users might be adjusting those size factors (i.e. map large to 1.1) then they'd have to do it in two places (for the sqrt spacing as well).

[...]

PR #1485, opened 2018-07-15
Fix \Large roots

ronkok:
Addresses issue #1476.

This PR changes how radicals are resized in response to sizing functions such as `\Huge`. It reverts behavior to what KaTeX did prior to release 0.8.0. That is, sizing changes are done by applying a scale factor to a `normalsize` radical. This opens up the right amount of horizontal space and roots then fit.

This PR does not change how radicals are sized in response to stacked elements such as fractions. It also does not change the response to style changes such as `\scriptstyle`.

During discussion of this issue, it was pointed out that sizing functions don't work in LaTeX math mode. This entire mode of behavior could stand to have a warning written when KaTeX is in `strict` mode. This PR does not do that warning. It only fixes the root.

This PR also makes no changes to delimiter sizing. I'm not sure if those should change or not.

Here's a screen shot of the updated behavior:
rootpick


ronkok:
@edemaine I've made a couple of changes. One fixes a logic error and one addresses your line weight concerns. I'll think you'll be pleased with the improvement.

edemaine:
The new screenshots look great!

PR #1486, opened 2018-07-16
Port buildMathML to @flow.

marcianx:


src/buildMathML.js
```diff
@@ -137,22 +159,29 @@ export const buildExpression = function(expression, options) {
...
+): MathDomNode {
     if (!group) {
         return new mathMLTree.MathNode("mrow");
     }
 
     if (groupBuilders[group.type]) {
         // Call the groupBuilders function
-        const result = groupBuilders[group.type](group, options);
+        // $FlowFixMe
+        const result: MathDomNode = groupBuilders[group.type](group, options);
```
ylemkimon: Why `$FlowFixMe`?

src/functions/symbolsOrd.js
```diff
@@ -37,7 +39,7 @@ defineFunctionBuilders({
     htmlBuilder(group, options) {
         return buildCommon.makeOrd(group, options, "textord");
     },
-    mathmlBuilder(group, options) {
+    mathmlBuilder(group: ParseNode<"textord">, options) {
```
ylemkimon: Why these generic types can't be inferred from `type`?

src/buildMathML.js
marcianx: There's no way in the type system to say that `groupBuilders[group.type]` has type `MathMLBuilder<TYPE>` where `group.type` (dynamic) `=== TYPE` (compile-time).

src/functions/symbolsOrd.js
marcianx: IMO, it should be able to, but it's confused by `getVariant(group, options)`. Since the latter takes `group`s of a variety of types, here it might be inferring that even though `type === "textord"`, `type: SymbolNodeType` where `SymbolNodeType` might be defined to be all possible types encompassed by `SymbolParseNode`.

src/buildMathML.js
ylemkimon: But isn't `groupBuilders` an array of `MathMLBuilder`, which is a function returns `MathDomNode`? So if `groupBuilders[group.type]` isn't undefined, wouldn't `groupBuilders[group.type](group, options)` be `MathDomNode`?

marcianx: Sorry for the confusion: the return to isn't the issue here, even though it was written out explicitly; it's the calling of the function (`groupBuilders[group.type](group, options)`) with the given `group` that requires the `$FlowFixMe`.

src/functions/href.js
```diff
@@ -35,7 +37,7 @@ defineFunction({
     },
     mathmlBuilder: (group, options) => {
         const math = mml.buildExpressionRow(group.value.body, options);
-        math.setAttribute("href", group.value.href);
+        assertType(math, MathNode).setAttribute("href", group.value.href);
```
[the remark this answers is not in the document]
marcianx: It would have before as well, since `MathNode` is the only one that has `setAttribute`.

src/buildMathML.js
```diff
@@ -103,29 +119,35 @@ export const getVariant = function(group, options) {
...
+                   lastGroup.getAttribute('mathvariant')) {
+                lastGroup.children.push(...group.children);
+                continue;
+            // Concatenate adjacent <mn>s
+            } else if (group.type === 'mn' && lastGroup.type === 'mn') {
+                lastGroup.children.push(...group.children);
+                continue;
+            // Concatenate <mn>...</mn> followed by <mi>.</mi>
+            } else if (group.type === 'mi' && group.children.length === 1 &&
+                       assertType(group.children[0], TextNode).text === '.' &&
```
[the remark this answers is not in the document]
marcianx: It definitely seems to be the case. I've not been able to come up with a counterexample so far, but I also have not definitively prove it in the `\operatorname` case. In any case, I changed this to be more conservative about the assumptions.

src/functions/href.js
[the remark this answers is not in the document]
marcianx: Yes, that bug is reproducible via `\href{http://google.com}{\sin}`. Could that bug be fixed as a separate PR with test? The purpose of the type-change PR is not to change the fundamental logic.

PR #1487, opened 2018-07-16
\@binrel command and refactor

edemaine:
New `\@binrel{x}{y}` command renders `y` as if it's a bin/rel/ord like `x`. This is equivalent to AMSTeX's `\binrel@{x}\binrel@@{y}`. It will hopefully be useful for macros, such as `\pmb` (#1418).

Also refactor the (already duplicated) code for implementing this functionality into `buildCommon.binrelClass`.

This PR tests via parse trees (as they are equivalent already at this point).  Here's a quick visual demo:

image


src/buildCommon.js
```diff
@@ -774,7 +774,22 @@ const staticSvg = function(value: string, options: Options): SvgSpan {
...
+    // atom type directly.
+    const atomType = (arg.type === "ordgroup" &&
+        arg.value.length ? arg.value[0].type : arg.type);
+    if (/^(bin|rel)$/.test(atomType)) {
+        return "m" + atomType;
+    } else {
+        return "mord";
+    }
+};
+
```
ylemkimon: I think this should be in `src/functions/mclass.js` since it's mclass-related or in `src/defineFunction.js` like `ordargument`.

edemaine: True, 2 out of 3 uses are in `src/functions/mclass.js`.  I originally wrote it this way, but switched for some reason.  Switched back.  This PR should be ready for review now.

src/functions/mclass.js
```diff
@@ -41,6 +42,37 @@ defineFunction({
...
+
+// \@binrel{x}{y} renders like y but as mbin/mrel/mord if x is mbin/mrel/mord.
+// This is equivalent to \binrel@{x}\binrel@@{y} in AMSTeX.
+defineFunction({
+    type: "mclass",
+    names: ["\\@binrel"],
+    props: {
+        numArgs: 2,
+    },
+    handler({parser}, [baseArg, renderArg]) {
```
ylemkimon: Array destructuring polyfill is expensive. (#1469 is pending review)

src/functions/mclass.js
```diff
@@ -52,21 +84,12 @@ defineFunction({
...
-            if (/^(bin|rel)$/.test(atomType)) {
-                mclass = "m" + atomType;
-            } else {
-                // This may capture some instances in which the baseArg is more
-                // than just a single symbol. Say a \overset inside an \overset.
-                // TODO: A more comprehensive way to determine the baseArg type.
-                mclass = "mord";
-            }
+            // LaTeX applies \binrel spacing to \overset and \underset.
+            mclass = binrelClass(baseArg);
```
kevinbarabash: nice cleanup

test/katex-spec.js
```diff
@@ -2999,6 +2999,23 @@ describe("\\tag support", function() {
...
 });
 
+describe("\\@binrel automatic bin/rel/ord", () => {
+    it("should generate proper class", () => {
+        expect("L\\@binrel+xR").toParseLike("L\\mathbin xR");
+        expect("L\\@binrel=xR").toParseLike("L\\mathrel xR");
+        expect("L\\@binrel xxR").toParseLike("L\\mathord xR");
+        expect("L\\@binrel{+}{x}R").toParseLike("L\\mathbin{{x}}R");
+        expect("L\\@binrel{=}{x}R").toParseLike("L\\mathrel{{x}}R");
+        expect("L\\@binrel{x}{x}R").toParseLike("L\\mathord{{x}}R");
```
kevinbarabash: I think I get why we need `{{x}}` here.

Issue #1483, opened 2018-07-15
Document cli.js


edemaine:
There's no documentation for command-line interface `cli.js` in the README, but we should at least mention it.  At the same time, we should mention how to run it from a git clone, as described in https://github.com/Khan/KaTeX/issues/894#issuecomment-404758297

ylemkimon:
`cli.js` arguments seem to be outdated:
https://github.com/Khan/KaTeX/blob/9783ac1466186504f823076c708c04473f8b6761/cli.js#L53-L59

PR #1496, opened 2018-07-22
Switch from nomnom to commander

### LINKED ISSUE #1489: Change `nomnom` to `commander.js` for CLI
KaTeX currently uses `nomnom` for various CLIs. However, `nomnom` is no longer maintained and uses old dependencies, which may bloat `node_modules`. `commander.js` provides nice and simple alternative and is used by many packages.

edemaine:
* Fix #1489
* ~~`--macro` repeatable option replaced with single `--macros` option with ";" separator (until https://github.com/tj/commander.js/issues/674)~~ Never mind, this feature was already added, so `--macro` is now back to normal.
* Add missing `--max-expand` option
* Replace expired `--unicode-text-in-math-mode` option with `--strict` (https://github.com/Khan/KaTeX/issues/1483#issuecomment-405108291)
* Add instructions for how to run in the opening comments
* Clean up option doc strings


ylemkimon:
`dockers/Screenshotter/screenshotter.js` and `test/symgroups.js` also uses `nomnom`.

cli.js
```diff
@@ -1,103 +1,53 @@
 #!/usr/bin/env node
 // Simple CLI for KaTeX.
 // Reads TeX from stdin, outputs HTML to stdout.
+// To run this from the repository, you must first build KaTeX by running
+// `npm install` and `npm run dist`.
+
 /* eslint no-console:0 */
 
 const katex = require("./");
+const json = require("./package.json");
```
ylemkimon: I think the name `package` or using object destructuring, `{version}` is better.

cli.js
```diff
@@ -1,103 +1,53 @@
...
-        full: "output",
-        abbr: "o",
-        metavar: "path",
-        default: null,
-        help: "Write html output to the given file.",
-    })
-    .parse();
+const options = require("commander")
+    .version(json.version)
+    .option("-d, --display-mode", "displayMode",
```
ylemkimon: The second argument seems to be passed by a mistake.

cli.js
```diff
@@ -1,103 +1,53 @@
 #!/usr/bin/env node
 // Simple CLI for KaTeX.
 // Reads TeX from stdin, outputs HTML to stdout.
+// To run this from the repository, you must first build KaTeX by running
+// `npm install` and `npm run dist`.
+
 /* eslint no-console:0 */
 
 const katex = require("./");
```
ylemkimon: How about wrapping this with try...catch and printing helpful message when the module is not found, i.e., KaTeX is not built?

package-lock.json
```diff
@@ -2053,16 +2048,6 @@
         "lazy-cache": "^1.0.3"
       }
     },
-    "chalk": {
```
ylemkimon: Could you try to run `npm dedupe` or deduplicate `chalk` and `ansi-style`? It seems they are duplicated many times.

edemaine:
Ah, I see. Running `screenshotter.sh` (and thus `webpack`) from the `dockers/Screenshotter` directory fails, while running `dockers/Screenshotter/screenshotter.sh` from the root directory succeeds.  Is this a bug in https://github.com/gnandretta/babel-plugin-version-inline/blob/master/src/index.js#L3?  I'm not sure how paths are supposed to work in webpack.  @ylemkimon?

dockers/Screenshotter/screenshotter.js
```diff
@@ -109,11 +67,13 @@ if (opts.exclude) {
...
-let seleniumURL = opts.seleniumURL;
-let seleniumIP = opts.seleniumIP;
-let seleniumPort = opts.seleniumPort;
-let katexURL = opts.katexURL;
-let katexIP = opts.katexIP;
+const browser = opts.browser || "firefox";
+const numAttempts = opts.attempts || 5;
+let seleniumURL = opts.seleniumUrl;
+let seleniumIP = opts.seleniumIp;
+let seleniumPort = opts.seleniumPort || 4444;
```
ylemkimon: You can pass the default value as the third argument or fourth argument(if collector function is present).

dockers/Screenshotter/screenshotter.js
```diff
@@ -27,74 +27,32 @@ const newDir = path.normalize(
...
+        "Comma-separated list of test cases to process")
+    .option("-x, --exclude <tests>",
+        "Comma-separated list of test cases to exclude")
+    .option("--reload", "Reload page for each test")
+    .option("--verify", "Check whether screenshot matches current file content")
+    .option("--diff", "With `--verify`, produce image diffs when match fails")
+    .option("--new",
+        "With `--verify`, generate new screenshots when match fails")
+    .option("--attempts <n>",
+        "Retry this many times before reporting failure (default 5)")
```
ylemkimon: Shouldn't `--selenium-port`, `--katex-port`, and `--attempts` be collected as an integer(`parseInt`)?

ylemkimon:
@edemaine Most `babel` plugins depend on the working directory(`cwd`) being the root, where `package.json` and `.babelrc` resides. The working directory should be root and one can run screenshot tests using `npm run test:screenshots`.

.circleci/config.yml
```diff
@@ -21,7 +21,7 @@ defaults: &defaults
 
     - run:
         name: Verify screenshots and generate diffs and new screenshots
-        command: node dockers/Screenshotter/screenshotter.js --seleniumIP localhost -b $CIRCLE_JOB --verify --diff --new
```
ylemkimon: How did this work?

ylemkimon:
@edemaine Most `babel` plugins depend on the working directory(`cwd`) being the project root, where `package.json` and `.babelrc` resides. Maybe we should `pushd` the project root at the start of `dockers/Screenshotter/screenshotter.sh` and `popd` at the end.

dockers/Screenshotter/screenshotter.js
edemaine: Indeed, you're right.  Thanks!  That wasn't obvious to me from the documentation...

.circleci/config.yml
edemaine: We could probably add an alias for the old form, but given that the screenshotter is internal to developers, I doubt it will be too much trouble.

dockers/Screenshotter/screenshotter.js
edemaine: There's actually a bug in the four-argument case: https://github.com/tj/commander.js/issues/686. But I was able to work around it.

edemaine:
@ylemkimon All comments implemented, and I also restore the `-m`/`--macro` behavior of accumulating multiple macros into an array, reading over the commander.js documentation more closely.

cli.js
```diff
@@ -1,103 +1,64 @@
...
+        0, parseInt)
+    .option("-e, --max-expand <n>",
+        "Limit the number of macro expansions to the specified number, to " +
+        "prevent e.g. infinite macro loops.  If set to Infinity, the macro " +
+        "expander will try to fully expand as in LaTeX.",
+        (n) => (n === "Infinity" ? Infinity : parseInt(n)))
+    .option("-m, --macro <def>",
+        "Define custom macro of the form '\\foo:expansion' (use multiple -m " +
+        "arguments for multiple macros).",
+        (def, defs) => defs.push(def),
```
ylemkimon: `Array.push` returns the new `length`. This should be
```js
(def, defs) => {
    defs.push(def);
    return defs;
}
```

cli.js
```diff
@@ -1,103 +1,64 @@
...
 
-const katex = require("./");
+let katex;
+try {
+    katex = require("./");
+} catch (e) {
+    console.log("KaTeX could not import, likely because dist/katex.js is missing.");
+    console.log("Please run 'npm install' and 'npm run dist' before running");
+    console.log("cli.js from the KaTeX repository.");
+    console.log();
```
ylemkimon: I think `console.error` is more appropriate, as it print to `stderr`.

package.json
```diff
@@ -83,7 +83,7 @@
     "dist:dist": "rimraf dist/ && cp -r build/katex/ dist/"
   },
   "dependencies": {
-    "nomnom": "^1.8.1"
+    "commander": "^2.16.0"
```
ylemkimon: The file permissions of `package.json` and `package-lock.json` have been changed to 755.

edemaine:
Thanks. All fixed now.  (I thought I'd tested the macro arguments, but I guess not since I removed the braces and thereby changed the return value from `null`.  Tested again.)

I also changed the build instructions from `npm run dist` to `npm run build` given that #1500 is about to merge.

cli.js
```diff
@@ -1,103 +1,66 @@
...
+// `npm install` and `npm run build`.
+
 /* eslint no-console:0 */
 
-const katex = require("./");
+let katex;
+try {
+    katex = require("./");
+} catch (e) {
+    console.error("KaTeX could not import, likely because dist/katex.js is missing.");
```
ylemkimon: >  13:1  error  Line 13 exceeds the maximum line length of 84  max-len

edemaine:
@ylemkimon Thanks for the reviews!

PR #1497, opened 2018-07-22
Fix \not vertical alignment

### LINKED ISSUE #1491: not equal sign sometimes renders incorrectly
This formula is interesting as it shows the case with a correct render and wrong render at the same time:

```\begin{aligned} & { z(z-2) \ne 0 } \Rightarrow \left\{ \begin{matrix} z\ne 0 \\ z\ne 2 \\ \end{matrix} \right. \end{aligned}```

image

I used the KaTeX's github.io renderer for the image.

Looks like it has something to do with the height of the matrix as it will be fixed by removing it.   
We can fix it by bringing `matrix` out of the `aligned` environment but it also causes the `rightarrow` and the `{` to get unaligned as it is not really a correct LaTeX formula.  

image

Version: 0.10.0-beta

edemaine:
~~Fix #1491 by changing from `position: absolute` to `position: relative`.~~ This turned out to be buggy; see https://github.com/Khan/KaTeX/pull/1497#issuecomment-406881529 below.

The new implementation defines `\not` to the macro `\mathrel{\mathrlap{\@not}}` where `\@not` is the symbol formerly known as `\not`.  (It's possible to name both the macro and symbol that way, using our new `\char` feature, and https://github.com/Khan/KaTeX/pull/1497/commits/60dd846aee1e68bea26c46f2482c52ac6adff9aa shows how that would look.  But I think it's cleaner to avoid overlapping names.)

Basically, we already have a good way to make a character zero width and spilling over to the right (`\mathrlap`), so why not use it?  This idea was originally discussed in #1267.

I haven't checked the screenshot tests yet, but the test in #1491 now renders as follows:

image

Here's what it looked like with the `position: relative` hack (seems identical):

image


ylemkimon:
Firefox:

edemaine:
I've switched the implementation to use `\mathrlap` -- see the revised PR description.  Screenshot tests are now identical!

kevinbarabash:
I'm amazed that we made a change that didn't change at least one pixel.

PR #1499, opened 2018-07-22
Update fonts to remove \neq and \notin glyphs

edemaine:
See https://github.com/KaTeX/katex-fonts/pull/53


ylemkimon:
We really should move the fonts in to the main repo 😄

edemaine:
@ylemkimon Is repo bloat not a concern?

ylemkimon:
@edemaine I’m not sure how much font files bloat the repo. At most, they are few MBs. I’ve seen repos storing static assets like images, videos, and audios without problems.

But if they bloat the repo, I think we can store only font files on another repo or artifacts storage, like CircleCI build artifacts, possibly deployed using CI, as suggested by @kevinbarabash:

> I guess if we moved font creation into the main repo, we could avoid checking the built fonts in and instead grab the most recent artifact so that people don't have to build fonts themselves when doing KaTeX development.

edemaine:
@ylemkimon That sounds cool!  I like the idea of storing cached built versions somehow, even in another repo (or branch?), with the ability/backup of building the fonts locally (with a few more dependencies -- which maybe we can improve via a Docker image).  Having lived with the split repo for a while, the double PR thing with restricted testing functionality is definitely not ideal.

kevinbarabash:
We'll have to think about what the flow should look like both when switching branches as well as when someone actually wants to make a change to a font.

PR #1484, opened 2018-07-15
KaTeX website & documentation

ylemkimon:
Uses [docusaurus](https://docusaurus.io/), which is a markdown-based static documentation site generation tool. Run `yarn start` to start test server and `yarn build` to build the site.

**Test site:** https://ylemkimon.github.io/KaTeX/ (You can access the documentation by clicking `Installation` or `Documentation`, `Note: this is not an official documentation of KaTeX` is inserted to distinguish from official documentation)

Related: #1330 
Fixes #1483 

- [ ] Dedupe documentations, most documentations are copied from `README.md`s, source codes, and the wiki. Either remove original docs or copy(auto-generate) at the build
- [ ] Setup CircleCI to build and deploy website
- [ ] Add redirects from original documents
- [ ] More examples: #275, #661, #1064 
- [ ] Improve Options(`options.md`) documentation
- [ ] Sandbox/Demo/REPL: #1244 
- [ ] Mention auto-render extension in `api.md`: #378 
- [ ] Better macro documentation: #1096, #1176
- [ ] `Who is using KaTeX` list: #305
- [ ] Mobile-friendly(responsive) main(`index`) page
- [ ] Documentation license like `CC BY-SA`, attribute original author, maybe preserve history?
- [ ] Supported function list/support matrix(maybe auto-generated): #103, #276, #1480 
- [ ] Include other `contrib` documentation
- [ ] Mention ES module support, if #1479 is merged

PR #1500, opened 2018-07-22
Change build directory to `dist` and cleanup NPM scripts

### LINKED ISSUE #1498: Improve NPM script design
I'm finding the split between the `build` and `dist` NPM scripts to be difficult to work with.  Recall the current definitions:

```js
    "build": "npm run prestart && rimraf build/* && webpack",
    "dist": "npm test && npm run build && npm run dist:copy && npm run dist:zip && npm run dist:dist",
    "dist:copy": "cd build && mkdirp katex && cp -r katex.js katex.min.js katex.css katex.min.css contrib fonts ../README.md katex",
    "dist:zip": "cd build && tar czf katex.tar.gz katex/ && zip -rq katex.zip katex/",
    "dist:dist": "rimraf dist/ && cp -r build/katex/ dist/"
```

In simpler terms:
* `build` makes the `build` subdirectory via webpack
* `dist` runs `test` (which takes a long time), `build`, and three other `dist` scripts:
* `dist:copy` copies the things in the `build` subdirectory into the `build/katex` subdirectory (I guess to flatten the directory structure? but it leaves the old copies!)
* `dist:dist` copies those same files into the `dist` subdirectory
* `dist:zip` makes a zip for the Github release

This is annoying because we also define `"main": "dist/katex.js"`.  So to `require("./")` (as in `cli.js` for example) to get the current code, we need to either
1. `npm run dist`, which takes a long time because of `npm test`, or
2. `npm run build && npm run dist:copy && npm run dist:dist`, which seems rather excessive.

I'd like to suggest that `npm run build` builds the files directly into `dist`, or at least does the equivalent by building into `build`, then copying to `dist` (ideally avoiding the second copy into build), and then deletes `build`.  Then the `dist` script should just be `npm test && npm run build && npm run dist:zip`.

@ylemkimon I think you are the last to tweak the NPM script design.

ylemkimon:
Fixes #1498.

- Changes the build directory to `dist` and do not duplicate files.
- Cleanups NPM scripts(**bold** are changed behaviors):
  * `build` makes the **`dist`** subdirectory via **copying** and webpack
  * `dist` runs `test`, `build`, and `dist:zip`
  * `dist:zip` makes a tar and a zip for the GitHub release **at the root**

I think having built files at `build`may make more sense, but it would introduce breaking changes to codes directly using `./node_modules/katex/dist/katex.js` or CDN urls `https://cdn.jsdelivr.net/npm/katex@version/dist/katex.min.js`

This would change archives(`katex.zip` and `katex.tar.gz`) directory structure, but I'm not sure there is anyone relying on this.


.gitignore
```diff
@@ -1,4 +1,6 @@
 build
```
edemaine: Should this `build` be removed?

edemaine: Hmm, now I worry slightly about someone with an existing repo, discovering a new `build` directory marked unknown.  But maybe that's good, encouraging people to delete it.  (`bower` might even have an issue with directory `build` existing, now that it's removed from other lists?)

ylemkimon: @edemaine I originally thought this ignores `website/build/` generated by #1484 too, but its PR adds it as separate entry. I'll remove it.

package.json
```diff
@@ -70,17 +70,15 @@
     "test:screenshots": "npm run test:screenshots:update -- --verify",
     "test:screenshots:update": "npm run prestart && dockers/Screenshotter/screenshotter.sh",
     "test:perf": "npm run prestart && NODE_ENV=test node test/perf-test.js",
-    "clean": "rm -rf build/* node_modules/",
+    "clean": "rm -rf dist/* node_modules/",
```
edemaine: This is old, but perhaps we should use `rimraf` instead of `rm -rf` (for Windows users, presumably).  Also, I don't see why we delete `dist/*` instead of `dist/` here and in `build`.  It seems slightly better to delete the entire directory instead of leaving it blank.

ylemkimon: @edemaine As `rimraf` resides in the `node_modules`, it is impossible to delete `node_modules` with `rimraf` 😄

ylemkimon: I'll change `dist/*` to `dist/`.

ylemkimon:

kevinbarabash:
Thanks for simplifying our `npm` scripts.

PR #1502, opened 2018-07-24
Screenshotter chdir to KaTeX root directory

### LINKED ISSUE #1501: Screenshotter doesn't run from its own directory
Since #1475, the screenshotter needs to be run from the repository's root directory (via `dockers/Screenshotter/screenshotter.sh`); running `screenshotter.sh` from within `dockers/Screenshotter` results in the following (path) error:

```
    Module build failed: Error: ENOENT: no such file or directory, open 'package.json'
        at Object.fs.openSync (fs.js:646:18)
        at Object.fs.readFileSync (fs.js:551:33)
        at Object.<anonymous> (/afs/csail.mit.edu/u/e/edemaine/Projects/KaTeX/node_modules/babel-plugin-version-inline/lib/index.js:27:39)
```

This might be considered a bug in https://github.com/gnandretta/babel-plugin-version-inline/blob/master/src/index.js#L3, but @ylemkimon suggests in https://github.com/Khan/KaTeX/pull/1496#issuecomment-406930265 that we should just change directories before running webpack.

edemaine:
Fix #1501.  This seems to be required for webpack, specifically `babel-plugin-version-inline`, to run correctly.

### DIFF
--- a/dockers/Screenshotter/screenshotter.js
+++ b/dockers/Screenshotter/screenshotter.js
@@ -19,7 +19,7 @@ const data = require("../../test/screenshotter/ss_data");
 
 // Change to KaTeX root directory so that webpack (in particular
 // babel-plugin-version-inline) runs correctly.
-process.chdir(path.join(__dirname, "..", ".."))
+process.chdir(path.join(__dirname, "..", ".."));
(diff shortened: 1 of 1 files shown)


kevinbarabash:
Interesting... babel is supposed to look in parent dirs for .babelrc.

PR #1512, opened 2018-07-28
Add redirect from function-support.html

ylemkimon:


function-support.html
```diff
@@ -0,0 +1,10 @@
+<!DOCTYPE html>
+<html lang="en">
+ <head>
+   <title>Function Support in KaTeX</title>
+   <meta http-equiv="refresh" content="0;URL='./docs/supported.html'" />
```
kevinbarabash: TIL, cool stuff.  For those following along: https://www.w3.org/TR/WCAG20-TECHS/H76.html.

PR #1515, opened 2018-07-28
Improve README.md

ylemkimon:
* Fix KaTeX logo
* Add `npm` badge
* Change CircleCI badge style
* Remove links to original external documentation

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,5 @@
 # [<img src="https://cdn.rawgit.com/Khan/KaTeX/84189cd3adae24d92e766d14eb80d6e54f3c7dca/katex-logo.svg" width="130" alt="KaTeX">](https://khan.github.io/KaTeX/)
+[![npm](https://img.shields.io/npm/v/katex.svg)](https://www.npmjs.com/package/katex)
 [![Build Status](https://travis-ci.org/Khan/KaTeX.svg?branch=master)](https://travis-ci.org/Khan/KaTeX)
 [![CircleCI](https://circleci.com/gh/Khan/KaTeX.svg?style=svg)](https://circleci.com/gh/Khan/KaTeX)
 [![codecov](https://codecov.io/gh/Khan/KaTeX/branch/master/graph/badge.svg)](https://codecov.io/gh/Khan/KaTeX)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -12,7 +13,7 @@ KaTeX is a fast, easy-to-use JavaScript library for TeX math rendering on the we
  * **Self contained:** KaTeX has no dependencies and can easily be bundled with your website resources.
  * **Server side rendering:** KaTeX produces the same output regardless of browser or environment, so you can pre-render expressions using Node.js and send them as plain HTML.
 
-KaTeX supports all major browsers, including Chrome, Safari, Firefox, Opera, Edge, and IE 9 - IE 11. More information can be found on the [list of supported commands](https://khan.github.io/KaTeX/function-support.html) and on the [wiki](https://github.com/khan/katex/wiki).
+KaTeX supports all major browsers, including Chrome, Safari, Firefox, Opera, Edge, and IE 9 - IE 11.
```
edemaine: While I agree that these were the wrong place to link to these, I do wonder whether we should offer some deep links into the documentation.  The supported functions list is quite useful, in particular.  Maybe add a mention under "print quality" (when comparing to TeX?)  "See a list of supported LaTeX functions"?

README.md
```diff
@@ -31,15 +32,10 @@ Call `katex.render` with a TeX expression and a DOM element to render into:
...
 ```
 
-To avoid escaping the backslash (double backslash), you can use
-[`String.raw`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/raw)
-(but beware that `${`, `\u` and `\x` may still need escaping):
-```js
-katex.render(String.raw`c = \pm\sqrt{a^2 + b^2}`, element);
-```
-
 If KaTeX can't parse the expression, it throws a `katex.ParseError` error.
```
edemaine: Is it worth saying "see https://khan.github.io/KaTeX/docs/api.html for details"? (to cover the removed part)

README.md
```diff
@@ -31,15 +32,10 @@ Call `katex.render` with a TeX expression and a DOM element to render into:
...
 
+To render all math expressions on the page, see [Auto-render Extension](https://khan.github.io/KaTeX/docs/autorender.html).
```
edemaine: This is definitely good to add!  I wonder about no longer having the quick start that most people would use (1 link tag, 2 script tags, then call renderMathInElement). I guess that's what's in https://khan.github.io/KaTeX/docs/autorender.html, but it still might be nice to have the minimal version here...

README.md
ylemkimon: @edemaine I don't think `String.raw` is as important as others.

README.md
ylemkimon: @edemaine I'm not sure about the other two comments.

README.md
ylemkimon: @edemaine I've added the link to the list of supported commands.

README.md
ylemkimon: @edemaine I've added a section about the extension.

ylemkimon:
@edemaine Done.

PR #1517, opened 2018-07-28
Add size graphic to function support page, fix vertical alignment

ronkok:
Add a paragraph to the function support page that was omitted earlier. This shows the effect of size and style commands by drawing `\rule` in contrasting styles.


docs/supported.md
```diff
@@ -565,3 +565,13 @@ where:
 F = (font size of surrounding HTML text)/(10 pt)
 
 G = 1.21 by default, because KaTeX font-size is normally 1.21 × the surrounding font size. This value [can be over-ridden](font.md#font-size-and-lengths) by the CSS of an HTML page.
+
+<hr>
```
ylemkimon: You can(and should, as markdown parser sees it as HTML open tag) use `---`.

ronkok:
I've also change the table cell vertical alignment to `middle`. The environment table looks better now.

PR #1518, opened 2018-07-28
Add `markdown-help` command and auto-generate CLI documentation

### LINKED ISSUE #1508: cli self-documentation
https://github.com/Khan/KaTeX/edit/master/docs/cli.md is currently written by hand, but all the option documentation is redundant with https://github.com/Khan/KaTeX/blob/master/cli.js (and currently doesn't incorporate option rewriting that I did in the latter).  Can we automatically include the option documentation from the JavaScript into the Markdown (or vice versa)?

ylemkimon:
Fixes #1508.

* Added `markdown-help` command, which generates markdown documentation, and update CLI documentation
* Auto-generate cli.md before `start`, `build`, and `publish`


cli.js
```diff
@@ -60,8 +60,22 @@ const options = require("commander")
...
+/* program.command('markdown-help')
+    .action(() => {
+        console.log(program.options.map(option => `### \`${option.flags}\`
+${option.description}${((option.bool && option.defaultValue !== undefined)
+    ? ' (default: ' + option.defaultValue + ')' : '')}
+`)
+            .concat(['### `-h, --help`\nOutput usage information'])
+            .join('\n'));
+        process.exit();
+    }); */
```
edemaine: Nice idea on how to do this!  Not fully automatic, but close enough.

I'd suggest making this command always available (i.e. uncomment it out), provided that doesn't cause `cli.js` to otherwise break (I assume it runs command-free otherwise, but haven't tested).  And then we can add a comment to the effect "if you change the options or their documentation, please run this command and also update docs/cli.md".

edemaine:
I was wondering whether there was some way to output this usage information to a file that gets included into `docs/cli.md`.  https://github.com/facebook/Docusaurus/issues/599 suggests that this is best done via a plugin.  But I didn't see a matching plugin... maybe we could write one later.

cli.js
ylemkimon: @edemaine It outputs the command in the help, i.e., `node cli.js --help`.

ylemkimon:
@edemaine We can run scripts to generate docs before building the docs like copying the files in #1520.

docs/cli.md.template
```diff
@@ -0,0 +1,20 @@
...
+
+```bash
+npx katex
+```
+
+> Above uses the `npx` command to run the locally installed executable.
+You can execute with the relative path: `./node_modules/.bin/katex`
+
+> To use CLI from local clone, you need to build the project first by
+running `npm run dist`
```
edemaine: This should be `npm run build` (since your PR fixing up build scripts).

docs/cli.md.template
```diff
@@ -0,0 +1,20 @@
...
+
+## Usage
```
edemaine: Would "Options" or "Command-Line Options" be better than "Usage"? (Not sure.)

ylemkimon: @edemaine Changed to `Options`.

ylemkimon:
@edemaine Thanks for reviews!

Issue #1509, opened 2018-07-27
Provide multiple versions of documentation


kevinbarabash:
This is so that users can look up docs for the version that they're using even if it isn't the most recent version.

ylemkimon:
Docusaurus supports documentation versioning: https://docusaurus.io/docs/en/versioning.

kevinbarabash:
Awesome!

ylemkimon:
https://github.com/Khan/KaTeX/issues/1505#issuecomment-408504458 by @edemaine 
> When we publish to gh-pages (which I guess we do every release, not every commit, so that we document the release not the master branch?), 

Another question is should we document the `master` branch, i.e., publish the website(`gh-pages`) every commit?

ylemkimon:
I think there can be two workflows:
1. Publish website every commit to `master` and every release
* Use CircleCI to build and deploy the website using GitHub token or SSH deploy key
* Run `deploy` job on `master` branch commit builds and Git tag build
* Commit and use local built CSS and fonts or use CircleCI build artifacts for `master` documentation
  * **CORS is not enabled for CircleCI build artifacts, so it needs a proxy**
* Use CSS and fonts from CDN for release version documentation
* Latest release version documentation is shown by default
* `master` documentation can be accessed by selecting `next`
* **May bloat the repo**

2. Publish website every release
* Publish the website manually in the release script
* Use CSS and fonts from CDN

Or we can switch to `netlify`, which provides with following limitations:

> Network Bandwidth: 100GB/month - Soft
Storage: 100GB - Soft
API requests: 500 requests/minute, 3 deploys/minute* - Hard

* Use netlify to build and deploy the website
* Use local built CSS and fonts
* Use CSS and fonts from CDN for release version documentation
* Latest release version documentation is shown by default
* `master` documentation can be accessed by selecting `next`
* PR preview is available through #1510

ronkok:
> Publish website every release

The function support page has been updated on that schedule for 1½ years. It works fine. I'm not sure about the other pages. We may find a problem and wish to get new docs out in between releases.

edemaine:
I think we're currently in Plan 2 (though adding to release script sounds nice), but with this issue completed, Plan 1 makes sense. This is why I was thinking it might be good for the published compiled documentation (not the doc source) to live in a separate repo, to avoid any bloat.

[...]

PR #1520, opened 2018-07-29
[WIP] Use local built CSS on master documentation

ylemkimon:
* Copy CSS and fonts before `start`, `build`, and `publish`


ylemkimon:
The concern is commiting CSS and fonts may bloat the repo.

docs/supported.md
```diff
@@ -2,8 +2,9 @@
 id: supported
 title: Supported Functions
 ---
+<link rel="stylesheet" href="/KaTeX/static/katex.min.css"/><!-- must use absolute path -->
```
ylemkimon: - [ ] Change to use `baseUrl` from siteConfig

website/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "scripts": {
-    "prestart": "cp ../docs/cli.md.template ../docs/cli.md && node ../cli.js markdown-help >> ../docs/cli.md",
+    "prestart": "cp ../docs/cli.md.template ../docs/cli.md && node ../cli.js markdown-help >> ../docs/cli.md && cp -r ../dist/katex.min.css ../dist/fonts static/static",
```
ylemkimon: Instead of copying files, we can use build artifacts from Circle CI build. However this needs a CORS proxy.

ronkok:
@ylemkimon I fully approve of the content in the `supported.md` file. I also agree completely with the objectives of the other changes in this PR and they all seem to work smoothly emprically.

But regarding build scripts, I'm probably not qualified to pass judgment. Perhaps @kevinbarabash  or @edemaine could weigh in.

edemaine:
I guess the main concern here is repo bloat in the `gh-pages` branch.

I'd like to reraise the question of whether we should have a separate repo for the documentation/website publication (not the source files).  We could move the website to `http://katex.github.io/` which would be served from a repository called `katex.github.io` within the KaTeX project.  (See https://help.github.com/articles/user-organization-and-project-pages/#user-and-organization-pages-sites)  Then we wouldn't have to worry about repo size, and we would be free to put all the versions of documentation there (which really wants to have the CSS/JS/fonts copied).

@kevinbarabash Thoughts?  Should we discuss this in a separate issue?

ylemkimon:
I think we can discuss in #1509.

ylemkimon:
@ronkok This PR only changes two lines of `supported.md`, so I think it'll be easy to resolve merge conflicts.

website/package.json
kevinbarabash: Using build artifacts from Circle CI sounds nice.  What's involved with a CORS proxy?

ylemkimon:
@kevinbarabash 
> Using build artifacts from Circle CI sounds nice. What's involved with a CORS proxy?

Sorry for late reply, I missed your comment. The build artifacts from CircleCI [doesn’t have CORS header set](https://discuss.circleci.com/t/add-cors-support-to-artifacts-server/806), i.e., allow loading from other domains. So we need a proxy like [cors-anywhere](https://cors-anywhere.herokuapp.com) or a third party object storage like AWS S3 to serve files with correct headers. However, I cannot find reliable proxy. So we need to setup a CORS proxy or an object storage. (CORS proxy with cache is essentially same as object storage)

kevinbarabash:
@ylemkimon it seems like using `netlify` might be easier for hosting `master` docs, but we should probably secure a domain first.  I asked our IT dept about purchasing the domain name and they can totally do it but they need manager approval and my manager is on vacation.  @edemaine you mentioned being able to get the domain and transfer it later.  I'm not sure how that's done, but that's maybe the best option to move things forward.

edemaine:
@kevinbarabash Preference on `katex.io` vs. `katexjs.org`?  See https://github.com/Khan/KaTeX/issues/1509#issuecomment-409772045 (and https://github.com/Khan/KaTeX/issues/1509#issuecomment-411278202 for an alternative).

ylemkimon:
@edemaine I think the shorter, the better. If you and @kevinbarabash don't mind, I'd like to set up the domain to arrange with ongoing PRs.

PR #1521, opened 2018-07-29
Improve release script & enable versioning documentation

### LINKED ISSUE #1428: Improve release script
We should consider adding the following features to the release script:
1. Update `gh-pages` branch's `index.html` to new version. (mentioned in #1424)
2. Update `README`s even when using 1-argument form. (mentioned in #1423)

As a side note, this will only work for people who can directly push to the repo.  I think this should include all NPM package maintainers, but it currently doesn't include me.  Is there a way to only make this possible with an extra argument to `git push`, so that it cannot (easily) be done accidentally, except from the release script?

### LINKED ISSUE #1509: Provide multiple versions of documentation

ylemkimon:
* Enable documentation versioning: Fixes #1509
  * Add version selection page
  * Generate new version(bump) in the release script
* Update CDN urls in the documentation: Fixes #1428 
* Update CDN urls even when using 1-argument form: Fixes #1428


release.sh
```diff
@@ -98,56 +103,86 @@ git checkout "$BRANCH"
...
 git clean -fdx dist
 yarn dist
 sed -i.bak -E '/^\/dist\/$/d' .gitignore
 rm -f .gitignore.bak
 git add .gitignore dist/
 
-# Update the version number in CDN URLs included in the README files,
+# Edit docs to use CSS from CDN
+grep -l '{@stylesheet: static/katex.min.css}' docs/*.md | xargs sed -i.bak \
+    's|{@stylesheet: static/katex.min.css}|<link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/katex@./dist/katex.min.css" integrity="sha256-katex.min.css" crossorigin="anonymous"/>|'
```
ylemkimon: This will fail without #1520, as no documentation containing `{@stylesheet: static/katex.min.css}` exists.

release.sh
```diff
@@ -98,56 +103,86 @@ git checkout "$BRANCH"
...
+
+# Update the version number in CDN URLs included in the README and the documentation,
 # and regenerate the Subresource Integrity hash for these files.
-node update-sri.js "${VERSION}" README.md contrib/*/README.md dist/README.md
+node update-sri.js "${VERSION}" README.md contrib/*/README.md dist/README.md \
+    docs/*.md docs/*.md.bak website/pages/index.html
+
+# Generate a new version of the docs and publish the website
+pushd website
+npm run version "${VERSION}"
```
edemaine: Should this be `yarn version "${VERSION}"`?

release.sh
```diff
@@ -98,56 +103,86 @@ git checkout "$BRANCH"
...
+
+# Restore docs to use local built CSS
+# for file in ../docs/*.md.bak; do
+#     mv -f "$file" "${file%.bak}"
+# done
+
+USE_SSH=true npm run publish-gh-pages
```
edemaine: Similarly, should this be `yarn publish-gh-pages`?

website/pages/en/versions.js
```diff
@@ -0,0 +1,109 @@
...
+            </p>
+          </div>
+        </Container>
+      </div>
+    );
+}
+
+Versions.title = 'Versions';
+
+module.exports = Versions;
```
edemaine: I didn't see this in the netlify preview, I guess because there's no `versions.json` yet.  How should I test this?

release.sh
```diff
@@ -98,56 +103,86 @@ git checkout "$BRANCH"
...
+popd
 
 # Make the commit and tag, and push them.
-git add package.json bower.json README.md contrib/*/README.md dist/README.md
+git add package.json README.md contrib/*/README.md dist/README.md \
+    docs/*.md website/
```
edemaine: I'm a little confused why we're adding `website` to the tag.  Should this just be `website/pages/index.html` from the `update-sri.js` command above?

release.sh
```diff
@@ -98,56 +103,86 @@ git checkout "$BRANCH"
...
-# Update the version number in CDN URLs included in the README files,
+# Edit docs to use CSS from CDN (#1520)
+# grep -l '{@stylesheet: static/katex.min.css}' docs/*.md | xargs sed -i.bak \
+#     's|{@stylesheet: static/katex.min.css}|<link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/katex@./dist/katex.min.css" integrity="sha256-katex.min.css" crossorigin="anonymous"/>|'
+
+# Update the version number in CDN URLs included in the README and the documentation,
 # and regenerate the Subresource Integrity hash for these files.
-node update-sri.js "${VERSION}" README.md contrib/*/README.md dist/README.md
+node update-sri.js "${VERSION}" README.md contrib/*/README.md dist/README.md \
+    docs/*.md docs/*.md.bak website/pages/index.html
```
edemaine: Why `docs/*.md.bak`? Are these actually checked in?

release.sh
```diff
@@ -98,56 +103,86 @@ git checkout "$BRANCH"
...
     rm -f package.json.bak
+    git add package.json
+fi
 
-    # Refer to the just-released version in the documentation of the
-    # development branch, too.  Most people will read docs on master.
-    git checkout "v${VERSION}" -- README.md contrib/*/README.md
+# Refer to the just-released version in the documentation of the
+# development branch, too.  Most people will read docs on master.
+git checkout "v${VERSION}" -- README.md contrib/*/README.md docs/*.md website/
```
edemaine: Is "Most people will read docs on master" still true?  It seems to make sense to give them the `nextversion-pre` release number...  But I'm not exactly sure what's best.

website/pages/en/versions.js
ylemkimon: @edemaine Run 
```bash
cd website
yarn version 1.0.0
```

release.sh
ylemkimon: @edemaine `yarn version` command copies current `docs` and sidebar into `website/versioned_docs` and `website/versioned_sidebars`, respectively, and edits `versions.json`.

release.sh
ylemkimon: @edemaine They will be restored after `yarn version`, which clones the current documentation.

release.sh
ylemkimon: @edemaine It'll be given `next` version tag, which one can access by selecting in the version page.

website/pages/en/versions.js
edemaine: I get the following (first half looks good, second half looks bad):

```sh
$ yarn version --new-version 0.10.0-beta
yarn version v1.9.2
warning package.json: No license field
warning package.json: No license field
info New version: 0.10.0-beta
$ docusaurus-version
No version number specified!
Pass the version you wish to create as an argument.
Ex: 1.0.0
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/version for documentation about this command.
```

ylemkimon: @edemaine `--new-version` is not needed for `version`.

ylemkimon: @edemaine Yarn has built-in version command: https://yarnpkg.com/en/docs/cli/version. It should be `yarn run version`, sorry.

edemaine: Ahh, got it now, thanks.

release.sh
edemaine: It *might* be preferable to more explicitly add the things you want (although there are other checks for uncommitted changes, I guess).

website/pages/en/versions.js
ylemkimon: @edemaine Also the `version` script doesn't run `prestart`, so it will probably give an error about `cli.md` not found. I'll commit the changes.

edemaine:
I've gotten it working in testing, but I'm still having a hard time seeing how this will work in terms of build.  Should `versions.json` be checked in so that it gets added to over time?  (That's not part of the release script, it seems.)  Does `publish-gh-pages` (for example) really load past versions of the Github repo and load the Markdown files from them to build `versioned_docs`?  Via tags or ...?

edemaine:
OK, I understand better now -- thanks for the explanation!

I'm a little concerned about the entire `website` being checked in.  For example, will this include `website/node_modules` (around 61MB)?  Even if not, we're talking about 2.4MB per release.  Many of these files (like `website/build/KaTeX/img`, which is 685KB) will end up identical -- will Git realize that and store only one copy internally?  The `versioned_docs` is nice and tiny (67KB) -- docusaurus even seems to notice when documents haven't changed and only stores one copy of them -- but I'm worried about the rest.

It makes more sense to me for these built products to be consistently in the `gh-pages` branch.  Instead of checking in a fresh new `website` each time, we're adding to the existing one.  Then Git will definitely do the right thing, only storing changed files, and doing diffs within files.  Is this possible?

ylemkimon:
@edemaine 

`website/node_modules` and built files (`website/build`) will not be committed (checked in) to the `master` branch, as they are ignored by `.gitignore`:

https://github.com/Khan/KaTeX/blob/fd6690b98871a4ef408d9bbf5a576311f3098399/.gitignore#L23-L25

I've written `git add website/` (https://github.com/Khan/KaTeX/pull/1521#discussion_r207258662) because #1520 is pending, and we might update the CSS. Sorry for the confusion.

edemaine:
I don't see a response to the following questions:

> I'm a little concerned about the entire website being checked in. [...] we're talking about 2.4MB per release. Many of these files (like `website/build/KaTeX/img`, which is 685KB) will end up identical -- will Git realize that and store only one copy internally? The `versioned_docs` is nice and tiny (67KB) -- docusaurus even seems to notice when documents haven't changed and only stores one copy of them -- but I'm worried about the rest.
> 
> It makes more sense to me for these built products to be consistently in the gh-pages branch. Instead of checking in a fresh new website each time, we're adding to the existing one. Then Git will definitely do the right thing, only storing changed files, and doing diffs within files. Is this possible?

All this would be easier if we could decide to publish the website to another repo...

ylemkimon:
@edemaine By 'checked in', do you mean they are checked in to the `gh-pages` branch? If so, Git will not commit unchanged files from last revision. So if static pages and assets are not changed, only one copy will be stored.

edemaine:
I still might be a bit happier with explicitly doing `git add website/versioned_docs` (and possibly adding to that list later) instead of `git add website` (which feels like it could break if we forget to `.gitignore` something).

PR #1522, opened 2018-07-29
Switch npm to Yarn

### LINKED ISSUE #1490: use yarn instead of npm
`npm` has changed the format of package-lock.json between v5 and v6.  The yarn.lock format is intended to be stable so even if the organization of `node_modules` changes between versions, yarn.lock should remain unchanged.  Also, `yarn install` doesn't update its lock file whereas you have to pass `npm` the `--no-save` option if you want to prevent changes to `package-lock.json.  The downside of this change is that people will have to install yet another package manager.

ylemkimon:
Fixes #1490 

Yarn seems to be definitely faster.


.circleci/config.yml
```diff
@@ -55,23 +55,23 @@ jobs:
...
-          command: npm test -- -- --coverage
+          command: yarn test --coverage
       - run:
           name: Upload code coverage reports to Codecov
           command: ./node_modules/.bin/codecov
 
       - run:
           name: Build KaTeX
-          command: npm run build
+          command: yarn run build
```
kevinbarabash: This could be `yarn build`.

.circleci/config.yml
```diff
@@ -55,23 +55,23 @@ jobs:
...
       - save_cache:
-          key: *npm_cache_key
+          key: *node_modules_cache_key
           paths:
             - node_modules
 
       - run:
           name: Run tests
-          command: npm test -- -- --coverage
+          command: yarn test --coverage
```
kevinbarabash: I didn't realize you could do that with `yarn`.

release.sh
```diff
@@ -74,10 +74,6 @@ if ! command git diff --stat --exit-code HEAD; then
     echo "Please make sure you have no uncommitted changes" >&2
     : $((++INSANE))
 fi
-if ! command npm owner ls katex | grep -q "^$(command npm whoami) <"; then
-    echo "You don't seem do be logged into npm, use \`npm login\`" >&2
-    : $((++INSANE))
-fi
```
kevinbarabash: Is there no equivalent `yarn` command for this?

edemaine: There is `yarn login`: https://yarnpkg.com/lang/en/docs/publishing-a-package/
I'm not sure about `whoami` but it ought to be possible.  `yarn info` might also be a possible replacement for `npm owner ls`...

edemaine:
Do we want to force Yarn upon users (developers I guess) or present both as options?  I like how the current docs present both options under `node.md`...  On the other hand, our build chain needs to do just one.

kevinbarabash:
I think strongly suggesting `yarn` is probably a good direction to go.  Otherwise people may end up committing `package-lock.json` files by accident when submitting PRs.

edemaine:
Relatedly, I find it a bit annoying that using `npm` without `yarn` is now impossible.  E.g., `npm test` will try to run `yarn`.  This might work if we added `yarn` as a `devDependency`, though...

.circleci/config.yml
ylemkimon: @kevinbarabash I thought only certain commands can pass through, but it seems every commands can be pass through:

> It’s also possible to leave out the run in this command, each script can be executed with its name.
Note that built-in cli commands will have preference over your scripts, so you shouldn’t always rely on this shortcut in other scripts.

One place where our script may clash with built-in commands is `clean`, but Yarn won't be using it: https://github.com/yarnpkg/yarn/issues/2438.

edemaine:
@kevinbarabash We could add `package-lock.json` to `.gitignore`.

release.sh
ylemkimon: @edemaine Yarn's login is different from npm's login. It's more like `git config --global user.name "John Doe"; git config --global user.email johndoe@example.com`. 

> Later when you run a command that requires authentication such as `yarn publish`, you will have to enter your password to do so.

ylemkimon:
@edemaine I think we should force developers to use Yarn to produce reproducible builds.

release.sh
edemaine: I found that `yarn config get username` works like `npm whoami` and `yarn info katex maintainers` is pretty close to `npm owner ls katex`.  Unfortunately, `yarn info katex maintainers` is out of date (missing me in particular), though the same was true of `npm owner ls` until I bugged NPM.

ylemkimon: @edemaine Have you tried  `yarn owner list katex`?

edemaine: Oh, nice!  That list is up-to-date for me.

ylemkimon: @edemaine AFAIK, the purpose of this check is that one can't `npm publish` without `npm login`. However, it is possible to `yarn publish` without `yarn login`.

PR #1524, opened 2018-07-29
Add codecov as devDependencies

### LINKED ISSUE #1504: Add codecov to package.json as a devDependency
If we do this, we can cache it on circleci and avoid a separate install.

ylemkimon:
Fixes #1504


edemaine:
@ylemkimon I think you want to remove `sudo npm install -g codecov` from `.circleci/config.yml`.  And maybe replace `codecov` with `npx codecov`?

edemaine:
Incidentally, what is the point of CircleCI running `codecov` in the `test` job?  We're not requiring that it pass, and `codecov/project` is already running such a test and posting about it, right?  Can't we just remove this step of the job altogether?

ylemkimon:
@edemaine 
> I think you want to remove `sudo npm install -g codecov` from `.circleci/config.yml`.  And maybe replace `codecov` with `npx codecov`?

Of course, I forgot 😄 Changed to `./node_modules/.bin/codecov`.

> Incidentally, what is the point of CircleCI running `codecov` in the `test` job?  We're not requiring that it pass, and `codecov/project` is already running such a test and posting about it, right?  Can't we just remove this step of the job altogether?

No, `codecov/project` only reports the result. Individual CIs like CircleCI is responsible for running tests and uploading the report.

Issue #977, opened 2017-11-21
Custom environments


edemaine:
It's be nice to allow the user to define custom environments, such as `unskipmatrix` in #971.

In LaTeX, `\begin{foo} ... \end{foo}` is equivalent to `\foo ... \endfoo`.  So one way would be for `\begin{foo}` to check whether `\foo` and `\endfoo` are defined, and if so, just expand to that.

Alternatively, we could add a custom environments argument to complement the custom macros.

kevinbarabash:
If `\begin{foo}` and `\end{foo}` map to `\foo` and `\endfoo` respectively, how do arguments passed to `\begin{foo}` get transmitted to `\endfoo`?  From the description of `\newenvironment` I would expect any args to be available in both `begdef` and `enddef`, http://www.emerson.emory.edu/services/latex/latex_20.html.

edemaine:
That page says `\enddef` "may not contain any argument parameters. Only the `\begindef` gets the arguments.

muzimuzhi:
Just to provide a more detailed doc about "custom environments". See Sec. 12.6 of [_LaTeX2e: An unofficial reference manual_](http://mirrors.concertpass.com/tex-archive/info/latex2e-help-texinfo/latex2e.html#g_t_005cnewenvironment-_0026-_005crenewenvironment)

aterenin:
Was this ever implemented? The README says that it is possible to map `align` to `aligned`, but I can't find a way to actually do this.

kevinbarabash:
@aterenin good catch.  I should not have included that suggestion without trying it out first.

Lakedaemon:
Hello, I used TeX and ConTeXT a lot (always hated LaTeX)

Is there a way to have 

\startMatrix instead of \begin{matrix} 
\stopMatrix instead of \end{Matrix}

Also, is there a way to define macro like in TeX like
\def\myMacro#1,#2\S{arg1 is #1 and arg2 is #2}

which is able to gobble stuff like \myMacro Wow, great!\S
and spit out arg1 is Wow and arg2 is   great

The TeX macro system is really powerfull (developped a xml parser for Jmdict and Kanjidic and a japanese dictonary with XeTeX once), but is KaTeX macro system on par or is it missing things ?

edemaine:
@Lakedaemon Can you ask your questions in a separate Discussion thread? It's not really related to this issue.

Lakedaemon:
sure, sorry

[...]

PR #1527, opened 2018-07-29
Small tweaks to documentation

edemaine:
I made a quick pass over all the new documentation and corrected a few typos etc.

I think `options.md` might be nice to split into sections now that it's its own document, but I'll leave that for another PR.


docs/issues.md
```diff
@@ -13,8 +13,7 @@ title: Common Issues
   `\\[0.1em]` instead of the standard line separator distance.
 - KaTeX does not support the `align` environment because LaTeX doesn't support
   `align` in math mode.  The `aligned` environment offers the same functionality
-  but in math mode, so use that instead or define a macro that maps `align` to
-  `aligned`.
+  but in math mode, so use that instead.
```
edemaine: We've never been able to define such a macro. See #977.

PR #1528, opened 2018-07-29
Replace `yarn run` and `yarn install` with `yarn`, remove building with `npm` instructions

ylemkimon:
* Replaced `yarn run ...` with `yarn ...`

From https://github.com/Khan/KaTeX/pull/1522#discussion_r205983374:
I thought only certain commands can pass through, but it seems every commands can pass through:

> It’s also possible to leave out the run in this command, each script can be executed with its name.
Note that built-in cli commands will have preference over your scripts, so you shouldn’t always rely on this shortcut in other scripts.

One place where our script may clash with built-in commands is `clean`, but Yarn won't be using it: https://github.com/yarnpkg/yarn/issues/2438.

* Replaced `yarn install` with `yarn`

> Default Command
Running yarn with no command will run `yarn install`, passing through any provided flags.

* Updated `node.md`: remove building with `npm` instructions, it's impossible to build without Yarn installed and for reproducible builds

### DIFF
--- a/cli.js
+++ b/cli.js
@@ -2,7 +2,7 @@
 // Simple CLI for KaTeX.
 // Reads TeX from stdin, outputs HTML to stdout.
 // To run this from the repository, you must first build KaTeX by running
-// `yarn install` and `yarn run build`.
+// `yarn install` and `yarn build`.
 
 /* eslint no-console:0 */
 
@@ -12,7 +12,7 @@ try {
 } catch (e) {
     console.error(
         "KaTeX could not import, likely because dist/katex.js is missing.");
-    console.error("Please run 'yarn install' and 'yarn run build' before running");
+    console.error("Please run 'yarn install' and 'yarn build' before running");
     console.error("cli.js from the KaTeX repository.");
     console.error();
     throw e;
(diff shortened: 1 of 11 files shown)


cli.js
```diff
@@ -2,7 +2,7 @@
 // Simple CLI for KaTeX.
 // Reads TeX from stdin, outputs HTML to stdout.
 // To run this from the repository, you must first build KaTeX by running
-// `yarn install` and `yarn run build`.
+// `yarn install` and `yarn build`.
```
edemaine: I read that `yarn install` is the default behavior, so the first could be replaced by `yarn`.  Similarly elsewhere in this PR.

ylemkimon: @edemaine Done!

docs/node.md
```diff
@@ -22,7 +22,7 @@ yarn global add katex
 
 ## Building from Source
 
-To build you will need Git, Node.js 6.9 or later, and npm or Yarn.
+To build you will need Git, Node.js 6.9 or later, and Yarn(recommended) or npm.
```
edemaine: Please add a space before the parenthesis.

ylemkimon:
@edemaine I removed the `npm` part from the `Building from Source`.

PR #1529, opened 2018-07-29
 Comments without terminating newlines, \href fixes, \url support

### LINKED ISSUE #1506: Single line comment error
`\[ x = 1 % comment \]` results in an error. The error goes away if a linefeed is placed after the comment.

comment

edemaine:
1. **Comments without terminating newlines in nonstrict mode:** Fix #1506 by allowing single-line comments (`%` without terminating newline) in nonstrict mode.  `Lexer` and `MacroExpander` now store the `Settings` object, so the `Lexer` can complain about missing newline according to the `strict` setting.  I filtered this out from the snapshot tests with a slightly different `replacer`.

2. **Reimplement `\href` like `\verb`:** Major restructuring to lex URL arguments directly, e.g. to support `\href%{hello}` and `\href{http://foo.com/#test%}{hello}`.  The new URL parsing code is simpler, but involves a special case in `parseSymbol` like `\verb`.

3. **Add support for `\url`** while we're here.

### DIFF
--- a/src/Parser.js
+++ b/src/Parser.js
@@ -890,6 +890,10 @@ export default class Parser {
             return newFunction(nucleus);
         } else if (/^\\(href|url)[^a-zA-Z]/.test(text)) {
             const match = text.match(urlFunctionRegex);
+            if (!match) {
+                throw new ParseError(
+                    `Internal error: invalid URL token '${text}'`, nucleus);
+            }
             const funcName = match[1];
             // match[2] is the only one that can be an empty string,
             // so it must be at the end of the following or chain:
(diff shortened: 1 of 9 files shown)


src/Parser.js
```diff
@@ -957,6 +888,45 @@ export default class Parser {
...
+            // hyperref package allows backslashes alone in href, but doesn't
+            // generate valid links in such cases; we interpret this as
+            // "undefined" behaviour, and keep them as-is. Some browser will
+            // replace backslashes with forward slashes.
+            const url = rawUrl.replace(/\\([#$%&~_^{}])/g, '$1');
+            const protocol = /^\s*([^\\/#]*?)(?::|&#0*58|&#x0*3a)/i.exec(url);
+            const allowed = this.settings.allowedProtocols;
+            if (!utils.contains(allowed,  "*") && !utils.contains(allowed,
+                    protocol != null ? protocol[1] : "_relative")) {
+                throw new ParseError('Not allowed \\href protocol', nucleus);
```
ylemkimon: Now we have `\url` too.

edemaine:
@ylemkimon Fixed the error message and flow errors.  Thanks for the review!

test/katex-spec.js
```diff
@@ -2451,33 +2452,69 @@ describe("operatorname support", function() {
...
+    it("should allow empty URLs", function() {
+        expect`\href{}{example here}`.toBuild();
+        expect("\\url{}").toBuild();
+    });
+
+    it("should allow single-character URLs", () => {
+        expect`\href%end`.toParseLike("\\href{%}end");
+        expect`\href %end`.toParseLike("\\href{%}end");
+        expect("\\url%end").toParseLike("\\url{%}end");
+        expect("\\url %end").toParseLike("\\url{%}end");
```
kevinbarabash: Why the mix of 
```
expect`...`
```
and 
```
expect(...)
```
?

edemaine: `\u` doesn't work in a raw string, because ES thinks it's the beginning of a Unicode escape. (This is mentioned in our documentation.) This means we need to use the old fashioned method for `\url`, sadly. (`\\u` also doesn't work.)

kevinbarabash: There was a revision to the template string spec for ES2018.  I assumed that babel supported it but after checking http://kangax.github.io/compat-table/es2016plus/ it appears not.

PR #1532, opened 2018-07-30
Typo in \hspace*, cleanup TeX commands

edemaine:
Two small fixes:
1. `\hspace*` was accidentally rendering as `\hspace\*` (extra escaping within back ticks)
2. Although technically correct, I think wrapping the two-sided TeX commands `\over` and `\choose` in braces makes it clearer how to use them. Happy to revert these if people prefer them as they are.


ronkok:
> wrapping the two-sided TeX commands \over and \choose in braces 

I have often thought of doing that. I never got around to writing it into the file.

PR #1533, opened 2018-07-31
Improve Unicode docs

ronkok:
Restore a feature to the function support page in which the Unicode Mathematical Alphanumeric Symbols section renders its letters in the relevant font.


kevinbarabash:
@ronkok thanks for adding this back in.  Fractur and Double-struck aren't fully supported.  Maybe we should show only the supported character ranges for those type faces.

screen shot 2018-07-30 at 11 44 16 pm

ronkok:
@kevinbarabash Thanks for the note. In Fractur and blackboard-bold, we actually do have the range that the table indicates. But we have it only in math mode. A few of the characters are not supported in text mode. So I had to change how the table calls them.

Also, I've found a bug in sans-serif bold and with some numerals. I'll fix it and turn in a PR after #1534 lands.

table

Issue #1492, opened 2018-07-19
Refactor ParseNode to be plain objects


marcianx:
Original brainstorm in https://github.com/Khan/KaTeX/issues/892.

This will just change `class ParseNode<TYPE>` to `type ParseNode<TYPE> = ...` which would be a union of all possible `ParseNode` types. To avoid needing to do this all at once, will require some type indirection so that we don't refer to the Class version of `ParseNode` all over the place. If that doesn't end up working out, we might have to do an all-at-once style change.

marcianx:
Yea...I think it might have to be done all at once to avoid getting into "this type is not compatible with that type" style errors which provide not additional insight and dramatically progress.

kevinbarabash:
I wonder if we should introduce an `AtomNode` to encapsulate `bin`, `close`, `op` etc. but inside of something that looks like:
```
{|
   type: "atom",
   family: "bin" | "close" | "op" | ...,
   nucleus?: ParseNode,
   subscript?: ParseNode,
   superscript?: ParseNode,
|}
```
This makes our `ParseNode` type more regular and should make flattening `ParseNode`,  getting rid of `value`, easier (I assume this is what was meant by "plain objects").

marcianx:
Oh yes, thanks for the reminder. That should be done first.

kevinbarabash:
> And indeed, I mean plan objects.

Sorry, should've read the original post more carefully.  Changing `class ParseNode<TYPE>` to `type ParseNode<TYPE> = ...` seems pretty safe.

marcianx:
@kevinbarabash It's worth noting that the `"atom"` refactor requires merging the various `src/functiosn/symbols*.js` into one.

marcianx:
Actually, I can keep them in separate files and import them all into a new `symbol.js` that dispatches to these handlers.

marcianx:
@kevinbarabash Before I go further (I've already gone pretty far, actually), do you think there should be separate "atom" types for "mathord", "textord", "op", and "spacing"?
Also, we don't have any HTML/MathML builder handlers for "accent-token" and "op-token". I'm not 100% certain (haven't done a deep audit) how precisely those are handled and whether we should do anything about separating them. Any insights would be appreciated.

kevinbarabash:
[...]

PR #1534, opened 2018-07-31
Convert ParseNode to struct

marcianx:
- This converts `ParseNode` to flat objects per #892 and fixes additional type errors discovered by flow.
- Pro: This works better with the flow type system (I think the use of `NodeValue<TYPE>` used to constrain flow's reasoning sometimes).
- Con: It is less obvious/easily-greppable to find where `ParseNode`s are created in the source code.

**NOTE:** This PR has a higher chance of conflicts due to the unavoidable extensiveness of this change. So I'd like to get this in sooner rather than later. This does NOT flatten the objects yet! This reduces the area of merge conflict; and also, flattening is a rabbit hole since sometimes a handful of `ParseNode` types need to be converted together.


kevinbarabash:
Cool.  This is going to take some time to review.  I will attempt to do so tomorrow evening.  Also, we can put other PRs that make code changes on hold until this get's merged.

src/ParseNode.js
```diff
@@ -61,309 +31,535 @@ export type SymbolParseNode =
...
-    ParseNode<"hphantom"> |
-    ParseNode<"vphantom"> |
-    ParseNode<"raisebox"> |
-    ParseNode<"rule"> |
-    ParseNode<"sizing"> |
-    ParseNode<"smash"> |
-    ParseNode<"sqrt"> |
-    ParseNode<"underline"> |
-    ParseNode<"xArrow">;
+export type AnyParseNode = $Values<ParseNodeTypes>;
```
ylemkimon: Is it possible to deduplicate `mode` and `loc` using intersection types(`&`)?

ylemkimon:
```diff
diff --git a/src/ParseNode.js b/src/ParseNode.js
index 26b2bbe..7f8cc10 100644
--- a/src/ParseNode.js
+++ b/src/ParseNode.js
@@ -7,7 +7,14 @@ import type {Token} from "./Token";
 import type {Measurement} from "./units";

 export type NodeType = $Keys<ParseNodeTypes>;
-export type ParseNode<TYPE: NodeType> = $ElementType<ParseNodeTypes, TYPE>;
+type ParseNodeBase ={|
+    mode: Mode,
+    loc?: ?SourceLocation,
+|}
+export type ParseNode<TYPE: NodeType> = {|
+    ...$ElementType<ParseNodeTypes, TYPE>,
+    ...ParseNodeBase
+|};

 export type LeftRightDelimType = {|
     type: "leftright",
@@ -31,20 +38,19 @@ export type SymbolParseNode =
     ParseNode<"textord">;

 // Union of all possible `ParseNode<>` types.
-export type AnyParseNode = $Values<ParseNodeTypes>;
+export type AnyParseNode = {|
+    ...$Values<ParseNodeTypes>,
+    ...ParseNodeBase
+|};
```
seems to work, not sure it's correct.

marcianx: Nice find on the type spread operator!

Regarding the `ParseNode` definition above.
1. I would really like `ParseNode` to be covariant with its `TYPE` parameter. This means not looking up the value type within its definition. I think that might have been the cause of a lot of "___ is not compatible with ___" style errors which are completely unhelpful. Flow seems to be able to deal with an explicit union of explicitly-defined structs quite easily. An alternative is to replicate `...ParseNodeBase` in every definition, which I can do.
2. As a minor nit: if we _not_ make `ParseNodeBase` a _strict_ object type, then every `ParseNode<TYPE>: ParseNodeBase` as well. Not sure if we would every use that fact, though.
3. The `AnyParseNode` as defined above is not technically a union of all the structs and the behavior of Flow's type checking (especially when it gives an error) might get very confusing in such cases. E.g. Flow will allow you to do the following with a union that type it won't with the type defined above.
   ```javascript
   type AllTypes = {| type: "foo", foo: number |} |
                   {| type: "bar", bar: string |};
   function(val: AllTypes)  {
     let len;
     if (val.type === "foo") {
       len = val.foo + 1;
     } else {
       len = val.bar.length;
     }
   }
   ```
   [Playground](https://flow.org/try/#0C4TwDgpgBAggNnAKuCBnKBeKBvAPlUSALigCIAzAe0tIBoorKSA7AVwFsAjCAJylwC+-HPkIQSpTgEMedKNJ4lUwHgEtmAc34CA3ACg95VswDGwVZWZQAjqwAeACgBuUuCXhIUqAJRQcevzgIYCgg5n0-VXIoZ1cAOjFMDCwKalJfbAC-UIgrLBc4OMYoAGooAEYIqCEIOFRoTOycvKgCuIU4sI1gAAsqgT0BoA)

   I was thus hoping to eventually change uses of `checkNodeType` to be simple `type` checks.

marcianx: @ylemkimon Actually, using
```
type ParseNodeBase = {
    mode: Mode,
    loc?: ?SourceLocation,
};

type ParseNodeTypes = {
    "array": {|
        type: "array",
        ...ParseNodeBase,
        value: ArrayEnvNodeData,
    |},
    ... // and same on the rest
````
seems to freeze the flow type checker very consistently (it's stuck at 24% progress at 100% CPU utilization). So let's skip that for now...

ylemkimon:
As we don't have `ParseNode` class anymore, we should consider move type declarations to `types.js` and functions to `utils.js`.

marcianx:
@ylemkimon IMO, `ParseNode` is central enough that it would really help to keep everything related to it -- it's functions and "methods" (free functions in this case) -- together for the sake of disciverability and learnability. Also, as it's immense, it may overshadow other standalone types in `types.js`, making them less discoverable.

But I'm open to any feedback you may have on any advantages of splitting it up.

ylemkimon:
@marcianx I was thinking the PascalCase file name may give an impression that it default-exports a class named `ParseNode`, but I cannot think of better alternative.

marcianx:
That's a valid concern. How about `parsenode.js` then?

edemaine:
I think `parseNode.js` would be more consistent.

marcianx:
SGTM. Renamed to `parseNode.js` and also rebased to head, migrating the relevant code.
Bizarre thing: flow is strange ... renaming `ParseNode.js` to `parseNode.js` started to ignore one of my explicit `$FlowFixMe`s in `environment/array.js`. I had to reorder code around to make it accept the code again!

marcianx:
This was towards #1492.

kevinbarabash:
@ylemkimon thanks for reviewing and approving.

PR #1541, opened 2018-08-02
Introduce "atom" parse node to coalesce various symbol nodes.

marcianx:
Towards #1492.

### DIFF
--- a/src/functions/mclass.js
+++ b/src/functions/mclass.js
@@ -45,15 +45,14 @@ defineFunction({
     mathmlBuilder,
 });
 
-export const binrelClass = (arg: AnyParseNode) => {
+export const binrelClass = (arg: AnyParseNode): string => {
     // \binrel@ spacing varies with (bin|rel|ord) of the atom in the argument.
     // (by rendering separately and with {}s before and after, and measuring
     // the change in spacing).  We'll do roughly the same by detecting the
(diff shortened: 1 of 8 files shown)


src/functions/mclass.js
```diff
@@ -45,15 +45,14 @@ defineFunction({
...
     // \binrel@ spacing varies with (bin|rel|ord) of the atom in the argument.
     // (by rendering separately and with {}s before and after, and measuring
     // the change in spacing).  We'll do roughly the same by detecting the
     // atom type directly.
-    const atomType = (arg.type === "ordgroup" &&
-        arg.value.length ? arg.value[0].type : arg.type);
-    if (/^(bin|rel)$/.test(atomType)) {
-        return "m" + atomType;
+    const atom = (arg.type === "ordgroup" && arg.value.length ? arg.value[0] : arg);
+    if (atom.type === "atom" && (atom.family === "bin" || atom.family === "rel")) {
```
marcianx: In case you are wondering, I changed from a regex to string equality comparison mostly based my failing to find this while looking for `"bin"` and `"rel"` throughout the codebase. Otherwise, it was getting difficult to find the cause of the jest errors. I eventually guessed to look for all instances of `"m"` which is how I ran into this.

PR #1542, opened 2018-08-03
Flatten ParseNodes: "supsub" and related ("accent", "accentUnder", "horizBrace", "xarrow", "op").

marcianx:
Towards #1492.


kevinbarabash:
@marcianx this is awesome!

Issue #1546, opened 2018-08-04
\neq ligature vs line wrap


ccorn:
Using katex-0.10.0-beta.
When rendering some text with an inline formula `i\neq\frac{d}{2}`, the line got wrapped around where the symbol for `\neq` was supposed to be. Funnily, the `neq` symbol got split into / (at the end of the first line) and = (at the beginning of the following line):

![strange_neq](https://user-images.githubusercontent.com/25403527/43675943-01701cae-97d8-11e8-97f5-2680fb81b68a.png)

When rescaling (Ctrl+)  such that the lines get wrapped elsewhere, the correctly composed `\neq` symbol reappears. Which makes the issue look like a breaking ligature to me.

ronkok:
@ccorn Thank you for the report. We'll fix it in the next release. For now, you can substitute `\mathrel{\neq}`.

Line wraps are still new to KaTeX and we have to get accustomed to using such wraps in the macros we write internally. Looking at the code, I see that `\ne`, `\nottin`, and some others also need such a wrap.

ccorn:
Thanks @ronkok for looking into this. Yes, using `\mathrel` fixes the issue.

PR #1548, opened 2018-08-04
Fix \neq line breaks

ronkok:
Wrap the `\neq` macro inside a `\mathrel`, to prevent an inadvertent line break.  Fixes issue #1546.


kevinbarabash:
We should probably add a test to avoid future regressions.

edemaine:
That would be nice. It's a little tricky to design line break tests, but maybe we can come up with something...

ronkok:
Can we use `jest` to prove that `\neq` produces a group? That's enough to prevent a line break.

edemaine:
Yes, that's a good idea. (An htmlmathml containing an ordgroup.) We could test it explicitly or use a snapshot test. Either way, it'll be easy to test various macro-based operators.

PR #1549, opened 2018-08-04
Fix wide characters

ronkok:
This PR fixes problems with Unicode wide characters that I discovered while creating the new Support Table page for the documentation. This PR:

* adds two classes to the CSS
* fixes an offset error for wide numerals
* adds a screenshotter test to prevent recurrance


kevinbarabash:
Good call adding digits to those tests.

PR #1550, opened 2018-08-05
Fix \fbox

ronkok:
In LaTeX, `\fbox`, like the typical TeX box, will accept only text,  not math. I was unaware of that when I wrote PR #670, so I allowed it to accept math.  This PR corrects that error and permits only text in an `\fbox`. 

`\boxed` still accepts math, as it should.

This is a breaking change only for `\fbox` and it has never, so far, been written into the documentation. I'm not sure how that should affect version numbering. The upcoming Support Table page will be the first appearance of `\fbox` in the documentation, so maybe this PR should be merged before @kevinbarabash publishes to gh-pages.


ronkok:
And to be clear, this bug and the wide character bug fixed in PR #1549 are the only two bugs that I saw during creation of the Support Table page. I don't know of any more existing bugs.

kevinbarabash:
@ronkok thanks for fixing this.  It just means that the next publish will be 0.11.  We're still at 0.10-beta.  I wonder about the utility of the -beta moniker. 🤔

ronkok:
> I wonder about the utility of the -beta moniker.

I think KaTeX has fewer bugs right now than several earlier releases that had no alpha/beta label. I would say the "alpha" is unnecessary.

edemaine:
Agreed. I think we only need to use alpha/beta/pre for testing the release infrastructure. But hopefully it's not even necessary at this point, and we can do bug fix (0.11.1 etc.) releases if needed.

kevinbarabash:
Okay... so once we test the new release script and verify that it works correctly we can do a non-beta release.

PR #1551, opened 2018-08-06
Flatten a bunch of non-pervasive ParseNode types (part 1)

marcianx:
Towards #1492.
Also, this renames some of the `value` sub-fields to something more descriptive.

### DIFF
--- a/src/functions/delimsizing.js
+++ b/src/functions/delimsizing.js
@@ -147,10 +147,7 @@ defineFunction({
         return {
             type: "leftright-right",
             mode: context.parser.mode,
-            value: {
-                type: "leftright-right",
-                value: checkDelimiter(args[0], context).value,
-            },
+            delim: checkDelimiter(args[0], context).value,
         };
     },
 });
@@ -182,7 +179,7 @@ defineFunction({
             mode: parser.mode,
             body,
             left: delim.value,
-            right: assertNodeType(right, "leftright-right").value.value,
+            right: assertNodeType(right, "leftright-right").delim,
         };
     },
     htmlBuilder: (group, options) => {
--- a/src/functions/delimsizing.js
+++ b/src/functions/delimsizing.js
@@ -52,7 +52,7 @@ const delimiters = [
     ".",
 ];
 
-type IsMiddle = {value: string, options: Options};
+type IsMiddle = {delim: string, options: Options};
 
 // Delimiter functions
 function checkDelimiter(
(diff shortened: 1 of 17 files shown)


src/parseNode.js
```diff
@@ -418,30 +370,19 @@ type ParseNodeTypes = {
...
     "vphantom": {|
         type: "vphantom",
         mode: Mode,
         loc?: ?SourceLocation,
-        value: {|
-            type: "vphantom",
-            body: AnyParseNode,
-            value: AnyParseNode[],
-        |},
+        body: AnyParseNode,
```
kevinbarabash: Why don't we need the `AnyParseNode[]` that was originally on the phantoms?

src/parseNode.js
```diff
@@ -344,57 +316,37 @@ type ParseNodeTypes = {
...
         type: "mclass",
         mode: Mode,
         loc?: ?SourceLocation,
-        value: {|
-            type: "mclass",
-            mclass: string,
-            value: AnyParseNode[],
-        |},
-    |},
-    "mod": {|
```
kevinbarabash: It's nice that we can get rid whole nodes, but where did `mod` go?

src/functions/mathchoice.js
```diff
@@ -6,19 +6,16 @@ import Style from "../Style";
...
+import type {ParseNode} from "../parseNode";
 
+const chooseMathStyle = (group: ParseNode<"mathchoice">, options) => {
+    switch (options.style.size) {
+        case Style.DISPLAY.size: return group.display;
+        case Style.TEXT.size: return group.text;
+        case Style.SCRIPT.size: return group.script;
+        case Style.SCRIPTSCRIPT.size: return group.scriptscript;
+        default: return group.text;
+    }
```
kevinbarabash: nice cleanup

src/functions/delimsizing.js
```diff
@@ -134,11 +127,10 @@ defineFunction({
 });
 
 
-function leftRightGroupValue(group: ParseNode<"leftright">): LeftRightDelimType {
-    if (!group.value.body) {
+function assertParsed(group: ParseNode<"leftright">) {
```
kevinbarabash: I guess since we're in delimsizing.js this can be a more general name.

src/parseNode.js
edemaine: `mod` was replaced with macros a while back. I guess I forgot to remove the type.

src/parseNode.js
marcianx: `value = ordargument(body)`. There was no reason to store both.

Plus I've gotten allergic to anything named `value`. :) We should probably rename the rest to have names that say what they are.

PR #1552, opened 2018-08-06
Flatten a bunch of non-pervasive ParseNode types (part 2)

marcianx:
Towards #1492.


src/parseNode.js
```diff
@@ -34,11 +34,8 @@ type ParseNodeTypes = {
         type: "color",
         mode: Mode,
         loc?: ?SourceLocation,
-        value: {|
-            type: "color",
-            color: string,
-            value: AnyParseNode[],
-        |},
+        color: string,
+        body: AnyParseNode[],
```
kevinbarabash: Much better!

kevinbarabash:
@ylemkimon thanks for reviewing all of these large refactors.

PR #1555, opened 2018-08-06
Add ignore option for class names

otacke:
This change adds another option for ignoring nodes for rendering similar to `ignoreTags`.
The new option `ignoreClasses` can hold an array of strings representing class names that should be ignored if they are in a node's class list. This can be desirable in dynamic environments.


otacke:
[clabot:check]

contrib/auto-render/auto-render.js
```diff
@@ -61,9 +61,24 @@ const renderElem = function(elem, optionsCopy) {
             elem.replaceChild(frag, childNode);
         } else if (childNode.nodeType === 1) {
             // Element node
-            const shouldRender = optionsCopy.ignoredTags.indexOf(
+            let shouldRender = optionsCopy.ignoredTags.indexOf(
                 childNode.nodeName.toLowerCase()) === -1;
```
ylemkimon: IE9 doesn't support `classList`. You can do:
```js
// normalize className to use indexOf for searching
const className = ' ' + childNode.className + ' ';
const shouldRender = optionsCopy.ignoredTags.indexOf(
    childNode.nodeName.toLowerCase()) === -1 &&
    optionsCopy.ignoreClasses.every(
        x => className.indexOf(' ' + x + ' ') === -1)
```

otacke: Thanks for the hint! Didn't know that your support covers this old browser version, too. Will fix it.

otacke: @ylemkimon I changed the code. Thanks again for the hint! I also learned that every is supported by IE9, so no need to do everything "manually" with the for-loops.

ylemkimon:
The default should be an empty array or a null check should be added.

otacke:
I'll gladly add that.

otacke:
@ylemkimon This should do now. Thanks for investing all this time into the project!

PR #1556, opened 2018-08-06
Sync support pages

ronkok:
This PR adds 17 functions, mostly dots and negative spaces, to `supported.md`. These functions turned up during creation of the Support Table page.  It also fixes a color link, removes one un-supported function (`\skip`) and does a little rearrangement of the symbols table.


ronkok:
When I run `yarn start` on this branch, the resulting page does not render the `\url` function properly. If this page were to be published before the next KaTeX release, how would `\url` render on the public web?

ylemkimon:
> If this page were to be published before the next KaTeX release, how would \url render on the public web?

I am thinking of using the release version of KaTeX from the CDN.

Two things not directly related to this PR:
* What do you think about using KaTeX logo(`\KaTeX`) in the `\href` example instead of plain text?
* There seems to be an overflow on the smaller screens. Do you have any thoughts?

ronkok:
> I am thinking of using the release version of KaTeX from the CDN.

The current release does not include `\url`. So `\url` would not render in that scenario. As a general rule, should we wait until a new release before adding functions to the documentation?

> KaTeX logo in the `\href` example

Good idea. I'll give that a try.

> overflow on the smaller screens

My phone can render it pretty well, but only in landscape mode, not portrait mode. That was also true of the old function support page, so this is nothing new. I don't really have any ideas to make it better.

ronkok:
@ylemkimon I changed the `\href` example so that it displays the KaTeX logo. It works well. Nice touch.

mbourne:
The problem with the page on a phone is due to (in part) a few `<br>`s on the "Direct input" sections (where there are long lists of characters, which are presumably being split for  exactly this reason) being interpreted by Docusaurus as actual `&lt;br&gt;`s instead. For example, there are 2 in this section:

image

These are pushing the page to be super wide, so the tables' `overflow:auto` gets messed up (visually).

When it doesn't really need to be a table (which I believe is the case here), and can just be a list of functions, what I usually do (rather than make phone users scroll wide tables left and right and therefore probably miss a lot of things), is to give the `td`s a `display:inline-block`.

So here's my version of your page which works on both wide and narrow screens: 
https://www.intmath.com/tmp/fns-tbl.html (it's rough and has some loading errors, but the idea's there.)

The `<br>`s have been fixed, and the `<table>`s have:

`@media screen and (max-width:480px){
	table tr:nth-child(2n){background:#fff}
	table td{display:inline-block;background:#fff}
}`

The borders still could be tidied up, but hopefully it helps.

ronkok:
@mbourne Thank you for the tip! I'll get to work on it.

ronkok:
I'm working out the idea suggested by @mbourne for a responsive table. It's not ready to publish yet, but it looks very promising. But there is this: IE 9 does not support `display: block` for table elements.

Should I try to implement a browser hack that keeps the `display: block` trick away from IE 9, or is this the moment when we decide we no longer support IE 9?

edemaine:
Is it just the responsiveness that fails?  It'd be nice if *something* renders in IE 9, just not as beautiful as otherwise.

ronkok:
On second thought, IE 9 should not be a problem. The special CSS will be activated only when `@media screen and (max-width:480px)`. I don't many people are using IE 9 on their phone.

ylemkimon:
@ronkok Also, mostly developers will be using the site and they won't be using IE9 😄 

Do you want to work on this PR or separate PR?

ronkok:
@ylemkimon I'm working it on this PR. I also have some other time demands so this may take a day or two.

mbourne:
@ronkok I meant to add that you might want to consider "break-word" for those long strings of non-space characters. No need to add `<br>`, and it's cleaner. 

` overflow-wrap: break-word;
  word-wrap: break-word`

See: https://css-tricks.com/almanac/properties/o/overflow-wrap/

ylemkimon:
@ronkok Could you change `In master` functions to display the actual function?

ylemkimon:
@mbourne The support table (https://katex.netlify.com/docs/support_table.html) also has overflowing tables but doesn't overflow on smaller screen. I wonder what's the difference.

ronkok:
> Could you change In master functions to display the actual function?

Yep. On it.

> consider "break-word"

Nice idea. I'll see how it affects things.

ronkok:
Status: I've eliminated all "In master" and replaced them with function displays. Ive adopted `break-word` and eliminated all `<br>` tags inside `<p>` elements. I've also done two CSS changes to the tables in the `supported.md` page. These changes are done to facilitate narrow mobile displays and they vary in their success. 

Good parts: Sideways scrolling is completely unnecessary and the tables with headings turned out well. 

Questionable part: The tables without headings are now, on a narrow screen, broken down so that each cell acts like its own `<div>`.  So it all fits, but it is the zaniest hopscotch board you have ever seen.

Take a look at https://deploy-preview-1556--katex.netlify.com and tell me if I need to change it. 

Tomorrow, I'll add the `texvc` functions to `supported.md`. So far they're only in `support_table.md`. (Unless we decide they're not needed in `supported.md`. They're all redundant function. I added them mostly to make the support table look better.)

mbourne:
@ylemkimon On the page you linked to, there are no super-wide non-breaking text elements like this, which the page I massaged does have:

image

@ronkok Your latest link goes to a mock-up of the main intro page, not the page with functions. But there is an oddity on that last expression (the q condition) on a 320 px wide screen, which `white-space:nowrap` might fix:

image

(Those things are a PIA, I know...)

ylemkimon:
@mbourne @ronkok

You can access the supported functions page at https://deploy-preview-1556--katex.netlify.com/docs/supported.html.

The example in the main page and the supported functions page overflows at screen width smaller than 359px, but it's inevitable, latest devices (iPhone > 5, Galaxy > S2) have larger screens and there won't be much use cases of accessing the documentation on screens too small, so I think it's OK.

edemaine:
> But there is an oddity on that last expression (the q condition) on a 320 px wide screen, which white-space:nowrap might fix

You could also just wrap the math in `{...}` to prevent the line break.  Or use display math `\[...\]` (why aren't we using display math?).

ronkok:
The latest `texvc` functions are now in `supported.md`. I think this PR is now ready for review.

docs/supported.md
```diff
@@ -85,25 +95,30 @@ The `{array}` environment supports `|` and `:` vertical separators.
...
 | $\url{https://khan.github.io/KaTeX/}$ | `\url{https://khan.github.io/KaTeX/}` |
 
 ## Letters and Unicode
 
 **Greek Letters**
 
-Direct Input: $Α Β Γ Δ Ε Ζ Η Θ Ι Κ Λ Μ Ν Ξ Ο Π Ρ Σ Τ Υ Φ Χ Ψ Ω$<br>
-$α β γ δ ϵ ζ η θ ι κ λ μ ν ξ o π ρ σ τ υ ϕ χ ψ ω ε ϑ ϖ ϱ ς φ$
+Direct Input: $Α Β Γ Δ Ε Ζ Η Θ$ $Ι Κ Λ Μ Ν Ξ Ο Π Ρ Σ Τ Υ Φ Χ Ψ Ω$
+$α β γ δ ϵ ζ η θ ι κ λ μ ν ξ o$ $π ρ σ τ υ ϕ χ ψ ω ε ϑ ϖ ϱ ς φ$
```
edemaine: You can add `\allowbreak` instead of exiting and leaving math mode.

docs/support_table.md
```diff
@@ -823,7 +823,7 @@ If you know the shape of a character, but not its name, [Detexify](http://detexi
 |\r|$\text{\r{a}}$|`\text{\r{a}}`|
 |\raise|<span style="color:firebrick;">Not supported</span>|see `\raisebox`|
 |\raisebox|$h\raisebox{2pt}{ighe}r$|`h\raisebox{2pt}{ighe}r`|
-|\rang|$\angle A\rang$|`\angle A\rang`|
+|\rang|$\langle A\rang$|`\langle A\rang`|
```
edemaine: It looks a little weird to have the `\rang` example so different from the `\rangle` example right below it.  I'd suggest unifying them one way or another.  Ditto for `\lang` and `\langle`.

docs/supported.md
```diff
@@ -194,57 +220,63 @@ $\displaystyle\sum_{\mathclap{1\le i\le j\le n}} x_{ij}$ `\sum_{\mathclap{1\le i
...
-|$~$                                   | `\global\def\macroname#1#2…{definition}`
-|$~$                                   | `\newcommand\macroname[numargs]{definition}`
-|$~$                                   | `\renewcommand\macroname[numargs]{definition}`
-|$~$                                   | `\providecommand\macroname[numargs]{definition}`
+|                                      | `\global\def\macroname#1#2…{definition}`
+|                                      | `\newcommand\macroname[numargs]{definition}`
+|                                      | `\renewcommand\macroname[numargs]{definition}`
+|                                      | `\providecommand\macroname[numargs]{definition}`
 
 Macros can also be defined in the KaTeX [rendering options](options.md).
```
edemaine: The next line says "up to ten arguments".  That should really be nine, as we start counting at 1.

docs/supported.md
```diff
@@ -254,7 +286,7 @@ Macros accept up to ten arguments: #1, #2, etc.
 
 Available functions include:
 
-`\mathchoice` `\TextOrMath` `\@ifstar` `\@ifnextchar` `\@firstoftwo` `\@secondoftwo` `\relax`
+`\mathchoice` `\TextOrMath` `\@ifstar` `\@ifnextchar` `\@firstoftwo` `\@secondoftwo` `\relax` `\allowbreak` `\nobreak`
```
edemaine: I'm not sure this is the right place for the `break` commands.  Maybe in "Line Breaks and Vertical Layout" around `\\` and `\newline`?

edemaine: `\char` is another recent addition for this list.

docs/supported.md
```diff
@@ -471,24 +504,26 @@ $\textcolor{#228B22}{F=ma}$ `\textcolor{#228B22}{F=ma}`<br>
 $\colorbox{aqua}{A}$ `\colorbox{aqua}{A}`<br>
 $\fcolorbox{red}{aqua}{A}$ `\fcolorbox{red}{aqua}{A}`
 
-For color definition, KaTeX color functions will accept the standard HTML [predefined color names](https://www.w3schools.com/colors/colors-names.asp). They will also accept an RGB argument in CSS hexa­decimal style.
+For color definition, KaTeX color functions will accept the standard HTML [predefined color names](https://www.w3schools.com/colors/colors_names.asp). They will also accept an RGB argument in CSS hexa­decimal style.
```
edemaine: It might be better to link somewhere other than w3schools.  Perhaps https://developer.mozilla.org/en-US/docs/Web/CSS/color_value#Color_keywords?

docs/supported.md
```diff
@@ -471,24 +504,26 @@ $\textcolor{#228B22}{F=ma}$ `\textcolor{#228B22}{F=ma}`<br>
...
-|$\textsf{AB}$ `\textsf{AB}`  |$\mathtt{AB}$ `\mathtt{AB}`  |$\mathfrak{AB}$ `\mathfrak{AB}`
-|$\sf AB$ `\sf AB`|$\texttt{AB}$ `\texttt{AB}`  |$\mathcal{AB}$ `\mathcal{AB}`
-|  |$\tt AB$ `\tt AB`|$\mathscr{AB}$ `\mathscr{AB}`
+|:------------------------------|:------------------------------|:-----
+|$\mathrm{Ab0}$ `\mathrm{Ab0}`  |$\mathbf{Ab0}$ `\mathbf{Ab0}`  |$\mathit{Ab}$ `\mathit{Ab}`
+|$\textrm{Ab0}$ `\textrm{Ab0}`  |$\textbf{Ab0}$ `\textbf{Ab0}`  |$\textit{Ab}$ `\textit{Ab}`
+|$\rm Ab0$ `\rm Ab0`            |$\bf Ab0$ `\bf Ab0`            |$\it Ab$ `\it Ab`
+|$\textnormal{Ab0}$ `\textnormal{Ab0}`|$\bold{Ab0}$ `\bold{Ab0}`|$\Bbb{AB}$ `\Bbb{AB}`
+|$\text{Ab0}$ `\text{Ab0}`      |$\boldsymbol{Ab}$ `\boldsymbol{Ab}`|$\mathbb{AB}$ `\mathbb{AB}`
+|$\mathsf{Ab0}$ `\mathsf{Ab0}`  |$\bm{Ab}$ `\bm{Ab}`            |$\frak{Ab0}$ `\frak{Ab0}`
```
edemaine: Is there a reason that a few of these don't have 0s added?

docs/supported.md
```diff
@@ -502,55 +537,59 @@ One can stack font family, font weight, and font shape by using the `\textXX` v
...
+|$\text{\textbraceright}$ `\text{\textbraceright}`|$\Diamond$ `\Diamond`|$\text{\textdegree}$ `\text{\textdegree}`
+|$\text{\P}$ `\text{\P}`|$\lozenge$ `\lozenge`|$\mho$ `\mho`
+|$\text{\S}$ `\text{\S}`|$\blacklozenge$ `\blacklozenge`|$\diagdown$ `\diagdown`
+|$\text{\sect}$ `\text{\sect}`|$\star$ `\star`|$\diagup$ `\diagup`
+|$\copyright$ `\copyright`|$\bigstar$ `\bigstar`|$\flat$ `\flat`
+|$\circledR$ `\circledR`|$\clubsuit$ `\clubsuit`|$\natural$ `\natural`
+|$\text{\textregistered}$ `\text{\textregistered}`|$\clubs$ `\clubs`|$\sharp$ `\sharp`
+|$\circledS$ `\circledS`|$\diamondsuit$ `\diamondsuit`|$\heartsuit$ `\heartsuit`
+|$\text{\textcircled a}$ `\text{\textcircled a}`|$\diamonds$ `\diamonds`|$\hearts$ `\hearts`
+|$\maltese$ `\maltese`|$\spadesuit$ `\spadesuit`|$\spades$ `\spades`
```
edemaine: FYI, there are screen widths where this does not render very well:

image

The cells are forced to be pretty wide, so I think the only solution would be to switch to two columns...  Eh, other tables look bad at this size too:

image

Not sure we want to deal with this now.

docs/supported.md
```diff
@@ -559,28 +598,39 @@ Direct Input: $£ ¥ ∇ ∞ · ∠ ∡ ∢ ♠ ♡ ♢ ♣ ♭ ♮ ♯ ✓ …
...
+</div>
 
 where:
 
 <div style="margin-left: 1.5em;">
+
 F = (font size of surrounding HTML text)/(10 pt)
 
 G = 1.21 by default, because KaTeX font-size is normally 1.21 × the surrounding font size. This value [can be over-ridden](font.md#font-size-and-lengths) by the CSS of an HTML page.
+
```
edemaine: "overridden" doesn't need a hyphen.

website/static/css/custom.css
```diff
@@ -26,6 +26,102 @@
...
+        display: block;
+    }
+
+    /* Hide table headers (but not display: none;, for accessibility) */
+    .katex-cards thead tr {
+        display: block;
+        position: absolute;
+        top: -9999px;
+        left: -9999px;
+    }
```
edemaine: I confirmed that this is indeed best practice: https://webaim.org/techniques/css/invisiblecontent/#absolutepositioning

These (dis)appearing headers are really cool!

docs/supported.md
ronkok: > Is there a reason that a few of these don't have 0s added?

Yes, those fonts don't contain a "0" glyph.

docs/supported.md
ronkok: >  there are screen widths where this does not render very well:

Currently, both sidebars disappear at the same quite narrow break point. I'll see if I can make just one sidebar disappear at an intermediate break point.  

I'd rather not make the tables more narrow than they already are. If users have to scroll up and down a lot to see just one topic, then I'd prefer to get rid of this page and put a topic filter on the Support Table page.

docs/supported.md
ronkok: `\boldsymbol` and `\bm do render a 0, but it's not bold-*italic*.  It's just bold.

ronkok:
The latest commit picks up review comments. To address the troublesome screen width, I made the right sidebar a little more narrow. This doesn't solve the problem, but it does mitigate it. A solution would require a more extensive CSS modification than I am up to at the moment. What I have here is probably my best effort for the present.

kevinbarabash:
@ronkok thanks for keeping this page up to date.

PR #1557, opened 2018-08-06
ScrollSpy for table of contents

### LINKED ISSUE #1537: ScrollSpy in documentation
It'd be nice, especially in `supported.md`, to have the table of contents on the right automatically highlight the section you're in.  In Bootstrap land this is called [ScrollSpy](https://getbootstrap.com/docs/3.3/javascript/#scrollspy); there's also a [react-scrollspy](https://github.com/makotot/react-scrollspy).  As far as I can tell, [markdown-toc](https://github.com/jonschlinkert/markdown-toc) doesn't have an automatic option for this, while [some other systems](https://ndossougbe.github.io/strapdown/) do (not that I'm suggesting switching -- but we could request the feature upstream).  Perhaps we could add a plugin that transforms the TOC's HTML into a react-scrollspy version?  I'm actually having trouble figuring out where markdown-toc gets used.  Any pointers?

edemaine:
Fix #1537 by adding automatic highlighting of currently visible topmost table of contents item.

This sometimes doesn't do an initial highlight, I guess because the table of contents isn't rendered at the beginning?  Is there a way to fix this?  But it works when scrolling and resizing.

Based on https://github.com/Khan/KaTeX/issues/1537#issuecomment-409616103 and https://github.com/makotot/scrollspy/blob/master/src/js/modules/scrollspy.js

### DIFF
--- a/website/siteConfig.js
+++ b/website/siteConfig.js
@@ -66,7 +66,10 @@ const siteConfig = {
 
     markdownPlugins: [require('./remarkableKatex'), require('./empty_thead')],
 
-    scripts: ['https://buttons.github.io/buttons.js'],
+    scripts: [
+        'https://buttons.github.io/buttons.js',
+        baseUrl + 'js/scrollspy.js',
+    ],
     stylesheets: ['https://cdn.jsdelivr.net/npm/katex@0.10.0-beta/dist/katex.min.css'],
 
     separateCss: ['static/static', 'static\\static'],
--- a/website/static/js/scrollspy.js
+++ b/website/static/js/scrollspy.js
@@ -7,6 +7,8 @@ const onScroll = () => {
     timer = setTimeout(() => {
         let found = false;
         const headings = document.querySelectorAll('.toc-headings > li > a');
+        // Some computations based on
+        // https://github.com/makotot/scrollspy/blob/master/src/js/modules/scrollspy.js
         const scrollTop = document.documentElement.scrollTop ||
             document.body.scrollTop;
         const scrollBottom = scrollTop + window.innerHeight;
--- a/website/static/js/scrollspy.js
+++ b/website/static/js/scrollspy.js
@@ -1,0 +1,35 @@
+const epsilon = 10;
+let timer;
+const onScroll = () => {
+    if (timer) {  // throttle
+        return;
+    }
+    timer = setTimeout(() => {
+        let found = false;
+        const headings = document.querySelectorAll('.toc-headings > li > a');
+        const scrollTop = document.documentElement.scrollTop ||
+            document.body.scrollTop;
+        const scrollBottom = scrollTop + window.innerHeight;
+        for (let i = 0; i < headings.length; i++) {
+            // if !found and i is the last element, highlight the last
+            let current = !found;
+            if (!found && i < headings.length - 1) {
+                const next = headings[i + 1].href.split('#')[1];
+                const nextHeader = document.getElementById(next);
+                const top = nextHeader.getBoundingClientRect().top + scrollTop;
+                current = top > scrollTop + epsilon;
+            }
+            if (current) {
+                found = true;
+                headings[i].className = "active";
+            } else {
+                headings[i].className = "";
+            }
+        }
+        timer = null;
+    }, 50);
+};
+document.addEventListener('scroll', onScroll);
+document.addEventListener('resize', onScroll);
+document.addEventListener('ready', onScroll);
+onScroll();
--- a/website/static/css/custom.css
+++ b/website/static/css/custom.css
@@ -1,3 +1,17 @@
+ul.toc-headings > li {
+    padding-bottom: 0px;   /* moved to li > a */
+}
+
+.toc-headings > li > a {
+    display: block;
+    padding: 4px;
+}
+
+.toc-headings > li > a.active {
+    background-color: rgba(27, 31, 35, 0.05);
+    font-weight: bold;
+}
+
 .fixedHeaderContainer header img {
     height: 80%;
 }


website/siteConfig.js
```diff
@@ -66,7 +66,10 @@ const siteConfig = {
 
     markdownPlugins: [require('./remarkableKatex'), require('./empty_thead')],
 
-    scripts: ['https://buttons.github.io/buttons.js'],
+    scripts: [
+        'https://buttons.github.io/buttons.js',
+        baseUrl + 'js/scrollspy.js',
```
ylemkimon: You can do `/js/scrollspy.js'.

website/static/js/scrollspy.js
```diff
@@ -0,0 +1,37 @@
...
+                headings[i].className = "";
+            }
+        }
+        timer = null;
+    }, 50);
+};
+document.addEventListener('scroll', onScroll);
+document.addEventListener('resize', onScroll);
+document.addEventListener('ready', onScroll);
+onScroll();
```
ylemkimon: Wrap this in a closure (IIFE) to prevent polluting global scope.

website/static/js/scrollspy.js
```diff
@@ -0,0 +1,37 @@
...
+            } else {
+                headings[i].className = "";
+            }
+        }
+        timer = null;
+    }, 50);
+};
+document.addEventListener('scroll', onScroll);
+document.addEventListener('resize', onScroll);
+document.addEventListener('ready', onScroll);
```
ylemkimon: `ready` is only available on jQuery. You should use `DOMContentLoaded`.

website/static/js/scrollspy.js
ylemkimon: The script is included in the `<head>`, so calling `onScroll` here has no effect.

website/static/js/scrollspy.js
```diff
@@ -0,0 +1,37 @@
+const epsilon = 10;
```
ylemkimon: I think the name `offset` is better and as it's a constant, ALL_CAPS is better.

website/static/js/scrollspy.js
```diff
@@ -0,0 +1,37 @@
...
+let timer;
+const onScroll = () => {
+    if (timer) {  // throttle
+        return;
+    }
+    timer = setTimeout(() => {
+        let found = false;
+        const headings = document.querySelectorAll('.toc-headings > li > a');
```
ylemkimon: You can cache (pre-calculate) this value after `DOMContentLoaded`. You can also probably cache positions of their headers, but I'm not sure `DOMContentLoaded` waits for fonts to be loaded and rendered.

website/static/js/scrollspy.js
```diff
@@ -0,0 +1,37 @@
...
+        // Some computations based on
+        // https://github.com/makotot/scrollspy/blob/master/src/js/modules/scrollspy.js
+        const scrollTop = document.documentElement.scrollTop ||
+            document.body.scrollTop;
+        const scrollBottom = scrollTop + window.innerHeight;
```
ylemkimon: `scrollBottom` is not used.

website/static/js/scrollspy.js
```diff
@@ -0,0 +1,37 @@
...
+    if (timer) {  // throttle
+        return;
+    }
+    timer = setTimeout(() => {
+        let found = false;
+        const headings = document.querySelectorAll('.toc-headings > li > a');
+        // Some computations based on
+        // https://github.com/makotot/scrollspy/blob/master/src/js/modules/scrollspy.js
+        const scrollTop = document.documentElement.scrollTop ||
+            document.body.scrollTop;
```
ylemkimon: All browsers except IE<9, which Docusaurus and we don't support, supports `window.pageYOffset` (https://developer.mozilla.org/en-US/docs/Web/API/Window/pageYOffset).

website/static/js/scrollspy.js
ylemkimon: I think we can cache positions of headers based on the `scrollHeight`. i.e., update cached positions if `scrollHeight` has been changed. (AFAIK, modern browsers support `document.documentElement.scrollHeight`)

website/static/js/scrollspy.js
```diff
@@ -0,0 +1,37 @@
...
+            }
+            if (current) {
+                found = true;
+                headings[i].className = "active";
+            } else {
+                headings[i].className = "";
+            }
+        }
+        timer = null;
+    }, 50);
```
ylemkimon: I wonder 50ms is too fast.

website/static/js/scrollspy.js
edemaine: I've added the query cache (though if we ever load pages via AJAX, this will break).  I don't think we need to cache the positions of the headers -- these should be effectively cached by the web browser.  This is also standard for scrollspy.

edemaine:
@ylemkimon Thanks for all the comments and pointers.  I've implemented them all, except for caching header positions.

website/static/js/scrollspy.js
```diff
@@ -0,0 +1,44 @@
+// Based in part on
+// https://github.com/makotot/scrollspy/blob/master/src/js/modules/scrollspy.js
```
ylemkimon: I think it is no longer based on it.

website/static/js/scrollspy.js
```diff
@@ -0,0 +1,44 @@
...
+(function () {
+    const OFFSET = 10;
+    let timer;
+    let headingsCache;
+    const findHeadings = () =>
+        document.querySelectorAll('.toc-headings > li > a');
```
ylemkimon: I think we can move caching logics into `findHeadings`, i.e.:
```js
let headings;
const findHeadings = () => {
    if (!headings) headings = document.querySelectorAll('.toc-headings > li > a');
    return headings;
};
```

website/static/js/scrollspy.js
```diff
@@ -0,0 +1,44 @@
...
+            // scrollspy uses
+            // document.documentElement.scrollTop || document.body.scrollTop;
+            for (let i = 0; i < headings.length; i++) {
+                // if !found and i is the last element, highlight the last
+                let current = !found;
+                if (!found && i < headings.length - 1) {
+                    const next = headings[i + 1].href.split('#')[1];
+                    const nextHeader = document.getElementById(next);
+                    const top = nextHeader.getBoundingClientRect().top;
+                    current = top > OFFSET;
```
ylemkimon: This should be `current = top > scrollTop + OFFSET;`

website/static/js/scrollspy.js
ylemkimon: This will break if `onScroll` is called before `DOMContentLoaded`.

website/static/js/scrollspy.js
edemaine: Maybe "inspired by"?  I could remove the link and just mention that ScrollSpy is a thing.  I also wonder whether this file should be called "scrollspy.js" or something else...

website/static/js/scrollspy.js
edemaine: It used to be `top = nextHeader.getBoundingClientRect().top + scrollTop` and then `current = top + scrollTop > scrollTop + OFFSET`.  The `scrollTop`s cancel.

website/static/js/scrollspy.js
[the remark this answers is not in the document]
edemaine: I don't think so.  In my experience, it takes multiple seconds before the DOM content is fully loaded, and someone could easily scroll during that time.  We don't want to accidentally cache at this time or we might miss some of the elements...

Alternatively, I could bind the `scroll` etc. events only after DOM content is fully loaded.  But I like that the current code does its best effort while the page is still loading, in the case that someone scrolls early.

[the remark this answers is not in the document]
edemaine: Oops, I see now that you were suggesting the cache could be read here, not written.  I'll do that.

website/static/js/scrollspy.js
```diff
@@ -0,0 +1,46 @@
...
+        document.querySelectorAll('.toc-headings > li > a');
+    const onScroll = () => {
+        if (timer) {  // throttle
+            clearTimeout(timer);
+        }
+        timer = setTimeout(() => {
+            timer = null;
+            let found = false;
+            let headings = findHeadings();
+            const scrollTop = window.pageYOffset;
```
ylemkimon: Then `scrollTop` seems to be not used, too.

website/siteConfig.js
ylemkimon: @edemaine Sorry, it turns out `baseUrl` is needed for scripts path. I'll open a PR.

Issue #1544, opened 2018-08-03
Release 0.10.0


kevinbarabash:
I was thinking with all of the changes to our docs that I should probably publish to gh-pages again.  Does anyone have any concerns about that?

edemaine:
Sounds good to me.  There are more changes to come, but lots of good stuff already, so seems like a good time.

ylemkimon:
It’d be best to get #1547 merged before.

edemaine:
I'm now of the opinion that we shouldn't publish to gh-pages except when releasing...

ylemkimon:
I've created a Version 0.11.0 milestone, consisting of mostly documentation tasks.

edemaine:
@ylemkimon That is a nice list of documentation issues, and I'll try to help on them. (Let's self-assign issues that we're working on.) But I wonder if they're all necessary for 0.11.0.  We have an important bugfix (`\neq`) and rather large documentation changes in the ready, and I'd rather we release soon (@kevinbarabash said this weekend) and update as needed.

I added the new issue in #1428 which seems important for releasing, though @kevinbarabash (as admin) could still use the current version which directly publishes to `gh-pages`.

The rest seem like "nice to haves" but shouldn't necessarily block a new release; they could easily be added e.g. in "bug-fix" releases.  Or they could move to a 0.12.0 milestone.  (I may have also misinterpreted the list; maybe you had in mind that this weekend's release is 0.10.0 and these are for the next version?)  What do you and others think?

ylemkimon:
@edemaine I was just listing issues that'd be nice to be resolved until the weekend (soft milestone, to-do list), as it's first time we're releasing new version with the website. They are not meant to block the release 😄

ylemkimon:
BTW, the version number should be `0.10.0`, as last release is `0.10.0-beta`. My mistake.

kevinbarabash:
I thought we were going to `0.11.0` b/c there was a breaking change that was going to be deployed.

ylemkimon:
@kevinbarabash Semantic Versioning states:
> A pre-release version indicates that the version is unstable and might not satisfy the intended compatibility requirements as denoted by its associated normal version. 

But as people may be using `^0.10.0-beta`, I think `0.11.0` makes sense.

kevinbarabash:
I didn't realize that.  Cool, let's go with 0.10.0 in that case.

ylemkimon:
[...]

PR #1558, opened 2018-08-07
Support mediawiki

ronkok:
Wikipedia and other mediawiki sites have their own system for generating images of LaTeX expressions. That system has accrued several aliases for LaTeX functions, resulting in a design that their own [math road map](https://www.mediawiki.org/wiki/Extension:Math/Roadmap) calls "somewhat chaotic".

Those additional functions are supported by the LaTeX package [texvc](https://www.ctan.org/pkg/texvc).

This PR adds to KaTeX all of the `texvc` functions that have not been [deprecated](https://en.wikipedia.org/wiki/Help:Displaying_a_formula#Deprecated-syntax). (except `\O`, which conflicts with a pre-existing function, and `\varcoppa`, for which we do not have font support.)

Below is visual confirmation that this PR renders the whole motley group.

mediawiki

### DIFF
--- a/docs/support_table.md
+++ b/docs/support_table.md
@@ -69,15 +69,15 @@ If you know the shape of a character, but not its name, [Detexify](http://detexi
 |\acute|$\acute e$|`\acute e`|
 |\AE|$\text{\AE}$|`\text{\AE}`|
 |\ae|$\text{\ae}$|`\text{\ae}`|
-|\alef| |In master|
-|\alefsym| |In master|
+|\alef|$\alef$||
+|\alefsym|$\alefsym$||
 |\aleph|$\aleph$||
 |{align}|<span style="color:firebrick;">Not supported</span>|see `{aligned}`|
 |{aligned}|$\begin{aligned}a&=b+c\\d+e&=f\end{aligned}$|`\begin{aligned}`<br>&nbsp;&nbsp;&nbsp;`a&=b+c \\`<br>&nbsp;&nbsp;&nbsp;`d+e&=f`<br>`\end{aligned}`|
 |{alignat}|<span style="color:firebrick;">Not supported</span>|see `{alignedat}`|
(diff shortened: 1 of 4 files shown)


ylemkimon:
As there is `texvc` package, I think this qualifies to be in the main KaTeX.

edemaine:
Do we really want the "(in master)" notes in the documentation? We already have features listed in the docs that are master only. (I'd also like to release soon.)

ronkok:
> Do we really want the "(in master)" notes in the documentation? 

No. But what we want even less is an error message, in red, saying that a function is not supported. 

I am writing "In master" on the assumption that the docs may be published to gh-pages at any moment. The documentation is in a shakedown period at present. After things settle down, it may be possible to delay publishing `supported.md` until after a release. In that scenario, we could write feature directly into the docs.

Is it possible to update some but not all of the documentation pages? If, during the period between releases, we want to update some part of the documentation, can we update everything except `supported.md` and `suppport_table.md`? That would fix the problem once we get past this shakedown period.

edemaine:
This is related to #1544.  Perhaps we should only publish at release times, for now.

ronkok:
> Perhaps we should only publish at release times, for now.

That would allow us to update the docs with new functions as they are written. But I worry that we may someday discover that something in the documentation is just wrong and needs to be fixed right away.

ylemkimon:
@edemaine @ronkok If we decide to publish every commit, we can use built CSS (#1520) and `in master` notes are not necessary. Documentation can be updated regardless of whether we publish every release or commit.

ronkok:
@ylemkimon Thank you for the clarification. That changes how I view things. I'll get started on changing "In master" to displaying the actual function.

We need to get #1520 merged before we publish the docs.

ronkok:
I've changed the Support Table, deleting "In master" and instead writing in code to display the function.

When I run `yarn start`, every one of them shows an error. I don't understand why. These functions should run okay with the CSS from the current release. I don't understand why PR #1520 would help them.

ylemkimon:
You probably should build the KaTeX again, by running yarn build at the root.

ronkok:
@ylemkimon Thank you. Your suggestion did the trick and I am very much reassured.

edemaine:
We can publish docs every release once #1521 is merged (soon), then publish every commit once #1520 gets merged (but that, I fear, might want to wait till we get a domain name...).

kevinbarabash:
Wowza... so many symbols!  I'll publishing 0.11.0 and new docs this weekend.

PR #1559, opened 2018-08-07
Flatten "styling", "text", and "array" ParseNodes.

### LINKED ISSUE #1492: Refactor ParseNode to be plain objects

marcianx:
Fixes #1492.

### DIFF
--- a/src/environments/array.js
+++ b/src/environments/array.js
@@ -16,38 +16,14 @@ import type Parser from "../Parser";
 import type {ParseNode, AnyParseNode} from "../parseNode";
 import type {StyleStr} from "../types";
 import type {HtmlBuilder, MathMLBuilder} from "../defineFunction";
-import type {Measurement} from "../units";
 
 // Data stored in the ParseNode associated with the environment.
-type AlignSpec = { type: "separator", separator: string } | {
+export type AlignSpec = { type: "separator", separator: string } | {
     type: "align",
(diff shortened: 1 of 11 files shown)


ylemkimon:
Thank you! 😄

src/environments/array.js
```diff
@@ -16,39 +16,15 @@ import type Parser from "../Parser";
...
 // Data stored in the ParseNode associated with the environment.
-type AlignSpec = { type: "separator", separator: string } | {
+export type AlignSpec = { type: "separator", separator: string } | {
     type: "align",
     align: string,
     pregap?: number,
     postgap?: number,
 };
 
-export type ArrayEnvNodeData = {|
```
kevinbarabash: So flattening the "array" `ParseNode` made all this disappear?  Cool!

marcianx: Not really. It's still in the second parameter of `parseArray()` (which is now destructured). It's the only place it was used.

PR #1563, opened 2018-08-08
Do not polyfill JSON.stringify

ylemkimon:
Removes redundant `JSON.stringify` polyfill.

Related: #1469, #1470

### DIFF
--- a/website/static/css/custom.css
+++ b/website/static/css/custom.css
@@ -1,13 +1,23 @@
-ul.toc-headings > li {
-    padding-bottom: 0px;   /* moved to li > a */
+.onPageNav ul li,
+.onPageNav ul ul li {
+    padding-bottom: 0px; /* moved to ul ul, ul a, ul ul a */
 }
 
-.toc-headings > li > a {
+.onPageNav ul ul {
+    padding-top: 1.5px;
+    padding-bottom: 6.5px;
+}
+
+.onPageNav ul a {
     display: block;
     padding: 4px;
 }
 
-.toc-headings > li > a.active {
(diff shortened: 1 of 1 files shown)


edemaine:
Why is `JSON.stringify` polyfilled at all? Is there a browser that doesn't support it?

ylemkimon:
@edemaine It is wrapped to support stringifying/parsing `Symbol` polyfill.

edemaine:
@ylemkimon Ah, I see.  Can we remove symbol support from Babel, given that we don't use it?  Or add a lint test that avoids this in the future?

ylemkimon:
@edemaine I think we can do that with Babel 7.

kevinbarabash:
It looks like `Symbol` is being used in generated code:
```
/******/ 	// define __esModule on exports
/******/ 	__webpack_require__.r = function(exports) {
/******/ 		if(typeof Symbol !== 'undefined' && Symbol.toStringTag) {
/******/ 			Object.defineProperty(exports, Symbol.toStringTag, { value: 'Module' });
/******/ 		}
/******/ 		Object.defineProperty(exports, '__esModule', { value: true });
/******/ 	};
```
I wonder if we should generate a separate build for ES6 modules using rollup.

ylemkimon:
@kevinbarabash It uses `typeof Symbol !== 'undefined'` to detect Symbol support, and it justs sets `toString()` method.

kevinbarabash:
Ah... so this wasn't triggering the polyfill of `Symbol` then?  Is guarding against it just to avoid accidentally using `Symbol` in our code in the future?

PR #1569, opened 2018-08-08
Partial list of KaTeX users

### LINKED ISSUE #305: "Built with KaTeX" page
I was thinking it might be cool to show off other projects that are built using KaTeX.  Three.js does this, imo, to great effect: http://threejs.org/.

I saw a cool project on HN this morning that uses KaTeX: 
- https://news.ycombinator.com/item?id=9863152 
- https://github.com/AndrewBelt/hack.chat

Does anyone know of other projects?  Is there a way to search for projects that us KaTeX?

edemaine:
Fix #305 (though work should be ongoing) by incorporating users from that issue.  Also render project names.  Screenshot:

image


website/siteConfig.js
```diff
@@ -14,12 +14,84 @@ const baseUrl = process.env.BASE_URL || (process.env.CONTEXT ? '/' : '/KaTeX/');
...
+        pinned: true,
+    },
+    {
+        caption: 'hack.chat',
+        image: 'https://hack.chat/apple-icon-180x180.png',
+        infoLink: 'https://hack.chat/',
+        pinned: true,
+    },
+    {
+        caption: 'Khan Academy',
```
ylemkimon: I think Khan Academy should be on a separate section or the first, as this project is maintained by them.

ylemkimon:
Many of logos are trademark and some may have a guideline, which may not allow using them in our site. I think we should search for logo guidelines and contact them if necessary.

Also image links can break someday. I think we need to use image hosting such as Imgur or commit them to the repo.

ylemkimon:
I wonder using logos here qualifies as a Fair use.

ylemkimon:
* Observable (https://beta.observablehq.com)
* Rocket.Chat (https://rocket.chat/)
* pandoc (https://pandoc.org/)

edemaine:
> Many of logos are trademark and some may have a guideline, which may not allow using them in our site. I wonder using them here qualifies as a Fair use. If not, I think we should search for logo guidelines and contact them if necessary.

Good point.  https://en.wikipedia.org/wiki/Wikipedia:Logos discusses this issue at length.  One relevant quote:

> Note that it is not necessary to seek formal permission from the owner in advance of using its logo, so long as the usage is [fair use](https://en.wikipedia.org/wiki/Fair_use), does not create any impression that the logo is associated with Wikipedia or endorses either Wikipedia or the article in which the logo appears, and does not create any reasonable grounds for complaint by the owner.

Fair use in the US is governed by four factors: whether the use is commercial (no for KaTeX) and/or educational (probably yes); the nature of the work (a logo); the amount of the work used (all, though often low-resolution); and the effect on the market value (I believe we're only increasing the linked product value by linking to them, and I doubt the logo itself would ever be sold).  There are no hard-and-fast rules, and I am not a lawyer, but my impression is that these uses would qualify (and Wikipedia's read seems to be the same).

Regarding the "association": I think it's clear that the linked logos are from the linked projects, not ours.  (And even more so with the text labels.)  So should be good here.

Regarding the "reasonable grounds": If we listed a project that doesn't actually use KaTeX, or if their use of KaTeX were somehow a secret, I could imagine a complaint.  Otherwise, I don't expect one, and I think we can reasonably include the images and take them down if we get a take-down notice.

I tried to link to official logo URLs whenever possible, and otherwise Twitter/Tumblr profile pictures, but it's true that any of them could move and we'd have to monitor that.  (We have the same issue with badges, though their APIs are usually better defined.)  I believe it's simpler copyright-wise if we link to remote pictures instead of including copies on our own repo/website (but the trademark issue is unchanged), so I'd personally prefer to keep these as-is.

Here are the logo guidelines I've been able to find:
* https://en.facebookbrand.com/assets/messenger-toolkit-2 seems fine, except it says not to use the word "Facebook".  I'll remove that.
* https://www.dropbox.com/branding mentions it's fine to use the logo to link to Dropbox, though it does say we should write them about "promotional" use, which this may qualify as.

Probably good to have @kevinbarabash weigh in here.

website/siteConfig.js
edemaine: OK, maybe KA first and the rest alphabetical (i.e. as is)?

edemaine:
I added many of the projects that @ylemkimon listed.

I'm not sure what to do with the ones that have plugins.  For now, I've listed ones that have an "official" plugin for KaTeX, but not the others.  I'm not sure I know the right definition of "official" though...

Pandoc doesn't seem to have a logo, though [it's been discussed](https://groups.google.com/forum/#!topic/pandoc-discuss/1bKIuyBnWaQ%5B1-25%5D).  Not sure how we want to list projects without a logo...  Maybe a second list at the bottom of projects without images?

New screenshot:

image

website/siteConfig.js
```diff
@@ -22,12 +22,126 @@ embed.register('stylesheet',
...
         pinned: true,
     },
+    {
+        caption: 'Gitter',
+        image: 'https://assets.gitlab-static.net/uploads/-/system/project/avatar/3601513/gitter_logo.png',
+        infoLink: 'https://gitter.im/',
+        pinned: true,
+    },
+    {
+        caption: 'Gatsby',
```
ylemkimon: `Gatsby` is alphabetically before `GitLab`.

website/siteConfig.js
```diff
@@ -22,12 +22,126 @@ embed.register('stylesheet',
...
+    },
+    {
+        caption: 'Spinning Numbers',
+        image: 'https://spinningnumbers.org/i/sn_logo2.svg',
+        infoLink: 'https://spinningnumbers.org/',
+        pinned: true,
+    },
+    {
+        caption: 'StackEdit',
+        image: 'https://cdn.rawgit.com/benweet/stackedit/master/static/landing/favicon.ico',
```
ylemkimon: High resolution logo seems to be available in https://github.com/benweet/stackedit/blob/master/src/assets/iconStackedit.svg and https://github.com/benweet/stackedit/blob/master/src/assets/favicon.png.

kevinbarabash:
Here's the correct logo for Khan Academy.  It's mainly a color change.

q1lgyhsq

edemaine:
@kevinbarabash Is there a public URL for that logo?

kevinbarabash:
@edemaine their resizing algorithm leaves much to be desired.  It should be okay 🤞 since we're scaling image down.  I'm sure we're hosting it some in fastly too... I'll try to track down the URL but the one you posted is fine for now.

ylemkimon:
- Gitter: https://gitlab.com/gitlab-org/gitter/webapp/issues/1966

I've created issues on where public issue tracker is available.

ylemkimon:
https://github.com/bokeh/bokeh/issues/8167#issuecomment-412944140
> Just FYI Bokeh does not use KaTex directly, at least not yet. We have published one example that shows how users can create a custom extension themselves, that integrates KaTeX output into a Bokeh plot. We'd like to make KaTeX output a fully supported buit-in feature, but I'm not sure when that might be able to happen.

edemaine:
@ylemkimon Can you ask for a linkable URL? Or is this goal futile and we should we add this one to the repo?

ylemkimon:
@edemaine It seems it's from the internal guideline. As we got permissions to use them, I think we can commit image files to the repo.

website/siteConfig.js
```diff
@@ -22,15 +22,123 @@ embed.register('stylesheet',
 
 /* List of projects/orgs using your project for the users page */
 const users = [
```
ylemkimon: The `pinned` property of users are not used.

website/siteConfig.js
```diff
@@ -22,15 +22,123 @@ embed.register('stylesheet',
...
+    },
+    {
+        caption: 'Gitter',
+        image: 'https://assets.gitlab-static.net/uploads/-/system/project/avatar/3601513/gitter_logo.png',
+        infoLink: 'https://gitter.im/',
+        pinned: true,
+    },
+    {
+        caption: 'Gradescope',
+        image: '../img/gradescope_logo.png',
```
ylemkimon: This should be `baseUrl + 'img/gradescope_logo.png'`

kevinbarabash:
Looks good except the Rocket.Chat logo URL is broken.

screen shot 2018-08-16 at 10 00 02 pm

website/siteConfig.js
edemaine: I've removed all `pinned` properties.

edemaine:
I implemented @ylemkimon's comments. @kevinbarabash That image loads fine for me, in local testing and on netlify.  Are you using Chrome or another browser?  It is an SVG image, but I thought that worked on all browsers...

image

ylemkimon:
Maybe it's because RawGit is recommending using a specific tag or commit hash in the CDN URL (not a branch).

edemaine:
@ylemkimon Good catch. I've switched to a specific commit.

mbourne:
You may wish to add https://www.intmath.com/
Here's a logo: https://www.intmath.com/intmath-logo.svg

kevinbarabash:
@mbourne PRs are welcome. 🙂

edemaine:
Added in #1635

PR #1575, opened 2018-08-09
Enable stylelint-config-standard, lint CSS files

ylemkimon:


website/static/static/index.css
```diff
@@ -63,17 +64,14 @@ b {
     background: rgba(255, 255, 255, 0.1);
 }
 
-.demo {
```
kevinbarabash: I thought for a second it detected an unused style, but then I see it merged styles.

PR #1577, opened 2018-08-09
Improve main page (index.html) speed

ylemkimon:
* Inline Font Awesome icons using SVG
  * Upgrade to Font Awesome 5.2.0, which provides SVG for icons
  * Instead of including whole webfont and stylesheet, inline SVG and CSS rules
* Preload KaTeX fonts
  * Only supported by few browsers (https://caniuse.com/#feat=link-rel-preload), which all support WOFF2 (https://caniuse.com/#feat=woff2)
  * https://developer.mozilla.org/en-US/docs/Web/HTML/Preloading_content

Both prevents FOUT and FOIT.

* Defer loading javascripts
  * Supported by most browsers (https://caniuse.com/#feat=script-defer)
  * https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script#Attributes

I'll add to documentation what I've learned here.


ylemkimon:
@kevinbarabash Is Khan Academy still using Google Analytics?

website/static/js/index.js
```diff
@@ -1,6 +1,6 @@
 /* eslint-disable no-var */
 /* global katex: false */
-window.startup = function() {
+document.addEventListener("DOMContentLoaded", function() {
```
ylemkimon: This is to support browsers that don't support `defer`, but I'm not sure it's necessary.

website/pages/index.html
```diff
@@ -1,4 +1,9 @@
 <!DOCTYPE html>
+<!-- This page uses following open source softwares:
```
edemaine: "softwares" should be "software" (collective noun)

website/static/js/index.js
```diff
@@ -31,4 +31,4 @@ window.startup = function() {
     });
 
     doDemo();
-};
+})();
```
edemaine: I take it the `<script defer`s are guaranteed to execute in order, and only after the DOM has loaded?  [Yep](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script).  Cool.

website/pages/index.html
```diff
@@ -75,9 +96,18 @@ <h1 class="logo tex" data-expr="\KaTeX">KaTeX</h1>
...
+          View on GitHub
+        </a>
```
edemaine: It's a shame these have to be repeated, but we are saving 76KB in the woff2 file load (and the roundtrips for the fontawesome CSS and font files).

ylemkimon: @edemaine They probably will be compressed on most servers.

edemaine: Oh, true, gzip should deal with this well.

By the way -- is there some HTML linter we ought to be using? (not for this PR, just reviewing HTML made me wonder)

ylemkimon: @edemaine Not that I can think of. There seems to be [htmlhint](https://www.npmjs.com/package/htmlhint), but I haven't used it.

kevinbarabash:
> @kevinbarabash Is Khan Academy still using Google Analytics?

@ylemkimon I believe we are.  Why do you ask?

ylemkimon:
@kevinbarabash I was wondering whether I could remove GA.

kevinbarabash:
I didn't even realize there was Google Analytics code on this.  I doubt anyone uses this data, but let me ask our data analytics team first.

PR #1581, opened 2018-08-09
Upload minified files to CircleCI build artifacts

ylemkimon:


.circleci/config.yml
```diff
@@ -75,10 +75,10 @@ jobs:
 
       - store_artifacts:
           path: dist/katex.js
-          destination: katex.js
+          destination: katex.min.js
```
edemaine: The sources and destinations have mismatching filenames.  Did you mean `dist/katex.min.js`?  Ditto below

ylemkimon: @edemaine Yes, thank you!

ylemkimon:
This is thousandth commit!

kevinbarabash:
I could've sworn it was in the 700s last time I looked. 😅

kevinbarabash:
@ylemkimon are we using these artifacts for anything at the moment or do you have future plans for them?

ylemkimon:
@kevinbarabash No, we're not using them for now, and I think it's highly unlikely we'll use them in the future, as we have Netlify builds. I've removed them in #1658.

PR #1583, opened 2018-08-10
[0] Cleanup documentation/website build process

ylemkimon:
* Export `commander` program if required as module
* Move `prestart` (cli.md generation) to build.js
* Change `npm run` to `yarn`, normalize script names


website/package.json
```diff
@@ -1,11 +1,11 @@
 {
   "scripts": {
-    "prestart": "cp ../docs/cli.md.template ../docs/cli.md && node ../cli.js markdown-help >> ../docs/cli.md",
+    "prestart": "node lib/build.js",
```
edemaine: Doesn't this run `build.js` from the `website` directory, in which case `../docs/cli.md` path is incorrect?  Perhaps we should `chdir` within `build.js` for safety in any case (like we recently added to `screenshotter`).

ylemkimon: @edemaine `docs` is at the same level of `website`, so `../docs/cli.md` is correct path. I'm not sure cases of running `build.js` directly.

website/package.json
```diff
@@ -2,10 +2,10 @@
   "scripts": {
     "prestart": "node lib/build.js",
     "start": "docusaurus-start",
-    "build": "yarn prestart && docusaurus-build",
-    "publish-gh-pages": "yarn prestart && docusaurus-publish",
+    "build": "node lib/build.js && docusaurus-build",
+    "publish-gh-pages": "node lib/build.js && docusaurus-publish",
     "write-translations": "docusaurus-write-translations",
-    "version": "yarn prestart && docusaurus-version",
+    "version": "node lib/build.js && docusaurus-version",
```
edemaine: Should we remove the "prestart" script then?

ylemkimon: @edemaine It is run before `start`, but since it's single command and to be consistent with others, we can move `node lib/build.js` to `start`. What do you think?

edemaine: I'd say either yes, move it into `start`, or just revert that last commit and have everyone call `yarn prestart`.  Slight reference for the latter.

ylemkimon: I've committed the request change in #1584. I'll commit the changes here, if it gets disapproved.

Issue #1525, opened 2018-07-29
Built docs should use relative paths


kevinbarabash:
There are a few places where we assume that the docs will live inside `/KaTeX`.  This works when deploying to `khan.github.io` b/c the site lives in `khan.github.io/KaTeX`, but this won't work when deploying to netlify since we want `index.html` to live at `katex.netlify.com` instead of `katex.netlify.com/KaTeX`.

edemaine:
Relatedly, I found it confusing that the docusaurus server run via `npm run start` would not respond to `http://localhost:3000/` (error message `Cannot GET /`), but rather required `http://localhost/KaTeX`.  I expected everything to live at the root level of the test server (which would require relative linking), but barring that, we should at least improve the error message.

ylemkimon:
All urls are generated using `baseUrl` in `siteConfig.js`:
https://github.com/Khan/KaTeX/blob/edede0867dde7250d2aab07bb54312fe256f86f5/website/siteConfig.js#L25

edemaine:
> Doesn't running npm start automatically open a browser with http://localhost:3000/KaTeX?

I'm running `npm start` via ssh into a Linux machine, and in that setting, it doesn't.  (Cool to hear that it does that when running locally though.)

edemaine:
https://github.com/facebook/Docusaurus/issues/448 seems to be @kevinbarabash's original issue.

ylemkimon:
One place where we have to use absolute paths is docs, e.g., in #1520:
> https://github.com/Khan/KaTeX/pull/1520/files#diff-4c5caa110a8eefbc2536f38bd144c8fcR5
```html
<link rel="stylesheet" href="/KaTeX/static/katex.min.css"/><!-- must use absolute path -->
```

I'm not sure how to make it work using the environment variable.

edemaine:
Why does that path have to be absolute? Does `../static/katex.min.css` not work?

ylemkimon:
@edemaine If versioning is enabled(#1521) or translations are enabled, the path may be nested one or two levels further.

ylemkimon:
I've opened #1526.

ylemkimon:
Closing as #1526 is merged and it's an upstream issue.

PR #1584, opened 2018-08-10
[1] Use local built CSS and fonts for master documentation

ylemkimon:
Depends on #1583. Succeeds #1520.

* 7ee44df: copies local built CSS and fonts before building for `master` documentation


release.sh
```diff
@@ -129,10 +129,10 @@ node update-sri.js "${VERSION}" README.md contrib/*/README.md dist/README.md \
 pushd website
 yarn run version "${VERSION}"
 
-# Restore docs to use local built CSS
-# for file in ../docs/*.md.bak; do
-#     mv -f "$file" "${file%.bak}"
-# done
+Restore docs to use local built CSS
```
edemaine: Missing `#`.

ylemkimon: @edemaine Fixed.

docs/support_table.md
```diff
@@ -2,7 +2,7 @@
 id: support_table
 title: Support Table
 ---
-<link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/katex@0.10.0-beta/dist/katex.min.css" integrity="sha384-9tPv11A+glH/on/wEu99NVwDPwkMQESOocs/ZGXPoIiLE8MU/qkqUcZ3zzL+6DuH" crossorigin="anonymous">
+{@stylesheet: katex.min.css}
```
edemaine: Where is this feature documented? It seems a bit weird, e.g., the output includes `<meta property="og:description" content="{@stylesheet: katex.min.css}">`.  (Maybe we can manually include a description in the YAML header?)

Is there a reason to do it this way vs. via a `<link>` tag?  Avoiding the absolute path issue?  Or does it embed the CSS in another way?  (I haven't actually figured out how the CSS gets in there...)

ylemkimon: @edemaine This is `remarkable-embed` plugin and is to fix absolute path issue (#1525 and #1520). The `<link>` tag is in `body > div.navPusher > div > div.container.mainContainer > div > div.post > article > div > span > p > link`.

edemaine: Got it; all makes sense now.  Can we fix the description thing?

ylemkimon: @edemaine I think we can move stylesheet after the first paragraph.

ylemkimon: That seems to fix the issue.

ylemkimon:
@edemaine Yes 😄

website/siteConfig.js
```diff
@@ -12,6 +12,14 @@
 // If on netlify, use '/'. Otherwise use '/KaTeX/'.
 const baseUrl = process.env.BASE_URL || (process.env.CONTEXT ? '/' : '/KaTeX/');
 
+// Plugin for Remarkable to inject variables
+const {Plugin: Embed} = require('remarkable-embed');
+const embed = new Embed();
```
kevinbarabash: So this replaces `program.command('markdown-help')`?

release.sh
```diff
@@ -116,9 +116,9 @@ sed -i.bak -E '/^\/dist\/$/d' .gitignore
 rm -f .gitignore.bak
 git add .gitignore dist/
 
-# Edit docs to use CSS from CDN (#1520)
-# grep -l '{@stylesheet: static/katex.min.css}' docs/*.md | xargs sed -i.bak \
-#     's|{@stylesheet: static/katex.min.css}|<link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/katex@./dist/katex.min.css" integrity="sha256-katex.min.css" crossorigin="anonymous"/>|'
+# Edit docs to use CSS from CDN
+grep -l '{@stylesheet: katex.min.css}' docs/*.md | xargs sed -i.bak \
+    's|{@stylesheet: katex.min.css}|<link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/katex@./dist/katex.min.css" integrity="sha384-katex.min.css" crossorigin="anonymous"/>|'
```
kevinbarabash: Does the plugin inject the actual SHA384 for `katex.min.css`?

website/siteConfig.js
```diff
@@ -12,6 +12,14 @@
...
+
+ // {@stylesheet: path}
+embed.register('stylesheet',
+    path => `<link rel="stylesheet" href="${baseUrl}static/${path}"/>`);
```
kevinbarabash: I was wondering where the `/static` went to but here it is. :)

website/siteConfig.js
ylemkimon: @kevinbarabash No, this is for the stylesheet inclusion.

release.sh
ylemkimon: @kevinbarabash The `update-sri.js` script updates the hash.

PR #1585, opened 2018-08-10
[2-1] Do not publish master documentation to GitHub Pages

ylemkimon:
Part of https://github.com/Khan/KaTeX/issues/1509#issuecomment-412148173.

* Do not publish `master` (next) documentation to GitHub Pages
  * Does not publish local built CSS and fonts

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,6 +17,7 @@ diff.png
 /test/screenshotter/unicode-fonts
 coverage/
 docs/cli.md
+docs.bak
 lib/core/metadata.js
 lib/core/MetadataBlog.js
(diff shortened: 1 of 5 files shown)


edemaine:
Just to clarify: you want to avoid publishing master/next docs because we're not yet making the built master CSS/JS/fonts available, right? (and then 2-2 fixes this?)

edemaine:
(I believe this means we should get to this PR before next release.)

ylemkimon:
@edemaine Please see https://github.com/Khan/KaTeX/issues/1509#issuecomment-412148173 😄

PR #1591, opened 2018-08-11
Remove Object.assign() in auto-render which requires expensive polyfill, update LICENSE

ylemkimon:
* Remove Object.assign() in auto-render which requires expensive polyfill
  * This reduces auto-render size from 26 KB (9 KB minified) to 12 KB (5 KB minified).
* Update LICENSE
  * `_.extend` of `underscore.js` was used in https://github.com/Khan/KaTeX/commit/cd9bca4a89eddd77c542e23a4a9dc2deff2ab72c#diff-61e0bdf7e1b43c5c93d9488b22e04170, but later removed in #656, but LICENSE was not updated
  * Renamed LICENSE.txt to LICENSE, which seems to be more common
  * Updated copyright year, the initial commit (https://github.com/Khan/KaTeX/commit/708a36502bea3b289fa4eac46c93478a5af3eca3) was 2013

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -71,7 +71,7 @@
         "require-jsdoc": 0
     },
     "overrides": [{
-        "files": ["katex.js", "src/**/*.js"],
+        "files": ["katex.js", "src/**/*.js", "contrib/**/*.js"],
         "excludedFiles": "unicodeMake.js",
         "rules": {
             "no-restricted-syntax": [2, "ForOfStatement", "ClassDeclaration[superClass]", "ClassExpression[superClass]"]
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,6 @@
 The MIT License (MIT)
 
-Copyright (c) 2015 Khan Academy
+Copyright (c) 2013-2018 Khan Academy
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
 of this software and associated documentation files (the "Software"), to deal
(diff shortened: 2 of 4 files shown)


LICENSE
```diff
@@ -1,12 +1,6 @@
 The MIT License (MIT)
 
-Copyright (c) 2015 Khan Academy
-
-This software also uses portions of the underscore.js project, which is
-MIT licensed with the following copyright:
-
-Copyright (c) 2009-2015 Jeremy Ashkenas, DocumentCloud and Investigative
-Reporters & Editors
+Copyright (c) 2013-2018 Khan Academy
```
edemaine: This looks right to me, but perhaps someone from KA can confirm.

Also, is it best practice to call the file `LICENSE` or `LICENSE.txt`? I thought the latter but I don't know.

.eslintrc
```diff
@@ -71,7 +71,7 @@
         "require-jsdoc": 0
     },
     "overrides": [{
-        "files": ["katex.js", "src/**/*.js"],
+        "files": ["katex.js", "src/**/*.js", "contrib/**/*.js"],
```
edemaine: Nice addition. I'm now happy with the polyfill part of this PR.

LICENSE
ylemkimon: @edemaine Most JavaScript library repos I know use `LICENSE` and GitHub initializes the repo with `LICENSE`.

Issue #1593, opened 2018-08-11
Fraction bars disappear when printing webpages into PDFs.


riteme:
Hello, I'm using *KaTeX 0.10.0-beta* and a fresh installation of Google Chrome Beta 69.0.3497.23 on Ubuntu 18.04 (to make sure that all preferences are defaults). I'm considering printing webpages with math formulas into PDFs. But I came across some issues with fractions.

For example:

```html
<!DOCTYPE html>
<html>
<head>
  <link href="katex/katex.min.css" type="text/css" rel="stylesheet">
  <script src="katex/katex.min.js" type="text/javascript"></script>
  <script src="katex/contrib/auto-render.min.js" type="text/javascript"></script>
  <script>
    document.addEventListener("DOMContentLoaded", function() {
      renderMathInElement(document.body, options = {delimiters: [{ left: "$$", right: "$$", display: true }, { left: "$", right: "$", display: false }]});
    });
  </script>
</head>
<body>
<p>inline math: ${2 \over n}$.</p>
<p>display math:
$$
f^{(n)}(x) = \begin{cases}
x + nb &amp; (a = 1) \\
a^n\ x + {a^n - 1 \over a - 1} \cdot b &amp; (a &gt; 1)
\end{cases}
$$
</p>
</body>
</html>
```

Print it via Chrome and will get:

![](https://gitee.com/riteme/blogimg/raw/master/other/katex-print-1.png)

Press "F12" to add the following CSS code:

```css
.katex {
    font-size: 1em;  /* or even smaller */
}
```

Reprint and get:

![](https://gitee.com/riteme/blogimg/raw/master/other/katex-print-2.png)

I have also tried <https://khan.github.io/KaTeX/> and set `font-size` to `0.3em` and came across the same issue.  I wonder why printed PDFs would be different from original webpages. Is that a bug in Chrome or KaTeX?

riteme:
Fraction bars begin to show after the CSS added. However it seems that horizontal lines are too wide:

![](https://gitee.com/riteme/blogimg/raw/master/other/katex-print-4.png)

Is that a bug with Chrome? Since the underline of "display math" has the same width.

ronkok:
> Is that a bug with Chrome?

We've had no reports that I recall of this type from Firefox or Edge. Safari has similar problems.

>  it seems that horizontal lines are too wide.

Try `0.5px`. Let me know how this turns out. I'll write a PR that adds the best results to KaTeX.

[...]

PR #1594, opened 2018-08-11
Add min-height CSS to print

ronkok:
In release 0.10.0-alpha, we reverted frac-lines to span borders and applied `min-height: 1px` CSS. But that CSS was applied only to screens, not to print. 

This PR removes the media query so that the same `min-height` is applied to both screen and print.

Addresses issue #1593.


ronkok:
I'm going to withdraw this PR for now. I think we need more information before proceeding. Can the problem reported in issue #1593 be avoided by just using a minimum 10 pt font size? Can it avoided by using Firefox instead of Chrome? Is a `1px` minimum line thickness a net loss for printed material?

Until we know the answers to those questions, we probably should not merge this PR.

edemaine:
While I wish Chrome didn't do any rounding so we could be more TeX-accurate, I think it's reasonable for the printed form to match the screen form in this case, given that the rounding bug sometimes occurs in printed form.

src/katex.less
```diff
@@ -411,6 +404,7 @@
             display: inline-block;
             margin: 0 -0.025em;
             border-right: 0.05em solid;
+            min-width: 1px;
```
edemaine: Now that this line is far away from the rest, maybe it needs a comment too?

ronkok: Done.

PR #1595, opened 2018-08-11
Upgrade to Babel 7

ylemkimon:
Babel 7 RC has been released and the final version is expected in 1-2 weeks.

Blocked on Babel 7 release.


babel.config.js
```diff
@@ -0,0 +1,35 @@
+module.exports = api => {
+    const nodeEnv = api.cache(() => {
+        switch (process.env.NODE_ENV) {
```
ylemkimon: TODO: `rollup.config.js` sets the `BABEL_ENV`, not `NODE_ENV`.

kevinbarabash:
7.0 is out. 🎉

ylemkimon:
Updated!

package.json
```diff
@@ -16,17 +16,19 @@
...
-    "babel-core": "^6.26.3",
-    "babel-eslint": "^8.1.2",
+    "@babel/core": "^7.0.0",
+    "@babel/plugin-proposal-class-properties": "^7.0.0",
+    "@babel/plugin-transform-runtime": "^7.0.0",
+    "@babel/preset-env": "^7.0.0",
+    "@babel/preset-flow": "^7.0.0",
+    "@babel/register": "^7.0.0",
+    "@babel/runtime": "^7.0.0",
+    "babel-core": "^7.0.0-bridge.0",
```
kevinbarabash: Thanks for figuring out how to run jest after upgrading to babel 7.

rollup.config.js
```diff
@@ -1,6 +1,6 @@
 import babel from 'rollup-plugin-babel';
 
-process.env.BABEL_ENV = 'esm';
+process.env.NODE_ENV = 'esm';
```
kevinbarabash: I didn't realize that node scripts could also set environment variables, cool!

src/functions/delimsizing.js
```diff
@@ -64,8 +64,8 @@ function checkDelimiter(
     } else {
         throw new ParseError(
             "Invalid delimiter: '" +
-            // $FlowFixMe, do not polyfill
-            (symDelim ? symDelim.text : JSON["stringify"](delim)) +
+            // eslint-disable-next-line transform-runtime-aliasing/no-transform-runtime-aliasing
+            (symDelim ? symDelim.text : JSON.stringify(delim)) +
```
kevinbarabash: The reason why babel doesn't polypill this anymore is b/c we're using browserlist right?

src/macros.js
```diff
@@ -215,6 +215,7 @@ const def = (context, global: boolean) => {
             throw new ParseError(`Invalid argument number "${arg[0].text}"`);
         }
         numArgs++;
+        // eslint-disable-next-line transform-runtime-aliasing/no-transform-runtime-aliasing
```
kevinbarabash: Adding this wherever we have a `parseInt` or `parseFloat` is kind of a pain.  Do we still need this lint rule now we're using `browserlist`?

src/functions/delimsizing.js
ylemkimon: (The "polyfill" or "polyfilling" here refers to `core-js aliasing`)

@kevinbarabash Unfortunately, `@babel/plugin-transform-runtime` doesn't use Browserslist (yet), so it includes every polyfill even if all target environments support it or it isn't needed. So I think it's better to whitelist polyfills than blacklist them.

> After careful consideration:
> - Polyfills are generally expensive
> - ES modules of core-js@2 is not available, it needs to be wrapped in ES module to use in the ESM build (rollup)
> - We have removed codes that require polyfill in #1469, #1470, #1563, and #1591 
> - We added a warning (linter) against code that requires polyfill in #1597
> - Babel 7 separated out Babel's helpers from its polyfilling behavior in runtime, `transform-runtime` no longer polyfills by default

ylemkimon: Therefore, I've disabled automatic core-js aliasing, which is default in Babel 7, and changed to warn that it is not polyfilled and adding polyfill may be needed.

src/macros.js
ylemkimon: @kevinbarabash I've added an `ignore` option to `no-transform-runtime-aliasing` and ignored `JSON.stringify`, `parseInt`, and `parseFloat`. which clearly don't need polyfill.

babel.config.js
```diff
@@ -0,0 +1,25 @@
+module.exports = api => {
```
kevinbarabash: I assume this callback format for `babel.config.js` is new in v7.  I'm sure there are lots of cool things a person could do with this.

Issue #1495, opened 2018-07-22
Remove support for TTF fonts


ylemkimon:
Most browsers including IE 9 support WOFF font format. (https://caniuse.com/#feat=woff) The only concern is `< Android 4.4` but its usage is about 0.2%.

EOT fonts have been removed for similar reasons: https://github.com/Khan/KaTeX/issues/893.

Does anyone have any concerns about removing this font format?

edemaine:
WOFF seems OK for the web, but are there reasons we might want to install the KaTeX fonts at the operating system level?  It seems plausible.  And I think that means we'd need TTF or OTF (which we don't currently have).

ylemkimon:
@edemaine `TTF` and `OTF` would be still generated as `woff` and `woff2` are generated from them. I was thinking removing them from KaTeX distribution and CSS.

edemaine:
@ylemkimon Ah, I see. That seems more reasonable: still include them in the fonts repository, but don't link to them in the CSS.  I can't think of a reason not to do this, though it'd be helpful to have other opinions on this.  (What I've read on Google seems to agree with the idea.)  I also can't think of a major advantage, except slightly shorter CSS, but it does seem cleaner.

ronkok:
If understand issue #557 correctly, then this PR would be a benefit. An unusual case, but still a win.

ylemkimon:
I originally thought we can get rid of whole TTF build process, but it turns out we can't. This would be as simple as just setting following `false`:
https://github.com/KaTeX/katex-fonts/blob/a09ea0bbe0c6a7dd18118207449654f929388eba/fonts.less#L2

SmashManiac:
Unless someone can come up with a useful and complete use case for TTF fonts over WOFF fonts in KaTeX, I believe keeping both of them would be technical debt.

PR #1600, opened 2018-08-12
Enable environment variable USE_TTF to disable bundling TTF fonts

ylemkimon:
This can be used to reduce bloating CircleCI build artifacts and GitHub Pages repo if we commit local built files. Also, later default of `use-ttf` can be changed to `false`, and use this to custom-build with TTF fonts. (#1495)


ylemkimon:
@kevinbarabash http://lesscss.org/usage/#less-options-modify-variables

PR #1603, opened 2018-08-12
Update release script to use pull request

### LINKED ISSUE #1428: Improve release script

ylemkimon:
Fixes #1428

New release process consists of two steps:

* Release (`./release.sh <VERSION_TO_RELEASE> [NEXT_VERSION]`)
  * Make a release branch(`v<VERSION_TO_RELEASE>-release`)
  * Build generated files
  * Edit package.json to the next version
  * Edit docs to use CSS from CDN
  * Update the version number in the README and the documentation, and regenerate the SRI hash
  * Generate a new version of the docs
  * Make a commit and push
  * **[manual]** Open a pull request

If KaTeX codes get changed during PR review, update SRI hashes by `./release.sh <VERSION_TO_RELEASE>` on the generated release branch.

After the PR gets merged:
* Publish (`./release.sh -p <VERSION_TO_RELEASE>`)
  * Make a new detached HEAD
  * Edit package.json to the release version
  * Build generated files
  * Make the commit and tag, and push them
  * Publish to NPM and GitHub pages
  * **[manual]** Create a GitHub release


release.sh
```diff
@@ -98,95 +109,119 @@ if [[ "$CONFIRM" != "y" ]]; then
...
-# Restore docs to use local built CSS
-for file in ../docs/*.md.bak; do
-    mv -f "$file" "${file%.bak}"
-done
 
-USE_SSH=true yarn publish-gh-pages
-popd
+if [[ $BRANCH != "master" ]]; then
+    # Build generated files
+    yarn build
```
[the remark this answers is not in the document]
ylemkimon: @edemaine `set -e -o pipefail` is set and the script will fail if any command returns non-zero.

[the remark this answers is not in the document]
ylemkimon: @edemaine Arguments of `node update-sri.js` and `git add` differ from when running on the master branch (for the first time).

ylemkimon:
@edemaine 
> One policy question I wonder: should we delete the release branch afterward?  I notice some projects (e.g. https://github.com/meteor/meteor) keep `release-*` branches, but this may also get to be annoyingly many.

As we do not separate master branch and development branch, and do not have that many commits to do so, I don't think it's necessary.

release.sh
```diff
@@ -79,12 +86,16 @@ if ! command git diff --stat --exit-code HEAD; then
     echo "Please make sure you have no uncommitted changes" >&2
     : $((++INSANE))
 fi
-if [[ $BRANCH != @(v*|master) ]]; then
+if [[ $BRANCH != @(v*-release|master) ]]; then
```
kevinbarabash: Limiting which branches can be released from makes sense.

release.sh
```diff
@@ -98,95 +109,119 @@ if [[ "$CONFIRM" != "y" ]]; then
...
+    # Build generated files
+    yarn build
+
+    # Regenerate the Subresource Integrity hash in the README and the documentation
+    node update-sri.js "${VERSION}" README.md contrib/*/README.md \
+        docs/*.md website/pages/index.html website/versioned_docs/version-$VERSION/*.md
+
+    # Make the commit and push
+    git add README.md contrib/*/README.md \
+        docs website/pages/index.html website/versioned_docs/
```
kevinbarabash: I wonder if it's worth checking for any unstaged changes or new files after this command.

release.sh
```diff
@@ -98,95 +109,119 @@ if [[ "$CONFIRM" != "y" ]]; then
...
+    if [ ! -z "$NEXT_VERSION" ]; then
+        # Edit package.json to the next version
+        sed -i.bak -E 's|"version": "[^"]+",|"version": "'$NEXT_VERSION'-pre",|' package.json
+        rm -f package.json.bak
+        git add package.json
+    fi
+
+    # Edit docs to use CSS from CDN
+    grep -l '{@stylesheet: katex.min.css}' docs/*.md | xargs sed -i.bak \
+        's|{@stylesheet: katex.min.css}|<link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/katex@./dist/katex.min.css" integrity="sha384-katex.min.css" crossorigin="anonymous"/>|'
```
kevinbarabash: At this point these files aren't on the CDN yet, but if the PR is accepted, then the publish will occur and they will be on the CDN, right?

release.sh
```diff
@@ -98,95 +109,119 @@ if [[ "$CONFIRM" != "y" ]]; then
...
+        docs/*.md docs/*.md.bak website/pages/index.html
+
+    # Generate a new version of the docs
+    pushd website
+    yarn run version "${VERSION}"
+    popd
+
+    # Restore docs to use local built CSS
+    for file in docs/*.md.bak; do
+        mv -f "$file" "${file%.bak}"
```
kevinbarabash: `%.bak` means "without `.bak`"?

release.sh
```diff
@@ -98,95 +109,119 @@ if [[ "$CONFIRM" != "y" ]]; then
...
+    # Make the commit and push
+    git add package.json README.md contrib/*/README.md \
+        docs website/pages/index.html website/versioned_docs/ \
+        website/versioned_sidebars/ website/versions.json
+    if [[ -z "$NEXT_VERSION" ]]; then
+        git commit -n -m "Release v$VERSION"
+    else
+        git commit -n -m "Release v$VERSION" -m "Bump $BRANCH to v$NEXT_VERSION-pre"
+    fi
+    git diff --stat --exit-code # check for uncommitted changes
```
kevinbarabash: :+1:

release.sh
```diff
@@ -98,95 +109,119 @@ if [[ "$CONFIRM" != "y" ]]; then
...
+    sed -i.bak -E 's|"version": "[^"]+",|"version": "'$VERSION'",|' package.json
     rm -f package.json.bak
-    git add package.json
-fi
 
-# Refer to the just-released version in the documentation of the
-# development branch, too.  Most people will read docs on master.
-git checkout "v${VERSION}" -- README.md contrib/*/README.md docs/*.md website/
+    # Build generated files and add them to the repository
+    git clean -fdx dist
```
kevinbarabash: Why not `yarn clean`?

update-sri.js
```diff
@@ -20,12 +22,8 @@ Promise.all(process.argv.slice(3).map(file =>
...
     .then(body => {
         // Replace size badge url
-        // 1 - url prefix: https://img.badgesize.io/Khan/KaTeX/
-        // 2 - url suffix: /dist/katex.min.js?compression=gzip 
-        const badgeRe = /(https:\/\/img\.badgesize\.io\/Khan\/KaTeX\/v)(?:.+)(\/dist\/katex\.min\.js\?compression=gzip)/g;
-        body = body.replace(badgeRe, (m, pre, post) => {
-            return pre + version + post;
-        });
+        // eslint-disable-next-line max-len
+        body = body.replace(/(https:\/\/img\.badgesize\.io\/Khan\/KaTeX\/v)(?:.+)(\/dist\/katex\.min\.js\?compression=gzip)/g, `$1${version}$2`);
```
kevinbarabash: I didn't realize that you could use `$1`, `$2`, etc. for the capture groups in the replacement string, cool!

update-sri.js
```diff
@@ -35,24 +33,37 @@ Promise.all(process.argv.slice(3).map(file =>
         // 5 - integrity opening quote: "
         // 6 - old hash: sha384-…
         // 7 - integrity hash algorithm: sha384
-        const cdnRe = /((["'])https?:\/\/cdn\.jsdelivr\.net\/npm\/katex@)[^\/"']+(\/([^"']+)\2(?:\s+integrity=(["'])(([^-]+)-[^"']+)\5)?)/g;
+        // eslint-disable-next-line max-len
+        const cdnRe = /((["'])https?:\/\/cdn\.jsdelivr\.net\/npm\/katex@)[^/"']+(\/([^"']+)\2(?:\s+integrity=(["'])(([^-]+)-[^"']+)\5)?)/g;
```
kevinbarabash: I wish GitHub had better diffing.  This looks exactly the same.

update-sri.js
```diff
@@ -35,24 +33,37 @@ Promise.all(process.argv.slice(3).map(file =>
...
+                if (version === "check" && hash !== newHash) {
+                    throw new Error("SRI mismatch! " +
+                        "Please run the release script again.");
+                }
             })
-        )).then(() => write(file, body));
+        ));
+
+        if (version !== "check") {
+            promise = promise.then(() => write(file, body));
```
kevinbarabash: Minor nit: this could just return `promise.then(...)` here then `promise` could be a constant.

release.sh
```diff
@@ -98,95 +109,119 @@ if [[ "$CONFIRM" != "y" ]]; then
...
 
-# Go back to original branch to bump
-git checkout "$BRANCH"
+    echo ""
+    echo "The automatic parts are done!"
+    echo "Now create a pull request against master from 'v$VERSION-release'"
+    echo "Visit https://github.com/Khan/KaTeX/pulls to open a pull request."
+    echo "After it gets merged, run './release.sh -p $VERSION'!"
+    echo "Note that if KaTeX source code is changed after running this script,"
+    echo "you have to run the release script again."
```
kevinbarabash: If you run `./release.sh 0.6.3 0.6.4` will it push a new commit to the `v0.6.3-release` branch created by the first run?

release.sh
ylemkimon: @kevinbarabash Yes.

release.sh
ylemkimon: @kevinbarabash Yes, it's bash substitution.

release.sh
ylemkimon: @kevinbarabash Yes, and the next version is not used for the second run, so it can be omitted.

release.sh
ylemkimon: @kevinbarabash `yarn clean` removes the `node_modules`, and actually `yarn dist` includes cleaning the `dist`. But as it's the Git that manages files, I assume it's for creating a pristine working tree.

update-sri.js
ylemkimon: @kevinbarabash The backslash escaping the slash in the character group, `[^\/"']`, is removed as it's not necessary.

release.sh
ylemkimon: @kevinbarabash I move the check to the top level.

release.sh
kevinbarabash: We also have a `clean` script in package.json which I guess would have to be run using `yarn run clean`.  It just deletes `dist` though.  I think these accomplish the same thing so either is fine.

ylemkimon: @kevinbarabash Counterintuitive, but our `clean` script deletes the `node_modules`:
https://github.com/Khan/KaTeX/blob/46e7939fa921128d7c11f15b24541c3501e7d006/package.json#L83

PR #1606, opened 2018-08-13
Rename .value to .text or .color where applicable.

marcianx:
Post #1492 cleanup.
This is toward making field names more specific to make it easier to follow. Currently, `.value` is used to mean too many different things. So this part of a sequence of renames to make the names more specific.


kevinbarabash:
This is great! I wonder if there will be some point at which we'll be happy enough with the AST that we could publish and provide an API for third party code to provide the an AST which it generated for us to render. 🤔

marcianx:
@kevinbarabash I'm always hesitant to make such a commitment, especially as KaTeX still good to evolve to support, e.g., SVG output. But perhaps that wouldn't requires a change in the tree.

Before stabilizing the tree API, though, I would strongly recommend defining and documenting standard structural patterns (e.g. `text` fields would always be strings or some such) and updating the tree to conform to these strong patterns.

PR #1610, opened 2018-08-13
Improve prestart and tests

### LINKED ISSUE #1166: Don't run tests if only documentation changes have been made
It takes a while for tests to run and it would be nice if we could avoid running tests when we don't need to.

### LINKED ISSUE #1589: Consider using `yarn check` instead of `check-dependencies`
> [`yarn check`] [v]erifies that versions of the package dependencies in the current project’s package.json match those in yarn’s lock file.

`yarn check` is slower than `check-dependencies`, but it takes the lockfile into account and no dependency is required.

ylemkimon:
* Screenshotter
  * replace promisify(fs) and mkdirp with fs-extra
  * replace selenium.promise with builtin Promise
* unicodeMake
  * replace console.log with writeFile(string)
  * check timestamp, do not build if up-to-date
  * exclude from coverage
* update-sri
  * replace promisify(fs) with fs-extra
* Replace check-dependencies with yarn check: fixes #1589 
* Lint all JavaScript files in the root
* Skip CircleCI if only unrelated files have been changed: fixes #1166 
* Add `babel-core` dependency: fixes unmet peer dependency warning


kevinbarabash:
I think when I reviewed this I was only looking a one commit instead of all the changes.  I'm sure it's fine but I should probably do another pass even though this is already merged.

ylemkimon:
@kevinbarabash Sorry, there was an error in bash variable, see #1629.

PR #1615, opened 2018-08-14
Improve demo (REPL, sandbox) on the main page

### LINKED ISSUE #1244: Sandbox/Demo/REPL for testing on local and the website
Currently, KaTeX has a demo at [GitHub pages](https://khan.github.io/KaTeX/), which provides very basic demo, and a local test page on `npm run`, where some options are configurable using URL parameters.

It'd be useful to have KaTeX sandbox/demo for _both_ local and the website with:
- GUI to configure KaTeX (#1225, #1193)
- ability to run previous versions or test a PR. (#1187, #1171, #1235)

There are many existing similar services and it'd be nice to adopt existing framework, but I can't seem to find a suitable one:
- [CodePen](https://codepen.io/pen/)
- [JSFiddle](https://jsfiddle.net/)
- [Babel REPL](http://babeljs.io/repl)
- [CodeSandbox](https://codesandbox.io/)
- [Ace Kitchen Sink](https://ace.c9.io/build/kitchen-sink.html)

ylemkimon:
Fixes #1244.

Instead of making new page for the demo (REPL, sandbox), I think we can extend the main page demo.

I didn't include `maxSize`, `maxExpand`, and `allowedProtocols` because I think there is no use case of them.

Not supported on small screens (yet).

I'm really bad at UI, so any suggestion or modification are welcome!

image


ylemkimon:
@kevinbarabash 

> changing the values in the options panel should trigger a re-render

It listens to the `input` event, but it seems quite a few of browsers don't fire it on the checkbox: https://caniuse.com/#search=input. I'll change it to `change`.

> in maximized view we may want to stack the editor/render views vertically especially as the screen gets narrower

I agree.

kevinbarabash:
Safari unfortunately doesn't support `type="color"` on `<input>`.  That's okay, but it would be nice to constrain its width a bit.  I found that setting `max-width: 5em;` looked alright and didn't affect the color picker on Chrome.

screen shot 2018-08-16 at 10 06 24 pm

website/static/static/index.css
```diff
@@ -253,3 +320,21 @@ b {
...
+
+@keyframes slide-left {
+    0% { transform: translateX(0%); }
+    100% { transform: translateX(-110%); }
+}
+
+@-webkit-keyframes slide-left {
+    0% { -webkit-transform: translateX(0%); }
+    100% { -webkit-transform: translateX(-110%); }
+}
```
kevinbarabash: I wonder if we need `-moz-` versions too.  I checked Firefox and it works as is, but my FF is pretty recent.

website/static/static/index.css
```diff
@@ -74,10 +85,21 @@ b {
     margin: 50px 0 0;
 }
 
+.maximized .demo {
+    flex-direction: column;
```
kevinbarabash: Do we need `-ms-flex-direction` here as well.  We may want to consider adopting a CSS preprocessor to help with this in the future.

website/static/js/index.js
```diff
@@ -8,12 +9,72 @@
...
+            // eslint-disable-next-line no-console
+            console.warn(e);
+        }
+    }
+
+    if (data.code) {
+        demoInput.value = data.code;
+    }
+
+    var katexOptions = ["displayMode", "throwOnError", "errorColor", "strict",
```
kevinbarabash: One thing I noticed when toggling `displayMode` is that it takes a bit of time for the font with the smaller integral to load.  We may want to pre-load fonts for the demo in the future.

website/static/js/index.js
```diff
@@ -8,12 +9,72 @@
...
+        "macros"].map(function(id) {
```
kevinbarabash: I wasn't sure what format was need for the macros, it'd be good to have a couple of pre-populate macros... maybe one that takes an argument.

website/static/static/index.css
ylemkimon: @kevinbarabash FireFox >= 16 supports without the prefix: https://caniuse.com/#feat=css-animation

ylemkimon:
@kevinbarabash Done.

edemaine:
Just got to try this.  Wow, very cool!  A few questions for possible follow-up:
* It's cool to demo the macros feature, though I wonder if we should be encouraging people to use that, or instead show use of `\def`/`\newcommand`, as most KaTeX end-users won't be able to modify `options`?
* Should we add a "Copy HTML code" button (output of `renderToString`)?
* Also unclear to me whether `Copy KaTeX code` should output `render` (as currently) or `renderToString`.  But this is fine.

PR #1621, opened 2018-08-15
Fix scrollspy.js & use query-string@5 to support IE

ylemkimon:


ylemkimon:
@edemaine Thank you for reviews!

PR #1624, opened 2018-08-16
Generate ECMAScript module for contrib

ylemkimon:


contrib/copy-tex/copy-tex.webpack.js
```diff
@@ -0,0 +1,6 @@
+/**
+ * This is the webpack entry point for KaTeX. As ECMAScript doesn't support
+ * CSS modules natively, a separate entry point is used.
+ */
+import './copy-tex.css';
+import './copy-tex.js';
```
kevinbarabash: This change doesn't seem strictly necessary.  Was thine thinking to make it follow the same pattern as the main katex build?

rollup.config.js
```diff
@@ -1,14 +1,21 @@
...
+    input: entry.replace('.webpack', ''),
     output: {
-        file: 'dist/katex.mjs',
+        file: `dist/${name}.mjs`,
         format: 'es',
     },
     plugins: [
         babel({runtimeHelpers: true}),
+        alias({
+            katex: '../katex.mjs',
```
kevinbarabash: Does this require `katex` to appear first in `targets`?  Also, where is this used?

webpack.common.js
```diff
@@ -27,7 +27,7 @@ const targets /*: Array<Target> */ = [
     },
     {
         name: 'contrib/copy-tex',
-        entry: './contrib/copy-tex/copy-tex.js',
+        entry: './contrib/copy-tex/copy-tex.webpack.js',
```
kevinbarabash: If we're using `rollup.config.js` to build all of these targets, why is `copy-tex` the only one that needed to be changed to use `.webpack.js`?

rollup.config.js
ylemkimon: @kevinbarabash It transforms `import katex from ‘katex’` to `import katex from ‘../katex.mjs’` as browsers cannot resolve a module by its name by spec.

webpack.common.js
ylemkimon: @kevinbarabash `copy-tex` imports a CSS module that ESM doesn’t support. So we use separate entry point for webpack.

rollup.config.js
kevinbarabash: After I asked the question I ended up having to so something similar for another project.

PR #1629, opened 2018-08-17
Fix line break and variable expansion in CHANGED variable

ylemkimon:


.circleci/config.yml
```diff
@@ -13,8 +13,8 @@ post_checkout: &post_checkout
         fi
         if [[ $BASE_COMMIT ]]; then
           CHANGED=$(git diff --name-only $BASE_COMMIT)
-          echo $CHANGED
-          echo 'export CHANGED="$CHANGED"' >> $BASH_ENV
```
kevinbarabash: Does bash do something different with single quotes?

ccorn: Shells, Perl, and Ruby allow almost no escapes and no interpolation in single-quoted strings. (In contrast to e.g. Python and JavaScript which treat single and double-quotes similarly.)

Issue #1618, opened 2018-08-14
Remove documentFragment or separate it from HtmlDomNode


ylemkimon:
`documentFragment` causes many problems because they differ from normal `HtmlDomNode`, but it implements `HtmlDomNode`.

https://github.com/Khan/KaTeX/pull/1471#issuecomment-404193446 by @edemaine 
> This seems fine, and simpler to me, but I'll reraise a question I asked in https://github.com/Khan/KaTeX/issues/1439#issuecomment-399452762:
>> The one scary thing might be someone doing `myDocumentFragment.classes.push('something')`, which will then act as a no-op...
>
> I think the real issue here is we're abusing the type interface.  But this was true before, too.  Perhaps the proper way would be to have two interfaces, one with style and one without (the former could subclass from the latter).  We could save this change for another PR if you prefer.

https://github.com/Khan/KaTeX/pull/1486#pullrequestreview-138069240

https://github.com/Khan/KaTeX/blob/536eb9f3dc68503129515600de9486e09700039c/src/functions/href.js#L34-L35

assumes `buildExpressionRow` returning a documentFragment.

There are many other instances of assigning a class, a style, or an attribute to documentFragment.

edemaine:
It seems like `HtmlDomNode` should be a subtype of `documentFragment` (representing the special case of a single-node fragment), whereas currently it's the other way around.

kevinbarabash:
If `documentFragment`s methods are a strict subset of `HtmlDomNode`'s methods than that sounds reasonable.

kevinbarabash:
Looking at `hast` and it's related ecosystem... it doesn't look particularly active so I don't see a particular strong reason for changing how we structure our HTML AST.  That being said, our HTML AST is quite different from how structure our LaTeX AST.  I think we might benefit from making out HTML AST look more like our LaTeX AST.  In particular:
- instead of using ES6 classes, everything could be POJOs (plain old javascript objects)
- each object would have a `type`, e.g. "svg", "span", "path", "line", etc.
- each object would have a `classes`, `attributes`, `children` and `styles` except for `fragment` which would have something like `nodes` which is a list of top-level (within the fragment) HTML nodes
- there would be a couple of render functions to generate HTML text or an HTML DOM

This side steps the question subclasses while making the code more regular.

kevinbarabash:
[...]

PR #1633, opened 2018-08-18
simplify combining chars

### LINKED ISSUE #1299: Combine `mathDefault` with `makeOrd`
Having font selection logic all in a single function seems like it would make things easier to understand.

kevinbarabash:
In preparation of refactoring domTree classes into POJOs as described in my comments in #1618, this PR removes the `tryCombine` method all of these classes.  All of the checks we were doing in `symbolNode`'s `tryCombine` method weren't actually necessary b/c we were only combining characters within the same group which means that they all had the same style/options to begin with.

This diff also merges `mathDefault` into `makeOrd` (fixes #1299) which makes it easier to see more of the font picking logic in one place.


src/functions/text.js
```diff
@@ -61,9 +61,11 @@ defineFunction({
     },
     htmlBuilder(group, options) {
         const newOptions = optionsWithFont(group, options);
+        // TODO(kevinb): split group.body into combinable parts to avoid having
+        // complicated checks in buildCommon.tryCombineChars.
```
kevinbarabash: The only things we can combine are `textord`s and `spacing`s so if we create groups of consecutive `textord`s and `spacing`s we can pass each of those groups off to `buildExpression` and combine the `symbolNodes` that are returned without doing all of complex checks we're doing right now in `tryCombineChars`.

src/buildCommon.js
```diff
@@ -282,21 +240,80 @@ const makeOrd = function<NODETYPE: "spacing" | "mathord" | "textord">(
...
+                domTree.createClass(next.classes)) {
+                continue;
+            }
+
+            let skip = false;
+
+            for (const style in prev.style) {
+                if (prev.style.hasOwnProperty(style)
+                    && prev.style[style] !== next.style[style]) {
+                    skip = true;
```
edemaine: Instead of using `skip` (and should probably use `break`) here, you can just `continue charloop` if you label the for loop block with `charloop:`.  (I've never done this myself, but see [these docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/break).)

By the way, odd that there is no `domTree.createStyle`...

src/buildCommon.js
```diff
@@ -282,21 +240,80 @@ const makeOrd = function<NODETYPE: "spacing" | "mathord" | "textord">(
...
 const tryCombineChars = function(chars: HtmlDomNode[]): HtmlDomNode[] {
     for (let i = 0; i < chars.length - 1; i++) {
-        if (chars[i].tryCombine(chars[i + 1])) {
+        const prev = chars[i];
+        const next = chars[i + 1];
+        if (prev instanceof domTree.symbolNode
+            && next instanceof domTree.symbolNode) {
+
+            if (domTree.createClass(prev.classes) !==
+                domTree.createClass(next.classes)) {
```
edemaine: I realize that this check is somewhat oversensitive, because the classes aren't sorted.  But I guess that's fine.

src/buildCommon.js
```diff
@@ -282,21 +240,80 @@ const makeOrd = function<NODETYPE: "spacing" | "mathord" | "textord">(
...
+                    skip = true;
+                }
+            }
+            if (skip) {
+                continue;
+            }
+
+            prev.text += next.text;
+            prev.height = Math.max(prev.height, next.height);
+            prev.depth = Math.max(prev.depth, next.depth);
```
edemaine: You seem to be missing `prev.italic = next.italic;`

src/buildCommon.js
```diff
@@ -282,21 +240,80 @@ const makeOrd = function<NODETYPE: "spacing" | "mathord" | "textord">(
...
     } else {
-        return mathDefault(text, mode, options, classes, type);
+        throw new Error("unexpected type: " + type + " in makeOrd");
     }
 };
 
 /**
- * Combine as many characters as possible in the given array of characters
- * via their tryCombine method.
+ * Combine consequetive domTree.symbolNodes into a single symbolNode.
```
edemaine: Maybe add a comment that `chars` is modified in place, as well as returned?  The former could be dangerous (and now the code that uses `tryCombineChars` doesn't make it obvious that it's changing in-place.)

src/functions/text.js
edemaine: I worry that some other functions might return the equivalent of a `textord` that wouldn't be caught before running the function...

src/buildCommon.js
kevinbarabash: Good call.  I too have never done that either.  I might move these checks into a separate function.

src/buildCommon.js
kevinbarabash: Yeah... I made a note that this function mutates the input.  Maybe I can make a copy fo the array so that we don't mutate the input.

kevinbarabash:
@edemaine I've assigned this to you to indicate that it's ready for review again.

ylemkimon:
There seems to be no commit pushed after https://github.com/Khan/KaTeX/pull/1633#pullrequestreview-147429135.

kevinbarabash:
I forgot to push.  I've pushed the changes I had locally now.

src/buildCommon.js
```diff
@@ -282,21 +240,84 @@ const makeOrd = function<NODETYPE: "spacing" | "mathord" | "textord">(
...
-        if (chars[i].tryCombine(chars[i + 1])) {
+        const prev = chars[i];
+        const next = chars[i + 1];
+        if (prev instanceof SymbolNode
+            && next instanceof SymbolNode
+            && canCombine(prev, next)) {
+
+            prev.text += next.text;
+            prev.height = Math.max(prev.height, next.height);
+            prev.depth = Math.max(prev.depth, next.depth);
```
ylemkimon: https://github.com/Khan/KaTeX/pull/1633#discussion_r211074256 by @edemaine 

> You seem to be missing `prev.italic = next.italic;`

kevinbarabash: The previous code check if `this.italic > 0` which felt weird so I left it out.  `prev.italic = next.italic` seems like a better way to go.

src/buildCommon.js
```diff
@@ -282,21 +240,88 @@ const makeOrd = function<NODETYPE: "spacing" | "mathord" | "textord">(
...
+ * Note: this function mutates the argument.
  */
-const tryCombineChars = function(chars: HtmlDomNode[]): HtmlDomNode[] {
+const tryCombineChars = (chars: HtmlDomNode[]): HtmlDomNode[] => {
     for (let i = 0; i < chars.length - 1; i++) {
-        if (chars[i].tryCombine(chars[i + 1])) {
+        const prev = chars[i];
+        const next = chars[i + 1];
+        if (prev instanceof SymbolNode
+            && next instanceof SymbolNode
```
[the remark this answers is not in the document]
kevinbarabash: I wanted to originally, but if we do then `flow` doesn't know that `prev` and `next` are `SymbolNodes` inside of `tryCombineChars` which means it doesn't recognize the props that we use on `prev` and `next` following this check.  It's a bummer b/c the code would look a lot cleaner.

It would be cool if `flow` had a way to describe "if this function returns `true` then the args of of these types".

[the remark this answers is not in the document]
ylemkimon: @kevinbarabash AFAIK, after `if (!(prev instanceof SymbolNode) || !(next instanceof SymbolNode)) return;`, flow would [refine](https://flow.org/en/docs/lang/refinements/) them as `SymbolNode`.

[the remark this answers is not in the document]
kevinbarabash: Ugh, I forgot to change the param types to `canCombine` to `prev: any, next: any`.  Also, I am impressed with `flow` ability to make a type refinement based on that check.  Thank you for prodding me on this.

kevinbarabash:
@edemaine can you have another look when you have a chance?

src/functions/op.js
```diff
@@ -95,7 +95,8 @@ export const htmlBuilder: HtmlBuilderSupSub<"op"> = (grp, options) => {
             base = inner[0];
             base.classes[0] = "mop"; // replace old mclass
         } else {
-            base = buildCommon.makeSpan(["mop"], inner, options);
+            base = buildCommon.makeSpan(
+                ["mop"], buildCommon.tryCombineChars(inner), options);
```
edemaine: This is a new call to `tryCombineChars`, right? This suggests to me that it should be called in `buildExpression`, not anywhere else. For example, `\text{{hello}}` probably won't combine the characters in the current code, because of the ordgroup in the way, but would if we called it in `buildExpression`.

Perhaps it should only be called in text mode, though? How is that currently detected? (Vague worry that that's what the `italic > 0` check was for...)

kevinbarabash: It is a new call.  It would be nice to handle cases like `\text{{hello}}`.  I tried collating `ordgroup`s but I had taken some of the checks out.  I should try again now that all of the checks are back in.  There are situations where I could see us not wanting to combine `ordgroups` though.  As an example: `{xyz}`... in this situation we might want to keep those characters separate in case someone wants to be able to highlight each variable on hover or click to select an individual variable in the term.  I think calling it for this particular case of `op` makes sense b/c it handles things like `sin`, `cos`, etc. and I think we should be treating those things as atomic units as opposed to separate letters.

kevinbarabash: I get what you're saying... try to combine characters inside of ordgroup if we're in text mode.  I'll make an issue for it.  Having fewer `span`s should improve performance.  The fewer DOM nodes we have to create the better.

PR #1635, opened 2018-08-18
Add intmath (Interactive Mathematics) to users list

edemaine:
As discussed in https://github.com/Khan/KaTeX/pull/1569#issuecomment-413801933


edemaine:
@ylemkimon Is it OK to merge other PRs while you're working on the release? (Feel free to merge if so.) Or should we wait?

ylemkimon:
@edemaine All changes except those to `docs/` and KaTeX source codes can be safely merged.

PR #1636, opened 2018-08-18
make all classes in domTree.js PascalCase and make them named exports

kevinbarabash:
The purpose of this refactor is make our more consistent as well as follow common naming conventions.


kevinbarabash:
@ylemkimon I'm assigning this to you since it's ready for review again.

Issue #1616, opened 2018-08-14
Convert `for` array iterations to array functions


ylemkimon:
From [Airbnb style guide](https://github.com/airbnb/javascript):

> Use `map()` / `every()` / `filter()` / `find()` / `findIndex()` / `reduce()` / `some()` / ... to iterate over arrays

```javascript
const numbers = [1, 2, 3, 4, 5];

// bad
let sum = 0;
for (let num of numbers) {
  sum += num;
}
sum === 15;

// good
let sum = 0;
numbers.forEach((num) => {
  sum += num;
});
sum === 15;

// best (use the functional force)
const sum = numbers.reduce((total, num) => total + num, 0);
sum === 15;

// bad
const increasedByOne = [];
for (let i = 0; i < numbers.length; i++) {
  increasedByOne.push(numbers[i] + 1);
}

// good
const increasedByOne = [];
numbers.forEach((num) => {
  increasedByOne.push(num + 1);
});

// best (keeping it functional)
const increasedByOne = numbers.map(num => num + 1);
```

Ref: https://github.com/Khan/KaTeX/pull/1469#discussion_r203921764

VinVaz:
Can I work on this one?

kevinbarabash:
@VinVaz got for it. :)

kevinbarabash:
One thing to call out is that these functions will make `flow` typing harder.

ylemkimon:
@kevinbarabash Actually, it'll be easier to flow-type, as it understands those functions cannot mutate the array.

kevinbarabash:
Things like `filter` don't work quite as expect if you're trying to filter out elements of a certain type in the hope that the resultant array will be typed correctly.  If you know how to get around this and other similar issues I would be forever grateful.

SakaRicky:
Hi,
Is this issue still open for I'm interested in contributing through it as my first contribution.

SakaRicky:
Sorry I now see the close tag

Santosh1176:
Hi, is this issue still open? As I am seeing the open tag, can take a stab at this?

PR #1641, opened 2018-08-19
To array functions

VinVaz:
In regard of issue #1616.
Some of the for..in loop were also changed to object methods, but there were situations that it was not clear if it was the best thing to do, since the object iterated, been an object variable referenced by "this", could have its value as undefined in some cases throwing an error when processed. So I have only changed a portion of them, I hope that's not a problem, but it's something to be aware of.


ylemkimon:
`find()` and `findIndex()` are not supported by IE and old browsers, and they are not polyfilled as Babel cannot determine the type at compile time. I apologize I forgot to mention it in the issue.

contrib/auto-render/auto-render.js
```diff
@@ -21,39 +20,38 @@ const renderMathInText = function(text, optionsCopy) {
...
-    }
+    });
 
     return fragment;
 };
 
 const renderElem = function(elem, optionsCopy) {
-    for (let i = 0; i < elem.childNodes.length; i++) {
-        const childNode = elem.childNodes[i];
+    elem.childNodes.forEach((childNode, i) => {
```
ylemkimon: `NodeList.prototype.forEach()` is not supported by IE and old browsers.

contrib/copy-tex/copy-tex.js
```diff
@@ -12,10 +12,8 @@ document.addEventListener('copy', function(event) {
         return;  // default action OK if no .katex-mathml elements
     }
     // Preserve usual HTML copy/paste behavior.
-    const html = [];
-    for (let i = 0; i < fragment.childNodes.length; i++) {
-        html.push(fragment.childNodes[i].outerHTML);
-    }
+    const html = fragment.childNodes.map(node => node.outerHTML);
```
ylemkimon: `fragment.childNodes` is a `NodeList` and it doesn't have `map` method. You can use `Array.prototype.map.call`.

src/functions/verb.js
```diff
@@ -21,17 +21,16 @@ defineFunction({
     },
     htmlBuilder(group, options) {
         const text = buildCommon.makeVerb(group, options);
-        const body = [];
         // \verb enters text mode and therefore is sized like \textstyle
         const newOptions = options.havingStyle(options.style.text());
-        for (let i = 0; i < text.length; i++) {
-            let c = text[i];
+        const body = text.map((element) => {
```
ylemkimon: `text` is a `String`, and it doesn't have `map` method.

src/Parser.js
```diff
@@ -310,11 +310,10 @@ export default class Parser {
      */
     handleUnsupportedCmd(): AnyParseNode {
         const text = this.nextToken.text;
-        const textordArray = [];
 
-        for (let i = 0; i < text.length; i++) {
-            textordArray.push({type: "textord", mode: "text", text: text[i]});
-        }
+        const textordArray = text.map((element) => {
```
kevinbarabash: This loop can't be converted b/c `text` is a `string`.

src/buildCommon.js
```diff
@@ -276,11 +276,10 @@ const makeOrd = function<NODETYPE: "spacing" | "mathord" | "textord">(
         } else if (ligatures.hasOwnProperty(text) &&
                    fontName.substr(0, 10) === "Typewriter") {
             // Deconstruct ligatures in monospace fonts (\texttt, \tt).
-            const parts = [];
-            for (let i = 0; i < text.length; i++) {
-                parts.push(makeSymbol(text[i], fontName, mode, options,
-                                      classes.concat(fontClasses)));
-            }
+            const parts = text.map((element) => {
```
kevinbarabash: This loop can't be converted b/c `text` is a `string`.

src/domTree.js
```diff
@@ -60,24 +60,20 @@ const toNode = function(tagName: string): HTMLElement {
     node.className = createClass(this.classes);
 
     // Apply inline styles
-    for (const style in this.style) {
-        if (this.style.hasOwnProperty(style)) {
-            // $FlowFixMe Flow doesn't seem to understand span.style's type.
-            node.style[style] = this.style[style];
-        }
-    }
+    Object.keys(this.style).map((style) => {
```
kevinbarabash: Please use `Object["key"]` to avoid babel polyfilling this for us.

src/domTree.js
```diff
@@ -60,24 +60,20 @@ const toNode = function(tagName: string): HTMLElement {
...
+        node.style[style] = this.style[style];
+    });
 
     // Apply attributes
-    for (const attr in this.attributes) {
-        if (this.attributes.hasOwnProperty(attr)) {
-            node.setAttribute(attr, this.attributes[attr]);
-        }
-    }
+    Object.keys(this.attributes).map((attr) => {
```
kevinbarabash: Please use `Object["key"]` to avoid babel polyfilling this for us.

src/functions/verb.js
```diff
@@ -21,7 +21,6 @@ defineFunction({
     },
     htmlBuilder(group, options) {
         const text = buildCommon.makeVerb(group, options);
-        const body = [];
```
kevinbarabash: Why was this removed?

VinVaz:
I can see there is still an error, I'm working on it

VinVaz:
Since my first PR every time I send a push to my fork it seems it updates the PR for you, sorry for the noise

kevinbarabash:
@VinVaz no worries.  You should be able to run `yarn test` locally which should help with identify some of the issues before pushing.

VinVaz:
The local test only throws a Spawn Error and I have not been able to find its source.

kevinbarabash:
My guess is that there's something wrong with your dev environment.  I'm guessing you're on Windows.  In order to narrow things down, I would try running each command separately that the `test` script from package.json runs and then Google for the failure in whichever one fails.  In the meantime, circleci has reported the following test failures:
```
Error ---------------------------------------------------------------------------------- src/environments/array.js:244:9

Cannot assign `outrow` to `body[r]` because:
 - an indexer property is missing in `span` [1].
 - an indexer property is missing in `span` [2].

   src/environments/array.js:244:9
   244|         body[r] = outrow;
                ^^^^^^^

References:
   src/buildCommon.js:348:4
   348| ): DomSpan {
           ^^^^^^^ [1]
   src/buildCommon.js:523:67
   523| const makeVList = function(params: VListParam, options: Options): DomSpan {
                                                                          ^^^^^^^ [2]
```

kevinbarabash:
Also, something must not have worked with `CRLF to LF` since there's still 255 changes files according to GitHub.

ylemkimon:
The best way to fix these problems would be to save the work elsewhere, delete the whole project, and cloning a fresh copy. (https://xkcd.com/1597/ 😄)

kevinbarabash:
Closing due to lack of activity.  @VinVaz feel free to reopen a fresh PR if you have time.

PR #1642, opened 2018-08-20
Fix favicon.ico 404

### LINKED ISSUE #1625: Several 404s
I realise the documentation is in development, but there are several broken things currently published.

On this page: https://khan.github.io/KaTeX/, the comparison gif, https://khan.github.io/img/katex-comparison.gif returns 404.

On several of the inner pages, the "docs/en/" and favicon links are not working. For example:

https://khan.github.io/KaTeX/docs/api.html
	https://khan.github.io/KaTeX/https://khan.github.io/favicon.ico
	  \_____ error code: 404 (not found)
	https://khan.github.io/KaTeX/docs/en/node.html
	  \_____ error code: 404 (not found)
	https://khan.github.io/KaTeX/docs/en/api.html
	  \_____ error code: 404 (not found)
	https://khan.github.io/KaTeX/docs/en/options.html
	  \_____ error code: 404 (not found)
	https://khan.github.io/KaTeX/docs/en/support.html
	  \_____ error code: 404 (not found)

ylemkimon:
Fixes #1625.

### DIFF
--- a/website/siteConfig.js
+++ b/website/siteConfig.js
@@ -151,7 +151,7 @@ const siteConfig = {
     /* path to images for header/footer */
     headerIcon: 'img/katex-logo.svg',
     footerIcon: 'img/katex-logo.svg',
-    favicon: 'https://khan.github.io/favicon.ico',
+    favicon: '../favicon.ico',
 
     disableHeaderTitle: true,
 


edemaine:
On Netlify, this generates:
```html
<link rel="shortcut icon" href="/../favicon.ico">
```
Is this important enough to fix? One approach would be to have our own `favicon.ico` in `img` and then set `favicon: 'img/favicon.ico'`... Or we could ask Docusaurus to fix support for absolute URLs in `favicon`.

ylemkimon:
@edemaine The parent directory of the root is the root (by the URL spec). I'm not sure we should have a Khan Academy favicon on Netlify.

edemaine:
Sounds good. Incidentally, this may change once we move to `katex.org`, where we don't need the `/KaTeX` root directory.

PR #1643, opened 2018-08-20
Fix screenshotter race conditions

ylemkimon:
I somehow thought `Promise`s are lazily evaluated.


ylemkimon:
@kevinbarabash Unfortunately, `eslint-plugin-promise` including `no-return-wrap` doesn't lint against these issues.

kevinbarabash:
@ylemkimon yes, you're right.  I was so excited to find an eslint plugin for promises in general that I got ahead of myself.  I think `flow` might be able to check for this b/c `execFile` returns a `ChildProcess` which is an object, not a callback, and `then()` should only accept a callback.  https://github.com/facebook/flow/blob/v0.79.1/lib/core.js#L583

Updating screenshotter.js to use `flow` is pretty low priority and we don't use `Promise`s anywhere in the main codebase.

ylemkimon:
Also it is valid syntax to pass a non-function to `then`. However, it's evaluation is not deferred after Promise resolution.

PR #1654, opened 2018-08-22
Release v0.10.0-rc.1

ylemkimon:
Bump master to v0.10.0-pre

Remove 0.10.0-rc website


edemaine:
@ylemkimon The SHA384 hashes seem incorrect.

ylemkimon:
@edemaine Of what file?

ylemkimon:
@edemaine No, it seems to be correct. If you compared to built files after checkout, you should change `package.json` version before building. Also, if you're on Windows, you should disable `autocrlf` feature of Git. If not, could you let me know which hash is incorrect?

PR #1658, opened 2018-08-23
Improve & fix CircleCI scripts

### LINKED ISSUE #1655: Code coverage drops if screenshotter tests are skipped

ylemkimon:
### Added
* Add Codecov flags: https://docs.codecov.io/docs/flags: https://github.com/Khan/KaTeX/issues/1655#issuecomment-415110112
* Add comments for #1590 workaround

### Changed
* Upgrade to [CircleCI 2.1 configuration](https://github.com/CircleCI-Public/config-preview-sdk)
  * Use reusable commands and jobs with parameters, instead of aliases

### Fixed
* Upgrade Yarn before installing dependencies, if needed: fixes issue #1648, which is caused by yarnpkg/yarn#5723
* Skip Flow, Jest, and coverage report upload if only website and documentation are changed: fixes #1655
* Set `CIRCLE_PREVIOUS_BUILD_NUM` empty before running greenkeeper-lockfile: it detects the correct build by checking it's the first job of the first workflow. As we run jobs parallel, the correct build may not be the first job: https://github.com/Khan/KaTeX/issues/1648#issuecomment-414671854

### Removed
* Remove CircleCI build artifacts, we use Netlify


.circleci/config.yml
```diff
@@ -76,62 +92,54 @@ jobs:
...
 
       - run:
-          name: Run tests
-          command: yarn test --coverage
-      - *codecov
-
+          name: Lint code
+          command: yarn test:lint
+      - skip_if_only_changed:
+          filter: '^static/|^website/'
```
edemaine: Why not `filter: '^docs/|^static/|^website/|^LICENSE|\\.md$'` as above?

.circleci/config.yml
```diff
@@ -76,62 +92,54 @@ jobs:
...
-
-      - *restore_node_modules_cache
-      - *yarn_install
-      - save_cache:
-          key: *node_modules_cache_key
-          paths:
-            - node_modules
+      - checkout_repo
+      - skip_if_only_changed:
+          filter: '^docs/|^LICENSE|\.md$'
```
edemaine: Why not `filter: '^docs/|^static/|^website/|^LICENSE|\\.md$'` as above?

.circleci/config.yml
```diff
@@ -1,68 +1,84 @@
...
+          # do not upload if screenshotter tests are skipped
+          command: ./node_modules/.bin/codecov -F "<< parameters.flag >>"
 
-    - *restore_node_modules_cache
-    - *yarn_install
+  screenshotter:
+    steps:
+      - checkout_repo
+      - skip_if_only_changed:
+          filter: '^docs/|^static/|^website/|^LICENSE|\.md$'
```
edemaine: Does `\.` need to be `\\.`?

.circleci/config.yml
ylemkimon: @edemaine It's because we lint JS and CSS in `static` and `website`.

.circleci/config.yml
ylemkimon: @edemaine YAML and Bash don't process escapes in single-quoted string, so it'll be safely parsed as regex escaping a dot.

.circleci/config.yml
```diff
@@ -1,68 +1,84 @@
...
+  install_dependencies:
+    steps:
+      - restore_cache:
+          keys:
+            - yarn-deps-v2-{{ checksum "yarn.lock" }}
+            - yarn-deps-v2-
+      - run:
+          name: Install dependencies
+          # upgrade Yarn before installing dependencies, if needed
+          command: yarn check || { curl -o- -L https://yarnpkg.com/install.sh | bash && yarn; }
```
kevinbarabash: I see.  Another option would be changing which image we're using for node.  The node 10 image contains yarn 1.9.2 https://github.com/nodejs/docker-node/blob/master/10/jessie/Dockerfile.  I guess that's a bigger change so we can upgrade yarn ourselves in the meantime.

.circleci/config.yml
```diff
@@ -1,68 +1,84 @@
...
-node_modules_cache_key: &node_modules_cache_key
-  yarn-deps-v2-{{ checksum "yarn.lock" }}
-
-restore_node_modules_cache: &restore_node_modules_cache
-  restore_cache:
-    keys:
-      - *node_modules_cache_key
-      - yarn-deps-v2-
+commands:
+  checkout_repo:
```
kevinbarabash: Wow... I didn't realize you could replace `checkout` with your own checkout command, cool!

.circleci/config.yml
```diff
@@ -1,68 +1,84 @@
...
+          name: Get changed files
+          command: |
+              if [[ $CIRCLE_PULL_REQUEST ]]; then
+                BASE_COMMIT=$(curl -s -H "Authorization: token a61ecb2fc5b72da54431""1b3db3875c96854958a8" \
+                  https://api.github.com/repos/$CIRCLE_PROJECT_USERNAME/$CIRCLE_PROJECT_REPONAME/pulls/$CIRCLE_PR_NUMBER \
+                  | jq -r ".base.sha | select(. != null)")...
+              else
+                BASE_COMMIT=HEAD^
+              fi
+              if [[ $BASE_COMMIT ]]; then
```
kevinbarabash: I thought `circleci` was only running on PRs, I guess that isn't actually the case.

.circleci/config.yml
```diff
@@ -1,68 +1,84 @@
...
+          command: |
+              git submodule sync
+              git submodule update --init --recursive
+      - run:
+          name: Get changed files
+          command: |
+              if [[ $CIRCLE_PULL_REQUEST ]]; then
+                BASE_COMMIT=$(curl -s -H "Authorization: token a61ecb2fc5b72da54431""1b3db3875c96854958a8" \
+                  https://api.github.com/repos/$CIRCLE_PROJECT_USERNAME/$CIRCLE_PROJECT_REPONAME/pulls/$CIRCLE_PR_NUMBER \
+                  | jq -r ".base.sha | select(. != null)")...
```
kevinbarabash: What does `jq -r ...` do?

.circleci/config.yml
```diff
@@ -1,68 +1,84 @@
...
-    - run:
-        name: Skip screenshotter if no KaTeX code has been changed
-        command: |
-            if [[ $CHANGED ]]; then
-              echo "$CHANGED" | grep -qvE '^contrib/|^docs/|^static/|^website/|^LICENSE|\.md$' || circleci step halt
-            fi
+  codecov:
+    parameters:
+      flag:
+        type: string
```
kevinbarabash: I didn't realize commands could be parameterized.  This is super powerful.

.circleci/config.yml
```diff
@@ -76,65 +92,56 @@ jobs:
...
+              yarn test:jest --coverage
+      - codecov:
+          flag: test
 
   firefox:
     docker:
       - image: circleci/node:6
       - image: selenium/standalone-firefox:2.48.2
-    <<: *screenshotter
+    steps:
```
kevinbarabash: How does "stacking" of images work?  I'm guessing that `selenium/standalone-firefox:2.48.2` doesn't include `node` and yet the `screenshotter` command runs `node`.

.circleci/config.yml
ylemkimon: @kevinbarabash But we may (unwittingly) break something in development environment on Node 6, if we test it on Node 10.

.circleci/config.yml
ylemkimon: @kevinbarabash It parses `json`, it is equivalent to
```js
var result = JSON.parse(pipe).base.sha;
if (result != null) return result;
return "";
```
in JS.

.circleci/config.yml
ylemkimon: @kevinbarabash Custom commands and parameters are new features in CircleCI 2.1.

.circleci/config.yml
ylemkimon: @kevinbarabash

> It is possible to specify multiple images for your job. Specify multiple images if, for example, you need to use a database for your tests or for some other required service. In a multi-image configuration job, all steps are executed in the container created by the first image listed. All containers run in a common network and every exposed port will be available on localhost from a primary container.

.circleci/config.yml
ylemkimon: @kevinbarabash We also run them for pushed commits.

.circleci/config.yml
kevinbarabash: Nice!

.circleci/config.yml
kevinbarabash: I was thinking about how I might use this for `vite`, but since there's a node script that needs to run from inside the container it uses even when running it locally I needed to include `node` in the custom container.  I was thinking of using one of circleci's so-called "browsers" images that include Chrome and Firefox and starts Xvfb... then I'd only need to download the `vite` code.

ylemkimon:
@kevinbarabash Thank you for the reviews!

PR #1660, opened 2018-08-23
Improve & fix release script

ylemkimon:
I've created a wiki page about release process: https://github.com/Khan/KaTeX/wiki/Releasing-KaTeX.

Tested with 0.10.0-rc.1 release (#1654).

## Added
* Allow specifying origin name other than `origin` via `ORIGIN` environment variable
* Add fallback to npm publish, if Yarn cannot authenticate, e.g., 2FA
* If gh-pages branch is protected, push to another branch

## Changed
* Replace master branch check with release branch check: allow creating a release from branch named other than `master`
* Remove redundant quotes and braces, replace -z with !

## Fixed
* Fix `package.json` version number when updating SRI hashes: `package.json` version number was not updated before updating SRI hashes resulting in wrong hashes


release.sh
```diff
@@ -204,14 +202,27 @@ else
...
 
     # Publish the website
+    # If gh-pages branch is protected, push to another branch
     pushd website
-    USE_SSH=true yarn publish-gh-pages
+    PUBLISH_GH_PAGES=$(USE_SSH=true yarn publish-gh-pages 2>&1 | tee /dev/tty || true)
+    if echo "$PUBLISH_GH_PAGES" | grep -qEi 'GH006|protected branch'; then
+        pushd build/KaTeX-gh-pages
+        git checkout -B "v$VERSION-gh-pages"
+        git push -u origin "v$VERSION-gh-pages"
```
edemaine: Should `origin` be `"$ORIGIN"`? Ah, I guess not -- this is meant for the forked repo.

release.sh
```diff
@@ -105,29 +106,16 @@ if [[ $INSANE != 0 ]]; then
...
-    # Make the commit and push
-    git add README.md contrib/*/README.md \
-        docs website/pages/index.html website/versioned_docs/
-    git commit -n -m "Update SRI hashes"
-    git push
-elif [[ ! $PUBLISH ]]; then
+if [[ ! $PUBLISH ]]; then
     # Make a release branch
-    git checkout -b "v$VERSION-release"
+    git checkout -B "v$VERSION-release"
```
edemaine: Nice, I didn't know about `-B`.

release.sh
```diff
@@ -204,14 +202,27 @@ else
...
     # Update npm (cdnjs update automatically)
-    yarn publish --new-version "${VERSION}"
+    # Fallback to npm publish, if yarn cannot authenticate, e.g., 2FA
+    yarn publish --new-version "$VERSION" || npm publish
 
     # Publish the website
+    # If gh-pages branch is protected, push to another branch
     pushd website
-    USE_SSH=true yarn publish-gh-pages
+    PUBLISH_GH_PAGES=$(USE_SSH=true yarn publish-gh-pages 2>&1 | tee /dev/tty || true)
```
edemaine: Will `yarn publish-gh-pages` respect `ORIGIN`?

release.sh
```diff
@@ -204,14 +202,27 @@ else
     git commit -n -m "v$VERSION"
     git diff --stat --exit-code # check for uncommitted changes
     git tag -a "v$VERSION" -m "v$VERSION"
-    git push origin "v$VERSION"
+    git push "$ORIGIN" "v$VERSION"
 
     # Update npm (cdnjs update automatically)
-    yarn publish --new-version "${VERSION}"
+    # Fallback to npm publish, if yarn cannot authenticate, e.g., 2FA
+    yarn publish --new-version "$VERSION" || npm publish
```
edemaine: Will this work in DRYRUN mode? I guess the first operation will always succeed, but I'd feel a little safer if we defined a function `npm() { echo "npm $*" }`.

release.sh
ylemkimon: @edemaine It will clone GitHub Pages repository fresh, which is Khan/KaTeX for KaTeX, and therefore it is set to `origin`.

release.sh
ylemkimon: @edemaine I've added `npm() { echo "npm $*" }`.

ylemkimon:

> @kevinbarabash will have different experience than the rest of us because he can publish directly to the `gh-pages` branch (even though it's protected).  I assume that's fine, but maybe worth discussing briefly?  Should the script put us all through the PR step, instead of trying to publish directly to gh-pages?)

I think the purpose of protecting the `gh-pages` branch is to prevent malicious code being pushed. As the website is generated from `docs`, there seems to be no need for reviewing its content.

release.sh
kevinbarabash: TIL

PR #1666, opened 2018-08-27
Add shouldRender and allow a function to be passed to ignoredTags.

### LINKED ISSUE #691: Suggestion: add a property called shouldRender to renderMathInElement options
I would like to suggest adding an optional property called `shouldRender` to the `renderMathInElement` options which if provided would accept the current HTML element being considered and return a Boolean indicating whether or not to render math inside it. This would be used in place of `ignoreTags`. Alternatively you could make `ignoreTags` optionally a function in the same way but the name makes less sense. This would neatly cover any scenario for opting out of math rendering and users could come up with their own conventions (like putting the attribute `data-render-math="false"` on elements they don't want rendered).

leungbk:
Closes #691.

### DIFF
--- a/contrib/auto-render/auto-render.js
+++ b/contrib/auto-render/auto-render.js
@@ -62,10 +62,28 @@ const renderElem = function(elem, optionsCopy) {
         } else if (childNode.nodeType === 1) {
             // Element node
             const className = ' ' + childNode.className + ' ';
-            const shouldRender = optionsCopy.ignoredTags.indexOf(
-                childNode.nodeName.toLowerCase()) === -1 &&
-                    optionsCopy.ignoredClasses.every(
-                        x => className.indexOf(' ' + x + ' ') === -1);
+            const inRenderedClass = optionsCopy.ignoredClasses.every(
+                x => className.indexOf(' ' + x + ' ') === -1);
+
+            let shouldRender;
+
+            if (inRenderedClass) {
+                if (optionsCopy.ignoredTags) {
+                    if (Array.isArray(optionsCopy.ignoredTags)) {
+                        shouldRender = optionsCopy.ignoredTags.indexOf(
+                            childNode.nodeName.toLowerCase()) === -1;
+                    } else if (typeof optionsCopy.ignoredTags === "function") {
+                        shouldRender = !optionsCopy.ignoredTags(childNode);
+                    }
+                } else {
+                    if (Array.isArray(optionsCopy.shouldRender)) {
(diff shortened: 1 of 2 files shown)


leungbk:
[clabot:check]

contrib/auto-render/auto-render.js
```diff
@@ -102,9 +120,17 @@ const renderMathInElement = function(elem, options) {
         // treating it as if it were the start of a KaTeX math zone.
         {left: "\\[", right: "\\]", display: true},
     ];
-    optionsCopy.ignoredTags = optionsCopy.ignoredTags || [
-        "script", "noscript", "style", "textarea", "pre", "code",
-    ];
+
+    if (optionsCopy.ignoredTags && optionsCopy.shouldRender) {
+        throw new Error(
+            "Please specify either ignoredTags or shouldRender, not both"
```
ylemkimon: Why not both?

leungbk: I felt it was more intuitive to specify only `.ignoredTags` (`.shouldRender`) and have everything not classified as such implicitly satisfy `.shouldRender` (`.ignoredTags`).

leungbk:
I'll try to figure out the decrease in code coverage.

leungbk:
It seems that even if I checkout the `master` branch, running `yarn test:jest:coverage` locally still gives only about `85%` coverage. The codecov bot on GitHub seemed to report `93%` (after reporting about `85%` yesterday) after I merged earlier, and continues to report `93%`. I'm not sure why `codecov` is still reporting `85%` on my computer, and why the codecov bot reported `85%` yesterday and `93%` today.

ylemkimon:
@leungbk Codecov result includes coverage from screenshot tests and we don't strictly enforce code coverage changes, especially on `contrib` extensions. So you don't need to worry about them.

contrib/auto-render/auto-render.js
ylemkimon: But the user may want to ignore some tags and other elements by another criterion.

leungbk: Would you provide an example of what you have in mind? The way I see it, if a user specifies both `optionsCopy.ignoredTags = ["script", "noscript", "style", "textarea", "pre", "code"]` and `optionsCopy.shouldRender = ["foo", "bar"]`, I wouldn't know what to do with a `<baz>` tag, whereas providing only one of the two properties would make it clearer.

ylemkimon:
Closing due to inactivity. Please feel free to reopen or create a new PR.

PR #1674, opened 2018-08-28
Include only necessary fonts for target environment specified by Browserslist

### LINKED ISSUE #1495: Remove support for TTF fonts

### LINKED ISSUE #1667: Document custom building
- `USE_TTF` option to exclude TTF in added in #1600 
- Browserslist added in #1662 and will be used by #1595

ylemkimon:
Fixes #1667.
Closes #1495.

\+ Allow WOFF and WOFF2 to be controlled using environment variables like TTF. (`USE_TTF`) These environment variables will override target environment.

Uses caniuse data to detect support. No additional dependencies are required, as they are already included in other dependencies.


docs/node.md
```diff
@@ -36,5 +36,15 @@ yarn
 yarn build
 ```
 
+It will automatically transpile code and include only necessary fonts for
+target environments specified by [Browserslist config](https://github.com/browserslist/browserslist#environment-variables).
+For example, if you are making a web app for a kiosk with Chrome 68, run
+`BROWSERSLIST="Chrome 68" yarn build` and it will produce build with no
```
kevinbarabash: I ran this command got the following error:
```
error "browserslist#node-releases@^1.0.0-alpha.11" doesn't satisfy found match of "node-releases@1.0.0-alpha.10"
error "browserslist#node-releases" is wrong version: expected "1.0.0-alpha.11", got "1.0.0-alpha.10"
```

ylemkimon: @kevinbarabash This PR updates some package versions by `yarn.lock`. You need to `yarn install` first. (If you did, maybe due to recent NPM downtime: https://status.npmjs.org/)

kevinbarabash: If there a difference between running `yarn` and `yarn install`?  It looks like the `circleci: test` check is running into this issue too.

ylemkimon: @kevinbarabash AFAIK, `yarn` and `yarn install` is equivalent: https://yarnpkg.com/en/docs/cli/#toc-default-command. The CircleCI build failures are due to old Yarn version (which #1658 addresses).

ylemkimon:
Test failures are due to the same issue in #1648, which'll be fixed by #1658.

docs/node.md
ylemkimon: @kevinbarabash The error seems to be fixed!

ylemkimon:
@kevinbarabash There was a bug in #1595, where Browserslist config is not read even if `isESMBuild` is false. Also, I've upgraded `uglifyjs-webpack-plugin` to `terser-webpack-plugin`, to be able to minify ES6 output.

PR #1680, opened 2018-08-30
Move to katex.org

ylemkimon:
* Change links to katex.org
* Remove baseUrl hacks

### DIFF
--- a/release.sh
+++ b/release.sh
@@ -135,8 +135,8 @@ if [[ ! $PUBLISH ]]; then
         fi
 
         # Edit docs to use CSS from CDN
-        grep -l '{@stylesheet: katex.min.css}' docs/*.md | xargs sed -i.bak \
-            's|{@stylesheet: katex.min.css}|<link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/katex@./dist/katex.min.css" integrity="sha384-katex.min.css" crossorigin="anonymous"/>|'
+        grep -l '/static/' docs/*.md | xargs sed -i.bak \
+            's|/static/|https://cdn.jsdelivr.net/npm/katex@./dist/katex.min.css" integrity="sha384-|'
 
         # Update the version number in CDN URLs included in the README and the documentation,
         # and regenerate the Subresource Integrity hash for these files.
--- a/release.sh
+++ b/release.sh
@@ -136,7 +136,7 @@ if [[ ! $PUBLISH ]]; then
 
         # Edit docs to use CSS from CDN
         grep -l '/static/' docs/*.md | xargs sed -i.bak \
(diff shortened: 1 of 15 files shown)


release.sh
```diff
@@ -135,8 +135,8 @@ if [[ ! $PUBLISH ]]; then
         fi
 
         # Edit docs to use CSS from CDN
-        grep -l '{@stylesheet: katex.min.css}' docs/*.md | xargs sed -i.bak \
-            's|{@stylesheet: katex.min.css}|<link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/katex@./dist/katex.min.css" integrity="sha384-katex.min.css" crossorigin="anonymous"/>|'
+        grep -l '/static/' docs/*.md | xargs sed -i.bak \
+            's|/static/|https://cdn.jsdelivr.net/npm/katex@./dist/katex.min.css" integrity="sha384-|'
```
edemaine: This seems a little fragile going forward. What if we used an image from `/static/`? Maybe we should use a regex (with `egrep` and `sed`) like `static/[^ ]*css` or `static/katex.min.css`. Oh, also, the replacement string is incorrect as is, but would be correct with such a query.

website/pages/index.html
```diff
@@ -12,8 +12,8 @@
     <meta name="Description" content="KaTeX – The fastest math typesetting library for the web">
     <meta property="og:title" content="KaTeX – The fastest math typesetting library for the web">
     <meta property="og:type" content="website">
-    <meta property="og:url" content="http://khan.github.io/KaTeX/">
-    <meta property="og:image" content="http://khan.github.io/KaTeX/og_logo.png">
+    <meta property="og:url" content="https://katex.org/">
+    <meta property="og:image" content="https://katex.org/og_logo.png">
```
edemaine: This link isn't working.

release.sh
ylemkimon: @edemaine I've changed to only update URLs prefixed with `katex`, e.g., `/static/katex.min.css`.

edemaine: Isn't the second half of the `s` command still incorrect?

ylemkimon: @edemaine Sorry, I forgot the original replacement string was `{@stylesheet: katex.min.css}`. Other static assets than KaTeX and `index.css` are not served from `/static/`, so I think it's safe to assume it's built KaTeX files. I've changed to capture the file name.

edemaine: Did you forget to push, or are you still working on this?

ylemkimon: @edemaine I've amended the last commit. (4dbf4fc)

edemaine:
Cool, another change to the release script that needs testing. 😄

ylemkimon:
@edemaine Thank you!

@kevinbarabash Could you change the link on the repo description?

kevinbarabash:
I've updated the link on the repo description.  Thanks for updating all of our links!

Issue #1507, opened 2018-07-27
"Docs" links in the footer 404


kevinbarabash:
Repro steps:
- go to https://khan.github.io/KaTeX/docs/api.html
- scroll to the bottom
- click on any of the links in the footer under the "Docs" heading

<img width="214" alt="screen shot 2018-07-27 at 5 49 57 pm" src="https://user-images.githubusercontent.com/1044413/43347950-952e5546-91c5-11e8-9cfd-554697060dec.png">

PR #1683, opened 2018-08-30
Move to Netlify

ylemkimon:
* Remove GitHub Pages publish script (and from release script)
* Redirect katex.netlify.com to katex.org
* Include includeSubDomains and preload directive in HSTS header
* Cache static image assets

I've opened this PR separate from #1680, so that we can simply revert this PR if Netlify is no longer available or we choose to roll back to use GitHub Pages.


website/static/_redirects
```diff
@@ -0,0 +1,6 @@
+https://katex.netlify.com/* https://katex.org/:splat 301!
+
+/function-support.html /docs/supported.html 301
+
+/docs/en/* /docs/:splat 301
+/docs /docs/node.html 301
```
kevinbarabash: Are redirects needed for each subdirectory?

ylemkimon: @kevinbarabash We don't have `index.html` for `docs` directory, so people going to https://katex.org/docs will have 404. Also previously, there were 404s due to wrong URLs (`docs/en`, #1507), so this redirects `docs/en/XXX` to `docs/XXX`.

PR #1684, opened 2018-08-30
Add Flow types to CssStyle.

### LINKED ISSUE #1677: add better flow types to CssStyle
Currently it's
```
export type CssStyle = {[name: string]: string};
```
which isn't all that helpful.  It could be made more precise specifying all of the possible keys.  To make these easier use https://github.com/Khan/wonder-blocks/blob/master/flow-typed/aphrodite.flow.js as starting point.

leungbk:
Closes #1677.

I wanted to use an exact object type, but Flow wouldn't allow an empty, unsealed object to be assigned. Is there a better way of doing this?

### DIFF
--- a/src/domTree.js
+++ b/src/domTree.js
@@ -125,7 +125,33 @@ const toMarkup = function(tagName: string): string {
     return markup;
 };
 
-export type CssStyle = {[name: string]: string};
+// Making the type below exact with all optional fields doesn't work due to
+// - https://github.com/facebook/flow/issues/4582
+// - https://github.com/facebook/flow/issues/5688
+// However, since *all* fields are optional, $Shape<> works as suggested in 5688
+// above.
+// This type does not include all CSS properties. Additional properties should
+// be added as needed.
+export type CssStyle = $Shape<{
+    backgroundColor: string,
+    borderBottomWidth: string,
+    borderColor: string,
+    borderRightWidth: string,
+    borderTopWidth: string,
+    bottom: string,
+    color: string,
+    height: string,
+    left: string,
+    marginLeft: string,
+    marginRight: string,
+    marginTop: string,
+    minWidth: string,
+    paddingLeft: string,
+    position: string,
+    top: string,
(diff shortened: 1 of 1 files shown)


src/domTree.js
```diff
@@ -125,7 +125,26 @@ const toMarkup = function(tagName: string): string {
     return markup;
 };
 
-export type CssStyle = {[name: string]: string};
+export type CssStyle = {
```
kevinbarabash: Please change this to an exact object type, e.g.
```
export type CssStyle = {|
    backgroundColor?: string,
    ...
|}
```
This will cause `flow` to complain about any misspelled properties.

src/domTree.js
```diff
@@ -125,7 +125,31 @@ const toMarkup = function(tagName: string): string {
     return markup;
 };
 
-export type CssStyle = {[name: string]: string};
+// Making the type below exact with all optional fields doesn't work due to
+// - https://github.com/facebook/flow/issues/4582
+// - https://github.com/facebook/flow/issues/5688
+// However, since *all* fields are optional, $Shape<> works as suggested in 5688
+// above.
```
ylemkimon: Could you add a comment this is not comprehensive?

leungbk: I amended the commit to include `// Optional settings for styling.` on the line above this snippet (included in the current submission, though you'll have to view the rest of the file to confirm)---does that suffice?

leungbk: OK, I've added your recommendation to the line just below this snippet.

leungbk:
@kevinbarabash What additional properties did you have in mind? I don't mind adding more; the reason the list is so sparse is that I simply used `ripgrep` to search for properties that are presently used on the objects of type `CssStyle`.

kevinbarabash:
@leungbk I thought about it more and we can add more props to `CssStyle` later and if it does catch people off guard we can always add all them at that point.

kevinbarabash:
Thanks for the PR.

PR #1685, opened 2018-08-30
Redirect khan.github.io/katex to katex.org

ylemkimon:
`jekyll-redirect-from` plugin generates redirect pages with `refresh` meta tag and `canonical` link tag.

I've left image assets, as it is impossible to create 301 redirects and there may be hot-linked images.

Also, enabled Google Webmasters tools to improve SEO.


kevinbarabash:
@ylemkimon is there a way to test this locally?  Also, with respect to "Google Developer Tools", is that what `google-site-verification: google4e8111332337b304.html` is for?  How does it improve SEO?

kevinbarabash:
@ylemkimon I've tried a couple of pages on https://ylemkimon.github.io/KaTeX/ and the redirects seem to be working.  It looks like this has be deployed before we can change an settings in Google Developer Tools.

PR #1688, opened 2018-08-31
Enable Algolia Docsearch (documentation search)

### LINKED ISSUE #1513: Enable search on the documentation
It is possible to enable search on Docusaurus, using [Algolia DocSearch](https://community.algolia.com/docsearch/): https://docusaurus.io/docs/en/search

CC @kevinbarabash

ylemkimon:
Fixes #1513.


website/siteConfig.js
```diff
@@ -185,6 +185,14 @@ const siteConfig = {
 
     separateCss: ['static/static', 'static\\static'],
 
+    algolia: {
+        apiKey: '46ecd80046d78d4e5d9a5c06f559dfaa',
```
kevinbarabash: Where did this key come from?  It feels a little unsafe to include a plaintext API key like this.

ylemkimon: @kevinbarabash It is provided Algolia. It is Search-only API Key, which is safe to use in front-end code and in production. Also, Docusaurus documentation suggests entering the API key to its configuration: https://docusaurus.io/docs/en/search.

ylemkimon:
The configuration for our DocSearch is available at: https://github.com/algolia/docsearch-configs/blob/master/configs/katex.json.

One question is should we include supported functions in `table`s, i.e., `td`s?

kevinbarabash:
> One question is should we include supported functions in tables, i.e., tds?

Not sue.  It looks things are set up to use `td` so let's give that a try.  I assume we can change it later if need be.

Issue #1657, opened 2018-08-22
Fix \textcolor and \color to match what the color package does


jessebett:
I'm lost as to how I'm supposed to programmatically use custom colours. Especially since `\definecolor` is not supported.

To get around this I tried using `\newcommand` to define a new colour. 
`\newcommand{\myblue}[1]{\textcolor{#0074D9}{#1}}` but this is not valid because the `#` symbol needed for `textcolor` is parsed as an invalid number for the argument `#`. Any ideas on how to get around this?

Further, does `\newcommand` not make the command global? How do I have any new colours defined for the entire page?

kevinbarabash:
I'm not sure if it's possible to work around this issue with `\newcommand`.  Since https://github.com/Khan/KaTeX/issues/750 is already requesting support for `\definecolor` I'm going to close this as a duplicate.

kevinbarabash:
The `HTML` color model from https://en.wikibooks.org/wiki/LaTeX/Colors#Defining_new_colors shows that `\definecolor` using a hex color without the leading `#`.  It looks like `\color` and `\textcolor` require the user to specify the color model unless using a defined color so our current implementation of these commands appears to be incorrect.

kevinbarabash:
`color` package reference: http://ctan.mirror.rafal.ca/macros/latex/required/graphics/color.pdf

kevinbarabash:
I skimmed the PDF but it doesn't make mention of all of the color models listed on https://en.wikibooks.org/wiki/LaTeX/Colors.  I think it's fine to support more color models but it should be done in a way that's compatible with the `color` package.

edemaine:
@jessebett You can fix the `#` issue as follows:

```latex
\newcommand{\myblue}[1]{\textcolor{##0074D9}{#1}}
```

(This is how `#`s get escaped in TeX macros.)

If you want a global definition, you can use `\gdef`:

```latex
\gdef\myblue#1{\textcolor{##0074D9}{#1}}
```

(Unfortunately, LaTeX has no global version of `\newcommand`.)

Both of these are generic advice about LaTeX not specific to KaTeX. Specific to KaTeX, there's another way to specify global macros: you can specify the macro in the `macros` option when calling KaTeX.

I agree that we should probably deprecate the `#` if it's not there in LaTeX. KaTeX's color support has always been HTML focused, and we should instead bring it into alignment with LaTeX.

jessebett:
@edemaine your suggestion throws an error:

```
$$
[...]

PR #1690, opened 2018-08-31
Allow 6-digit color spec w/o #

ronkok:
This PR takes a minimalist approach to addressing issue #1657. This does **not** completely change the color specification syntax to align with the `color` package or the `xcolor` package. It does not accommodate RGB notation or a cmyk color model.  It does, however, allow a 6-digit HTML color specification to pass without a leading `#` symbol, as in `xcolor`'s HTML color model.

### DIFF
--- a/src/Parser.js
+++ b/src/Parser.js
@@ -749,7 +749,7 @@ export default class Parser {
         }
         let color = match[0];
         if (/^[0-9a-f]{6}$/i.test(color)) {
-            // We allow a 6-digit HTML color spec without a leading "#". 
+            // We allow a 6-digit HTML color spec without a leading "#".
             // This follows the xcolor package's HTML color model.
             // Predefined color names are all missed by this RegEx pattern.
             color = "#" + color;
--- a/src/Parser.js
+++ b/src/Parser.js
@@ -743,14 +743,25 @@ export default class Parser {
         if (!res) {
             return null;
         }
-        const match = (/^(#[a-f0-9]{3}|#[a-f0-9]{6}|[a-z]+)$/i).exec(res.text);
+        const match = (/^(#[a-f0-9]{3}|#?[a-f0-9]{6}|[a-z]+)$/i).exec(res.text);
         if (!match) {
             throw new ParseError("Invalid color: '" + res.text + "'", res);
+        }
+        let color = match[0];
+        if (color.charAt(0) !== "#" && color.length === 6) {
+           // Check for HTML predefined color words that are 6 letters long.
+           // If not predefined, allow a 6-digit color specification without
+           // a leading #, as in the xcolor package's HTML color model.
+           const colorWordRegEx = RegExp("^(silver|maroon|purple|yellow|orange" +
+               "|bisque|indigo|orchid|salmon|sienna|tomato|violet)$",'i');
+           if (!colorWordRegEx.test(color)) {
+               color = "#" + color;
+           }
         }
         return newArgument({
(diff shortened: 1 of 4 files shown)


src/Parser.js
```diff
@@ -743,14 +743,25 @@ export default class Parser {
...
         if (!match) {
             throw new ParseError("Invalid color: '" + res.text + "'", res);
         }
+        let color = match[0];
+        if (color.charAt(0) !== "#" && color.length === 6) {
+           // Check for HTML predefined color words that are 6 letters long.
+           // If not predefined, allow a 6-digit color specification without
+           // a leading #, as in the xcolor package's HTML color model.
+           const colorWordRegEx = RegExp("^(silver|maroon|purple|yellow|orange" +
+               "|bisque|indigo|orchid|salmon|sienna|tomato|violet)$",'i');
```
kevinbarabash: Since we're passing through strings whose length aren't 6 maybe we check that `color` matches a hex 6-tuple that way we're not prefix things like `cherry`.  What's LaTeX's behaviour if it gets a named color that isn't defined?  Also, you said this PR was minimal so may we can deal with those situations in subsequent PR.

ylemkimon: Every supported 6-letter color names seem to be out of hexadecimal digits (`[0-9a-f]`). I think it'd be better to test for `[0-9a-f]{6}`. Also, why not support 3-digit color codes?

ronkok: >What's LaTeX's behaviour if it gets a named color that isn't defined? 

There is an error message such as: `!Package xcolor Error: Undefined color 'cherry'.` quicklatex.com will still render a black expression if the `Report LaTeX errors` box is unchecked. So the error behavior is somewhat similar to KaTeX. That is, KaTeX also will render a black expression.

> maybe we check that color matches a hex 6-tuple

I'll see if that can be done fairly easily.

ronkok:
The latest commit revises the RegEx pattern as recommended by @ylemkimon. (Thanks!) This simplifies the code but it does not raise a parse error on a mis-typed predefined color name. `\color{cherry}` will render in black.

I'd like to raise an error on a mis-typed predefined color name, but the only way I can think to accomplish that is via a RegEx pattern that enumerates all 150 of the [HTML predefined colors](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value#Color_keywords). Is that something I should do?

src/Parser.js
```diff
@@ -743,14 +743,21 @@ export default class Parser {
...
+            // We allow a 6-digit HTML color spec without a leading "#". 
+            // This follows the xcolor package's HTML color model.
+            // Predefined color names are all missed by this RegEx pattern.
+            color = "#" + color;
+        }
         return newArgument({
             type: "color-token",
             mode: this.mode,
-            color: match[0],
+            color: color,
```
ylemkimon: This can be just `color,`.

ylemkimon:
> I'd like to raise an error on a mis-typed predefined color name, but the only way I can think to accomplish that is via a RegEx pattern that enumerates all 150 of the [HTML predefined colors](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value#Color_keywords). Is that something I should do?

I think it is impossible without access to DOM APIs and there is no need to check the color name is valid.

src/Parser.js
ronkok: Done.

jessebett:
Just wanted to come in here and say thanks for this!

PR #1634, opened 2018-08-18
Release v0.10.0-rc

ylemkimon:
As we have many infrastructure changes, I think it'd be safer to release RC version first.

PR #1692, opened 2018-09-02
Added correct hash for katex.min.css

iSumitG:
The hash for katex.min.css was incorrect, hence breaking the rendering if copied from the official website.

### DIFF
--- a/docs/browser.md
+++ b/docs/browser.md
@@ -15,7 +15,7 @@ Use CDN to deliver KaTeX to your project:
 KaTeX also provides minified versions:
 
 ```html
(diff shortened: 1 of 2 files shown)


iSumitG:
[clabot:check]

ylemkimon:
Thank you for reporting and sending a PR! It seems hashes for both files got mixed in https://github.com/Khan/KaTeX/pull/1634/commits/7f3c0f247eb92ae8c8d84eab1911357a95646507 during v0.10.0-rc release, and as we're using previous hashes for the replacement key of new hases, it produced same hashes for both files. I'm not sure what caused hash mix-up in the first place, but I'll continue to investigate.

PR #1693, opened 2018-09-02
Add sitemap.xml

ylemkimon:
This improves SEO, by setting katex.org as canonical URLs in `sitemap.xml`.

### DIFF
--- a/sitemap.xml
+++ b/sitemap.xml
@@ -1,0 +1,32 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<urlset xmlns="http://www.sitemaps.org/schemas/sitemap/0.9" xmlns:news="http://www.google.com/schemas/sitemap-news/0.9" xmlns:xhtml="http://www.w3.org/1999/xhtml" xmlns:mobile="http://www.google.com/schemas/sitemap-mobile/1.0" xmlns:image="http://www.google.com/schemas/sitemap-image/1.1" xmlns:video="http://www.google.com/schemas/sitemap-video/1.1">
+<url> <loc>https://khan.github.io/KaTeX/users.html</loc> <changefreq>weekly</changefreq> <priority>0.5</priority> <xhtml:link rel="canonical" href="https://katex.org/en/users.html" /> </url>
+<url> <loc>https://khan.github.io/KaTeX/versions.html</loc> <changefreq>weekly</changefreq> <priority>0.5</priority> <xhtml:link rel="canonical" href="https://katex.org/en/versions.html" /> </url>
+<url> <loc>https://khan.github.io/KaTeX/docs/next/api.html</loc> <changefreq>hourly</changefreq> <priority>1.0</priority> <xhtml:link rel="canonical" href="https://katex.org/docs/next/api.html" /> </url>
+<url> <loc>https://khan.github.io/KaTeX/docs/next/autorender.html</loc> <changefreq>hourly</changefreq> <priority>1.0</priority> <xhtml:link rel="canonical" href="https://katex.org/docs/next/autorender.html" /> </url>
+<url> <loc>https://khan.github.io/KaTeX/docs/next/browser.html</loc> <changefreq>hourly</changefreq> <priority>1.0</priority> <xhtml:link rel="canonical" href="https://katex.org/docs/next/browser.html" /> </url>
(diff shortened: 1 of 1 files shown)


ylemkimon:
@edemaine For Google, sitemap can be submitted using Google Search Console (formerly Google Webmaster Tools).

mbourne:
@edemaine: The sitemap.xml is correctly at the root level.
@ylemkimon: Some comments:
(1) The `&lt;changefreq&gt;`  tag is not really necessary - Google ignores it and crawls the pages within the site whenever it feels like it. It could be removed.
(2) Having all `&lt;priority&gt;` tags set to 1 doesn't help the search engines to prioritise any page over another. This tag is only possibly used by search engines if 2 pages end up having an identical SEO score (highly unlikely) and it helps to decide where in the list of katex.org pages it will come in the search results. But in reality, it has little importance, and I suggest it be removed.
(3) Are there translations of the site already? Or are they intended? If so, the `rel="alternate" hreflang` part would be necessary, but until then, probably not.
(4) The search engines look for sitemap.xml in the root by default, so there's not really a need to submit in Google Search Console. It doesn't hurt for a new site, though! 
(5) It is probably worth adding a robots.txt with a link to sitemap.xml.

FYI, these are the pages Google has already indexed to date:
https://www.google.com/search?q=site:katex.org

ylemkimon:
@mbourne Thank you for your great insights!

This PR is for our old site (https://khan.github.io/KaTeX), which Google marked as canonical (authoritative) version of our new site (https://katex.org). I assumed if I add the new site as canonical in the sitemap as well as HTML link tag, Google will mark the new site as canonical. If you have any suggestions on how to deprecate the old site, please let us know.

(1)(2) These tags are auto-generated by our documentation tool, Docusaurus. If they hurt SEO, could you report the issue at https://github.com/facebook/Docusaurus/issues?

(3) We only have English translations, and they are added as alternate link with hreflang en in our main site’s sitemap: https://katex.org/sitemap.xml

(4) The URL of the old site is not root, i.e., https://khan.github.io/KaTeX. Does Google look at the subdirectories?

(5) If we’d like to put no restrictions on crawling, is robots.txt necessary?

Thank you again!

mbourne:
@ylemkimon I've had a closer look at what you've done on the old and new sites and I think it looks good. It appears to be working as intended.

I'd only suggest the following on the old site pages: Change the order of the javascript redirect and the meta-refresh, so it's like this:

`<meta http-equiv="refresh" content="0; url=https://katex.org/"><script>location="https://katex.org/"</script>`

(With the javascript one coming first, the browser will start working on that before it works on the meta refresh. For the search robots, a meta-refresh with time 0 is equivalent to a 301 - which is what we want -  while a script-based redirect is not as robust for some robots. Actually, I would personally leave out the script one, but if you have included it for a reason, fine!)

Your canonical tag is good on both old and new sites, and will pass the SEO "juice" appropriately to the new pages. I've noticed Google can be a bit slow to list the new pages - it's as though it wants to make sure you've really moved it before it changes the search index entry.

So in summary, I think what you've done will deprecate the old site well - nothing more to do.

For items (1) and (2), it's not really a case of them hurting SEO - it's just they don't help and take up (a little bit of) bandwidth, so could go. Not a biggie.

For (3) - what I meant was the alternate tag is not really necessary if you only have English versions. If there are other languages added later, then you would add the `hreflang` tag. Once again it doesn't hurt, but it doesn't help, either.

(4) You can actually put the sitemap.xml anywhere (root or not) and point to it from robots.txt, which has to be in the root (which you won't have access to for khan.github.io, and not necessary). But with this tag in the old pages, `<meta name="robots" content="noindex">`, you have already addressed this and it's now a non-issue.

(5) The answer is "no", but I've had cases where someone pointed to text-based files and they showed up in the search results as mostly gibberish, and you may find that later. But for now, I believe it would be worth adding a robots.txt on the new site with simply this:

`User-agent: *`
`Sitemap: https://katex.org/sitemap.xml`

kevinbarabash:
> These tags are auto-generated by our documentation tool, Docusaurus.

Does that mean the whole file is generated by Docusaurus?

ylemkimon:
@kevinbarabash Yes, it is not documented behavior, but it generates `sitemap.xml`.

kevinbarabash:
Is there a command you have to manually run to generate it?

ylemkimon:
@kevinbarabash It is part of build process(`yarn build`), so it is automatically deployed.

PR #1700, opened 2018-09-05
Fix `\mathit` font and italic correction and add `\mathnormal`

### LINKED ISSUE #584: \mathit uses the wrong font
Consider: 

```tex
$helfo x \; \mathit{helfo x}$
```

Here’s what `\showlists` says. Note that `\mathit` uses a different font (`\fam4`).

```
\mathord
.\fam1 h
\mathord
.\fam1 e
\mathord
.\fam1 l
\mathord
.\fam1 f
\mathord
.\fam1 o
\mathord
.\fam1 x
\glue(\mskip) 5.0mu plus 5.0mu
\mathord
.\mathord
..\fam4 h
.\mathord
..\fam4 e
.\mathord
..\fam4 l
.\mathord
..\fam4 f
.\mathord
..\fam4 o
.\mathord
..\fam4 x
```

Here’s the texcmp diff (red = katex).

image

This highlights the spacing diffs but `\mathit` has different glyphs too—the narrower glyphs of the text italic font.

SOLUTION: Basically I think `\mathit` should use `font-family: KaTeX_Main`, rather than `font-family: KaTeX_Math`. Also, `\mathit` should not apply italic corrections—none of the `\math…` font commands should. But I'm not clear on any finer differences between KaTeX_Main and KaTeX_Math fonts.

### LINKED ISSUE #1250: \mathnormal vs. \mathit
TeX has two "math italic" fonts with different italic corrections. Here's a comparison from quicklatex for `fff\mathnormal{fff)\mathit{fff)`:

ql_f9de3eed425b22fbe468db5d6af6db20_l3

KaTeX doesn't support `\mathnormal` but renders all italic text as if it is mathnormal. This is correct in the default, but is *not* how `\mathit` is supposed to work. 

This might be why we have two italic fonts (Main and Math) but I'm not sure. We should investigate how TeX does this. 

P.S. We might also look at what MathJax does, as it gets this right.

ylemkimon:
Fixes #584 and fixes #1250.

KaTeX uses the same font (`Math-Italic`, Computer Modern Math Italic[`cmmi`], class `mathit` for **alphabets** and `Main-Regular`, Computer Modern Roman[`cmr`] for **digits**) for math atoms outside font commands (default) and in `\mathit`.

However, `\mathit` in LaTeX uses text italic font (`Main-Italic`, Computer Modern Text Italic[`cmti`], class `mainit`) for both **alphabets and digits**, which has narrow widths (spacing) and no italic correction, same as `\textit`. This PR changes `\mathit` font to `Main-Italic` and remove its italic correction. (0d52009) This also fixes old font command `\it`, as it uses `mathit` class.

```latex
\DeclareMathAlphabet{\mathit}{OT1}{cmr}{m}{it}
```

mathit

(Green is LaTeX, red is KaTeX, aligned manually. The spacing around accent character seems to be still wrong but is in the right direction, I'll open an issue after this is merged. Wrong accent position is tracked by #1099, incorrect spacing around dotless i by #354, and \KaTeX mismatch by #1703)

Also, I've renamed the name `mathit` to `mathdefault` and `mainit` to `mathit` to avoid confusion. (f354da8) We should consider changing the font name.

Furthermore, I've added `\mathnormal`, which is similar to default math font but uses old style (`Caligraphic`) font for **digits** and math italic (`Math-Italic`) font for uppercase Greek alphabets as well as other letters that is in math italic font. (702c454)

mathnormal

(Green is LaTeX, red is KaTeX, incorrect spacing around dotless i is tracked by #354 and \KaTeX mismatch by #1703)


src/buildMathML.js
```diff
@@ -85,7 +85,7 @@ export const getVariant = function(
     }
 
     const font = options.font;
-    if (!font) {
+    if (!font || font === "mathnormal") {
```
ylemkimon: I'm not sure what variant `\mathnormal` should use, but as it acts similar to default math font, I used `null`.

src/buildCommon.js
```diff
@@ -210,15 +239,15 @@ const makeOrd = function<NODETYPE: "spacing" | "mathord" | "textord">(
...
-            const fontData = mathit(text, mode, options, classes);
+        if (fontOrFamily === "boldsymbol" || fontOrFamily === "mathnormal") {
+            const fontData = fontOrFamily === "boldsymbol"
+                ? boldsymbol(text, mode, options, classes)
+                : mathnormal(text, mode, options, classes);
             fontName = fontData.fontName;
             fontClasses = [fontData.fontClass];
+        } else if (utils.contains(mathitLetters, text)) {
+            fontName = "Main-Italic";
+            fontClasses = ["mathit"];
```
kevinbarabash: This feels redundant.  Shouldn't the `mathdefault` call below take care of this case?

src/buildCommon.js
```diff
@@ -72,7 +72,7 @@ const makeSymbol = function(
     let symbolNode;
     if (metrics) {
         let italic = metrics.italic;
-        if (mode === "text") {
+        if (mode === "text" || (options && options.font === "mathit")) {
             italic = 0;
```
kevinbarabash: Interesting. The combining of text glyphs we do should have a similar affect, shouldn't it. I guess it still keeps the italic value from the last glyph so it still changing the spacing at the boundaries between font types whereas this won't.

src/buildCommon.js
```diff
@@ -146,16 +145,46 @@ const mathit = function(
...
+    mode: Mode,
+    options: Options,
+    classes: string[],
+): {| fontName: string, fontClass: string |} {
+    if (utils.contains(mathitLetters, value)) {
+        return {
+            fontName: "Main-Italic",
             fontClass: "mathit",
         };
+    } else if (/[0-9]/.test(value.charAt(0))) {
```
kevinbarabash: It'd be nice to have some helper functions for some of these tests, maybe `isLatinLetter`, `isNumber`, etc.

src/buildCommon.js
```diff
@@ -705,11 +734,17 @@ const fontMap: {[string]: {| variant: FontVariant, fontName: string |}} = {
         variant: "italic",
         fontName: "Main-Italic",
     },
+    "mathit": {
+        variant: "italic",
+        fontName: "Main-Italic",
```
kevinbarabash: What were you thinking in terms of renaming fonts?

src/buildMathML.js
kevinbarabash: That should be fine.

docs/supported.md
```diff
@@ -515,14 +515,14 @@ For color definition, KaTeX color functions will accept the standard HTML [pred
...
-|$\sf Ab0$ `\sf Ab0`            |$\texttt{Ab0}$ `\texttt{Ab0}`  |$\mathcal{AB0}$ `\mathcal{AB0}`
-|                               |$\tt Ab0$ `\tt Ab0`            |$\mathscr{AB}$ `\mathscr{AB}`
+|$\mathnormal{Ab0}$ `\mathnormal{Ab0}`|$\textbf{Ab0}$ `\textbf{Ab0}`  |$\textit{Ab}$ `\textit{Ab}`
+|$\textrm{Ab0}$ `\textrm{Ab0}`  |$\bf Ab0$ `\bf Ab0`            |$\it Ab$ `\it Ab`
+|$\rm Ab0$ `\rm Ab0`            |$\bold{Ab0}$ `\bold{Ab0}`|$\Bbb{AB}$ `\Bbb{AB}`
+|$\textnormal{Ab0}$ `\textnormal{Ab0}`|$\boldsymbol{Ab}$ `\boldsymbol{Ab}`|$\mathbb{AB}$ `\mathbb{AB}`
+|$\text{Ab0}$ `\text{Ab0}`      |$\bm{Ab}$ `\bm{Ab}`            |$\frak{Ab0}$ `\frak{Ab0}`
+|$\mathsf{Ab0}$ `\mathsf{Ab0}`  |$\mathtt{Ab0}$ `\mathtt{Ab0}`  |$\mathfrak{Ab0}$ `\mathfrak{Ab0}`
+|$\textsf{Ab0}$ `\textsf{Ab0}`  |$\texttt{Ab0}$ `\texttt{Ab0}`  |$\mathcal{AB0}$ `\mathcal{AB0}`
+|$\sf Ab0$ `\sf Ab0`            |$\tt Ab0$ `\tt Ab0`            |$\mathscr{AB}$ `\mathscr{AB}`
```
edemaine: Any guesses why I'm not seeing these changes on https://deploy-preview-1700--katex.netlify.com/docs/supported.html ? Is there something wrong with our Netlify builds?

src/buildCommon.js
edemaine: Even better (faster) would be to use `hasOwnProperty` on an object with the appropriate keys. We should replace all/most uses of `utils.contains` with this. But not necessarily for this PR.

docs/supported.md
ylemkimon: @edemaine It's because versioned docs are not updated. You can check them in https://deploy-preview-1700--katex.netlify.com/docs/next/supported.html.

edemaine: Of course! Thanks.

Now I see a bug, though: the `0` in `\mathnormal` his rendering in the incorrect font.

image

According to inspect, it's rendering with the HTML `<span class="mord mathcal">0</span>` whereas the letters are rendering as e.g. `<span class="mord mathdefault">A</span>`.

edemaine: Oops, I guess that's exactly what LaTeX does.  Sorry, `\mathnormal` does not do what I thought it did!

```latex
\mathnormal{Ab0}\mathit{Ab0}
```
image

I do get the feeling that there's too much space after the `A` in all other KaTeX examples though...

edemaine: Comparing more with LaTeX, everything seems like a good match. Which your texcmp outputs confirm. 😄 

I do remain confused about what the initial/default mode of LaTeX is. I thought it was `\mathnormal`, but that differs on numerals. Is there a name for it?

ylemkimon:
I'll merge this PR, as I'm planning to completely refactor font generation and selection.

src/buildCommon.js
kevinbarabash: Not sure, but it's something we could investigate.

src/buildCommon.js
kevinbarabash: That's clever.  We'd want to construct the object at load time though.

docs/supported.md
kevinbarabash: `\mathdefault`?  I don't even though if that's a thing or not, but there must be some way of getting back to that mode from inside of `\mathnormal`.

kevinbarabash:
@ylemkimon what kinds of changes are you planning?

PR #1706, opened 2018-09-07
Refactor `buildExpression` to depth-first traverse groups (nodes), fix spacings and \begingroup...\endgroup

### LINKED ISSUE #1108: Spacing inside \href is wrong
`\href` should not affect spacing inside it.

However, `href` builds expression with `isRealGroup = false`, but it doesn't get flattened in its parent `buildExpression` and spacings inside `\href` are not added.

`a \href{.}{+b}`
js-spacing
(Black `a` and purple is KaTeX and black is LaTeX.)

### LINKED ISSUE #1679: \begingroup messes up spacing
this is a standard issue in latex, see e.g.
https://tex.stackexchange.com/questions/85033/colored-symbols/85035#85035
try the following
```
\begin{gathered}
U \oplus V\\
U {\color{red}\oplus} V\\
U\begingroup\color{red}\oplus\endgroup V
\end{gathered}
```
in LaTeX vs KaTeX. in LaTeX the 2nd has wrong spacing, but the third line is okay; in KaTeX both 2nd and 3rd lines are incorrect.

ylemkimon:
Fixes #1108 and fixes #1679. This also allows the implementation of semi-simple (partial, transparent, i.e., doesn't affect spacing around) groups like `\class` and `\cssId` in #1437.

This PR refactors `buildExpression` to depth-first traverse (DFS) groups (nodes), going into `Anchor` and possibly other nodes in the future.

Also, this PR fixes `\begingroup...\endgroup`, which only can be ended by `\endgroup` and is semi-simple group.


src/buildHTML.js
```diff
@@ -88,146 +64,170 @@ export const buildExpression = function(
...
     }
     return node;
 };
 
 // Return math atom class (mclass) of a domTree.
+// If `side` is given, it will get the type of the outermost node at given side.
 export const getTypeOfDomTree = function(
     node: ?HtmlDomNode,
-    side: Side,
+    side: ?Side,
```
ylemkimon: It'd be very nice to get rid of its only usage:
https://github.com/Khan/KaTeX/blob/f71f469d4b3ca97654702a0807e736d176f4c4c5/src/functions/supsub.js#L183-L187

src/buildHTML.js
```diff
@@ -88,146 +64,174 @@ export const buildExpression = function(
...
-                            glueOptions = options.havingStyle(styleMap[node.style]);
-                        }
-                    }
-
-                    groups.push(buildCommon.makeGlue(space, glueOptions));
-                }
+        const result = callback(node, prev.node);
+        if (result) {
+            if (prev.insertAfter) {
+                prev.insertAfter(result);
```
ylemkimon: I've tried inserting before the current node, but it inserts spacing after the`\newline` and probably breaks other things.

src/Parser.js
```diff
@@ -153,7 +153,7 @@ export default class Parser {
         return parse;
     }
 
-    static endOfExpression = ["}", "\\end", "\\right", "&"];
+    static endOfExpression = ["}", "\\endgroup", "\\end", "\\right", "&"];
```
kevinbarabash: Why isn't `]` in this list?  Are the contents of an optional group an expression?

src/Parser.js
```diff
@@ -815,43 +815,51 @@ export default class Parser {
...
-        } else {
-            // Otherwise, just return a nucleus, or nothing for an optional group
-            if (mode) {
-                this.switchMode(mode);
-            }
+            groupEnd = optional ? "]" : "}";
+        } else if (!optional && this.nextToken.text === "\\begingroup") {
+            // A group formed by /begingroup...\endgroup is a semi-simple group,
+            // which doesn't affect spacing in math mode, i.e., is transparent
+            semisimple = true;
```
kevinbarabash: Can you add a reference to the TeXBook or similar resource?

src/buildHTML.js
```diff
@@ -88,146 +64,173 @@ export const buildExpression = function(
...
+        })(i);
     }
+    if (next) {
+        nodes.pop();
+    }
+};
 
-    return groups;
+// Check if given node is a partial group, i.e., does not affect spacing around,
+// like fragment.
```
kevinbarabash: I mis-parsed this comment on the first read.  The intent is that partial groups behave the same as fragment, i.e. they don't affect the spacing around it.

src/buildHTML.js
```diff
@@ -88,146 +64,173 @@ export const buildExpression = function(
...
+const checkPartialGroup = function(
+    node: HtmlDomNode,
+): ?(DocumentFragment<HtmlDomNode> | Anchor) {
```
kevinbarabash: Maybe we should define `type PartialGroup = DocumentFragment<HtmlDomNode> | Anchor`.

src/buildHTML.js
```diff
@@ -88,146 +64,173 @@ export const buildExpression = function(
...
-
-        const right = getOutermostNode(nonSpacesI, "right");
-        if (right.classes[0] === "mbin" &&
-                isBinRightCanceller(nonSpaces[i + 1], isRealGroup)) {
-            right.classes[0] = "mord";
+    }, {node: dummyPrev}, dummyNext);
+
+    traverseNodes(groups, (node, prev) => {
+        const prevType = getTypeOfDomTree(prev);
+        const type = getTypeOfDomTree(node);
```
kevinbarabash: Why don't we need the `side` argument here?

src/buildHTML.js
```diff
@@ -88,146 +64,173 @@ export const buildExpression = function(
...
+            node.classes[0] = "mord";
         }
-
-        const right = getOutermostNode(nonSpacesI, "right");
-        if (right.classes[0] === "mbin" &&
-                isBinRightCanceller(nonSpaces[i + 1], isRealGroup)) {
-            right.classes[0] = "mord";
+    }, {node: dummyPrev}, dummyNext);
+
+    traverseNodes(groups, (node, prev) => {
```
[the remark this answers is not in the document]
kevinbarabash: My preference is for multiple passes with simpler code.

src/buildHTML.js
```diff
@@ -88,146 +64,173 @@ export const buildExpression = function(
...
+    // to avoid processing groups multiple times.
+    if (!isRealGroup) {
+        return groups;
+    }
+
+    let glueOptions = options;
+    if (expression.length === 1) {
+        const node = checkNodeType(expression[0], "sizing") ||
+            checkNodeType(expression[0], "styling");
+        if (!node) {
```
kevinbarabash: Not being able define variables inside `if` blocks is a drag.  It's too bad we can't do:
```
if (const node = checkNodeType(expression[0], "sizing")) {
   ...
} else if (const node = checkNodeType(expression[0], "styling")) {
   ...
}
```
oh well.

src/buildHTML.js
```diff
@@ -88,146 +64,173 @@ export const buildExpression = function(
...
-    ];
+    // Dummy spans for determining spacings between surrounding atoms.
+    // If `expression` has no atoms on the left or right, class "leftmost"
+    // or "rightmost", respectively, is used to indicate it.
+    const dummyPrev = makeSpan([surrounding[0] || "leftmost"], [], options);
+    const dummyNext = makeSpan([surrounding[1] || "rightmost"], [], options);
+
+    // TODO: These code assumes that a node's math class is the first element
+    // of its `classes` array. A later cleanup should ensure this, for
+    // instance by changing the signature of `makeSpan`.
```
kevinbarabash: Thanks for calling this out.

src/buildHTML.js
```diff
@@ -88,146 +64,173 @@ export const buildExpression = function(
...
+    }, {node: dummyPrev}, dummyNext);
+
+    return groups;
+};
+
+// Depth-first traverse non-space `nodes`, calling `callback` with the current and
+// previous node as arguments, optionally returning a node to insert after the
+// previous node. `prev` is an object with the previous node and `insertAfter`
+// function to insert after it. `next` is a node that will be added to the right.
+const traverseNodes = function(
```
kevinbarabash: It might be worth unit testing this.  Also the name is very general but it some logic that's super specific to modifying spacing.  Can you change the name and update the comment to reflect this?

src/buildHTML.js
kevinbarabash: Isn't that what the else branch does?

kevinbarabash: Could you explain the `\newline` situation a bit more?

test/katex-spec.js
```diff
@@ -329,6 +329,34 @@ describe("A group parser", function() {
...
+        const ord = parse[0];
+
+        expect(ord.type).toMatch("ord");
+        expect(ord.body).toBeTruthy();
+        expect(ord.semisimple).toBeTruthy();
+    });
+
+    it("should not affect spacing in math mode", function() {
+        const built = getBuilt`\begingroup x+ \endgroup y`;
+        expect(built).toMatchSnapshot();
```
kevinbarabash: Maybe compare this build tree with that of `x + y`,

test/katex-spec.js
```diff
@@ -2572,6 +2600,11 @@ describe("href and url commands", function() {
             allowedProtocols: [],
         }));
     });
+
+    it("should not affect spacing around", function() {
+        const built = getBuilt`a\href{http://example.com/}{+b}`;
+        expect(built).toMatchSnapshot();
```
kevinbarabash: Snapshot looks good.

src/buildHTML.js
ylemkimon: @kevinbarabash The callback function will be only called on non-partial groups.

test/katex-spec.js
ylemkimon: @kevinbarabash Fixed.

src/buildHTML.js
ylemkimon: @kevinbarabash I've renamed the function to `traverseNonSpaceNodes` and added comments about its usage. We really need a way to unit-test functions.

src/Parser.js
kevinbarabash: It sounds like after we see a `[` then we treat `]` specially.

src/Parser.js
kevinbarabash: Did you determine the behavior by seeing what LaTeX does for different test cases?

src/buildHTML.js
kevinbarabash: That makes sense, thanks.

src/buildHTML.js
kevinbarabash: The comment makes it sound kind of hacky.  Is this how LaTeX resets text-align?  Any thoughts on a better way to do this?

src/Parser.js
edemaine: FWIW, this behavior matches my experience testing with LaTeX. I don't see it documented in The TeXbook. But the TeX source code definitely calls groups started with `\begingroup` "semi-simple groups" (search for `begin_group`), and it definitely does something special with them (search for `semi_simple_group`). But I'm no expert at that code...

ylemkimon: @kevinbarabash I've added a link to stack exchange answer, as it has enough up-votes.

src/buildHTML.js
kevinbarabash: That makes sense, but it's not obvious.  Could you add a comment here before merging this?

kevinbarabash:
> LGTM. Requesting one additional comment about why we don't need to pass the side argument to getTypeOfDomTree on lines 120-121 of buildHTML.js.

I forgot about this.  I guess it wasn't that important.

PR #1707, opened 2018-09-08
 Wrap document fragment where classes, attributes, or styles are applied

ylemkimon:
Fixes issues mentioned in #1618. Also a step toward #1618, separating document fragments from DOM nodes.

### DIFF
--- a/src/buildCommon.js
+++ b/src/buildCommon.js
@@ -430,6 +430,20 @@ const makeFragment = function(
     return fragment;
 };
 
+/**
+ * Wraps group in a span if it's a document fragment, allowing to apply classes
+ * and styles
+ */
+const wrapFragment = function(
+    group: HtmlDomNode,
+    options: Options,
+): HtmlDomNode {
+    if (group instanceof DocumentFragment) {
+        return makeSpan([], [group], options);
+    }
+    return group;
+};
+
 
 // These are exact object types to catch typos in the names of the optional fields.
 export type VListElem = {|
@@ -777,6 +791,7 @@ export default {
     makeLineSpan,
     makeAnchor,
     makeFragment,
+    wrapFragment,
     makeVList,
     makeOrd,
     makeGlue,
--- a/src/functions/arrow.js
+++ b/src/functions/arrow.js
@@ -44,15 +44,19 @@ defineFunction({
         // Build the argument groups in the appropriate style.
         // Ref: amsmath.dtx:   \hbox{$\scriptstyle\mkern#3mu{#6}\mkern#4mu$}%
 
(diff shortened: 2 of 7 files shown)


test/katex-spec.js
```diff
@@ -2386,6 +2386,23 @@ describe("A smash builder", function() {
     });
 });
 
+describe("A document fragment", function() {
+    it("should have paddings applied inside an extensible arrow", function() {
+        const markup = katex.renderToString("\\tiny\\xrightarrow\\textcolor{red}{x}");
+        expect(markup).toContain("x-arrow-pad");
```
kevinbarabash: These should maybe be snapshot tests so it's easier to see the structure of the markup.

ylemkimon: @kevinbarabash But the markup result is too complicated, with a lot of unnecessary parts.

src/functions/arrow.js
```diff
@@ -44,15 +44,19 @@ defineFunction({
         // Build the argument groups in the appropriate style.
         // Ref: amsmath.dtx:   \hbox{$\scriptstyle\mkern#3mu{#6}\mkern#4mu$}%
 
+        // Some groups can return document fragments.  Handle those by wrapping
+        // them in a span.
         let newOptions = options.havingStyle(style.sup());
-        const upperGroup = html.buildGroup(group.body, newOptions, options);
+        const upperGroup = buildCommon.wrapFragment(
+            html.buildGroup(group.body, newOptions, options), options);
```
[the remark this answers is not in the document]
ylemkimon: @kevinbarabash The spacing would not be applied properly, (can be fixed by #1706) and the result would be more complicated.

src/buildCommon.js
```diff
@@ -430,6 +430,20 @@ const makeFragment = function(
...
+/**
+ * Wraps group in a span if it's a document fragment, allowing to apply classes
+ * and styles
+ */
+const wrapFragment = function(
+    group: HtmlDomNode,
+    options: Options,
+): HtmlDomNode {
+    if (group instanceof DocumentFragment) {
+        return makeSpan([], [group], options);
```
edemaine: I guess this is equivalent to
```js
return makeSpan([], group.children, options);
```
? Not sure there are any advantages either way, though.

test/katex-spec.js
kevinbarabash: Makes sense, thanks.

src/functions/arrow.js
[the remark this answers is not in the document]
kevinbarabash: Makes sense, thanks.

PR #1710, opened 2018-09-12
Remove greenkeeper-lockfile

ylemkimon:
Greenkeeper now has built-in support for updating lockfiles: https://blog.greenkeeper.io/announcing-native-lockfile-support-85381a37a0d0.


ylemkimon:
@edemaine Probably, as we’re not using it.

PR #1711, opened 2018-09-12
Add `raw` string group, change URL group parser

ylemkimon:
This adds a `raw` string group parser(`parseStringGroup` with the last argument `true`), which returns the raw string inside a group and allows:
* single character string without braces, e.g., `\command x`, `\command#`
* nested braces, e.g., `\command{a={[b,{c}],d}}`
* delegating argument group parser to individual functions: useful for #1620, #1437, #1698

This also changes URL parser to use it. However, this introduces a regression, `%` (percent sign) without escaping allowed in URLs. But I think benefits outweigh the cost, as:
* Lexer and Parser is much simpler, as it doesn't require defining special cases and duplicating function parsing code
* URL argument can be used without modifying the code, in other places like #1620
* New URL parser was never included in the stable version
* It allows nested braces


ylemkimon:
@ronkok Nice idea! If you don't mind, I'd like give it a try.

> 3. Edit Parser's `parseString` function so that any non-URL groups that contain a % will throw an error. URL groups will not throw the error. That will enable the URL escape.

Instead of throwing an error, the LaTeX behavior seems to be continue parsing to the next line.

ronkok:
> I'd like give it a try.

Please, feel free.

> LaTeX behavior seems to be continue parsing to the next line.

I haven't written any of this idea into code yet. But I want to keep the behavior that @edemaine created for comments. So in an expression like `\frac{a}{b % s}`, the group does not close and it should throw an error. That's my intent. Any way we get there would be okay with me.

ylemkimon:
@ronkok Thank you for the suggestion! I’ve implemented it and it have made code much simpler.

src/Parser.js
```diff
@@ -674,35 +682,81 @@ export default class Parser {
...
-                    firstToken.range(this.nextToken, str));
+        while ((raw && nested > 0) || this.nextToken.text !== groupEnd) {
+            switch (this.nextToken.text) {
+                case "EOF":
+                    throw new ParseError(
+                        "Unexpected end of input in " + modeName,
+                        firstToken.range(lastToken, str));
+                case "%":
+                    if (!raw) { // allow % in raw string group
+                        this.consumeComment();
```
kevinbarabash: I guess that's kind of hard to do that lever level.

src/Parser.js
```diff
@@ -674,35 +682,81 @@ export default class Parser {
...
+                return nextToken;
+            }
         }
         const outerMode = this.mode;
         this.mode = "text";
-        this.expect(optional ? "[" : "{");
+        this.expect(groupBegin);
         let str = "";
         const firstToken = this.nextToken;
+        let nested = 0; // allow nested braces in raw string group
```
kevinbarabash: If it's truly "raw" we might want to allow mismatched braces.  In practice that's probably not all that useful so maybe add a comment describing the limitations of raw arg types.

src/Parser.js
```diff
@@ -674,35 +682,81 @@ export default class Parser {
...
      */
     parseStringGroup(
         modeName: ArgType,  // Used to describe the mode in error messages.
         optional: boolean,
+        raw?: boolean,
     ): ?Token {
-        if (optional && this.nextToken.text !== "[") {
-            return null;
+        const groupBegin = optional ? "[" : "{";
+        const groupEnd = optional ? "]" : "}";
```
kevinbarabash: Good call!

src/Parser.js
```diff
@@ -674,35 +682,81 @@ export default class Parser {
         }
     }
 
+    consumeComment() {
+        // the newline character is normalized in Lexer, check original source
+        while (this.nextToken.text !== "EOF" && this.nextToken.loc &&
+                this.nextToken.loc.getSource().indexOf("\n") === -1) {
```
kevinbarabash: Why not `this.nextToken.text.indexOf("\n") === -1`?

src/types.js
```diff
@@ -12,13 +12,15 @@ export type Mode = "math" | "text";
...
+//   - "raw": A string, allowing single character, percent sign,
+//            and nested braces
 //   - "original": The same type as the environment that the
 //                 function being parsed is in (e.g. used for the
 //                 bodies of functions like \textcolor where the
 //                 first argument is special and the second
 //                 argument is parsed normally)
 //   - Mode: Node group parsed in given mode.
-export type ArgType = "color" | "size" | "url" | "original" | Mode;
+export type ArgType = "color" | "size" | "url" | "raw" | "original" | Mode;
```
kevinbarabash: Now that we have `raw`, I wonder if we should move the parsing of `color`, `size`, and `url` into functions/*.js.

test/katex-spec.js
```diff
@@ -1591,6 +1591,16 @@ describe("A comment parser", function() {
...
 
+    it("should parse comments between subscript and superscript", () => {
+        expect("x_3 %comment\n^2").toParseLike`x_3^2`;
+    });
+
+    it("should parse comments in size and color groups", () => {
+        expect("\\kern{1 %kern\nem}").toParse();
+        expect("\\kern1 %kern\nem").toParse();
+        expect("\\color{#f00%red\n}").toParse();
+    });
```
kevinbarabash: We should probably have some tests for parsing `raw` arg types.  It would be cool if we had a test that defines a function that accepts one `raw` arg and one optional `raw` arg then we could test that including `[]` or `{}` in the raw string works as expected.

src/Parser.js
ylemkimon: @kevinbarabash Lexer normalizes whitespaces to a single space(` `). I think it's better to look at the source, instead of separating whitespace token and newline token.

src/Parser.js
ylemkimon: @kevinbarabash If we allow unmatched braces, it's impossible to determine where the group ends, without forbidding a right brace.

src/Parser.js
ylemkimon: @kevinbarabash Yes, it's context-dependent.

test/katex-spec.js
ylemkimon: @kevinbarabash Currently, it seems there is no way to define a function cleanly in the tests. I've updated existing test case below.

src/Parser.js
kevinbarabash: That makes sense.  It might be worth adding a comment here about that.

test/katex-spec.js
kevinbarabash: Maybe that's something we can do in the future.  I've opened an issue for it https://github.com/Khan/KaTeX/issues/1740.

test/katex-spec.js
```diff
@@ -2532,7 +2532,7 @@ describe("href and url commands", function() {
     });
 
     it("should allow balanced braces in url", function() {
-        const url = "http://example.org/{too}";
+        const url = "http://example.org/{{}t{oo}}";
```
kevinbarabash: Cool!

PR #1721, opened 2018-09-19
Fix tag positioning to live within katex-html parent

edemaine:
`.tag` has `position: absolute`, but this positions it relative to the nearest parent with a non-static `position` (see https://developer.mozilla.org/en-US/docs/Web/CSS/Containing_block).  (My mistake in #1309.)

This PR makes `.katex-display > .katex > .katex-html` that parent.

Before this PR (katex.org), the tag appears in the blue margin:

image

After this PR, the tag appears where it belongs:

image


src/katex.less
```diff
@@ -571,6 +571,7 @@
 
         > .katex-html {
             display: block;
+            position: relative;
```
kevinbarabash: I love easy fixes.

willymcallister:
In katex 0.10 is it possible to render a left-justified equation on its own line and include a \tag{} or \tag*{}?.

edemaine:
@willymcallister Please ask in a new issue.

Issue #428, opened 2016-01-14
Add basic error recovery


xymostech:
Sometimes, LaTeX will find an error but is still able to recover from it, such as the example in #427. The TeXbook talks about this in chapter 27.

This would make KaTeX print errors still, but attempt to recover from them by guessing what the user meant.

kevinbarabash:
https://github.com/Khan/KaTeX/issues/247 is another case of this.

PR #1723, opened 2018-09-21
Refactor Parser

### LINKED ISSUE #538: Improve error message on unsupported command

ylemkimon:
* Moved unsupported command (undefined control sequence) handlings to `parseGroup`
* Changed `parseGivenFunction` to parse only function, remove `parseFunction`, and rename `parseGivenFunction` to `parseFunction`. Moved argument greediness check to `parseFunction`
* Merged `parseImplicitGroup` and `parseGroup`
* Removed `ParsedFunc` and `ParsedArg`
* Moved `parseFunction` call from `parseSymbol` to `parseGroup`
* Throw an error if the group type is unknown: https://github.com/Khan/KaTeX/pull/1620#discussion_r216034688
* Throw "undefined control sequence" error for undefined control sequence (unknown/unsupported command): fixes #538 

This is a step toward #428.


src/Parser.js
```diff
@@ -429,115 +381,38 @@ export default class Parser {
     }
 
     /**
-     * Parses an implicit group, which is a group that starts at the end of a
-     * specified, and ends right before a higher explicit group ends, or at EOL. It
-     * is used for functions that appear to affect the current style, like \Large or
-     * \textrm, where instead of keeping a style we just pretend that there is an
-     * implicit grouping after it until the end of the group. E.g.
-     *   small text {\Large large text} small text again
+     * Same as parseFunction(), except that the base is provided.
```
ylemkimon: TODO: fix comment

src/Parser.js
```diff
@@ -643,29 +498,31 @@ export default class Parser {
...
+            case "math":
+            case "text":
+                return this.parseGroup(name, optional, greediness, undefined, type);
+            case "original":
+            case null:
+            case undefined:
+                return this.parseGroup(name, optional, greediness);
+            default:
+                throw new ParseError(
+                    "Unknown group type as " + name, this.nextToken);
```
ylemkimon: TODO: add test

src/Parser.js
```diff
@@ -812,15 +669,24 @@ export default class Parser {
      * If `mode` is present, switches to that mode while parsing the group,
      * and switches back after.
```
ylemkimon: TODO: add comments

kevinbarabash:
I'm excited about the removal of `ParsedArg` and `ParsedFunc`.  There's a lot going on in this PR so it may take me a while to review.

src/environments/array.js
```diff
@@ -114,11 +114,8 @@ function parseArray(
             }
             break;
         } else if (next === "\\cr") {
-            const cr = parser.parseFunction();
-            if (!cr) {
-                throw new ParseError(`Failed to parse function after ${next}`);
-            }
-            rowGaps.push(assertNodeType(cr, "cr").size);
+            const cr = assertNodeType(parser.parseFunction(), "cr");
+            rowGaps.push(cr.size);
```
edemaine: I guess this simpler error checking is fine because `parseFunction` will return null only if `\cr` is not a defined function, which shouldn't happen.  Ditto below for `\right`.

test/errors-spec.js
```diff
@@ -29,7 +29,7 @@ describe("Parser:", function() {
         it("rejects \\sqrt as argument to ^", function() {
             expect`1^\sqrt{2}`.toFailWithParseError(
                    "Got function '\\sqrt' with no arguments as superscript" +
-                   " at position 2: 1^̲\\sqrt{2}");
+                   " at position 3: 1^\\̲s̲q̲r̲t̲{2}");
```
edemaine: Why did the position here change from the `\` to the `s`? I guess other errors already pointed there.

test/errors-spec.js
```diff
@@ -106,28 +106,17 @@ describe("Parser:", function() {
...
-    describe("#parseArguments", function() {
-        it("complains about missing argument at end of input", function() {
-            expect`2\sqrt`.toFailWithParseError(
-                   "Expected group after '\\sqrt' at end of input: 2\\sqrt");
-        });
-        it("complains about missing argument at end of group", function() {
-            expect`1^{2\sqrt}`.toFailWithParseError(
-                   "Expected group after '\\sqrt'" +
-                   " at position 10: 1^{2\\sqrt}̲");
-        });
```
edemaine: Why did the above two tests disappear? They seem useful, as they repeated elsewhere.

ylemkimon: @edemaine Three tests were duplicate of lines 98 to 114.

test/errors-spec.js
ylemkimon: @edemaine Actually the position is 1-based, and it has changed from the `^` to the `\`, and I think it makes more sense.

ylemkimon:

src/Parser.js
```diff
@@ -986,20 +864,14 @@ export default class Parser {
...
+    parseSymbol(): ?AnyParseNode {
         const nucleus = this.nextToken;
         let text = nucleus.text;
 
-        if (functions[text]) {
-            // If there exists a function with this name, we return the
-            // function and say that it is a function.
-            // The token will be consumed later in parseGivenFunction
-            // (after possibly switching modes).
-            return newFunction(nucleus);
```
kevinbarabash: Getting rid of this sure allowed you clean up a lot of things.

src/environments/array.js
kevinbarabash: `assertNodeType` will throw if `parser.parseFunction()` returns null.

src/functions/environment.js
```diff
@@ -2,29 +2,60 @@
...
         }
-        let name = "";
+        let envName = "";
         for (let i = 0; i < nameGroup.body.length; ++i) {
-            name += assertNodeType(nameGroup.body[i], "textord").text;
+            envName += assertNodeType(nameGroup.body[i], "textord").text;
         }
+
+        if (funcName === "\\begin") {
+            // begin...end is similar to left...right
```
kevinbarabash: This is easier to follow.  I'm glad that you were able to make this look a little more like delimsizing.js.

Issue #897, opened 2017-09-20
used a fixed number of decimal places in output


kevinbarabash:
Some of the numbers we output are needlessly long e.g.
```
<span class="strut bottom" style="height:1.1764999999999999em;vertical-align:-0.481108em;"></span>
```
```
<span style="top:-2.6550000000000002em;">...</span>
```
Maybe we can do `.toFixed(3)` or `.toFixed(4)` when outputting values to keep the output smaller.

kevinbarabash:
@kokopelli314 it sounds like you're on Windows.  I'm not sure what editor you're using but there should be some way to configure which line endings it's using.

grigoriy-reshetniak:
Most of the work related to this issue was done in https://github.com/KaTeX/KaTeX/pull/2460, but it wasn't linked.

PR #1747, opened 2018-10-19
Fixed number of decimal places in styles

kokopelli314:
Fixes [#897](https://github.com/Khan/KaTeX/issues/897).

From initial testing, trimming the numbers in domTree's `toNode` function seems to apply everywhere. It's a little more processor-intensive than rounding numbers before they become strings, but this seems to allow us to only round in this one place and catch everything. 

I set the values to a max of 4 decimal places, but this can trivially be changed to any precision.


kevinbarabash:
@kokopelli314 thanks for fixing those line endings.  I don't see any discernible changes in the screenshots so I think it's okay to replace the existing ones with the new ones from https://circleci.com/gh/Khan/KaTeX/5269#artifacts/containers/0 and  https://circleci.com/gh/Khan/KaTeX/5271#artifacts/containers/0.

@ylemkimon I'm surprised that nothing showed up in the "checks" tab.  Is the service that's running the screenshot app down?

edemaine:
Question: does this make us less accurate compared against LaTeX? Or is there no hope of being identical given layout engines that we shouldn't worry about <1/10000 ems? (Is that the correct epsilon for this PR?)

kokopelli314:
@edemaine Yup, I think 1/10,000 em would be the precision rounding with 4 decimal places.

To put it in perspective, if someone uses KaTeX on a billboard with em's 100 feet tall, the rounding would be to the nearest ~1/10th inch.

It's an interesting visual to imagine, but I'm definitely open to increasing the precision a couple orders of magnitude if that makes more sense.

edemaine:
The mathematician in me is uncomfortable throwing away precision for no reason. I guess the advantage is cleaner input. Is this for humans reading the output, or to speed up HTML parsers down the line? The costs seem to be possible inaccuracy (maybe only visible in printed output?), and maybe decreased performance (though not sure how measurable that is). Sorry, should have posted these questions on the original issue before the PR, but at least now we can try to evaluate them.

edemaine:
The mathematician in me is uncomfortable throwing away precision for no reason. I guess the advantage is cleaner input. Is this for humans reading the output, or to speed up HTML parsers down the line? The costs seem to be possible inaccuracy (maybe only visible in printed output?), and maybe decreased performance (though not sure how measurable that is). Sorry, should have posted these questions on the original issue before the PR, but at least now we can try to evaluate them.

kevinbarabash:
@kokopelli314 sorry abut the duplicate comments, GitHub was having some issues yesterday.

edemaine:
By the way, I'm reminded that TeX lengths have some fixed-point precision. Maybe worth tracking that down and (roughly) matching it?

ylemkimon:
Those long numbers will most come from converting units and I think it's enough to round the number there:
https://github.com/Khan/KaTeX/blob/0b7b4db7fc6649ffc316016b81386f4e1cebaee0/src/units.js#L58-L59

ylemkimon:
@kevinbarabash It seems the bot froze. I've restarted the bot, for future reference:
```bash
now scale bot.katex.org 0
now scale bot.katex.org 1
```

edemaine:
I found what TeX does: see [page 58 of the TeXbook](http://www.ctex.org/documents/shredder/src/texbook.pdf#page=69):
> TeX represents all dimensions internally as an integer multiple of the tiny units called sp. Since the wavelength of visible light is approximately 100 sp, rounding errors of a few sp make no difference to the eye.

According to [our own table](https://github.com/Khan/KaTeX/blob/0b7b4db7fc6649ffc316016b81386f4e1cebaee0/src/units.js#L28), 1sp = 1/65516pt = 1/6551.6em with our `ptPerEm = 10`. I think these means that this PR's 1/10000 tolerance is less than TeX's, so seems like a fine choice. A more radical choice would be to change KaTeX's internal representation to `sp`, but I'm not sure that's worth the effort.

Have you tried measuring the performance hit with `yarn test:perf`?

Do we care that `render` and `renderToString` will use different measurements (only the latter will round)?  @ylemkimon's suggestion of changing `calculateSize` instead would fix this.

kokopelli314:
It looks like `calculateSize` is a better (and much more efficient) place to do the rounding. I'll try making that change this evening, and run performance tests.

@kevinbarabash With the TeX info that @edemaine found, do you think it's still beneficial to make the precision configurable?

kevinbarabash:
@kokopelli314 this is good for now, we can always add it in the future.

ylemkimon:
You'd need to adjust:
https://github.com/Khan/KaTeX/blob/b2432e8ad238dbb723d5f3c83c530fe31fadfee6/src/mathMLTree.js#L177-L178
and
https://github.com/Khan/KaTeX/blob/b2432e8ad238dbb723d5f3c83c530fe31fadfee6/src/mathMLTree.js#L185-L186
.

src/units.js
```diff
@@ -94,5 +94,6 @@ export const calculateSize = function(
             scale *= unitOptions.sizeMultiplier / options.sizeMultiplier;
         }
     }
-    return Math.min(sizeValue.number * scale, options.maxSize);
+    const scaled = Math.min(sizeValue.number * scale, options.maxSize);
+    return +scaled.toFixed(4);
```
edemaine: Maybe we can add the TeXbook reference in a comment to justify the choice of 4?

kokopelli314: Sounds good, added your observation and link in there.

kokopelli314:
@ylemkimon Thanks for pointing out those lines!

It looks like there are some DOM tree elements that don't run their styles through `calculateSize`, meaning they're not rounded now. In particular I've noticed `vlist` and `strut` are not rounded. Would this require a more comprehensive refactor of the `buildCommon.makeVList` function?

src/units.js
kevinbarabash: Instead of converting to fixed units here, I think we should consider doing the conversion in `toNode` and `toMarkup` in domTree.js.  This would mean we wouldn't have to make the changes we're making to mathMLTree.js.

kevinbarabash: @kokopelli314 this would also address the issue you raised:
> It looks like there are some DOM tree elements that don't run their styles through calculateSize, meaning they're not rounded now. In particular I've noticed vlist and strut are not rounded. Would this require a more comprehensive refactor of the buildCommon.makeVList function?

ylemkimon: @kevinbarabash I think it's better to handle rounding when calculating the number. If we do it in `toNode`, we need to detect numbers in string, involving converting the string back to number and may give false positives.

kevinbarabash: That's a good point.  I forgot that they're already strings by time we get to `toNode`.

edemaine: @kevinbarabash The PR was originally done that way, and it was messier; see https://github.com/Khan/KaTeX/pull/1747/files/bcd69c54ad850a8149c247ac5995b7e92c396591

It's a little sad that this approach (still) converts from number to string to number and later back to string, but I guess it'd be hard to return a string here...

kevinbarabash:
@kokopelli314 how much work would it be to round those values that don't go through `calculateSize`?  Is it just `vlist` and `strut` or are there other places we need to add rounding?

kevinbarabash:
@kokopelli314 is it necessary to use`calculateSize` to do the rounding?  Could we do the rounding right inside of `makeVList` and `staticSvg` themselves?

ronkok:
Hey, @kokopelli314, if testing has become frustrating, remember that you can test locally and work things out before pushing a commit to the KaTeX repository. On the command line, you can run: `yarn run test`.

If you are developing on a Windows box like me, that doesn't work so well. Git will occasionally change all my font file line endings to `CRLF`. That's no problem until `eslint` writes 26,000 error messages. So I usually don't run `eslint` locally.

Instead, I write: `yarn run test:jest`, and `yarn run test:flow` locally. For the linter, I usually wait until CircleCI does its run. On a small PR, the number of true lint errors is usually pretty small.

Hope that helps.

kokopelli314:
@kevinbarabash That seems easier. I just made the change, moving the rounding function to utils.js. Does that look good?

@ronkok Thanks for the tip! The 26,000 eslint error messages show up when I run `git commit`. I get around it by running `dos2unix` before each commit, but that has the side effect of marking nearly all files as modified. Know of a way around the commit auto-linting?

ronkok:
`git commit -n -m '<message>'` will commit in a way that evades hooks/pre-commit

src/buildCommon.js
```diff
@@ -608,7 +609,7 @@ const makeVList = function(params: VListParam, options: Options): DomSpan {
             pstrutSize = Math.max(pstrutSize, elem.maxFontSize, elem.height);
         }
     }
-    pstrutSize += 2;
+    pstrutSize = utils.round(pstrutSize + 2);
```
kevinbarabash: I'm glad to see you didn't need to add `utils.round` in too many places.

src/buildCommon.js
```diff
@@ -636,10 +637,10 @@ const makeVList = function(params: VListParam, options: Options): DomSpan {
             }
 
             realChildren.push(childWrap);
-            currPos += elem.height + elem.depth;
+            currPos += utils.round(elem.height + elem.depth);
         }
-        minPos = Math.min(minPos, currPos);
-        maxPos = Math.max(maxPos, currPos);
+        minPos = utils.round(Math.min(minPos, currPos));
+        maxPos = utils.round(Math.max(maxPos, currPos));
```
ylemkimon: It seems rounding errors accumulate over the loop. Could you move the rounding out of the loop?

src/buildCommon.js
```diff
@@ -636,10 +637,10 @@ const makeVList = function(params: VListParam, options: Options): DomSpan {
             }
 
             realChildren.push(childWrap);
-            currPos += elem.height + elem.depth;
+            currPos += utils.round(elem.height + elem.depth);
```
ylemkimon:
```suggestion
            currPos += elem.height + elem.depth;
```
It seems `currPos` is not used directly and rounding is not necessary.

src/buildCommon.js
```diff
@@ -608,7 +609,7 @@ const makeVList = function(params: VListParam, options: Options): DomSpan {
             pstrutSize = Math.max(pstrutSize, elem.maxFontSize, elem.height);
         }
     }
-    pstrutSize += 2;
+    pstrutSize = utils.round(pstrutSize + 2);
     const pstrut = makeSpan(["pstrut"], []);
     pstrut.style.height = pstrutSize + "em";
```
ylemkimon:
```suggestion
     pstrut.style.height = utils.round(pstrutSize) + "em";
```

src/buildCommon.js
ylemkimon:
```suggestion
    pstrutSize += 2;
```

kevinbarabash:
@kokopelli314 I'm going to close this for now due to lack of activity.  Feel free to reopen this PR if you get some time to address the edits that @ylemkimon suggested.

Issue #1682, opened 2018-08-30
Add helpful 404 page


ylemkimon:
Netlify has an ability to set custom 404 pages using `_redirects`.

It'd be nice to have 404 page matching our site style and providing helpful information.

kokopelli314:
I'd like to give this one a go. I can put the PR up shortly for a basic page, and add helpful information from there.

Issue #222, opened 2015-04-20
ReferenceError: require is not defined


franciscop:
When using katex with contrib/auto-render, the file "./splitAtDelimiters" is attempted to be included using the back-end library `require()`, but it doesn't work in the browser.

Code from grunt to do it:

    concat: {
      dist: {
        src: [
          'static/katex.min.js',     // Katex math rendering
          'static/katex-render.js',  // Automatic katex render
          ...
          'static/javascript.js'  // This specific file
        ],
        dest: 'public/javascript.js',
      }
    },

Then when loading javascript.js, the Console says:

![problem](https://cloud.githubusercontent.com/assets/2801252/7224336/cad9e5a0-e76b-11e4-8e8b-eae4a3eda5fa.png)

That line is this:

    var splitAtDelimiters = require("./splitAtDelimiters");

![line](https://cloud.githubusercontent.com/assets/2801252/7224360/0444a578-e76c-11e4-965d-71f434932ddb.png)

My call to the library is by doing this, although I don't think this changes anything

    renderMathInElement(document.body, { delimiters: [{ left: '$$', right: '$$', display: true }] });

So, while this could be solved by messing with that file and just concatenating the previous file, the problem would continue for anyone following the readme.md.

Proposed solution: either include the other file in the same one or edit the instructions about how to include it.

xymostech:
How are you trying to load the auto-render file? You should be including the built auto-render.min.js file, not the un-built auto-render.js file, unless you are including it via browserify/webpack/etc. to serve on your page.

I should be more explicit about including that file. If you use the built file (for example, [from cdnjs](https://cdnjs.cloudflare.com/ajax/libs/KaTeX/0.3.0/contrib/auto-render.min.js)), does it work?

franciscop:
That solved my problem, thank you @xymostech

PR #1748, opened 2018-10-20
Website custom 404 page

kokopelli314:
Fixes [#1682](https://github.com/Khan/KaTeX/issues/1682). Implements one custom 404 page, but Netlify also allows you to make different pages by language, etc.

Happy to make any changes to add/remove content, styling, or math jokes. 😁


kevinbarabash:
Thanks for the PR.  I really like the concept.  I'm have a couple of suggestions though.  Instead of "Page Not Found", maybe it could say something along the lines: "Sorry, we couldn't find the page you were looking for.  Try one of these links below:".  Also, the `\lim_{x \leftarrow 404}...` part could be a lot bigger to fill the space.  Also the text and the math should probably be off blank instead of pure black, maybe #222222.  Here's my attempt at a mock (I wasn't able to update the math color because I was using Preview.app to make these changes):

screen shot 2018-10-20 at 3 25 17 pm

kokopelli314:
Thanks! I've just updated the styles and text similar to your screenshot, with #222 text color. For small screens, the equation will be smaller.

I also made the references to JS and CSS resources absolute (prepending `/` to the `static` path), so that a 404 HTTP status can be returned along with the HTML page, while not breaking for long paths (such as `katex.org/doesnt/exist/sub/path`).

.browserslistrc
```diff
@@ -1,11 +1,11 @@
...
+last 2 versions
+not dead
+
+# Browsers with more than 0.1% global usage, excluding old browsers
+> 0.1%
+not chrome < 45
+not ie < 9
+
+# Currently maintained Node.js versions
+maintained node versions
```
kevinbarabash: @kokopelli314 what changed in this file?

kokopelli314: Looks like it was just a formatting issue. Reverted the file.

kokopelli314:
Quite odd, after reverting the file back to its original state, the deploy/netlify step failed. Looking through the Git blame, I'm not seeing any changes from what I previously had to the original commit, either.

.browserslistrc
ylemkimon: Sorry, this is probably on me 😢 I use Windows and have created file with CRLF line endings.

website/static/static/404.css
```diff
@@ -0,0 +1,53 @@
+body {
+    background: #329894;
+}
+
+.main {
+    display: flex;
```
ylemkimon:
```suggestion
    display: -ms-flexbox;
    display: flex;
```

website/static/static/404.css
```diff
@@ -0,0 +1,53 @@
...
+    flex-direction: column;
```
ylemkimon:
```suggestion
    -ms-flex-direction: column;
    flex-direction: column;
```

website/static/static/404.css
```diff
@@ -0,0 +1,53 @@
...
+    justify-content: space-between;
```
ylemkimon:
```suggestion
    -ms-flex-pack: justify;
    justify-content: space-between;
```

kevinbarabash:
Long term I wonder if we should add some sort of build step for the CSS to auto prefix stuff.

kevinbarabash:
> Quite odd, after reverting the file back to its original state, the deploy/netlify step failed. Looking through the Git blame, I'm not seeing any changes from what I previously had to the original commit, either.

Not sure what was going on there but I don't think it was anything you did.

PR #1750, opened 2018-10-23
Support Unicode double square brackets

ronkok:
This PR adds support for the Unicode characters ⟦ and ⟧. The proposed additions are not variable sized.

I tried and failed to make variable sized delimiters. I believe that new font glyphs will be necessary in order to make them variable sized.

I do not include any function names for the delimiters in this PR. I can't use names like `\ldbrack` or `\lBrack` because those are names for variable sized delimiters. There are a couple of packages that define function names for text-mode brackets, but they don't get delimiter spacing and they look remarkably out of place without that spacing.

So the PR uses only the Unicode characters. This follows the `semantic` package, which provide a non-variable sized delimiter with delimiter spacing. They render it by using a glyph from the `bbold` package. We don't have that glyph so I wrote a macro with a little negative spacing.

### DIFF
--- a/src/macros.js
+++ b/src/macros.js
@@ -803,6 +803,18 @@ defineMacro("\\limsup", "\\DOTSB\\mathop{\\operatorname{lim\\,sup}}\\limits");
 defineMacro("\\liminf", "\\DOTSB\\mathop{\\operatorname{lim\\,inf}}\\limits");
 
 //////////////////////////////////////////////////////////////////////
+// semantic
+
+// The semantic package renders the next two items by calling a glyph from the
+// bbold package. Those glyphs do not exist in the KaTeX fonts. Hence the macros.
+
+defineMacro("\u27e6", "\\mathopen{[\\mkern-3.2mu[}");  // blackboard bold [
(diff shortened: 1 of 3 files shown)


src/macros.js
```diff
@@ -802,6 +802,18 @@ defineMacro("\\notni", "\\html@mathml{\\not\\ni}{\\mathrel{\\char`\u220C}}");
...
 defineMacro("\\liminf", "\\DOTSB\\mathop{\\operatorname{lim\\,inf}}\\limits");
 
+//////////////////////////////////////////////////////////////////////
+// semantic
+
+// The semantic package renders the next two items by calling a glyph from the
+// bbold package. Those glyphs do not exist in the KaTeX fonts. Hence the macros.
+
+defineMacro("\u27e6", "\\mathopen{[\\mkern-3.2mu[}");  // blackboard bold [
+defineMacro("\u27e7", "\\mathclose{]\\mkern-3.2mu]}"); // blackboard bold ]
```
kevinbarabash: These macros are awesome!

PR #1752, opened 2018-10-24
Add \>

ylemkimon:
Add `\>`: fixes partly #164.

`\>` is defined as `\medmuskip`, but uses a fixed value, like `\:`. (Added TODO)


kevinbarabash:
@ylemkimon thanks for adding this. 👍 to closing #164.

Issue #1708, opened 2018-09-10
blackboard letters should only be substituted with \mathbb in math mode


pzinn:
```
\text{ℤ}
```
produces
KaTeX parse error: Can't use function '\mathbb' in text mode at position 1: \̲m̲a̲t̲h̲b̲b̲{Z}
More generally, I think many of the macros defined in macros.js (L327-348) should only work in math mode, i.e., probably should be defined using defineSymbol in symbols.js (only guessing, don't have time to look in more detail).

kevinbarabash:
> More generally, I think many of the macros defined in macros.js (L327-348) should only work in math mode

It sounds like this is already the case since we're throwing a parse error.  What am I missing?

ronkok:
PR #1757 has resolved the issue with Unicode blackboard bold letters. They all now work in either text mode or math mode. There are a few script letters and three Fractur letters that are still implemented via macros that work only in math mode. So I've edited the title of this issue.

PR #1751, opened 2018-10-24
v0.10.0 release

ylemkimon:
Closes #1544.

* Fix multiple occurring hash replacement

Sorry this took so long. As @edemaine said in https://github.com/Khan/KaTeX/issues/1544#issuecomment-429547721, I think it's better to fix bugs later, if any.

PR #1757, opened 2018-10-28
Support blackboard bold in text mode

ronkok:
Support text mode for Unicode characters ℂℍℕℙℚℝℤℎ. Partially addresses issue #1708. 

As suggested by @pzinn, these definitions are now moved from `macros.js` to `symbols.js`. Issue #1708 also requested text mode for certain script and Fractur Unicode characters now defined in `macros.js`. `symbols.js` is not set up to easily accommodate script or Fractur, so I've left those to be done later.

Visual confirmation, given input ℂℍℕℙℚℝℤℎ\text{ℂℍℕℙℚℝℤℎ}:

bb


kevinbarabash:
I'm holding off on merging this until #1751 is merged.

PR #1759, opened 2018-10-29
Fix documentation CORS error

ylemkimon:
Subresource Integrity: The resource 'https://cdn.jsdelivr.net/npm/katex@0.10.0/dist/katex.min.css' has an integrity attribute, but the resource requires the request to be CORS enabled to check the integrity, and it is not. The resource has been blocked because the integrity cannot be enforced.


ylemkimon:
@kevinbarabash Probably it got removed at some point.

PR #1763, opened 2018-10-29
Use reusable executors in CircleCI

ylemkimon:


kevinbarabash:
I assume that this will speed things up for future builds.

PR #1767, opened 2018-10-30
Remove Supported Functions onPageNav style

ylemkimon:
The page is right justified, which is inconsistent with other pages.

CC @ronkok


ylemkimon:
@ronkok Justification was a bad choice of the word, the page is shifted to the right.

Before:
image

After:
image

ronkok:
>  the page is shifted to the right.

Ah, I see now. Thank you for the screen shot. It should be possible to do something that both centers the page and allows a wider main container. I'll look into it and get back to you.

ronkok:
@ylemkimon I have two suggestions:

1. I agree that it would be a good idea to remove the `onPageNav` rule from the `support_table.md` page.
2. On the `supported.md` page, we can re-center the page by changing one line of code from `flex: 0 0 200px;` to `flex: 1 0 200px;`. That would center the page and also retain the slightly wider `mainContainer` that was my original motivation for the rule.

kevinbarabash:
It looks a little weird for the highlight on the right sidebar to extend to the edge of the page.  I wonder if for really wide browsers we could have a max width for the content container (which contains all three columns) and then have `margin: auto` on it to centre it.

ylemkimon:
@ronkok There seems to be no `onPageNav` rule on the `support_table.md`. Also changing `flex-grow` to 1 still shifts the page a bit, which may be bad UX. If you think it's OK, I'll change it 😄

ronkok:
> changing flex-grow to 1 still shifts the page a bit

You're right, now that I look closely. So your preference has a tiny benefit on a wide window and my preference has a tiny benefit on a small window. We're picking nits.

I don't feel strongly about this, so I'll let you choose. I'll approve this PR and you can either merge it or close it just as your prefer.

PR #1781, opened 2018-11-10
Replace Khan/KaTeX with KaTeX/KaTeX

### LINKED ISSUE #1686: Remove `.arcconfig`, `.arclint`, and `lint_blacklist.txt`
https://github.com/Khan/KaTeX/issues/1411#issuecomment-414739113 by @kevinbarabash 
> [W]e use phabricator internally for other projects and may have used it for the initial KaTeX development work. We can get rid of these on this repo.

### LINKED ISSUE #1718: Remove check-node-version.js
The node version is checked when installing dependencies using `engines` field in `package.json`, which is more stricter than version checked by `check-node-version.js`.

ylemkimon:
* Removed RawGit links: [RawGit is shutting down](https://rawgit.com/)
* Removed old files: fixes #1686 and fixes #1718


ylemkimon:
@kevinbarabash Could you check Gitter, Greenkeeper, and Netlify?

edemaine:
Thanks for the heads up about RawGit! FWIW, https://www.jsdelivr.com/rawgit suggests that we could have replaced the links to use jsDelivr, but this is more future-proof.

kevinbarabash:
For gitter I'm going to email them and see if they can transfer the existing chat room.

PR #1783, opened 2018-11-18
Support mhchem in KaTeX core

ronkok:
This PR supports `mhchem` as a core part of KaTeX. I offer this as an alternative to PR #1436.

Towards this purpose, I have modified the code from <https://github.com/mhchem/MathJax-mhchem/blob/master/unpacked/mhchem.js> in several ways, listed below. The first four items enable KaTeX to run `mhchem`. Only item 4 changes mhchem behavior. Item 5 tweaks the vertical alignment of dashed bonds and items 7 through 14 enable the code to pass the KaTeX linter.

1. Changed the interface to `mhchem` so that it runs within KaTeX rather than as an extension to MathJax.
2. Replaced `\rlap` and `\llap` with `\mathrlap` and `\mathllap`.
3. Modified four lines of code in order to change `\raise` to `\raisebox`.
4. Simplified the code for the reaction arrows, since KaTeX provides a full set of extensible arrows and work around hacks are not needed.
5. Slightly raised the vertical alignment of `\tripledash`, to address an [earlier comment]( https://github.com/KaTeX/KaTeX/pull/1436#issuecomment-435615141).
6. Included a preamble that is meant to address licensing issues.
7. Changed indentation to 4 spaces and max line length to 80. `Prettier` did most of this work.
8. Changed each `var` to either `const` or `let`.
9. `eslint` found \~50 unnecessary escapes in RegEx literals. I removed them.
10.   At each `for (key in foo)` statement, I added a condition that guards us from calling a `key` that might be attached to the prototype instead of the object.
11. Five variable declarations were located within a `case`. I moved those declarations above the `switch` statement.
12. `throw` statements now throw a KaTeX \`ParseError\` object instead of plain text.
13. There were a few locations where Prettier did not get line length below the limit. I
    made some minor rearrangements, mostly of comments, at those locations. Some of
    those long lines were `@type` statements apparently there to enable `JSDoc`. Rather than mangle just a few of the long `@type` lines, I removed all the `@type` lines.
14.  I disabled `eslint` at a few lines, (a) to allow an `Array.isArray()` call, (b) to allow a `while (true)` statement, and (c) to avoid messing with some long complicated RegEx literals.

I’ve added only minimal documentation to this repo. For more complete explanation, I have a link to the [mhchem Manual](https://mhchem.github.io/MathJax-mhchem/). I plan to write a PR to that page so that it acknowledges KaTeX as well as MathJax.

The `jest` parser tests are derived from <https://github.com/mhchem/MathJax-mhchem/blob/master/tests/mhchem-tests.js>. I also wrote a few `jest` build tests that will get us coverage of some arrow SVGs.

I’ll add screenshot tests for the reaction arrows and bonds that KaTeX renders slightly differently than MathJax. @ylemkimon, per your [comment](https://github.com/KaTeX/KaTeX/pull/1436#pullrequestreview-174720793), I have revised one arrow SVG so that its "second brush stroke" ([Ref](https://github.com/KaTeX/KaTeX/blob/f4e24183d6d8f302e55bd0d588468489c2ba4e27/src/stretchy.js#L90)) applies to the correct arrow shaft.

I invite comments from everyone, but especially from @mhchem.


ronkok:
RE: the circleci failure: I'm the wrong person to edit a bash script. If someone can correct that, feel free.

danielhollas:
@ronkok It seems that this condition `$CIRCLE_PROJECT_USERNAME != "Khan"` is not longer valid since this repo was moved to its own KaTeX project? So I think you could try simply rewriting "Khan" to "KaTeX" and see whether it resolves it?

ronkok:
> try simply rewriting "Khan" to "KaTeX"

Yes, I also read the error message. But I have no idea where that string should be modified. Rather than have me blunder about trying to fix something that I don't understand, I'll wait for someone who knows what they are doing.

mhchem:
Hi Ron,

it seems you've put a lot of work into this. Some changes are definitely necessary, come are good, some are personal taste, some will make my life more complicated. I would prefer to keep the mhchem core aligned between different MathJax and KaTeX. With things like 'different linebreaks' you make this task very hard. Similar argument for `const` and `let`. I am not sure I want to change that for MathJax too, because that would exclude IE<11. This might not be a priority any more, but that was my reason to use `var`. (Yes, really, MathJax/mhchem runs in IE 6. ☺)

So, again, I'd prefer a (first) PR where the core mhchem can be kept identical between MathJax and KaTeX. This makes it easier to keep mhchem working identical everywhere. From there, we can discuss improvements like shorter lines or `const`.

mhchem:
I'd be okay with that approach (I maintain just one version, you converting it on every change).

I am not so sure about every change. For instance, 9, I find it much less error-prone to say `\.` in a regular expression if I mean a literal dot, even if I could omit the backslash in some instances. The backslash doesn't hurt, decrease performance or anything. 10, I have to look into it. I'd like to do a performance measurement to see how this additional condition in the for loop works. 11 the same, I have to look into it. Contrary to some other coding styles, I'd like to have variable declarations visible on the same screen, I know that the position of the declaration does not change anything. But I am not sure we should do the discussions here at this point in time.

Let's discuss the fundamental idea first: KaTeX using an adapted version of mhchem core, where every change has to be re-done by hand.

kevinbarabash:
@ronkok thanks for exploring an alternative approach for mhchem support.  My preference would be for #1436.  It sounds like it'll be easier to keep it in sync but more importantly keeping it as an extension saves bytes.  My hope is that eventually we end up with more extensions like it and a smaller core so that users can pick and choose what they want.

PR #1784, opened 2018-11-19
Add optional pre-process callback to the auto-renderer.

### LINKED ISSUE #450: Add optional pre-process callback to the auto-renderer
Hello,
It would be great adding an option for rendering inline math in display style (but not centering it). In MathJax, it suffices to add:

```
<script type="text/x-mathjax-config"> 
        MathJax.Hub.Register.StartupHook("TeX Jax Ready",function () { 
          var TEX = MathJax.InputJax.TeX; 
          var PREFILTER = TEX.prefilterMath; 
          TEX.Augment({ 
            prefilterMath: function (math,displaymode,script) { 
              math = "\\displaystyle{"+math+"}"; 
              return PREFILTER.call(TEX,math,displaymode,script); 
            } 
          }); 
        }); 
        </script>
```

Thank you in advance!

AlbertHilb:
Fixes #450.

Cheers

### DIFF
--- a/contrib/auto-render/auto-render.js
+++ b/contrib/auto-render/auto-render.js
@@ -26,11 +26,14 @@ const renderMathInText = function(text, optionsCopy) {
             fragment.appendChild(document.createTextNode(data[i].data));
         } else {
             const span = document.createElement("span");
-            const math = data[i].data;
+            let math = data[i].data;
             // Override any display mode defined in the settings with that
             // defined by the text itself
             optionsCopy.displayMode = data[i].display;
             try {
+                if (optionsCopy.preProcess) {
+                    math = optionsCopy.preProcess(math);
+                }
                 katex.render(math, span, optionsCopy);
             } catch (e) {
                 if (!(e instanceof katex.ParseError)) {
--- a/contrib/auto-render/test/auto-render-spec.js
+++ b/contrib/auto-render/test/auto-render-spec.js
@@ -237,15 +237,15 @@ describe("A delimiter splitter", function() {
 });
 
 describe("Pre-process callback", function() {
-    it("replace `²` with `^2 `", function() {
+    it("replace `-squared` with `^2 `", function() {
         const el1 = document.createElement('div');
-        el1.textContent = 'Circle equation: $x² + y² = r²$.';
+        el1.textContent = 'Circle equation: $x-squared + y-squared = r-squared$.';
         const el2 = document.createElement('div');
         el2.textContent = 'Circle equation: $x^2 + y^2 = r^2$.';
         const delimiters = [{left: "$", right: "$", display: false}];
         renderMathInElement(el1, {
             delimiters,
-            preProcess: math => math.replace(/²/g, '^2'),
+            preProcess: math => math.replace(/-squared/g, '^2'),
         });
         renderMathInElement(el2, {delimiters});
         expect(el1.innerHTML).toEqual(el2.innerHTML);
--- a/contrib/auto-render/test/auto-render-spec.js
+++ b/contrib/auto-render/test/auto-render-spec.js
@@ -4,6 +4,7 @@
 /* global describe: false */
 
 import splitAtDelimiters from "../splitAtDelimiters";
+import renderMathInElement from "../auto-render";
 
 beforeEach(function() {
     expect.extend({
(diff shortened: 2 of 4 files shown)


kevinbarabash:
@AlbertHilb thanks for the PR.  I'm curious, did you make this PR against `Khan/KaTeX` or `KaTeX/KaTeX`? (we're having a minor issue with our circleci setup)

docs/autorender.md
```diff
@@ -99,6 +99,9 @@ in addition to two auto-render-specific keys:
 - `errorCallback`: A callback method returning a message and an error stack
   in case of an critical error during rendering. The default uses `console.error`.
 
+- `preProcess`: A callback function, `(math: string): string`, used to process
```
kevinbarabash: I think `(math: string) => string` would be a bit clearer.

contrib/auto-render/auto-render.js
```diff
@@ -26,11 +26,14 @@ const renderMathInText = function(text, optionsCopy) {
...
         } else {
             const span = document.createElement("span");
-            const math = data[i].data;
+            let math = data[i].data;
             // Override any display mode defined in the settings with that
             // defined by the text itself
             optionsCopy.displayMode = data[i].display;
             try {
+                if (optionsCopy.preProcess) {
+                    math = optionsCopy.preProcess(math);
```
kevinbarabash: Could you add a test for this?

docs/autorender.md
AlbertHilb: Done.

contrib/auto-render/auto-render.js
AlbertHilb: Added. Let me know if it's OK. :-)

AlbertHilb:
HI @kevinbarabash.

> did you make this PR against `Khan/KaTeX` or `KaTeX/KaTeX`?

I think `KaTeX/KaTeX`. How can I verify?

AlbertHilb:
Check failure fixed.

contrib/auto-render/auto-render.js
```diff
@@ -1,6 +1,6 @@
 /* eslint no-console:0 */
 
-import katex from "katex";
+import katex from "../../katex";
```
kevinbarabash: Can you run `npm run build` and see if this ends up including katex within the built file?

contrib/auto-render/test/auto-render-spec.js
```diff
@@ -234,3 +235,19 @@ describe("A delimiter splitter", function() {
...
+describe("Pre-process callback", function() {
+    it("replace `²` with `^2 `", function() {
+        const el1 = document.createElement('div');
+        el1.textContent = 'Circle equation: $x² + y² = r²$.';
+        const el2 = document.createElement('div');
+        el2.textContent = 'Circle equation: $x^2 + y^2 = r^2$.';
+        const delimiters = [{left: "$", right: "$", display: false}];
+        renderMathInElement(el1, {
+            delimiters,
+            preProcess: math => math.replace(/²/g, '^2'),
```
kevinbarabash: Does the test fail without this?  I know we handle some unicode chars, but I forget if we handle superscript chars properly without the preprocessor you've defined here.  It might be worth writing a more extreme preprocessor such as `/-squared/g, '^2'` and then run it on `x-squared + y-squared = r-squared`.  The rationale is that even if we don't handle superscript chars right now, we might at some point in the future which would mean that this test would loose its power at that point.

contrib/auto-render/auto-render.js
AlbertHilb: Yes, you are right. With that change `katex` is included in the `auto-render` build.
I tried leaving `katex` absolute imported here and included in the `test/auto-render-spec` build, but still check fails.
I should confess I have no idea how to fix this. :disappointed:.

contrib/auto-render/test/auto-render-spec.js
AlbertHilb: Yes the test fails without `preProcess`, however I understand your concern. So I update the test in order to be more future-proof.

contrib/auto-render/auto-render.js
AlbertHilb: Moreover, unfortunately, the check doesn't fail locally. So I find a bit difficult to test potential fixes.

AlbertHilb: Hi @kevinbarabash.

Finally I managed to fix the check failure.
`auto-render` is built without including `katex`. During test execution, since we are in a `node.js` environment, it tries to import `katex`as `node.js` module but no such module can be found in the default paths.
My solution consists in adding the root folder to the `NODE_PATH` environment variable before executing `jest`.

PR #1787, opened 2018-11-22
Fix argument font sizing in \fbox and \raisebox, fix font sizing in \TeX, \LaTeX, \KaTeX

### LINKED ISSUE #1786: \TeX is rendered incorrectly with \Huge and \footnotesize things
katex

The size of 'E' does not change. Same problem occurs with 'A' in \LaTeX and \KaTeX.

edemaine:
Fix #1786 (incorrect size of "E" in `\TeX`) by fixing `\raisebox` to keep the current font size.

Previously, `\raisebox` was rendering its second argument as if it were in `\normalsize\textrm{...}`; now, the argument is just as if it were in `\textstyle\textrm{...}`.

I've introduced a new "hbox" argument type that automatically wraps the argument in the equivalent of `\textstyle\textrm{...}`.  I imagine this will be useful to fix for other commands we support with text arguments, but I haven't done a thorough review yet.  I'm open to other approaches, though.

I also fixed some sizing issues with `\LaTeX` and `\KaTeX`.  I reread the actual LaTeX macro, and it's trying to switch to the font size equivalent to math-mode `\scriptstyle` instead of my previous interpretation of `\scriptsize`.  This fixes the size of the A to be relative when in a different font size (e.g. `\large \LaTeX`).  I also had to make `.sizing` (which uses `display:inline-block`) use `width:min-content` to fix our usual Chrome negative-space bug.


edemaine:
Hmm, the sizing in subscripts (I guess with `\scriptstyle`) doesn't match LaTeX anymore.  QuickLaTeX rendering of the screenshot test looks more like the old screenshot, not the new one:

image

I'll need to mess with font sizing then... Presumably this is what happens with `\hbox` instead of `\text`...

edemaine:
OK, this should be ready for review.  Added since original posting:
> I've introduced a new "hbox" argument type that automatically wraps the argument in the equivalent of `\textstyle\textrm{...}`.  I imagine this will be useful to fix for other commands we support with text arguments, but I haven't done a thorough review yet.  I'm open to other approaches, though.

src/functions/styling.js
```diff
@@ -49,15 +49,6 @@ defineFunction({
     },
     mathmlBuilder(group, options) {
         // Figure out what style we're changing to.
-        // TODO(kevinb): dedupe this with buildHTML.js
-        // This will be easier of handling of styling nodes is in the same file.
-        const styleMap = {
-            "display": Style.DISPLAY,
-            "text": Style.TEXT,
```
kevinbarabash: I didn't realize this was defined earlier in this file.  Thanks for cleaning this up.

src/Parser.js
```diff
@@ -527,6 +527,28 @@ export default class Parser {
...
+            case "hbox": {
+                const group = this.parseGroup(
+                    name, optional, greediness, undefined, "text");
+                if (!group) {
+                    return group;
+                }
+                /* "styling" group below suffices to transition to text mode
+                const textGroup = {
+                    type: "text",
+                    mode: this.mode,
```
kevinbarabash: Can we get rid of this commented out code?

src/functions/raisebox.js
```diff
@@ -1,20 +1,20 @@
...
-import * as sizing from "./sizing";
 
 // Box manipulation
 defineFunction({
     type: "raisebox",
     names: ["\\raisebox"],
     props: {
         numArgs: 2,
-        argTypes: ["size", "text"],
+        argTypes: ["size", "hbox"],
```
kevinbarabash: It feels a little weird to introduce a new arg type just to wrap the arg in `\textstyle\textrm{...}`.  Would it be possible to use macro for this?

kevinbarabash:
It would be good to have a test for this to avoid regressions.  A screenshot tests seems like it would be the most efficient way to check that this is working as intended.

src/functions/raisebox.js
edemaine: In LaTeX, `\text` is a macro that enters `\hbox` mode, then attempts to restore sizing and font to match the current values in the surrounding math mode.  So it feels like there are fundamentally two different ways to enter text mode: `\text` which preserves everything and `\hbox` which resets everything.  A lot of AMS macros (such as `\text`) behave like the former, while a lot of TeX and LaTeX macros (such as `\raisebox`) behave like the latter.  Hence the second arg type.

I just reviewed all the functions we currently support for those that use `\hbox` behavior.  Here are the ones I could find:

1. `\verb` already has code to deal with this, which could be simplified with the new `hbox` argument type (though is admittedly pretty simple):

   ```js
           // \verb enters text mode and therefore is sized like \textstyle
           const newOptions = options.havingStyle(options.style.text());
   ```

2. `\fbox` does *not* correctly use hbox style; a change from argument type `text` to `hbox` would fix this bug.  For example, `a_{xyz\fbox{xyz}\boxed{xyz}}` currently renders as 
katex.org rendering but should render as 
QuickLaTeX rendering (and does so with the `hbox` fix).

3. Other macros that we might one day support that would benefit include `\hbox`, `\makebox`, `\mbox`, `\settoheight`.  I'm not sure how important these are, as a lot of other box macros (e.g. `\settowidth`, `\setbox`/`\box`/`\copy`) would be difficult for us to implement at present.

So I'd suggest either changing `\verb` to use the new functionality, or changing `\raisebox` and `\fbox` to match `\verb`.  I have a vague feeling that using a separate arg type would make it easier to avoid bugs (forces you to make a decision).

edemaine: Another argument for keeping this as is is that it's helpful to have this reset code centralized.  For example, I just found some slight discrepancies from LaTeX:

CORRECT: `a_{\bf xyz\fbox{xyz}\raisebox{0pt}{xyz}\boxed{xyz}}` renders with this PR as
KaTeX render in KaTeX vs. QuickLaTeX render in LaTeX.

INCORRECT: `a_{\textbf{xyz\fbox{xyz}\raisebox{0pt}{xyz}\boxed{xyz}}}` renders with this PR as
KaTeX render in KaTeX vs. QuickLaTeX render in LaTeX.

It looks like AMS's `\text` (and `\textbf`) actually redefine the font size of the argument to the script size.  This is a bug for another PR, but I feel like we have a better chance of fixing this across `\fbox` and `\raisebox` (and any other future operations) if they all use common code for shaping an `hbox` argument.

edemaine:
@kevinbarabash I added a simple screenshot test: `\large \LaTeX` and similar macros. To make it look right, I also fixed some sizing issues with `\LaTeX` and `\KaTeX`.  I reread the actual LaTeX macro, and it's trying to switch to the font size equivalent to math-mode `\scriptstyle` instead of my previous interpretation of `\scriptsize`.  This fixes the size of the A to be relative when in a different font size (e.g. `\large \LaTeX`).  I also had to make `.sizing` (which uses `display:inline-block`) use `width:min-content` to fix our usual Chrome negative-space bug.

src/Parser.js
```diff
@@ -528,6 +528,22 @@ export default class Parser {
...
+                const group = this.parseGroup(
+                    name, optional, greediness, undefined, "text");
+                if (!group) {
+                    return group;
+                }
+                const styledGroup = {
+                    type: "styling",
+                    mode: group.mode,
+                    body: [group],
+                    style: "text", // simulate \textstyle
```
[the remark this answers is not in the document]
edemaine: Removing that sizing group (`\normalsize` simulation) was the original point of this PR. In LaTeX, `\raisebox` (or more precisely, `\hbox`) puts its contents in text mode, in the equivalent of `\textstyle`, but it doesn't change the font size (from `\small`/`\large`/etc.).  So this PR corrects KaTeX's behavior to match LaTeX's, which in particular fixes things like `\large \LaTeX` rendering the wrong size of `A`.

src/functions/raisebox.js
edemaine: I agree that it could be nice to define everything using a new function `\hbox`, though it's a bit tricky currently because macros and functions are a bit annoying to mix. We'd have to define `\fbox` and `\raisebox` as macros that call a function `\hbox` and an auxiliary function like `\@fbox` and `\@raisebox` that do the actual rendering (as in `\def\fbox#1{\@fbox{hbox{#1}}}`).

Perhaps we could refactor `hbox` argument type into a separate `defineFunction` mechanism for preprocessing arguments according to macros, like `argMap: [`\hbox{#1}`]` meaning "for my first argument, please give me `\hbox{#1}` instead of `#1`".  This would be a lot more general, and would keep function definitions very convenient, without having to define macros and intermediate functions like `\@fbox`, and separating out the functionality of `hbox`-style "argument reformattings" from other argument types' "argument parsing".

I think this could be done as a pretty clean PR after this one -- it would be a drop-in replacement for `hbox` argument type, so it's OK if we end up using it more.

src/Parser.js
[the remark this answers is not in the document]
kevinbarabash: Got it.  I sometimes get confused about font size vs. `\textstyle`, `\displaystyle`, etc.

edemaine:
@kevinbarabash Thanks for the review!  I've opened issues (#2050 and #2051) for the follow-ups we discussed above.

src/functions/raisebox.js
kevinbarabash: That's an interesting idea.  Definitely much more flexible that what I was suggesting.  I like it.

PR #1791, opened 2018-11-24
xmlClass function

daniel3735928559:
Resurrecting #507, and possibly a part of #1437 in the current architecture (which, by the way, is very pleasant to work with--thanks for your work on that).

These changes might be somewhat naive--I've just added `xmlClass` as new function (as I did in 507), this time by putting a new file in src/functions/. As I understand it there might be some mechanism for doing this from contrib instead (since it is not standard LaTeX), and additionally something enabling it only in certain non-strict/unsafe modes (since there might be XSS concerns), but that suspicion is just from reading various discussions and I could not find any documentation to elaborate. In particular, I don't claim that this is the final form the change should take, but would like to restart this discussion here.


daniel3735928559:
Referencing the discussion on #1792, it is unclear to me how to access the KaTeX internal objects currently used by this code from contrib.  Currently, these are: 

* `defineFunction` and `ordargument` from defineFunction.js
* `buildCommon` from buildCommon.js
* `mathMLTree` from mathMLTree.js
* buildHTML.js
* buildMathML.js

Some of these like `defineFunction` I can expose as `katex.__defineFunction`, which seems consistent with the mhchem example. And indeed, an approach is just to likewise expose all the other functions I need through `katex`. I'm happy to do this, but am skeptical that exposing more fragile internal interfaces is a good idea and so am curious what the recommended approach would be?

(For instance, some, like `ordargument`, I could in principle just reimplement, and some may be accessible by object spelunking, although I didn't identify any obvious paths.)

ylemkimon:
`\class` in #1437 seems to have the same function as `\xmlClass`. We're planning to include it in non-strict mode, but as there is security issues involved and I'm working on refactoring the parser, I'd like to work on those first.

kevinbarabash:
One benefit of implementing this as an extension is as that it side-steps some of the requires of having to check whether the user is using the strict setting or not.

> And indeed, an approach is just to likewise expose all the other functions I need through katex. I'm happy to do this, but am skeptical that exposing more fragile internal interfaces is a good idea and so am curious what the recommended approach would be?

I think in the short term this is probably the way to go.  In the future though, we'd like to:
- have an intermediate representation between the TeX AST and the HTML tree
- standardize our DOM tree possibly using HAST or something very similar

The latter would all you to generate DOM node data structures without having to import our helper functions.

> (For instance, some, like ordargument, I could in principle just reimplement, and some may be accessible by object spelunking, although I didn't identify any obvious paths.)

This seems just as fragile with the added disadvantage of being less explicit about depending on internal data structures.

kevinbarabash:
Closing in preference of #2082.

PR #1792, opened 2018-11-24
Nop function

daniel3735928559:
Add a `\nop` function which creates a zero-width element that doesn't render at all and doesn't affect the spacing of anything else, but does exist in the DOM. As with #1791 this is probably better suited to contrib. 

This is useful (along with `\xmlClass`) in the hack I've done to get a caret working in my [WYSIWYG editor project](https://github.com/daniel3735928559/guppy).  As I maintain that project, it would be useful to have these changes upstream so that I don't have to maintain a modified KaTeX.  They both seem generic enough that they might be useful for others also.  

Alternatively, if there are plans for a plugin system, that could make sense as well.  At any rate, the changes I have to maintain are now relatively contained (no more changes to buildHTML, e.g.), so thanks again for that!


kevinbarabash:
Given the non-standard nature of this command I'd suggest reworking this PR to provide `\nop` as an extension in the contrib folder.  Please see contrib/mhchem as an example.  You'll also need to expose classes from mathMLTree.js in a similar way to how domTree.js classes are exposed on the `katex` object.  Instead of `__domTree` could you add a `__mathMLTree`.  You should then be able to write `\nop` as an extension.

daniel3735928559:
Apologies--After some further Googling, `\hspace{0pt}` actually accomplishes exactly the same thing in standard LaTeX and, in KaTeX, behaves exactly like this extension at the DOM level.  In particular, this PR is entirely redundant with existing functionality, thus I will close and only need #1791. Is that also preferred to be in contrib?

kevinbarabash:
@daniel3735928559 #1791 would be better in contrib as well.

PR #1793, opened 2018-11-24
Remove double encoding in MathML, and unescape escaped text

### LINKED ISSUE #1782: katex.render double encodes HTML entities in MathML
`katex.render` double encodes HTML entities in MathML. For example, the following formula:

```latex
x < y
```

Produces an element with the following MathML:

```html
<math>
  <semantics>
    <mrow><mi>x</mi><mo>&amp;lt;</mo><msup><mi>y</mi><mn>2</mn></msup></mrow>
    <annotation encoding=\"application/x-tex\">x &amp;lt; y^2</annotation>
  </semantics>
</math>
```

`&amp;lt;` is produced, `&lt;` is expected.

`katex.renderToString` is not affected by the issue (`element.innerHTML = katex.renderToString` produces expected output).

edemaine:
Fix #1782 by changing `toNode()` to not call `toText()` which would use `utils.escape()` in some cases.

Awkwardly, if we've already escaped the text, we need to unescape before giving it to `document.createTextNode`.  I couldn't find a simpler way.

Apologies to @ylemkimon if he was already working on this.


src/mathMLTree.js
```diff
@@ -137,7 +137,9 @@ export class TextNode implements MathDomNode {
      * Converts the text node into a DOM text node.
      */
     toNode(): Node {
-        return document.createTextNode(this.toText());
+        // createTextNode escapes given text, so unescape if already escaped.
+        return document.createTextNode(
+            this.needsEscape ? this.text : utils.unescape(this.text));
```
kevinbarabash: Since we're not calling `toText()` which was doing the escaping, why do we need to additional unescaping here?

kevinbarabash:
I'm not sure what's up with the tests.  First the firefox tests were running into an issue and now the chrome tests are having issues.

kevinbarabash:
@edemaine I'm noticing the same issue with the Chrome tests in this PR as with your other PR, #1796.

edemaine:
Replaced by #1813.

PR #1794, opened 2018-11-24
[breaking] trust setting to indicate whether input text is trusted

### LINKED ISSUE #1771: Security setting
We have several features/features in waiting that should probably only be used when the LaTeX code is from a trusted source:
* embedding arbitrary images (#1620, merged) could be used for tracking users 
* arbitrary `class/id` attributes (#1437) could break pages with duplicate `id`s in particular, allow unintended formatting (?)
* arbitrary HTML (#1596) could be a big security hole

I think we need some kind of `security` or `trust` option (like we did with `strict`) before these should be released. The simplest thing would be to have a Boolean `trusted` flag that turns on all of these features. Might we want more granularity?  Maybe `trusted: ['images', 'class']` turns on images + class but not ID and HTML?

Links and embedded images are quite different, so our existing URL restrictions don't suffice. On the other hand, we might want to replace our URL restrictions with this security/trust mechanism...?  For example, we might want to allow images from `http:` but not `file:`, like we can do for links.

edemaine:
Here is a draft of a `trust` setting to indicate whether the input text is trusted (and thus fix #1771), also allowing a function (like we do with `strict`) to depend on the specific function that we're worried about (advanced usage).

I did not add support for an array of trusted things like I suggested in #1771, in order to keep this simple, but happy to add something like that if we want.

I also have not added any tests; wanted to see if people like this interface first.


ylemkimon:
I was thinking of a dictionary like:
```js
{
    allowedProtocols: ["https", "http"],
    allowedClass: /katex-.+/,
    ....
}
````
but I think the function approach is better.

File: docs/options.md [hunk 1]
```diff
@@ -26,6 +26,8 @@ You can provide an object of options as the last argument to [`katex.render` and
   - `"newLineInDisplayMode"`: Use of `\\` or `\newline` in display mode
     (outside an array/tabular environment).  In strict mode, no line break
     results, as in LaTeX.
+- `trust`: `boolean` or `function` (default: `false`). If `false` (do not trust input), prevent any commands like `\includegraphics` that could enable adverse behavior, rendering them instead in `errorColor`. If `true` (trust input), allow all such commands. Provide a custom function `handler(command, ...)` to customize behavior depending on the command and possibly its arguments (e.g. a URL).  A list of such commands:
+  - `"\\includegraphics"`, with URL argument
```
ylemkimon: What do you think of deprecating `allowedProtocols` and allow setting allowed protocols here?

src/functions/includegraphics.js
```diff
@@ -85,6 +85,10 @@ defineFunction({
             alt = alt.substring(0, alt.lastIndexOf('.'));
         }
 
+        if (!parser.settings.isTrusted("\\includegraphics", src)) {
```
ylemkimon: I think we should provide context information as the second argument like:
```
{
    type: "url",
    url: "https://katex.org/image.png",
    protocol: "https",
    ...
}
```
or
```
{
    type: "html",
    class: "my-class",
}
```
to allow setting across commands and prevent mis-parsing the url.

edemaine:
@ylemkimon Sorry for the long delay on looking at this. I've just pushed a new proposal where each call to the `trust` function gets a `TrustContext` which includes the `command` but also e.g. `url` and `protocol` as you suggested.  What do you think?

Here are some sample uses: (which maybe I should add to documentation)
* Forbid specific command: `trust: (context) => context.command !== '\includegraphics`
* Allow specific command: `trust: (context) => context.command === '\includegraphics`
* Allow specific protocol: `trust: (context) => context.protocol === 'http'`
* Forbid specific protocol: `trust: (context) => context.protocol !== 'file'`

I agree that this could replace `allowedProtocols`, if I test for trust in `\url` and \href` as well.  In that case, however, the default `trust` of `false` will forbid **all** `\url` and `\href` calls.  Is that what we want?  Maybe, actually.

ylemkimon:
> Is that what we want? Maybe, actually.

I think we should be as conservative as possible, not allowing anything by default.

edemaine:
@ylemkimon Agreed, safest does seem like a good default. This will break the existing release, though we can give a setting `trust: (context) => ['\\url ', '\\href'].includes(context.command) && ['http', 'https', 'mailto', '_relative'].include s(context.protocol)` (already almost an example in the documentation) that mimics the current trust behavior.

This change breaks a lot of tests, which I still need to fix.  Also we should either remove or deprecate `allowedProtocols`.

I think we should try to settle something soon, because #1842 is a pretty critical bug, so once fixed, we're going to want to release -- but I also think we can't release with `\includegraphics` (already merged) before we have a trust setting.

kevinbarabash:
@edemaine this is a big change.  Thanks for tackling this!  ❤️

docs/security.md
```diff
@@ -9,10 +9,12 @@ Of course, it is always a good idea to sanitize the HTML, though you will need
...
 
-Use `maxSize` option for preventing large width/height visual affronts,
-use `maxExpand` for preventing infinite macro loop attacks, and
-use `allowedProtocols` for preventing certain protocols in `\href`. Please
-refer to [Options](options.md) for more details.
+A variety of options give finer control over the security of KaTeX
+with untrusted inputs; refer to [Options](options.md) for more details.
+* `maxSize` can prevent large width/height visual affronts.
+* `maxExpand` can prevent infinite macro loop attacks.
+* `allowedProtocols` can prevent certain protocols in URLs (e.g., with `\href`)
```
ylemkimon: I think as we're in 0.x (major version zero) stage, we can remove `allowedProtocols` without deprecation.

src/utils.js
```diff
@@ -91,11 +91,21 @@ export const assert = function<T>(value: ?T): T {
     return value;
 };
 
+/**
+ * Return the protocol of a URL, or "_relative" if the URL does not specify a
+ * protocol (and thus is relative).
+ */
+export const urlToProtocol = function(url: string): string {
```
ylemkimon: `protocolFromUrl` or `getProtocol(FromUrl)`

kevinbarabash:
I resolved the merge conflicts which were due to this PR modifying the `/includegraphics` docs and the docs being deleted on `master`.  The resolution was to delete them here as well.  We can add the docs back in the future once we re-enable `/includegraphics` after this PR is merged.

kevinbarabash:
@ylemkimon I can make the change you requested.

kevinbarabash:
Looks like there are some flow errors after the merge.  I'll fix those up too.

docs/options.md
```diff
@@ -31,10 +32,27 @@ You can provide an object of options as the last argument to [`katex.render` and
...
+
+  - `{command: "\\url", url, protocol}`
+  - `{command: "\\href", url, protocol}`
+  - `{command: "\\includegraphics", url, protocol}`
+
+  Here are some sample trust settings:
+
+  - Forbid specific command: `trust: (context) => context.command !== '\\includegraphics'`
+  - Allow specific command: `trust: (context) => context.command === '\\url'`
+  - Allow multiple specific commands: `trust: (context) => ['\\url', '\\href'].includes(context.command)`
```
kevinbarabash: @edemaine I really like these examples.

docs/security.md
```diff
@@ -9,10 +9,11 @@ Of course, it is always a good idea to sanitize the HTML, though you will need
...
 
-Use `maxSize` option for preventing large width/height visual affronts,
-use `maxExpand` for preventing infinite macro loop attacks, and
-use `allowedProtocols` for preventing certain protocols in `\href`. Please
-refer to [Options](options.md) for more details.
+A variety of options give finer control over the security of KaTeX
+with untrusted inputs; refer to [Options](options.md) for more details.
+* `maxSize` can prevent large width/height visual affronts.
+* `maxExpand` can prevent infinite macro loop attacks.
+* `trust` can allow certain commands that are not always safe (e.g., `\includegraphics`)
```
kevinbarabash: I was thinking of changing to something else, but I think we can just re-enable `\includegraphics` once this PR is merged.

src/Parser.js
```diff
@@ -803,7 +793,8 @@ export default class Parser {
                     throw new ParseError(
                         "Undefined control sequence: " + text, firstToken);
                 }
-                result = this.handleUnsupportedCmd();
+                result = this.formatUnsupportedCmd(text);
+                this.consume();
```
kevinbarabash: The `consume` was moved out of `formatUnsupported` to here.

src/Settings.js
```diff
@@ -16,6 +16,26 @@ export type StrictFunction =
...
     ?(boolean | string);
 
+export type TrustContextTypes = {
+    "\\href": {|
+        command: "\\href",
+        url: string,
+        protocol?: string,
+    |},
+    "\\includegraphics": {|
+        command: "\\includegraphics",
```
kevinbarabash: I'm going to leave this in since we want to re-enable `\includegraphics` once this lands.

src/Settings.js
```diff
@@ -16,6 +16,26 @@ export type StrictFunction =
...
+        url: string,
+        protocol?: string,
+    |},
+    "\\url": {|
+        command: "\\url",
+        url: string,
+        protocol?: string,
```
kevinbarabash: While this could be restructured to dedupe the `url` and `protocol` we may have other commands that are insecure in other ways in the future.

src/defineFunction.js
```diff
@@ -21,7 +22,7 @@ export type FunctionHandler<NODETYPE: NodeType> = (
     context: FunctionContext,
     args: AnyParseNode[],
     optArgs: (?AnyParseNode)[],
-) => ParseNode<NODETYPE>;
+) => UnsupportedCmdParseNode | ParseNode<NODETYPE>;
```
kevinbarabash: I had to flip the order to make flow happy.  I'll add a link to the issue in the code.

src/utils.js
kevinbarabash: I'm going to go with `protocolFromUrl` since we use `fooFromBar` in a number of places already.

src/Settings.js
```diff
@@ -146,6 +165,23 @@ class Settings {
...
-}
 
-export default Settings;
+    /**
+     * Check whether to test potentially dangerous input, and return
+     * `true` (trusted) or `false` (untrusted).  The sole argument `context`
+     * should be an object with `command` field specifying the relevant LaTeX
+     * command (as a string starting with `\`), and any other arguments, etc.
+     * If `context` has a `url` field, a `protocol` field will automatically
+     * get added by this function (changing the specified object).
```
kevinbarabash: This comment is great!

kevinbarabash:
I renamed this PR to having [breaking] in the title as a reminder to ourselves to bump the version when publishing.  @ylemkimon I think I've addressed all of your concerns.  Please have another look when you have some time.

edemaine:
Thanks @kevinbarabash for picking up my slack here. (And sorry I've been too busy lately to do it myself.)

Given that `\includegraphics` was blocking only on this, perhaps the resolution should have been to put it back in? Up to you though.

test/katex-spec.js
```diff
@@ -2717,27 +2717,10 @@ describe("href and url commands", function() {
...
-        }));
-    });
-
-    it("should not allow protocols not in allowedProtocols", function() {
-        expect("\\href{javascript:alert('x')}{foo}").not.toParse();
-        expect("\\href{relative}{foo}").not.toParse(new Settings({
-            allowedProtocols: [],
-        }));
-    });
-
```
edemaine: It would be nice to replace these with tests of the corresponding `trust` settings (as in the documentation examples). In general, it would be good to add `trust` tests. Because `trust` violations don't throw errors, you could probably just use snapshot tests... Or test for the appropriate parse node.

kevinbarabash: Good idea.  Will do.

kevinbarabash:
> Given that \includegraphics was blocking only on this, perhaps the resolution should have been to put it back in? Up to you though.

I'd like to keep this PR focused on the new `trust` setting.  Re-enabling `\includegraphics` will require adding screenshots and tests that are unrelated to rest of this PR.  I can create a stacked PR so we can see what those changes look like before this is merged.

edemaine:
Makes sense. We'll have the old commits anyway to guide the small changes needed for `\includegraphics`. I don't think it's necessary to stack PRs.

test/katex-spec.js
kevinbarabash: I've re-added the tests but I had to refactor them a bit to be snapshot tests since `\href` with `trust: false` parses, but produces a different parse tree from when `trust: true` is set.

I've also set the default for the unit tests to `trust: false`.  I can probably remove that setting though since `false` is the default.

PR #1798, opened 2018-11-26
Update lockfile, fix test reporting, throwOnError on `yarn start` test page

ylemkimon:
Minor changes 😄


static/main.js
```diff
@@ -14,7 +14,7 @@ function init() {
     input.addEventListener("input", reprocess, false);
     permalink.addEventListener("click", setSearch);
 
-    const options = {displayMode: true, throwOnError: false};
+    const options = {displayMode: true, throwOnError: true};
```
kevinbarabash: This feels like a breaking change.  What's the rationale for this change?

test/helpers.js
```diff
@@ -163,15 +163,15 @@ export const expectEquivalent = (actual, expected, settings, mode, expand) => {
...
+                `${printExpected(r(expected))} are equivalent`
             : () => {
                 const diffString = diff(expectedTree, actualTree, {
                     expand,
                 });
 
-                return `${mode.Verb} trees of ${printReceived(actual)} and ` +
-                `${printExpected(expected)} are not equivalent` +
+                return `${mode.Verb} trees of ${printReceived(r(actual))} and ` +
+                `${printExpected(r(expected))} are not equivalent` +
```
kevinbarabash: These changes look like they're improving the output of our test output, nice!

static/main.js
ylemkimon: @ylemkimon As `yarn start` is for the debugging, I think it's better to throw on error and debug on the developer console.

Issue #1805, opened 2018-12-19
hphantom not exactly as wide as visible expression


bergwerf:
It appears the `hphantom` does not have exactly the same width as the content that is hidden. To reproduce, put `a\frac{1}{2}=a\\a\hphantom{\frac{1}{2}}=a` in the demo box at katex.org. Replace `hphantom` with `phantom` to see that the normal phantom command does show correct behavior.

| `\hphantom` | `\phantom` |
|--|--|
| ![image](https://user-images.githubusercontent.com/4458965/50190757-c4fdfc00-032a-11e9-84cc-329411c5431c.png) | ![image](https://user-images.githubusercontent.com/4458965/50190778-d941f900-032a-11e9-8b06-4758b0670ce2.png)

ronkok:
Yep, that;s a bug. I think I see the problem. The `\hphantom` should be given a `mord` class, just like what it contains. That would cause the correct spacing to the relation to its right. 

I think I wrote the `\hphantom` code. I'll see if I work something up to fix it.

PR #1806, opened 2018-12-22
Fix \hphantom width

ronkok:
Fixes issue #1805. I also applied the same fix to `\smash`, since it had the same problem.

### DIFF
--- a/src/functions/smash.js
+++ b/src/functions/smash.js
@@ -4,6 +4,7 @@ import defineFunction from "../defineFunction";
 import buildCommon from "../buildCommon";
 import mathMLTree from "../mathMLTree";
 import {assertNodeType} from "../parseNode";
+import {binrelClass} from "./mclass";
 
 import * as html from "../buildHTML";
 import * as mml from "../buildMathML";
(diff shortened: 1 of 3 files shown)


src/functions/smash.js
```diff
@@ -85,10 +86,14 @@ defineFunction({
         // makeVList applies "display: table-cell", which prevents the browser
         // from acting on that line height. So we'll call makeVList now.
```
kevinbarabash: It seems like we could use a `\strut` to achieve this if we supported `\strut`.  Is that a correct way of thinking about this?

ronkok: > screenshots test

Will do.

> use a `\strut`...?

`\strut`'s serve a different purpose. They ensure that the box height does not fall below a minimum. The `table-cell` is used to prevent an out-of-control HTML line height from creating an unintended maximum.

kevinbarabash: I see... a fraction inside a `\hphantom` might be quite tall and makes sure we clamp the height.

ronkok:
Well now, I have thoroughly messed up this feature branch in my local repository. I'm going to close this PR and open a new one from a new branch.

Issue #1795, opened 2018-11-24
\includegraphics{file} fails


edemaine:
@ronkok @ylemkimon There seems to be an issue (on master) with the current raw/url parser setup with `\includegraphics`: namely, `\includegraphics[height=5]{filename}` parses fine, as does `\includegraphics[]{filename}`, but `\includegraphics{filename}` does not parse ("Expected raw group").

We should add a test for this too (ensure that various forms of `\includegraphics` parse).

PR #1807, opened 2018-12-23
Allow \includegraphics to omit the optional argument

ronkok:
Fixes issue #1795.


edemaine:
Thanks @ronkok for fixing this!

PR #1808, opened 2018-12-23
update flow to 0.89

kevinbarabash:
Since #1765 is blocked on a greenkeeper issue I decided to upgrade flow manually.


kevinbarabash:
@edemaine I ran into this issue with a different project and I was able to fix it by forcing jest to run in single-threaded mode since the circleci instances we're running on are single core instances.

PR #1811, opened 2018-12-26
Fixes #1810: jest problem on windows

### LINKED ISSUE #1810: Jest problem on Windows
Since PR #1798 was merged, I can no longer run `yarn test:jest`. I get the error message:
```
yarn run v1.12.3
$ NODE_PATH=./:$NODE_PATH jest
'NODE_PATH' is not recognized as an internal or external command,
operable program or batch file.
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

I'm working on a Windows box, if that matters. `flow` still works.

kevinbarabash:
I forgot that windows doesn't support bash variables.


kevinbarabash:
@ronkok could you try this branch out and see if it fixes the issue you were seeing?

kevinbarabash:
@ylemkimon good suggestion.  I'll do that instead.

kevinbarabash:
I guess I can't rely on `dist/katex.js` being built.  I'm surprised the other path worked, since the "main" field in `package.json` points to `dist/katex.js`.

edemaine:
@kevinbarabash I added a commit, https://github.com/KaTeX/KaTeX/pull/1811/commits/e8ba2f49fd2aaaff208a4c9f5402fe61b1034560, which should fix the problem (still works on Windows).  This should be what the old "adding `.` to path` did (as we `import 'katex'` not `import './'`).

kevinbarabash:
@edemaine thanks for updating this PR.

ylemkimon:
Would it be possible to mock the KaTeX module?

kevinbarabash:
`jest` provides a way to mock modules.  I'll give that a try.

kevinbarabash:
@ylemkimon thanks for merging this.  Sorry for not actually getting around to mocking the module. 😞

PR #1813, opened 2018-12-28
Remove double encoding in MathML via Unicode MathML spaces

### LINKED ISSUE #1782: katex.render double encodes HTML entities in MathML

edemaine:
We used to have a complex mechanic for escaping strings, marking them as `needsEscape = false`, and then not escaping those strings (for combining strings in `\operatorname`).  But this doesn't really work with `render`ing directly to a node, as `document.createTextNode` can't be stopped from escaping.

I've thus removed this mechanic, which required the following changes:

* Switch MathML "smart space" encoding to use Unicode instead of `&LongNames;` (which weren't working with `render` for the same reason).
* ~~Hack our HTML/MathML serializer to not use `String.trim`, which wrecks havoc with emitted Unicode spaces. (See also https://github.com/rayrutjes/diffable-html/issues/14)~~ Now fixed thanks to https://github.com/rayrutjes/diffable-html/pull/15

Now `toText()` doesn't escape, so strings concatenate in unescaped form, and `toHTML()` only does the necessary escaping.  Thus fix #1782.

This PR replaces #1793.


src/mathMLTree.js
```diff
@@ -126,34 +126,32 @@ export class MathNode implements MathDomNode {
...
      * Converts the text node into escaped HTML markup
      * (representing the text itself).
      */
     toMarkup(): string {
-        return this.toText();
+        return utils.escape(this.toText());
     }
 
     /**
      * Converts the text node into an escaped string
```
ylemkimon:
```suggestion
     * Converts the text node into a string
```

src/utils.js
```diff
@@ -95,6 +95,7 @@ export default {
     contains,
     deflt,
     escape,
+    unescape,
```
ylemkimon:
```suggestion
```

test/setup.js
```diff
@@ -8,7 +9,29 @@ import {
...
+// generating extra spaces.
+expect.addSnapshotSerializer({
+    print(val) {
+        const trim = String.prototype.trim;
+        String.prototype.trim = function() { return this; };
+        try {
+            return toDiffableHtml(val);
+        } finally {
+            String.prototype.trim = trim;
+        }
```
ylemkimon: Is there any other library that can do this?

edemaine: We could search.  But also a new version of this library should now do what we want, thanks to https://github.com/rayrutjes/diffable-html/pull/15

PR #1814, opened 2018-12-28
leqno and fleqn support

### LINKED ISSUE #1776: Support for leqno and fleqn rendering options
In katex 0.10+ is it possible to render a left-justified equation on its own line and include a \tag{} or \tag*{}?

edemaine:
Fix #1776 by adding two boolean display-math rendering options to the Settings object:
* `leqno`: Put tags on the left instead of right
* `fleqn`: Flush equations left instead of center

Both are implemented by a CSS class on the `.katex-display` element. Settings seemed to be the right place for these, but feel free to suggest alternatives (e.g. a list of classes to apply to display elements? that seemed a bit weird...).

See [amsmath.pdf](https://mirror.hmc.edu/ctan/macros/latex/required/amsmath/amsmath.pdf) for some documentation of these features.


ylemkimon:
Could you update https://github.com/KaTeX/KaTeX/blob/master/website/pages/index.html and https://github.com/KaTeX/KaTeX/blob/master/website/static/js/index.js (options panel), too?

PR #1825, opened 2019-01-06
Fix nested \dfrac

### LINKED ISSUE #1822: \dfrac and \cfrac are rendered too big inside \xrightarrow
```latex
\dfrac a b
\xrightarrow [2.\, \dfrac c d]
             {1.\, \displaystyle\frac c d}
\xrightarrow [4.\, \normalsize \dfrac c d]
             {3.\,\Large \displaystyle\frac c d}
{ \displaystyle \frac e f }
```
yields 
image
even in the display mode.

`2.\, \dfrac c d` is rendered like `\Large` applied. (2. and 3. have the same size, and `\normalsize` in 4. restore that.)

ronkok:
Adjust the size of a `\dfrac` that is nested inside a `sub` or `sup` style. Fixes #1822.


edemaine:
@ronkok This looks like a nice change! I thought I'd check the example from #1822:

```latex
\dfrac a b
\xrightarrow [2.\, \dfrac c d]
             {1.\, \displaystyle\frac c d}
\xrightarrow [4.\, \normalsize \dfrac c d]
             {3.\,\Large \displaystyle\frac c d}
{ \displaystyle \frac e f }
```

LaTeX renders this as:

image

KaTeX currently renderS like this:

image

Your PR fixes case 2:

image

I'm guessing that the Case 4 mismatch is because LaTeX doesn't do anything with `\normalsize` whereas KaTeX does.  Can you confirm that this is the intended behavior?

kevinbarabash:
`\dfrac`'s behavior is more complicated that I realized.  @edemaine thanks for posting that test case.  That's something we may want to include as a screenshot test.

kevinbarabash:
I don't think case 3 needs to be solved as part of this PR though.  We can open a new issue to deal with case 3 separately.

ronkok:
@edemaine When I wrote this PR, I was focused on case 2 since it seemed to demonstrate the point of the OP's request.  I did not check case 3 (`\Large`) against LaTeX and did not even attempt to match LaTeX behavior in that case. At this point, I would agree with @kevinbarabash that case 3 is a topic for another PR.

@kevinbarabash, I have added case 2 to the screenshot tests and gotten a successful Chrome test. My kludge of getting CircleCI to run the Firefox screenshot did not work this time and I have never mastered Docker. So the the screenshot test shows an error. Sorry.

ylemkimon:
The error is due to screenshot being too tall.

ronkok:
Screenshot tests have now passed, thanks to a hint from @ylemkimon.

PR #1827, opened 2019-01-07
Implemented the actuarial angle notation

perfaram:
This PR adds support for the actuarial angle notation (`\angl`, and `\angln` as a macro for the very common `\angle{n}`).
I've not yet added the functions to `docs/supported.md` because I'm not sure in which section it belongs, if it belongs to an existing section at all. Need help there on this little detail.

Anyway, here what it looks like : 
aanglen

(I'm afraid that the vertical bar is somewhat too thin when rendering at low sizes, but it's the same width as `\overline` so I'm trusting you guys there).


edemaine:
@perfaram Which LaTeX package defined these macros?

perfaram:
@edemaine A bunch of standalone LaTeX macros for this exist on the internet, but my design comes from the `lifecon` package (which is unmaintained, as it seems). 

A currently maintained package providing the \angl is https://vigou3.gitlab.io/actuarialsymbol/

ronkok:
@perfaram This is really cool. You got to this one before I did. 

I don't know which is more commonly used:  `\angl` from the `actuarialangle` package, or `\lcroof` from the `lifecon` package. If you have domain specific knowledge and can tell me that `\lcroof` is more common, then I would be okay implementing it.

If we don't have any such knowledge, then I would be more comfortable using the function that has a maintained package on CTAN. That would be `\angl`.

As to its place in the Supported Functions page, I suggest that we expand the Annotation section to read Annotation and Enclosed Math. Maybe someday we can add `\phase` from the Steinmetz package and some sort of `\longdiv`.

Do you have any deeper knowledge on whether `\angl` or `\lcroof` is more common?

perfaram:
First thing, I must say that I find `\lcroof`'s design (according to its user guide – that is, where the top bar of the angle is thicker than its descender) way more elegant than the uniformly thick angle of `actuarialangle`, which provides `\angl`. In this respect, it may be somewhat treacherous to call the function I added `\angl` and not have it behave like the usual `\angl` but like `\lcroof`. 

Regarding the common use, the newer, maintained package is more often used nowadays (I made a couple of exact-match google and GitHub searches to conclude this). However, a lot of (older ?) people obviously stick with `\lcroof` (at least it seems so, judging by what I see in my uni). 

Therefore, since the two notations are 100% compatible with each other, it would be easy to rename the function I added with a more generic name, make its style ("calligraphic" or uniform-thickness) customisable, and add two macros, `\lcroof` and `\angl`, each rendering the right style.

My suggestion is : 
* `\actuarialangle{hbar thickness}{vbar thickness}{text}`
* `\lcroof{text}` as a macro for `\actuarialangle{0.08em}{0.04em}{text}`
* `\angl{text}` as a macro for `\actuarialangle{0.4pt}{0.4pt}{text}`

If it seems suitable, tell me and I'll do it.

ronkok:
@perfaram First, a comment on the philosophy behind this library: It tries hard to maintain compatibility between KaTeX and LaTeX. I occasionally submit something that I think is an improvement on LaTeX. Those submissions generally do not get approved or merged into master.

Inventing a new function name is not in line with that philosophy. You'll want to use a name from an existing package and it should provide 100% round trip copy-and-paste compatibility with a LaTeX package. You can submit either `\angl` or `\lcroof`, or both, but not a brand new function name.

Then the rendering should be as close as possible to the way it is rendered in LaTeX. So for instance, in `\angl` you will want to match these rendering features:
1. Top line thickness that matches `\hrule`. The *TeXbook* p 221 defines a standard `\hrule` as 0.4pt thick.  KaTeX is based on a 10 pt font, so your CSS can emulate a 0.4pt rule with a 0.04em CSS border.
2. 1mu space before and after the argument. That can be emulated by 0.0556em padding. `actuarialangle` has an option for a 2mu space after the argument, so if you prefer 2mu, I think that stands a good chance of getting approved.
3. A 1mu space to the right of the vertical line. Maybe a  CSS margin would be the easiest way to do this.
4. `\angl` always runs its vertical leg as least as low as the depth of a parenthesis, even if the argument's glyph has no descender. `opentype.js` tells me that the depth of a paren is 0.349em. You will need to check the depth of the argument and make that adjustment when necessary.

That depth must be added both the the TeX-like `depth` datum that KaTeX maintains for its calculations and also to the HTML, since you are using a span border for your rendering. You can do both in one step by using the KaTeX `makeVList` function. Maybe the `\underline` [example](https://github.com/KaTeX/KaTeX/blob/master/src/functions/underline.js#L23-L41) would be a good go-by. You'll need a lower kern and the original group as arguments to `makeVList`.

Now those comments pertain to `\angl`. I agree with you that the thicker top line of `\lcroof` is better. If you submit `\lcroof` and not `\angl`, it will need rendering code just as finicky and precise as `\angl`.

ronkok:
Also, @kevinbarabash will ask you to write unit tests. In `test/katex-spec.js`, I suggest that you should at least ensure that you can expect your new function `toBuild()`. 

Rendering is tested by adding to the `test/screenshotter/ss_data.yaml` file.  The KaTeX [Contributing page](https://github.com/KaTeX/KaTeX/blob/master/CONTRIBUTING.md) describes how to use Docker to create the screenshots. If you, like me, are not a Docker user, let me know and I can show you a trick to create KaTeX screenshots without Docker.

ronkok:
I notice now that you proposed `\actangle` only as a means to define `\angl` and `\lcroof`, not as a function for use by itself. If that technique allows you to accurately render both `\angl` and `\lcroof`, then I think it is an excellent idea.

ronkok:
Correction: the descender of a parenthesis is 0.25em, not 0.349em. I need to look at the glyph in the correct font, which is KaTeX_Main-Regular.

perfaram:
Yup, but what if we're running with a custom, different font ? Hardcoding values such as the depth of a parenthesis seems dodgy to me...

perfaram:
(BTW, thank you VERY much for the help ! And yes, I'll gladly see you trick to avoid using Docker – but in due time, I'm gonna get everything running and pixel-perfect before.

ronkok:
> what if we're running with a custom, different font ?

KaTeX doesn't use any custom different fonts except in cases such as CJK characters that do not exist in the KaTeX fonts. One of the reasons that KaTeX is fast is that it never has to query the browser to find the width or height of a math expression. KaTeX has detailed metrics for every glyph in the KaTeX fonts and it uses this information to set all vertical alignments. It can finish its work and shut down without multiple cycles of render - measure - render - measure.

And yes, this does result in tight coupling between the code and the font. It's not ideal but the speed gained is a valuable trade off.

perfaram:
Quick update !

> I notice now that you proposed `\actangle` only as a means to define `\angl` and `\lcroof`, not as a function for use by itself. If that technique allows you to accurately render both `\angl` and `\lcroof`, then I think it is an excellent idea.

That's it, and so far so good ! I think it makes the first point :

> 1. Top line thickness that matches `\hrule`. The _TeXbook_ p 221 defines a standard `\hrule` as 0.4pt thick.  KaTeX is based on a 10 pt font, so your CSS can emulate a 0.4pt rule with a 0.04em CSS border.

 ...irrelevant, since it the thickness will be set in the macro (building `\angl` and `\lcroof` atop a single `\\actangle` function, that won't be available to the user), but tell me if I misunderstood.


> 2. 1mu space before and after the argument. That can be emulated by 0.0556em padding. `actuarialangle` has an option for a 2mu space after the argument, so if you prefer 2mu, I think that stands a good chance of getting approved.
> 3. A 1mu space to the right of the vertical line. Maybe a  CSS margin would be the easiest way to do this.

These were actually more or less done – I had 'em, but with totally wrong widths. This led me to add two more parameters to my `\\actangle` function – the margins. This is starting to look as heavy as `\\genfrac`.


> 4. `\angl` always runs its vertical leg as least as low as the depth of a parenthesis, even if the argument's glyph has no descender. `opentype.js` tells me that the depth of a paren is 0.349em. You will need to check the depth of the argument and make that adjustment when necessary.

Thanks a lot about this one. It made me rework the whole code actually, and it's cleaner (and actually accurate, I think) now.

perfaram:
Reference (using `actuarialangle`) :
reference
(`a_{\angl{n})\angl{K}\angl{1_{j_j}}}`, so with a 2mu-thick space before the descender)

Current KaTeX implementation, uniform-width-style (matching `\angl`) : 
kateximpl

perfaram:
Okay, I built the macros anyway (not really macros in the end, but the same function label replacement logic used in genfrac.js 
For the time being, I left `\cmangle` (which allows me to directly invoke the underlying, customisable `\actangle` function), but I'll remove it once we're done (are we ? Save for the docker stuff and unit tests ?)

ronkok:
@perfaram I've looked at your code and to me it looks like you did an excellent job. I especially like the way you defined `\\\\actangle` in a way that cannot be written by an end-user. My two review comments are (1) yes, you'll need to remove `\cmangle` from the code, and (2) I'd like to see a couple of specific screenshot tests to test spacing: 
* `a_{\angl{n}` as the numerator and denominator of a fraction,
* `1 + a_{x:\angl{n} =`.

KaTeX is, for me, a retirement hobby. This sort of thing is not my profession and when I read reviews done by @kevinbarabash or @ylemkimon they sometimes turn their attention to issues that I did not know existed. So I will leave final review and approval to others.

Now the screenshotter trick. We can evade Docker by getting CircleCI to do the screenshots for us. In order to do this, you'll need to [log in](https://circleci.com/vcs-authorize/) to CircleCI. If you don't have an account there, you can [sign up](https://circleci.com/signup/) with your GitHub name and password.

In `test/screenshotter/ss_data.yaml`, you can see definitions for screen shots organized under headings like "Accents" or "AccentsText".  Then the `images` folder contains PNG files that have names derived from those headings, like `Accents-chrome.png` and `Accents-firefox.png` The steps you need to take are: 

1. Write a new section into `ss_data.yaml` that will create an example of your work under a heading such as "ActuarialAngle". 
2. Make two copies of an existing png file in the `images` folder and give them the names `ActuarialAngle-chrome.png` and `ActuarialAngle-firefox.png`. These are entirely phony files but are needed for the next step. Put the two bogus files into `test/screenshotter/images/` with the other screenshot files.
3. Stage the files, commit your work, and push to origin.
4. CircleCI will create two screenshot files from your `ss_data.yaml` and will compare them to the phony files that you committed. Naturally, the screenshot tests will fail.
5. Where the CircleCI test list says, "Screenshotter - Chrome", click the "Details" link to its right.
6. In the subsequent CircleCI screen, click on the "Artefacts" tab. You need to be logged in to CircleCI in order to see that tab.
7. The next screen will include a link to the `ActuarialAngle-chrome.png` file that CircleCI created. Click on that link and you'll get a view of the picture. If you're in Chrome, you can then right-click on the picture and download the PNG file. Use it to replace the phony file in your local folder.
8. Repeat steps 5 thru 7 for Firefox and then you will be ready to stage, commit, and push the real screenshots.

ronkok:
@perfaram I'm having second thoughts on a couple of issues. First, it was my idea to rename the "Annotation" section of the Support Table to "Annotation and Enclosed Math". Now it occurs to me that someone out there may have written a deep link to that section, a link which now would be broken. I now think we should leave the heading name alone.

I am also reconsidering the code. Not the correctness, but the style. It is now written in the style of `genfrac.js`, which makes it quite verbose. In `genfrac.js`, there is a reason for that verbosity. KaTeX supplies several fraction functions, allowing an author custom control over style, padding, and line thickness. Some KaTeX fractions are infix functions, so they have to use the `replacewith`  technique to pass values back and forth between `genfrac.js` and `Parser.js`.

But `\angl` and `\lcroof` are not infix functions and the only parameter that changes is the top line thickness. (`\angl` is supposed to have a 1mu left padding.) 

In your first commit, many of the dimensions were hard coded into `katex.less`. By doing that, you wrote the whole thing with much less code. Some of those dimensions needed an adjustment, but I like the brevity of that first commit. KaTeX is already quite a large library and we should not ship more code than necessary.

perfaram:
I understand your concerns. I suggest simply removing `\\\\actangle`, moving back in CSS everything that is not constant across `\lcroof` and `\angl`, and choosing the right settings (bar width, etc) in htmlBuilder.

perfaram:
Done, although I kept using `calculateSize` to translate between `pt` and CSS `em` units instead of hardcoding, for the bar thickness. All the rest moved back in the stylesheet.
I also noticed a spacing issue (vertical spacing, depth-wise), that I fixed. 
Next thing is taking the screenshots, but first, is the current code style fit in your opinion ?

ronkok:
@perfaram I really like what you've done. It looks to me like there are a few lines in `parseNode.js` that could be deleted. And then you're good.

perfaram:
Okay, I think everything's right now 🎉

src/functions/enclose.js
```diff
@@ -34,6 +35,40 @@ const htmlBuilder = (group, options) => {
...
+        inner.depth = Math.max(inner.depth, parenDepth);
+
+        imgShift += parenDepth;
+
+        // Height of the angle computation :
+        // 1. Take inner's height + depth of a parenthesis,
+        //    because we only want to extend this far down.
+        const totalInnerHeight = inner.height + parenDepth;
+        // 2. Account for the hline (for thicker hlines especially)
+        // 3. Add 15/24rd (hey, magic number) of the parenthesis' descender's
```
ylemkimon: Could you add comments about these constants, including `0.4pt`, `0.6pt`, and `0.25` above?

perfaram: See #baf4c6e

src/katex.less
```diff
@@ -547,6 +547,22 @@
         border: 0.04em solid black;   // \fboxrule = 0.4pt
     }
 
+    .actangle-pad {
+        padding-right: @muspace * 2; //2mu
+        padding-left: @muspace;
```
kevinbarabash: We actually don't use `@muspace` nor any of the other "space" constants in this file.  I'd like to actually get rid of `@muspace` et al.  Could you create an internal `@actangle` command that renders the symbol without the padding and then define a macro for the actual command that includes the the padding/margin using `\mkern`?

ylemkimon:
[clabot:check]

jameswasher:
Any progress on this?  I'm happy to help if I can.

perfaram:
Hi to all ! Unfortunately, I don't currently have the time to work on this. I'm not sure how much KaTeX internals have changed since I started this PR, it might be way outdated. **I do encourage @jameswasher or anyone else interested to either take up where I left off, or simply start anew.**

On the other hand, depending on how a project (that makes heavy use of KaTeX) I'm working on progresses, there might be a renewed need for the actuarial angle to be implemented – in which case I'll likely try to get this merged. However, I don't really expect this to happen before a couple of months *at least*.

In conclusion : if you have a need for this to be implemented, please, continue my work or just start a new PR and close this one. If you don't, I'll likely eventually do it, but it might take some time.

perfaram:
[clabot:check]

perfaram:
Just came across #2334 – maybe we should just close this PR ?

ylemkimon:
@perfaram Thank you for the PR!

PR #1833, opened 2019-01-12
Fix invalid UTF-8 in mhchem/README.md

varkor:
This causes issues when trying to use KaTeX with GitHub Pages.

### DIFF
--- a/contrib/mhchem/README.md
+++ b/contrib/mhchem/README.md
@@ -4,7 +4,7 @@ This extension adds to KaTeX the `\ce` and `\pu` functions from the [mhchem](htt
 
 ### Usage
 
-This extension isn't part of core KaTeX, so the script should be separately included. Write the following line into the HTML page’s `<head>`. Place it *after* the line that calls `katex.js`.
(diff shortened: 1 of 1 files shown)


edemaine:
Hmm, upon further study, I see two similar `’` characters in the root `README.md` which has been there for a long time:
```md
* **Print quality:** KaTeX’s layout is based on Donald Knuth’s TeX, the gold standard for math typesetting.
```

We should probably change those too, or change none at all and leave a mix of `'` and `’`.  Either way, I do not see how these would affect GitHub Pages...

varkor:
> We should probably change those too, or change none at all and leave a mix of `'` and `’`.

I've made these consistent too.

> Either way, I do not see how these would affect GitHub Pages...

It wasn't a bare apostrophe: there was some encoding problem with that character (it's only evident in some text editors). GitHub Pages refuses to build any site that has a markdown file that's not correct UTF-8 encoded, so if the repository is used as a submodule in a GitHub Pages site, it will fail to build.

Replacing it with a correctly-encoded apostrophe would also have worked, but as you say, this is more consistent with the other occurrences.

edemaine:
Thanks for the clarification and fix!

ylemkimon:
[clabot:check]

varkor:
[clabot:check]

PR #1841, opened 2019-01-31
Reset to leftmost spacing mode after newline

### LINKED ISSUE #1828: Small alignment issue with minner
The following have a span of class `minner` containing the entire things.
image
However, making a nonempty new line, the line _above_ changes: a span of class `mspace` has created. (`<span class="mspace" style="margin-right: 0.166667em;"></span>`)
image
Differences:
image

Was it intended? I think the new line should not affect the line above.

I checked that the line 57 of cr.js occurs the issue:
```js
const span = buildCommon.makeSpan(["mspace"], [], options);
```

edemaine:
Newlines have class "`mspace newline`" but shouldn't be fully ignored: they should reset the spacing mode back to "leftmost" so as to not accidentally create trailing spaces at end of lines.  Fix #1828.

At least, I'm pretty sure no kerns are generated before `\\`/`\newline`, but it wouldn't hurt to verify that this is indeed the case in (La)TeX.


edemaine:
The failing screenshot test diff shows a possible problem/question with this PR:

image

The test includes the line `a+b+c+{d+\\e}+f+g` where the `\\` isn't supposed to do anything because it's not top-level, but internally there's still an `<span class="mspace newline">` so the new spacing algorithm assumes a new line starts there...

I'm unable to find a good LaTeX reference for correct behavior.  `$$a+b+c+{d+\\e}+f+g$$` just ignores the `\\` (i.e. the old behavior of a space between `+` and `e`), but it ignores top-level `\\` too.  `\begin{align*} a+b+c+{d+\\e}+f+g \end{align*}` throws an error.  `\begin{eqnarray*} a+b+c+{d+\\e}+f+g \end{eqnarray*}` actually breaks the line between `+` and `e`!  Is that the behavior we should be aiming for...?  (I'm not sure how we could actually do it, and even `eqnarray*` throws an error when ` \\` is in a `\frac`.)

Perhaps best would be to detect `\\` in non-topmost environments (where it wouldn't do anything) and throw a (different) error/warning and not generate `<span class="mspace newline">` in this case (or at least prevent the `newline` class).  I'm not quite sure how to detect "topmost" (in the `\newline` function) though... essentially we want to be outside all other "functions" (things that render, as opposed to macros).  Alternatively, we could leave `\newline` alone and detect it in the new spacing algorithm in this PR, but again I don't know how to detect "topmost" there.

kevinbarabash:
@edemaine thanks for raising the issue of `\\` inside groups.  That's a little unfortunate that different environments treat this situation differently.  Ignoring it and logging a warning in non-top-level situations seems reasonable.

edemaine:
It's been a long time, but I thought I'd investigate this further.  MathJax turns `\\` in the middle of any expression into a vertical box stacking, so it locally inserts a newline wherever.  The following patch would reproduce roughly this behavior, though we then get incorrect height computations so things are pretty off.  Also, I don't think we really want to reproduce MathJax behavior that is impossible to do in LaTeX.

```diff
diff --git a/src/katex.less b/src/katex.less
index 546dc7c8..482494f7 100644
--- a/src/katex.less
+++ b/src/katex.less
@@ -30,13 +30,6 @@
         overflow: hidden;
     }

-    .katex-html {
-        /* \newline is an empty block at top level, between .base elements */
-        > .newline {
-            display: block;
-        }
-    }
-
     .base {
         position: relative;
         display: inline-block;
@@ -236,6 +229,9 @@
     .mspace {
         display: inline-block;
     }
+    .mspace.newline {
+        display: block;
+    }

     .llap,
     .rlap,
```

Instead, I've opted for the LaTeX-style behavior of ignoring `\\` in groups, pretty much how the code works now.  I've just extended that policy to the spacing algorithm, treating `\\` as starting a new leftmost surrounding only when `\\` is at the outermost level, using some simple "am I at the root?" detection via parameter passing.  Assuming screenshot tests pass, I think this is ready for review again.

edemaine:
I believe this is ready to go. Can someone review? (I can't because I wrote it.) It'd be good to get this bugfix into the next release (#2287).

edemaine:
Thanks for the review, @ylemkimon!

src/buildHTML.js
```diff
@@ -151,27 +153,31 @@ const traverseNonSpaceNodes = function(
...
+                callback, prev, null, isRoot);
             continue;
         }
 
         // Ignore explicit spaces (e.g., \;, \,) when determining what implicit
         // spacing should go between atoms of different classes
-        if (node.classes[0] === "mspace") {
-            continue;
-        }
-
```
ronkok, line 163: Might I suggest that we keep the `continue statement` and change line 160 to:
~~~
if (node.classes[0] === "mspace" && !(isRoot && node.hasClass("newline")))
~~~

The current PR adds a superfluous `<span class="mspace newline"></span>` element after bins. It doesn't change the rendering, but also doesn't need to be there.

By retaining the `continue`, some of the `if (nonspace)` logic can be deleted.

edemaine: @ronkok Could you elaborate on the problem here? I've been unable to reproduce it, trying both `x+\\y` and `x+y`.  In the former, the `<span class="mspace newline"></span>` is needed to force the line break, and the rendering hasn't changed since the last release.

ronkok: Sorry. I should have been more clear. Given the input: `{d +\\ e}`, this PR inserts two space elements after the "+" sign, one with classes  `mspace newline`, and a second one with the desired spacing. The first element is redundant.

As currently written, this PR will always execute lines 181 - 184. By keeping the `continue` statement, that execution could be evaded and no redundant element inserted.

edemaine: I tried reinserting the `continue`, and it didn't make any difference. Indeed, the issue you point out is in the current release too, so it's old. The only code that executes without the continue is a redefinition of `prev.insertAfter`, so I believe it just affects where the next nonspace token will get placed (if it has any affect at all -- I haven't seen one).

ronkok: Well, it's not important. KaTeX writes a lot of elements. I withdraw my suggestion.

PR #1843, opened 2019-02-04
Fix \KaTeX macro definition in tests

### LINKED ISSUE #1703: \KaTeX of KaTeX and test.tex (texcmp) does not match
mathrm

https://github.com/Khan/KaTeX/blob/f71f469d4b3ca97654702a0807e736d176f4c4c5/src/macros.js#L711-L714

https://github.com/Khan/KaTeX/blob/f71f469d4b3ca97654702a0807e736d176f4c4c5/test/screenshotter/test.tex#L16-L22

edemaine:
Update `test/screenshotter/test.tex`'s definition of `\KaTeX` macro to match the metrics of `src/macros.js`'s definition of `\KaTeX`, and thereby fix #1703. (Thanks @ylemkimon for spotting this! Sorry for the delay in fixing.)

I just changed the `\kern` metrics to match the update from #974, and left the font selection code to match LaTeX's definition of `\LaTeX`.

New texcmp:

image

Here's what it used to look like (from #1703):

image


kevinbarabash:
Looks great!

PR #1845, opened 2019-02-07
\color affects following \right, put array cells in their own groups

### LINKED ISSUE #1844: Colored bracket.
LaTeX renders

$$\ln\left(x\color{red}\right)y$$

as expected. The closing bracket is red. It's the only character that is red.

With KaTeX there is no red bracket. How can I achieve this?

edemaine:
Fix #1844 by giving `leftright` nodes a `rightColor` attribute for how to color the right bracket.  Now `\color` sets the macro `\current@color` in the current environment (in particular, resetting after `\right`), just like `color.sty` does in LaTeX.  This is used to specially pass the current color into any following `\right` and then into the `leftright` parse node.

To confirm LaTeX behavior, here's how it renders the new ColorImplicit test:

image

(I tried `texcmp`, but it doesn't work very well with colored text; suffice to say the letters are also aligned with our test.)

Along the way, I discovered environment leakage from one cell of an array to the next, unlike in LaTeX. This PR also fixes an issue, so `\def`s no longer incorrectly span multiple cells of an array.


edemaine:
Failing screenshot tests show that there's currently environment leakage from one cell of an array to the next, unlike in LaTeX.  I'll need to fix this before this can be reviewed/merged.

edemaine:
Fixed! (Turned out to be really easy.) So this PR also fixes an issue with `\def`s incorrectly spanning multiple cells of an array.

edemaine:
OK, group isolation wasn't quite as easy as I thought, because of the messyness that is `nextToken`.  I fixed most cases, but the very first character after `\begin{matrix}` is not in a group like it should be. Fixing that will be painful... and this PR is definitely an improvement (currently there is no isolation!), so probably worth merging for now.

test/katex-spec.js
```diff
@@ -3066,6 +3066,21 @@ describe("A macro expander", function() {
             .toParseLike`11\sqrt[2]{2}11`;
     });
 
+    it("array cells generate groups", () => {
+        expect`\def\x{1}\begin{matrix}\x&\def\x{2}\x&\x\end{matrix}`
+            .toParseLike`\begin{matrix}1&2&1\end{matrix}`;
+    });
+
+    // TODO: This doesn't yet work; before the environment gets called,
```
kevinbarabash: Could open an issue for this and include a link to the issue from this comment?

src/functions/color.js
```diff
@@ -102,7 +102,13 @@ defineFunction({
     handler({parser, breakOnTokenText}, args) {
         const color = assertNodeType(args[0], "color-token").color;
 
-        // If we see a styling function, parse out the implicit body
+        // Set macro \current@color in current namespace to store the current
+        // color, mimicking the behavior of color.sty.
+        // This is currently used just to correctly color a \right
+        // that follows a \color command.
+        parser.gullet.macros.set("\\current@color", color);
```
kevinbarabash: macros defined within a group get popped as we exit the group right?

src/functions/delimsizing.js
```diff
@@ -145,10 +145,16 @@ defineFunction({
         // \left case below triggers parsing of \right in
         //   `const right = parser.parseFunction();`
         // uses this return value.
+        const color = context.parser.gullet.macros.get("\\current@color");
+        if (color && typeof color !== "string") {
```
kevinbarabash: Do we give `\current@color` a default value somewhere?

test/katex-spec.js
```diff
@@ -3066,6 +3066,21 @@ describe("A macro expander", function() {
             .toParseLike`11\sqrt[2]{2}11`;
     });
 
+    it("array cells generate groups", () => {
+        expect`\def\x{1}\begin{matrix}\x&\def\x{2}\x&\x\end{matrix}`
+            .toParseLike`\begin{matrix}1&2&1\end{matrix}`;
```
kevinbarabash: Interesting... so before this change, the output would've been `\begin{matrix}1&2&2\end{matrix}`?

JMSS-Unknown:
May I ask what is holding up this pull request? I need to be able to colour one side of a bracket without knowing the size of the contents, so I can't use `\big`, etc...

src/functions/delimsizing.js
edemaine: No, the default value is `undefined`, which implicitly translates to `color: inherit` in CSS (i.e., don't change the color, use whatever CSS sets it to).  Note that `rightColor` has a type of `?string`.  I'll add a comment about the meaning of `undefined`.

test/katex-spec.js
edemaine: Yes, exactly!

image

src/functions/color.js
edemaine: Correct. I think this is implied by the current comment `Set macro \current@color in current namespace` but feel free to suggest a better wording.

FWIW, we're already using this mechanism to redefine the meaning of `\\` within array environments. It's quite useful!

test/katex-spec.js
edemaine: Done.

edemaine:
Sorry for the delay; I hadn't realized this was blocked on me. I just implemented @kevinbarabash's comments, so it's ready for (re)review.

src/functions/delimsizing.js
kevinbarabash: Thanks!

test/katex-spec.js
kevinbarabash: Thanks!

PR #1851, opened 2019-02-20
Upgrade dependencies and remove yarn check

ylemkimon:
- Remove Yarn check: it is buggy (https://github.com/yarnpkg/yarn/issues/6427) and will be removed in the future (https://github.com/yarnpkg/rfcs/pull/106)
- Upgrade dependencies


.circleci/config.yml
```diff
@@ -55,13 +55,7 @@ commands:
...
       - run:
           name: Install dependencies
-          # upgrade Yarn before installing dependencies, if needed
-          command: |
-              if ! yarn check; then
-                curl -o- -L https://yarnpkg.com/install.sh | PROFILE=$BASH_ENV bash
-                source $BASH_ENV
-                yarn
-              fi
+          command: yarn
```
edemaine: Should we use `yarn install --frozen-lockfile` instead, as mentioned in https://github.com/yarnpkg/yarn/issues/6427#issuecomment-435831173 ?  I admit I don't really understand the implications either way, but it seems this would force it to respect `yarn.lock` and not drift?

ylemkimon:
@edemaine Making another branch for the v0.10.1 seems like a good way to go. I've modified the PR and added `--frozen-lockfile`.

PR #1852, opened 2019-02-20
Fix \not and \neq on Chrome via new Unicode symbol U+E020

### LINKED ISSUE #1842: \not (and \neq) broken in ~~Chrome 72~~ Safari
$`\neq`$ or $`\not =`$ doesn't work correctly. Please take a look into the documentation too.

edemaine:
Fix #1842


edemaine:
Some screenshots changed, but in an invisible way. Not sure why tests not involving `\not` or `/` changed, but given no discernible difference, I assume all is fine.

edemaine:
Of note: https://deploy-preview-1852--katex.netlify.com/docs/next/supported.html displays the symbol correctly now.

Unfortunately, we do break old documentation, because we're not versioning the fonts on Netlify. 
 E.g. https://deploy-preview-1852--katex.netlify.com/docs/supported.html fails.

kevinbarabash:
Thanks for finding a workaround that works in both Chrome and Safari.

PR #1856, opened 2019-02-21
Release v0.10.1

ylemkimon:
v0.10.1 will be released from [v0.10.1-release](https://github.com/KaTeX/KaTeX/tree/v0.10.1-release), where `\includegraphics` has been reverted.


edemaine:
Many thanks @ylemkimon!

kevinbarabash:
Was there a bug in `\includegraphics`?

kevinbarabash:
I see, it was to get a hot-fix out for the `\neq` issue.  I assume the concern around release `\includegraphics` is that we want to address the security concerns we had.

edemaine:
@kevinbarabash Exactly. I think it would be dangerous to release `\includegraphics` until we have a `trust` (or similar, e.g. #1794) setting that lets people turn it off for untrusted inputs. Not as dangerous as `\cssId` or `\class`, but I think it's still a security issue.  It'd be great to get more feedback/review on #1794 (or another approach if that's preferred) so that releases become easier. 😄

PR #1875, opened 2019-03-16
[TEST]Test on Chrome 73

ylemkimon:
https://github.com/KaTeX/KaTeX/pull/1875/checks?check_run_id=79405876

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -8,10 +8,7 @@ executors:
   chrome:
     docker:
       - image: circleci/node:8
-      - image: selenium/standalone-chrome:2.48.2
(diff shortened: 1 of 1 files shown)


kevinbarabash:
Thanks for updating the chrome selenium image.  It thought it might require changes to our test code since it's a major version bump.

I found a couple of the diffs hard to interpret because the content is rendered using either red or green, but those are also the colors we using for different.  I think everything's okay though.  I was a little surprised by how much the Unicode example changed though it's not enough to move forward with this diff.

Are they really using PI for their versioning scheme or is it just a coincidence that the current version if 3.141.59?

PR #1882, opened 2019-03-21
Improve MathML for fractions

ronkok:
The HTML part of `genfrac.js` has long had some code that adjusts for `\dfrac`, `\tfrac`, etc. This PR applies the same code to the fraction's MathML.


src/functions/genfrac.js
```diff
@@ -11,8 +11,7 @@ import * as html from "../buildHTML";
 import * as mml from "../buildMathML";
 import {calculateSize} from "../units";
 
-const htmlBuilder = (group, options) => {
-    // Fractions are handled in the TeXbook on pages 444-445, rules 15(a-e).
+const adjustStyle = (group, options) => {
```
ylemkimon:
```suggestion
const adjustStyle = (size, originalStyle) => {
```
I think this can take only `group.size` and `options.style` as arguments.

ronkok: Done.

ronkok: I've reverted to the previous arguments because it's too late to fight `flow`. If this is important to you, I'll do it tomorrow.

ronkok: Arguments are now revised as suggested.

src/functions/genfrac.js
```diff
@@ -11,24 +11,29 @@ import * as html from "../buildHTML";
...
+const adjustStyle = (size, originalStyle) => {
     // Figure out what style this fraction should be in based on the
     // function used
-    let style = options.style;
-    if (group.size === "display") {
+    let style = originalStyle;
+    if (size === "display") {
         // Get display style as a default.
         // If incoming style is sub/sup, use style.text() to get correct size.
         style = style.id > 3 ? style.text() : Style.DISPLAY;
```
ylemkimon: Part of previous PR, but
```suggestion
        style = style.id >= style.SCRIPT.id ? style.text() : Style.DISPLAY;
```

ronkok: Done.

src/functions/genfrac.js
```diff
@@ -11,24 +11,29 @@ import * as html from "../buildHTML";
...
     // function used
-    let style = options.style;
-    if (group.size === "display") {
+    let style = originalStyle;
+    if (size === "display") {
         // Get display style as a default.
         // If incoming style is sub/sup, use style.text() to get correct size.
-        style = style.id > 3 ? style.text() : Style.DISPLAY;
-    } else if (group.size === "text" &&
+        style = style.id > Style.SCRIPT.id ? style.text() : Style.DISPLAY;
```
ylemkimon: Shouldn't this be `>=`?

ronkok: Done.

ronkok:
@ylemkimon Thank you for the rapid reviews. I've updated the page at [webmathcomparison.net](http://webmathcomparison.net/). It's been fun to see improvement happen so quickly.

PR #1888, opened 2019-03-22
[TEST]Test fonts

ylemkimon:


kevinbarabash:
@ylemkimon I tried pushing a change to update this PR with the latest changes from https://github.com/KaTeX/katex-fonts/pull/56 but ran into a permission error.  Could you update this PR when you have a chance?

PR #1889, opened 2019-03-22
Improve MathML for \coloneqq, \dblcolon, \eqcolon, and \eqqcolon

ronkok:
The four functions in the title all have an equivalent at a single Unicode code point. I exploited that to improve the MathML for these four functions.

I don't know if MathML has any way to create ligatures on the fly. If I find such a thing, I'll work on more of the coloneq series.

@edemaine, I know that you are fond of macros. ~~I tried, but could not find a way to make a macro do the work in this PR.~~


src/macros.js
```diff
@@ -731,13 +725,8 @@ defineMacro("\\coloneq",
 // \providecommand*\Coloneq{\dblcolon\mathrel{\mkern-1.2mu}\mathrel{-}}
 defineMacro("\\Coloneq",
     "\\mathrel{\\dblcolon\\mathrel{\\mkern-1.2mu}\\mathrel{-}}");
-// \providecommand*\eqqcolon{=\mathrel{\mkern-1.2mu}\vcentcolon}
-defineMacro("\\eqqcolon", "\\mathrel{=\\mathrel{\\mkern-1.2mu}\\vcentcolon}");
 // \providecommand*\Eqqcolon{=\mathrel{\mkern-1.2mu}\dblcolon}
 defineMacro("\\Eqqcolon", "\\mathrel{=\\mathrel{\\mkern-1.2mu}\\dblcolon}");
-// \providecommand*\eqcolon{\mathrel{-}\mathrel{\mkern-1.2mu}\vcentcolon}
-defineMacro("\\eqcolon",
-    "\\mathrel{\\mathrel{-}\\mathrel{\\mkern-1.2mu}\\vcentcolon}");
```
ylemkimon: We have a macro `\\html@mathml`, which renders the first argument to HTML and the second to MathML. For example, you can do `\\html@mathml{\\mathrel{\\mathrel{-}\\mathrel{\\mkern-1.2mu}\\vcentcolon}}{\\mathrel{\\char`\u2239}}`.

ronkok: Yes, I can use `\html@mathml` to define MathML  as `\mathrel{\u2254}`. But then we also have a macro:

```
defineMacro("\u2254", "\\coloneqq");
```

That looks to me like an infinite loop.  In order to avoid the infinite loop, I put the code into the function section instead. Is there some macro trick that avoids the infinite loop?

ylemkimon: @ronkok AFAIK, using `\char[backtick]` won't expand the character after it.

ronkok: @ylemkimon Once again, you are quite right.

There is a very small penalty to using `\char`. The MathML we seek for `\coloneqq` is `<mo>≔</mo>`. Using `\char`, the MathML that we get is 
```
<mo>
    <mi mathvariant="normal">≔</mi>
</mo>
```

The second version renders just fine in Firefox. I consider it to be acceptable and I will change this PR to the macro version that you suggest. Thank you for the help.

ronkok:
Okay, This PR has been redone with macros and is ready for review.

src/macros.js
```diff
@@ -732,12 +735,15 @@ defineMacro("\\coloneq",
...
+    "\\mathrel{=\\mathrel{\\mkern-1.2mu}\\vcentcolon}}" +
+    "{\\mathop{\\char\"2255}}"); // ≕
 // \providecommand*\Eqqcolon{=\mathrel{\mkern-1.2mu}\dblcolon}
 defineMacro("\\Eqqcolon", "\\mathrel{=\\mathrel{\\mkern-1.2mu}\\dblcolon}");
 // \providecommand*\eqcolon{\mathrel{-}\mathrel{\mkern-1.2mu}\vcentcolon}
-defineMacro("\\eqcolon",
-    "\\mathrel{\\mathrel{-}\\mathrel{\\mkern-1.2mu}\\vcentcolon}");
+defineMacro("\\eqcolon", "\\html@mathml{" +
+    "\\mathrel{\\mathrel{-}\\mathrel{\\mkern-1.2mu}\\vcentcolon}}" +
+    "{\\mathop{\\char\"2239}}");
```
ylemkimon: Are they all `\mathop` (`<mo>`)?

ronkok: The `\mathop` produces the MathML `<mo>` that we want. I tried writing `\mathrel` into the macro  but the resulting MathML did not include a `<mo>`. So, yes, all the scripts include a `\mathop`.

Issue #1800, opened 2018-12-02
Error with subscript or superscript


JMSS-Unknown:
Typing `\log_{10}\left(x\right)` into the preview at katex.org results in the error
`Invalid markup: Incorrect number of children for <msub/> tag.`
Same with
`\log^{10}\left(x\right)` 

I am using Firefox 63.0.3 on Windows. The error does not occur on Chrome or Edge.

kevinbarabash:
I tried reproducing these using the MacOS version of Firefox but wasn't able to.  Could you paste the MathML markup that you're seeing when you get these errors?

JMSS-Unknown:
For `\log_{10}\left(x\right)`
```html
<span class="katex-mathml">
   <math>
      <semantics>
         <mrow>
            <msub>
               <mi>log</mi>
               <mo>⁡</mo>
               <mn>10</mn>
            </msub>
            <mrow>
               <mo fence="true">(</mo>
               <mi>x</mi>
               <mo fence="true">)</mo>
            </mrow>
         </mrow>
         <annotation encoding="application/x-tex">\log_{10}\left(x\right)</annotation>
      </semantics>
   </math>
</span>
```

JMSS-Unknown:
It seems to be caused by the empty `mo` tag.
I extracted it to a new page to get:
```html
<!DOCTYPE html>

<html>
<head>
<meta charset="utf-8">
</head>
<body>
Equation:
<span class="katex-mathml">
   <math>
      <semantics>
         <mrow>
            <msub>
               <mi>log</mi>
               <mo>⁡</mo>
               <mn>10</mn>
            </msub>
            <mrow>
               <mo fence="true">(</mo>
               <mi>x</mi>
               <mo fence="true">)</mo>
            </mrow>
         </mrow>
         <annotation encoding="application/x-tex">\log_{10}\left(x\right)</annotation>
      </semantics>
   </math>
</span>
</body>
</html>
```

With the `mo` tag I get `invalid-markup`, without it the `log` renders correctly.

edemaine:
I don't expect that the `mo` tag is actually empty.  It should have an invisible character, U+2061 (`&ApplyFunction;`), inside it.  This is intentional (though arguably it should be after the `<msub>`...).

However, I'm also unable to reproduce the error, using Firefox 63.0.3 on Windows.  Could it be caused by a plugin?  Can you try in a private window?

JMSS-Unknown:
[...]

PR #1890, opened 2019-03-22
Fix ApplyFunction character

ronkok:
First, some background: In version 3.0 of W3C's MathML spec, they recommend placement of an invisible character, `&ApplyFunction;` (U+2061) between a function name and the function's argument. Ref: [MathML spec section 3.2.5.5](https://www.w3.org/TR/MathML3/chapter3.html#presm.invisibleops) This is done for accessibility reasons. The invisible character gives a hint to a screen reader that it should say something like "f of x" instead of "f times x".  

KaTeX does insert that character, but it did so incorrectly for functions with a subscript, such as `\lim_y x`. 

This PR revises the code such that the invisible character is omitted for any function name with a subscript. This matches MathJax behavior.  When the invisible character is omitted the function name gets a `<mo>` tag instead of a `<mi>` tag. This also matches MathJax behavior.

@ylemkimon Thank you for the expedited review on PR #1886. I ask for expedited review on this PR as well. It is also a prerequisite for `\operatorname*`.

### DIFF
--- a/src/functions/mclass.js
+++ b/src/functions/mclass.js
@@ -99,6 +99,7 @@ defineFunction({
             mode: baseArg.mode,
             limits: true,
             alwaysHandleSupSub: true,
+            parentIsSupSub: false,
             symbol: false,
             suppressBaseShift: funcName !== "\\stackrel",
             body: ordargument(baseArg),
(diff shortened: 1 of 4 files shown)


ronkok:
Also, this PR fixes isssue #1800.

PR #1891, opened 2019-03-23
Load fonts before running screenshotter

ylemkimon:
Reduces screenshotter errors due to unloaded fonts, producing more deterministic screenshots.


test/screenshotter/test.html
```diff
@@ -31,6 +31,21 @@
...
+            SansSerif:7,Size1:1,Size2:1,Size3:1,Size4:1,Typewriter:1};
+        for (var f in fonts) {
+          var ff = fonts[f];
+          if (ff & 1) document.fonts.load("10px KaTeX_" + f);
+          if (ff & 2) document.fonts.load("bold 10px KaTeX_" + f);
+          if (ff & 4) document.fonts.load("italic 10px KaTeX_" + f);
+          if (ff & 8) document.fonts.load("bold italic 10px KaTeX_" + f);
+        }
+        document.fonts.load("10px Mincho");
+        document.fonts.load("10px Batang");
```
ylemkimon: `10px` is arbitrarily selected value.

test/screenshotter/test.html
```diff
@@ -31,7 +31,35 @@
...
+        if (!document.fonts || !document.fonts.load) {
+          callback();
+          return;
+        }
+
+        var results = [];
+        function load_font(name, style) {
+          // 10px is arbitrarily selected value
+          results.push(document.fonts.load((style || "") + " 10px " + name));
+        }
```
kevinbarabash: Neat.  This is pretty straight forward.  If we find this doesn't work when we start testing on more browsers we can have a look at https://github.com/typekit/webfontloader.

test/screenshotter/test.html
```diff
@@ -31,7 +31,35 @@
...
+
+        var fonts = {AMS:1,Caligraphic:3,Fraktur:3,Main:15,Math:12,Script:1,
+            SansSerif:7,Size1:1,Size2:1,Size3:1,Size4:1,Typewriter:1}; // see below
+        for (var f in fonts) {
```
kevinbarabash: It would be nice if there was a way to know which fonts needed to be loaded for a particular test case so we don't have to always load all of them.  Not a big deal though.  I'd rather the tests be robust and consistent.

ylemkimon: @kevinbarabash Fortunately, we don't reload the page unless `reload` option is set.

test/screenshotter/test.html
ylemkimon: @kevinbarabash Modern browsers except Edge support this way of loading (https://caniuse.com/#feat=font-loading).

kevinbarabash: Cool.  Hopefully Edge will add this soon. 🤞

PR #1893, opened 2019-03-23
Add KaTeX version to stylesheet and troubleshooting guide

ylemkimon:
It is hard to know the KaTeX version from the stylesheet and many issues have been reported due to their version mismatch. This PR adds the KaTeX version to the stylesheet and a troubleshooting guide to check it.


docs/issues.md
```diff
@@ -17,3 +17,25 @@ title: Common Issues
...
+To check the stylesheet (katex.css) is properly loaded, add following code to
+anywhere in the document:
+
+```html
+<style>
+  .katex-version {display: none;}
+  .katex-version::after {content:"0.10.2 or earlier";}
+</style>
+<span class="katex">
+  <span class="katex-mathml">The KaTeX stylesheet is not loaded!</span>
```
ylemkimon: If the stylesheet is loaded, `.katex .katex-mathml` will be hidden.

docs/issues.md
```diff
@@ -17,3 +17,25 @@ title: Common Issues
...
+  <span class="katex-version rule">KaTeX stylesheet version: </span>
```
ylemkimon: If the stylesheet is loaded, `.katex .rule` will be shown.

src/katex.less
```diff
@@ -4,6 +4,8 @@
 // The mu unit is defined as 1/18 em
 @mu: 1/18em;
 
+@version: "";
```
kevinbarabash: Is there something that replaces the `""` with the actual version string from package.json?

ylemkimon: @kevinbarabash 
https://github.com/KaTeX/KaTeX/blob/db184677ea34b49ebbe0b9266816b7431291c5cd/webpack.common.js#L64-L66
sets the version.

kevinbarabash: Could you add a comment here before merging so people don't accidental "fix" this?

PR #1895, opened 2019-03-23
 Add more documentation on font loading & prefetch fonts on the website

ylemkimon:


docs/browser.md
```diff
@@ -31,11 +31,30 @@ asynchronously. Add [`async` attribute](https://developer.mozilla.org/en/HTML/El
...
+<script defer src="https://cdn.jsdelivr.net/npm/webfontloader@1.6.28/webfontloader.js" integrity="sha256-4O4pS1SH31ZqrSO2A/2QJTVjTPqVe+jnYgOWUVr7EEc=" crossorigin="anonymous"></script>
+```
+
+You can also use [`rel="preload"`](https://developer.mozilla.org/en-US/docs/Web/HTML/Preloading_content).
+Add `<link rel="preload" href=(path to WOFF2 font) as="font" type="font/woff2" crossorigin="anonymous">`
+to `head`. Note that [only few browsers support it](https://caniuse.com/#feat=link-rel-preload)
+and they all support WOFF2 so preloading WOFF2 fonts is enough.
+
+See [Google Web Fundamentals - Web Font Optimization](https://developers.google.com/web/fundamentals/performance/optimizing-content-efficiency/webfont-optimization)
+for more detail.
```
ylemkimon: I'm not sure what'd be the best practice.

kevinbarabash: I think it depends on which browsers a person is targetting.

docs/browser.md
```diff
@@ -31,11 +31,30 @@ asynchronously. Add [`async` attribute](https://developer.mozilla.org/en/HTML/El
 to `script` and use [`rel="preload"` and `onload` attribute](https://github.com/filamentgroup/loadCSS)
 on `link`.
 
-You can prefetch KaTeX fonts to prevent FOUT or FOIT. Use [Web Font Loader](https://github.com/typekit/webfontloader)
-or add [`<link rel="preload" href=(path to WOFF2 font) as="font" type="font/woff2" crossorigin="anonymous">`](https://developer.mozilla.org/en-US/docs/Web/HTML/Preloading_content)
-to `head`. (Note that only few browsers [support `rel="preload"`](https://caniuse.com/#feat=link-rel-preload)
-and they all support WOFF2 so preloading WOFF2 fonts is enough.) You can use
-Chrome DevTools Network panel or similar to find out which fonts are used.
+You can prefetch KaTeX fonts to prevent FOUT or FOIT. Use [Font Loading API](https://developer.mozilla.org/en-US/docs/Web/API/CSS_Font_Loading_API)
+or [Web Font Loader](https://github.com/typekit/webfontloader):
```
kevinbarabash: It seems like this section font-loading is independent of how katex itself is loaded.  Maybe it deserves its own header and could be moved to the bottom of this page.

kevinbarabash: I'm going to merge this as is so that we can get the 0.10.2 release out.  We can always tweak this later if we decide to.

kevinbarabash:
I'm not sure what's up with the codecov report.

PR #1897, opened 2019-03-23
Improve MathML for \limits

ronkok:
PR #1890 revised the treatment of `\limits` to match the MathJax behavior. It turns out that the MathJax behavior does not render properly in Firefox. It may be a Firefox bug because the MathML looks okay to me. An expression like `\lim\limits_y x` returns MathML of `<munder><mo>lim</mo><mi>y</mi></munder>`. And yet Firefox's renders it like a subscript. This picture is how Firefox renders `\lim x
\lim_y x
\lim\limits_y x`

limits

This PR modifies the MathML to `<munder><mo><mi>lim</mi><mo>⁡</mo></mo><mi>y</mi></munder>`. The inner `<mo>` contains the invisible ApplyFunction character. Such a change causes Firefox to render the same example as:
limits-2

### DIFF
--- a/src/functions/op.js
+++ b/src/functions/op.js
@@ -261,7 +261,7 @@ const mathmlBuilder: MathMLBuilder<"op"> = (group, options) => {
         if (group.parentIsSupSub) {
             node = new mathMLTree.MathNode("mo", [node, operator]);
         } else {
-            return mathMLTree.newDocumentFragment([node, operator]);
+            node = mathMLTree.newDocumentFragment([node, operator]);
         }
     }
 
--- a/src/functions/op.js
+++ b/src/functions/op.js
@@ -252,17 +252,15 @@ const mathmlBuilder: MathMLBuilder<"op"> = (group, options) => {
         // operator's name.
         // TODO(emily): Add a space in the middle of some of these
         // operators, like \limsup.
+        node = new mathMLTree.MathNode(
+            "mi", [new mathMLTree.TextNode(group.name.slice(1))]);
+        // Append an <mo>&ApplyFunction;</mo>.
+        // ref: https://www.w3.org/TR/REC-MathML/chap3_2.html#sec3.2.4
+        const operator = new mathMLTree.MathNode("mo",
+            [mml.makeText("\u2061", "text")]);
         if (group.parentIsSupSub) {
-            node = new mathMLTree.MathNode(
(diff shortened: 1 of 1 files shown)


src/functions/op.js
```diff
@@ -252,17 +252,15 @@ const mathmlBuilder: MathMLBuilder<"op"> = (group, options) => {
...
+            node = new mathMLTree.MathNode("mo", [node, operator]);
         } else {
-            node = new mathMLTree.MathNode(
-                "mi", [new mathMLTree.TextNode(group.name.slice(1))]);
-            // Append an <mo>&ApplyFunction;</mo>.
-            // ref: https://www.w3.org/TR/REC-MathML/chap3_2.html#sec3.2.4
-            const operator = new mathMLTree.MathNode("mo",
-                [mml.makeText("\u2061", "text")]);
-
             return mathMLTree.newDocumentFragment([node, operator]);
```
ylemkimon:
```suggestion
             node = mathMLTree.newDocumentFragment([node, operator]);
```

ronkok: Done.

kevinbarabash:
Even though this code is covered by tests, no tests broke due to the changes because we aren't actually testing MathML output.  For the screenshot tests we should probably prevent MathML from being generated at all.  That way the coverage reports will be more accurate and actionable.

BenLauwens:
The generated MathML `<munder><mo><mi>lim</mi><mo>⁡</mo></mo><mi>y</mi></munder>` does not render correctly in Safari or Chrome (and AHFormatterV66). The original MathML `<munder><mo>lim</mo><mi>y</mi></munder>` renders perfectly.

ronkok:
I can confirm that @BenLauwens is correct and that `\limits` is not rendering properly in Safari MathML. The problem is not the `&ApplyFunction;` character added by this PR. Instead, there is a unnecessary `<mi>` nested inside `<mo><mi>lim</mi><mo>`. Firefox deals successfully with that `<mi>` but Safari does not.

I'll try to get rid of the `<mi>`.

ronkok:
Hey, can I get an review on PR #1899? The problem reported here by @BenLauwens is easy to fix, but the fix will be a merge conflict with PR #1899. So I am going to open a ticket for this new issue and mark it as blocked.

 PR #1899 does some good stuff. I really do think it deserves a little more love.

PR #1899, opened 2019-03-24
Support \operatorname*

### LINKED ISSUE #1242: \limits after \operatorname
Currently, using \limits after \operatorname gives the following error:
```
ParseError: KaTeX parse error: Limit controls must follow a math operator
```

Desired behavior:
`$\operatorname{E}\limits_x$` should place the x underneath the E, just like `$\max\limits_x$`

ronkok:
Support `\operatorname*`. I believe this resolves #1242.

I also modified four macros, deleting `\limits` and calling `\operatorname*` instead of `\operatorname`. The macros affected are: `\argmin`, `\argmax`, `\limsup`, and `\liminf`.


ronkok:
The screenshotter tests have failed with a `YAMLException` at line 16, the beginning of the data. My additions are on lines 266 and 267, so I have no idea what the problem is. Any help would be appreciated.

test/screenshotter/ss_data.yaml
```diff
@@ -1,4 +1,4 @@
-# List of test cases for which we create screenshots and other documents.
+operatorname# List of test cases for which we create screenshots and other documents.
```
ylemkimon:
```suggestion
# List of test cases for which we create screenshots and other documents.
```

janosh:
@ylemkimon @edemaine @kevinbarabash Would really like to start using `\operatorname*`. Anything preventing this from being merged?

ronkok:
Note to reviewer: The diffs are large, but that is mostly because a lot of code has moved. All of file `assembleSupSub.js` is code copied without modification from `op.js`. And inside `operatorname.js`, I had to make `htmlBuilder` and `mathmlBuilder` into free-standing functions so that they could be called from `operatorname*`.

Once those moves are taken into account, the amount of new code is not all that large.

kevinbarabash:
@ronkok it sounds like there's some overlap between this PR and #1966, is that correct?

ronkok:
> some overlap between this PR and #1966, is that correct?

I don't see much overlap with PR #1966. The `op.js` file saw some changes in PR #1890 and PR #1897 that are pretty closely tied to this one.

kevinbarabash:
@ronkok can you resolve the conflicts?  I'll make sure this is the next PR I review.  I should be able to get to it sometime this week.

ronkok:
@kevinbarabash Do you know any good command line way to resolve screenshotter merge conflicts?

kevinbarabash:
@ronkok I don't have much experience with merge conflicts in binary files but it looks like https://codeyarns.com/2016/04/11/how-to-resolve-git-conflict-with-binary-files/ should do the trick.

kevinbarabash:
@ronkok can you double check that I resolved the merge conflicts correctly?

ronkok:
@kevinbarabash Yes, the `\plim` conflict resolution is fine. We still have a screenshotter issue, though.

ronkok:
Okay, we have a branch that is once again clear of merge conflicts.

src/functions/supsub.js
```diff
@@ -231,6 +241,10 @@ defineFunctionBuilders({
             if (base && base.type === "op" && base.limits &&
                 (options.style === Style.DISPLAY || base.alwaysHandleSupSub)) {
                 nodeType = "munder";
+            } else if (base && base.type === "operatorname" &&
+                base.alwaysHandleSupSub &&
+                (options.style === Style.DISPLAY || base.limits)) {
```
kevinbarabash: nit: can you make the order of this `||` match the one line line 234?

test/screenshotter/ss_data.yaml
```diff
@@ -270,7 +270,8 @@ OldFont: |
 OperatorName: |
     \begin{matrix}
         \operatorname g (z) + 5\operatorname{g}z + \operatorname{Gam-ma}(z) \\
-        \operatorname{Gam ma}(z) + \operatorname{\Gamma}(z) + \operatorname{}x
+        \operatorname{Gam ma}(z) + \operatorname{\Gamma}(z) + \operatorname{}x \\
+        \operatorname*{asin} x + \operatorname*{asin}_y x + \operatorname*{asin}\limits_y x
     \end{matrix}
 OpLimits: |
```
kevinbarabash: Was our rendering of `limsup` always wrong?  I'm confused about the image diff since what katex.org currently renders for `\limsup_{x \rightarrow \infty} x \stackrel{?}= \liminf_{x \rightarrow \infty}` is:
Screen Shot 2019-07-13 at 7 03 55 PM

src/functions/supsub.js
kevinbarabash: Is there a case where `base.alwaysHandleSupSub` and `base.limits` is falsy?

test/katex-spec.js
```diff
@@ -2651,6 +2651,9 @@ describe("An aligned environment", function() {
 describe("operatorname support", function() {
     it("should not fail", function() {
         expect("\\operatorname{x*Π∑\\Pi\\sum\\frac a b}").toBuild();
+        expect("\\operatorname*{x*Π∑\\Pi\\sum\\frac a b}").toBuild();
+        expect("\\operatorname*{x*Π∑\\Pi\\sum\\frac a b}_y x").toBuild();
+        expect("\\operatorname*{x*Π∑\\Pi\\sum\\frac a b}\\limits_y x").toBuild();
```
kevinbarabash: What's the difference between the last two examples?  It might be good to have a screenshot test with those two items.

test/__snapshots__/mathml-spec.js.snap
```diff
@@ -347,7 +347,7 @@ exports[`A MathML builder should output \\limsup_{x \\rightarrow \\infty} correc
             ∞
           </mi>
         </mrow>
-      </munder>
+      </msub>
```
kevinbarabash: When should we be using `munder` vs. `msub`?

src/functions/operatorname.js
```diff
@@ -3,115 +3,160 @@ import defineFunction, {ordargument} from "../defineFunction";
...
 
+    if (group.parentIsSupSub) {
+        return new mathMLTree.MathNode("mo", [identifier, operator]);
+    } else {
         return mathMLTree.newDocumentFragment([identifier, operator]);
+    }
+};
+
+// \operatorname
+// amsopn.dtx: \mathop{#1\kern\z@\operator@font#3}\newmcodes@
```
kevinbarabash: We don't actually follow this definition do we?

src/functions/operatorname.js
```diff
@@ -3,115 +3,160 @@ import defineFunction, {ordargument} from "../defineFunction";
...
+        // If we have limits, supsub will pass us its group to handle. Pull
+        // out the superscript and subscript and set the group to the op in
+        // its base.
+        supGroup = supSub.sup;
+        subGroup = supSub.sub;
+        group = assertNodeType(supSub.base, "operatorname");
+        hasLimits = true;
+    } else {
+        group = assertNodeType(grp, "operatorname");
+    }
```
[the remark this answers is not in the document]
kevinbarabash: After looking at the other places we do a similar check, each of them seems to be a little different.  Oh well.

src/functions/operatorname.js
```diff
@@ -3,115 +3,160 @@ import defineFunction, {ordargument} from "../defineFunction";
...
+    props: {
+        numArgs: 1,
     },
+    handler: ({parser, funcName}, args) => {
+        const body = args[0];
+        return {
+            type: "operatorname",
+            mode: parser.mode,
+            body: ordargument(body),
+            alwaysHandleSupSub: (funcName === "\\operatorname*"),
```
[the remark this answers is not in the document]
edemaine: In some styles and I believe this codebase, it's customary to write Boolean comparisons in parentheses like this, I believe to distinguish from `funcName = "\\operatorname*"`.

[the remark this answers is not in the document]
kevinbarabash: I hadn't picked up on that style. 😅 In that case let's keep this as is.

test/screenshotter/ss_data.yaml
edemaine: I spotted the following comments in `functions/op.js` which should be removed:
```js
        // TODO(emily): Add a space in the middle of some of these
        // operators, like \limsup
 ``` 
Instead, we already have the following macros in `macros.js` that add `\,` in the middle:
```js
defineMacro("\\limsup", "\\DOTSB\\mathop{\\operatorname{lim\\,sup}}\\limits");
defineMacro("\\liminf", "\\DOTSB\\mathop{\\operatorname{lim\\,inf}}\\limits");
```
If I understand correctly, this PR seems to be disabling underneath limits for `\limsup`, which seems like a bug?

ronkok: In AMS LaTeX, `\limsup`  behaves as does `\sum`. In text style, they both show a sub next to the operator. They show a sub below the operator if in display style or if the operator is followed by `\limits`. Here's an example from quicklatex.com that renders `\( \limsup_x y \: \limsup\limits_x y \: \sum_x^y \: \sum\limits_x^y \:\displaystyle \limsup_x y \: \sum_x^y \)
`


 The macro that exists in current KaTeX always shows the sub below a `\limsup` and thus is incorrect when in text style. This PR attempts to get the correct behavior. I'll double check to see that it does and if the screenshots don't show a text style distinction, I'll write a screenshot test case.

src/functions/assembleSupSub.js
```diff
@@ -0,0 +1,108 @@
+
```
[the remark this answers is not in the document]
ronkok: @kevinbarabash  We don't have a `utils` folder. If we did, then the `utils.js` file would belong there and I would prefer to avoid updating all the paths that point to `utils.js`. As an alternate, I propose to put this function back into `op.js` but expose it so that it can be called from `operatorname.js`.

Would that be okay?

src/Parser.js
```diff
@@ -315,15 +315,21 @@ export default class Parser {
...
                     opNode.limits = limits;
                     opNode.alwaysHandleSupSub = true;
                 } else {
-                    throw new ParseError(
-                        "Limit controls must follow a math operator",
-                        lex);
+                    opNode = checkNodeType(base, "operatorname");
+                    if (opNode && opNode.alwaysHandleSupSub) {
+                        const limits = lex.text === "\\limits";
+                        opNode.limits = limits;
```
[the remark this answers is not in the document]
ronkok: > So limits aren't allowed for `\operatorname`

It might be more precise to say that `\limits` is ineffective when applied to `\operatorname`.  You can write a `\limits` after an `\operatorname` without a log error, but the sub will only be shifted below in the case of `\operatorname*`.

src/functions/operatorname.js
```diff
@@ -3,115 +3,160 @@ import defineFunction, {ordargument} from "../defineFunction";
...
-        // ref: https://www.w3schools.com/charsets/ref_html_entities_a.asp
-        const operator = new mathMLTree.MathNode("mo",
-            [mml.makeText("\u2061", "text")]);
+    // \u2061 is the same as &ApplyFunction;
+    // ref: https://www.w3schools.com/charsets/ref_html_entities_a.asp
+    const operator = new mathMLTree.MathNode("mo",
+        [mml.makeText("\u2061", "text")]);
 
+    if (group.parentIsSupSub) {
+        return new mathMLTree.MathNode("mo", [identifier, operator]);
```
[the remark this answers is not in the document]
ronkok: Good question. In MathML, a function name is an identifier. That's why a screen reader needs the `&ApplyFunction;` character to distinguish it from a variable name. But MathML specifies a `<mo>` for anything that takes `\limits`.

Also, this code needs to be revised in response to issue #2028. The `<mo>` case does not need the `&ApplyFunction;`. I might just sneak that nuance into this PR.

test/screenshotter/ss_data.yaml
edemaine: Ah, I assumed the screenshot was in `\displaystyle`, but if it's in `\textstyle`, that makes more sense! I just checked the definition of `\limsup` and `\liminf` in `amsopn.sty`, and they are clearly just `\operatorname*{lim\,sup}` and `\operatorname*{lim\,inf}`, so I am definitely a fan of the new macros in this PR.

test/katex-spec.js
ronkok: It should build with or without `\limits`. So I wrote a test case both ways. And yes, a screen shot both ways would help.

test/__snapshots__/mathml-spec.js.snap
ronkok: The `<munder>` was a hack because we did not have a way to apply `\limits` to `\operatorname`. Now that we can apply `\limits`, we should expect to see `<msub>` in the snapshot of a text style example.

src/functions/operatorname.js
ronkok: Well, I've tried to emulate its behavior as closely as I can. I'll let others judge how well I did.

src/functions/assembleSupSub.js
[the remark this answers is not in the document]
ronkok: Never mind. It's done.

src/functions/operatorname.js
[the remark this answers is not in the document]
ronkok: Scratch that last comment. I'll address issue #2028 with another PR after this one lands.

src/functions/assembleSupSub.js
```diff
@@ -0,0 +1,108 @@
...
+import buildCommon from "../buildCommon";
+import * as html from "../buildHTML";
+
+// We break this function out because it is common to both op.js and opeartorname.js
```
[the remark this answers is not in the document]
ronkok: Done.

src/functions/assembleSupSub.js
```diff
@@ -0,0 +1,108 @@
...
+
+export const assembleSupSub = (
+    base,
+    supGroup,
+    subGroup,
+    options,
+    style,
+    slant,
+    baseShift,
```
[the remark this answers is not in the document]
ronkok: Done.

src/functions/operatorname.js
```diff
@@ -3,115 +3,160 @@ import defineFunction, {ordargument} from "../defineFunction";
...
-                if (child instanceof SymbolNode) {
-                    // Per amsopn package,
-                    // change minus to hyphen and \ast to asterisk
-                    child.text = child.text.replace(/\u2212/, "-")
-                        .replace(/\u2217/, "*");
-                }
+// NOTE: Unlike most `htmlBuilder`s, this one handles not only
+// "operatorname", but also  "supsub" since \operatorname* can
+// affect super/subscripting.
+export const htmlBuilder: HtmlBuilderSupSub<"operatorname"> = (grp, options) => {
```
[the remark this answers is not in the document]
ronkok: That's reasonable. In the upcoming effort for automatically numbered environments, I will break the work up into smaller PRs. In return, I will ask for quicker reviews.

src/functions/supsub.js
ronkok: > Is there a case where `base.alwaysHandleSupSub` and `base.limits` is falsy?

Yes. `\lim\nolimits_x y`

ronkok:
I have picked up review comments and I have added some variety to the screenshots in OperatorName.png and OpLimits.png. They now show items with and without limits, and in text style and display style.

I think this is ready for back check.

src/functions/operatorname.js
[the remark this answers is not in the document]
kevinbarabash: Thanks for raising a ticket for this.

src/functions/operatorname.js
kevinbarabash: Is `\kern\z@` a zero width kern to ensure a minimum height/depth?  I assume `\operator@font` is an alias for cmr.  I guess it useful to have the comment if we're not doing a literal translation of it.

src/utils/assembleSupSub.js
```diff
@@ -0,0 +1,112 @@
+// @flow
```
kevinbarabash: Thanks for adding flow.

test/katex-spec.js
kevinbarabash: Thanks for updating the screenshots to check for this.

src/functions/op.js
```diff
@@ -4,8 +4,9 @@ import defineFunction, {ordargument} from "../defineFunction";
 import buildCommon from "../buildCommon";
 import {SymbolNode} from "../domTree";
 import * as mathMLTree from "../mathMLTree";
-import utils from "../utils";
+import utils from "../utils/utils";
 import Style from "../Style";
+import {assembleSupSub} from "../utils/assembleSupSub";
```
kevinbarabash: I was thinking that `assembleSupSub` would go in `src/functions/utils/assembleSupSub.js`.  That would avoid unrelated changes to files importing `src/utils.js`.  Sorry for not being clearer in my original comment.  The rest of the PR looks good.  I'll push a commit to move things around and then approve this.

src/functions/operatorname.js
ronkok: Re: `\kern\z`. According to `amsopn`:

> In the interior of the \mathop we need a null object (we choose a zero kern for
minimum waste of main mem) in order to guard against the case where #3 is a
single letter; TeX will seize it and center it on the math axis if there is nothing
else inside the \mathop atom.

kevinbarabash: I didn't realize that TeX centered single letter on the math axis.  I just assumed that it used baseline alignment.  TIL.

PR #1902, opened 2019-03-24
Improve more coloneq relations

ronkok:
This is similar in effect to PR #1886, in that it improves the MathML for nine more symbols in the colonequals series. It uses a ligature instead of a single Unicode code point. In Firefox, they look pretty good!


kevinbarabash:
I love that we're able to do so much with macros now. 🙂

PR #1906, opened 2019-03-25
Improve MathML for \middle

ronkok:
`\middle` is a delimiter, not an operator. So suppress the operator spacing.


kevinbarabash:
[mfenced](https://developer.mozilla.org/en-US/docs/Web/MathML/Element/mfenced) looks like a better choice for delimiters in general.

ronkok:
Near the bottom of [this page](https://mathml.igalia.com/project/), it is stated that the Chromium MathML project does not intend to support `mfenced`, at least not right away. So I thought it best to stick with `<mo>`.

PR #1908, opened 2019-03-26
Improve MathML for characters in Unicode private use area

ronkok:
The KaTeX fonts contain a few glyphs at code points inside the Unicode private use area. Naturally, passing a private area code point to MathML is unsuccessful.

This PR installs a macro for each of the relevant functions. Each macro points the HTML to the KaTeX glyph and points the MathML to a (near) match among the standard Unicode characters.

Here are the functions in this PR:

|   **Function**  |   **KaTeX code point**   |   **Substitute code point**   |   **LaTeX character**   |   **Substitute character**  |
|:----------------|:----------:|:----------:|:----:|:----:|
|   \gvertneqq    |   U+E00D   |   U+2269   | gvertneqq     |   ≩  |
|   \lvertneqq    |   U+E00C   |   U+2268   |  lvertneqq    |   ≨  |
|   \ngeqq        |   U+E00E   |   U+2271   |  ngeqq    |   ≱  |
|   \ngeqslant    |   U+E00F   |   U+2271   |  ngeqslant    |   ≱  |
|   \nleqq        |   U+E011   |   U+2270   |  nleqq    |   ≰  |
|   \nleqslant    |   U+E010   |   U+2270   |  nleqslant    |   ≰  |
|   \nshortmid    |   U+E006   |   U+2224   |  nshortmid    |   ∤  |
| \nshortparallel |   U+E007   |   U+2226   |  nshortparallel    |   ∦  |
| \varsubsetneq   |   U+E01A   |   U+228A   | varsubsetneq     |   ⊊  |
| \varsubsetneqq  |   U+E017   |   U+2ACB   |  varsubsetneqq    |   ⫋  |
| \varsupsetneq   |   U+E01B   |   U+228B   |  varsupsetneq    |   ⊋  |
| \varsupsetneqq  |   U+E019   |   U+2ACC   |  varsupsetneqq  |   ⫌  |

`\not` also uses a glyph from the private use area. I’m not sure what to do about `\not`. That’s a topic for some later PR.


src/symbols.js
```diff
@@ -240,16 +241,16 @@ defineSymbol(math, ams, rel, "\u22e9", "\\succnsim", true);
 defineSymbol(math, ams, rel, "\u2aba", "\\succnapprox", true);
 // unicode-math maps \u2246 to \simneqq. We'll use the AMS synonym.
 defineSymbol(math, ams, rel, "\u2246", "\\ncong", true);
-defineSymbol(math, ams, rel, "\ue007", "\\nshortparallel");
+defineSymbol(math, ams, rel, "\ue007", "\\@nshortparallel");
 defineSymbol(math, ams, rel, "\u2226", "\\nparallel", true);
 defineSymbol(math, ams, rel, "\u22af", "\\nVDash", true);
 defineSymbol(math, ams, rel, "\u22eb", "\\ntriangleright");
 defineSymbol(math, ams, rel, "\u22ed", "\\ntrianglerighteq", true);
 defineSymbol(math, ams, rel, "\ue018", "\\nsupseteqq");
```
ylemkimon: Maybe `\u2288`?

src/symbols.js
```diff
@@ -240,16 +241,16 @@ defineSymbol(math, ams, rel, "\u22e9", "\\succnsim", true);
...
 defineSymbol(math, ams, rel, "\u228b", "\\supsetneq", true);
-defineSymbol(math, ams, rel, "\ue01b", "\\varsupsetneq");
+defineSymbol(math, ams, rel, "\ue01b", "\\@varsupsetneq");
 defineSymbol(math, ams, rel, "\u2acc", "\\supsetneqq", true);
-defineSymbol(math, ams, rel, "\ue019", "\\varsupsetneqq");
+defineSymbol(math, ams, rel, "\ue019", "\\@varsupsetneqq");
 defineSymbol(math, ams, rel, "\u22ae", "\\nVdash", true);
 defineSymbol(math, ams, rel, "\u2ab5", "\\precneqq", true);
 defineSymbol(math, ams, rel, "\u2ab6", "\\succneqq", true);
```
ylemkimon: Maybe `u2289`? (next line)

ronkok: Well, those are interesting suggestions. If I understand matters correctly, I am proposing Option A below and you are proposing Option B

| Function     | LaTeX | Option A | Option B |
|:------------|:--------:|:----------:|:----------:|
| \varsubsetneq |  varsubsetneq     |  ⊊         |    ⊈   |
| \varsupsetneq |  varsupsetneq     |  ⊋         |    ⊉   |

I find the LaTeX symbols to be ambiguous and annoying. Either Option A or Option B is less ambiguous but I am not completely sure what the LaTeX symbols are trying to convey. If I had to bet, I'd pick Option A. For what it's worth MathMax and TeXZilla made the same choice that I made, as you can see [here](http://webmathcomparison.net/#v).

BTW, I've updated that page with `\middle`. Thank you for the quick review.

edemaine: The name "subsetneq" suggests "subset but not equal" which seems to be Option A, while Option B means "not subset". Presumably the Unicode character names confirm this?

ronkok: Unicode character names for Option A are: "Subset of with Not Equal To" and "Superset of with Not Equal To", which I think is more corroboration.

ylemkimon: @ronkok I meant for the `\nsupseteqq` and `\nsubseteqq`, which also use PUAs.

ronkok: @ylemkimon Thank you for the clarification. I have revised this PR to include `\nsubseteqq` and `\nsupseteqq`.

PR #1910, opened 2019-03-27
Improve MathML for environments

ronkok:
This PR improves the MathML of array environments. Specifically, it improves
* row spacing
* column spacing
* column alignment
* vertical lines in `{array}` environments
* horizontal lines from `\hline` and `\hdashline`.

To aid review, I have changed the page at [webmathcomparison.net](http://webmathcomparison.net/) so that it runs a version of KaTeX that includes this PR.


src/environments/array.js
```diff
@@ -367,21 +369,123 @@ const htmlBuilder: HtmlBuilder<"array"> = function(group, options) {
...
                 }));
         }));
+
+    // Set column alignment, row spacing, column spacing, and
+    // array lines by setting attributes on the table element.
+
+    // Set the row spacing. In MathML, we specify a gap distance.
+    // We do not use rowGap[] because MathML automatically increases
+    // cell height with the height/depth of the element content.
+    const gap = 0.16 + group.arraystretch - 1 + (group.addJot ? 0.09 : 0);
```
ylemkimon: Could you add comments about these constants?

src/environments/array.js
```diff
@@ -367,21 +369,123 @@ const htmlBuilder: HtmlBuilder<"array"> = function(group, options) {
...
+                align += alignMap[cols[i].align];
+
+                if (prevTypeWasAlign) {
+                    columnLines += "none ";
+                }
+                prevTypeWasAlign = true;
+
+            } else if (cols[i].type === "separator") {
+                // MathML accepts only single lines between cells.
+                // So we read only the first of consecutive separators.
```
ylemkimon: Would it be possible to have an empty column/row between two lines?

ronkok: Yes, that is possible, but it would involve rewriting the existing, tightly nested code that now creates the `mtable` body. That is more than I am prepared to take on at the moment. Maybe some future PR can do it.

src/environments/array.js

src/environments/array.js
```diff
@@ -367,21 +369,127 @@ const htmlBuilder: HtmlBuilder<"array"> = function(group, options) {
...
+        table.setAttribute("columnalign", align.trim());
+
+        if (/[sd]/.test(columnLines)) {
+            table.setAttribute("columnlines", columnLines.trim());
+        }
+    }
+
+    // {aligned} and {alignedat} get 0 column spacing.
+    // The others get 1em.
+    table.setAttribute("columnspacing", group.isAlign ? "0em" : "1em");
```
ylemkimon: Is `postgap` somehow related to this?

ronkok: Yes. The HTML part of the code creates a `colsep` span whose width is either 0, or 1em (1qquad), or the default `arraycolsep`. The `align` environment gets a `pregap` defined as 1em, but I cannot see how this makes any difference. The end result of all the `colsep`, `arraycolsep`, logic works out in effect to just two choices and I chose to simplify it into this one line of code. 

I have plans to add more environments. I will revisit this line of code when I do that.

ronkok: I figured out what pregap does. The latest commit improves the column spacing for the `aligned` environment.

src/environments/array.js
```diff
@@ -25,6 +25,10 @@ export type AlignSpec = { type: "separator", separator: string } | {
     postgap?: number,
 };
 
+// Constants to indicate column separation
+const ALIGN = 1;
+const ALIGNAT = 2;
```
ylemkimon:
```suggestion
export type ColSeparationType = "align" | "alignat";
```

src/environments/array.js
```diff
@@ -25,6 +25,10 @@ export type AlignSpec = { type: "separator", separator: string } | {
     postgap?: number,
 };
 
+// Constants to indicate column separation
+const ALIGN = 1;
```
ylemkimon:
```suggestion
```

ylemkimon:
```suggestion
```

src/environments/array.js
ylemkimon:
```suggestion
export type ColSeparationType = "align" | "alignat";
```

src/environments/array.js
```diff
@@ -48,11 +52,12 @@ function getHLines(parser: Parser): boolean[] {
  */
 function parseArray(
     parser: Parser,
-    {hskipBeforeAndAfter, addJot, cols, arraystretch}: {|
+    {hskipBeforeAndAfter, addJot, cols, arraystretch, colSeparationType}: {|
         hskipBeforeAndAfter?: boolean,
         addJot?: boolean,
         cols?: AlignSpec[],
         arraystretch?: number,
+        colSeparationType?: number,
```
ylemkimon:
```suggestion
        colSeparationType?: ColSeparationType,
```

ylemkimon: Is this needed? It seems there is no `parseArray` calls with `colSeparationType`.

src/environments/array.js
```diff
@@ -367,15 +373,127 @@ const htmlBuilder: HtmlBuilder<"array"> = function(group, options) {
...
+    const gap = 0.16 + group.arraystretch - 1 + (group.addJot ? 0.09 : 0);
+    table.setAttribute("rowspacing", gap + "em");
+
+    // MathML table lines go only between cells.
+    // To place a line on an edge we'll use <menclose>, if necessary.
+    let menclose = "";
+    let align = "";
+
+    if (group.cols) {
+        // Find column alignment, column spacing, and  vertical lines.
```
ylemkimon:
```suggestion
        // Find column alignment, column spacing, and  vertical lines.
        const cols = group.cols;
```

src/environments/array.js
```diff
@@ -367,15 +373,127 @@ const htmlBuilder: HtmlBuilder<"array"> = function(group, options) {
...
+        let columnLines = "";
+        let prevTypeWasAlign = false;
+        let iStart = 0;
+        let iEnd = group.cols.length;
```
ylemkimon:
```suggestion
        let iEnd = cols.length;
```

src/environments/array.js
```diff
@@ -367,15 +373,127 @@ const htmlBuilder: HtmlBuilder<"array"> = function(group, options) {
...
+
```
ylemkimon:
```suggestion
```

src/environments/array.js
```diff
@@ -367,15 +373,127 @@ const htmlBuilder: HtmlBuilder<"array"> = function(group, options) {
...
+        const cols = group.cols;
```
ylemkimon:
```suggestion
```

src/environments/array.js
```diff
@@ -367,15 +373,127 @@ const htmlBuilder: HtmlBuilder<"array"> = function(group, options) {
...
+
+        for (let i = iStart; i < iEnd; i++) {
+            if (cols[i].type === "align") {
+                align += alignMap[cols[i].align];
+
+                if (prevTypeWasAlign) {
+                    columnLines += "none ";
+                }
+                prevTypeWasAlign = true;
+
```
ylemkimon:
```suggestion
```

src/environments/array.js
```diff
@@ -367,15 +373,127 @@ const htmlBuilder: HtmlBuilder<"array"> = function(group, options) {
...
+        table.setAttribute("columnalign", align.trim());
+
+        if (/[sd]/.test(columnLines)) {
+            table.setAttribute("columnlines", columnLines.trim());
+        }
+    }
+
+    // Set column spacing.
+    if (group.colSeparationType === ALIGN) {
```
ylemkimon:
```suggestion
    if (group.colSeparationType === "align") {
```

src/environments/array.js
```diff
@@ -367,15 +373,127 @@ const htmlBuilder: HtmlBuilder<"array"> = function(group, options) {
...
+        const cols = group.cols || [];
+        let spacing = "";
+        for (let i = 1; i < cols.length; i++) {
+            spacing += i % 2 ? "0em " : "1em ";
+        }
+        table.setAttribute("columnspacing", spacing.trim());
+    } else if (group.colSeparationType === ALIGNAT) {
```
ylemkimon:
```suggestion
    } else if (group.colSeparationType === "alignat") {
```

src/environments/array.js
```diff
@@ -367,15 +373,127 @@ const htmlBuilder: HtmlBuilder<"array"> = function(group, options) {
...
+            table.setAttribute("columnlines", columnLines.trim());
+        }
+    }
+
+    // Set column spacing.
+    if (group.colSeparationType === ALIGN) {
+        const cols = group.cols || [];
+        let spacing = "";
+        for (let i = 1; i < cols.length; i++) {
+            spacing += i % 2 ? "0em " : "1em ";
```
ylemkimon: Should this be from `1` to `cols.length - 1`?

src/environments/array.js
```diff
@@ -448,6 +566,7 @@ const alignedHandler = function(context, args) {
             postgap: 0,
         };
     }
+    res.colSeparationType = isAligned ? ALIGN : ALIGNAT;
```
ylemkimon:
```suggestion
    res.colSeparationType = isAligned ? "align" : "alignat";
```

src/parseNode.js
```diff
@@ -28,6 +28,7 @@ type ParseNodeTypes = {
         type: "array",
         mode: Mode,
         loc?: ?SourceLocation,
+        colSeparationType?: number,
```
ylemkimon:
```suggestion
        colSeparationType?: ColSeparationType,
```

src/parseNode.js
```diff
@@ -28,6 +28,7 @@ type ParseNodeTypes = {
         type: "array",
```
ylemkimon: I think it's better to make `ColSeparationType` a type and import it here like `AlignSpec`.

src/environments/array.js
ronkok: The current code result in the correct number of column spacing specs. There is not point in specifying a column space for before the first column or after the last column. I think we should keep this code as is.

ronkok:
The most recent commit pick ups the review comments.

ylemkimon:
@ronkok Thank you for your hard work! 😄

ronkok:
@ylemkimon  Thank you for the valuable reviews!

PR #1912, opened 2019-03-28
Improve MathML for \rule

ronkok:
Address a long standing TODO item.

This PR wraps up the current plans I have for MathML improvements. I'll gladly help resolve anything that comes up during review and I have a couple of outstanding Safari rendering bugs and a `\smallint` bug to address. But other than that, I go back on hiatus for a few months.

As MathML PRs are merged into master, I'll update [webmathcomparison.net](http://webmathcomparison.net/). I'm quite pleased with how the rendering has improved.

@kevinbarabash If you would like me to circle back and write some MathML snapshots, let me know which ones and I'll be glad to oblige.


src/functions/rule.js
```diff
@@ -55,10 +55,25 @@ defineFunction({
         return rule;
     },
     mathmlBuilder(group, options) {
-        // TODO(emily): Figure out if there's an actual way to draw black boxes
-        // in MathML.
-        const node = new mathMLTree.MathNode("mrow");
+        const width = calculateSize(group.width, options);
+        const height = calculateSize(group.height, options);
+        const shift = (group.shift) ? calculateSize(group.shift, options) : 0;
```
ylemkimon: I think code should be consistent with the HTML builder above, or common codes can be extracted to a function.

src/functions/rule.js
```diff
@@ -55,10 +55,25 @@ defineFunction({
...
+        let color = options.color && options.getColor();
+        color = color ? color : "black";
```
ylemkimon:
```js
const color = options.color && options.getColor() || "black";
```

src/functions/rule.js
```diff
@@ -55,10 +55,25 @@ defineFunction({
...
+        rule.setAttribute("mathbackground", color);
+        rule.setAttribute("width", width + "em");
+        rule.setAttribute("height", height + "em");
+
+        const wrapper = new mathMLTree.MathNode("mpadded", [rule]);
+        const prefix = shift >= 0 ? "+" : "";
+        wrapper.setAttribute("height", prefix + shift + "em");
+        if (shift < 0) {
+            wrapper.setAttribute("depth", "+" + Math.abs(shift) + "em");
+        }
```
ylemkimon:
```js
if (prefix >= 0) {
    wrapper.setAttribute("height", "+" + shift + "em");
} else {
    wrapper.setAttribute("height", shift + "em");
    wrapper.setAttribute("depth", "+" + (-shift) + "em");
}
```

src/functions/rule.js
ronkok: Done.

src/functions/rule.js
ronkok: Done.

kevinbarabash:
@ronkok thanks for all of the MathML improvements.  With regards to adding snapshots for MathML, I think it would be good to wait for https://github.com/KaTeX/KaTeX/pull/1894 so that we can use the coverage reports to help with determining what snapshots we need.

PR #1914, opened 2019-03-29
\Improve MathML for \colorbox and \fcolorbox

ronkok:
Add border color.

### DIFF
--- a/src/functions/enclose.js
+++ b/src/functions/enclose.js
@@ -112,7 +112,9 @@ const htmlBuilder = (group, options) => {
 
 const mathmlBuilder = (group, options) => {
     const node = new mathMLTree.MathNode(
-        "menclose", [mml.buildGroup(group.body, options)]);
+        (group.label === "\\fcolorbox") ? "mpadded" :"menclose",
+        [mml.buildGroup(group.body, options)]
+    );
     switch (group.label) {
         case "\\cancel":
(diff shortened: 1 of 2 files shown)


src/functions/enclose.js
```diff
@@ -127,8 +129,13 @@ const mathmlBuilder = (group, options) => {
...
             break;
         case "\\fcolorbox":
-            // TODO(ron): I don't know any way to set the border color.
-            node.setAttribute("notation", "box");
+            // <menclose> cannot set a border color. So use <mpadded> instead.
+            // Set some attributes that come included with <menclose>.
+            node.setAttribute("width", "+10px");
+            node.setAttribute("height", "+5px");
+            node.setAttribute("lspace", "5px");
+            node.setAttribute("style", "border: 1.5px solid " +
```
ylemkimon: Are these values documented somewhere?

ronkok: I used the same values as MathJax. If you would prefer the LaTeX2e padding, I would have no objection.

I do object to the LaTeX2e border width. On my screen, I can barely see it. So for this border, I chose a 1.5px value as a happy medium between the too-thin LaTeX2e (0.04em) border and the too-gaudy MathJax (2px) border.

ronkok:
I've added `\colorbox` to this PR. I've also changed the padding and border thickness to dimensions that should be thoroughly non-controversial.

PR #1915, opened 2019-03-29
Improve MathML for \tag

ronkok:
Contrary to what is stated on MDN, Firefox does not actually support `<mlabeledtr>`. Furthermore, `<mlabeledtr>` is explicitly omitted from the [project scope](https://mathml.igalia.com/project/) of the Chromium MathML project.  So this PR revises `\tag` in a way that will actually render.

### DIFF
--- a/src/functions/tag.js
+++ b/src/functions/tag.js
@@ -4,21 +4,22 @@ import mathMLTree from "../mathMLTree";
 
 import * as mml from "../buildMathML";
 
+const pad = () => {
+    const padNode = new mathMLTree.MathNode("mtd", []);
+    padNode.setAttribute("width", "50%");
+    return padNode;
+}
+
 defineFunctionBuilders({
     type: "tag",
     mathmlBuilder(group, options) {
-        const pad = new mathMLTree.MathNode("mtd", []);
-        pad.setAttribute("width", "50%");
-        const pad2 = new mathMLTree.MathNode("mtd", []);
-        pad2.setAttribute("width", "50%");
-
         const table = new mathMLTree.MathNode("mtable", [
             new mathMLTree.MathNode("mtr", [
-                pad,
+                pad(),
                 new mathMLTree.MathNode("mtd", [
                     mml.buildExpressionRow(group.body, options),
                 ]),
-                pad2,
+                pad(),
                 new mathMLTree.MathNode("mtd", [
                     mml.buildExpressionRow(group.tag, options),
                 ]),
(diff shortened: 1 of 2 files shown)


src/functions/tag.js
```diff
@@ -7,17 +7,22 @@ import * as mml from "../buildMathML";
...
+
         const table = new mathMLTree.MathNode("mtable", [
-            new mathMLTree.MathNode("mlabeledtr", [
+            new mathMLTree.MathNode("mtr", [
+                pad,
                 new mathMLTree.MathNode("mtd", [
-                    mml.buildExpressionRow(group.tag, options),
+                    mml.buildExpressionRow(group.body, options),
                 ]),
+                pad,
```
ylemkimon: I'm not sure it's safe to use the same instance twice.

ronkok: Okay, I've made a duplicate.

ronkok:
I just realized that I need to work a left-aligned tag into this PR. I'll get to that tomorrow.

ronkok:
A left-aligned tag turns out to need a fairly deep intervention, more than I am prepared to do at the moment. This is still a valuable stand-alone PR. It turns a tag that did not render into one that does render. I'd prefer to see this PR be merged now and I'll come back later to finish the left-alignment.

I've written a TODO in the code that explains the situation.

src/functions/tag.js
```diff
@@ -4,21 +4,37 @@ import mathMLTree from "../mathMLTree";
...
+    padNode.setAttribute("width", "50%");
+    return padNode;
+};
+
 defineFunctionBuilders({
     type: "tag",
     mathmlBuilder(group, options) {
         const table = new mathMLTree.MathNode("mtable", [
-            new mathMLTree.MathNode("mlabeledtr", [
+            new mathMLTree.MathNode("mtr", [
```
edemaine: I'm late to the party here, but I wonder if (in a future PR) `mtr` should be `mlabeledtr` for `\tag`s? See https://www.w3.org/TR/MathML/chapter3.html#id.3.5.3.3

ronkok: @edemaine I agree, but only when and if the major browsers actually support `<mlabeledtr>`. At present, Firefox has baked-in browser CSS that hides the text inside a `<mlabeledtr>`. The Chromium MathML project's [project scope](https://mathml.igalia.com/project/) explicitly omits `<mlabeledtr>`.

edemaine: Oops, sorry, missed that in your original post. Thanks for investigating this!

PR #1918, opened 2019-04-01
Remove redundant style attribute from .op-symbol

glebm:
`!hasLimits` implies that `base` is not a `supsub`. Non-zero `baseShift` then further implies that it's a symbol and has the `op-symbol` class (line 70).

The `op-symbol` class already has `position: relative` set in the stylesheet.


glebm:
Screenshotter fails so I must've missed something

glebm:
This is what I've missed, it doesn't have the `op-symbol` class if it comes from here:

https://github.com/KaTeX/KaTeX/blob/01a94d5f07ad39443438a4445f577a9281eaa80f/src/functions/op.js#L94-L95

Example: `\mathop{\delta}`

Issue #1867, opened 2019-02-27
\clap has visible width in Safari


KazimirPodolski:
In Safari:

![](https://gitlab.softaria.com/web-editor/web-editor/uploads/c07800e0605c3cabc607f653b0ef2063/image.png)
Should be like this (as it is in Chrome):

![](https://gitlab.softaria.com/web-editor/web-editor/uploads/69e34fa35410712fea0a9d3105168c84/image.png)

ronkok:
I grow weary of Safari weirdness. 

Sometime in the next week, I'll take a look and see if something can be done. No guarantee that I'll succeed.

ronkok:
@spontaliku-softaria This is just a note to say I have not forgotten this issue. I work on a Windows box, so my access to Safari is limited. Next week some new iPads will be available. I'll buy one and see if I can do something here.

ronkok:
@spontaliku-softaria I'm sorry. I tried many ways to squash this and every one failed. Safari is picking up an extraneous width on a span deep within KaTeX's main function for vertical alignment. There is a [standing issue](https://github.com/KaTeX/KaTeX/issues/1409) to rewrite that function using CSS flex. It's possible that such a change would fix this issue. Until then, I can think of no way to resolve this.

ronkok:
Back one last time to say that I've found a fix. This should work properly in the next release.

Issue #1409, opened 2018-06-05
Simplify layout using flex box


kevinbarabash:
Flex box would allow us to simplify the markup we produce as well as allowing us to support commands that rely on the stretchiness/shrinkiness of glue.  Here's a prototype that use flex box for layout:

**Rendering:**
<img width="291" alt="screen shot 2018-06-05 at 1 09 10 am" src="https://user-images.githubusercontent.com/1044413/40956227-1799798a-685d-11e8-81ba-09e7c0aee82a.png">

**HTML**
```html
<span>
  1 +
  <span class="vbox" style="vertical-align:15px;">
    <span class="hbox">
      <span class="glue"></span>
      <span>1</span>
      <span class="glue"></span>
    </span>
    <span class="rule"></span>
    <span class="hbox">
      <span class="glue"></span>
      <span>
        1 + 
        <span class="vbox" style="vertical-align:14px;">
          <span class="hbox">
            <span class="glue"></span>
            1
            <span class="glue"></span>
          </span>
          <span class="rule"></span>
          <span class="hbox">
            <span class="glue"></span>
            x + 1
            <span class="glue"></span>
          </span>
        </span>  
      </span>
      <span class="glue"></span>
    </span>
  </span>
</span>
```

**CSS**
```css
body {
  font-size: 48px;
}

.vbox {
  display: inline-flex;
  flex-direction: column;
}

.hbox {
  display: inline-flex;
  flex-direction: row;
  width: 100%;
}

.rule {
  height: 4px;
  background: black;
}

.glue {
  flex-grow: 1;
  flex-shrink: 1;
}
```

https://jsfiddle.net/evhxcym0/14/

ronkok:
Wow. That HTML is a *lot* simpler than what we get now from `makeVList`. 

IE apparently has a number of flex bugs. Looking at the list, though, they don't seem much more difficult to work around than the stuff that `makeVList` currently does to work within Safari.

kevinbarabash:
My hope is that the IE flex bugs aren't too bad.  We should probably make a small test suite of different layouts using this pattern and try it out on IE to see what breaks.  I'm also curious to see how our stretchy SVG interacts with this.

ylemkimon:
Unfortunately, IE9 doesn't support `flex`, and we should drop IE9 support first.

kevinbarabash:
Khan Academy has dropped support for IE9 and IE10.  I'm interested how others feel about dropping IE9 support.

[...]

PR #1919, opened 2019-04-01
Fix laps. Drop support for IE 9 and 10

ronkok:
This PR fixes issue #1867. The problem was actually present for all the laps, so this PR now prevents Safari from rendering a visible width on `\mathllap`, `\mathrlap`, `\mathclap`, `\llap`, `\rlap`, and `\clap`.

I used CSS `flex` to achieve this result. If we wish to merge this PR, we have to drop support for IE 9 and 10. IE 10 might work, but [caniuse](https://caniuse.com/#search=flex) reports many IE 10 bugs with flex. So I've edited the documentation to say that KaTeX supports IE 11.

Let me know if this drop in IE support is not acceptable.

cc: @kevinbarabash


docs/browser.md
```diff
@@ -2,7 +2,7 @@
 id: browser
 title: Browser
 ---
-> KaTeX supports all major browsers, including Chrome, Safari, Firefox, Opera, Edge, and IE 9–11.
+> KaTeX supports all major browsers, including Chrome, Safari, Firefox, Opera, Edge, and IE 11.
```
ylemkimon: I think we can support IE 10, as there is only few caveats:
>If you have to support IE 10, the best solution is to *always* set an explicit `flex-shrink` value on all of your flex items, or to always use the longhand form (rather than the shorthand) in `flex` declarations to avoid the gotchas shown in the table above. Demo [6.1.a](https://codepen.io/philipwalton/pen/myOYqW) shows how not setting any flexibility properties causes an error, and demo [6.2.a](https://codepen.io/philipwalton/pen/zvvQdB) shows how using `flex: 1` can have the same problem.

ylemkimon:
Related: #1409.

I think we should release a major version with pending PRs merged before dropping the support.

docs/browser.md
ronkok: If you want to do the work it would take to support IE 10, feel free. I don't intend to waste my time chasing the whimsies of a browser with usage that is 0.06% and falling.

src/functions/lap.js
```diff
@@ -44,19 +44,16 @@ defineFunction({
...
-        node = buildCommon.makeVList({
-            positionType: "firstBaseline",
-            children: [{type: "elem", elem: node}],
-        }, options);
-
-        // Get the horizontal spacing correct relative to adjacent items.
-        return buildCommon.makeSpan(["mord"], [node], options);
+        // This code resolves issue #1234
+        node = buildCommon.makeSpan(["thinbox"], [node], options);
+        return buildCommon.makeSpan(["mord", "vbox"], [node], options);
```
kevinbarabash: Are there other commands we can simplify in similar ways?

docs/browser.md
kevinbarabash: In terms of how we support browsers we could look at tools like [autoprefixer](https://github.com/postcss/autoprefixer).  It has support for browserslist which we're already using.  As specific flexbox bugs in older browsers, that autoprefixer page suggests some possible [polyfills](https://github.com/postcss/autoprefixer#does-it-add-polyfills).  It looks like [postcss-preset-env](https://github.com/csstools/postcss-preset-env) does both auto-prefixing and polyfilling.  I'm not sure how difficult it would be set this up, but it would allow us to use flexbox without having to worry about the details on older browsers.

src/katex.less
```diff
@@ -207,6 +207,34 @@
...
+        display: inline-flex;
+        flex-direction: row;
+        width: 0;
+        max-width: 0; // necessary for Safari
+    }
+
+    .glue {
+        flex-grow: 1;
+        flex-shrink: 1;
+    }
```
kevinbarabash: This doesn't seem to be used anywhere.

docs/browser.md
ronkok: > we could look at tools like ...

There's that pronoun again, "we".  If you want someone to research polyfills for IE 9 and IE 10, you'll have to find someone to do that. It won't be me.

src/functions/lap.js
ronkok: The big opportunity is in `makeVList()`. It writes a lot of HTML and you demonstrated in issue #1409 that it could be made much less verbose by use of `flex`.

src/katex.less
ronkok: Correct. `.glue` is not used here. It would be necessary for use in `makeVList()`. I was looking forward a little bit here. Do you want me to take it out?

docs/browser.md
kevinbarabash: @ronkok sorry, I didn't mean "we" to imply "you".  I'll open a new issue for the autoprefixer/postcss-preset-env proposal.  I do think the autoprefixer would simplify parts of this PR, but I think this PR can go forward without it.

src/functions/lap.js
kevinbarabash: I totally forgot about that, thanks for the reminder. 😅

src/katex.less
kevinbarabash: Yeah, let's take it out for now.  It'll be easy enough to add back later once we need it.

kevinbarabash:
> I think we should release a major version with pending PRs merged before dropping the support.

@ylemkimon were you thinking a 1.0 release before dropping support?  We could also defer merging this until https://github.com/KaTeX/KaTeX/issues/1946 gets done.

PR #1921, opened 2019-04-02
Support \textup and \textmd

### LINKED ISSUE #1909: Support \textup
`\textup{text}` renders text that is always upright (i.e. not italicised), regardless of its content.

edemaine:
This PR adds support for `\textup` and `\textmd`, which are the inverses of `\textit` and `\textbf`.

Unlike bare `\text`, they result in `textup` and `textmd` classes being applied, but those have no CSS rules, so they act the same as bare `\text`.

Fixes #1909.

Perhaps notably, this PR does not add support for `\upshape` or `\mdseries`, as we don't currently support `\itshape` or `\bfseries` (could be added easily in another PR). But it also doesn't support `\up` and `\md`, despite existing support `\it` and `\bf`. Not sure if anyone ever uses those old-style commands, though.  See [this cheat sheet](https://www.harding.edu/lmurray/latex_files/latexref/latexcheatsheet.pdf) for example.

### DIFF
--- a/docs/support_table.md
+++ b/docs/support_table.md
@@ -1007,6 +1007,7 @@ table td {
 |\textgreater|$\text{\textgreater}$|`\text{\textgreater}`|
 |\textit|$\textit{AaBb}$|`\textit{AaBb}`|
 |\textless|$\text{\textless}$|`\text{\textless}`|
+|\textmd|$\textmd{AaBb123}$|`\textmd{AaBb123}`|
 |\textnormal|$\textnormal{AB}$|`\textnormal{AB}`|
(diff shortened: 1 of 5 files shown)


edemaine:
I was just reminded that `support_table.md` conveniently is keeping track of commands we don't support in addition to those that we do.  In https://github.com/KaTeX/KaTeX/pull/1921/commits/8d03e440ba062ffc4efe5c8ea842acf044349ee6, I added all the font commands from that quick reference guide that we don't support, so it's easier to keep track of.  If someone could review that part, I think we can merge this.

PR #1922, opened 2019-04-02
Improve MathML for corners

ronkok:
This PR corrects the MathML for `\ulcorner`, `\urcorner`, `\llcorner`, and `\lrcorner`. The KaTeX fonts have the glyphs for those functions located at the box drawing corners. This PR points the MathML at the Unicode code points for corners in the Miscellaneous Technical part of the Unicode table.


src/macros.js
```diff
@@ -394,6 +394,13 @@ defineMacro("\u00A9", "\\copyright");
 defineMacro("\u00AE", "\\textregistered");
 defineMacro("\uFE0F", "\\textregistered");
 
+// The KaTeX fonts have corners at codepoints that don't match Unicode.
+// For MathML purposes, use the Unicode code point.
+defineMacro("\\ulcorner", "\\html@mathml{\\@ulcorner}{\\mathop{\\char\"231c}}");
+defineMacro("\\urcorner", "\\html@mathml{\\@urcorner}{\\mathop{\\char\"231d}}");
+defineMacro("\\llcorner", "\\html@mathml{\\@llcorner}{\\mathop{\\char\"231e}}");
+defineMacro("\\lrcorner", "\\html@mathml{\\@lrcorner}{\\mathop{\\char\"231f}}");
```
ylemkimon: If we have many symbols like this, maybe we can add another argument to `defineSymbol`.

ylemkimon:
In this case, I think we can update their codepoints in https://github.com/KaTeX/katex-fonts/blob/master/src/fonts/OTF/TeX/makeFF#L871-L880.

ronkok:
> we can update their codepoints

That would be a better solution, but I have never mastered Docker or the KaTeX font tool chain, so I cannot update those codepoints. And some font issues go unresolved [for years](https://github.com/KaTeX/KaTeX/issues/564). So unless you are planning to do that font work yourself, I'd suggest that this PR is a good short term solution.

kevinbarabash:
I think we should fix the codepoints in our fonts but I don't think it should block this change.  I opened a ticket to track this.

kevinbarabash:
Once the codepoints are updated in the fonts though this PR shouldn't be necessary, right?

ronkok:
If the font is corrected, then the macros of this PR would no longer be necessary, but the [relevant lines](https://github.com/KaTeX/KaTeX/blob/master/src/symbols.js#L312) in `symbols.js` would need to be edited to be in sync with the code points.

PR #1923, opened 2019-04-04
Improve MathML for \not

ronkok:
This PR resolves a long-standing TODO item by implementing `\not` in MathML. The code employed is similar to the code used for HTML `\not` prior to PR #1497.

### DIFF
--- a/src/buildMathML.js
+++ b/src/buildMathML.js
@@ -163,31 +163,6 @@ export const buildExpression = function(
         groups.push(group);
         lastGroup = group;
     }
-
-    // Combine \not with mrels and mords
-    for (let i = (groups.length - 2); i >= 0; i--) {
-        const group = groups[i];
-        if (group instanceof MathNode && group.type === 'mi' &&
-                group.children.length === 1) {
-            const child = group.children[0];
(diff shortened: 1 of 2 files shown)


src/buildMathML.js
```diff
@@ -151,7 +151,29 @@ export const buildExpression = function(
         lastGroup = group;
```
ylemkimon: I think we can loop only once with less or same number of checks:
```js
            // Combine \not with mrels and mords
/*L#147*/   } else if (lastGroup instanceof MathNode && lastGroup.type === 'mi' &&
                lastGroup.children.length === 1) {
                const lastChild = lastGroup.children[0];
                if (lastChild instanceof TextNode && lastChild.text === '\u0338' &&
                    group instanceof MathNode && (group.type === 'mo' ||
                    group.type === 'mi' || group.type === 'mn')) {
                    const child = group.children[0];
                    if (child instanceof TextNode && child.text.length > 0) {
                        // Overlay with combining character long solidus
                        child.text = child.text.slice(0, 1) + "\u0338" + child.text.slice(1);
                        groups.pop();
                    }
                }
            }
```

ronkok: Done.

Issue #1873, opened 2019-03-16
\pmb is awkward on Safari


ylemkimon:
![Pmb-safari2](https://user-images.githubusercontent.com/888148/54478510-4bf14080-4856-11e9-8c35-066bdc2fae7a.png)

CC @ronkok

ronkok:
@ylemkimon There may not be much I can do. Right now, I don't have much access to Safari. Next week some new iPads will be available. I'll buy one and see what is possible.

ronkok:
I'll take a look at this after PR #1919 lands.

PR #1924, opened 2019-04-04
Fix \pmb

ronkok:
Fixes issue #1873.


edemaine:
Does this mean that `\raisebox` is generally broken in Safari? By just a small amount or larger amount? I guess I'm wondering, should we just fix `\raisebox`/`makeVList` instead of changing the definition of `\pmb`?

ronkok:
I agree that the long term solution is to fix `makeVList`. I don't know how to do that quickly and I am not positive that I have an idea that works at all. I guess I should describe the problem.

When PR #768 rewrote `makeVList`, it had to overcome Safari problems. One was that the code needed a zero-width span with which to define depth. Safari refused to render a zero-width span and the work around was to define a span that was 2px wide and then give it `margin-right: -2px;` to correct the position. That works well until you want to run adjacent very narrow spans next to each other. Then the actual width of the span matters.

It's possible that `makeVList` can be written with CSS flex as @kevinbarabash proposed in issue #1409. @ylemkimon is of the opinion that before using flex we should first have a major release that deprecates old IE versions. So the flex option is not a quick fix.

And we need to remember why PR #768 was written in the first place. Prior to that PR, in math that contained a tall element, the HTML font ascent and line height distances would occasionally create something that looked like a large margin above the KaTeX rendering.  PR #768 fixed that problem by putting the math into a `display: inline-table;` span. I don't know if CSS flex accomplishes the same thing.

If CSS flex doesn't work, I have one more idea to try. Remember I said that "Safari refused to render a zero-width span". That may be fixable by declaring `max-width: 0;`. I don't know if it is possible on the relevant span because that span also has some other hacks going on to prevent a Chrome problem. And I'd really like the CSS flex idea to work first. The HTML would be much simpler.

src/macros.js
```diff
@@ -655,13 +655,10 @@ defineMacro("\\mod", "\\allowbreak" +
...
-// Ref: a rather lengthy macro in ambsy.sty
-defineMacro("\\pmb", "\\html@mathml{\\@binrel{#1}{" +
-    "\\mathrlap{#1}" +
-    "\\mathrlap{\\mkern0.4mu\\raisebox{0.4mu}{$#1$}}" +
-    "{\\mkern0.8mu#1}" +
-    "}}{\\mathbf{#1}}");
+// The version in ambsy.sty works by typesetting three copies of the argument
+// with small offsets. We use two copies. We omit the vertical offset because
+// of rendering problems that makeVList encounters in Safari.
+defineMacro("\\pmb", "\\@binrel{#1}{\\mathrlap{#1}\\kern0.5px#1}");
```
ylemkimon: What about MathML output?

ronkok: This code produces MathML that also renders as a simulated bold.  In the screenshot below, the three versions are from MathJax, KaTeX, and KaTeX-MathML:
pmb

ylemkimon: @ronkok Currently MathML is used for accessibility, and I think it'd be better to render only once, maybe with `\mathbf` like before.

ronkok: Yes, MathML is currently used only for accessibility, but that will only be true until #1928 is resolved. More importantly, `\mathbf` is currently ineffective in KaTeX MathML. You can verify this by looking at the entry for `\pmb` in http://webmathcomparison.net/#p. 

I haven't yet investigated why this occurs, but I suspect some sort of bug similar to what I address in PR #1929, which applies classes such as `\mathbin` in MathML.

ronkok: @ylemkimon I have revised the MathML as you requested, using `\mathbf`. So this PR can be merged. But I consider the larger matter unfinished at least until `\mathbf` works in MathML.

MONANREN:
DeclareRobustCommand{\LaTeX}{L\kern-.36em%
//         {\sbox\z@ T%
//          \vbox to\ht\z@{\hbox{\check@mathfonts
//                               \fontsize\sf@size\z@
//                               \math@fontsfalse\selectfont
//                               A}%

PR #1929, opened 2019-04-07
Improve MathML for classes

ronkok:
Implement classes such as `\mathbin` and `\mathrel` in MathML.


test/__snapshots__/mathml-spec.js.snap
```diff
@@ -414,32 +423,34 @@ exports[`A MathML builder should render boldsymbol with the correct mathvariants
...
-        </mi>
-        <mi mathvariant="bold-italic">
-          ı
-        </mi>
-        <mo mathvariant="bold-italic">
-          +
-        </mo>
-      </mrow>
+      <mi>
+        <mrow>
```
kevinbarabash: This change doesn't seem right.

test/__snapshots__/mathml-spec.js.snap
```diff
@@ -8,15 +8,21 @@ exports[`A MathML builder \\html@mathml makes clean symbols 1`] = `
...
-        ∉
-      </mi>
-      <mi mathvariant="normal">
-        ≘
-      </mi>
+      <mo>
+        <mi mathvariant="normal">
+          ≠
+        </mi>
+      </mo>
```
kevinbarabash: Is `<mo><mi>...</mi></mo>` an artifact of how these operators are described in the TeX code?  Would `<mo mathvariant="normal">≠</mo>` be more natural?

src/functions/mclass.js
```diff
@@ -18,7 +18,30 @@ function htmlBuilder(group: ParseNode<"mclass">, options) {
...
+    let node;
+    switch (group.mclass) {
+        case "minner":
+            return mathMLTree.newDocumentFragment(inner);
+        case "mord":
+            return new mathMLTree.MathNode("mi", inner);
+        default:
+            node = new mathMLTree.MathNode("mo", inner);
+            // Set spacing based on what is the most likely adjacent atom type.
+            // See TeXbook p170.
```
kevinbarabash: Thanks for linking to the source.

test/__snapshots__/mathml-spec.js.snap
ronkok: I agree. This one may take some work. I'll get back to you tomorrow.

test/__snapshots__/mathml-spec.js.snap
ronkok: Re: `<mo><mi>...</mi></mo>`

Take, for example, `\neq`. It is defined by a macro:
```
defineMacro("\\neq", "\\html@mathml{\\mathrel{\\not=}}{\\mathrel{\\char`≠}}");
```

Notice that the MathML part of that macro employs `\char ≠`. `\char`  is a way to define a character in a macro and avoid an infinite loop of macro expansion. So far, so good. But `\char` always expands its argument to a textord. That accounts for the `<mi>`. And so the macro wraps it with `\mathrel` to get the correct spacing.

> Would `<mo mathvariant="normal">≠</mo>` be more natural?

Just `<mo>` would be even more natural. To get that, we would have to add some code to `char.js`. At present, it returns everything as a `textord`. But it takes the Unicode code point as an argument. It could compare that code point to some ranges and return everything from, say, U+2238 to U+2294 ([Ref](https://unicode-table.com/en/#mathematical-operators)) as a `mrel`. Then the macro's wrapper `\mathrel` would be unnecessary. There would be similar ranges for `bin`, etc. It could end up as quite an extensive dictionary, like [this one](https://www.w3.org/TR/MathML3/appendixc.html#oper-dict.entries-table).

Is it worth the trouble? Would you like me to make such a modification to `char.js`?

ronkok: A thought occurs to me. In `mclass.js`, we could test if the argument is a `<mi>` and also apply the `utils.isCharacterBox` test. If an argument meets both of those tests, we could change the element from `<mi>` to `<mo>` instead of wrapping it with `<mo>`. `flow` probably will object to this sort of thing, but the end result is better.  I'll look into it.

kevinbarabash: Having to include a large lookup table would be unfortunate.  The `isCharacterBox` is worth a try.  Are there any situations where `isCharacterBox` would return a false positive for this use case?

edemaine: I agree that the code implementing `\mathrel` (etc.) and returning `<mo>` is the place to do this kind of simplification. I think it makes sense to do this, as `\mathrel` (etc.) are explicitly trying to make this kind of type cast. In all cases, it will still return an `<mo>` tag, so I don't see any problems with Flow...

test/__snapshots__/mathml-spec.js.snap
ronkok: @kevinbarabash Having looked closer at this, I now think it is correct. `\boldsymbol` makes a call to `\binrel` which return a mathord for argument that is not a single character. So to see this group wrapped by a `<mi>` is what we should expect.

test/__snapshots__/mathml-spec.js.snap
ronkok: I believe that I have now resolved this issue and that this PR is ready for back check.

ronkok:
@kevinbarabash Thank you for the review, and thank you for the commit. I now try to avoid the children of an argument defined by a `flow` type. But to address this problem, I could think of no other way to get it done. So again, thanks for your help.

PR #1930, opened 2019-04-07
Improve MathML for \Bbbk

ronkok:
Implements `\Bbbk` in MathML.


edemaine:
According to [this stackexchange](https://tex.stackexchange.com/questions/161159/math-times-pro-2-and-txfontsb), this new definition is the one used by `mtpro2.sty`.  They also look identical in current KaTeX:

```latex
\Bbbk\mathbb k
```
image

I think, once merged, we should remove this symbol from the AMS font. @ylemkimon

PR #1932, opened 2019-04-10
Include extensions mhchem & copy-tex in home-page

### LINKED ISSUE #1866: Mhchem not loaded on katex.org
Hi!

I tried to enable mhchem extension that shipped in the new 0.10.1 version on my site. It seems to work although some rendering seems a bit odd (will make another issue for that) so I wanted compare with katex.org test page. It seems that mhchem is not loaded there. I realize that might be by design since mhchem is an extension, but maybe in this case it is worth it to add?

mothas:
This PR fixes #1866. As of now, the interactive demo box in [home page](https://katex.org/) does not load extensions `copy-tex` and `mhchem`. With this fix, users will be able to try the following in home-page:
* copy the rendered KaTex expressions as latex source
* try out chemical equations


danielhollas:
Forgive me if I am completely wrong here, I was looking into this as well some time ago. Shouldn't this be handled in the same way as katex.min.js?
i.e. including only relative path in index.html and modifying `website/lib/build.js`?

mothas:
@edemaine I'll need some guidance here. If I have understood this correctly, I'll have to do the following:
* include `mhchem` and `copy-tex` files in `website/static` folder
  * which version should I copy over? The latest?
* modify `index.html` to use relative paths, instead of absolute paths
* modify `website/lib/build.js` to include `mhchem` and `copy-tex` files

I thought I'll check with you first before making changes, so that I don't end up using up your time.

danielhollas:
@mothas at least when I tried to do it locally, your task list should do the trick. Except for one thing: I do not think you need to copy the files over to the `static` directory manually. They are copied automatically from your local build from `../dist` when you run `yarn start`, if you modify the `lib/build.js` accordingly. In fact, what you need to do is to add these extra files into `.gitignore`, as you can see from the last two lines in that file.

btw: don't forget to modify the regex on line 23 in `build.js` for production deploy. I guess that one cannot be tested locally and you'll need to push here and hope for the best. :-)

danielhollas:
@mothas one small github-related tip that I learned recently 🙂 
If you delete the word _issue_ in your first post in this PR like this:
      
   This PR fixes ~~issue~~ #1866.

...the issue #1866 will be closed automagically when this PR gets merged to master. See https://help.github.com/en/articles/closing-issues-using-keywords

mothas:
@danielhollas Thanks for the great tip! I have made changes like you mentioned. Let me know if there are any more review comments.

mothas:
@danielhollas Not sure if I completely understood your question. I'm assuming `dist` folder will have the latest version. Maybe we should update the content in `dist` folder?

danielhollas:
If you compare the source of the preview and actual katex.org, it uses the latest official build served via CDN, not the local files from dist. (the substitution in index.html is handled at the bottom of lib/build.js where you updated regex that handles this). The question was meant more to the maintainers of the repo.

danielhollas:
Hmm, if I understand the code correctly, you could perhaps test the production version by running:
```
export CONTEXT=production
cd website && yarn start
```
(this will modify the index.html in your repo, which you should revert later)

mothas:
@danielhollas You are right! When `CONTEXT=production`, `index.html` was edited to use CDN links, instead of relative paths. That resulted in `mhchem` and `copy-tex` not loading correctly(refer screenshot below). And hence they didn't work.

Will this be a problem in production? Or is it a matter or fixing the version numbers?

KaTeX browser console error

website/lib/build.js
```diff
@@ -16,11 +16,17 @@ fs.copySync('../dist/katex.min.js', 'static/static/katex.min.js');
...
+fs.copySync('../dist/contrib/copy-tex.min.js', 'static/static/copy-tex.min.js');
+fs.copySync('../dist/contrib/mhchem.min.js', 'static/static/mhchem.min.js');
+
+
 // use KaTeX from CDN on the main page for Netlify production deploy
 if (process.env.CONTEXT === 'production') {
     const version = require('../versions.json')[0];
     let indexHtml = fs.readFileSync('pages/index.html', 'utf8');
-    indexHtml = indexHtml.replace(/(["'])static\/(katex|fonts)/g,
+    indexHtml = indexHtml.replace(/(["'])static\/(katex|fonts|copy-tex|mhchem)/g,
```
danielhollas: You'll need a separate `.replace()` here, because the correct CDN URL for plugins is slightly different. For mhchem, it is 
`https://cdn.jsdelivr.net/npm/katex@0.10.1/dist/contrib/mhchem.min.js`
as explained here:
https://github.com/KaTeX/KaTeX/tree/master/contrib/mhchem

mothas: Included these changes.

mothas:
@danielhollas Let me know if you have more comments.

website/lib/build.js
```diff
@@ -16,11 +16,20 @@ fs.copySync('../dist/katex.min.js', 'static/static/katex.min.js');
...
+fs.copySync('../dist/contrib/mhchem.min.js', 'static/static/mhchem.min.js');
+
+
 // use KaTeX from CDN on the main page for Netlify production deploy
 if (process.env.CONTEXT === 'production') {
     const version = require('../versions.json')[0];
     let indexHtml = fs.readFileSync('pages/index.html', 'utf8');
     indexHtml = indexHtml.replace(/(["'])static\/(katex|fonts)/g,
         `$1https://cdn.jsdelivr.net/npm/katex@${version}/dist/$2`);
+        //CDN pathway different for `copy-tex` and `mhchem`
```
danielhollas: very very minor: You could chain the replaces like this:
```suggestion
        `$1https://cdn.jsdelivr.net/npm/katex@${version}/dist/$2`)
         .replace(/(["'])static\/(copy-tex|mhchem)/g,
            `$1https://cdn.jsdelivr.net/npm/katex@${version}/dist/contrib/$2`);
```

mothas: I did think about chaining it. But I realized it might more readable if I break it up because each replace is quite long. Any thoughts?

kevinbarabash: I'm okay either way.

kevinbarabash:
@edemaine it looks like https://github.com/KaTeX/KaTeX/blob/master/release.sh#L138-L139 will update all `/static/` paths in the docs to use the the CDN versions.

edemaine:
I agree this looks good! Thanks again.

To make the substitution symmetric, we could (maybe) copy the contributed libraries into `static/static/contrib`, but I'm not sure it's worth the change.

mothas:
> @edemaine it looks like https://github.com/KaTeX/KaTeX/blob/master/release.sh#L138-L139 will update all `/static/` paths in the docs to use the the CDN versions and update-sri.js will process anything that starts with `https://cdn.jsdelivr.net/npm/katex@` so I think we're good.

I was wondering if this will handle the newly added `mhchem` and `copy-tex` packages. These packages have an additional `/contrib/` in the path. Just checking if there are changes to be made on `release.sh` to accommodate this.

kevinbarabash:
I'm in the process of doing a release so we may as well merge this and see what happens.

kevinbarabash:
Looking at https://github.com/KaTeX/KaTeX/pull/1961/files#diff-589987ee7f5e5f193621aa08073028e0 and the source of update-sri.js it looks like we needed a `integrity` attribute in the script tag.  I'll add a placeholder one manually and then run `update-sri.js` on the `contrib/mhchem/README.md` to update it.

kevinbarabash:
I've updated #1961 so that it has a integrity hash in the mhchem README.md.

ylemkimon:
[clabot:check]

Issue #750, opened 2017-06-29
Use macros instead of colorMap for KA specific colors


kevinbarabash:
See TODO in Options.js

edemaine:
These would most naturally be replaced by `\definecolor` calls (see [this description](https://en.wikibooks.org/wiki/LaTeX/Colors#Defining_new_colors)).  Unless you want to replace `\textcolor{katex-blue}{...}` with `\textcolor{\katexblue}{...}` and use a macro `\katexblue` (but sadly can't have a hyphen in its name).  By analogy to the `macros` option we currently have, we could have a `colors` option that provides a mapping of colors to RGB.  This map could then be moved to the KA project where KaTeX gets called, or it could be the default settings which uses can add additional colors to.

kevinbarabash:
Cool!  I didn't know about `\definecolor`.  I think implementing that to support at least the RGB and HTML models makes sense since we already have a `colorMap` object.

danielhollas:
I wanted to give this one a try, although I am not sure this a good first issue :-)
My idea was to create a customColorMap object that would look like this:
```js
let customColorMap = {
  "myColor": "#ffaaff"
}
```

and modify `parseColorGroup` in Parser.js to take into account this object:
https://github.com/KaTeX/KaTeX/blob/b9deec098f35919b9ad77baa6b11b7593473d488/src/Parser.js#L649
```
let inputColor = res.text;
if (colorMap.hasOwnProperty(inputColor)) {
     inputColor = colorMap[inputColor];
}
```
So far so good and this works, but now I don't know how to properly define a function whose sole purpose would be to extend the customColorMap object and nothing else (no htmlbuilder etc). Is there a precedent here that I could follow?

edemaine:
@danielhollas I believe this map would most naturally be an addition to the (recent) `Namespace` class/js file.  Looking at the definition of `\definecolor`, it looks like `\definecolor{red}{rgb}{1,0,0}` defines `\color@red` by calling `\color@rgb{\string\color@red}{1,0,0}`.  We could just simulate that behavior, or at least use the macro mapping by setting `\color@red` to the color.  Alternatively, you could try adding a special mapping for colors in the Namespace, though then you'll need to do more work when namespaces get pushed or popped.

Could you test whether `{\definecolor{myred}{rgb}{1,0,0}}\color{myred}` fails as I expect it should, i.e., the definition of color is local to the group?

danielhollas:
[...]

PR #1933, opened 2019-04-11
Move custom colors used by Khan into macros.js

danielhollas:
This is a step towards resolving #28. I moved the color definitions from `Options.js` and `functions/color.js` into `macro.js`.


danielhollas:
I wonder whether the color mapping should not be in a separate JSON file. There are other places in the Khan Academy codebase where these are used so this might be valuable...

kevinbarabash:
@danielhollas thanks for the PR.  This saves a little space so I think we should go ahead with this PR.  

Long term though I think we should move these custom colors out of the KaTeX codebase altogether.  The downside of doing so right now is that we'd have to reparse these colors each time we render any math on khanacademy.org on in the mobile apps.  I think we should implement `katex.globalEval` as suggested in #750 before moving these custom colors out of KaTeX.

src/Options.js
```diff
@@ -299,81 +299,13 @@ class Options {
...
+     * Gets the CSS color of the current options object
      */
     getColor(): string | void {
         if (this.phantom) {
             return "transparent";
-        } else if (
-            this.color != null &&
-            Options.colorMap.hasOwnProperty(this.color)
-        ) {
-            return Options.colorMap[this.color];
```
kevinbarabash: Nice

src/functions/color.js
```diff
@@ -55,42 +55,6 @@ defineFunction({
     mathmlBuilder,
 });
 
-// TODO(kevinb): define these using macros
-defineFunction({
```
kevinbarabash: yay

PR #1947, opened 2019-04-20
Add symbol double square brackets

ben-dyer:
I added a symbols for left/right double square brackets, unicode U+27E6 and U+27E7. For the properties I mimicked similar symbols such as floor/ceiling which were already included.

I tested with `yarn start` and got the error `No character metrics for '⟦' in style 'Main-Regular'`. I tried to see if I should address this and found the issues https://github.com/KaTeX/KaTeX/issues/374 and https://github.com/KaTeX/KaTeX/issues/1817 which gave me the impression I could overlook it.

The names `\llbrack` and `\rrbrack` are chosen to be consistent with KaTeX's `\lbrack` and `\rbrack` for single square brackets. They differ slightly from those found in the `stmaryrd` package for LaTeX (see [Double square brackets](https://tex.stackexchange.com/questions/107252/double-square-brackets) @ TeX SE) which uses `\llbracket` and `\rrbracket`.

I would also like to make them into delimiters to work with `\left` and `\right`, but didn't see similar unicode symbols to the upper/lower corner and extension symbols for single square brackets (U+23A1, U+23A2, U+23A3).

### DIFF
--- a/src/symbols.js
+++ b/src/symbols.js
@@ -625,6 +625,8 @@ defineSymbol(math, main, open, "[", "\\lbrack");
 defineSymbol(text, main, textord, "[", "\\lbrack");
 defineSymbol(math, main, close, "]", "\\rbrack");
 defineSymbol(text, main, textord, "]", "\\rbrack");
+defineSymbol(math, main, open, "\u27e6", "\\llbrack", true);
+defineSymbol(math, main, close, "\u27e7", "\\rrbrack", true);
 defineSymbol(math, main, open, "(", "\\lparen");
 defineSymbol(math, main, close, ")", "\\rparen");
 defineSymbol(text, main, textord, "<", "\\textless"); // in T1 fontenc
--- a/src/symbols.js
+++ b/src/symbols.js
@@ -625,8 +625,8 @@ defineSymbol(math, main, open, "[", "\\lbrack");
 defineSymbol(text, main, textord, "[", "\\lbrack");
 defineSymbol(math, main, close, "]", "\\rbrack");
 defineSymbol(text, main, textord, "]", "\\rbrack");
-defineSymbol(math, main, open, "\u27e6", "\\llbrack", true);
-defineSymbol(math, main, close, "\u27e7", "\\rrbrack", true);
+defineSymbol(math, main, open, "\u27e6", "\\llbracket", true);
+defineSymbol(math, main, close, "\u27e7", "\\rrbracket", true);
 defineSymbol(math, main, open, "(", "\\lparen");
 defineSymbol(math, main, close, ")", "\\rparen");
 defineSymbol(text, main, textord, "<", "\\textless"); // in T1 fontenc
--- a/docs/support_table.md
+++ b/docs/support_table.md
@@ -570,6 +570,7 @@ table td {
 |\limsup|$\limsup$||
 |\ll|$\ll$||
 |\llap|${=}\llap{/\,}$|`{=}\llap{/\,}`|
+|\llbracket|$\llbracket$||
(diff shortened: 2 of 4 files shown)


ronkok:
@ben-dyer Nice job finding a LaTeX function name for that character. I once looked for such a function name and did not find it.

The KaTeX convention has consistently been to use pre-existing LaTeX function names whenever possible rather than inventing new function names. I think that using the names `\llbracket` and `\rrbracket` would be a better choice.

Also, you can avoid the error message by means of a somewhat sneaky macro.  Take a look at the code in [macros.js](https://github.com/KaTeX/KaTeX/blob/f948947736fa0674c5edd6c5ca30fcbccdac75fc/src/macros.js#L836-L843). I suggest we modify that code so that it also includes your proposal. Something like this:
```
//////////////////////////////////////////////////////////////////////
// stmaryrd and semantic

// The stmaryrd and semantic packages render the next four items by calling a
// glyph. Those glyphs do not exist in the KaTeX fonts. Hence the macros.

defineMacro("\\llbracket", "\\html@mathml{" +
    "\\mathopen{[\\mkern-3.2mu[}}" +
    "{\\mathopen{\\char`\u27e6}}");
defineMacro("\\rrbracket", "\\html@mathml{" +
    "\\mathclose{]\\mkern-3.2mu]}}" +
    "{\\mathclose{\\char`\u27e7}}");

defineMacro("\u27e6", "\\llbracket");  // blackboard bold [
defineMacro("\u27e7", "\\rrbracket"); // blackboard bold ]
```

ronkok:
Also, could you add documentation to this PR? That would entail editing files `docs/support_table.md` and `docs/supported.md`.

ben-dyer:
Thank you for such helpful feedback! Yes I'll get working on these updates.

ice1000:
This sounds incredibly good! Can this be included in the coming release?

ice1000:
Can there be double-square braces (or even parenthesis)?

ronkok:
@ice1000 We could probably use a similar macro trick to emulate double-struck braces. Something like: `\mathopen{\{\mkern-4.3mu|}`. I don't see any such combination that would work for parenthesis. It looks like those will have to wait until someone can add new glyphs to the KaTeX fonts.

Until then, you can still use the Unicode characters `⦅` and `⦆`. That technique has some drawbacks. KaTeX will get a glyph from a system font and put up a console warning. Also the height of the glyph might not be ideal.  But it might be better than nothing.

ice1000:
@ronkok I've added this as a PR, #1953.
Could you please take a look at that?

ben-dyer:
@ronkok I updated the documentation and changed the names to \llbracket and \rrbracket. I also made the change to macros as suggested, and verified it's working and the error message is gone.

ronkok:
Looks good to me.

ronkok:
@ben-dyer Nice work. Thank you for the contribution.

ice1000:
Closed??

ronkok:
@ice1000 My bad. Fixed now.

src/symbols.js
```diff
@@ -625,6 +625,8 @@ defineSymbol(math, main, open, "[", "\\lbrack");
 defineSymbol(text, main, textord, "[", "\\lbrack");
 defineSymbol(math, main, close, "]", "\\rbrack");
 defineSymbol(text, main, textord, "]", "\\rbrack");
+defineSymbol(math, main, open, "\u27e6", "\\llbracket", true);
+defineSymbol(math, main, close, "\u27e7", "\\rrbracket", true);
```
ylemkimon: This results in duplicate definition of symbols. There's no need for `defineSymbol`s.

PR #1951, opened 2019-04-23
disable \includegraphics until trust settings is merged

kevinbarabash:
This is only temporary and I thought that simply disabling the this function would be easier that removing it and adding it back in later.

### DIFF
--- a/docs/support_table.md
+++ b/docs/support_table.md
@@ -479,7 +479,6 @@ table td {
 |\impliedby|$P\impliedby Q$|`P\impliedby Q`|
 |\implies|$P\implies Q$|`P\implies Q`|
 |\in|$\in$||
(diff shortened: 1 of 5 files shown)


edemaine:
For 0.10.1, we did [a commit](https://github.com/KaTeX/KaTeX/commit/d4383d03562535c3d4e98559894288617226564f) on the v0.10.1 release branch that reverted the change, without affecting master.  I'm not sure which is better:
1. Merge this with master so releasing is easier.
2. Retarget this to v0.10.2 release branch.

I'm fine either way. The disabling looks good.

kevinbarabash:
@edemaine I forgot about us disabling `\includegraphics` once before.  Merging this commit with `master` seems the best way to avoid confusion in the future.

PR #1953, opened 2019-04-23
Support double-square curly braces

ice1000:
Signed-off-by: ice1000 <ice1000kotlin@foxmail.com>

Addressed my own comment: https://github.com/KaTeX/KaTeX/pull/1947#issuecomment-485986880

PR #1954, opened 2019-04-25
Remove redundant symbols from #1947

edemaine:
https://github.com/KaTeX/KaTeX/pull/1947#pullrequestreview-230534608

I also notice a lack of tests...


ben-dyer:
@edemaine Thank you, I will add a test. I think it will just be an expect().toBuild() similar to what was done #1953 for double-braces.

PR #1965, opened 2019-05-03
Improve MathML fonts

ronkok:
Enable commands such as `\mathbf` and `\mathbb` in MathML.


test/katex-spec.js
```diff
@@ -1945,9 +1939,9 @@ describe("A MathML font tree-builder", function() {
         expect(markup).toContain("<mi mathvariant=\"sans-serif\">A</mi>");
         expect(markup).toContain("<mi mathvariant=\"sans-serif\">x</mi>");
         expect(markup).toContain("<mn mathvariant=\"sans-serif\">2</mn>");
-        expect(markup).toContain("<mi>\u03c9</mi>");                            // \omega
+        expect(markup).toContain("<mi mathvariant=\"sans-serif\">\u03c9</mi>"); // \omega
         expect(markup).toContain("<mi mathvariant=\"sans-serif\">\u03A9</mi>"); // \Omega
-        expect(markup).toContain("<mi>\u0131</mi>");                            // \imath
+        expect(markup).toContain("<mi mathvariant=\"sans-serif\">\u0131</mi>"); // \imath
```
kevinbarabash: I don't think we have the CM fonts for some of these characters in these variants so this may be one area where the MathML output is now ahead of the HTML output.

test/__snapshots__/mathml-spec.js.snap
```diff
@@ -185,28 +185,8 @@ exports[`A MathML builder ligatures render properly 1`] = `
         −
       </mo>
       <mtext mathvariant="monospace">
-        \`\`‘Hi----&#x27;&#x27;’
+        \`\`‘Hi----&#x27;&#x27;’\`\`‘Hi----&#x27;&#x27;’
```
kevinbarabash: Why did this snapshot change so much?

kevinbarabash: I think this is because everything has the same `mathvariant` set so our concatenation code is merging all of the `mtext`s into a single `mtext`.

Issue #1928, opened 2019-04-07
Add a way to use MathML rendering


ylemkimon:
Thanks to @ronkok, we now have a quite decent MathML output.

It'd be nice to have a way to easily switch to it.

kevinbarabash:
We'll probably want to have separate screenshot tests for MathML rendering as well.  Right now we always render MathML but hidden.  I wonder if we should switch to rendering HTML by default and then allow people to choose HTML, MathML, or both and leave it up to people to hide MathML if they want to.

kevinbarabash:
This can be closed now that #1966 has been merged.

PR #1966, opened 2019-05-04
Render MathML directly

ronkok:
This PR addresses #1928 by ~~exposing two new functions, `renderMathML()`, and `renderMathMLToString()`~~ adding a rendering option that allows a web site administrator to output KaTeX in HTML only, in MathML only, or in both HTML and MathML (default).  

You can review its effectiveness by using Firefox to view http://webmathcomparison.net/. The KaTeX MathML column on that page now is driven by a build from this branch. I'll be happy to add other tests that might help.


katex.js
```diff
@@ -151,6 +180,19 @@ export default {
      * for sending to the client.
      */
     renderToString,
+    /**
+     * KaTeX error, usually during parsing.
+     */
```
kevinbarabash: This comment exists below.  This copy should be removed.

src/buildTree.js
```diff
@@ -57,4 +57,18 @@ export const buildHTMLTree = function(
...
+    tree: AnyParseNode[],
+    expression: string,
+    settings: Settings,
+): DomSpan {
+    const options = optionsFromSettings(settings);
+    const mathMLNode = buildMathML(tree, expression, options);
+    // Flow doesn't think a DomSpan has children.
+    // $FlowFixMe
+    const node = (buildCommon.makeSpan([], [mathMLNode])).children[0].children[0];
+    node.setAttribute("xmlns", "http://www.w3.org/1998/Math/MathML");
```
kevinbarabash: Does `buildMathML` not include the namespace?  If not, it probably should and then we can remove this line here.

ronkok: > Does `buildMathML` not include the namespace?

Correct. It does not. I'll put the namespace there as you suggest.

ronkok: Done.

ronkok:
Control of the output is now moved from the main KaTeX API to the KaTeX rendering options. This should be good to go.

PR #1973, opened 2019-05-11
Fixed user-select range.

jun-sheaf:
The relative positioning consolidates the select range to the parsed area (as opposed to its parent container for example).

### DIFF
--- a/contrib/copy-tex/copy-tex.css
+++ b/contrib/copy-tex/copy-tex.css
@@ -6,6 +6,7 @@
  */
 .katex,
 .katex-display {
+    position: relative;
     user-select: all;
     -moz-user-select: all;
     -webkit-user-select: all;


kevinbarabash:
@jun-sheaf could you describe the problem a bit more and how to test your fix?  Can you post some screenshots if possible?

jun-sheaf:
Without the additional property, clicking on any LaTeX can sometimes result in the entire parent container being selected. e.g. consider the following
```html
<div>
 Some text
 <span class="katex">
  Some KaTeX elements
 </span>
</div>
```
Upon selecting the `span` element, the entire `div` is selected (more accurately, the click event bubbles till it reaches some relatively positioned parent).

edemaine:
@jun-sheaf Could you make an actual example where the current behavior doesn't work? I just made [this Codepen](https://codepen.io/edemaine/pen/PvKOwO) to test, and didn't find your change to be necessary (on Chrome on Windows).

jun-sheaf:
@edemaine I've learned this is an issue with how Safari handles selection. Your codepen does exhibit the issue, but not on Chrome. Here are some screenshots.
Screen Shot 2019-05-21 at 3 19 00 PM
Screen Shot 2019-05-21 at 3 19 18 PM

ylemkimon:
Converted to an issue: #2132.

PR #1976, opened 2019-05-16
DOC: Fix path to built file

danielhollas:
Apart from this minor fix in the doc file, I have a question about building KaTeX with a custom path to fonts.

Turns out when you modify `@font-folder` in `submodules/katex-fonts/fonts.less`, webpack will no longer find them in `submodules/katex-fonts/fonts/` and the `dist/fonts/` folder will be missing in the build. Two-step process is needed in this case:

1. Build with default `fonts.less` file, and copy `dist/fonts/` folder where you need.
2. Modify `fonts.less` and build again to get correct css files.

I wonder whether this could be improved, or perhaps at least stated a bit more explicitly in this doc?


kevinbarabash:
@danielhollas what kind of change were you making to `@font-folder`?

danielhollas:
@kevinbarabash I was rebuilding KaTeX for [Discourse Math plugin](https://meta.discourse.org/t/discourse-math-plugin/65770), which requires:

`@font-folder: "/plugins/discourse-math/katex/fonts";`

edemaine:
Thanks for the fix. Perhaps worth moving the rest to another issue? I imagine this is standard NPM/webpack stuff that has a solution, but not being a webpack expert, I don't know what it is.

ylemkimon:
[clabot:check]

PR #1999, opened 2019-06-08
Improve fonts selection

ylemkimon:
* Remove number checks in mathdefault, they are textords.


kevinbarabash:
@ylemkimon why did you close this?

Issue #2031, opened 2019-07-13
Overlapping expressions


fabiospampinato:
> Before reporting a bug
- [x] Check [common issues](https://katex.org/docs/issues.html).
- [x] Check the bug is reproducible in [the demo](https://katex.org). If not, check KaTeX is up-to-date and installed correctly.
- [x] Search for [existing issues](https://github.com/KaTeX/KaTeX/issues).

**Describe the bug:**

Rendered expressions don't seem to be properly contained within the `span.katex` element, so they may overlap with one another in some scenarios.

**(La)TeX code:**

The expression I'm trying is the following, but it needs to be rendered multi-times for the issue to manifest:

```latex
F_e=\dfrac{可改进部分占用时间}{改进前总时间}
```

**Expected behavior:**

Expressions shouldn't overlap because of their height.

**Screenshots:**

<img width="579" alt="Screen Shot 2019-07-13 at 17 44 31" src="https://user-images.githubusercontent.com/1812093/61173686-e3f9ff00-a596-11e9-8f83-dacf7cc4a505.png">

**Environment (please complete the following information):**
 - KaTeX Version: 0.10.2
 - Device: [e.g. Desktop, iPhone 8] Desktop
 - OS: [e.g. Windows 10, iOS 11] macOS 10.14
 - Browser: [e.g. Chrome, Safari] Chrome
 - Version: [e.g. 72] 70-something I believe, this shouldn't be relevant to the issue.

**Additional context:**

_Display_, rather than inline, expressions are actually rendered correctly with this example, but I think only because they happen to have an arbitrary extra margin applied to them.

If you want to reproduce the same exact environment I'm actually rendering the following inside [Notable](https://github.com/notable/notable/releases):

```markdown
# KaTeX Bug

### Single

$F_e=\dfrac{可改进部分占用时间}{改进前总时间}$

### Multiple

$F_e=\dfrac{可改进部分占用时间}{改进前总时间}$
$F_e=\dfrac{可改进部分占用时间}{改进前总时间}$
$F_e=\dfrac{可改进部分占用时间}{改进前总时间}$
```

ronkok:
I can get a similar effect on the KaTeX demo page with fractions separated by line breaks. The code I used is:
```
F_e=\dfrac{可改进部分占用时间}{改进前总时间 \\
F_e=\dfrac{可改进部分占用时间}{改进前总时间
```

The resulting rendering has the same problem that @fabiospampinato describes.

[...]

PR #2032, opened 2019-07-13
Clarify documentation for CJK characters

ronkok:
In partial response to #2031.


edemaine:
This may end up getting subsumed by #2040.

ronkok:
I'm withdrawing this PR so that #2040 can be used instead.

PR #2040, opened 2019-07-14
Unicode characters in math render in text mode

### LINKED ISSUE #1817: Warning: No character metrics for '中' in style 'Main-Regular'
When I write Chinese characters in a math code block:

```latex
$$
f(x) = 测试中文
$$
```

They will be rendered properly but I still got a warning like this in the console:

```node
katex.js:5308 No character metrics for '中' in style 'Main-Regular'
```

This is a screenshot:

image

Follow the call stack, I can see the warning was logged at: https://github.com/KaTeX/KaTeX/blob/8346294bf30b438379bea17f5e626fb546f55d62/src/buildCommon.js#L82-L85

I have tried setting `strict` option to `false`. But it only disables `LaTeX-incompatible` warning, the problem above still exists. Could you please explain what does this mean, and how can I avoid it?

edemaine:
Improve #2031 by rendering all supported Unicode text characters (via `supportedCodepoints`) in text mode, roughly mimicking wrapping them in `\text`, thereby using metrics of letter M as usual.

I confirmed that the tests only work with the new (one-line) change.

Here's what something like #2031's example looks like now; the remaining overlap is because the characters are actually taller than an M:

```latex
F_e=\dfrac{可改进部分占用时间}{\text{MMM}改进前总时间}\\
F_e=\dfrac{\text{M}可改进部分占用时间}{改进前总时间}
```

image

The Ms are just for illustrative purposes; with this PR, they don't affect the vertical layout:

image

Here's the without-Ms version with currently published KaTeX: substantially more overlap.

image

This PR will also fix #1817.


ronkok:
I'm not sure how this works. I would have thought that it would be necessary to edit l[ine 288 in fontMetrics.js](https://github.com/KaTeX/KaTeX/blob/master/src/fontMetrics.js#L228).

ronkok:
@edemaine Does this make PR #2032 unnecessary?

ronkok:
Oh, I see now how you get font metrics. In `Parser.js`, you force all the symbols with high Unicode code points into text mode.

It seems like that would change the CSS class selected for wide Unicode characters in `wide-characters.js`. It's the same font glyph, though. Do we care if the CSS class changes?

edemaine:
I had in mind that `可` ought to be simulating `\text{可}` (at least that's what it takes in LaTeX), hence the forced text mode.  I could see this being somewhat undesired, if you specified different fallback fonts for math and text in custom CSS perhaps.  I think that's a bit of an edge case, though (differing text and math fonts for wide characters seems not terribly useful), and this PR is an improvement on the current situation.

But I could also see changing [the line you mentioned](https://github.com/KaTeX/KaTeX/blob/master/src/fontMetrics.js#L228); is it safe to just remove `&& mode === 'text'`?  Is this just going to remove all "no character metrics" warnings?  Perhaps we should check `supportedCodepoint`.

ronkok:
@edemaine My only quibble with this is that now the parser is sending a message saying "this argument is in text mode" when the user did not actually specify text mode. That sort of approach seems like it could lead to unforeseen problems if some future code depends on accurate information about the mode. An honesty in arguments policy seems like a good idea.

If we want to print every CJK character regardless of mode, the revision in fontMetrics.js can get us there without coercing the mode.

ronkok:
I withdraw the objection in my previous comment. The maintenance issues raised by PR #2065 are worse than the quibble in this PR.

edemaine:
Given the failure of #2065, I've [updated the documentation](https://github.com/KaTeX/KaTeX/pull/2040/commits/a9684d88501b98b8b00e6c3f7fe87b06479d9a82) here to point out the quirks of this PR, as well as document some currently hidden `fallback` features.  @ronkok, could you give it a read?

ronkok:
The documentation looks good to me.

docs/supported.md
```diff
@@ -180,9 +180,14 @@ Direct Input: $∂ ∇ ℑ Ⅎ ℵ ℶ ℷ ℸ ⅁ ℏ ð$
 
 **Unicode**
 
-The letters listed above will render in any KaTeX rendering mode.
+The letters listed above will render properly in any KaTeX rendering mode.
 
-If the KaTeX rendering mode is set to `strict: false` or `strict:"warn"` (default), then KaTeX will accept all Unicode letters. The letters not listed above will be rendered from system fonts, not KaTeX-supplied fonts, so their typography may clash. They may also cause small vertical alignment issues. KaTeX has detailed metrics for glyphs in Latin, Greek, and Cyrillic, but other glyphs are treated as if they are each as tall as the letter M.
+In addition, Brahmic, Georgian, Chinese, Japanese, and Korean glyphs are always accepted in text mode. However, these glyphs will be rendered from system fonts (not KaTeX-supplied fonts) so their typography may clash.
+You can provide rules for CSS classes `.latin-fallback`, `.cyrillic-fallback`, `.brahmic-fallback`, `.georgian-fallback`, `.cjk-fallback`, and `.hangul-fallback` to provide fallback fonts for these languages.
```
kevinbarabash: Cool!

docs/supported.md
```diff
@@ -180,9 +180,14 @@ Direct Input: $∂ ∇ ℑ Ⅎ ℵ ℶ ℷ ℸ ⅁ ℏ ð$
...
+Use of these glyphs may cause small vertical alignment issues: KaTeX has detailed metrics for listed symbols and most Latin, Greek, and Cyrillic letters, but other accepted glyphs are treated as if they are each as tall as the letter M.
```
kevinbarabash: This is a little ambiguous.  Maybe `... as the letter M from one of KaTeX's fonts`.  I forget if we use different heights for bold, regular, italic, sans-serif, etc.

docs/supported.md
```diff
@@ -180,9 +180,14 @@ Direct Input: $∂ ∇ ℑ Ⅎ ℵ ℶ ℷ ℸ ⅁ ℏ ð$
...
+
+If the KaTeX rendering mode is set to `strict: false` or `strict:"warn"` (default), then KaTeX will accept all Unicode letters in both text and math mode.
```
kevinbarabash: super nit: `strict: "warn"`

src/Parser.js
```diff
@@ -973,7 +973,7 @@ export default class Parser {
             }
             symbol = {
                 type: "textord",
-                mode: this.mode,
+                mode: "text",
```
kevinbarabash: Can you add a comment here explaining why we're doing this?

src/buildCommon.js
```diff
@@ -80,9 +80,8 @@ const makeSymbol = function(
             metrics.width, classes);
     } else {
         // TODO(emily): Figure out a good way to only print this in development
-        typeof console !== "undefined" && console.warn(
-            "No character metrics for '" + value + "' in style '" +
-                fontName + "'");
+        typeof console !== "undefined" && console.warn("No character metrics " +
+            `for '${value}' in style '${fontName}' and mode '${mode}'`);
```
kevinbarabash: I wish that our tests actually tested which error messages are being generated for the `.not.toParse()` and `.not.toBuild()` test cases.  Not something we should worry about in this PR.  I'll open an issue for this.

docs/supported.md
edemaine: I checked the code; it uses M from the current KaTeX font.  (I didn't check whether the heights differ, but seems plausible especially that the typewriter font differs.)  Updating to say "as tall as the letter M in the current KaTeX font".

edemaine:
@kevinbarabash Comments implemented; thanks! Do you want to take a look at the added comment before merging?

src/Parser.js
kevinbarabash: LGTM.  Thanks.

kevinbarabash:
@edemaine changes look great!  Thanks.

Issue #953, opened 2017-10-26
Pass the `Options` object everywhere.


marcianx:
There are a number of places in the code base where `Options` is optional because it's not being passed through by the parent. This is a tracking bug to pass `Options` everywhere. We can confirm its completion after fully porting to flow if we manage to turn all optional `Options` parameters into mandatory ones.

786-aquib:
Hello..
I would like to contribute...
And i want to fix it..

PR #2041, opened 2019-07-15
Fix font choice in operators like \log (e.g. \boldsymbol{\log})

### LINKED ISSUE #1971: \boldsymbol{\log}
**Describe the bug:**
`\boldsymbol{\log}` renders a roman-font bold. According to quicklatex, it should render bold.

edemaine:
Fix #1971 by correctly passing arguments into the font choice for the HTML handler for operators like `\log`.

This turned out to be a very easy fix. 😃 

I've also changed `mathsym` to require its `options` argument to avoid this type of bug in the future, partially addressing #953.


ronkok:
This looks good to me. I'm not sure why GitHub says that I requested a review from @kevinbarabash. His reviews are always welcome, but such a request was not my intent. Perhaps I fat-fingered a button.

edemaine:
@ronkok Thanks for taking a look!  If you want to write an approving review, then we can merge this.  Or if you'd rather wait for someone else to review, you can leave it as is.  (I forget whether you're comfortable doing approvals.)

ronkok:
> whether you're comfortable doing approvals.

Well, this sort of thing is not my profession, and when I look at the reviews done by @ylemkimon or @kevinbarabash, I find that about a third of the comments are items that I would have never thought to flag. So I should stick to PRs that are pretty straight forward. This one qualifies, I think

PR #2044, opened 2019-07-16
Do not force sizing groups to display inline-block

### LINKED ISSUE #1788: makeVList fails immediately in sizing group
Dear `makeVList` experts: I'm encountering a bug where negative shifts do not work when they come just inside a sizing group (as made by `\large`, `\huge`, etc.).  Here is an example:

```latex
A\raisebox{-1ex}{B}
\large A\raisebox{-1ex}{B}
\huge A\raisebox{-1ex}{B}
\large {A\raisebox{-1ex}{B}}
\huge {A\raisebox{-1ex}{B}}
```

Rendering on katex.org (ignore wrong font size for B):

image

Rendering with #1787 applied (correct font size for B):

image

The middle examples like `\large A\raisebox{-1ex}{B}` aren't descending the `B` at all.

Any ideas what's wrong?  Everything works fine for positive vertical ascents (positive `\raisebox`).

edemaine:
Fix #1788 by removing `display:inline-block` rule from `.sizing`, as [suggested by @ronkok](https://github.com/KaTeX/KaTeX/issues/1788#issuecomment-441264959).  I don't see why a sizing change cares about the display mode, and it fixes the interaction between `makeVList` and sizing commands, and didn't affect any screenshots beyond three invisible changes specific to Chrome.  It should also allow me to remove a `.sizing { width:min-content }` rule that I had to add in #1787.

Here's how the example from #1788 now renders (without #1787): all `B`s properly descend.

```latex
A\raisebox{-1ex}{B}
\large A\raisebox{-1ex}{B}
\huge A\raisebox{-1ex}{B}
\large {A\raisebox{-1ex}{B}}
\huge {A\raisebox{-1ex}{B}}
```
image

Here are screenshot diffs.  As there are no red or green pixels, the changes are invisible:

MathChoice-chrome-diff

ModScript-chrome-diff

Sizing-chrome-diff


kevinbarabash:
It's interesting that only the Chrome screenshots were affected.  Different browsers different rendering engines I guess. 🤷‍♂️

PR #2049, opened 2019-07-18
Add <katex-expression> to docs

navsgh:
A web component, built with Stencil, to render KaTeX expressions. Stencil uses a dynamic loader to load the custom elements polyfill only on browsers that need it.


ylemkimon:
[clabot:check]

ylemkimon:
[clabot:check]

Issue #2047, opened 2019-07-17
\not =, \neq, \ne, etc. not aligned correctly inside iframe [pen included]


nsheep755:
> Before reporting a bug
- [x] Check [common issues](https://katex.org/docs/issues.html).
- [x] Check the bug is reproducible in [the demo](https://katex.org). If not, check KaTeX is up-to-date and installed correctly.
- [x] Search for [existing issues](https://github.com/KaTeX/KaTeX/issues).

**Describe the bug:**
As the title says,` \not =`, `\neq`, `\ne`, etc. are not aligned correctly inside the iframe even though it loads fine outside an iframe; other things don't seem to have issues, including fractions, which seems to have been reported before

**(La)TeX code:**
The code of (La)TeX you tried to render:
```latex
a \ne b
a \neq b
a \not = b
```
(in math mode, both display and inline)

**Expected behavior:**
i'd expect it to look the same inside and outside an iframe because everything else seems to do that

**Pen:**
https://codepen.io/anon/pen/XLvvvL

**Environment (please complete the following information):**
 - KaTeX Version: 0.10.2
 - Device: MacBook Air 2018
 - OS: OSX Mojave (10.14.5), also tested on Windows 10
 - Browser: FireFox 68.0, Safari 12.1.1,  Chrome 75.0.3770.142
 - Version: see above?

**Additional context:**
n/a; thanks!

kevinbarabash:
I'm really surprised that this issue is cross browser.  I'm not sure what's causing the difference in behavior, but my guess would be there's some default style applied to the content of iframes that is affecting our styles.  Have you noticed any other differences in KaTeX's output when rendering inside an iframe?

nsheep755:
I haven't noticed anything else that is different. I've actually been working a lot with KaTeX inside an iframe and so far, this is the only thing I've noticed.

while playing around a bit, I noticed that the `.strut` element coming before what seems to be the `\not` symbol had the following style:

```css
vertical-align: -0.19444em;
```

and removing that seemed to align it correctly for one of the cases but I don't know if it covers all of them.

Thanks for taking a look at it!

edemaine:
The problem is that the iframe doesn't have a `<!DOCTYPE html>`, so it's running in quirks mode.  Adding `iframe.write('<!DOCTYPE html>')` fixes the problem.

I forget what our current state is with respect to quirks mode (level of support), but I think rendering issues like this are not unexpected.

kevinbarabash:
[...]

PR #2052, opened 2019-07-18
Suggest <!DOCTYPE html> in documentation

edemaine:
The documentation currently doesn't mention `<!DOCTYPE>` at all.  Inspired by #2047.


ylemkimon:
[The starter template on the README](https://github.com/KaTeX/KaTeX#starter-template) mentions it. Maybe should we include it in the main page or documentation?

edemaine:
@ylemkimon Good idea! I've added the starter template to `docs/browser.md`, in place of the DOCTYPE warning.  (There's still a DOCTYPE warning in Common Issues, which seems appropriate.)  Also fixed a few things in the wording of that document.

kevinbarabash:
In the case of #2047 the specific issue was that iframes don't inherit the DOCTYPE so it needs to be set there as well.  Perhaps we should specifically mention iframes.

edemaine:
@kevinbarabash Good idea.  I've [added a mention](https://github.com/KaTeX/KaTeX/pull/2052/commits/86c95095c04e225b07ba30413e386f3d3b755a64).

In [my latest commit](https://github.com/KaTeX/KaTeX/pull/2052/commits/06e95540720e55410bfe9401f0653eea3c139182), I also changed the default docs page to https://katex.org/docs/browser.html instead of https://katex.org/docs/node.html, because I feel like this is a more useful starting point for a novice.  But if anyone disagrees, I am happy to undo that change.

kevinbarabash:
@edemaine thanks.  @ylemkimon thanks for reviewing.

PR #2053, opened 2019-07-19
Re includegraphics

### LINKED ISSUE #1892: Load \includegraphics test image from local

### LINKED ISSUE #2027: Re-enable \includegraphics
`\includegraphics` was disabled in #1951 because we didn't have a way to address security concerns at the time and we need to get a bug fix out.  Now that we have the `strict` setting we can re-enable this functionality.  As part of this work we need to add `\includegraphics` to `TrustContextTypes` and check if the command is trusted in `includegraphics.js`.  See `href.js` for how to do this.

edemaine:
Fix #2027 by re-enabling `\includegraphics`, now that `trust` setting (#1794) has merged.  I reverted 5806b24 and checked old versions of #1794 for anything else to add.  I added some snapshot tests similar to those for `\url` and `\href`, and cleaned

While I was here, also fix #1892 by including the `\includegraphics` image locally for testing (and use it on the website).

Not directly related: I modified the `yarn start` interactive test website to set `trust` to `true` by default, like it is on the katex.org demo, because the assumption is that someone entering text manually is doing so in a trustworthy fashion.  I added a `trust=0` option for testing that things break when `trust` is `false`.


kevinbarabash:
Thanks for re-enabling this functionality and for tweaking `yarn start` so that we can test everything on the test page locally.

PR #2054, opened 2019-07-19
Fix parse timing by separating consume() into fetch() and consume()

### LINKED ISSUE #1989: First character of array doesn't get its own group
The following should render 111, but it currently renders as 22:
```latex
\def\x{1}\begin{matrix}\def\x{2}&\x\end{matrix}\x
```

As described in `test/katex-spec.js` (see TODO), before the environment gets called, `{matrix}` gets consumed which means that the `\def` gets executed, before we can create a group. :-(

This is a general issue with the way `nextToken` works, so it's going to be difficult to fix.

edemaine:
Fix #1989, and generally cleanup parse timing (many fewer hoops to jump through) by defining two methods on parser:

1. `consume()` marks the current token (`nextToken`) as "done", but doesn't yet fetch the next token (setting `nextToken` to `null`).
2. `fetch()` fetches the next token if we don't already have one (e.g., if last token was `consume()`d).

Before this change, `consume()` did both actions.  By separating them, and allowing the parser to live in a state for a while where `nextToken` is `null`, it is far easier to change settings (in particular, math/text mode and catcodes) before reading the next token, in a way that depends on what we're parsing.  For example, if an argument should be treated in text mode, we can just set the mode in the argument parser, instead of when the previous token was consumed.  Similarly, if an argument should be treated as a URL, we can just set the catcode of `%` in the URL argument parser, and reset it after.  We no longer have to take care to reset things before calling `consume()`.

This change mostly involves changing `this.nextToken` to `this.fetch()`. In a perfect world, we could use slightly fewer calls to `this.fetch()`, but Flow doesn't realize that `this.nextToken` will be non-null after a
call to `this.fetch()`, so we need to use a few more calls to `this.nextToken()` or a few more local `nextToken` variables.

I'm kind of impressed how such a major change could be relatively easy to do.  It's been extremely helpful to have so many tests and type checking in our codebase.  For reference, I previously tried to change all the methods to load the token when they need them, and not bother marking used tokens as consumed, so that `nextToken` is never `null`.  This proved very difficult (maybe impossible) because we sometimes get nested deep within `parseExpression` and a single token needs to trigger all of them to terminate (e.g., `\fbox{$\displaystyle x$}`, which is what `\boxed{x}` expands to, the last `$` token ends the `\displaystyle` implicit group and the math mode). So I opted for this approach, which is a much smaller (and more obviously correct) change to the code.


kevinbarabash:
This is great!  I'll try to review this sometime this week.

ylemkimon:
Thank you for the PR. This would allow manipulation of `nextToken` on the Parser level, which is needed fixing issues like #924!

src/Parser.js
```diff
@@ -551,7 +550,10 @@ export default class Parser {
                 return styledGroup;
             }
             case "raw": {
-                if (optional && this.nextToken.text === "{") {
+                if (consumeSpaces) {
+                    this.consumeSpaces();
+                }
+                if (optional && this.fetch() === "{") {
```
ylemkimon:
```suggestion
                if (optional && this.fetch().text === "{") {
```

edemaine:
@ylemkimon Thanks for the review! I do think this will be a helpful step for parsing. I hadn't thought about #924 in a while, but it would be great if it helps with that!

PR #2062, opened 2019-07-31
Port katex-a11y.js to KaTeX

kevinbarabash:
katex-a11y.js was an add-on that Khan Academy created to generate a string that could be read by a screen reader to describe the math that KaTeX renders.  When attempting to update KaTeX we noticed that this custom add-on broke because of changes to the internal AST.  The goal of this PR is to bring the add-on up-to-date so it works with the current version of KaTeX.  I'm also adding tests and flow types to help keep it up to date or at the very least avoid breaking the functionality.  I'm not going to be able to everything that has a TODO, but at the very least I'd like to have tests that covers all of the node types without TODOs.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -75,7 +75,8 @@
     "env": {
         "es6": true,
         "node": true,
-        "browser": true
+        "browser": true,
+        "jest": true
     },
     "settings": {
         "react": {
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,0 +1,7 @@
+{
+    "trailingComma": "all",
+    "tabWidth": 4,
+    "semi": true,
+    "singleQuote": false,
+    "bracketSpacing": false
+}
--- a/contrib/to-a11y-string/test/to-a11y-string-spec.js
+++ b/contrib/to-a11y-string/test/to-a11y-string-spec.js
@@ -376,7 +376,7 @@ describe("renderString", () => {
         });
 
         test("1 \\ne 2", () => {
-            const result = renderString("1 \\neq 2");
+            const result = renderString("1 \\ne 2");
             expect(result).toMatchInlineSnapshot(`"1, does not equal, 2"`);
         });
 
@@ -402,7 +402,7 @@ describe("renderString", () => {
         });
 
         test("1 \\le 2", () => {
-            const result = renderString("1 \\leq 3");
+            const result = renderString("1 \\le 3");
             expect(result).toMatchInlineSnapshot(
                 `"1, is less than or equal to, 3"`,
             );
--- a/contrib/to-a11y-string/test/to-a11y-string-spec.js
+++ b/contrib/to-a11y-string/test/to-a11y-string-spec.js
@@ -1,0 +1,368 @@
+/* eslint-disable max-len */
+// @flow
+import {renderString} from "../to-a11y-string";
+
+describe("renderString", () => {
+    describe("basic expressions", () => {
+        test("simple addition", () => {
+            const result = renderString("1 + 2");
+            expect(result).toMatchInlineSnapshot(`"1, plus, 2"`);
+        });
+    });
+
+    describe("accent", () => {
+        test("\\vec", () => {
+            const result = renderString("\\vec{a}");
+            expect(result).toMatchInlineSnapshot(`"a, with, vector, on top"`);
+        });
+
+        test("\\hat{a}", () => {
+            const result = renderString("\\hat{a}");
+            expect(result).toMatchInlineSnapshot(`"a, with, hat, on top"`);
+        });
+    });
+
+    describe("accentUnder", () => {
+        test("\\underleftarrow", () => {
+            const result = renderString("\\underleftarrow{1+2}");
+            expect(result).toMatchInlineSnapshot(
+                `"1, plus, 2, with, left arrow, underneath"`,
+            );
+        });
+
+        test("\\underlinesegment", () => {
+            const result = renderString("\\underlinesegment{1+2}");
+            expect(result).toMatchInlineSnapshot(
+                `"1, plus, 2, with, line segment, underneath"`,
+            );
+        });
+    });
+
+    describe("color", () => {
+        test("\\color{red}", () => {
+            const result = renderString("\\color{red}1+2");
+            expect(result).toMatchInlineSnapshot(
+                `"start color red, 1, end color red, plus, 2"`,
+            );
+        });
+
+        test("\\textcolor{red}", () => {
+            const result = renderString("\\textcolor{red}1+2");
+            expect(result).toMatchInlineSnapshot(
+                `"start color red, 1, end color red, plus, 2"`,
+            );
+        });
+    });
+
+    describe("delimiters", () => {
+        test("simple parens", () => {
+            const result = renderString("(1 + 3)");
+            expect(result).toMatchInlineSnapshot(
+                `"left parenthesis, 1, plus, 3, right parenthesis"`,
+            );
+        });
+
+        test("simple brackets", () => {
+            const result = renderString("[1 + 3]");
+            expect(result).toMatchInlineSnapshot(
+                `"open bracket, 1, plus, 3, close bracket"`,
+            );
+        });
+
+        test("nested parens", () => {
+            const result = renderString("(a + (b + c))");
+            expect(result).toMatchInlineSnapshot(
+                `"left parenthesis, a, plus, left parenthesis, b, plus, c, right parenthesis, right parenthesis"`,
+            );
+        });
+
+        test("stretchy parens around fractions", () => {
+            const result = renderString("\\left(\\frac{1}{x}\\right)");
+            expect(result).toMatchInlineSnapshot(
+                `"left parenthesis, start fraction, 1, divided by, x, end fraction, right parenthesis"`,
+            );
+        });
+    });
+
+    describe("enclose", () => {
+        test("\\cancel", () => {
+            const result = renderString("\\cancel{a}");
+            expect(result).toMatchInlineSnapshot(
+                `"start cancel, a, end cancel"`,
+            );
+        });
+
+        test("\\fbox", () => {
+            const result = renderString("\\fbox{a}");
+            expect(result).toMatchInlineSnapshot(`"start box, a, end box"`);
+        });
+
+        test("\\sout", () => {
+            const result = renderString("\\sout{a}");
+            expect(result).toMatchInlineSnapshot(`"start strikeout, a, end strikeout"`);
+        });
+    });
+
+    describe("exponents", () => {
+        test("simple exponent", () => {
+            const result = renderString("e^x");
+            expect(result).toMatchInlineSnapshot(
+                `"e, start superscript, x, end superscript"`,
+            );
+        });
+
+        test("^{\\circ} => degrees", () => {
+            const result = renderString("90^{\\circ}");
+            expect(result).toMatchInlineSnapshot(`"90, degrees"`);
+        });
+
+        test("^{\\degree} => degrees", () => {
+            const result = renderString("90^{\\degree}");
+            expect(result).toMatchInlineSnapshot(`"90, degrees"`);
+        });
+
+        test("^{\\prime} => prime", () => {
+            const result = renderString("f^{\\prime}");
+            expect(result).toMatchInlineSnapshot(`"f, prime"`);
+        });
+
+        test("^2 => squared", () => {
+            const result = renderString("x^2");
+            expect(result).toMatchInlineSnapshot(`"x, squared"`);
+        });
+
+        test("^3 => cubed", () => {
+            const result = renderString("x^3");
+            expect(result).toMatchInlineSnapshot(`"x, cubed"`);
+        });
+
+        test("log_2", () => {
+            const result = renderString("\\log_2{x+1}");
+            expect(result).toMatchInlineSnapshot(
+                `"log, start subscript, 2, end subscript, x, plus, 1"`,
+            );
+        });
+    });
+
+    describe("genfrac", () => {
+        test("simple fractions", () => {
+            const result = renderString("\\frac{2}{3}");
+            expect(result).toMatchInlineSnapshot(
+                `"start fraction, 2, divided by, 3, end fraction"`,
+            );
+        });
+
+        test("nested fractions", () => {
+            const result = renderString("\\frac{1}{1+\\frac{1}{x}}");
+            // TODO: this result is ambiguous, we need to fix this
+            expect(result).toMatchInlineSnapshot(
+                `"start fraction, 1, divided by, 1, plus, start fraction, 1, divided by, x, end fraction, end fraction"`,
+            );
+        });
+
+        test("binomials", () => {
+            const result = renderString("\\binom{n}{k}");
+            expect(result).toMatchInlineSnapshot(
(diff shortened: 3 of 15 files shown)


kevinbarabash:
I better fix those tests so that we can get some coverage reports.

.eslintrc
```diff
@@ -75,7 +75,8 @@
     "env": {
         "es6": true,
         "node": true,
-        "browser": true
+        "browser": true,
+        "jest": true
```
kevinbarabash: This will allow as to drop the:
```
/* global expect: false */
/* global it: false */
/* global describe: false */
```
comments that are at the top of all of our spec files.

.prettierrc
```diff
@@ -0,0 +1,7 @@
+{
+    "trailingComma": "all",
+    "tabWidth": 4,
+    "semi": true,
+    "singleQuote": false,
+    "bracketSpacing": false
+}
```
kevinbarabash: This is only used by jest when formatting tests after inserting inline snapshots.  It is not used for formatting any of the code in `src`.

src/unicodeMake.js
```diff
@@ -11,7 +11,7 @@ const target = path.join(__dirname, 'unicodeSymbols.js');
 const targetMtime = fs.statSync(target).mtime;
 if (fs.statSync(__filename).mtime <= targetMtime && fs.statSync(
         path.join(__dirname, 'unicodeAccents.js')).mtime <= targetMtime) {
-    return;
```
kevinbarabash: Wallaby compiles this file and complained about this `return`.  I'm kind of surprised eslint didn't pick this up sooner.  The issue is that it's a `return` that appears outside of a function.

wallaby.js
```diff
@@ -0,0 +1,89 @@
+const babelConfig = require("./babel.config.js");
+
+module.exports = function(wallaby) {
+    const tests = [
+        // "test/*-spec.js",
```
kevinbarabash: I disabled running these tests b/c I hadn't created `.prettierrc` and it was reformatting the file.  It should be safe to re-enable this now.

src/unicodeMake.js
edemaine: Both versions work in Node because the top-level script is wrapped in a function. But it is an invalid "ECMAScript program". [Weird](https://stackoverflow.com/questions/28955047/why-does-a-module-level-return-statement-work-in-node-js). Anyway, fine to change.

contrib/to-a11y-string/test/to-a11y-string-spec.js
```diff
@@ -0,0 +1,368 @@
...
+        });
+    });
+
+    describe("color", () => {
+        test("\\color{red}", () => {
+            const result = renderString("\\color{red}1+2");
+            expect(result).toMatchInlineSnapshot(
+                `"start color red, 1, end color red, plus, 2"`,
+            );
+        });
```
edemaine: That looks like old `\color` behavior. It won't work without option `colorIsTextColor` set to `true`.

contrib/to-a11y-string/test/to-a11y-string-spec.js
```diff
@@ -0,0 +1,368 @@
...
+        test("\\cancel", () => {
+            const result = renderString("\\cancel{a}");
+            expect(result).toMatchInlineSnapshot(
+                `"start cancel, a, end cancel"`,
+            );
+        });
+
+        test("\\fbox", () => {
+            const result = renderString("\\fbox{a}");
+            expect(result).toMatchInlineSnapshot(`"start box, a, end box"`);
```
edemaine: Maybe worth testing newer`\boxed` etc.

contrib/to-a11y-string/test/to-a11y-string-spec.js
```diff
@@ -0,0 +1,368 @@
...
+
+        test("^3 => cubed", () => {
+            const result = renderString("x^3");
+            expect(result).toMatchInlineSnapshot(`"x, cubed"`);
+        });
+
+        test("log_2", () => {
+            const result = renderString("\\log_2{x+1}");
+            expect(result).toMatchInlineSnapshot(
+                `"log, start subscript, 2, end subscript, x, plus, 1"`,
```
edemaine: Two ideas for what I'd consider "more natural":
1. Specific to log, I could see saying `start base ... end base` instead of `start subscript ... end subscript`.
2. In general, when a subscript, `\cancel`, etc. is given a "short" argument (maybe one character?), you could imagine using the noun instead of start noun ... end noun.  For example, `x_2` could be `x, subscript, 2` instead of `x, begin subscript, 2, end subscript`.

Just ideas -- feel free to ignore, or just add as TODOs.

contrib/to-a11y-string/test/to-a11y-string-spec.js
```diff
@@ -0,0 +1,368 @@
...
+            // TODO: this result is ambiguous, we need to fix this
+            expect(result).toMatchInlineSnapshot(
+                `"start fraction, 1, divided by, 1, plus, start fraction, 1, divided by, x, end fraction, end fraction"`,
+            );
+        });
+
+        test("binomials", () => {
+            const result = renderString("\\binom{n}{k}");
+            expect(result).toMatchInlineSnapshot(
+                `"start binomial, left parenthesis, n, over, k, right parenthesis, end binomial"`,
```
edemaine: I wouldn't have expected parenthesis in addition to binomial.

contrib/to-a11y-string/test/to-a11y-string-spec.js
```diff
@@ -0,0 +1,368 @@
...
+            expect(result).toMatchInlineSnapshot(
+                `"1, is less than or equal to, 3"`,
+            );
+        });
+    });
+
+    describe("rule", () => {
+        test("\\rule", () => {
+            const result = renderString("\\rule{1em}{1em}");
+            expect(result).toMatchInlineSnapshot(`"rule"`);
```
edemaine: "rectangle" might be more intuitive... but "rule" is also fine (I assume it means something to typographers 😄).

flow-typed/npm/jest_v24.x.x.js
```diff
@@ -0,0 +1,1201 @@
...
+  arrayContaining(value: Array<mixed>): Array<mixed>,
+  clock(): JestClockType,
+  createSpy(name: string): JestSpyType,
+  createSpyObj(
+    baseName: string,
+    methodNames: Array<string>
+  ): { [methodName: string]: JestSpyType },
+  objectContaining(value: Object): Object,
+  stringMatching(value: string): string,
+};
```
edemaine: Does this need to be checked in? Presumably it can be included from an NPM package somewhere somehow?

contrib/to-a11y-string/test/to-a11y-string-spec.js
kevinbarabash: I should be able to do the `start base ... end base` as part of this PR.  I'll add a TODO for simplifying short subscripts.  I think having a single ticket that collects all of the TODOs is probably the way to go.

contrib/to-a11y-string/test/to-a11y-string-spec.js
kevinbarabash: Yeah, this isn't very good.  It looses all of the meaning.  I think having some way to leave things unexpanded would help with this but it would complicate ParseNode.  Also, not expanding certain things could result in an unparsable expression. :\

contrib/to-a11y-string/test/to-a11y-string-spec.js
kevinbarabash: Good call.

flow-typed/npm/jest_v24.x.x.js
kevinbarabash: Unfortunately flow-typed doesn't work like this... yet.  There are some tickets proposing this change, e.g. https://github.com/flow-typed/flow-typed/issues/990.  It's complicated by the fact that flow types are versioned both by flow version and by library version.

contrib/to-a11y-string/test/to-a11y-string-spec.js
kevinbarabash: Good call.

contrib/to-a11y-string/test/to-a11y-string-spec.js
kevinbarabash: That's for pointing that out.  I'll fix this.

kevinbarabash:
@edemaine I've asked @mpolyak, another developer from Khan Academy, to help review this code as well.  He's working on upgrading KaTeX on khanacademy.org.

contrib/to-a11y-string/test/to-a11y-string-spec.js
edemaine: I think a good solution for this might be to add macros before calling KaTeX, to customize the rendering of some commands, like we already do to make cleaner MathML outputs (check out the existing macros in `macros.js` that use `\html@mathml`).  I could imagine macro-defining `\binom` to something like `{\@speak{begin}\@speak{binomial}#1\@speak{choose}#2\@speak{end}\@speak{binomial}}` for a11y purposes.  Ideally then `\@speak` could be `defineFunction`d to create a parse node that then just gets output in a11y.

contrib/to-a11y-string/test/to-a11y-string-spec.js
kevinbarabash: I think `supsub`s for integrals and summations could be improved as well.

contrib/to-a11y-string/test/to-a11y-string-spec.js
```diff
@@ -0,0 +1,368 @@
...
+    });
+
+    describe("relations", () => {
+        test("1 \\neq 2", () => {
+            const result = renderString("1 \\neq 2");
+            expect(result).toMatchInlineSnapshot(`"1, does not equal, 2"`);
+        });
+
+        test("1 \\ne 2", () => {
+            const result = renderString("1 \\neq 2");
```
mpolyak: Should be `"1 \\ne 2"`?

contrib/to-a11y-string/test/to-a11y-string-spec.js
```diff
@@ -0,0 +1,368 @@
...
+
+        test("1 \\leq 2", () => {
+            const result = renderString("1 \\leq 3");
+            expect(result).toMatchInlineSnapshot(
+                `"1, is less than or equal to, 3"`,
+            );
+        });
+
+        test("1 \\le 2", () => {
+            const result = renderString("1 \\leq 3");
```
mpolyak: `"1 \\le 2"`?

contrib/to-a11y-string/to-a11y-string.js
```diff
@@ -0,0 +1,735 @@
+// @flow
+/**
+ * KaTeX A11y
+ * A library for converting KaTeX math into readable strings.
+ */
+
+// NOTE(jeresig): We need to keep this file as pure ES5 to avoid import
```
mpolyak: Are these still true?

contrib/to-a11y-string/to-a11y-string.js
```diff
@@ -0,0 +1,735 @@
...
+    } else if (type === "bin") {
+        ret = binMap[str] || str;
+    } else if (type === "rel") {
+        ret = relMap[str] || str;
+    } else {
+        ret = stringMap[str] || str;
+    }
+
+
+    // If nothing was found and it's not a plain string or number
```
mpolyak: We longer need to test this?

contrib/to-a11y-string/to-a11y-string.js
```diff
@@ -0,0 +1,735 @@
...
+                buildA11yStrings(tree.base, a11yStrings, atomType);
+                a11yStrings.push("with");
+                buildString(accentUnderMap[tree.label], "normal", a11yStrings);
+                a11yStrings.push("underneath");
+            });
+            break;
+        }
+
+        case "accent-token": {
+            throw new Error("TODO: accent-token");
```
mpolyak: Is it possible to prefix exceptions with something like `KaTeX a11y` to ease tracing back the error?

contrib/to-a11y-string/to-a11y-string.js
```diff
@@ -0,0 +1,735 @@
...
+                    regionStrings.push("over");
+                    buildA11yStrings(tree.denom, regionStrings, atomType);
+                    rightDelim && buildString(rightDelim, "close", regionStrings);
+                    regionStrings.push("end binomial");
+                }
+            });
+            break;
+        }
+
+        // katex: function(tree: ParseNode<"katex">, a11yStrings) {
```
mpolyak: Comment block can be deleted?

contrib/to-a11y-string/test/to-a11y-string-spec.js
kevinbarabash: `\ne` is an alias for `\neq`.

contrib/to-a11y-string/test/to-a11y-string-spec.js
kevinbarabash: `\le` is an alias for `\leq`.

contrib/to-a11y-string/to-a11y-string.js
kevinbarabash: I'll remove stale comments.

contrib/to-a11y-string/to-a11y-string.js
kevinbarabash: Good idea, will do!

contrib/to-a11y-string/to-a11y-string.js
kevinbarabash: It can.  I'll delete.  I don't we use `\katex` in any of our content anyways.

contrib/to-a11y-string/test/to-a11y-string-spec.js
edemaine: You have a test labeled `\neq` and another test labeled `\ne`, but both tests actually test `\neq`.  You could probably just remove the `\ne` test, or the tests should be made distinct.

kevinbarabash: I see... I'll fix that.

contrib/to-a11y-string/test/to-a11y-string-spec.js
kevinbarabash: I worry about injecting human readable strings into our macros.  It doesn't well with i18n.

kevinbarabash:
This is ready for review again.

contrib/to-a11y-string/test/to-a11y-string-spec.js
kevinbarabash: I think I get what you're saying now.  `renderA11yString` would be responsible for defining the macros.  That's a really cool idea.

contrib/to-a11y-string/to-a11y-string.js
kevinbarabash: Let me double check this.

kevinbarabash: This check is too restrictive.  KaTeX accepts pretty much all Unicode now.  I'll remove this commented out code.

contrib/to-a11y-string/test/to-a11y-string-spec.js
kevinbarabash: I've added a test for `\boxed`.

kevinbarabash:
I think all of the concerns have been addressed.

PR #2065, opened 2019-08-02
Wide Unicode character support in math mode

### LINKED ISSUE #1817: Warning: No character metrics for '中' in style 'Main-Regular'

edemaine:
Improve #2031 by rendering all supported Unicode text characters (via supportedCodepoints) in math mode in addition to existing text mode support (using metrics of letter M as usual).  Also fix #1817.

This is a replacement/alternative to #2040 and #2032. I just implemented @ronkok's [suggested one-line change](https://github.com/KaTeX/KaTeX/pull/2040#issuecomment-511919654), and copied over the tests from #2040.

Same screenshots:

```latex
F_e=\dfrac{可改进部分占用时间}{\text{MMM}改进前总时间}\\
F_e=\dfrac{\text{M}可改进部分占用时间}{改进前总时间}
```

image

Without the Ms, we get the same vertical layout:

image

Without this PR:

image

### DIFF
--- a/test/unicode-spec.js
+++ b/test/unicode-spec.js
@@ -9,7 +9,7 @@ import {strictSettings, nonstrictSettings} from "./helpers";
 describe("unicode", function() {
     it("should parse Latin-1 inside \\text{}", function() {
         expect`\text{ÀÁÂÃÄÅÈÉÊËÌÍÎÏÑÒÓÔÕÖÙÚÛÜÝàáâãäåèéêëìíîïñòóôõöùúûüýÿÆÇÐØÞßæçðøþ}`
-            .toParse();
+            .toBuild();
     });
 
     it("should not parse Latin-1 outside \\text{} with strict", function() {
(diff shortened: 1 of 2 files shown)


edemaine:
Interesting: `boldsymbol` (which determines the correct font for `\boldsymbol`) calls `lookupSymbol` to see if the symbol is in KaTeX-Math, which calls `getCharacterMetrics`, which is affected by this change, because `supportedCodepoint` seems to include at least some math characters, namely, U+0131 for `\imath`.

I see a few options:
1. Modify `getCharacterMetrics` to include a special field like `notInFont` or `missing` or `notReally` (joking) to indicate that the character wasn't in the font, and metrics are just simulated.
2. Try to modify `supportedCodepoint` to miss any characters in our fonts (seems risky).
3. Use #2040 instead.

Let me know your opinions!

ronkok:
You're right. `supportedCodepoint` has been written to include (among many other things) the entire Latin A and Latin B code block, U+0100 to U+024f. That overrides the KaTeX glyphs: ı Œ œ ȷ

I agree that we don't want to have to ensure that the KaTeX fonts and `supportedCodepoint` are disjoint sets every time the KaTeX fonts are revised. That would be a bigger maintenance headache than the one I brought up in a comment in PR #2040. 

And so I withdraw my quibbles about #2040. It seems like the better way to go.

kevinbarabash:
It not very obvious that the change you made in fontMetrics.js would have the undesired impact that it did.  I'm glad we had at least one screenshot that picked up this issue.  @edemaine thanks for working out the source of the problem.  I think in general the code we use t determine which character and font to use for a particular symbol is overly complicated and doesn't align very well with how TeX does things.  I'm not saying we have to be just like TeX, but being closer to TeX's model might help.

kevinbarabash:
@edemaine the screenshots need updating because of this change right?

test/unicode-spec.js
```diff
@@ -9,7 +9,7 @@ import {strictSettings, nonstrictSettings} from "./helpers";
 describe("unicode", function() {
```
kevinbarabash: The titles for these tests should change from `it("should parse...` to `it("should build...`.

ronkok:
Didn't we agree to use PR #2040 instead of this PR?

edemaine:
Replaced by PR #2040.

PR #2066, opened 2019-08-02
Use current font for accents

### LINKED ISSUE #353: fonts for accents don't match base font
See https://github.com/Khan/KaTeX/pull/349 for details.

### LINKED ISSUE #1099: accents on characters inside \textit{} are incorrectly positioned

edemaine:
Fix #353 using `buildCommon.makeSymbol` to automatically determine font for the accent symbol.

I'd forgotten about this issue but @kevinbarabash's recent post in it caused me to look at it again.  I believe the fix is quite simple with the current code (much revised since the issue was first raised).

Example:

```latex
\acute a\text{\'a}\textit{\'a}\texttt{\'a}\mathbf{\acute a}\boldsymbol{\acute a}
```

image

As a bonus, this also fixes #1099: turns out the issue there was not using the italic-font accents within `\textit`.

```latex
\textit{\r{A}\H{A}\.A\'A\"A}
```

image

The screenshot diffs highlight the improvements (green is new):

MathIt: MathIt-chrome-diff
MathBf: MathBf-chrome-diff
MathSf: MathSf-chrome-diff
MathTt: MathTt-chrome-diff


edemaine:
@ylemkimon Thanks for the review!

ylemkimon:
@edemaine Thank you for tackling the long-standing problem!

kevinbarabash:
This is awesome! 🚀

PR #2069, opened 2019-08-06
mathtex-script: Use html 'defer' attribute

Evidlo:
Instead of putting the extension script tag at the bottom, we can use the `defer` attribute to execute the script after all other content is loaded.  It has [very good](https://caniuse.com/#feat=script-defer) browser support.


kevinbarabash:
@Evidlo I've resolved the merge conflicts in your PR.  Can you have a look at them to double check I did the right thing?

Evidlo:
@kevinbarabash The checksums needed to be updated.  It works for me now.

kevinbarabash:
@Evidlo thanks for catching that.  Why are the checksums for katex.min.js and katex.min.css different from what's on master?

Evidlo:
@kevinbarabash Not sure, I'm using the links generated from [here](https://www.jsdelivr.com/package/npm/katex?path=dist), and it generates sha256 hashes instead of sha384.

kevinbarabash:
I see.  I think we should probably stick with the `sha384` hashes since that's what we use everywhere else.

kevinbarabash:
@Evidlo sorry for not merging this before the recent release.  Would you be able to resolve the merge conflict?

Evidlo:
@kevinbarabash Done

kevinbarabash:
@Evidlo thanks.  One more request.  This PR actually reverts the version in the URLs from `0.11.1` to `0.11.0`.  That needs to be fixed before we can merge this.

Evidlo:
Should be good now.

ylemkimon:
[clabot:check]

Evidlo:
[clabot:check]

PR #2075, opened 2019-08-16
Fix \gray's macro definition

### LINKED ISSUE #2074: \gray not supported
As part of an effort to upgrade KaTeX from version `0.9` to the current version `0.11` on www.khanacademy.org we ran an audit of published math content on the site, to compare parse errors using `KaTeX 0.11` vs the (currently used) `KaTeX 0.9` library version.

We found that that expressions using `\gray` fail to parse with `KaTeX 0.11`:
```
ParseError: KaTeX parse error: Expected '}', got '#' at position 18: …extcolor{gray}{#̲#1}
```

Here are some examples:
```
\gray{1}
```
```
\gray6
```
```
\gray 1
```
```
\gray9 \times \gray3
```
```
\gray{\text{apple}}
```

These all fail to parse with the above error.

edemaine:
Fix #2074 by fixing a typo in the macro definition of `\gray`. This seems to be the only color incorrectly defined.

### DIFF
--- a/src/macros.js
+++ b/src/macros.js
@@ -952,7 +952,7 @@ defineMacro("\\orange", "\\textcolor{##ffa500}{#1}");
 defineMacro("\\pink", "\\textcolor{##ff00af}{#1}");
 defineMacro("\\red", "\\textcolor{##df0030}{#1}");
 defineMacro("\\green", "\\textcolor{##28ae7b}{#1}");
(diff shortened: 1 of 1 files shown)


danielhollas:
Whoops, this one's on me, so sorry. :frowning_face:  Thanks @edemaine for fixing!

PR #2080, opened 2019-08-20
fix(builder): combine characters together in all expressions

### LINKED ISSUE #2079: \href{} Produces Huge number of elements (1 per character)
Katex is producing huge number of <span> elements when wrapping text within the `\href{}` directive. I can't see any immediate reason to suddenly produce 1 span per character, rather than just simply wrapping the range of identically styled characters in the <a> tag.
[jsfiddle](https://jsfiddle.net/0a7e6fx1/3/)

I am generating a lot of text, and don't want the overhead of adding hundreds of spans for each link.

**(La)TeX code:**
```latex
\text{\href{https://google.com}{abcdefghijk}}
```

**Expected behavior:**
image

 - KaTeX Version: 0.10.2
 - Device: Desktop
 - OS: macOS 10.13.6
 - Browser: Chrome
 - Version: 76

edemaine:
An attempt to fix #2079.  Call `tryCombineChars` in `buildExpression`, so all expressions get adjacent characters combined together.

Unlike past KaTeX builds, this happens even in math mode, in particular because `buildExpression` doesn't know (and maybe can't know) the current mode.  This is a significant deviation, so deserves discussion.  Is it bad that we render math-mode `hello` into 4 `mord` spans (`h`, `e`, `ll`, `o`)?  (I believe the multiple `mord`s are caused by the italic correction rule in `tryCombineChars`.)

Alternatively, perhaps we could tell `buildExpression` whether to combine according to the mode... or just review every call to `buildExpression` and see whether to add a call `tryCombineChars`?


kevinbarabash:
One concern that I have is how combining characters affects the a11y plugin.  Certain things should be combined for that plugin like `sin` and numbers and other things like the `xyz` in `2xyz + 5 = 0` should be separate in terms of what the plugin sees.  That being said, it's okay to combine `xyz` afterwards.

ronkok:
> the `xyz` in `2xyz + 5 = 0` should be separate in terms of what the [a11y] plugin sees. That being said, it's okay to combine `xyz` afterwards.

Doesn't the a11y plugin work on the parse tree? This PR works with elements in  the HTML builder. I don't see a conflict. Can this PR proceed to approval and merge?

kevinbarabash:
> Doesn't the a11y plugin work on the parse tree? This PR works with elements in the HTML builder. I don't see a conflict. Can this PR proceed to approval and merge?

@ronkok thanks for pointing this out.

kevinbarabash (approved):
I think it would be good to have a test for this.  A snapshot test should do.

kevinbarabash:
I checked out this branch to update the screenshots.  I was playing around with the test page and notice it doesn't produce quite the desired behavior described in #2079.  While there are still groups within the `href` text, this is a good start and I can be merged as is (once I add some snapshot tests).
Screen Shot 2020-12-28 at 1 20 27 PM

Looking at the italic corrections got me thinking: should we always be adding the correction like we currently do?  TBH, in some situations it makes it feel like some of the letters are too far apart (e.g. the space between `r` and `l`).  I think having the correct makes sense when a letter in italics are followed by a number or a super script.  If we dropped the italics corrections in `Hello, world` it would also allow us to group more of the letter together.  It's just a thought and is totally outside the scope of this PR.

kevinbarabash:
After updating the screenshots I found a couple of issues:
Screen Shot 2020-12-28 at 1 40 08 PM
Screen Shot 2020-12-28 at 1 39 53 PM

src/buildCommon.js
```diff
@@ -243,6 +243,15 @@ const canCombine = (prev: SymbolNode, next: SymbolNode) => {
...
     }
 
+    // If prev and next both are just "mbin"s or "mord"s we don't combine them
+    // so that the proper spacing can be preserved.
+    if (prev.classes.length === 1) {
+        const cls = prev.classes[0];
+        if (cls === "mbin" || cls === "mord") {
+            return false;
+        }
+    }
```
kevinbarabash, line 253: @edemaine are there any other classes that we should do this for?

kevinbarabash: I'm going to merge this as is and if we find other classes that need to be handled this way we can add them later.

edemaine:
Thanks for the review!  Looking back at this now, it's semantically a bit weird to group letters together at all in math mode, but I guess that's important only for MathML where semantics are important, whereas HTML is all about getting the rendering right.  Hopefully this PR will make for much smaller DOM trees in complex formulas!

Issue #1596, opened 2018-08-12
Add \html command to insert HTML


pzinn:
I would like to be able to insert some html into katex output (ex of use: inserting a svg image in the middle of latex code). 
Would it be possible to have a simple function, say \html, that takes as arguments the height/depth and the html code, and inserts it into the katex output?
I have actually managed to do it myself, but my code is a bad hack, and I won't even suggest using it as a basis for proper code...
(if this issue has already arisen, feel free to direct me to past discussions)

kevinbarabash:
I think https://github.com/Khan/KaTeX/issues/898 would cover your request for insert an SVG image in the middle of some LaTeX.  `\html` seems useful too.  I think we'd want to put some restrictions on the `\html`.  For instance with `\includegraphics` you can optionally include dimensions.  Making that mandatory would be helpful for both `\html` and `\includegraphics`.  In the latter case, we may not know how big an image is until it's loaded but we need that information during typesetting which happens first.

pzinn:
I hadn't thought of using `\includegraphics`, I'll give it a try.

edemaine:
This probably goes without saying, but `\html` and even `\includegraphics` create rather large security holes if the LaTeX source is untrusted.  (`\html{<script>...</script>}` most notably, but also including an image from an arbitrary host can enable broad tracking.)  We're wondering the same in #1437 with the much simpler notion of `class` and `id` attributes.  I feel like we need a `security` option...

pzinn:
As to contributing myself, I'd be happy to, but to be honest there's just not enough documentation for an outsider to be able to write proper code. For example, what are the different types of arguments to a function? is that explained anywhere? I'm reading defineFunction.js but don't see anything there (incidentally, I suspect L73 should read "math mode", not "text mode")

kevinbarabash:
It's a security hole for pages that render user submitted math input.

kevinbarabash:
The reason why it could be a security hole is for pages that allow users to submit math and then render it to other users.  Rendering arbitrary JS in your own browser isn't a big deal, but this would all a user of such a site to run JS in other people's browsers.

edemaine:
@pzinn You'll get a lot of mileage by looking at the types.  `defineFunction` refers to type `FunctionDefSpec`, and its `props` has type `FunctionPropSpec`, and its `argTypes` has type `ArgType[]` (admittedly, this is a deep chain).  `ArgType` is defined in `types.js` with a decent comment:

```js
[...]

PR #2082, opened 2019-08-20
Add HTML extension

### LINKED ISSUE #90: Support for non-standard macros \class and \cssId

### LINKED ISSUE #2029: Support for \style{}
> Before requesting a feature, please search for [existing issues](https://github.com/KaTeX/KaTeX/issues).

When I am trying to render a latex - `\frac{\style{font-size:14px}R\style{font-size:14px}{\mathit=}\style{font-size:14px}\rho\style{font-size:14px}{\mathit\;}\style{font-size:14px}l}{\style{font-size:14px}A}` 
Screenshot 2019-07-10 at 4 20 27 PM
The latex is generated using Wiris's MathType Plugin.

KaTeX's general goal is to render as LaTeX does.
When the same latex is entered here - http://www.wiris.com/editor/demo/en/developers - it renders as expected.

ylemkimon:
This PR adds four new commands that give access to some HTML features: `\htmlClass`, `\htmlId`, `\htmlStyle`, and `\htmlData`.

Fixes #90.
Fixes #2029.


src/functions/html.js
```diff
@@ -0,0 +1,99 @@
...
+import defineFunction, {ordargument} from "../defineFunction";
+import buildCommon from "../buildCommon";
+import {assertNodeType} from "../parseNode";
+
+import * as html from "../buildHTML";
+import * as mml from "../buildMathML";
+
+defineFunction({
+    type: "html",
+    names: ["\\htmlClass", "\\htmlId", "\\htmlStyle", "\\htmlData"],
```
ylemkimon: In MathJax, they are `\class`, `\cssId`, and `\style`, respectively, (they don't have one that can set data attributes) but I think their names are too generic and we have/will have multiple renderers, so I named them with the prefix `html`.

ylemkimon: `\htmlData` takes key/value pairs as the first argument and sets data attributes. Inspired by #1698.

src/functions/html.js
```diff
@@ -0,0 +1,99 @@
...
+        allowedInText: true,
+    },
+    handler: ({parser, funcName, token}, args) => {
+        const value = assertNodeType(args[0], "raw").string;
+        const body = args[1];
+
+        if (parser.settings.strict) {
+            parser.settings.reportNonstrict("htmlExtension",
+                "HTML extension is disabled on strict mode");
+        }
```
ylemkimon: I'm not sure they are necessary, as the `trust` setting controls them.

src/functions/html.js
```diff
@@ -0,0 +1,99 @@
...
+                    command: "\\htmlId",
+                    id: value,
+                };
+                break;
+            case "\\htmlStyle":
+                attributes.style = value;
+                trustContext = {
+                    command: "\\htmlStyle",
+                    style: value,
+                };
```
ylemkimon: I can't get Flow to work if I use `funcName` as `command` in trustContexts.

src/functions/html.js
kevinbarabash: I like the prefixing.  Thanks for adding `\htmlData` as well.

src/functions/html.js
kevinbarabash: `funcName` is typed as `string`.  It looks like it might be possible to introduce another generic type param to `FunctionDefSpec` and related types to make that work.  I can poke around a bit.

kevinbarabash:
@ylemkimon this is great!

vd-man:
Does not work with fraction, eg:

\htmlId{id1}{a}+\frac{1}{\htmlId{id2}{b}}-c

id1 woks, but id2 does not.

ylemkimon:
@vd-man Both `id1` and `id2` seem to work [here](https://deploy-preview-2082--katex.netlify.com/?data=%7B%22displayMode%22%3Atrue%2C%22leqno%22%3Afalse%2C%22fleqn%22%3Afalse%2C%22throwOnError%22%3Atrue%2C%22errorColor%22%3A%22%23cc0000%22%2C%22strict%22%3A%22ignore%22%2C%22trust%22%3Atrue%2C%22code%22%3A%22%5C%5ChtmlId%7Bid1%7D%7Ba%7D%2B%5C%5Cfrac%7B1%7D%7B%5C%5ChtmlId%7Bid2%7D%7Bb%7D%7D-c%22%7D). Could you give more detail? Thank you.

ylemkimon:
This is ready for review!

phlegmaticprogrammer:
Would it be possible to add some docs for what \htmlId, \htmlStyle, \htmlClass and in particular \htmlData actually do? How can I preview the docs? If I click on "Deploy preview ready!" I get to a preview which has no updated HTML section in the docs.

kevinbarabash:
@phlegmaticprogrammer see https://deploy-preview-2082--katex.netlify.com/docs/next/supported.html#html

phlegmaticprogrammer:
Now it works for me! This is great progress, although some CSS properties seem to be still under full KaTeX control, for example in 

`A*\htmlStyle{color: red; font-family: Arial, Helvetica, sans-serif; font-style: normal; font-size:50px;}{A}`

the color and font-size are applied, but the font-family and style are not. So a general \html would still be necessary I guess.

kevinbarabash:
@phlegmaticprogrammer there's a ticket for adding `\html`, see #1596.

docs/issues.md
```diff
@@ -22,6 +22,8 @@ title: Common Issues
 - MathJax defines `\color` to be like `\textcolor` by default; set KaTeX's
   `colorIsTextColor` option to `true` for this behavior.  KaTeX's default
   behavior matches MathJax with its `color.js` extension enabled.
+- Equivalents of MathJax `\class`, `\cssId`, and `\style` are `\htmlClass`,
+  `\htmlId`, and `\htmlStyle`, respectively, to avoid ambiguity.
```
kevinbarabash: Thanks for adding this.  If people want to use the MathJax variants they can set up some macros.

src/functions/html.js
kevinbarabash: I didn't have time to poke around.  This works.  This is something we can always revisit later.

contrib/render-a11y-string/render-a11y-string.js
```diff
@@ -655,6 +655,10 @@ const handleObject = (
             break;
         }
 
+        case "html": {
+            throw new Error("KaTeX-a11y: html not implemented yet");
+        }
```
kevinbarabash: The `\html*` commands wrap other commands.  I think we could probably call `buildString` on `tree.body`.  Do you see any issues with that approach?

docs/support_table.md
```diff
@@ -457,6 +457,10 @@ use `\ce` instead|
 |\hskip|$w\hskip1em i\hskip2em d$|`w\hskip1em i\hskip2em d`|
 |\hslash|$\hslash$||
 |\hspace|$s\hspace7ex k$|`s\hspace7ex k`|
+|\htmlClass|$\htmlClass{foo}{x}$|`\htmlClass{foo}{x}` Requires `trust` and `strict` [option](options.md)|
+|\htmlData|$\htmlData{foo=a, bar=b}{x}$|`\htmlData{foo=a, bar=b}{x}` Requires `trust` and `strict` [option](options.md)|
+|\htmlId|$\htmlId{bar}{x}$|`\htmlId{bar}{x}` Requires `trust` and `strict` [option](options.md)|
+|\htmlStyle|$\htmlStyle{color: red;}{x}$|`\htmlStyle{color: red;}{x}` Requires `trust` and `strict` [option](options.md)|
```
kevinbarabash: This makes it sound like we need to set `trust: true` and `strict: true`.  Maybe reword it to say something like:
```
Must enable `trust` and disable `strict` [option](options.md)
```

src/functions/html.js
```diff
@@ -0,0 +1,99 @@
...
+                break;
+            case "\\htmlData": {
+                const data = value.split(",");
+                for (let i = 0; i < data.length; i++) {
+                    const keyVal = data[i].split("=");
+                    if (keyVal.length === 2) {
+                        attributes["data-" + keyVal[0].trim()] = keyVal[1].trim();
+                    }
```
kevinbarabash: We should probably throw an error if `keyVal.length !== 2`.

mariusGundersen:
This seems to be merged, but not released yet. What is the timeline for having this in a version available for use? I would really like to use this feature soon

PR #2085, opened 2019-08-21
feat(macro): improve argument parsing

### LINKED ISSUE #611: \overline \mathrm command stacking
`\overline \mathrm x` works in KaTeX and plain LaTeX, and the `MathAtom2` test confirms this.  However, `\overline \mathrm x` does *not* work when `\usepackage{amsmath}` is in effect.  (This quirk is mentioned also here: https://github.com/mathjax/MathJax/issues/677) The trouble is that this quirk causes `texcmp` to fail on this test.  Perhaps we should add an option in the YAML to turn off `amsmath` in this test?

### LINKED ISSUE #924: Difficult interaction between MacroExpander and Parser

### LINKED ISSUE #1027: Single-symbol macro arguments

### LINKED ISSUE #1419: Optional argument checking shouldn't expand
I've found that `\lbrack`, despite being `\def`d to `[`, doesn't get treated as the beginning of an optional argument in LaTeX:
```latex
\begin{array}
...
\\\lbrack 2ex \rbrack
...
\end{array}
```
doesn't get treated the same as `...\\[2ex]` in LaTeX.  However, KaTeX treats them the same (with `\def\lbrack{[}`).  I believe this is because optional argument (`[`) detection uses `\futurelet` which grabs the next symbol, but doesn't expand it.  Assuming this gets verified, we should adjust optional argument detection to use the appropriate `MacroExpander` tools.

### LINKED ISSUE #1720: Macro Expansion and Function Arguments Don't Work Together
What do I expect: `\text\AA` to render as Å
What happens: `\text\AA` generates an error `(ParseError: KaTeX parse error: Got function '\r' as argument to '\text' at position 1: \̲r̲ ̲A)`

Why does this happen?  KaTeX expands macros before processing function arguments

Refs #924

### LINKED ISSUE #2119: \cr and \newline should not take an optional size argument
**Describe the bug:**
```latex
\begin{matrix}
  a&b \cr[10px]
  c&d
\end{matrix}
a\newline[10px]b
```

**Expected behavior:**
`\cr` and `\newline` should not take an optional size argument

### LINKED ISSUE #2319: Math operators require braces when used in superscript/subscript
On katex.org, enter `T_\max`

Resulting behavior: 

> KaTeX parse error: Got function '\max' with no arguments as subscript at position 3: T_\̲m̲a̲x̲

Expected behavior: This should be identical to `T_{\max}`, which does work fine in KaTeX. The `T_\max` does work in LaTeX and MathJax: http://jbergknoff.github.io/mathjax-sandbox/#JFRfXG1heCQ%3D

Superscripts and subscripts should not require braces.

ylemkimon:
**1. Primitive v.s. Macro**
LaTeX has two types of command (control sequence): TeX primitive and macro. TeX primitives are provided by TeX engine. Macros, in contrast, are defined using primitives in the LaTeX kernel (`latex.ltx`) and packages such as `amsmath`. (Macro may expand into primitive commands without taking arguments. They are considered as primitives hereinafter.) Primitives scan argument as it goes, i.e., expanding macros, whereas macros take a properly nested group or non-blank token without expanding. For instance, 
```tex
\def\foo{12}
\sqrt\foo
\underline\foo
```
is equivalent to
```tex
\sqrt12
\underline{12}
```

However, KaTeX uses primitive argument parsing for functions. This PR introduces a new argument type `primitive`, which uses the existing method, and uses it for parsing arguments of primitive commands. It also changes other group parsers such as `parseArgumentGroup` (new) and `parseStringGroup` to use macro argument parsing method, which scans the argument and denotes the end of the argument at the `MacroExpander` level before parsing. This fixes #924, fixes #1027, fixes #1720, and fixes #2319. Closes #611.

**2. Macro**
Delimited parameters in macro and brace handling are implemented in the new `consumeArg` function. Braced control sequence and unbraced replacement text no longer works.[1]

Optional arguments are implemented using a delimited parameter at the `MacroExpander` level. This fixes #1419 and #2288.

Whitespace handling is also improved to consume only in undelimited macro arguments and required/optional arguments. This broke `\\` and this PR fixes it along with fix #2119.

---

Some changes may result in **BREAKING CHANGEs** such as [1], and issues linked above, however they probably were originally wrong behavior.


src/MacroExpander.js
```diff
@@ -108,6 +109,30 @@ export default class MacroExpander implements MacroContextInterface {
...
+            }
+            this.popToken(); // don't include [
+            argument = this.consumeArg(["]"]);
+        } else {
+            argument = this.consumeArg();
+        }
+
+        // indicate the end of an argument
+        const pos = this.lexer.tokenRegex.lastIndex;
+        this.pushToken(new Token("EOF", new SourceLocation(this.lexer, pos, pos)));
```
ylemkimon: TODO: find a better way to indicate the end of an argument

src/MacroExpander.js
```diff
@@ -122,42 +147,82 @@ export default class MacroExpander implements MacroContextInterface {
...
+                --depth;
+                if (depth === -1) {
+                    throw new ParseError("Extra }", tok);
+                }
+            } else if (tok.text === "EOF") {
+                throw new ParseError("Unexpected end of input in a macro argument" +
+                    ", expected '" + (delims && isDelimited ? delims[match] : "}") +
+                    "'", tok);
+            }
+            if (delims && isDelimited) {
```
ylemkimon: `delims && ` is required for Flow to recognize `delims != null`.

src/Parser.js
```diff
@@ -476,31 +475,34 @@ export default class Parser {
...
-            // In KaTeX, for now, both spaces will get consumed.
-            // TODO(edemaine)
-                (i === 0 && !isOptional && this.mode === "math");
-            const arg = this.parseGroupOfType(`argument to '${func}'`,
-                argType, isOptional, baseGreediness, consumeSpaces);
-            if (!arg) {
-                if (isOptional) {
+
+            // \sqrt is a macro only if an optional argument exists
+            if (funcData.type === "sqrt" && i === 1 && optArgs[0] == null) {
```
ylemkimon: I can't seem to find a better way to handle `\sqrt` than hard-coding it.

src/functions/kern.js
```diff
@@ -14,7 +14,7 @@ defineFunction({
     names: ["\\kern", "\\mkern", "\\hskip", "\\mskip"],
     props: {
         numArgs: 1,
-        argTypes: ["size"],
+        argTypes: ["primitive_size"],
```
ylemkimon: BREAKING CHANGE: braced size arguments no longer work.

src/macros.js
```diff
@@ -198,39 +204,59 @@ defineMacro("\\char", function(context) {
 //     \def\macro#1#2#3#4#5#6#7#8#9{expansion}
 // Also the \gdef and \global\def equivalents
 const def = (context, global: boolean) => {
-    let arg = context.consumeArgs(1)[0];
-    if (arg.length !== 1) {
-        throw new ParseError("\\gdef's first argument must be a macro name");
+    let tok = context.popToken();
```
ylemkimon: BREAKING CHANGE: braced control sequence no longer works.

src/macros.js
```diff
@@ -198,39 +204,59 @@ defineMacro("\\char", function(context) {
...
+        } else if (tok.text === "EOF") {
+            throw new ParseError("Expected a macro definition");
+        } else {
+            delimiters[numArgs].push(tok.text);
         }
-        arg = context.consumeArgs(1)[0];
     }
-    // Final arg is the expansion of the macro
+    // replacement text, enclosed in '{' and '}' and properly nested
+    const tokens = context.consumeArg(null, preserveOutermostBraces);
```
ylemkimon: BREAKING CHANGE: unbraced replacement text no longer works

src/Parser.js
```diff
@@ -759,6 +704,36 @@ export default class Parser {
...
+        this.gullet.beginGroup();
+        const firstToken = this.fetch();
+        const expression = this.parseExpression(false, "EOF");
+        const lastToken = this.fetch();
+        // End group namespace before consuming symbol after close brace
+        this.gullet.endGroup();
+        const result = {
+            type: "ordgroup",
+            mode: this.mode,
+            loc: SourceLocation.range(firstToken, lastToken),
```
ylemkimon: This may be a regression, as `firstToken` is the first token after the opening brace, if it exists.

src/functions/genfrac.js
```diff
@@ -444,7 +444,7 @@ defineFunction({
     names: ["\\above"],
     props: {
         numArgs: 1,
-        argTypes: ["size"],
+        argTypes: ["primitive_size"],
```
ylemkimon: BREAKING CHANGE: braced size arguments no longer work.

src/Parser.js
```diff
@@ -698,19 +651,22 @@ export default class Parser {
...
         }
-        if (!res) {
+        if (res == null) {
             return null;
         }
-        if (!optional && res.text.length === 0) {
+        if (res.text.length === 0) {
             // Because we've tested for what is !optional, this block won't
             // affect \kern, \hspace, etc. It will capture the mandatory arguments
             // to \genfrac and \above.
```
ylemkimon: *

ylemkimon:
This breaks the example at the main page:
```tex
% \f is defined as f(#1) using the macro
\f{x} = \int_{-\infty}^\infty
    \hat \f\xi\,e^{2 \pi i \xi x}
    \,d\xi
```
Any suggestion to replace it is welcome!

src/macros.js
```diff
@@ -198,39 +214,63 @@ defineMacro("\\char", function(context) {
 //     \def\macro#1#2#3#4#5#6#7#8#9{expansion}
 // Also the \gdef and \global\def equivalents
 const def = (context, global: boolean) => {
-    let arg = context.consumeArgs(1)[0];
-    if (arg.length !== 1) {
-        throw new ParseError("\\gdef's first argument must be a macro name");
+    let tok = context.popToken();
+    const name = tok.text;
+    if (name[0] !== "\\") {
```
ylemkimon: TODO: allow active character macro

kevinbarabash:
> This breaks the example at the main page

Is there any way to change the example to make it work?  I tried the following:
```
\def\f#1{f(#1)}
\f{x} = \int_{-\infty}^\infty
    \hat \f\xi\,e^{2 \pi i \xi x}
    \,d\xi
```
on the deploy preview and got:
```
KaTeX parse error: Unexpected end of input in a macro argument, expected '}' at position 97: …infty \hat \̲f̲\xi\,e^{2 \pi i…
```
Quicklatex.com also produces an error which is a good sign.  My TeX-fu isn't very strong unfortunately.  `\def\f#1{f(#1)}` seems reasonable, but apparently it is not. :(

kevinbarabash:
I figured out the issue.  We have to use explicit groups for both `\f` and `\hat`.  The following works:
```
\def\f#1{f(#1)}
\f{x} = \int_{-\infty}^\infty
    \hat{\f\{xi\}},e^{2 \pi i \xi x}
    \,d\xi
```

kevinbarabash:
I'm glad to see that we closing more gaps between KaTeX and LaTeX.  I think this change is going to require some docs on how to migrate existing TeX that people may have written that will no longer work after these changes.

kevinbarabash:
What I typed before was wrong, I meant to write:
```
\def\f#1{f(#1)}
\f{x} = \int_{-\infty}^\infty
    \hat{\f{xi}},e^{2 \pi i \xi x}
    \,d\xi
```
Unfortunately, this doesn't give the output we want:
Screen Shot 2019-12-31 at 6 10 11 PM

On the bright side this is also what LaTeX does:
Screen Shot 2019-12-31 at 6 11 10 PM

What if we change `\def\f#1{f(#1)}` to `\def\func#1#2{#1(#2)}`?  Then the example would look like:
```
% \func is a macro defined to be {#1(#2)}
\func{f}{x} = \int_{-\infty}^\infty
    \func{\hat f}{xi}\,e^{2 \pi i \xi x}
    \,d\xi
```
The result has the hat in the right spot.
Screen Shot 2019-12-31 at 6 16 34 PM

ylemkimon:
I've updated `\f` to take two arguments to demonstrate a macro can take multiple arguments and a control sequence as an argument. I wanted to demonstrate KaTeX supports a delimited macro, but it's currently not possible to define a delimited macro. But I think this is out of this PR's scope. I've also added a migration guide.

Whadup:
I'm really looking forward to this merge =)

edemaine:
We should undo #2289 in this PR. 🙂

kevinbarabash:
@ylemkimon in #2288 you commented:
> Actually, #2085 will add support for delimited macros.

but earlier in this PR you also commented:
> I wanted to demonstrate KaTeX supports a delimited macro, but it's currently not possible to define a delimited macro. But I think this is out of this PR's scope. I've also added a migration guide.

Looking at the code it looks delimiters are supported.  Is that correct?  If so, could you add some tests that demonstrate this?

ylemkimon:
@kevinbarabash I meant it's not possible to define a delimited macro using `options.macros`.

test/katex-spec.js
```diff
@@ -3173,32 +3166,36 @@ describe("A macro expander", function() {
...
-        //expect`\gdef{\foo}{}`.not.toParse();
+    });
+
+    it("\\gdef defines macros with delimited parameter", function() {
+        expect`\gdef\foo|#1||{#1}\text{\foo| x y ||}`.toParseLike`\text{ x y }`;
+        expect`\gdef\foo#1|#2{#1+#2}\foo 1 2 |34`.toParseLike`12+34`;
+        expect`\gdef\foo#1#{#1}\foo1^{23}`.toParseLike`1^{23}`;
+        expect`\gdef\foo|{}\foo`.not.toParse();
+        expect`\gdef\foo#1|{#1}\foo1`.not.toParse();
+        expect`\gdef\foo#1|{#1}\foo1}|`.not.toParse();
```
ylemkimon, line 3198: Tests for delimited macro are added here.

kevinbarabash: I was wondering what the `|` were for.  I'm so used to brackets and parens as delimiters, I forgot that `|` is a delimiter as well. 🙃

src/MacroExpander.js
```diff
@@ -122,42 +148,89 @@ export default class MacroExpander implements MacroContextInterface {
...
+            }
+        } while (depth !== 0 || isDelimited);
+        // If the argument found ... has the form ‘{<nested tokens>}’,
+        // ... the outermost braces enclosing the argument are removed
+        if (start.text === "{" && tokens[tokens.length - 1].text === "}") {
+            tokens.pop();
+            tokens.shift();
+        }
+        tokens.reverse(); // to fit in with stack order
+        return {tokens, start, end: tok};
```
kevinbarabash, line 211: `start` and `end` can be outside the range of `tokens`, correct?

src/MacroExpander.js
```diff
@@ -122,42 +148,89 @@ export default class MacroExpander implements MacroContextInterface {
...
+    }
+
     /**
      * Consume the specified number of arguments from the token stream,
      * and return the resulting array of arguments.
      */
```
kevinbarabash, line 217: Can you update this comment to include that delimiters are consumed if specified?  Also, are there any constraints on the size of `delimiters` based on the value of `numArgs`?  If so, it might be good to add a check to make sure it meets that constraint.

src/MacroExpander.js
```diff
@@ -202,8 +271,8 @@ export default class MacroExpander implements MacroContextInterface {
                 "need to increase maxExpand setting");
         }
         let tokens = expansion.tokens;
+        const args = this.consumeArgs(expansion.numArgs, expansion.delimiters);
```
kevinbarabash, line 274: The reason this was a moved up is so that we consume delimiters even if there are no args, is that right?  What's an example of when this could happen?

src/Parser.js
```diff
@@ -480,31 +473,24 @@ export default class Parser {
...
-            // won't know what to do with them.  This can only happen with
-            // macros, e.g. \frac\foo\foo where \foo expands to a space symbol.
-            // In LaTeX, the \foo's get treated as (blank) arguments.
-            // In KaTeX, for now, both spaces will get consumed.
-            // TODO(edemaine)
-                (i === 0 && !isOptional && this.mode === "math");
+
+            if ((funcData.primitive && argType == null) ||
+                // \sqrt expands into primitive if optional argument doesn't exist
+                (funcData.type === "sqrt" && i === 1 && optArgs[0] == null)) {
```
kevinbarabash, line 481: Is `sqrt` the only macro that does this?

src/Parser.js
```diff
@@ -480,31 +473,24 @@ export default class Parser {
...
+                argType, isOptional, baseGreediness);
+            if (isOptional) {
+                optArgs.push(arg);
+            } else if (arg != null) {
+                args.push(arg);
+            } else { // should be unreachable
+                throw new ParseError("Null argument, please report this as a bug");
             }
-            (isOptional ? optArgs : args).push(arg);
         }
```
kevinbarabash, line 494: Thanks for reworking this.  Not having to do a `continue` and getting rid of the ternary is much cleaner.

src/Parser.js
```diff
@@ -518,64 +504,50 @@ export default class Parser {
         type: ?ArgType,
         optional: boolean,
         greediness: ?number,
-        consumeSpaces: boolean,
     ): ?AnyParseNode {
         switch (type) {
             case "color":
-                if (consumeSpaces) {
-                    this.consumeSpaces();
-                }
```
kevinbarabash, line 527: Where did all of these calls to `consumeSpaces()` go?

src/Parser.js
```diff
@@ -518,64 +504,50 @@ export default class Parser {
...
+                    throw new ParseError("Expected group as " + name, this.fetch());
                 }
+                return group;
             }
             case "original":
             case null:
             case undefined:
-                return this.parseGroup(name, optional, greediness,
-                    undefined, undefined, consumeSpaces);
+                return this.parseArgumentGroup(optional);
```
kevinbarabash, line 550: Having more specialized functions makes the code easier to understand.

src/MacroExpander.js
```diff
@@ -108,6 +108,32 @@ export default class MacroExpander implements MacroContextInterface {
         this.stack.push(...tokens);
     }
 
+    /**
+     * Find an macro argument without expanding tokens and append the array of
+     * tokens to the token stack. Uses Token as a container for the result.
+     */
```
kevinbarabash, line 114: If the argument is a group will the contents of the return token have to be reparsed?

src/Parser.js
```diff
@@ -598,49 +570,20 @@ export default class Parser {
...
+        while ((nextToken = this.fetch()).text !== "EOF") {
+            str += nextToken.text;
             this.consume();
         }
-        this.expect(groupEnd);
-        this.mode = outerMode;
-        return firstToken.range(lastToken, str);
+        this.consume(); // consume the end of the argument
+        argToken.text = str;
+        return argToken;
```
kevinbarabash, line 586: If we replace the `.text` won't the `argToken.loc` be incorrect?

src/Parser.js
```diff
@@ -705,7 +645,9 @@ export default class Parser {
     parseSizeGroup(optional: boolean): ?ParseNode<"size"> {
         let res;
         let isBlank = false;
-        if (!optional && this.fetch().text !== "{") {
+        // don't expand before parseStringGroup
+        this.gullet.consumeSpaces();
+        if (!optional && this.gullet.future().text !== "{") {
```
kevinbarabash, line 650: Is the comment referring to the `if` block?

src/Parser.js
```diff
@@ -678,7 +618,7 @@ export default class Parser {
      */
     parseColorGroup(optional: boolean): ?ParseNode<"color-token"> {
         const res = this.parseStringGroup("color", optional);
-        if (!res) {
+        if (res == null) {
```
kevinbarabash, line 621: This doesn't change the behaviour does it?

src/Parser.js
```diff
@@ -764,52 +706,60 @@ export default class Parser {
     }
 
     /**
-     * If `optional` is false or absent, this parses an ordinary group,
-     * which is either a single nucleus (like "x") or an expression
-     * in braces (like "{x+y}") or an implicit group, a group that starts
-     * at the current position, and ends right before a higher explicit
+     * Parses an argument.
+     */
+    parseArgumentGroup(optional: boolean, mode?: Mode): ?AnyParseNode {
```
kevinbarabash, line 711: Could the result be typed as `?ParseNode<"ordgroup">` or does that cause other issues with the flow types?

src/Parser.js
```diff
@@ -764,52 +706,60 @@ export default class Parser {
     }
 
     /**
-     * If `optional` is false or absent, this parses an ordinary group,
-     * which is either a single nucleus (like "x") or an expression
-     * in braces (like "{x+y}") or an implicit group, a group that starts
-     * at the current position, and ends right before a higher explicit
+     * Parses an argument.
+     */
```
kevinbarabash, line 710: Are arguments the only groups where we specify a mode?  If so, it might be helpful to add a comment saying such.

src/Parser.js
```diff
@@ -764,52 +706,60 @@ export default class Parser {
...
+            return null;
+        }
+        const outerMode = this.mode;
+        if (mode) { // Switch to specified mode
+            this.switchMode(mode);
+        }
+
+        this.gullet.beginGroup();
+        const expression = this.parseExpression(false, "EOF");
+        this.expect("EOF"); // expect the end of the argument
```
kevinbarabash, line 723: I wonder if we could use a different token to mark the end of arguments.  Overloading `EOF` seems like it could be a bit confusing.  Maybe something for a future PR given that there's already so much going on in this one.

src/defineFunction.js
```diff
@@ -223,6 +228,10 @@ export function defineFunctionBuilders<NODETYPE: NodeType>({
     });
 }
 
+export const normalizeArgument = function(arg: AnyParseNode): AnyParseNode {
+    return arg.type === "ordgroup" && arg.body.length === 1 ? arg.body[0] : arg;
+};
```
kevinbarabash, line 233: Is it possible that `arg` could be a series of nested `ordgroup`s?  If so, would we want to unwrap each of them?

src/environments/array.js
```diff
@@ -61,9 +61,10 @@ function parseArray(
     |},
     style: StyleStr,
 ): ParseNode<"array"> {
-    // Parse body of array with \\ temporarily mapped to \cr
     parser.gullet.beginGroup();
-    parser.gullet.macros.set("\\\\", "\\cr");
+    // \cr is equivalent to \\ without the optional size argument (see below)
+    // TODO: provide helpful error when \cr is used outside the array environment
```
kevinbarabash, line 66: Can you raise a ticket for this `TODO`?

src/environments/array.js
```diff
@@ -125,9 +126,18 @@ function parseArray(
...
-            const cr = assertNodeType(parser.parseFunction(), "cr");
-            rowGaps.push(cr.size);
+        } else if (next === "\\\\") {
+            parser.consume();
+            let size;
+            // \def\Let@{\let\\\math@cr}
+            // \def\math@cr{...\math@cr@}
+            // \def\math@cr@{\new@ifnextchar[\math@cr@@{\math@cr@@[\z@]}}
+            // \def\math@cr@@[#1]{...\math@cr@@@...}
+            // \def\math@cr@@@{\cr}
```
kevinbarabash, line 136: Does this collapse multiple `\cr`s in a row or something like that?

src/functions/def.js
```diff
@@ -88,41 +88,65 @@ defineFunction({
...
+                if (parseInt(tok.text) !== numArgs + 1) {
+                    throw new ParseError(
+                        `Argument number "${tok.text}" out of order`);
+                }
+                numArgs++;
+                delimiters.push([]);
+            } else if (tok.text === "EOF") {
+                throw new ParseError("Expected a macro definition");
+            } else {
+                delimiters[numArgs].push(tok.text);
```
kevinbarabash, line 132: Can `tok.text` be any string or should we be checking it against a list of allowed delimiters?

src/functions/def.js
```diff
@@ -88,41 +88,65 @@ defineFunction({
...
+                    throw new ParseError(`Invalid argument number "${tok.text}"`);
+                }
+                if (parseInt(tok.text) !== numArgs + 1) {
+                    throw new ParseError(
+                        `Argument number "${tok.text}" out of order`);
+                }
+                numArgs++;
+                delimiters.push([]);
+            } else if (tok.text === "EOF") {
+                throw new ParseError("Expected a macro definition");
```
kevinbarabash, line 130: How would this case arise?

src/macros.js
```diff
@@ -240,7 +253,7 @@ defineMacro("\\char", function(context) {
 // \renewcommand{\macro}[args]{definition}
 // TODO: Optional arguments: \newcommand{\macro}[args][default]{definition}
 const newcommand = (context, existsOK: boolean, nonexistsOK: boolean) => {
-    let arg = context.consumeArgs(1)[0];
+    let arg = context.consumeArg().tokens;
```
kevinbarabash, line 256: Returning an object with named properties is much nicer.  Thank you.

test/errors-spec.js
```diff
@@ -166,7 +167,7 @@ describe("Parser.expect calls:", function() {
     describe("#parseSpecialGroup expecting braces", function() {
         it("complains about missing { for color", function() {
             expect`\textcolor#ffffff{text}`.toFailWithParseError(
-                   "Expected '{', got '#' at position 11:" +
+                   "Invalid color: '#' at position 11:" +
```
kevinbarabash, line 170: Nice!

test/errors-spec.js
```diff
@@ -94,11 +94,12 @@ describe("Parser:", function() {
     describe("#parseArguments", function() {
         it("complains about missing argument at end of input", function() {
             expect`2\sqrt`.toFailWithParseError(
-                   "Expected group after '\\sqrt' at end of input: 2\\sqrt");
+                   "Expected group as argument to '\\sqrt'" +
+                   " at end of input: 2\\sqrt");
```
kevinbarabash, line 98: I like the new wording.

test/errors-spec.js
```diff
@@ -176,23 +177,23 @@ describe("Parser.expect calls:", function() {
         // Can't test for the [ of an optional group since it's optional
         it("complains about missing } for color", function() {
             expect`\textcolor{#ffffff{text}`.toFailWithParseError(
-                   "Invalid color: '#ffffff{text' at position 12:" +
-                   " \\textcolor{#̲f̲f̲f̲f̲f̲f̲{̲t̲e̲x̲t̲}");
+                   "Unexpected end of input in a macro argument," +
+                   " expected '}' at end of input: …r{#ffffff{text}");
```
kevinbarabash, line 181: Removing the position information may make it hard to track down errors in some cases.  The underlining was also helpful because it made it really easy to see the range of the issue.  Looking at some of the test cases below, we still have both of these features in those test cases.  Is it hard to maintain those features in these test cases because how macro expansion has changed?

test/errors-spec.js
```diff
@@ -293,22 +296,22 @@ describe("Lexer:", function() {
         it("reject 3-digit hex notation without #", function() {
             expect`\textcolor{1a2}{foo}`.toFailWithParseError(
                    "Invalid color: '1a2'" +
-                   " at position 12: \\textcolor{1̲a̲2̲}{foo}");
+                   " at position 11: \\textcolor{̲1̲a̲2̲}̲{foo}");
```
kevinbarabash, line 299: Why is there an extra space inside of `\textcolor{}` in the error message?

test/katex-spec.js
```diff
@@ -2566,23 +2581,6 @@ describe("A smash builder", function() {
...
-    it("should have paddings applied inside an enclose", function() {
-        const markup = katex.renderToString(r`\fbox\textcolor{red}{x}`);
-        expect(markup).toContain("boxpad");
-    });
-
-    it("should have paddings applied inside a square root", function() {
-        const markup = katex.renderToString(r`\sqrt\textcolor{red}{x}`);
-        expect(markup).toContain("padding-left");
-    });
-});
```
kevinbarabash, line 2584: Why were these test cases removed?

test/katex-spec.js
```diff
@@ -2990,12 +2983,17 @@ describe("A macro expander", function() {
...
             "\\foo": "(#1)",
-            "\\bar": "x",
+            "\\bar": "xyz",
         }}));
     });
 
+    it("should allow properly nested group for macro argument", function() {
+        expect`\foo{e^{x_{12}+3}}`.toParseLike("(e^{x_{12}+3})",
+            new Settings({macros: {"\\foo": "(#1)"}}));
+    });
```
kevinbarabash, line 2995: Nice!

test/katex-spec.js
```diff
@@ -3015,9 +3013,8 @@ describe("A macro expander", function() {
             new Settings({macros: {"\\foo": "x"}}));
         // \frac is a macro and therefore expandable
         expect`\noexpand\frac xy`.toParseLike`xy`;
-        // TODO(ylem): #2085
         // \def is not expandable, so is not affected by \noexpand
-        // expect`\noexpand\def\foo{xy}\foo`.toParseLike`xy`;
+        expect`\noexpand\def\foo{xy}\foo`.toParseLike`xy`;
```
kevinbarabash, line 3017: Yay!

docs/migration.md
```diff
@@ -0,0 +1,24 @@
...
+## `\def`
+`\def` no longer accepts a control sequence enclosed in braces. For example,
+`\def{\foo}{}` no longer works and should be changed to `\def\foo{}`.
+
+It also no longer accepts replacement text not enclosed in braces. For example,
+`\def\foo1` no longer works and should be changed to `\def\foo{1}`.
+
+## `\newline` and `\cr`
+`\newline` and `\cr` no longer takes an optional size argument. To specify vertical
+spacing, `\\` should be used.
```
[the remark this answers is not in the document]
ylemkimon: @kevinbarabash Partly, yes. Those changed space handling and they are updated to match their definitions.

src/MacroExpander.js
```diff
@@ -122,42 +148,89 @@ export default class MacroExpander implements MacroContextInterface {
...
+        // The argument for an undelimited parameter is the next nonblank
+        // token, unless that token is ‘{’, when the argument will be the
+        // entire {...} group that follows.
+        const tokens: Token[] = [];
+        const isDelimited = delims && delims.length > 0;
+        if (!isDelimited) {
+            // Ignore spaces between arguments.  As the TeXbook says:
+            // "After you have said ‘\def\row#1#2{...}’, you are allowed to
+            //  put spaces between the arguments (e.g., ‘\row x n’), because
+            //  TeX doesn’t use single spaces as undelimited arguments."
```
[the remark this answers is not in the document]
ylemkimon: @kevinbarabash They are intended, but if you think they are unnecessary, I'll remove them.

src/MacroExpander.js
ylemkimon: @kevinbarabash No, they are always the first and last or the second and second to last tokens.

src/MacroExpander.js
ylemkimon: @kevinbarabash Any non-alphabetic characters following the control sequence is considered to be delimiters. For instance, `\def\abc|2c{123}` has delimiter `|2c`.

src/Parser.js
ylemkimon: @kevinbarabash Yes, for functions we currently have and probably most LaTeX macros.

src/Parser.js
ylemkimon: They have been moved to match TeX behavior: https://github.com/KaTeX/KaTeX/blob/35470fd7db11202e01d3097bbc57fe2a1eb67acd/src/MacroExpander.js#L115-L120 https://github.com/KaTeX/KaTeX/blob/35470fd7db11202e01d3097bbc57fe2a1eb67acd/src/MacroExpander.js#L155-L170

src/MacroExpander.js
ylemkimon: @kevinbarabash They're not parsed here, but only fetched (buffered).

src/Parser.js
ylemkimon: @kevinbarabash Normally, they should match since the end of the argument is denoted using `EOF`. Otherwise, it's an error and will be caught by the `EOF` mismatch.

src/Parser.js
ylemkimon: @kevinbarabash Yes, but also to `this.gullet.consumeSpaces();`, on why it's used instead of `this.consumeSpaces();`.

src/Parser.js
ylemkimon: @kevinbarabash It doesn't.

src/Parser.js
ylemkimon: @kevinbarabash I agree. This is my major concern, too. This will probably cause a lot of issues that are hard to debug, but I can't seem to find a simple better way.

src/defineFunction.js
ylemkimon: @kevinbarabash I'm not sure whether it's possible, but that's not caused by changes in this PR.

src/environments/array.js
ylemkimon: @kevinbarabash No, they are just how `\\` is defined in LaTeX and amsmath.

src/functions/cr.js
```diff
@@ -5,43 +5,29 @@ import defineFunction from "../defineFunction";
...
 import {calculateSize} from "../units";
-import ParseError from "../ParseError";
 import {assertNodeType} from "../parseNode";
 
-// \\ is a macro mapping to either \cr or \newline.  Because they have the
-// same signature, we implement them as one megafunction, with newRow
-// indicating whether we're in the \cr case, and newLine indicating whether
-// to break the line in the \newline case.
-
+// \DeclareRobustCommand\\{...\@xnewline}
```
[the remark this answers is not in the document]
ylemkimon: @kevinbarabash They are omitted commands for clarity.

src/functions/def.js
```diff
@@ -88,41 +88,65 @@ defineFunction({
...
+        let insert;
+        const delimiters = [[]];
+        // <parameter text> contains no braces
+        while (parser.gullet.future().text !== "{") {
+            tok = parser.gullet.popToken();
+            if (tok.text === "#") {
+                // If the very last character of the <parameter text> is #, so that
+                // this # is immediately followed by {, TeX will behave as if the {
+                // had been inserted at the right end of both the parameter text
+                // and the replacement text.
```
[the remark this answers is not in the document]
ylemkimon: @kevinbarabash https://github.com/KaTeX/KaTeX/blob/35470fd7db11202e01d3097bbc57fe2a1eb67acd/test/katex-spec.js#L3195

src/functions/def.js
ylemkimon: @kevinbarabash AFAIK, every valid TeX token can be used as delimiters.

src/functions/def.js
ylemkimon: @kevinbarabash If the end is reached without the macro body, e.g., `\def\abc`.

test/errors-spec.js
ylemkimon: @kevinbarabash In case of macros, arguments should be scanned (fetched) first, so scanning errors are detected before parsing errors, hence changes in error position.

test/errors-spec.js
ylemkimon: @kevinbarabash It is probably combining underline not shown properly.

test/katex-spec.js
ylemkimon: @kevinbarabash They're test cases added in #1707, which relied on edge-case and previous (wrong) behaviors.

src/Parser.js
ylemkimon: @kevinbarabash Yes, I'll update the PR.

src/environments/array.js
kevinbarabash: I don't think I'll ever get used to reading .dtx files. 😓

src/functions/def.js
kevinbarabash: This seems really obvious now. 😅

test/errors-spec.js
kevinbarabash: I see, this is just how we've been reporting parse for some time now.

test/errors-spec.js
kevinbarabash: I don't see an extra space anymore. 🤷‍♂️

kevinbarabash (approved):
Thanks for answering all my questions and for all of your hard work on this PR.

src/MacroExpander.js
ylemkimon: @kevinbarabash Sorry, my mistake. `end` can be outside the range of tokens, when it's delimited. In that case, `end` is the last delimiter.

src/Parser.js
ylemkimon: @kevinbarabash Their lengths may not match, due to whitespace normalization, etc.

ylemkimon:

docs/migration.md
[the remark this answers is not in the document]
ronkok: This is a late comment, sorry. But there is a breaking change in that `\raisebox` previously accepted unbraced dimensions and now it does not. That might be worth a note in the migration guide.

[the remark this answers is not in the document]
ylemkimon: @ronkok Thank you for finding it! That was an unexpected change, but might be fixed with #2139. I'll add a note to the migration guide.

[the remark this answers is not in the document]
ronkok: I don't think there is anything to "fix".  LaTeX `\raisebox` does not accept unbraced dimensions, so the new behavior is not a bug. It's just needs a note in the migration guide.

PR #2086, opened 2019-08-22
Added “Notable” to the list of users

fabiospampinato:


fabiospampinato:
Thanks for merging 👍 I'll probably send a second PR for changing the url if that's not too much trouble, as I'm in the process of launching the actual website of the app.

PR #2091, opened 2019-09-01
Create globalGroup option to place definitions in global scope

cirosantilli:
If true, this allows \newcommand definitions to persist across calls to
renderToString.

Fix https://github.com/KaTeX/KaTeX/issues/2070

If this is functionally good let me know the bits I've missed, e.g. docs, unit tests, etc and I'll do them.

Manually tested with:

```
#!/usr/bin/env nodejs

var assert = require('assert');
var katex = require('.')

// globalGroup \def
macros = {}
console.log(katex.renderToString("\\def\\foo{myfoo}", {macros: macros, throwOnError: true, globalGroup: true}));
console.log(katex.renderToString("\\foo", {macros: macros, throwOnError: true}));

// !globalGroup \def
macros = {}
console.log(katex.renderToString("\\def\\foo{myfoo}", {macros: macros, throwOnError: true}));
raised = false;
try {
  console.log(katex.renderToString("\\foo", {macros: macros, throwOnError: true}));
} catch(error) {
  raised = true;
}
assert(raised);

// globalGroup \newcommand
macros = {}
console.log(katex.renderToString("\\newcommand{\\foo}[0]{myfoo}", {macros: macros, throwOnError: true, globalGroup: true}));
console.log(katex.renderToString("\\foo", {macros: macros, throwOnError: true}));

// !globalGroup \newcommand
macros = {}
console.log(katex.renderToString("\\newcommand{\\foo}[0]{myfoo}", {macros: macros, throwOnError: true}));
raised = false;
try {
  console.log(katex.renderToString("\\foo", {macros: macros, throwOnError: true}));
} catch(error) {
  raised = true;
}
assert(raised);
```


kevinbarabash:
Cool!  It's nice to see this behavior wasn't too hard to add.  This will need tests and documentation before this can be accepted.  To fix the flow error you'll have to add `settings: Settings` to the `MacroContextInterface` interface.  I think the name of this new setting should be something that more clearly communicates what it's doing, maybe something like `retainMacros` or `shareMacros`.

edemaine:
@kevinbarabash Neither of those names are an accurate description of what this option does. Passing in a shared `macros` object is what allows sharing/preserving macros. It just currently only works for gloval macros. This option makes definitions at the top of math mode (but not nesfed groups) global, so that those will also be shared. Hence the current name. But open to other suggestions on better names!

edemaine:
Ah, I wrore that before reading the code. It turns out that this option makes *all* user definitions global. This seems bad to me. I think it would be better to just not call beginGrouo and endGroup at the top level of each math mode, when the option is on.

Both behaviors are reasonably described by "globalGroup", suggesting we do need a better name...

edemaine:
Alternatively, I could see making just `\newcommand` working globally. But defining `\def` to work the same as `\gdef` seems wrong to me.

cirosantilli:
> Alternatively, I could see making just `\newcommand` working globally. But defining `\def` to work the same as `\gdef` seems wrong to me.

I will also try to remove the scope creation with this option. But is there any functional difference between removing the scopes and making both `\gdef` and `\def` do the same and be global?

cirosantilli:
Moved the if checks to `this.gullet.beginGroup();`. This does feel like the better approach.

cirosantilli:
Pushed doc and a few tests.

cirosantilli:
Ping. Let me know if I missed anything :+1:

kevinbarabash:
After re-reading the docs in this diff, the name is making more sense.  We can always add an alias if we come up with a better name later.

cirosantilli:
No worries! Thanks for looking into this!

PR #2103, opened 2019-09-27
Set border-collapse: collapse in vlist

### LINKED ISSUE #2100: Not equals in table row does not align correctly [pen included]
https://codepen.io/HughGrovesArup/pen/OJLadYp

As can be seen when in a table row the / is misaligned.

image

ylemkimon:
Fixes #2100.

`vlist` uses a table layout and `border-collapse` which affects it is set to `separate` in `table` tags. This resets it.


kevinbarabash:
Thanks for figuring this out.

PR #2116, opened 2019-10-09
Add support for \cal

### LINKED ISSUE #2115: \cal support
KaTeX supports the new-style `\mathcal{...}` but not the old-style `{\cal ...}` font format.  We should ideally support both, as we do for `\rm`/`\mathrm` etc.; see the [support table](https://katex.org/docs/supported.html#style-color-size-and-font).

This oversight was recently mentioned to me by the author of a project using KaTeX as a common issue with KaTeX where they needed to fall back to MathJax, so I think it's relevant to the real world.

edemaine:
Fix #2115 by adding \cal to list of old-style font commands. (Surprisingly tiny change!)
Add tests for all old-style font commands.

Example with tester:
```latex
\mathcal{ABC} ABC \cal ABC \rm ABC
```
image

### DIFF
--- a/test/katex-spec.js
+++ b/test/katex-spec.js
@@ -1601,6 +1601,15 @@ describe("A font parser", function() {
         const built = getBuilt`a\boldsymbol{}b\boldsymbol{=}c\boldsymbol{+}d\boldsymbol{++}e\boldsymbol{xyz}f`;
         expect(built).toMatchSnapshot();
     });
+
+    it("old-style fonts work like new-style fonts", () => {
+        expect`\rm xyz`.toParseLike`\mathrm{xyz}`;
+        expect`\sf xyz`.toParseLike`\mathsf{xyz}`;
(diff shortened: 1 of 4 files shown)


test/katex-spec.js
```diff
@@ -1601,6 +1601,15 @@ describe("A font parser", function() {
...
     });
+
+    it("old-style fonts work like new-style fonts", () => {
+        expect`\rm xyz`.toParseLike`\mathrm{xyz}`;
+        expect`\sf xyz`.toParseLike`\mathsf{xyz}`;
+        expect`\tt xyz`.toParseLike`\mathtt{xyz}`;
+        expect`\bf xyz`.toParseLike`\mathbf{xyz}`;
+        expect`\it xyz`.toParseLike`\mathit{xyz}`;
+        expect`\cal xyz`.toParseLike`\mathcal{xyz}`;
+    });
```
kevinbarabash: nice!

PR #2122, opened 2019-10-12
Add support for \expandafter, \noexpand, \edef, \let, and \long

### LINKED ISSUE #1413: support \let
`\let` copies the content of a command into a new command.

ylemkimon:
I'm surprised that they are fairly simple to implement! Fixes #1413.


src/MacroExpander.js
```diff
@@ -282,6 +310,26 @@ export default class MacroExpander implements MacroContextInterface {
...
+     */
+    expandTokens(tokens: Token[]): Token[] {
+        const output = [];
+        const oldStackLength = this.stack.length;
+        tokens = tokens.slice(); // make a shallow copy
+        tokens.reverse();
+        this.pushTokens(tokens);
+        while (this.stack.length > oldStackLength) {
+            // expand the token following \noexpand to itself
+            const expanded = this.expandOnce(true);
```
ylemkimon: TODO: Primitive commands such as `\def` and `MathSymbol` defined using macros should not be expanded here.

src/MacroExpander.js
```diff
@@ -342,4 +387,11 @@ export default class MacroExpander implements MacroContextInterface {
             symbols.text.hasOwnProperty(name) ||
             implicitCommands.hasOwnProperty(name);
     }
+
+    isExpandable(name: string): boolean {
+        return this.macros.has(name) ? this.macros.hasCurrent(name) // overriden
+                || !unexpandableMacros.hasOwnProperty(name)
+            : functions.hasOwnProperty(name)/* && !functions[name].primitive*/
```
ylemkimon: #2085

ylemkimon:
This is ready for review.

kevinbarabash:
Since this depends on #2138, maybe set the base of this PR to be `internal` instead of `master.

src/Token.js
```diff
@@ -23,6 +23,7 @@ export interface LexerInterface {input: string, tokenRegex: RegExp}
 export class Token {
     text: string;
     loc: ?SourceLocation;
+    noexpand: ?number; /* 1: treat as \relax, 2: pass to the parser */
```
kevinbarabash: Having constants for these would make the code clearer.

src/functions/def.js
```diff
@@ -4,37 +4,86 @@ import ParseError from "../ParseError";
...
+            unexpandable: !parser.gullet.isExpandable(tok.text),
+        };
+    }
+    parser.gullet.macros.set(name, macro, global);
+};
+
+// <assignment> -> <non-macro assignment>|<macro assignment>
+// <non-macro assignment> -> <simple assignment>|\global<non-macro assignment>
+// <macro assignment> -> <definition>|<prefix><macro assignment>
+// <prefix> -> \global|\long|\outer
```
kevinbarabash: What does this mean?

src/functions/def.js
```diff
@@ -4,37 +4,86 @@ import ParseError from "../ParseError";
...
 defineFunction({
     type: "internal",
-    names: ["\\global"],
+    names: [
+        "\\global", "\\long",
+        "\\\\globallong", // can’t be entered directly
+    ],
```
kevinbarabash: Why are `\global` and `\long` grouped together?

src/functions/def.js
```diff
@@ -4,37 +4,86 @@ import ParseError from "../ParseError";
...
+        }
+    }
+    return tok;
+};
+
+const letCommand = (parser, name, tok, global) => {
+    let macro = parser.gullet.macros.get(tok.text);
+    if (macro == null) {
+    // if macro is undefined at this moment, set noexpand to 2
+    // and unexpandable to not expand it later and pass to the parser
```
kevinbarabash: Is this to handle a case like `\let\foo=\bar \def\bar{hello, world} \foo` where the value of the `let` is defined after the `let`?

src/functions/def.js
```diff
@@ -65,12 +114,66 @@ defineFunction({
...
+            type: "internal",
+            mode: parser.mode,
+        };
+    },
+});
 
+// <simple assignment> -> <let assignment>
+// <let assignment> -> \futurelet<control sequence><token><token>
+//     | \let<control sequence><equals><one optional space><token>
+// <equals> -> <optional spaces>|<optional spaces>=
```
kevinbarabash: Is this a description of possible expansions?  ~Where do we match the `equals`?~ I see, `getRHS` handles this.

src/functions/def.js
```diff
@@ -65,12 +114,66 @@ defineFunction({
             }
             arg = parser.gullet.consumeArgs(1)[0];
         }
+        if (funcName === "\\edef" || funcName === "\\xdef") {
+            arg = parser.gullet.expandTokens(arg);
+            arg.reverse();
+        }
```
kevinbarabash: Can you add a comment about why we reverse `arg` here?

src/functions/def.js
```diff
@@ -65,12 +114,66 @@ defineFunction({
...
+        numArgs: 0,
+        allowedInText: true,
+    },
+    handler({parser, funcName}) {
+        const name = checkControlSequence(parser.gullet.popToken());
+        const middle = parser.gullet.popToken();
+        const tok = parser.gullet.popToken();
+        letCommand(parser, name, tok, funcName === "\\\\globalfuture");
+        parser.gullet.pushToken(tok);
+        parser.gullet.pushToken(middle);
```
kevinbarabash: After reading the description of `\futurelet` this makes sense to me.

src/MacroExpander.js
```diff
@@ -266,16 +272,24 @@ export default class MacroExpander implements MacroContextInterface {
...
+    expandTokens(tokens: Token[]): Token[] {
         const output = [];
         const oldStackLength = this.stack.length;
-        this.pushToken(new Token(name));
+        this.pushTokens(tokens);
         while (this.stack.length > oldStackLength) {
-            const expanded = this.expandOnce();
+            const expanded = this.expandOnce(true);
             // expandOnce returns Token if and only if it's fully expanded.
             if (expanded instanceof Token) {
```
kevinbarabash: So whenever `expandOnce()` returns an array it's not considered fully expanded?

src/functions/def.js
```diff
@@ -4,37 +4,86 @@ import ParseError from "../ParseError";
...
+const getRHS = (parser) => {
+    let tok = parser.gullet.popToken();
+    if (tok.text === "=") { // consume optional equals
+        tok = parser.gullet.popToken();
+        if (tok.text === " ") { // consume one optional space
+            tok = parser.gullet.popToken();
+        }
+    }
+    return tok;
+};
```
kevinbarabash: It seems like this function should throw if the `tok.text === "="` check fails.  Am I missing something?

src/macros.js
```diff
@@ -99,6 +113,28 @@ export function defineMacro(name: string, body: MacroDefinition) {
...
+    }
+    return {tokens: [t], numArgs: 0};
+});
+
+defineMacro("\\expandafter", function(context) {
+    // TeX first reads the token that comes immediately after \expandafter,
+    // without expanding it; let’s call this token t. Then TeX reads the
+    // token that comes after t (and possibly more tokens, if that token
+    // has an argument), replacing it by its expansion. Finally TeX puts
+    // t back in front of that expansion.
```
kevinbarabash: This behavior feels similar to `\futurelet` in some ways.

test/katex-spec.js
```diff
@@ -2975,6 +2975,25 @@ describe("A macro expander", function() {
         }}));
     });
 
+    it("should delay expansion if preceded by \\expandafter", function() {
+        expect`\expandafter\foo\bar`.toParseLike("x+y", new Settings({macros: {
+            "\\foo": "#1+#2",
+            "\\bar": "xy",
+        }}));
+        expect`\expandafter\foo\def\foo{x}`.not.toParse();
+        expect`\def\bar{\def\foo{}}\expandafter\bar\foo`.not.toParse();
```
kevinbarabash: The first one shouldn't parse because there's nothing to expand with `\def`, right?  I'm not sure about why the second one shouldn't parse though.

test/katex-spec.js
```diff
@@ -2975,6 +2975,25 @@ describe("A macro expander", function() {
...
+    });
+
+    it("should not expand if preceded by \\noexpand", function() {
+        expect`\noexpand\foo y`.toParseLike("y",
+            new Settings({macros: {"\\foo": "x"}}));
```
kevinbarabash: It looks like the `\foo` is ignored in this case.  Is that correct?

test/katex-spec.js
```diff
@@ -3156,6 +3175,14 @@ describe("A macro expander", function() {
         //expect`\gdef{\foo}{}`.not.toParse();
     });
 
+    it("\\xdef should expand definition", function() {
+        expect`\def\foo{a}\xdef\bar{\foo}\def\foo{}\bar`.toParseLike`a`;
+        expect`\def\foo{a}\xdef\bar{\def\noexpand\foo{}}\foo\bar\foo`.toParseLike`a`;
+        expect`\def\foo{a}\xdef\bar{\expandafter\foo\noexpand\foo}\def\foo{b}\bar`
+            .toParseLike`ab`;
```
kevinbarabash: I get the first one, but the second two are a bit mind warping.

test/katex-spec.js
```diff
@@ -3177,13 +3204,23 @@ describe("A macro expander", function() {
             "\\x\\def\\x{5}\\x}\\x").toParseLike`1{2{34}35}3`;
     });
 
-    it("\\global needs to followed by \\def", () => {
+    it("\\global needs to followed by macro prefixes, \\def or \\edef", () => {
         expect`\global\def\foo{}\foo`.toParseLike``;
+        expect`\global\edef\foo{}\foo`.toParseLike``;
         expect`\def\DEF{\def}\global\DEF\foo{}\foo`.toParseLike``;
+        expect`\global\global\def\foo{}\foo`.toParseLike``;
+        expect`\global\long\def\foo{}\foo`.toParseLike``;
```
kevinbarabash: It looks like multiple global/long commands in a row are the same as having a single one of those commands, is that right?

test/katex-spec.js
```diff
@@ -3225,6 +3262,24 @@ describe("A macro expander", function() {
         expect(macros["\\foo"]).toBeTruthy();
     });
 
+    it("\\let copies the definition", () => {
+        expect`\let\foo=\frac\def\frac{}\foo12`.toParseLike`\frac12`;
+        expect`\def\foo{1}\let\bar\foo\def\foo{2}\bar`.toParseLike`1`;
+        expect`\let\foo=\kern\edef\bar{\foo1em}\let\kern=\relax\bar`.toParseLike`\kern1em`;
+        expect`\let\foo{\frac\foo1}{2}`.toParseLike`\frac{1}{2}`;
```
kevinbarabash: I didn't realize the `=` was optional when using `\let`.  In the first one it looks like `\frac` is being redefined, but the result doesn't include this redefinition.  I think it would be useful to have comments for some of these tests to help people that aren't as familiar with the interections between `let` and `def`.

test/__snapshots__/katex-spec.js.snap
```diff
@@ -562,6 +562,42 @@ exports[`A font parser \\boldsymbol should inherit mbin/mrel from argument 1`] =
...
+        "body": {
+          "type": "ordgroup",
+          "body": [
+            {
+              "type": "mathord",
+              "loc": {
+                "end": 39,
+                "lexer": {
+                  "input": "\\\\def\\\\bold{\\\\bgroup\\\\bf\\\\let\\\\next= }\\\\bold{a}",
+                  "lastIndex": 40
```
kevinbarabash: It's nice seeing the input here.  It makes reviewing snapshot tests easier.

src/functions/def.js
ylemkimon: @kevinbarabash It's the (formal) grammar of TeX from the TeXbook

src/functions/def.js
ylemkimon: @kevinbarabash Tthey work in the same manner as a prefix to the macro assignment.

src/functions/def.js
ylemkimon: This is to handle cases where non-macros are overwritten after `\let`, e.g., `\let\foo=\frac \def\frac{123} \foo12`.

src/MacroExpander.js
ylemkimon: Yes.

src/functions/def.js
ylemkimon: The equal sign (`=`) is optional in assignments.

test/katex-spec.js
ylemkimon: I'll add comments on test cases.

test/katex-spec.js
ylemkimon: Yes.

src/Token.js
```diff
@@ -23,6 +23,8 @@ export interface LexerInterface {input: string, tokenRegex: RegExp}
 export class Token {
     text: string;
     loc: ?SourceLocation;
+    noexpand: ?boolean; // don't expand the token
+    treatAsRelax: ?boolean; // used in \noexpand
```
kevinbarabash: Is any combination of these valid?

ylemkimon: It's possible in theory, but there is no use yet.

PR #2123, opened 2019-10-13
\relax should stop Parser from expanding

ylemkimon:
Add `scanning` mode.


kevinbarabash:
@ylemkimon why did you close this?

PR #2127, opened 2019-10-14
add spacing on left of fleqn display math

kzvi:
The `fleqn` option from amsmath doesn't only left-align display math but also adds an indent. This makes Katex's behavior with the `fleqn` option more similar to what happens in latex.


kzvi:
OK I did it.

kevinbarabash:
[clabot:check]

kzvi:
[clabot:check]

PR #2134, opened 2019-10-21
feat(function): add `allowedInArgument` instead of `greediness` property

### LINKED ISSUE #2046: ^\underline and greediness
```latex
error 
\text{Cu + 2}{{\text{H}}_2}\text{S}{{\text{O}}_4}~^\underline{\underline{\Delta}}~\text{CuS}{{\text{O}}_4}\text{ + S}{{\text{O}}_2}\uparrow + \ 2{{\text{H}}_2}\text{O}
ok
\text{Cu + 2}{{\text{H}}_2}\text{S}{{\text{O}}_4}~^{\underline{\underline{\Delta}}}~\text{CuS}{{\text{O}}_4}\text{ + S}{{\text{O}}_2}\uparrow + \ 2{{\text{H}}_2}\text{O}
problem:
~^ => ~^{\underline{\underline{\Delta}}}  why?
How to use macros to solve this problem？
```

ylemkimon:
Use boolean `allowedInArgument` property instead, which is more consistent with LaTeX.

BREAKING CHANGE: \cfrac, \color, \textcolor, \colorbox, \fcolorbox are no longer allowed in arguments.

Fixes #2046.


kevinbarabash:
> BREAKING CHANGE: \cfrac, \color, \textcolor, \colorbox, \fcolorbox are no longer allowed in arguments.

I tried `\sqrt[\color{red}{3}]{x+1}` on https://deploy-preview-2134--katex.netlify.com/#demo and it worked.  Can you provide an example using `\color` that breaks under the proposed changed?

kevinbarabash:
I see, `\sqrt\textcolor{red}{x}` fails now and to make it work we have to do `\sqrt{\textcolor{red}{x}}`.  Although the code changes are different the breaking behavior feels similar to that of #2085 for a user perspective.  I think we'll want to document changes people will need to make just in case they have any TeX code that breaks with the proposed changes.

ylemkimon:
I've added a migration guide in #2085.

src/functions/genfrac.js
```diff
@@ -319,6 +317,31 @@ defineFunction({
     mathmlBuilder,
 });
 
+defineFunction({
+    type: "genfrac",
+    names: ["\\cfrac"],
+    props: {
+        numArgs: 2,
+    },
```
kevinbarabash, line 325: I didn't realize that `\cfrac` wasn't allowed as an argument, but it makes sense that it isn't.

test/katex-spec.js
```diff
@@ -849,13 +849,6 @@ describe("A color parser", function() {
         expect(newColorExpression).toParse();
     });
 
-    it("should have correct greediness", function() {
```
kevinbarabash, line 852: I think there may be another reference to `greediness` in the `A style change parser` describe block.

src/functions/genfrac.js
```diff
@@ -374,7 +397,7 @@ defineFunction({
     names: ["\\genfrac"],
     props: {
         numArgs: 6,
-        greediness: 6,
```
kevinbarabash, line 377: `allowedInArgument` is easier to understand than greediness especially given that we had to give it different numbers to work properly.

kevinbarabash (approved):
LGTM.  Thanks for making this change.

ylemkimon:
@kevinbarabash Thank you for the reviews. I've created and pinned a migration guide issue.

PR #2135, opened 2019-10-22
Add \message, \errmessage, and \show for debugging

ylemkimon:


ylemkimon:
@kevinbarabash Mostly internal developers and advanced users. I can't seem to find a good way to add tests.

kevinbarabash:
@ylemkimon macros are executed during parsing so we should be able to write a parse test and mock `console.log` and do an assertion.  I'll check our your branch and see what I can do.

kevinbarabash:
I didn't add a test for `\show` since I was worried about some of the unicode characters in the output.

PR #2136, opened 2019-10-22
Drop IE 9/10 support

ylemkimon:
There are about three months left until IE 10 end-of-life and I think their usages have dropped to reasonable rate. We've released v0.11.1 last month, but maybe we can release another version, if necessary. (Related: #1919, #1409)


ronkok:
@ylemkimon I realize that line of code has a comment stating that it is there only to resolve an IE problem. But I think the comment is wrong. The screenshots clearly show a problem if the line of code is removed. 

I think we should keep the line of code and remove the comment that only IE made it necessary.

.browserslistrc
```diff
@@ -4,8 +4,8 @@ not dead
 
 # Browsers with more than 0.1% global usage, excluding old browsers
 > 0.1%
-not chrome < 45
-not ie < 9
+not ie < 11
+not safari < 9
```
ylemkimon: Safari 5.1 shows high usage, which is probably due to Safari for Windows, which is no longer supported.

kevinbarabash: @ylemkimon this seems reasonable.  I am curious about what you're looking at for usage numbers.  Can you post a link to your source?

src/functions/utils/assembleSupSub.js
```diff
@@ -17,8 +17,6 @@ export const assembleSupSub = (
     slant: number,
     baseShift: number,
 ): DomSpan => {
-    // IE 8 clips \int if it is in a display: inline-block. We wrap it
-    // in a new span so it is an inline, and works.
```
kevinbarabash: Do we still need to wrap it?

ronkok: @kevinbarabash  One commit early in this PR removed that wrap. The result was that the `subGroup` below an `\int` or a `\sum` had their top padding removed and were squashed up against the big operator.  That happened in both Chrome and Firefox. 

@ylemkimon then put the wrap span back in and removed the comment stating that it was there only for IE. I think that was the right move.

.browserslistrc
ylemkimon: @kevinbarabash `Can I use` has a browser usage table: https://caniuse.com/usage-table. `browserslist` also uses this via package `caniuse-lite`.

kevinbarabash: I've used caniuse.com for quite some time, but I've never seen that particular chart.  The Chrome for Android number seems a little suspicious.  Regardless, cool chart!

PR #2137, opened 2019-10-26
Remove checkNodeType and assert/checkAtomFamily

ylemkimon:
We used functions with $FlowFixMe as Flow couldn't refine ParseNode using its type. It seems the issue has been fixed and complicated function calls can be removed.


src/Parser.js
```diff
@@ -329,21 +327,18 @@ export default class Parser {
...
-                    opNode.alwaysHandleSupSub = true;
+                    base.limits = limits;
+                    base.alwaysHandleSupSub = true;
+                } else if (base && base.type === "operatorname"
+                        && base.alwaysHandleSupSub) {
+                    const limits = lex.text === "\\limits";
+                    base.limits = limits;
                 } else {
-                    opNode = checkNodeType(base, "operatorname");
-                    if (opNode && opNode.alwaysHandleSupSub) {
```
kevinbarabash: This was awkward.  I'm glad we're able to get rid of this.

src/functions/genfrac.js
```diff
@@ -382,17 +382,10 @@ defineFunction({
...
-
-        let rightNode = checkNodeType(args[1], "atom");
-        if (rightNode) {
-            rightNode = assertAtomFamily(args[1], "close");
-        }
-        const rightDelim = rightNode ? delimFromValue(rightNode.text) : null;
+        const leftDelim = args[0].type === "atom" && args[0].family === "open"
+            ? delimFromValue(args[0].text) : null;
+        const rightDelim = args[1].type === "atom" && args[1].family === "close"
+            ? delimFromValue(args[1].text) : null;
```
kevinbarabash: This is a big improvement

ylemkimon:

PR #2138, opened 2019-10-27
Move \global and \def to functions

ylemkimon:
TeX primitive commands are more like functions, so move to functions and return new parseNode type `internal`, which is skipped during `parseExpression`.


edemaine:
@ylemkimon Could you clarify the sense in which `\def` is more like a function than a macro? I assumed functions were a match for things that produced output, but `\def`s do not. Could you show an example that parses more correctly this way? (Apologies if this is mentioned in another issue/PR -- I'm far from caught up.)  Would also make sense to add this as a test.

ylemkimon:
@edemaine `\def` is a primitive TeX command, which is unexpandable and consumed in the TeX's "stomach", whereas macros are expandable and consumed in the TeX's "gullet". Currently in KaTeX, `\def` is defined as a macro that expands into an empty string before reaching the "stomach". So, for instance, `2^\def\foo{1}2` expands into `2^2` in KaTeX, but doesn't work in TeX.

This is required in two pending PRs:
- #2122: `\def` should not be affected by `\noexpand` or `\expandafter`
- #2139: Register (variable) command will be defined using functions and `\global` should work with them.

test/katex-spec.js
```diff
@@ -3523,10 +3523,10 @@ describe("The maxSize setting", function() {
 
 describe("The maxExpand setting", () => {
     it("should prevent expansion", () => {
-        expect`\gdef\foo{1}\foo`.toParse();
-        expect`\gdef\foo{1}\foo`.toParse(new Settings({maxExpand: 2}));
-        expect`\gdef\foo{1}\foo`.not.toParse(new Settings({maxExpand: 1}));
```
kevinbarabash: It looks like the behavior was wrong before this PR.  It it safe to say that the following will pass now?
```
expect`\gdef\foo{1}\foo`.toParse(new Settings({maxExpand: 1}));
```

contrib/render-a11y-string/render-a11y-string.js
```diff
@@ -655,6 +655,10 @@ const handleObject = (
             break;
         }
 
+        case "internal": {
+            break;
+        }
```
kevinbarabash: In reality we should never have to handle this case since all all internal nodes in the AST should get expanded, right?

src/functions/assignment.js
```diff
@@ -0,0 +1,79 @@
+//@flow
```
kevinbarabash: The name of this file feels a bit misleading.  Could you renamed it to `internal.js`?

edemaine:
@ylemkimon Thanks for the explanation! Ideally we should mention in a comment somewhere that KaTeX macros correspond to TeX gullet and KaTeX functions correspond to TeX stomach.  I was (mostly) aware of this, but wasn't really aware of how `\def` fit in. `2^\def\foo{1}2` is a great example to think about, and I think we should add it as a test too (unless you already did and I missed it).

contrib/render-a11y-string/render-a11y-string.js
edemaine: I believe the `\def` function returns an `internal` node (at least that's what the PR main text says), and this throws it away.

src/functions/assignment.js
edemaine: I think `assignment` refers to `\def` and `\gdef` which is the purpose of this file.  (The idea is that other commands might also generate `internal`.)  I could see calling it `def.js`.  The name `assignment.js` is presumably future-proofing for an eventual `\let` definition? But I'd still lean toward `def.js`.

kevinbarabash: The reason why I suggested internal.js was that it also handles `\global`, but if `\global` can only be used with `\def` then def.js should be fine as well.

test/katex-spec.js
ylemkimon: @kevinbarabash That seems better!

src/functions/assignment.js
ylemkimon: Changed to `def.js`.

test/katex-spec.js
```diff
@@ -3174,8 +3179,7 @@ describe("A macro expander", function() {
 
     it("\\global needs to followed by \\def", () => {
         expect`\global\def\foo{}\foo`.toParseLike``;
-        // TODO: This doesn't work yet; \global needs to expand argument.
-        //expect`\def\DEF{\def}\global\DEF\foo{}\foo`.toParseLike``;
+        expect`\def\DEF{\def}\global\DEF\foo{}\foo`.toParseLike``;
```
ylemkimon: It turns out this fixes another test case.

PR #2139, opened 2019-10-27
feat(parser): add integer, dimen, glue group parsers

### LINKED ISSUE #1936: Kern commands accept braced size argument
In LaTeX, kern commands, such as `\kern`, `\mkern`, `\hskip`, `\hspace`, and `\mskip`, don't accept braced size argument, i.e., don't allow `\kern{1mu}`.

I'm currently working on argument parsing, and it'd be better to separate unbraced and braced size argument parsing, but it'd be a breaking change.

ylemkimon:
**Size argument**
Primitive commands, such as `\above` and kern commands don't not allow braced size argument. This PR introduces a new argument type `size_primitive` to parse it. For backward compatibility, braced size arguments are allowed in non-strict mode. (**BREAKING CHANGE**) Fixes #1936.

Furthermore, a blank size argument was allowed for most commands. This PR introduces a new argument type `size_or_blank`, which allows a blank argument, and use it only for `\genfrac`.[2]


kevinbarabash:
@ylemkimon the tests are passing now.  Is there more that you wanted to do with this PR or is it ready for review now?

ylemkimon:
This incorporates changes from #2085, so it'd be preferable to review it first.

PR #2146, opened 2019-11-08
[TEST] Use Type 1 Euler fonts from amsfonts

ylemkimon:


ronkok:
I don't understand everything that this PR does. I will point out that the functions `\oiint` and `\oiiint` are done by superimposing an SVG oval over the glyphs ∬ and ∭ (U+222C and U+222C) from the fonts KaTeX_Size1-Regular.ttf or KaTeX_Size2-Regular.ttf. So the horizontal alignment of those ovals is hard-coded and tightly linked to the glyph geometry in those existing fonts.

If this PR substitutes a different glyph for ∬ or ∭, then the horizontal alignment of those ovals will break.

ronkok:
Also, a few KaTeX glyphs are located at Unicode code points that may not align with other fonts. A substitution of a new font needs a line-by-line review of `symbols.js` and `macros.js` to avoid bugs.

ylemkimon:
@ronkok Sorry for lack of explanation. This was just for testing the font. `katex-fonts` originally traced MetaFont sources for Euler fonts. However, they were [reshaped as of version 3.0](http://www.tug.org/TUGboat/tb29-2/tb92hagen-euler.pdf) and since the updates were made directly to the Type 1 files, the MetaFont sources have been removed. This PR will update to use the new font. Fortunately, metrics have not been changed, so I think it'll be compatible without any changes.

PR #2147, opened 2019-11-08
[TEST] Use modified xbbold for tracing msbm10 at 1000 mag

ylemkimon:


ylemkimon:
I'm not sure there are any noticeable changes here too.

kevinbarabash:
Thanks for trying some different tracing settings.

PR #2151, opened 2019-11-11
Minor symbols.js formatting changes

ylemkimon:


src/symbols.js
```diff
@@ -693,7 +687,6 @@ defineSymbol(text, main, textord, "\u0131", "\\i", true);
 defineSymbol(text, main, textord, "\u0237", "\\j", true);
 defineSymbol(text, main, textord, "\u00df", "\\ss", true);
 defineSymbol(text, main, textord, "\u00e6", "\\ae", true);
-defineSymbol(text, main, textord, "\u00e6", "\\ae", true);
```
kevinbarabash: How did that sneak in, lol.

src/symbols.js
```diff
@@ -879,18 +868,9 @@ for (let i = 0; i < 10; i++) {
 // but they are not actually in the font, nor are they supported by the
 // Unicode accent mechanism, so they fall back to Times font and look ugly.
 // TODO(edemaine): Fix this.
-export const extraLatin = "ÇÐÞçþ";
```
kevinbarabash: Having unicode codes is safer since it avoids accidentally using the wrong symbol.  Is that what prompted these changes?

src/symbols.js
```diff
@@ -621,22 +617,22 @@ defineSymbol(text, main, textord, "}", "\\}");
...
-defineSymbol(text, main, textord, "<", "\\textless"); // in T1 fontenc
-defineSymbol(text, main, textord, ">", "\\textgreater"); // in T1 fontenc
+defineSymbol(math, main, open, "[", "\\lbrack", true);
+defineSymbol(text, main, textord, "[", "\\lbrack", true);
+defineSymbol(math, main, close, "]", "\\rbrack", true);
+defineSymbol(text, main, textord, "]", "\\rbrack", true);
+defineSymbol(math, main, open, "(", "\\lparen", true);
+defineSymbol(math, main, close, ")", "\\rparen", true);
+defineSymbol(text, main, textord, "<", "\\textless", true); // in T1 fontenc
+defineSymbol(text, main, textord, ">", "\\textgreater", true); // in T1 fontenc
```
kevinbarabash: Makes sense to me to set `acceptUnicodeChar` to `true` for these.

src/symbols.js
```diff
@@ -665,8 +661,6 @@ defineSymbol(math, main, op, "\u2a02", "\\bigotimes");
 defineSymbol(math, main, op, "\u2a01", "\\bigoplus");
 defineSymbol(math, main, op, "\u2a00", "\\bigodot");
 defineSymbol(math, main, op, "\u222e", "\\oint");
-defineSymbol(math, main, op, "\u222f", "\\oiint");
-defineSymbol(math, main, op, "\u2230", "\\oiiint");
```
kevinbarabash: Why did these get removed?

src/symbols.js
```diff
@@ -746,10 +739,6 @@ defineSymbol(text, main, mathord, "\u00a3", "\\textsterling", true);
 defineSymbol(math, ams, textord, "\u2720", "\\maltese");
 defineSymbol(text, ams, textord, "\u2720", "\\maltese");
 
-defineSymbol(text, main, spacing, "\u00a0", "\\ ");
-defineSymbol(text, main, spacing, "\u00a0", " ");
-defineSymbol(text, main, spacing, "\u00a0", "~");
```
kevinbarabash: These are dupes, right?

src/symbols.js
ylemkimon: @kevinbarabash Currently, there are no glyphs for them, but instead they are created by overlaying SVG ovals onto the glyphs from \iint and \iiint: see https://github.com/KaTeX/KaTeX/pull/1430.

src/symbols.js
ylemkimon: @kevinbarabash I didn't think of that, but that would be another upside. This is for ASCII-safety and consistency with other symbols.

src/symbols.js
ylemkimon: @kevinbarabash Yes, they have duplicates in:
https://github.com/KaTeX/KaTeX/blob/0d8830af304c722814df6060a170faa5d94d0592/src/symbols.js#L592-L594

src/symbols.js
kevinbarabash: Thanks for noticing this and cleaning it up.

PR #2154, opened 2019-11-14
auto-render: do not touch text nodes w/o formulas

nag5000:
Currently `renderMathInElement` replaces all text nodes w/o formula(s) with new ones, which don't really need to be recreated. This PR fixes this.

Example:
```html
<div>
  <p>foo bar</p>
  <p>\( x = \pm\sqrt{a^2 + b^2} \)</p>
</div>
```

In this DOM fragment there are 5 text nodes: `"↵  "`, `"foo bar"`, `"↵  "`, `"\( x = \pm\sqrt{a^2 + b^2} \)"`, `"↵"`. Currently `renderMathInElement(div)` processes and replaces all of them, but I expect that only text nodes with formula(s) should be modified. E.g. text node `"foo bar"` is replaced with new text node with the same content "foo bar".

Why I dig into it?
auto-render somehow breaks ember.js rendering in my application in some specific cases because all text nodes are replaced (= new references), not only text nodes with formula(s).

Also, it seems like this PR gives a tiny performance boost (I tested dev build only, ~25 samples): 
image

### DIFF
--- a/contrib/auto-render/auto-render.js
+++ b/contrib/auto-render/auto-render.js
@@ -19,6 +19,13 @@ const splitWithDelimiters = function(text, delimiters) {
  */
 const renderMathInText = function(text, optionsCopy) {
     const data = splitWithDelimiters(text, optionsCopy.delimiters);
+    if (data.length === 1 && data[0].type === 'text') {
+        // There is no formula in the text.
+        // Let's return null which means there is no need to replace
+        // the current text node with a new one.
+        return null;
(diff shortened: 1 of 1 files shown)


nag5000:
[clabot:check]

nag5000:
auto-render exports only `renderMathInElement` function.

kevinbarabash:
Nice!  Perf improvements are always welcome. 🙂

PR #2160, opened 2019-11-28
Test was checking numerator instead of denominator

ktraunmueller:
This pull request fixes what seems to be a small copy/paste issue.

### DIFF
--- a/test/katex-spec.js
+++ b/test/katex-spec.js
@@ -585,7 +585,7 @@ describe("An over/brace/brack parser", function() {
     it("should create a demonimator from the atoms after \\over", function() {
         const parse = getParsed(complexOver)[0];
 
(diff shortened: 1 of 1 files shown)


ktraunmueller:
[clabot:check]

test/katex-spec.js
```diff
@@ -585,7 +585,7 @@ describe("An over/brace/brack parser", function() {
     it("should create a demonimator from the atoms after \\over", function() {
```
edemaine: Let's fix this typo at the same time:
```suggestion
     it("should create a denomimator from the atoms after \\over", function() {
```

ktraunmueller: Right, I'll change it `denominator`.

PR #2162, opened 2019-12-01
Add bra-ket notation

ruiokada:
Hello,

I would like support for some bra-ket notation macros. I've tested the macros on KaTeX.org and the following screenshot is the result I got:

Screen Shot 2019-12-01 at 12 27 08 PM

Please let me know if there is anything else that needs to be done. Thanks!


ruiokada:
[clabot:check]

src/macros.js
```diff
@@ -960,6 +960,15 @@ defineMacro("\\argmin", "\\DOTSB\\operatorname*{arg\\,min}");
...
 defineMacro("\\plim", "\\DOTSB\\mathop{\\operatorname{plim}}\\limits");
 
+//////////////////////////////////////////////////////////////////////
+// braket.sty
+// http://ctan.math.washington.edu/tex-archive/macros/latex/contrib/braket/braket.pdf
+
+defineMacro("\\bra", "\\mathinner{\\langle{#1}|}");
+defineMacro("\\ket", "\\mathinner{|{#1}\\rangle}");
+defineMacro("\\braket", "\\mathinner{\\langle{#1}\\rangle}");
+defineMacro("\\ketbra", "\\mathinner{|{#1}\\rangle\\langle{#2}|}");
```
ylemkimon: It seems `braket` doesn't define `\ketbra`:
> NOT defined is "\ketbra" (for projection operators) because I prefer \ket{ } \bra{ }.

ruiokada: Hi, I've removed `\ketbra` and added `\Bra` and `\Ket` to be more faithful to the original `braket`.

ronkok:
Good stuff.

Please also update the documentation file at `./docs/supported.md`.

Another help would be a test in `./test/katex-spec.js` to ensure that this builds.

ruiokada:
Sorry for the billion commits, a lot unexpected things happened with the docs. I added everything to `./docs/supported.md` under a new section called "Special Notation" since I don't think bra-kets qualify as explicit functions or operators. I added tests for bra-kets in `./test/katex-spec.js` by following suit for the statmath package and adding tests for what `\ket{\psi}` etc. should parse as.

kevinbarabash:
@ruiokada don't worry about the extra commits, we squash merge all PRs.  Thanks for updating the docs and adding tests.

src/macros.js
```diff
@@ -930,6 +930,16 @@ defineMacro("\\argmin", "\\DOTSB\\operatorname*{arg\\,min}");
 defineMacro("\\argmax", "\\DOTSB\\operatorname*{arg\\,max}");
 defineMacro("\\plim", "\\DOTSB\\mathop{\\operatorname{plim}}\\limits");
 
+//////////////////////////////////////////////////////////////////////
+// braket.sty
+// http://ctan.math.washington.edu/tex-archive/macros/latex/contrib/braket/braket.pdf
```
kevinbarabash: Thanks for reference the ctan package and linking to its docs.

spino17:
Can anyone please tell me how to use this with var html = katex.renderToString("c = \\ket{\\Phi}", {
    throwOnError: false
}); 
I am not able to get bra-ket notation rendered.

Yue-Zhengyuan:
Is there any plan to provide support for `\Braket` (with **uppercase B**)? Current `\braket` does not adjust its size with the contents.

kevinbarabash:
@Yue-Zhengyuan you could probably define `\Bracket` using a macro.  I'm not sure how `\Bracket` adjusts the size of its contents though.

ruiokada:
@kevinbarabash The `\Braket` implementation in braket resizes content by using`\right` + `\left` (to resize the angle brackets) and what seems like some search/replace function (to resize pipes (`|`) and `\vert`s by prepending them with `\middle`). The `\right` + `\left` part can be easily replicated in KaTeX with the macro `\gdef\Braket#1{\left\langle{#1}\right\rangle}`. The search and replace will take a bit more work within KaTeX itself but I'm not familiar enough with KaTeX's codebase to see how readily this can be implemented.

edemaine:
Note that there's already a prototype implementation in #3171.

PR #2175, opened 2019-12-14
feat(mathml): add soft line breaks to MathML

### LINKED ISSUE #2169: MathML automatic soft line breaks
Thanks to @edemaine and PR #1287, KaTeX HTML renders with a soft line break after a top-level binary operator or relational operator. This aligns with p. 173 of __The TeXbook__.

This issue is written to request that KaTeX's MathML output have the same feature.

Note that the MathML `linebreakstyle` attribute is not supported by Firefox, nor will it be supported by the Chromium MathML project, [at least not anytime soon](https://github.com/mathml-refresh/mathml/issues/112#issuecomment-562985420). So a MathML line break implementation will have to be closely  patterned after KaTeX's current HTML method. That is, it will need to break the top level into multiple `mrow`s, with each break occurring at a binary or relational operator. That would map `a + b = d` into:

    <mrow><mi>a</mi><mo>+</mo></mrow>
    <mrow><mi>b</mi><mo>=</mo></mrow>
    <mrow><mi>d</mi></mrow>

I'll get around to this eventually, but I'm very busy for the next month or two. If anyone else wants to do it first, feel free.

ronkok:
This PR revises KaTeX MathML so that it renders with a soft line break after a top-level binary operator or relational operator. This aligns with KaTeX HTML and with p. 173 of __The TeXbook__. It also resolves #2169.

Unfortunately, the MathML `linebreakstyle` attribute is not supported by Firefox or by the upcoming Chromium MathML project. So in order to achieve soft line breaks, this PR acts in a way similar to KaTeX HTML. It breaks the top level expression into a series of `<mrow>` elements. Each element ends with a binary or relational operator and the visual result is soft line breaks after those operators.

This tactic has other consequences. The `<semantics>` element only recognizes one presentation child. Multiple top level `<mrows>` are not rendered if inside `<semantics>`. So this PR removes the `<semantics>` tag and its other child, the `<annotation>` element. I've taken the text content that was previously in the `<annotation>` element and put it into a `data-tex` attribute of the span element that wraps the `<math>`.

So the wrapper tag now looks something like: `<katex-mathml data-tex="a = b + c">`.

Before removing the `<semantics>` element, I tried to see if that one-child rule is enforced by user-agent CSS and could therefore be countermanded by KaTeX CSS. It turns out that Firefox once did implement this rule via CSS, but no more. It must now be written into the browser source code. It was after that attempt failed that I resorted to the `data-tex` tactic.

To see an example in action, use Firefox to view the MathML [test page](http://webmathcomparison.net/#symbols).


ronkok:
Note to reviewer: One line of code is written:

    const isOpenDelimiter = node.attributes.stretchy &&
            // $FlowFixMe
            "([{⌈⌊⎰⌜⌞⟦⟨⟪⟬⟮⦃⦅".indexOf(node.children[0].text) > -1;

I could instead attach an attribute of `direction="open"` or `direction="close"` to the node while in `symbols.js` or `delimsizing.js`. The MathML spec does not include a `direction` attribute, but it would not cause an error. The browser just ignores it. If I had that attribute, I would not need to examine the text and I would not need `$FlowFixMe`. 

Let me know if you prefer the attribute or the text.

kevinbarabash:
Thanks for the PR.  Screen readers don't seem to handle `<semantics>` containing presentational MathML very well so it's probably okay to get rid of this.  I am curious whether screen readers will still read our MathML at all after this change.

ronkok:
Yeah, it would be nice if browsers supported the  MathML `linebreakstyle` attribute, but they don't. My hope is that the effect of this PR on screenreaders is nothing worse than a comma pause after an operator.

Think of John Wayne reading math.

ronkok:
Note to reviewer: While working on this PR, I discovered that KaTeX MathML emitted space characters without enclosing elements. That is, it emitted `<mi>a</mi> <mi>b</mi>` instead of `<mi>a</mi><mspace> </mspace><mi>b</mi>`. That isn't right and it prevented this PR from working. So I fixed it.

Then I had to change `\operatorname` to catch `<mspace>` and revert back to a character, in order to deal with operators such as `\limsup`. 

So there is a little more code in this PR than one would expect.

kevinbarabash:
@ronkok the switch of `<annotation>` to `data-tex` breaks contrib/copy-tex/katex2tex.js.  Was this switch necessary to add soft line breaks?  Using `<annotation>` seems like a more official way of embedding non-MathML data in MathML.  It would be nice if we could keep using it.

ronkok:
> Was this switch necessary to add soft line breaks?

Yes, `<annotation>`'s parent element, `<semantics>`, takes exactly two children. Only the first child is rendered visually. So all the visual elements must be lumped together in that first element, and this prevents soft line breaks.

This entire PR is irrelevant when HTML is the visual output. Perhaps it would be best if I were to withdraw this PR.

kevinbarabash:
KaTeX supports rendering just the MathML so maybe some people are using that for visual output.

kevinbarabash:
> Unfortunately, the MathML linebreakstyle attribute is not supported by Firefox or by the upcoming Chromium MathML project.

@ronkok do you know if there any plans to support it in the future or have they explicitly stated that they won't be adding this feature?

ronkok:
The [current version](https://github.com/mathml-refresh/mathml/issues/112#issuecomment-647935099) of MathML Core will not support `linebreakstyle`.  Future versions may support it. But such support would make it necessary for the layout algorithm to make a [second pass](https://github.com/mathml-refresh/mathml/issues/112#issuecomment-562913234). Given browser vendors emphasis on speed, this may never happen.

PR #2180, opened 2019-12-23
auto-render: add "option" to ignoredTags

nag5000:
We should ignore `<option>` by default, because it's a Text node, it can't contain html.

Currently `renderMathInElement` transforms `<select><option>Hello $x+y$</option></select>` image into invalid html
```html
<select>
  <option>Hello <span><span class="katex">...</span></span></option>
</select>
```
image


ylemkimon:
Thank you for the PR! I wonder there is other nodes that need to be excluded.

Issue #2042, opened 2019-07-16
Idea: top-level environment option


edemaine:
KaTeX currently essentially only supports `$...$` and `$$...$$` for entering math mode.  It doesn't support:
* `\begin{equation}...\end{equation}` (#445, which includes automatic equation numbering #350)
* `\begin{eqnarray}...\end{eqnarray}` (requested once in #1931)
* `\begin{align}...\end{align}` (similar effect to `$$\begin{aligned}...\end{aligned}$$`, as described in one of our [common issue](https://katex.org/docs/issues.html), but the effect is actually very different with equation numbering, so this will be more relevant with #350, and [it's even an issue with `\tag`](https://github.com/KaTeX/KaTeX/issues/1345#issuecomment-477917444))
* `\begin{split}...\end{split}` (#1345)

I propose a new `render` option with a name like `mathEnv` or `topEnv` or `env` or something, or else a redefinition of `displayMode` to allow values other than `false` (meaning `$...$`) and `true` (meaning `$$...$$`).  Either way, I'd like to allow values like `equation`, `eqnarray`, `eqnarray*`, `align`, `align*`, `split`, `split*`, as well as `math` (equivalent of inline `$...$`) and `displaymath` or `equation*` (equivalent of display `$$...$$`).  Then we could add a matching option to auto-render's `delimiters` (or replacing the existing `display` option) so that `\begin\s*{equation}...\end\s*{equation}` would end up calling `render` with the top-level environment set to `equation`.

This would finally let us support other top-level math environments in what I believe is a natural way.  Assuming you like the idea, I'd appreciate some suggestions on the name for the option, or whether to overload `displayMode`, or whether to deprecate `displayMode` (as it's effectively redundant with this option).

ronkok:
@edemaine Good stuff. I wholly agree with what you propose.

As to the details, I think that deprecating `displayMode` would be too drastic. It would break too many existing documents. And for that reason my inclination would be to overload the existing `displayMode` option.  When the system is fully built out, `displayMode` would be settable to:

`true` | `false` | `equation`  | `equation*` | `align` | `align*`  | `gather`  | `gather*`  | `flalign` | `flalign*`| `multline` | `multline*`| `alignat` | `alignat*` 

That list comes from section 3.1 of the AMS Users Guide. I did not include `split` because `split` occurs inside the `equation` environment.

`displayMode` would continue to be optional, with a default of `false`.

[...]

Issue #2121, opened 2019-10-11
Support TeX register (counter)


ylemkimon:
Required for equation numbering (#350, #571), etc.

kevinbarabash:
I'm curious about your plans for this.  It would be nice if it could be done in such a way that it'll work with React component wrappers which may render the same piece of math more than once.  We'd still want it to render the same equation/reference number.

ylemkimon:
@kevinbarabash I'm thinking of using macros (see #2139), which can be shared across function calls. How does React call KaTeX several times?

kevinbarabash:
It depends. If a component is designed well it won't call `KaTeX` unless the `math` that's being passed in changes.  If it does changes though, it will get re-rendered in which case we'd want the numbering to be the same.  The tricky part of this, is that only one of a number of instances may get re-rendered.  The re-render case is probably something that component maintainers will have to provide an API for.  Instead of a bunch of `<KaTeX>` instances being render and having `katex` figure out what the number should be maybe people using react will have to pass the equation number to `<KaTeX>`, e.g.
```
<KaTeX eqnNumber={1}>2x + 5 = 10</KaTeX>
<KaTeX eqnNumber={2}>2x + 5 - 5 = 10 - 5</KaTeX>
...
```
or something like that.

kevinbarabash:
Because katex is being used by various rendering engines (our auto-renderer, various React and Vue KaTeX components, etc.) maybe it makes sense to provide hooks that each of these renderers can use to decide what the numbers should be.  I don't think we can rely on the page being rendered once and done.  If we did, it would restrict the utility of auto-numbering considerably.

ylemkimon:
@kevinbarabash Another way I think we can use is [CSS counter](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Lists_and_Counters/Using_CSS_counters), as suggested by https://github.com/KaTeX/KaTeX/issues/350#issuecomment-160906578. It depends on the location on the DOM and updates automatically when it is mutated.

ylemkimon:
Manually specifying the equation number can be done using `\tag`. I think the problem is using in environments: #604.

Issue #1983, opened 2019-05-21
\tag is expected to be always flushed at right when equation is overflowed


AstroH-Peng:
> Before reporting a bug
- [x] Check [common issues](https://katex.org/docs/issues.html).
- [x] Check the bug is reproducible in [the demo](https://katex.org). If not, check KaTeX is up-to-date and installed correctly.
- [x] Search for [existing issues](https://github.com/KaTeX/KaTeX/issues).

**Describe the pool rendering:**
For a long equation set to `overflow-x: auto`, the right-end `\tag{}` will overlap and scroll with the equation.
You can use Eqs. 7.85 and 7.86 as a test case (with a narrow window), available [here](https://spaceresearch.top/blog/97ae8555/#7-2-relevance-vector-machines).
```css
// allow katex equations scrolling horizontally, without vertical scrollbar
.katex-display > .katex {
    height: auto;
    overflow: hidden;
    overflow-x: auto;
}
```

**Expected behavior:**
The `\tag` will be flush right no matter the equation is scrollable or not.
I'm not familiar with CSS. I tried to play around
```css
.katex-display>.katex>.katex-html>.tag {
    position: absolute;
    right: 0;
}
```
A feature of `position` mixturing `absolute` and `static` should be perfect. `sticky` doesn't fit in current setting, it is a mixturing of `relative` and `fixed` as I understand.

AstroH-Peng:
Sorry that I do not know how to remove the `bug` label. This is not a bug, just a small defect.

invkrh:
I have the same `\tag` problem here. Any idea on this?

kevinbarabash:
<img width="1181" alt="Screen Shot 2019-06-16 at 5 49 55 PM" src="https://user-images.githubusercontent.com/1044413/59569999-2e9f5e80-905f-11e9-92c5-4b41782f39d2.png">

I think we can solve the issue by introducing a new wrapper around everything before the `.tag` inside of `.katex-html`.  The tricky part though is keeping this new wrapper centered on the page while not allowing it to overlap the tag.

martinpflaum:
Hello
    .katex-display>.katex>.katex-html>.tag {
    position: relative;
    right: 0;
}
worked for me fine - even with scroll bars which is reall cool.

mbourne:
I notice tags still overlap equations as of 0.15.6 and there's no "float" to the next line if the screen is narrow.

@martinpflaum 's suggestion above still has the overlapping tags issue.

[...]

PR #2183, opened 2019-12-28
Support {gather} and {gather*}

ronkok:
This is the first in a series of pull requests aimed at support of AMS environments with automatic equation numbering. This PR supports `gather` and `gather*`. It establishes some methods for how to go about this sort of thing.

After (if?) this PR is approved and merged, subsequent PRs will address environments `equation`, `equation*`,  `align`, `align*`, `alignat`, and `alignat*`, `split`, `multline`, and `multline*`.

You can see both the HTML and the MathML if you use Firefox to view a [demo page](http://webmathcomparison.net/#g).

### Top Level Environment

In LaTeX, `{gather}` is itself a way to enter math mode. It will not render if called from math mode.  But KaTeX begins execution already in math mode. To reconcile the two, this PR adds a new rendering option, `topEnv`.  This enables an auto-render enhancement as suggested by @edemaine in #2042. The extension would pattern-match `\begin{gather}…\end{gather} ` and invoke KaTeX with the code and with rendering options `topEnv` and `displayMode` both set to true.

I expect that Markdown lexers will always use `$$…$$` and `$…$ ` to detect LaTeX. To support them, this PR also renders `{gather}` if invoked from display mode while `strict` is false or “warn”. Calling `{gather}` from inline mode always results in a KaTeX parse error.

The auto-render enhancement is not part of this PR. That’s for some other PR. And if anyone else wants to write it, feel free.

### CSS Counter

Automatic numbering is provided via CSS counters. There are two parallel counters, one for HTML and one for MathML.

CSS counters have several good features:

* They work in either client-side or server-side rendering.

* They can be augmented by an HTML page’s own CSS to display chapter numbers or section numbers along with the equation number.

* They evade all the difficulties that @kevinbarabash describes [here](https://github.com/KaTeX/KaTeX/issues/2121#issuecomment-548973467). 

CSS counters also have a couple of bad parts.

1. Multiple CSS counters need careful coordination. KaTeX CSS has a line that reads: `body { counter-reset: katexEqnNo mmlEqnNo; }`. Any subsequent counter-reset declaration will overwrite the KaTeX resets. So if an HTML page has its own CSS counter(s) they would have to write their counter-reset declaration to include both their own counter and the KaTeX counters, separated by spaces,.

2. A CSS counter cannot be the target of a link. But that can be fixed by a run-time extension. Say that a page is rendered sever-side. Here’s run time code that adds HTML id’s:

```js
const eqns = document.getElementsByClassName("eqn-num");
for (let i = 0; i < eqns.length; i++) {
    eqns[i].id = "eqn-num-" + String(i + 1);
}
```

Extensions could also be written to support  `\ref`, `\eqref`, and `\label`. Core KaTeX could write a span for each with its own class. Then a run-time extension could make two passes through the document. The first pass would add HTML ids to each counter, and record any `\label`s. The second pass would add an anchor element to each `\ref` that targets the appropriate id.

React fields and Vue fields would probably need a different sort of extension, if they support `\ref` at all. That extension is outside my expertise.

### To Do

Among the items left for future PR’s are:

- [ ] `equation`, `equation*`

- [ ] `split`

- [ ] `multline` and `multline*`

- [ ] `align`, `align*`, `alignat`, `alignat*`

- [ ] Upgrade of the auto-render extension

- [ ] `\ref`, `\eqref`, and `\label`

- [ ] `\notag`, `\shoveleft`, `\shoveright`

- [ ] Maybe someday `CD`


ronkok:
And yes, I know that flow has an objection. I have no idea what it is talking about.

ronkok:
Well, I fixed the typo but CircleCI is still running flow on the code from the previous commit and showing a flow error that should no longer exist. Odd.

kevinbarabash:
This is great.  Restricting text-only environments to display mode makes a lot of sense.  As for the flow issue, the tests seems to be passing so it must have resolved itself.

edemaine:
Wow, I just took my first look at the description of this PR, and it's really amazing! I didn't realize there was already a draft implementation of `topEnv` *and* equation numbering. Those are probably the most requested KaTeX features I've heard!  I'll try to review this one as well (though not today -- quickly running out of time).

ronkok:
You know, I don't think we should rush this one. In this PR, I did the minimum that could get one of the environments to work. My plan was to finish the others after the general strategy was approved. I think it would be a good idea to do all the auto-numbering environments in the same release. That way, I could be more certain to avoid breaking changes later when we do `equation` and `split`.

edemaine:
Yep, agreed. I'll still try to review it, but not before 0.12.0 release. 🙂

danielhollas:
Is the demo page referenced in the OP running the latest commit? Looks like positioning is off
obrazek

In any case, once this gets merged I'd be happy to help write/test the modification for autorenderer. :-)

ronkok:
> Looks like positioning is off

Could you be more specific?

danielhollas:
Sure. The generated labels are overlapping with the equations in the KaTeX column on the screenshot above.
(tested in latest Firefox and Chromium on Linux)

ronkok:
>  generated labels are overlapping with the equations in the KaTeX column

So they are, but this PR generates the KaTeX *MathML* column. The overlap you point out is rendered in the standard KaTeX HTML output. It's a valid comment, but it belongs in an issue to itself.

ronkok:
@danielhollas Strike my last comment. I had something else on my mind. What I should have said is that the overlap occurs in both the proposed `{gather}`  environment and the pre-existing `tag` function. I suspect that the overlap can be eliminated with some better CSS, but I wanted to fix one problem at a time. This PR already has enough moving parts without trying to fix pre-existing problems on the fly.

ronkok:
Note to reviewer: This is to explain why I (1) changed the treatment of `leqno` and (2) changed the MathML nested map functions into a for-loop. The pre-existing treatment of `leqno` happens at a high level in the KaTeX code.

In order to automatically place a tag node in the proper position, I needed access to `leqno`  within the `htmlBuilder` and  `mathmlBuilder` functions of `array.js`.  So I had to first change the treatment of `leqno` and attach it to the group that is passed to `htmlBuilder` and  `mathmlBuilder`. I unwound the MathML nested map functions for the same reason. I needed access to `leqno` inside a deeply nested callback and I could think of no other way to get it.

ylemkimon:
Sorry for the delayed review. I was thinking about alternative approaches to CSS counter, but I couldn't think of one and months have passed :cry: I'll try to review this as soon as possible.

src/mathMLTree.js
```diff
@@ -104,6 +115,10 @@ export class MathNode implements MathDomNode {
             }
         }
 
+        if (this.classes.length > 0) {
+            markup += ` class ="${createClass(this.classes)}"`;
```
ylemkimon, line 119:
```suggestion
            markup += ` class ="${utils.escape(createClass(this.classes))}"`;
```

src/environments/array.js
```diff
@@ -817,17 +868,31 @@ defineEnvironment({
...
     props: {
         numArgs: 0,
     },
     handler(context) {
+        if (utils.contains(["gather", "gather*"], context.envName)) {
+            const settings = context.parser.settings;
+            if (!settings.displayMode) {
+                throw new ParseError(`{${context.envName}} cannot be used inline.`);
+            } else if (settings.strict === "warn" && !settings.topEnv) {
+                // eslint-disable-next-line no-console
```
ylemkimon, line 881:
```suggestion
```

src/environments/array.js
```diff
@@ -817,17 +868,31 @@ defineEnvironment({
...
+                console.log(`{${context.envName}} called from math mode.`);
```
ylemkimon, line 882:
```suggestion
                settings.reportNonstrict("textEnv",
                    `{${context.envName}} called from math mode.`);
```

src/environments/array.js
```diff
@@ -817,17 +868,31 @@ defineEnvironment({
...
+    names: ["gathered", "gather", "gather*"],
     props: {
         numArgs: 0,
     },
     handler(context) {
+        if (utils.contains(["gather", "gather*"], context.envName)) {
+            const settings = context.parser.settings;
+            if (!settings.displayMode) {
+                throw new ParseError(`{${context.envName}} cannot be used inline.`);
+            } else if (settings.strict === "warn" && !settings.topEnv) {
```
ylemkimon, line 880:
```suggestion
            } else if (settings.strict && !settings.topEnv) {
```

src/katex.less
```diff
@@ -635,3 +635,23 @@
...
+}
+
+.mml-eqn-num::before {
+    counter-increment: mmlEqnNo;
+    content: "(" counter(mmlEqnNo) ")";
+}
+
+.mtr-glue {
+    width: 50%;
+}
```
ylemkimon, line 657: I think these should be a `contrib` extension for now.

ronkok: I agree that we should be careful with CSS counters. First, just to make sure that you and I understand each other, let me describe what I think is the potential problem. It lies in the CSS code `body { counter-reset: katexEqnNo mmlEqnNo; }`. The trouble is that a counter-reset statement can over-write all counters. This line of CSS would either over-write a page's existing counters or be over-written, depending on the order in which CSS files are loaded.

Now let me suggest another approach.

1. Remove from this PR the CSS line `body { counter-reset: katexEqnNo mmlEqnNo; }`.
2. Let the other CSS in this PR stand. It does no harm.
3. Amend the setup documentation to strongly suggest that web site administrators should include the `body { counter-reset: katexEqnNo mmlEqnNo; }` CSS in their page.
4. For convenience, create a `contrib` CSS file that has only one line, that being `body { counter-reset: katexEqnNo mmlEqnNo; }`.
5. Test this PR without a counter-reset and make sure that the worst outcome is the omission of automatic numbering and perhaps some console messages.

Am I addressing the same problem that you see? What do you think of my suggestion?

ylemkimon: I was thinking moving all styles related to numbering out to a `contrib` module, but now I think it'd make KaTeX hard to configure. We could rather add an option to choose which numbering to use, e.g., CSS counter, Javascript, none, etc. (I'll send a PR later) I think `body { counter-reset: katexEqnNo mmlEqnNo; }` wouldn't cause too much trouble, so I think it's okay to leave it here.

Other styles, `.eqn-num::before`, `.mml-eqn-num::before`, and `.mtr-glue`, should be scoped, i.e., inside `.katex { ... }`.

ylemkimon:
@danielhollas The overlap issue is tracked at #1983.

ronkok:
In the most recent commit, I came across an odd behavior. Eslint now objects to any use of `console.error`, regardless of what sort of `eslint-disable` comment I apply. I even tried  `/*eslint-disable */` and `/*eslint-enable */`. In order to get a clean test run, I had to change `console.error` to `console.log`. This occurs in the `macro.js` terminal command `\errmessage`.

src/mathMLTree.js
ronkok: Done.

src/environments/array.js

src/environments/array.js

src/katex.less
ronkok: Agree 100% on scoping the styles inside `.katex{...}`. On it.

ronkok: Styles are scoped inside `.katex{...}`.

If we let the CSS counter-reset remain as is, then I think the KaTeX setup documentation should be amended. That can perhaps be a later PR. I would like all the automatic numbering environments to go into the same KaTeX release.

ylemkimon:
> In the error-handling commit, I came across an odd behavior. Eslint now objects to any use of `console.error`, regardless of what sort of `eslint-disable` comment I apply. I even tried  `/*eslint-disable */` and `/*eslint-enable */`. In order to get a clean test run, I had to change `console.error` to `console.log`. This occurs in the `macro.js` terminal command `\errmessage`.

It seems to work fine for me and [CircleCI](https://app.circleci.com/pipelines/github/KaTeX/KaTeX/2043/workflows/62715a0a-b4f3-4614-a07b-d764c083f29e/jobs/9721). Maybe wrong version of eslint got installed?

ronkok:
> It seems to work fine for me and CircleCI. 

Okay, since it seems to be a local problem, I'll revert the `console.error` and continue my investigation locally.

ronkok:
Yay! Thank you for the review. This one is pretty opinionated. There are decisions made and I'm glad you took the time to review them carefully.

edemaine:
One question (maybe) before this merges: have you done a pros/cons analysis of using CSS counters vs. using the existing LaTeX macro system (extended to do LaTeX counters)?

ronkok:
> pros/cons analysis of using CSS counters vs. using the existing LaTeX macro system (extended to do LaTeX counters)?

I did consider using some sort of Javascript counter. It would only have to carry an integer, so it could have been a lot simpler than the existing macro system. As you point out, this could be coordinated with the auto-render extension to provide a working system.

The problem that I could not overcome was the existence of fields in React, Vue, etc., that only update locally. A CSS counter handles this without drama. A Javascript counter would need a lot of extra infrastructure to update after those individual fields update.

I'm not married to CSS counters. If someone can show a reasonable system to handle React and Vue, I would be willing to change course.

ylemkimon:
I cannot think of a single method that accommodates every use cases. I think, the best we could do is provide multiple implementations and thoroughly document it.

edemaine:
Yes, good points. CSS counters do seem to do pretty much what they want. (If only they supported automatic `\label`/`\ref`! But sadly their values can't be saved in CSS variables.)

One advantage of implementing this via LaTeX counters is that would get support for LaTeX commands like the following:
```
\setcounter{equation}{10} % manually jump to equation number 10
\addtocounter{equation}{5}
```
(Assuming we actually implemented #2121, anyway.)

I like @ylemkimon's idea of offering multiple strategies that can be configured in an option. That means we can merge this now, and consider an alternative in a future PR.

edemaine:
By the way, discussion in #2121, in particular https://github.com/KaTeX/KaTeX/issues/2121#issuecomment-548973467 by @kevinbarabash highlights the difficulty with React. I see the problem now -- not easy. Indeed, @ylemkimon suggests using counters in that thread.

ronkok:
It seems to me that TeX counters and `\ref` will have to be in extensions, separate from core KaTeX.  If this PR is merged, then the auto-render extension could create HTML ids with code as simple as this:

```
const eqns = document.getElementsByClassName("eqn-num");
for (let i = 0; i < eqns.length; i++) {
    eqns[i].id = "eqn-num-" + String(i + 1);
}
```

Then a second pass through the document would update `\ref`s.

A TeX counter in React or in [Observable](https://observablehq.com/) would need a different sort of extension, and I don't know how it would work.

PR #2186, opened 2020-01-03
Fix delimiter error message

### LINKED ISSUE #2184: \right{...} causes circular structure in error reporting
**Describe the bug:**
Testing on `katex.org`, here is a minimal bad example based on #2174:
```latex
\newcommand{\myright}[1]{\right{#1}}
\cos\left(x\myright)
```
Error:
> Converting circular structure to JSON --> starting at object with constructor 't' | property 'settings' -> object with constructor 't' | property 'macros' -> object with constructor 'Object' | ... | property 'loc' -> object with constructor 't' --- property 'lexer' closes the circle

**Expected behavior:**
Unless I'm missing something, `\myright` should behave like `\right`.

**Additional context:**
The following code works fine, so it has to do with macro arguments.
```latex
\newcommand{\myright}{\right}
\cos\left(x\myright)
```
Perhaps #2085 fixes this bug, but I haven't checked. If so, we should add a test.

edemaine:
Avoid "circular structure" error message by avoiding `JSON.stringify` on a parse structure when getting an invalid delimiter type (e.g. ordgroup). Instead, report type of delimiter when it's invalid.  Fixes #2184.

### DIFF
--- a/src/functions/delimsizing.js
+++ b/src/functions/delimsizing.js
@@ -62,11 +62,12 @@ function checkDelimiter(
     const symDelim = checkSymbolNodeType(delim);
     if (symDelim && utils.contains(delimiters, symDelim.text)) {
         return symDelim;
-    } else {
+    } else if (symDelim) {
         throw new ParseError(
-            "Invalid delimiter: '" +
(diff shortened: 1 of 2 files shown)


kevinbarabash:
Nice!

PR #2193, opened 2020-01-31
Change katex import to relative path

cjh9:
In order to use mhchem on the frontend with native modules we need the exact relative path.


ronkok:
@cjh9 Many sites serve their own KaTeX rather than using the KaTeX CDN. Such sites may use a different relative path. So hard-coding an exact path into `mhchem.js` presents problems.

I'm open to suggestions, but no good alternatives occur to me right now.

ylemkimon:
Refer to https://github.com/KaTeX/KaTeX/tree/master/contrib/mhchem on how to include mhchem extension.

Issue #2200, opened 2020-02-14
Please add the Plimsoll symbol (circle with a horizontal line through it)


LloydLS:
Hi!

Please consider adding the Plimsoll symbol (https://en.wikipedia.org/wiki/Plimsoll), as provided by the chemstyle package (called `\standardstate` in that package). `\barcirc` would also work instead.

`\ominus` is similar, but not exactly the same (the horizontal line extends beyond the circle in the Plimsoll symbol) 

Chemstyle package documentation (page 4 has the relevant section):
http://mirrors.ibiblio.org/CTAN/macros/latex/contrib/chemstyle/chemstyle.pdf

Many thanks!

ronkok:
@LloydLS  Please try this macro code on the KaTeX demo page. Let me know if it fits your needs. If it does, I'll write up a PR. Also, which name is preferable: `\standardstate` or `\barcirc`?

```
\gdef\barcirc{{\mathrlap{\mathchoice{\kern{0.145em}}{\kern{0.145em}}{\kern{0.1015em}}{\kern{0.1015em}}\circ}{-}}}

\barcirc

G^\barcirc \\
\int_{T^\barcirc}^{T_\text{out}}

```

LloydLS:
Hi @ronkok 

That looks fantastic! 

I'd probably suggest the name `\barcirc`, just because sometimes a superscript circle (without a bar) is used to denote a standard state.

Thanks so much for this!

edemaine:
The [comprehensive symbol list](http://tug.ctan.org/info/symbols/comprehensive/symbols-a4.pdf#page=216) mentions `\barcirc` from a `comp.text.tex` post, but no packages apparently call it that. (It also gives a macro definition for it, but it uses `\hfil` so we can't use it directly; @ronkok's definition seems to compose the same symbols, `\circ` and minus.) It does mention that `stmaryrd.sty` as well as `boisik.sty` define `\minuso` with a similar symbol, so `\minuso` is another reasonable choice for the name -- perhaps the best one?

LloydLS:
Hi @edemaine 

I wasn't sure whether you were replying to me, so apologies for the delay. 

I'm not sure what the guidelines are for naming new symbols in KaTeX; presumably we want to be somewhat consistent with LaTeX? 

If we could name it anything we want, then maybe `\plimsoll` would be appropriate (hopefully wouldn't conflict with any nautical LaTeX users...)

Ultimately, anything you like to call it is fine by me.

[...]

PR #2213, opened 2020-03-08
Support \minuso

ronkok:
Support `\minuso` from the `stmaryrd` package. Addresses #2200.


kevinbarabash:
I tried it out using different styles and it looks great!
Screen Shot 2020-03-08 at 9 11 59 PM

src/macros.js
```diff
@@ -845,6 +845,12 @@ defineMacro("\\rBrace", "\\html@mathml{" +
 defineMacro("\u2983", "\\lBrace"); // blackboard bold {
 defineMacro("\u2984", "\\rBrace"); // blackboard bold }
 
+// The stmaryrd function `\minuso` provides a "Plimsoll" symbol that
+// superimposes the characters \circ and mathminus. Used in chemistry.
+// TODO: A MathML `\minuso` that is not tightly bound to the KaTeX font.
+defineMacro("\\minuso", `{\\mathrlap{\\mathchoice{\\kern{0.145em}}{\\kern{0.145em}}
+{\\kern{0.1015em}}{\\kern{0.0725em}}\\circ}{-}}`);
```
kevinbarabash: We could use the `\\html@mathml` command to define a unicode character, `\u29B5` in this case, for the MathML rendering.

ronkok: Good catch. I did not know that character existed.

kevinbarabash:
@ronkok thanks updating the macro definition to support MathML.  I tried the following on quicklatex.com and compare it to the output on the netlify preview and the spacing is off.

**quicklatex**
Screen Shot 2020-03-09 at 7 28 52 PM

**katex**
Screen Shot 2020-03-09 at 7 28 48 PM

I was testing using:
```
1 \minuso 2 \\
1 + 2
```

ronkok:
My bad. I just checked the `stmaryrd` code and it gets `bin` spacing. Change coming up.

src/macros.js
```diff
@@ -845,6 +845,12 @@ defineMacro("\\rBrace", "\\html@mathml{" +
 defineMacro("\u2983", "\\lBrace"); // blackboard bold {
 defineMacro("\u2984", "\\rBrace"); // blackboard bold }
 
+// The stmaryrd function `\minuso` provides a "Plimsoll" symbol that
+// superimposes the characters \circ and mathminus. Used in chemistry.
+defineMacro("\\minuso", `\\mathbin{\\html@mathml{{\\mathrlap{\\mathchoice{
+    \\kern{0.145em}}{\\kern{0.145em}}{\\kern{0.1015em}}{\\kern{0.0725em}}
+    \\circ}{-}}}{\u29B5}}`);
```
edemaine, line 852: Instead of `\u29B5`, we should probably write `\char`⦵` to be consistent with other macros defined in this file.

Also, you're currently not mapping Unicode input of ⦵ to this character. Perhaps you want to add `defineMacro("⦵", "\\minuso")`?  (This will only work with the added `\char` escape above.)

ronkok: I'll make the \char⦵ change.

The mapping of the Unicode code point takes place in `symbols.js`, where I have set the `acceptUnicodeChar` to `true`.

ronkok: Oops, strike that last comment. I had this confused with a different PR. 

Yes, I should map that Unicode value. On it.

edemaine: There are now two copies of `defineMacro("\\minuso"`.  Could you delete the old one?

ronkok: ...slaps forehead...

Done.

docs/supported.md
```diff
@@ -632,9 +632,9 @@ In cases where KaTeX fonts do not have a bold glyph, `\pmb` can simulate one. Fo
 |$\circledR$ `\circledR`|$\diamondsuit$ `\diamondsuit`|$\heartsuit$ `\heartsuit`
 |$\text{\textregistered}$ `\text{\textregistered}`|$\diamonds$ `\diamonds`|$\hearts$ `\hearts`
 |$\circledS$ `\circledS`|$\spadesuit$ `\spadesuit`|$\spades$ `\spades`
-|$\text{\textcircled a}$ `\text{\textcircled a}`|$\maltese$ `\maltese`||
+|$\text{\textcircled a}$ `\text{\textcircled a}`|$\maltese$ `\maltese`|$\minuso$ `\minuso`|
 
-Direct Input: $£ ¥ ∇ ∞ · ∠ ∡ ∢ ♠ ♡ ♢ ♣ ♭ ♮ ♯ ✓ …  ⋮  ⋯  ⋱  !$ ‼
+Direct Input: $£ ¥ ∇ ∞ · ∠ ∡ ∢ ♠ ♡ ♢ ♣ ♭ ♮ ♯ ✓ …  ⋮  ⋯  ⋱  !$ ‼ ⦵
```
edemaine, line 637: This one *can* go inside `$`s, right?

ronkok: Not if we want to support copy-and-paste from the page. KaTeX rendering will return an "o" superimposed with a "-". That doesn't provide what a copying user needs.

Issue #2198, opened 2020-02-08
3x3 matrices render with jagged delimiters when used with pmatrix


ganeshbelgur:
While rendering a 3x3 matrix, a part of the delimiters appear to be protruding. This nit picky detail is an issue when the matrix is viewed on a huge screen in front of a huge audience.
![3x3](https://user-images.githubusercontent.com/10693043/74090358-6cf6fb80-4ad0-11ea-8e5d-ddb8073e1630.png)

**LaTeX code:**
```latex
\begin{pmatrix}
    a & b & c\\
    a & b & c\\
    a & b & c\\
\end{pmatrix}
```

**Expected behavior:**
Did a quick render of the same Latex code using [QuickLaTeX](https://www.quicklatex.com/) to generate this image:
![3x3_right](https://user-images.githubusercontent.com/10693043/74090540-7719f980-4ad2-11ea-8702-9a21c5739ddf.png)

**Environment:**
 - KaTeX Version: Version used in https://katex.org/ website's "Type an expression" editor
 - Device: Desktop
 - OS: Ubuntu 18.04
 - Browser: Google Chrome 80.0.3987.87

**Additional context:**
Originally, I noticed this issue in Joplin, a markdown text editor supporting Katex in it. To verify if the issue was in Joplin or the Katex engine, I used the "Type an expression" feature at https://katex.org/ to render a test image and found the issue.

ronkok:
@ganeshbelgur I see what you mean. This issue is probably made worse by PR #1986, which I wrote to fix issue #1975.

I can't just remove the overlaps created in PR #1986. If anything, they may have to be increased because I have still seen a gap now and then.

So, to make this issues's jaggedness go away, we would have to stop building up tall parens out of font-glyph components and instead build them up out of SVGs, similar to what we do now with horizontal stretchy elements. I could write the SVG paths in way that ensures that the overlap always occurs at a straight section of the paren. Then the overlap would never be seen.

@kevinbarabash Such a change would add new SVG paths and add some finicky branching code to `delimiter.js`, similar to what we do now for radicals. `delimiter.js` is going to end up pretty hard to read, because it will contain so many edge cases. What do you think? Is it worth the effort?

kevinbarabash:
@ronkok could you say a bit more about the edge cases you're thinking about?  I think one of the things that makes `sqrtSvg` complicated is that we use SVG for all sizes of roots, but in this situation I think we could get away with only using SVG for the largest size of delimiter.

[...]

PR #2215, opened 2020-03-10
Fix jaggedness in tall parentheses

ronkok:
This PR fixes the jaggedness noted in #2198. It does so by using SVGs for parentheses. All other tall delimiters are unchanged.


adekau:
Being an SVG, will the parenthesis still be a highlightable character or when you right click it will you see "Open Image in New Tab", etc?

ronkok:
@adekau You bring up a very good point. An SVG parenthesis would indeed be a character that you could not highlight. I did not think of that.

 I'm going to look further into this problem and see if I can fix this parenthesis in some other way.

kevinbarabash:
It's sad that they removed support for defining custom fonts in SVG.  It really could've helped with this use case.

PR #2220, opened 2020-03-22
Support MathML display mode

### LINKED ISSUE #2219: Display mode doesn't work for MathML output
**Describe the bug:**
Display mode does not work for MathML output (it does for HTML output).

**To Reproduce:**
Render any expression, using the options `{displayMode: true, output: 'mathml'}`.
See [JSFiddle](https://jsfiddle.net/2b497Ls0/).

**Expected behavior:**
Expression renders in display mode, i.e.&nbsp;display-style math centred on its own line.

**Actual behavior:**
Expression appears to render in inline mode.

**Screenshot:**

**Environment:**
 - KaTeX Version: 0.11.1
 - Browser: any browser which supports MathML

ronkok:
Fix #2219.

### DIFF
--- a/src/buildMathML.js
+++ b/src/buildMathML.js
@@ -238,6 +238,7 @@ export default function buildMathML(
     tree: AnyParseNode[],
     texExpression: string,
     options: Options,
+    isDisplayMode: boolean,
     forMathmlOnly: boolean,
 ): DomSpan {
     const expression = buildExpression(tree, options);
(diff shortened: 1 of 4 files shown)


ronkok:
Below is visual evidence (from Firefox) that this PR fixes the problem. This comes from the code:
```
<p>\(\sum_{i=0}^{100}x^2\)</p>
<p>\[\sum_{i=0}^{100}x^2\]</p>
```

DisplayMML

There are also two tests in the MathML snapshots.

ronkok:
@ylemkimon Thank you for the review!

phpdev1995:
is katex convert `MathML to mathematics formulas` ? like mathjax

kevinbarabash:
@ronkok thanks for adding snapshots!  The display rendering of the sigma feels a little over the top.  Do you know if browsers that support MathML will render it using a different font if the font has a `MATH` table in it?

ronkok:
I believe that MathML-capable browsers will allow CSS to set a font for the `<math>`. If the webpage CSS makes no declaration, then the browser picks a system font. The screenshot in my comment above used the Cambria Math font installed on my machine.

PR #1986, opened 2019-05-27
Prevent gaps in tall delimiters

ronkok:
Fixes #1975 by typesetting tall delimiters with a small (0.005em) overlap where glyphs meet.

As evidence, I offer two screenshots. The first one is before this PR:
gap

The second screenshot is after this PR:
nogap

The value of 0.005em is entirely a guess on my part and may need later adjustment.

Issue #1975, opened 2019-05-12
Break in tall delimiters


OisinMoran:
- [x] Check [common issues](https://katex.org/docs/issues.html).
- [x] Check the bug is reproducible in [the demo](https://katex.org). If not, check KaTeX is up-to-date and installed correctly.
- [x] Search for [existing issues](https://github.com/KaTeX/KaTeX/issues).

**Describe the bug:**
Left and right floor brackets, when containing a large expression, appear to to break in the middle. This disappears on zooming but makes for suboptimal rendering.

**(La)TeX code:**
```latex
5 = \left\lfloor{\sqrt{\sqrt{\sqrt{\sqrt{\sqrt{(4!)!}}}}}}\right\rfloor
```

**Expected behavior:**
Expected vertical bar portion of left and right floor braces to be continuous.

**Screenshots:**
Taken from [the demo](https://katex.org).
![image](https://user-images.githubusercontent.com/8939851/57576342-3ca50280-7455-11e9-8b66-26c231afeb04.png)

**Environment (please complete the following information):**
 - KaTeX Version: 0.10.1
 - Device: Desktop
 - OS: Ubuntu 18.10
 - Browser: Chrome
 - Version: 74.0.3729.131 (Official Build) (64-bit)

kevinbarabash:
We could probably use SVG to avoid this situation.  @OisinMoran are you seeing similar issues with tall parens and tall square brackets?

OisinMoran:
Yes, with
```latex
5 \ne \left[{\sqrt{\sqrt{\sqrt{\sqrt{\sqrt{(4!)!}}}}}}\right]
```
I get:
![image](https://user-images.githubusercontent.com/8939851/57618311-f34ed300-757a-11e9-887d-826cff257438.png)

And with
```latex
5 \ne \left({\sqrt{\sqrt{\sqrt{\sqrt{\sqrt{(4!)!}}}}}}\right)
```
I get:
![image](https://user-images.githubusercontent.com/8939851/57618362-11b4ce80-757b-11e9-864b-97b2c428e891.png)

kevinbarabash:
Thanks for confirming my suspicions.

ronkok:
I think this can be fixed without SVG.  We just need to adjust the vertical alignment calculation in `makeStackedDelim()`. That calculation should include a small overlap where glyphs meet. Otherwise floating point rounding error can result in the gap we see in @OisinMoran's screenshots. 

I'll take a look sometime in the next week.

PR #2234, opened 2020-04-05
Fix jagged parentheses

### LINKED ISSUE #2198: 3x3 matrices render with jagged delimiters when used with pmatrix

ronkok:
Fixes #2198. 

As you may recall, an earlier issue (#1975) complained of occasional gaps in built-up vertical delimiters. PR #1986 introduced small overlaps at each junction, then added one extra repeat element to cover the last gap. For parentheses around a 3x3 matrix, that extra repeat element runs past the curve in the parenthesis and looks jagged.

This PR uses a shorter repeat element to cover that last gap.


src/delimiter.js
```diff
@@ -375,9 +375,19 @@ const makeStackedDelim = function(
 
     // To cover the gap create by the overlaps, insert one more repeat element,
     // at a position that juts 0.005 above the bottom of the top element.
-    inners.push({type: "kern", size: shiftOfExtraElement});
-    inners.push(makeInner(repeat, font, mode));
-    inners.push(lap);
+    if ((repeat === "\u239c" || repeat === "\u239f") && repeatCount === 0) {
```
edemaine, line 378: Why are parentheses special here? It seems like it would be a good idea to use a small connector whenever `repeatCount === 0`.

Perhaps there are some symbols where a vertical line segment isn't the right choice, but I feel like this might still be a better choice than the `else` case... Or at least perhaps we should do a little analysis of all the delimiters to decide?

src/delimiter.js
```diff
@@ -375,9 +375,19 @@ const makeStackedDelim = function(
...
+        // Parentheses need a short repeat element in order to avoid an overrun.
+        // We'll make a 0.3em tall element from a SVG.
+        inners.push({type: "kern", size: -0.15});
```
edemaine, line 381: Why 0.3em? It seems like the existing code uses 0.005em for the overlap shift, so I would have imagined using an 0.01em tall SVG with negative vertical shifts of 0.005em on either side.

Also, it would seem nicer to use the 0.3 (or whatever) constant in `svgdata.leftParenInner[2]` instead of repeating it here.

src/delimiter.js
ronkok: I don't get to pick any arbitrary small connector. If I don't use an SVG, then I am limited to font glyphs. And the selection of font glyphs is not ideal. The short ones were too thick, or had the wrong xMin distance or the wrong advance distance. After trying some alternative glyphs that made sense, none of them worked. I gave up and went the SVG route. It's a sure thing.

src/delimiter.js
ronkok: > Why 0.3em?

It's been awhile and I'm not 100% positive I remember this correctly, but I think 0.03em has a pretty big tolerance for adjustment. Since the SVG is used only in this one case where parens barely touch, it could be significantly bigger or smaller and still work.

I opted for the number at the bigger end of the range partly because I think I got the original 0.005em number wrong. I've seen some cases on my own screen where there is still a gap. I did not want to go too small a second time.

src/delimiter.js
edemaine: Yes, that makes sense. I was wondering why we're fixing the case of parentheses but not fixing the case of e.g. curly braces (which use `\u23aa` for their `repeat` character). Would bad things happen if we replaced this `if` statement with
```js
 if (repeatCount === 0) {
```
?

src/delimiter.js
edemaine: OK, and sounds safe enough in this case, but then maybe we should adjust the 0.005em number? (Can be in a separate PR effort.)

Do you agree with changing the `0.3` constant with `svgdata.leftParenInner[2]`?

ronkok: > Also, it would seem nicer to use the 0.3 (or whatever) constant in svgdata.leftParenInner[2] instead of repeating it here.

Well, I certainly would do something of that nature if the SVG had multiple use cases. We have some other paths in `svgGeometry.js` that can grow or shrink, and we could have done that. But since this does get used for only one case, I thought that the constant number `0.3em` was easy to read.

I don't feel strongly about this one.

src/delimiter.js
ronkok: Curly braces have a longer straight section built into the glyph. I checked, and they do not need a correction similar to this PR. The existing short glyph remains hidden within the curly brace glyph.

src/delimiter.js
ronkok: Oops, I misunderstood your comment. You were tying this constant to a piece of data from `src/buildCommon.js`, not `svgGeometry.`. Sure. I have no problem with that. I'll make a change.

ronkok: You've got quicker fingers that I do. You have approved this PR. Do you still want me to change the 0.3 constant?

edemaine: Sure, now that we agree, please go ahead. I'll merge after checking the commit.

src/delimiter.js
```diff
@@ -378,11 +378,12 @@ const makeStackedDelim = function(
     if ((repeat === "\u239c" || repeat === "\u239f") && repeatCount === 0) {
         // Parentheses need a short repeat element in order to avoid an overrun.
         // We'll make a 0.3em tall element from a SVG.
-        inners.push({type: "kern", size: -0.15});
+        const overlap = buildCommon.svgData.leftParenInner[2] / 2
```
edemaine: This needs a semicolon (for lint).

ronkok:
Thank you for the review!

PR #2239, opened 2020-04-13
Submodule fix for "build from source" instructions

brentyi:
Need submodules to build from source :slightly_smiling_face:


brentyi:
[clabot:check]

Issue #2265, opened 2020-06-04
New `\html*` commands


junghoocho:
PR #2082 adds new `\html*` commands. They are great! Now I can do things that I could never do before. 

But I also have one issue regarding these new commands: The current implementation adds one enclosing span per every `\html*` command. For example, if I have a command like `\htmlData{a=1}{\htmlClass{b}{x}}` the generated HTML is roughly `<span data-a="1"><span class="b">x</span></span>`. But the program that I use requires that both the class and data attributes appear in the same HTML element. 

So here is my question: When a sequence of `\html*` commands are directly nested like the above example, will it be possible to generate a single enclosing span like `<span data-a="1" class="b">x</span>`, instead of creating one `span` per command? 

If this can be done, it will be super awesome. Thank you.

edemaine:
@ylemkimon [wrote](https://github.com/KaTeX/KaTeX/pull/2266#issuecomment-657186601):

> I'm against adding `\htmlAttr` as (1) validation of attributes is impossible and (2) it may open to vulnerabilities and blacklisting is not enough.

An alternative would be to define some coalescing behavior for `<span class="enclosing">`, so that `\htmlId{hi}{\htmlData{x=1}{hello}}` produces `<span id="hi" data-x="1" class="enclosing">...hello...</span>`.

junghoocho:
`\htmlAttr` is in the same security domain as other `\html*` commands, so it is disabled by default. The user has to explicitly enable it in order to use it, so it is unlikely that someone is tripped up by the vulnerability by mistake. But I do see the concern, and I understand if the consensus judgment is not to include it. I just wanted to contribute back what I coded for my own need in case you find it will be useful.

Regarding coalescing, coalescing will address my need as does `\htmlAttr`. `\htmlAttr` was just so much easier for me to code up quickly since I am not familiar with the overall KaTeX codebase. I will be excited if someone who knows the codebase well can implement the coalescing idea if anyone has the time and interest. Thank you very much!

PR #2266, opened 2020-06-04
feat: add \htmlAttr command

junghoocho:
This pull request is related to issue #2265. In #2265, I suggested adding `\htmlAttr`, so that we can add any HTML attributes into a single enclosing `<span>` tag. This pull request includes the `\htmlAttr` that I suggested in #2265. My implementation is a very minor modification of the existing `\htmlData` command.


junghoocho:
[clabot:check]

junghoocho:
If this command is considered too dangerous because it allows an arbitrary attribute to be inserted, we can add an explicit check of the added attribute names, so that only id, class, style, and data- attributes are added. What I hope to get is to have a mechanism so that all needed attributes are included in a single enclosing span tag.

junghoocho:
I looked at the MDN doc on the span element to see on what attributes an attacker may potentially insert JavaScript code. From my review, it seems clear that event-handler attributes with names like `onX` are the only places with potential JavaScript code. In my third commit 4bc5626d9508b92d4ba66c63f6dd6daa496d9d0f, I added an explicit check to disallow setting any attributes starting with on via `\htmlAttr`. Of course, since `\htmlAttr` has its own TrustContext, people can configure exactly what attributes are allowed through the command, but I suspect we don't want to allow inserting JavaScript through KaTeX command under any scenario. I hope this will make the addition of \htmlAttr safe enough to be merged to the mainline.

ylemkimon:
I'm against adding `\htmlAttr` as (1) validation of attributes is impossible and (2) it may open to vulnerabilities and blacklisting is not enough.

junghoocho:
To make `\htmlAttr` logically equivalent to the union of existing `\htmlId``\htmlClass` `\htmlStyle`, and `\htmlData` commands,  I can add the following validation logic to `\htmlAttr`:

(1) limit allowed attribute names to `id`, `class`, `style`,`data-*`
(2) for `id`, `class`, `style`, `data-*`, let the attribute values pass the existing validation logics for `htmlid`, `htmlclass` and `htmlstyle`, `htmldata` values, respectively.

If you think this will be an acceptable implementation, I will be happy to add the above logic to the pull request for review.

ylemkimon:
@junghoocho That seems to be OK. If you don't mind, could you update the PR? I'll leave #2265 open for discussions on allowing other attributes.

junghoocho:
@ylemkimon Thank you very much. I just updated the pull request to allow only id, style, class, and data- attributes in `\htmlAttr`

Regarding attribute-value validation, I find that the existing code does not perform any validation on its own and relies on the user-provided `trust` function. Given this, the updated pull request simply calls the user-provided trust function for validation as well. 

Let me know if it looks OK or if I missed anything. Thank you.

src/functions/html.js
```diff
@@ -66,6 +66,29 @@ defineFunction({
...
+                    if (keyVal.length !== 2) {
+                        throw new ParseError(
+                            "Error parsing key-value for \\htmlAttr");
+                    }
+                    // allow "id", "class", "style" and "data-*" attributes only
+                    const key = keyVal[0].trim().toLowerCase();
+                    if (key !== "id" && key !== "class" && key !== "style" &&
+                        !key.startsWith("data-")) {
+                        return parser.formatUnsupportedCmd(funcName);
+                    }
```
eric-wieser, line 82: Can't we let `isTrusted` handle this?

eric-wieser (commented):
LGTM, thanks! This would close #3733

PR #2279, opened 2020-06-21
Update API language on auto-render script

xymostech:
I noticed that the docs said there were "two auto-render-specific keys" and then proceeded to list 5 extra keys that could be passed. :) I updated the language in all of the versions of docs because they were all incorrect.

I changed them all to say "several" so that this doesn't accidentally get out of date again when another key is added in the future, but I'm happy to change it to say "five"/"four" in the different versions if that seems better.

Also 👋

### DIFF
--- a/website/versioned_docs/version-0.10.0/autorender.md
+++ b/website/versioned_docs/version-0.10.0/autorender.md
@@ -57,7 +57,7 @@ nodes inside this element and render the math in them.
 
 `options` is an optional object argument that can have the same keys as [the
 object passed to `katex.render`](https://github.com/KaTeX/KaTeX/#rendering-options),
-in addition to two auto-render-specific keys:
+in addition to several auto-render-specific keys:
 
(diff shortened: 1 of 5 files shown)


csilvers:
Sadly I don't have write access, but I can at least respond to your commit message:

> Also wave

We miss you!

edemaine:
It looks like the CircleCI tests aren't running, perhaps because only Markdown files changed, but Github won't let the PR merge without them (it says "Expected — Waiting for status to be reported"). @ylemkimon any ideas?

edemaine:
> It looks like the CircleCI tests aren't running, perhaps because only Markdown files changed, but Github won't let the PR merge without them (it says "Expected — Waiting for status to be reported").

Seems to have just been transient -- merging master in caused the tests to actually run. Yay!

ronkok:
@xymostech I just want to add my appreciation for the work you did during the creation of KaTeX. It's an awesome library.

kevinbarabash:
@xymostech 👋 🙂

Issue #1766, opened 2018-10-30
\fcolorbox with math inside


basilgass:
Hi !
(this is my first issue - hope I'm doing it correctly)

I'm using this code to highlights some part of an equation:

    \fcolorbox{#007bff}{white}{16x^2}

But it seems to not working. If I remove the "power 2", e.g.

    \fcolorbox{#007bff}{white}{16x}

it is then working. Am I doing something wrong ? Is this a limitation ?

Thanks!

edemaine:
In LaTeX and KaTeX, `\fcolorbox` renders its third argument as text.  You need to switch back to math mode:
```latex
\fcolorbox{#007bff}{white}{$16x^2$}
```
![image](https://user-images.githubusercontent.com/2218736/47722896-c4e65780-dc29-11e8-8c1d-2aebb1c1b0fd.png)

edemaine:
Yes, this behavior matches LaTeX. It is implicit in the "Supported Functions" documentation in that the letter A appears upright and not italic. But I agree that is quite subtle, and we could perhaps make it more explicit. For example, we could make the example text include a space character. @ronkok, thoughts?

ronkok:
Yes, "Supported Functions" has always been a page that shows examples but doesn't explain them. It covers a lot of material that way, but it falls short as a tutorial. I think @edemaine's  suggestion is a good one. We'll change the example into one that contains math. Maybe `\colorbox{aqua}{$16 x^2$}`. 

There are probably other TeX boxes on that page and on the "Support Table" page that would benefit from similar treatment. I'll look into it.

I've long had the urge to augment that page with some links to explanations of TeX behavior. For instance, in the delimiter section, the page could show a <img src="https://user-images.githubusercontent.com/16403058/47855192-1e24c700-dda1-11e8-98e4-ec17db5d26fe.PNG" width="24" alt="question"> button which links to, say, [here](https://en.wikibooks.org/wiki/LaTeX/Mathematics#Automatic_sizing). Any thoughts on the links-to-explanations idea?

edemaine:
Links to relevant LaTeX documentation seems like a neat idea. For example, for features that come from a specific LaTeX package, we could link to that package documentation on CTAN. The wikibook is a reasonable other source. Nice idea!

qcho:
Had the exact same issue.
It seems  the change proposed by @ronkok never reached the documentation. Sending a PR.

ronkok:
Fixed via PR #2288.

PR #2280, opened 2020-06-28
Add detail about colorbox third text parameter

qcho:
As stated in issue #1766


qcho:
[clabot:check]

qcho:
Wording is perfect now! No further changes seem necessary.

I see what you mean with `remarkable-katex` and it's `braceLevel` issue. Hope it gets fixed soon by the maintainer. 

If that doesn't happen:
1. you may change the configuration of it `{delimiter: '$'}` for other one in the Katex documentation and avoid the problem alltoguether.
2. maybe you or me or someone can send the PR to `remarkable-katex`. We'll see :)

bradhowes:
Hi everyone. Developer for the remarkable-katex plugin. Thanks for raising the issue. I've implemented a fix and my test case ` mdWithDollar.render('Equation $\\colorbox{aqua}{$F=ma$}$.')` passes. I've pushed a new version 1.1.5. Please let me know if there are still issues.

edemaine:
@bradhowes Wow, that was fast! Thanks for taking that on and fixing!

I just pushed a new version and https://deploy-preview-2280--katex.netlify.app/docs/next/supported.html#style-color-size-and-font now looks good to me. If someone (@ronkok?) could spot check the rest to make sure there aren't any new issues, that would be appreciated.

@ylemkimon I changed the way that `remarkable-katex` is embedded, ignoring it in `.eslintignore` instead of changing its style to conform to ours, so it's easier to upgrade in the future. (I tried to use it as an NPM package, but I see the issue is that we need to `require("../..")` instead of `require("katex")`.) Let me know if you disagree with that approach / have a better idea!

Issue #2282, opened 2020-06-30
Add \origof and \imageof


sven-frotscher:
**Is your feature request related to a problem? Please describe.**
The `\origof` and `\imageof` commands to insert the &origof; (U+22B6 ORIGINAL OF) and &imof; (U+22B7 IMAGE OF) characters respectively aren't recognized.

**Describe the solution you'd like:**
Insert the respective glyph as relation symbols when using `\origof` or `\imageof`.

**Link to or name of a (La)TeX package that provides the same feature:**
The `stix` package/font provides these two glyphs. Certain symbol fonts may support them too. If you see them in the first section of this issue, your browser already has one such font at its disposal.

**Describe alternatives you've considered:**
You can put the characters (&origof; or &imof;) or their SGML escape sequences (`&origof;` and `&imof;`) inside a formula between or two formulae. (The escape sequences is how I got the characters into this issue message, written in Markdown.) It should go without saying that that is just a workaround, since both of these methods also mess with `\stackrel`.

**Additional context:**
These symbols are often used in system theory to denote transformations (e.g. Fourier, Laplace, z, ... transforms). The hollow "bubble" usually denotes the time domain signal, whereas the filled "bubble" denotes the transformed (usually frequency) domain signal. I have also seen L and z put above these to clarify which transformation is meant. In TeX, one can use e.g. `\stackrel{L}{\origof}` for the Laplace transform.

ronkok:
@sven-frotscher I've turned in PR #2283 to implement these two functions. 

The KaTeX fonts do not contains these two characters, so these function will get their glyph from a system font. They will also write a console error every time they are used. I don't have the skills to add these glyphs to the fonts, so there isn't much I can do about the error messages.

ronkok:
Closing via PR #2283

PR #2283, opened 2020-07-01
Support \origof and \imageof

ronkok:
Address #2282. 

LaTeX package `fdsymbol` includes `\origof` and `\imageof` as relations. So does this PR. I've written this so as to also support input of the Unicode characters U+22B6 and U+22B7. Here's a screenshot for the code `A \origof B \imageof C ⊶ D ⊷ E`:
imageof

Compare that to this image from quicklatex.com: imageof

I have omitted a screenshot test. These two characters are not in the KaTeX fonts, so they will get their glyph from a system font. They will also write a console error every time they are used. I don't have the skills to add these glyphs to the fonts, so there isn't much I can do about the error messages.


sven-frotscher:
Thank you for implementing this!

Should these two glyphs be added to the KaTeX fonts? If so, from where should they be taken and which font should accomodate them?

ronkok:
Yes, they should be added to the KaTeX fonts. That would get rid of the console error messages.

The KaTeX font repo is [here](https://github.com/KaTeX/katex-fonts). The STIX symbol seems reasonable to me. I won't be the one reviewing that PR, though.

PR #2284, opened 2020-07-01
Update docs/supported.md

sven-frotscher:
Put some "direct input" characters (back) inside the $ $.
They were outside and therefore not rendered in math mode.


sven-frotscher:
Putting `‼` inside $ $ apparently breaks the formula it is in completely.
See the Netlify page, the rest of the page is apparently consumed inside `<span class="katex-mathml">`.

In the demo on katex.org, `‼` works in math mode, `\text{‼}` doesn't (`Can't use function '\mathclose' in text mode at position 1: \mathclose{!\mkern-0.8mu!…`).

Strange.

sven-frotscher:
That seems to fix page building. Maybe the handling of `‼` is worth a separate issue.
Or it should just be taken out of the docs.

sven-frotscher:
Trying something for the CI checks

ronkok:
@sven-frotscher Those characters were left outside the `$…$` delimiters for a reason. I expect that some people will copy-and-paste from the Supported Functions page to their text editor. The KaTeX fonts have those glyphs at Unicode code points that do not match Unicode. If those characters were rendered somehow by KaTeX, then a copy-and-paste would result in the wrong character.

That was my reasoning, be it good or bad. You are in good company to question it. @edemaine did the same thing when we created that page.

sven-frotscher:
@ronkok I've stumbled into a(nother) font rabbit hole, haven't I?

ronkok:
Yep. I have never mastered Docker, which is a pre-requisite to modifying the fonts. If you were willing to take that on, you could make some good improvements.

sven-frotscher:
I installed docker and ran the updateFonts script in a fresh clone of the repo. What I got was a bunch of modified font files and a confused look on my face. It seems I need to learn Perl and Make first to understand what's going on there.
The katex-fonts docs are quite lacking for newcomers...
Thanks for your words of encouragement though.

edemaine:
Part of the problem is that there are a bunch of outstanding PRs in the fonts repo that will clean up the font build process and in particular make font files consistently (so if there aren't any changes, the font files won't change). Partly my bad for not reviewing them. They should likely be reviewed before further PRs get created.

sven-frotscher:
I have, but am still unsure as to which font file(s) I should modify before buildFonts.
src/fonts/generate_fonts.py leads me to believe it's the TTFs, is that correct?

edemaine:
@sven-frotscher You definitely don't change the TTF files! Those are generated, not the source. Generally you change the following files:
* `src/fonts/OTF/TeX/makeFF`
* `src/metrics/mapping.pl`

ronkok:
Since this PR has diverted from documentation to a discussion of fonts, I'm going to close it. Font discussions should go into its own issue.

sven-frotscher:
A way around the font ... irregularities might be to put the "direct input" characters in the table.
For example, both `$ℂ$` and `$\mathbb C$` render as "ℂ in math mode".
If I have time, I might make a PR for that.

ronkok:
That would make no difference. KaTeX renders `$ℂ$` with something that looks like "ℂ" (U+2102), but it is really supplying a letter at Unicode code point U+0043, which usually renders as "C". It only looks like "ℂ" because the KaTeX fonts have glyphs placed in the old-fashioned pre-Unicode arrangement.

sven-frotscher:
So `$ℂ$` and `$\mathbb C$` both render as "something that looks like ℂ" (but really is a C in a different font).
I think someone copy-pasting characters into markup code is usually more concerned with how the page looks than what font is used to render which codepoint(s).

My possible future PR would place both ways to produce e.g. a ℂ (or rather a ℂ lookalike) in the "Logic and Set Theory" table or near the `\mathbb` in the "Font" table.
You are, of course, welcome to discuss said PR if and when I get to it :)

Issue #2288, opened 2020-07-09
Macros parameter text with delimiters


vmx:
> Before reporting a bug
- [x] Check [common issues](https://katex.org/docs/issues.html).
- [x] Check the bug is reproducible in [the demo](https://katex.org). If not, check KaTeX is up-to-date and installed correctly.
- [x] Search for [existing issues](https://github.com/KaTeX/KaTeX/issues).

**Describe the bug:**

If the parameter text in a macro (via `/def`) contains a delimiter, KaTex errors.

**To Reproduce:**
Use the input box on https://katex.org/ and insert

```tex
\def\foo ABC#1{#1}
\foo ABC bar
```

The error is:  `KaTeX parse error: Expected 'EOF', got '#' at position 13: \def\foo ABC#̲1{#1} \foo ABC …`

**Expected behavior:**

It should render as `bar`, the same way as without a delimiter:

```tex
\def\foo #1{#1}
\foo bar
```

**Environment (please complete the following information):**
 - KaTeX Version: https://katex.org/ on 2020-07-09
 - Device: Desktop
 - OS: Debian
 - Browser: Firefox
 - Version: 76.0.1

**Additional context:**

It came up when porting a document from MathJax to KaTex and the definition for a ternary operator looked like this:

```tex
\def\tern#1?#2:#3{#1\ \text{?}\ #2 \ \text{:}\ #3}
```

*Update*: I forgot to add a link to good resource about delimiters in macro parameters: https://www.overleaf.com/learn/latex/How_TeX_macros_actually_work:_Part_4#.3Cparameter_text.3E_and_macro_delimiters

edemaine:
You're right that `\def` doesn't support delimiters, sadly. This will be a substantial undertaking because the existing `MacroParser` doesn't support delimiters, and that will need to be generalized first.  Pull requests are definitely welcome if you're willing to work on this!

I will add a line to [the documentation](https://katex.org/docs/supported.html#macros) to clarify that this isn't supported presently.

ylemkimon:
Actually, #2085 will add support for delimited macros.

alvorithm:
Hi, I am struggling to define a macro with custom delimiters such as

 ```latex
\def\tern#1?#2:#3{#1\ \text{?}\ #2 \ \text{:}\ #3}

as an entry for [the macros object](https://katex.org/docs/options.html). 

[...]

PR #2289, opened 2020-07-09
Document \def doesn't support delimiters (#2288)

edemaine:
`\def` has never supported delimiters, but I don't think we mention it anywhere, which recently led to #2288.  Added mention to function support documentation.

See https://deploy-preview-2289--katex.netlify.app/docs/next/supported.html#macros


edemaine:
Thanks @ronkok!

PR #2294, opened 2020-07-11
Link to Netlify on front page

edemaine:
Netlify's [Open Source Policy](https://www.netlify.com/legal/open-source-policy/) requires (and it seems like a good idea to do) the following:
> Must feature a link to our service on your main page, or all internal pages. You have two options:
> * We have premade badges for your convenience, or
> * You may create your own link, which should read “This site is powered by Netlify”, and include a link back to our home page.

I tried both and found that the latter fit best with our existing footer styling.

Once this is merged, we can [apply for an open source exemption](https://opensource-form.netlify.com/) to the limit we've come close to hitting in the past. (I just got notification that we're halfway there for the month, given the recent activity.) @kevinbarabash You could apply, or I can apply on your behalf. Let me know your preference.


kevinbarabash:
Thanks for meeting Netlify's requirements and for applying. 😄

PR #2303, opened 2020-07-13
Drop support for IE 11, Chrome 49, Firefox 52, Safari < 11, Android 4.4 and Samsung Internet 4

ylemkimon:
* IE 11
  * Initial release: 2013
  * Still supported, but Microsoft recommends using Microsoft Edge as the default browser
  * Global usage: 1.38%

All browsers below have reached their end-of-life and are no longer supported.

* Chrome 49
  * Last release: 2016
  * The last version supported on Windows XP, Windows Vista, Mac OS X 10.6, 10.7, and 10.8.
  * Global usage: 0.33%
* Firefox 52
  * Last release: 2018
  * Was an Extended Support Release (ESR)
  * Global usage: 0.11%
* Safari < 11
  * Last release: 2017
  * Global usage: ~0.5%
* Android 4.4
  * Last release: 2014
  * Global usage: 0.14%
* Samsung Internet 4
  * Last release: 2016
  * Global usage: 0.28%

If you have any concerns, please feel free to leave a comment.


ylemkimon:
Blocked on #2304.

edemaine:
I worry about dropping IE support altogether. 1% is a lot of users. https://www.microsoft.com/en-us/download/internet-explorer.aspx claims that "Internet Explorer 11 is the recommended browser for Windows 7" and I imagine there are still a lot of people on Windows 7.

ylemkimon:
@edemaine The support for Windows 7 has ended on January 14, 2020. But I agree that still a significant number of people are using Windows 7 and/or IE 11. However, I think most people who use IE probably have other browsers installed and it has become too much technical debt to support IE. IMHO, v0.12.0 is stable enough to be the last version to support IE 11, and maybe we can backport important changes, if any.

ronkok:
When we dropped support for IE 9, there was an obvious payoff. We could then use CSS flexbox. I'm not quite as clear what benefit we receive from dropping IE 11. What do we gain?

ylemkimon:
ES6 transpilation, such as `let`, `const`, arrow functions, and classes won't be needed anymore, so performance might increase. (I'm not sure, though.) Maybe we can release both versions, transpiled and non-transpiled, or add documentation on transpiling for IE 11, until we break something for sure.

edemaine:
I think the fundamental problem is that someone hosting a website doesn't know what browsers users will use to visit. Meteor has a way of including one codeset or another depending on how old the browser is; I guess that would be the ideal solution? For our purposes, we could use offer the transpiled + non-transpiled versions -- is the latter different from the `.mjs` version?

ronkok:
If transpilation is the major (or only) difference, I would prefer to keep IE for the time being. We already make life sufficiently complex for web site administrators. We don't need to add more complexity for such small stakes.

When IE drops to a sufficiently small share, we can drop it altogether and make a clean break.

ylemkimon:
@edemaine I totally forgot about ES modules (`.mjs`).

@ronkok I entirely agree.

PR #2304, opened 2020-07-13
Run screenshotter using Chrome 83 and Firefox 76

ylemkimon:
Versions of browsers running the screenshotter are too old. I think we should have a versioning policy on the screenshotter.


ylemkimon:
I've reviewed each screenshot and updated those not changed significantly and two screenshots which seem more correct.

* **Exponents:** vertical shift on Chrome and horizontal shift on Firefox
Chrome
Firefox

---

* **Unicode:** this seems to be caused by changes in Cyrillic rendering rather than KaTeX
Chrome
Firefox

---

* **Sizing:** vertical shift on Firefox
Firefox

---

@ronkok Could you check whether those changes are acceptable?

ronkok:
@ylemkimon I agree that we are better off using updated browsers for the screenshots. And I also agree with your diagnosis of the Cyrillic spacing. This all looks to me like a win.

ylemkimon:
@ronkok Thank you.

---
I've updated screenshots and this PR is ready for review.

PR #2305, opened 2020-07-14
Run screenshotter on Safari 13.1

ylemkimon:


kevinbarabash:
@ylemkimon this is neat.  Did you run into issues with BrowserStack?

ylemkimon:
@kevinbarabash No, I've opened another PR: #2306, which add support for Browserstack to the screenshotter itself.

PR #2309, opened 2020-07-15
chore(ci): use base revision provided by CircleCI

ylemkimon:
instead of GitHub API.

It is tested in #2307 and #2308.


.circleci/config.yml
```diff
@@ -19,21 +19,6 @@ commands:
           command: |
               git submodule sync
               git submodule update --init --recursive
-      - run:
-          name: Get changed files
-          command: |
-              if [[ $CIRCLE_PULL_REQUEST ]]; then
-                BASE_COMMIT=$(curl -s -H "Authorization: token a61ecb2fc5b72da54431""1b3db3875c96854958a8" \
-                  https://api.github.com/repos/$CIRCLE_PROJECT_USERNAME/$CIRCLE_PROJECT_REPONAME/pulls/$CIRCLE_PR_NUMBER \
```
kevinbarabash, line 27: If we revoke this token, people will have to merge `master` into their PRs right?

ylemkimon: @kevinbarabash I was going to, but you're right. Thank you for pointing that out. I'll leave it for the time being.

PR #2301, opened 2020-07-12
build(deps): update dependencies

ylemkimon:
* Flow
  * https://medium.com/flow-type/making-flow-error-suppressions-more-specific-280aa4e3c95c
* Jest
  * `jest-serializer-html`: https://github.com/rayrutjes/jest-serializer-html/pull/17
* Explicitly declare dependencies
* Commit `/website/yarn.lock`

Closes #2259.

Issue #2035, opened 2019-07-14
Change dependabot settings to only open PRs for security issues


kevinbarabash:
The number of PRs we're getting from dependabot is a little ridiculous.  I've enabled the following settings:
<img width="473" alt="Screen Shot 2019-07-13 at 8 09 31 PM" src="https://user-images.githubusercontent.com/1044413/61177753-32fd5f80-a5aa-11e9-89a7-a8730d15b3dc.png">

If anyone feels that we should turn off these settings let me know.

kevinbarabash:
I'm going to close all non-security related PRs from dependabot.

kevinbarabash:
I changed it again since PRs for security issues are important even if they require a change to package.json and not just yarn.lock.  There's some secondary settings about when to make changes to package.json.  Currently the default is selected.
<img width="550" alt="Screen Shot 2019-07-13 at 8 12 28 PM" src="https://user-images.githubusercontent.com/1044413/61177771-92f40600-a5aa-11e9-9bc4-dc29eda9d935.png">

kevinbarabash:
One library I wouldn't mind getting automatic updates for even if there aren't security issues is `flow`.  Maybe dependabot has some sort of config file where that can be specified.

edemaine:
Part of this is probably because we haven't been running greenkeeper for a while, but Friday's updates might be something more.  Maybe we should remove some dependencies 😃.  Is there a way to just update for "major" revisions (second number, not third), to cut down?

Regarding "Update strategy", I think it'd be a little nicer to update `package.json`, but not a big deal either way.

ylemkimon:
I've opened #2311 to enable GitHub Dependabot:

> I think we should enable it for all dependencies as (1) KaTeX build process may break even if it's minor update as in #2301, (2) it's easier to review and fix when there are few changes to the dependency, and (3) performance and compatibility might improve.
>
> By default, it raises a maximum of five pull requests, and this value is configurable. I think this is a reasonable amount.

ylemkimon:
Closing as #2311 is merged.

PR #2311, opened 2020-07-15
chore: enable GitHub Dependabot

ylemkimon:
Dependabot is now part of GitHub and can be configured within the repo.

Regarding #2035, I think we should enable it for all dependencies as (1) KaTeX build process may break even if it's minor update as in #2301, (2) it's easier to review and fix when there are few changes to the dependency, and (3) performance and compatibility might improve.

By default, it raises a maximum of five pull requests, and this value is configurable. I think this is a reasonable amount. There seems to be an auto-merge feature, if the backlog of PRs is a concern.

As most of our dependencies are outdated, I think we should get #2301 merged first, before enabling Dependabot.


ronkok:
@ylemkimon I agree with this PR and I agree that PR #2301 should be merged first. If you could update these branches with the base branch, I would be happy to give them a positive review.

In the meantime, I have a question. In my local repository, I just tried to run `git fetch upstream`. The command failed and the message told me that it "cannot lock ref 'refs/remotes/upstream/dependabot'". It had a similar problem with the `typescript` branch. Do you have any idea how I should proceed?

ylemkimon:
@ronkok Try `git gc --prune=now`, `git remote prune origin`, and deleting `.git/refs/remote/origin` (from [Stack Overflow](https://stackoverflow.com/questions/2998832/git-pull-fails-unable-to-resolve-reference-unable-to-update-local-ref)).

ylemkimon:
@ronkok Try `git gc --prune=now`, `git remote prune upstream`, and deleting `.git/refs/remote/upstream` (from [Stack Overflow](https://stackoverflow.com/questions/2998832/git-pull-fails-unable-to-resolve-reference-unable-to-update-local-ref), edited). I've updated #2301.

ylemkimon:
@ronkok Thank you for the review!

ylemkimon:
I can confirm the Dependabot is running on GitHub: https://github.com/KaTeX/KaTeX/network/updates/39449665.

kevinbarabash:
Do dependabot builds also trigger netlify builds?

ylemkimon:
@kevinbarabash Yes, they do. I was thinking of a way to disable them.

kevinbarabash:
In different project I tried using https://docs.netlify.com/configure-builds/file-based-configuration/#ignore-builds and checking the author of the commits.  It kind of worked, but whenever I updated a PR before merging it would run a netlify build because the author of the merge commit was me. 😞

PR #2313, opened 2020-07-16
feat(css): use postcss-preset-env

### LINKED ISSUE #1946: Use postcss-preset-env to make flexbox easier to work with
Flexbox could be used to simplify some of our layouts (#1409) and is being used to fix bugs (#1919) but it can be a bit of a pain to get it right across multiple browsers.  This is because some browser versions require prefixing of style properties/values.  Some browsers also have various bugs in old implementations of flexbox.  [postcss-preset-env](https://github.com/csstools/postcss-preset-env) promises to handle this for us.  It also integrates with browserslist allowing us to easily specify which browsers we'd like to support.

ylemkimon:
Closes #1946.

### DIFF
--- a/webpack.common.js
+++ b/webpack.common.js
@@ -56,13 +56,15 @@ const targets /*: Array<Target> */ = [
  */
 function createConfig(target /*: Target */, dev /*: boolean */,
         minimize /*: boolean */) /*: Object */ {
-    const cssLoaders /*: Array<Object> */ = [{loader: 'css-loader'}];
+    const cssLoaders /*: Array<Object> */ = [{
+        loader: 'css-loader',
+        options: {importLoaders: 1},
+    }, {
+        loader: 'postcss-loader',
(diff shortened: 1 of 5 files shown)


webpack.common.js
```diff
@@ -56,13 +56,15 @@ const targets /*: Array<Target> */ = [
  */
 function createConfig(target /*: Target */, dev /*: boolean */,
         minimize /*: boolean */) /*: Object */ {
-    const cssLoaders /*: Array<Object> */ = [{loader: 'css-loader'}];
+    const cssLoaders /*: Array<Object> */ = [{
+        loader: 'css-loader',
+        options: {importLoaders: 1},
+    }, {
+        loader: 'postcss-loader',
+        options: {plugins: [require('postcss-preset-env')()]},
```
kevinbarabash, line 64: I assume this reads our .browserslistrc.

ylemkimon: @kevinbarabash Yes, it does.

PR #2316, opened 2020-07-16
chore: migrate to Yarn 2

ylemkimon:
Yarn 2 supports PnP (Plug'n'Play). (https://yarnpkg.com/features/pnp)


.circleci/config.yml
```diff
@@ -112,7 +112,7 @@ jobs:
 
       - run:
           name: Verify screenshots and generate diffs and new screenshots
-          command: node dockers/screenshotter/screenshotter.js --selenium-ip localhost -b $CIRCLE_JOB --verify --diff --new --coverage
+          command: yarn node dockers/screenshotter/screenshotter.js --selenium-ip localhost -b $CIRCLE_JOB --verify --diff --new --coverage
```
ylemkimon, line 115: https://yarnpkg.com/advanced/migration#call-your-scripts-through-yarn-node-rather-than-node:
> **Call your scripts through `yarn node` rather than `node`**
> We now need to inject some variables into the environment for Node to be able to locate your dependencies. In order to make this possible, we ask you to use `yarn node` which transparently does the heavy lifting.

.eslintrc
```diff
@@ -93,6 +93,14 @@
...
             "jest": false
         }
+    }, {
+        "files": ["website/**/*.js"],
+        "extends": [
+            "plugin:react/recommended"
+        ],
+        "rules": {
+            "react/prop-types": 0
+        }
```
ylemkimon, line 103: ESLint has issues with shared configs and PnP (https://github.com/yarnpkg/berry/issues/8), so I've moved `website/.eslintrc` to here.

.yarnrc.yml
```diff
@@ -0,0 +1 @@
+yarnPath: ".yarn/releases/yarn-berry.js"
```
ylemkimon, line 1: https://yarnpkg.com/getting-started/install:
> Using a single package manager across your system has always been a problem. To be stable, installs need to be run with the same package manager version across environments, otherwise there's a risk we introduce accidental breaking changes between versions - after all, that's why the concept of lockfile was introduced in the first place! And with Yarn being in a sense your very first project dependency, it should make sense to "lock it" as well.
>
> For this reason, Yarn 2 and later are meant to be managed on a by-project basis. Don't worry, little will change!

.circleci/config.yml
```diff
@@ -112,7 +113,7 @@ jobs:
...
```
ylemkimon, line 116: https://yarnpkg.com/getting-started/install:
> Using a single package manager across your system has always been a problem. To be stable, installs need to be run with the same package manager version across environments, otherwise there's a risk we introduce accidental breaking changes between versions - after all, that's why the concept of lockfile was introduced in the first place! And with Yarn being in a sense your very first project dependency, it should make sense to "lock it" as well.
>
> For this reason, Yarn 2 and later are meant to be managed on a by-project basis. Don't worry, little will change!

webpack.analyze.js
```diff
@@ -1,5 +1,6 @@
 // @flow
 const {targets, createConfig} = require('./webpack.common');
+// $FlowIgnore
```
ylemkimon, line 3: Flow has issues resolving dependencies with PnP (https://github.com/yarnpkg/berry/issues/634), but as we don't have any dependencies in main codebase, I think it's an acceptable loss.

website/package.json
```diff
@@ -11,5 +11,9 @@
     "fs-extra": "^9.0.1",
     "react": "^16.8.4"
   },
-  "dependencies": {}
+  "dependenciesMeta": {
+    "docusaurus": {
+      "unplugged": true
```
ylemkimon, line 16: Docusaurus requires write access to `node_modules`, so we cannot use PnP and should unplug it. (https://github.com/yarnpkg/berry/issues/818#issuecomment-579185574)

bower.json
```diff
@@ -1,36 +0,0 @@
-{
```
kevinbarabash, line 1: Is removing this file necessary to migrate to yarn 2?

.circleci/config.yml
kevinbarabash: Is `yarn` monkey patching `node`?

website/package.json
kevinbarabash: If `docusaurus` is "unplugged" does this mean that it and all of its dependencies are installed into `node_modules`?

webpack.common.js
```diff
@@ -1,11 +1,17 @@
...
 const MiniCssExtractPlugin = require('mini-css-extract-plugin');
+// $FlowIgnore
+const PnpWebpackPlugin = require('pnp-webpack-plugin');
 
 const {version} = require("./package.json");
 
+// $FlowIgnore
 const browserslist = require('browserslist')();
+// $FlowIgnore
 const caniuse = require('caniuse-lite');
```
kevinbarabash, line 15: Why do we need `$FlowIngore`s for these imports now?

.gitignore
```diff
@@ -27,3 +27,16 @@ website/static/static/katex.*
 website/static/static/fonts
 website/static/static/copy-tex.*
 website/static/static/mhchem.*
+
+# https://yarnpkg.com/advanced/qa#which-files-should-be-gitignored
+.yarn/*
+!.yarn/releases
+!.yarn/plugins
+!.yarn/sdks
+!.yarn/versions
```
kevinbarabash, line 36: I don't see any files checked into for this director or `.website`, did you forget to add some files to this PR?

bower.json
ylemkimon: @kevinbarabash No, I had to update the Yarn version in Netlify, where I realized it installs Bower and runs `bower install`. I've separated changes to #2372 and it's merged.

.circleci/config.yml
ylemkimon: @kevinbarabash Sorry, I've quoted the wrong documentation. https://yarnpkg.com/advanced/migration#call-your-scripts-through-yarn-node-rather-than-node:
> We now need to **inject some variables into the environment** for Node to be able to locate your dependencies. In order to make this possible, we ask you to use `yarn node` which transparently does the heavy lifting.
>
> Note: this section only applies to the _shell CLI_. The commands defined in your `scripts` are unaffected, as we make sure that `node` always points to the right location, with the right variables already set.

webpack.common.js
ylemkimon: @kevinbarabash See https://github.com/KaTeX/KaTeX/pull/2316#discussion_r456137525 above.

.gitignore
ylemkimon: @kevinbarabash They are files recommended by [the doc](https://yarnpkg.com/advanced/qa#which-files-should-be-gitignored). Yarn automatically uses Yarn in parent directories, so there are currently no files for `website`.

website/package.json
ylemkimon: @kevinbarabash AFAIK, only the dependency itself is unplugged. Its dependencies remain PnP-ed.

.circleci/config.yml
ylemkimon: Looking at the source, it seems it just adds `--require .pnp.js` to `NODE_OPTIONS` variable.

edemaine:
I was looking at Yarn 2 the other day. It's annoying that it requires checking in the entire yarn binary (2MB). PnP looks pretty cool though!

ylemkimon:
@edemaine But I think it's in the right direction. It's possible to use a single, even outdated, Yarn across projects. Git nowadays handles large files really smoothly.

webpack.analyze.js
ylemkimon: Since we don't have `flow-typed` for them and they don't have type information, so nothing is changed.

webpack.common.js
kevinbarabash: Thanks.

ylemkimon:
I wonder Dependabot will work with Yarn 2 (https://github.com/dependabot/dependabot-core/issues/1297).

website/package.json
kevinbarabash: That's pretty cool.

kevinbarabash:
🤞

PR #2321, opened 2020-07-17
build: use semantic-release to automate release

ylemkimon:
Closes #2011. I'm really excited and surprised that it's really simple to use the `semantic-release`.

## Previous release process
* Run `release.sh` with the version to release
  * (1) Create a release branch
  * (2) Update `package.json` to the version to release
  * (3) Build
  * (4) Update `package.json` to the next version
  * (5) Update the README and documentation
  * (6) Generate a new version of the docs
  * (7) Make a release commit and push
* (8) Update CHANGELOG
* (9) Open PR
* (10) Update SRI hashes if files are changed
* Run `release.sh -p`
  * (11) Update `package.json` to the version to release
  * (12) Build distribution (`yarn dist`)
  * ~(13) Make another release commit~ **Removed in #2385**
  * (14) Make a tag and push
  * (15) Publish to `npm`
* (16) Create a GitHub release and upload artifacts

## New release process (fully automated)
* A PR is merged to a release branch, e.g., `master` ([Triggering a release](https://github.com/semantic-release/semantic-release#triggering-a-release))
* Determine the type of release based on commits added since the last release. ([@semantic-release/commit-analyzer](https://github.com/semantic-release/commit-analyzer))
  * [x] **All PRs should follow [Angular Commit Message Conventions](https://github.com/angular/angular.js/blob/master/DEVELOPERS.md#-git-commit-guidelines)**
  * [x] **Updated CONTRIBUTING.md and added a pull request template (Fixes #1647)**
  * [x] ~@kevinbarabash I've sent a request to install https://github.com/apps/semantic-pull-requests to enforce the convention~
* Generate release notes for commits added since the last release. ([@semantic-release/release-notes-generator](https://github.com/semantic-release/release-notes-generator))
* (8) Update CHANGELOG ([@semantic-release/changelog](https://github.com/semantic-release/changelog))
* (2, 11) Update `package.json` to the version to release ([@semantic-release/npm](https://github.com/semantic-release/npm))
  * `postversion`
    * (3, 12) Build distribution (`yarn dist`)
    * (5) Update the README and documentation
* (7) Make a release commit and push ([@semantic-release/git](https://github.com/semantic-release/git))
  * [x] **Permission to push to protected branch is required**
[...]

Issue #1411, opened 2018-06-06
Restructure the repo as a mono-repo


kevinbarabash:
What might this look like:
- /
- /packages
- /packages/katex
- /packages/katex-auto-render
- /packages/katex-copy-tex
- /packages/katex-mathtex-script-type
- /packages/katex-define-font
- /devserver
- /website

Each of the folders within `packages` would have their own package.json file which would allow us to publish separate packages for `katex` itself as well as each of our extensions.  Right now we bundle all of the extensions with each update to `katex` but the code in `contrib` hardly ever changes.

The `devserver` folder would have it's own package.json which would allow us to develop a more fully featured local testing app using React.  Although it has it's own package.json we wouldn't publish this or `website` to npm.

Many of these packages will need to require each other as dependencies, e.g. `katex-define-font` would have `katex` as a dependency and `devserver` would require `katex` along with all the extension packages.

All of these dependencies can be managed relatively easily using yarn workspaces.  The gist of workspaces is that all sub-package dependencies are hoisted into the root's node_modules folder.  We've started doing this for a new project at Khan Academy and I've been quite pleased so far with the set up.  See https://yarnpkg.com/lang/en/docs/workspaces/ for more details about yarn workspaces.

ylemkimon:
I was thinking the exactly same thing. Major projects like [Babel](https://github.com/babel/babel), [React](https://github.com/facebook/react), [Jest](https://github.com/facebook/jest), and many more already do this. And it can be easily done using tools like [Lerna](https://github.com/lerna/lerna).

I think we can move `katex-fonts` and possibly `katex-test-fonts` repo back to the main repo. And for consistency, I think we can have `devserver` and `website` under the `packages`, but exclude them from publishing.

ylemkimon:
Also, `katex-cli`.

ylemkimon:
Some thoughts:
```
KaTeX/
├── .circleci/
├── docs/
├── devserver/
├── packages/
│   ├── katex/
│   │   ├── src/
│   │   └── test/
│   ├── katex-auto-render/
│   ├── katex-copy-tex/
│   ├── katex-fonts/
│   ├── katex-mathtex-script-type/
│   ├── katex-screenshotter/
│   └── katex-texcmp/
├── scripts/
│   ├── check-node-version.js
│   ├── release.sh
│   └── update-sri.js
└── website/
```

kevinbarabash:
[...]

PR #2323, opened 2020-07-18
build: separate build and dist directory

ylemkimon:
In #1500, the build directory was merged into `dist`. This PR reverts this change and separates `build` and `dist` directory.

More specifically,
|      | `build` | `dist` |
| ---- | -------- | ------- |
| **Contains** | locally built files for the current code | built files for the **latest version** |
| **`.gitignore`** | ignored | **not** ignored; this enables (i) using KaTeX and CLI out of the box after cloning (#894) and (ii) using the repo directly as a dependency in `package.json` (#1229). This is also a step toward #1411. |
| **Generated by** | `yarn build` | `yarn dist`; the original issue, `yarn dist` taking too long (#1498), will be fixed in #2321 by removing the test stage from it |
| **Used in** | development, website | `package.json` main field (`require`), CLI, publish |
| **Release process** || Files were originally updated in a detached commit, which is then tagged with the version. This made version tags not reachable from `master`, i.e., `git tag --merged master` doesn't include version tags, and caused problems with tools relying on this. They are now included in the release commit, which is committed to `master`. As a side effect, the `package.json` version field points to the latest version, which makes more sense.  |


.gitignore
```diff
@@ -1,12 +1,12 @@
 katex.tar.gz
 katex.zip
+build
 node_modules
 npm-debug.log
 debug.log
 last.png
 diff.png
 /.npm-install.stamp
-/dist/
```
kevinbarabash, line 9: Is this because people shouldn't be running `yarn dist` locally?

cli.js
```diff
@@ -6,17 +6,7 @@
...
-    katex = require("./");
-} catch (e) {
-    console.error(
-        "KaTeX could not import, likely because dist/katex.js is missing.");
-    console.error("Please run 'yarn' and 'yarn build' before running");
-    console.error("cli.js from the KaTeX repository.");
-    console.error();
-    throw e;
-}
+const katex = require("./");
```
kevinbarabash, line 9: Is `dist` going to be checked in now?

release.sh
```diff
@@ -19,15 +18,12 @@ usage() {
         shift
     done
     echo "Usage:"
-    echo "./release.sh [OPTIONS] <VERSION_TO_RELEASE> [NEXT_VERSION]"
```
kevinbarabash, line 22: Are we no longer bumping the version to `NEXT_VERSION-pre`?

.gitignore
ylemkimon: @kevinbarabash They probably shouldn't and won't be needed.

cli.js
ylemkimon: @kevinbarabash Yes. This enables (i) using KaTeX and CLI out of the box after cloning (#894) and (ii) using the repo directly as a dependency in `package.json` (#1229). This is also a step toward #1411 (fonts are now checked in the main repo).

release.sh
ylemkimon: @kevinbarabash Yes. It was hard to determine which version to bump to. It's possible to bump the patch version, but it'd require two commits, release and bump, per a release. Another option is leave it `0.0.0-development` as packages with `semantic-release` do.

cli.js
ylemkimon: I was thinking of checking them in the next release, but it seems CLI and website needs it, so I've checked v0.12.0 in.

kevinbarabash: I don't think this is necessary for #1411.  jest, webpack, and flow have ways to defining module mappings so that name modules resolve to `/packages/foo` (or wherever).  Even though I filed #894, I'm not sure how big of a deal that is.  I assume the primary reason for someone to clone the repo is to do some development work of some sort.  We could maybe simplify the process a bit so that `yarn install` also builds things.

As for #1229, I forgot that you could install things using `khan/katex`.  Doing so seems like a bad idea since the resulting package.json contains `"katex": "khan/katex"` as a dependency which indicates that the dependency is non-versions.  I don't think we should bother supporting this type of installation.

Can you say a little more about the website needs `dict` checked in?

If we are checking in `dist` is the expectation that whenever a change is made to `src` that it should also be accompanied by a change to `dict`?  If so, I think it would be nice if we did the following:
- had a GitHub Action that commit files back to `dist` if there are any changes after the build step completes
- add an appropriate entry to .gitattributes so that files in `dist` are collapsed in code reviews so that they don't add extra noise

ylemkimon: @kevinbarabash #894 and #1229 are just extra benefits. The major motivation is include the release commit to `master` branch history.

> If we are checking in `dist` is the expectation that whenever a change is made to `src` that it should also be accompanied by a change to `dict`? 

No, `dist` will be only touched by the release process, i.e., semantic-release bot with #2011.

> add an appropriate entry to .gitattributes so that files in `dist` are collapsed in code reviews so that they don't add extra noise

I didn't know we can do that.

ylemkimon: Another way I can think of is not to check `dist` in at all. This would break installing KaTeX using Git tags, e.g., `"katex": "KaTeX/KaTeX#v0.12.0"` or `https://cdn.jsdelivr.net/gh/KaTeX/KaTeX@v0.12.0/dist/katex.min.js`.

ylemkimon: After careful consideration, I think it'd be better not to include `dist` in the release commit. Closing in favor of #2385.

release.sh
kevinbarabash: I think it's fine not bumping.  Thanks for confirming.

cli.js
kevinbarabash: I had a look at #2385.  I didn't realize that `dist` was being checked into a detached branch.  That's much better than what I thought was going on (checking `dist` into `master`).

> I was thinking of checking them in the next release, but it seems CLI and website needs it, so I've checked v0.12.0 in.

Is this still true?

PR #2324, opened 2020-07-19
test(screenshotter): move coverage to Jest

### LINKED ISSUE #1885: Move screenshotter coverage collection to jest
After #1644, I've realized there is no need to collect coverage in screenshotter tests, increasing overheads on the screenshotter.

ylemkimon:
Fixes #1885. After #1644, I've realized there is no need to collect coverage in screenshotter tests, increasing overheads on the screenshotter.

I'm not sure there is need to separate Jest tests coverage and screenshotter coverage, but it is possible to do it.


kevinbarabash:
The coverage went down a bit, but that's to be expected since it isn't exercise the DOM part of the code base.  We can always add some more snapshot tests later or something.

PR #2332, opened 2020-07-23
Support MathML \big, \bigg, \Big, and \Bigg

ronkok:
This PR applies the correct vertical size to delimiters specified with `\big`, `\Big`, `\bigg`, and `\Bigg`. Here's a Firefox screenshot that renders the code `(M) \big(M\big) \Big(M\Big) \bigg(M\bigg) \Bigg(M\Bigg)`:

bigg


ronkok:
In anticipation of a question, I'll point out that TeX does not adjust `\big` delimiter sizes for `scriptstyle` and `scriptscriptstyle`. Here's how quicklatex.com renders `\Bigg(M\Bigg) \; M^{M^{\Bigg(M\Bigg)}}`:



That makes this PR pretty easy.

src/functions/delimsizing.js
```diff
@@ -124,6 +124,9 @@ defineFunction({
             node.setAttribute("fence", "false");
         }
 
+        node.setAttribute("minsize", delimiter.sizeToMaxHeight[group.size] + "em");
+        node.setAttribute("maxsize", delimiter.sizeToMaxHeight[group.size] + "em");
```
ylemkimon, line 128: [MDN docs](https://developer.mozilla.org/en-US/docs/Web/MathML/Element/mo) suggest `minsize` and `maxsize` specifies the size only when `stretchy` is `true`. Is `stretchy` true by default?

ronkok: No. The default value of stretchy is false. I'll add stretchy = true.

ylemkimon: I wonder why it did work without it.

ronkok: Browsers are to render a lot of things that don't pass a W3C validation. I suppose this is just one of those cases.

ronkok:
@ylemkimon Thank you for the review. KaTeX MathML is getting close to parity with HTML.

Issue #2322, opened 2020-07-17
Katex \tag is overlapped when a formula is too long


langtutheky:
## Description

Adding a \tag to a formula, the \tag would overlap the formula when rendered if the formula is too long

## To reproduce

1. Type a very long formula into the [Katex's Checker](https://katex.org/)
2. Insert a \tag at the end of the formula
3. See the error

## Expected behaviors

No matter how long the formula is, the \tag should always be at the tail of the formula and not overlapping it

## Environments

 - Katex Version: 0.12.0
 - Device: Desktop
 - OS: MacOS Catalina 
 - OS Version: 10.15.5
 - Browser: Google Chrome
 - Broswer Version: 83.0.4103.116

## Screenshots

```Tex
\[\forall v \forall w \forall x \forall y \forall z [[Bar(z) \wedge Foo(y) \wedge [\neg Baz(y)] \wedge Bar(x, y) \wedge Foo(w, y, z) \wedge Bar(v, w)] \rightarrow v = x] \forall v \forall w \forall x \forall y \forall z [[Bar(z) \wedge Foo(y) \wedge [\neg Baz(y)] \wedge Bar(x, y) \wedge Foo(w, y, z) \wedge Bar(v, w)] \rightarrow v = x] \tag{99} \]
```
<img width="853" alt="Screen Shot 2020-07-17 at 11 20 54 AM" src="https://user-images.githubusercontent.com/9475206/87818561-9d7d5700-c81f-11ea-916c-4cfde21c465e.png">

## Note

This is a redirected issue from the original issue that I posted on [Markdown Preview Enhanced](https://github.com/shd101wyy/vscode-markdown-preview-enhanced/issues/389)

ylemkimon:
Thank you for the report! This issue is a duplicate of #1983. So I'll close the issue. Feel free to reopen or open a new issue if you have different issues.

PR #2343, opened 2020-07-28
Prevent tag overlaps

ronkok:
Prevent a `\tag` or an equation number from overlapping a long equation. This PR renders display mode using flexbox CSS. Glue elements are written to set the equation element and the tag element in their proper position.

Addresses issues #1983 & #2322.


ronkok:
Oops. There are glitches. This is not yet ready for review.

ronkok:
This is now ready for review. Here's a sample output of a long equation with a tag:

tag

Note that the equation is not centered on the page. It is centered on the section not occupied by the tag. I tried to center it on the page, but this is the best I could do. I think this is still a win. There will be no overlaps.

ylemkimon:
Wouldn't this result in different location of equations. if those with numbers and those without numbers are mixed?

ronkok:
> Wouldn't this result in different location of equations. if those with numbers and those without numbers are mixed?

Not if they were in the same environment. But otherwise, yes. That's bad, I suppose.

ylemkimon:
LaTeX pushes the tag to the next line, if the equation overflows:
image

Would it be possible to implement this behavior?

edemaine:
FYI, I investigated these options a lot when first implementing `\tag`; see #1309 and https://github.com/KaTeX/KaTeX/issues/899#issuecomment-335532235 (and other discussion in #899). My conclusion at that time was that it isn't possible to get all three desired properties:
1. Proper centering of equation
2. Tag on next line instead of overlapping
3. Proper vertical alignment with baseline

We already have the issue that (like LaTeX) display equations aren't responsive, so I figured giving up on (2) wasn't so bad. But I can imagine it's bad if you allow horizontal scrolling, say. I believe this PR gives up (1).

I do wonder about returning to the float approach. If we leave in the strut hack that's currently in KaTeX, perhaps we'd just give up (3) in the case of long equations, but that's probably better than overlap?

ronkok:
I'll open a different PR and implement a float. Then we can choose which one we prefer.

PR #2346, opened 2020-07-29
Set tags in float (WIP)

ronkok:
I offer this PR, before writing any tests, for discussion. It avoids tag overlap by setting tags as `float: right` or `float: left`. The result for `\tag` is pretty good:

tag

The problem will be in environments that apply multiple equation numbers. Like this `{gather}` environment:

gather

LaTeX would align the top equation with (1). As you can see, this PR gets it wrong.

KaTeX environments construct their HTML column-wise, not row-wise like a HTML table. When @edemaine created KaTeX tags, he inserted clever struts into the tag elements so that a tag would align with content. But KaTeX has no way of knowing when a environment equation will wrap. In that case, the struts are not enough.

I cannot think of any way to ensure alignment except to (1) allow overlaps as is currently done, or (2) rewrite `array.js` from scratch to build the HTML row-wise ~~(probably in tables)~~ instead of column-wise.

I do not volunteer to rewrite `array.js`.  We may just have to live with tag overlaps.

### DIFF
--- a/src/katex.less
+++ b/src/katex.less
@@ -622,17 +622,12 @@
     text-align: center;
 
     > .katex {
-        display: block;
-        text-align: center;
-        white-space: nowrap;
 
         > .katex-html {
-            display: block;
-            position: relative;
+            width: 100%;
 
             > .tag {
-                position: absolute;
-                right: 0;
+                float: right;
             }
         }
     }
@@ -640,8 +635,7 @@
 
 // Left-justified tags (default is right-justified)
 .katex-display.leqno > .katex > .katex-html > .tag {
-    left: 0;
-    right: auto;
+    float: left;
 }
 
 // Flush-left display math


edemaine:
KaTeX does not currently support `\tag` in multiline environments -- it's not your fault! (I get `KaTeX parse error: Multiple \tag` -- unless something changed?) I agree that this would need to be processed at the array level, and that's work best left to another PR. I hadn't realized that this would require row-based processing, but you're likely right. ☹️

ronkok:
> KaTeX does not currently support \tag in multiline environments

Point taken. But my intent in this PR was to write HTML element-locating code that would support both `\tag` and environments with multiple equation numbers. I think it is probably possible to get `\tag` pretty close to what we want. I do not think it is possible to get multiple equation numbers to work nearly as well without a major rewrite. So I'll probably close this PR.  

Maybe I could write code for `\tag` that is distinct from the code for equation numbers. Then try to improve just the `\tag` part. and let the equation numbers overlap. That might be worth an effort. I'll look at it tomorrow.

PR #2358, opened 2020-07-30
Prevent \tag overlap

ronkok:
Prevent a `\tag` from overlapping a long equation. Use `float: right;` or `float: left` to locate the tag. Addresses issues #1983 & #2322.  Screenshot:

tag

The automatic equation numbers in `{align}`, `{alignat}`, and `{gather}` will continue to have overlaps, as shown below.
gather

That's because I can only offset the two equation numbers as a group, not individually. The overlap is the lesser of two evils.


ronkok:
Nope. That did not work. After this effort, I have belatedly come to the same conclusion that @edemaine expressed in an earlier [comment](https://github.com/KaTeX/KaTeX/issues/899#issuecomment-335532235). I am giving up the search for a non-overlapped `\tag`. This PR is closing and I am moving on.

martinpflaum:
Don t give up!!

    .katex-display>.katex>.katex-html>.tag {
    position: relative;
    right: 0;
    }
works.

its derived from https://github.com/KaTeX/KaTeX/issues/1983

mbourne:
@ronkok I know you gave up on this.

I proposed a solution in #1983 but got no response there. 

Is my suggestion not possible in KaTeX, or something?

PR #2369, opened 2020-08-01
feat: support {equation}, {equation*}, and {split}

ronkok:
Three more AMS environments.

~~In AMS LaTeX, the `{split}` environment is allowed only inside certain other environments, like `{equation}` or `{gather}`. I've set it so that in KaTeX, the `{split}` environment is allowed only inside `{equation}`. That's where it will be most valuable. If we get requests for other nesting, we can revisit this code at some later time.~~

Here's a screenshot:
Equation


ylemkimon:
Isn't KaTeX in `equation*` environment by default?

ronkok:
No, KaTeX is in math mode by default. `{equation*}` only allows one line and one column. It will return an error if the input includes a `&` or a `\cr` and it will eat a `\\`.

ylemkimon:
One of default delimiters in `auto-render` is `\[` and `\]`, which are AFAIK macro to `equation*` environment:
https://github.com/KaTeX/KaTeX/blob/903b459ca42f6689fa960bc3bc1d38b16b57dbed/contrib/auto-render/auto-render.js#L115
I think we should allow `split` at the outermost (top) level.

ronkok:
You have a point. It seems that `\[…\]` is indeed equivalent to `{equation*}`. On the other hand, `$$…$$` is *not* equivalent to `{equation*}`. `$$…$$` does not allow `{split}` at the top level. 

That makes things a little awkward. The KaTeX option for `displayMode` does not distinguish between `\[…\]` and `$$…$$` and I really do not want to create yet another KaTeX option. That means we have to make KaTeX `{split}` behavior agree with either `\[…\]` or `$$…$$` but not both.

So I agree to fall in with your views. I'll rewrite `{split}` so that it is allowed at the top level whenever `displayMode` is set to true. That simplifies the code, which is an added benefit. `{split}` becomes just a version of `{align*}` with a restriction to no more than two columns.

If anyone has an objection to this plan, please let me know.

After this PR comes `multline` and after that PR, I'll turn in a documentation PR that tries to convey all these nuances.

edemaine:
I believe the only reason that `$$` doesn't support `split` is that it's (nearly) impossible (for amsmath) to redefine `$$`'s behavior in TeX. So I think it's fine to err on the side of allowing things, as if `$$` were in fact `\[`.

src/environments/array.js
```diff
@@ -290,6 +299,10 @@ const htmlBuilder: HtmlBuilder<"array"> = function(group, options) {
         setHLinePos(hLinesBeforeRow[r + 1]);
     }
 
+    if (group.maxNumCols && nc > group.maxNumCols) {
+        throw new ParseError("Too many tab characters: &");
+    }
```
ylemkimon, line 304: This can be enforced in `parseArray` above, e.g.:
https://github.com/KaTeX/KaTeX/blob/e3c01376c85c82180f2985a18550e481f7e06b3f/src/environments/array.js#L123-L125
```js
        const token = parser.fetch();
        const next = token.text;
        if (next === "&") {
            if (group.maxNumCols && row.length > group.maxNumCols) {
                throw new ParseError("Too many tab characters: &", token);
            }
            parser.consume();
```

src/parseNode.js
```diff
@@ -40,6 +40,8 @@ type ParseNodeTypes = {
         rowGaps: (?Measurement)[],
         hLinesBeforeRow: Array<boolean[]>,
         addEqnNum?: boolean,
+        maxNumRows?: 1,     // equation allows only 1 row.
+        maxNumCols?: 1 | 2, // equation: 1, split: 2.
```
ylemkimon, line 44:
```suggestion
```
`maxNumRows` is not used, i.e., not returned by `parseArray`, and `maxNumCols` can be removed with the change above.

src/environments/array.js

src/parseNode.js
ronkok: Done.

src/environments/array.js
```diff
@@ -122,6 +130,9 @@ function parseArray(
         row.push(cell);
         const next = parser.fetch().text;
         if (next === "&") {
+            if (maxNumCols && row.length === maxNumCols) {
+                throw new ParseError("Too many tab characters: &");
```
ylemkimon, line 134:
```suggestion
                throw new ParseError("Too many tab characters: &",
                                     parser.nextToken);
```
`ParseError` can provide position information if token is given as the second argument.

src/environments/array.js
```diff
@@ -934,6 +947,36 @@ defineEnvironment({
...
+        }
+        const res = {
+            addEqnNum: context.envName === "equation",
+            maxNumRows: 1,
+            maxNumCols: 1,
+            leqno: context.parser.settings.leqno,
+        };
+        const equation = parseArray(context.parser, res, "display");
+        if (equation.body.length > 1) {
+            throw new ParseError("Misplaced \\cr.");
```
ylemkimon, line 972: This also can be move to `parseArray`. Sorry for spread reviews, I didn't notice this at first.

src/environments/array.js
```diff
@@ -67,13 +69,19 @@ function parseArray(
         arraystretch?: number,
         colSeparationType?: ColSeparationType,
         addEqnNum?: boolean,
+        maxNumRows?: 1,
+        maxNumCols?: 1 | 2,
```
ylemkimon, line 73:
```suggestion
        maxNumRows?: number,
        maxNumCols?: number,
```
I see no reason to limit them to 1 or 2.

ronkok: My reasoning was:
* `{equation}` is limited to one column
* `{split}` is limited to no more than two columns
* Every other environment has no limit.

So an environment can either pass on putting `maxNumCols` into the payload, or specify 1 or 2. Anything else makes no sense.

src/environments/array.js
ronkok: It can move into `parseArray` but I don't know if it should. If it is is inside `parseArray`, then every environment does that check. As currently coded, only `{equation}` and `{equation*}` do that check. One method of getting good performance is to put less code into the inner loop.

src/environments/array.js
ylemkimon: Doesn't `{array}` have `maxNumCols` of the number of its alignment specifiers?

src/environments/array.js
ylemkimon: I don't think a simple truthiness check will affect much performance. IMHO, I think it's better to fail early with the token information.

src/environments/array.js
ronkok: You know, that would be a good thing to write into the KaTeX code for `{array}`. It is not there now. And then your point about `maxNumCols` type makes perfect sense. I'll do it.

ylemkimon: For `maxNumRows`, I cannot think of any use case that limits to more than one. If that's the case, I think we can make it a boolean, e.g., `singleRow`.

src/environments/array.js

src/environments/array.js

ronkok:
Well, I have screenshotter problems and CircleCI tells me that "This job has no artifacts." I'm stumped. And things were going so well.

Any suggestions would be welcome.

ronkok:
This PR now contains a **breaking change** in the `{array}` environment. If the environment body contains more columns than specified in the first argument, this PR now throws a ParseError (as does LaTeX).  Before this PR, KaTeX went ahead and created the rest of the columns and centered them.

This was driven home to me because one of the screenshotter `{array}` tests threw a ParseError. It thoroughly messed up the screenshotter tests. 

Is LaTeX purity worth a breaking change?

edemaine:
Perhaps a `strict` call would be more appropriate than throwing an error? (That's kinda what it's there for: things that don't work in LaTeX but are easy to support.) Further evidence for this is that MathJax silently accepts more columns than there are column specs.

ronkok:
> Perhaps a `strict` call...?

That seems reasonable for `{array}`. How about `{equation}` (max 1 column) and `{split}` (max 2 columns). I'm not sure why, but I think they maybe really should get a ParseError.

ronkok:
I'm not sure that the eslint check is working. I had to do a `yarn install` to get this commit to work. When I run `yarn test:lint`, I get a message: 

> Error: eslint-plugin-flowtype tried to access babel-eslint, but it isn't declared in its dependencies; this makes the require call ambiguous and unsound.

ylemkimon:
@ronkok You can safely ignore the error (actually, a warning).

ronkok:
Then this PR is ready for review.

src/environments/array.js
```diff
@@ -150,6 +169,10 @@ function parseArray(
         }
     }
 
+    if (singleRow && body.length > 1) {
+        throw new ParseError("Misplaced \\cr.", parser.nextToken);
```
ylemkimon, line 173: Could you move this after the line 157 (`if (next === "\\cr")`)?

src/environments/array.js
```diff
@@ -934,6 +959,32 @@ defineEnvironment({
...
+        numArgs: 0,
+    },
+    handler(context) {
+        const settings = context.parser.settings;
+        if (!settings.displayMode) {
+            throw new ParseError(`{${context.envName}} cannot be used inline.`);
+        } else if (settings.strict && !settings.topEnv) {
+            settings.reportNonstrict("textEnv",
+                `{${context.envName}} called from math mode.`);
+        }
```
ylemkimon, line 975: Would it be possible to extract this to a separate function?

src/environments/array.js

ronkok:
Thank you for the review!

I'll mention here that I no longer intend to submit a PR for the `{multline}` environment, for two reasons. (1) No one has ever requested it, and (2) I don't know how to reconcile our overlapping `\tag` problem with lines that are justified left or justified right.

So I will submit no more environments any time soon.  I do have some documentation to submit.

PR #2370, opened 2020-08-01
Update \color documentation

ronkok:
KaTeX's documentation asserts that the MathJax version of `\color` expects the content to be a function argument. That is no longer true, so  this PR removes that comment.


ronkok:
Good point. I took a shot at an edit to `options.md`.

docs/options.md
```diff
@@ -15,7 +15,7 @@ You can provide an object of options as the last argument to [`katex.render` and
...
 - `minRuleThickness`: `number`. Specifies a minimum thickness, in ems, for fraction lines, `\sqrt` top lines, `{array}` vertical lines, `\hline`, `\hdashline`, `\underline`, `\overline`, and the borders of `\fbox`, `\boxed`, and `\fcolorbox`. The usual value for these items is `0.04`, so for `minRuleThickness` to be effective it should probably take a value slightly above `0.04`, say `0.05` or `0.06`. Negative values will be ignored.
-- `colorIsTextColor`: `boolean`. If `true`, `\color` will work like LaTeX's `\textcolor`, and take two arguments (e.g., `\color{blue}{hello}`), which restores the old behavior of KaTeX (pre-0.8.0). If `false` (the default), `\color` will work like LaTeX's `\color`, and take one argument (e.g., `\color{blue}hello`).  In both cases, `\textcolor` works as in LaTeX (e.g., `\textcolor{blue}{hello}`).
+- `colorIsTextColor`: `boolean`. In early versions of both KaTeX and MathJax, the `\color` function expected the content to be a function argument, as in `\color{blue}{hello}`. In current KaTeX, `\color` is a switch, as in `\color{blue} hello`. This matches LaTeX behavior. If you want the old `\color` behavior, set option `colorIsTextColor` to true.
```
edemaine, line 18: Maybe worth preserving which versions: KaTeX < 0.8.0, and maybe also MathJax <= 2 to connect to colorV2?

ronkok: Version numbers are added.

docs/options.md
```diff
@@ -15,7 +15,7 @@ You can provide an object of options as the last argument to [`katex.render` and
...
-- `colorIsTextColor`: `boolean`. If `true`, `\color` will work like LaTeX's `\textcolor`, and take two arguments (e.g., `\color{blue}{hello}`), which restores the old behavior of KaTeX (pre-0.8.0). If `false` (the default), `\color` will work like LaTeX's `\color`, and take one argument (e.g., `\color{blue}hello`).  In both cases, `\textcolor` works as in LaTeX (e.g., `\textcolor{blue}{hello}`).
+- `colorIsTextColor`: `boolean`. In early versions of both KaTeX (<0.8.0) and MathJax (<2.0>), the `\color` function expected the content to be a function argument, as in `\color{blue}{hello}`. In current KaTeX, `\color` is a switch, as in `\color{blue} hello`. This matches LaTeX behavior. If you want the old `\color` behavior, set option `colorIsTextColor` to true.
```
edemaine, line 18: Sorry to keep nit-picking here:

```suggestion
- `colorIsTextColor`: `boolean`. In early versions of both KaTeX (<0.8.0) and MathJax (<3.0), the `\color` function expected the content to be a function argument, as in `\color{blue}{hello}`. In current KaTeX, `\color` is a switch, as in `\color{blue} hello`. This matches LaTeX behavior. If you want the old `\color` behavior, set option `colorIsTextColor` to true.
```

ronkok: Nope. MathJax changed the behavior of `\color` when they released version 2.0, not 3.0. [Version 2.0 release notes](https://docs.mathjax.org/en/v2.0-latest/whats-new-2.0.html#many-new-tex-additions-and-enhancements).

edemaine: I believe it's more complicated than that. In MathJax 2, `\color` works *by default* like `\textcolor`, for backwards compatibility with MathJax 1 (presumably). What the release notes says is that *if you add the `color` package*, `\color` gets LaTeX's switch behavior, which became the *default* only in MathJax 3 (which has a package `colorV2` to revert to MathJax 2's default). You can test MathJax 2 behavior on https://www.cs.odu.edu/~zeil/cs390/latest/Public/texmath/demo.html for example.

Given all this complexity, I'm not sure how to rewrite this sentence... I guess we could just omit the version number for MathJax. But we definitely shouldn't have a `>` in the version number. 🙂

ronkok: And I thought KaTeX options were too complex. I'm going to remove the MathJax version number. Our users don't need to know how many color options can dance on the head of a pin.

PR #2371, opened 2020-08-01
Enable a MathML option in the KaTeX demo.

### LINKED ISSUE #2221: MathML option for demonstration page
It would be nice if the KaTeX demonstration page had an option to render MathML instead of HTML.

ronkok:
I think this may enable a MathML option in the demo page. I'll have to see a deployed version to check.

Closes #2221.


ylemkimon:
You should also add `output` to `katexOptions`:
https://github.com/KaTeX/KaTeX/blob/903b459ca42f6689fa960bc3bc1d38b16b57dbed/website/static/js/index.js#L50-L51

You can run the website locally by `cd website && yarn start`.

ronkok:
> You should also add `output` to `katexOptions`:

Done.

> You can run the website locally by cd website && yarn start.

When I try that, I get an error message:

> 'docusaurus-start' is not recognized as an internal or external command, operable program or batch file.

ronkok:
Oh, I see. First I need to `yarn install` from the website subdirectory. It works! Thank you.

PR #2372, opened 2020-08-02
chore: delete bower.json

ylemkimon:
Bower is deprecated about three years ago and this unnecessarily [triggers `bower install` on Netlify builds](https://docs.netlify.com/configure-builds/manage-dependencies/#bower), which take about a minute.


ylemkimon:
@edemaine Thank you for the fast review!

PR #2385, opened 2020-08-04
build: don't include `dist` in the release commit

ylemkimon:
Built files (`dist`) were originally updated in a detached commit during a release, which is then tagged with the version. This made version tags not reachable from `master`, i.e., `git tag --merged master` doesn't include version tags, and caused problems with tools relying on this. The release commit is now included in the `master` branch history.

This also simplifies the release process to a single commit. The tagged commit wasn't even reviewed.

Furthermore,  mono-repo (#1411) doesn't have a 1:1 relation between a repo and a package, so it'd make no sense to install _a repo_.

This will have breaking changes, but they are minor use cases.

**BREAKING CHANGE:** The built files (`dist`) are no longer included in release tags.
1. To use KaTeX after cloning or checking out a tag, run `yarn build` to build necessary files:
```bash
git clone -b v0.12.0 https://github.com/KaTeX/KaTeX.git # or git checkout v0.12.0
yarn
yarn build
```

2. To install KaTeX, use the package available at npm. Change `package.json` from
```
{
  "dependencies": {
    "katex": "KaTeX/KaTeX#v0.12.0"
  }
}
```
to
```
{
  "dependencies": {
    "katex": "0.12.0"
  }
}
```

3. To link KaTeX, use CDN links for npm. Change
```
https://raw.githubusercontent.com/KaTeX/KaTeX/v0.12.0/dist/katex.min.js
https://cdn.jsdelivr.net/gh/KaTeX/KaTeX@v0.12.0/dist/katex.min.js
```
to
```
https://cdn.jsdelivr.net/npm/katex@0.12.0/dist/katex.min.js
```

4. To download KaTeX, use tarball or zip (not the Source Code) from [GitHub Releases](https://github.com/KaTeX/KaTeX/releases).


edemaine:
This makes sense to me. Keeping built files in Github releases, and NPM/CDN, seems sufficient.

release.sh
```diff
@@ -181,27 +181,20 @@ if [[ ! $PUBLISH ]]; then
     echo "After it gets merged, run './release.sh -p $VERSION'!"
     echo "Note that if KaTeX source code is changed after running this script,"
     echo "you have to run the release script again."
-else
-    # Make a new detached HEAD
-    git checkout --detach
 
+    git diff --stat --exit-code # check for uncommitted changes
```
kevinbarabash, line 185: nice

ylemkimon:
@kevinbarabash No, that was specific to #2323.

ylemkimon:
I'll open an issue about this breaking change and pin it for a while.

PR #2394, opened 2020-08-07
ci: autofix Dependabot commits

ylemkimon:
Temporary workaround for https://github.com/dependabot/dependabot-core/issues/2030.

Example: #2393 and #2391.

### DIFF
--- a/.github/workflows/dependabot.js.yml
+++ b/.github/workflows/dependabot.js.yml
@@ -16,16 +16,16 @@ jobs:
         node-version: '12'
     - name: Autofix lockfile
       run: |
-        # restore yarn.lock
+        # restore yarn.lock from the previous commit
         git checkout HEAD^ -- yarn.lock
-        
+
         # install yarn-plugin-deduplicate
         yarn plugin import https://raw.githubusercontent.com/eps1lon/yarn-plugin-deduplicate/latest/bin/%40yarnpkg/plugin-deduplicate.js
 
         # if package.json was not updated, upgrade the dependency
         git diff --name-only HEAD^ HEAD | grep -q 'package.json' || yarn up `git log -1 --pretty=%s | awk '{ print $3 }'`
 
-        # restore package.json
+        # restore package.json from the last commit
         git checkout HEAD -- package.json
 
         yarn install
--- a/.github/workflows/dependabot.js.yml
+++ b/.github/workflows/dependabot.js.yml
@@ -1,0 +1,40 @@
+name: Dependabot
+
+on:
+  push:
+    branches: [ dependabot/npm_and_yarn/** ]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Use Node.js 12.x
+      uses: actions/setup-node@v1
+      with:
+        node-version: '12'
+    - name: Autofix lockfile
+      run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/dependabot.js.yml
```diff
@@ -0,0 +1,41 @@
...
+
+        yarn install
+
+        # deduplicate lockfile
+        yarn deduplicate
+    - name: Commit changes
+      run: |
+        git add yarn.lock
+        git config --global user.name 'dependabot[bot]'
+        git config --global user.email 'support@github.com'
```
kevinbarabash, line 39: That's a cool way of having updates to the PR appear to be from `dependabot`.

.github/workflows/dependabot.js.yml
```diff
@@ -0,0 +1,41 @@
...
+    - name: Autofix lockfile
+      run: |
+        # restore yarn.lock from the previous commit
+        git checkout HEAD^ -- yarn.lock
+
+        # install yarn-plugin-deduplicate
+        yarn plugin import https://raw.githubusercontent.com/eps1lon/yarn-plugin-deduplicate/latest/bin/%40yarnpkg/plugin-deduplicate.js
+
+        # if package.json was not updated, upgrade the dependency
+        git diff --name-only HEAD^ HEAD | grep -q 'package.json' || yarn up `git log -1 --pretty=%s | awk '{ print $3 }'`
```
kevinbarabash, line 26: It seems like there's a disconnect between the comment and the command.  It looks like we're looking if there were changes made to `package.json` and upgrading those dependencies.  Am I misunderstanding the command?

.github/workflows/dependabot.js.yml
```diff
@@ -0,0 +1,41 @@
...
+
+        # restore package.json from the last commit
+        git checkout HEAD -- package.json
+
+        yarn install
+
+        # deduplicate lockfile
+        yarn deduplicate
```
kevinbarabash, line 34: I'm kind of surprised this is default behaviour built into `yarn`.  Do the other commands in this step making duplication more of an issue than usual?

.github/workflows/dependabot.js.yml
ylemkimon: @kevinbarabash If `grep` has succeeded, the `package.json` was updated, i.e., the version range is already updated, so it will install upgraded dependency on `yarn install` step and there is no need to run `yarn up`.

.github/workflows/dependabot.js.yml
ylemkimon: @kevinbarabash It turns out it doesn't. I've updated the email address.

.github/workflows/dependabot.js.yml
ylemkimon: @kevinbarabash Yarn doesn't dedupe dependencies when upgrading. For instance, if we have
```json
{
  "dependencies": {
    "A": "^1.2.0"
  }
}
```
```yaml
A@^1.1.0, A@^1.2.0
  version: "1.2.1"
```
and run `yarn up A`, it gives
```yaml
A@^1.1.0
  version: "1.2.1"

A@^1.2.0
  version: "1.3.0"
```

kevinbarabash: TIL 🙂

ylemkimon:

PR #2326, opened 2020-07-20
[WIP] chore: migrate to Typescript

ylemkimon:
Closes #2110.

Issue #1834, opened 2019-01-17
Katex commutative diagram


Gwynever:
Hi, commutative diagram \xymatrix does not work in Katex

 $$ \xymatrix{
FX \ar[r]^-{Ff} \ar[d]_{\eta_X} & FY \ar[d]^{\eta_Y} \\
GX \ar[r]_-{Gf} & GY\\} $$

This is a very frequently used diagram in mathematics!

ronkok:
@Gwynever Would the AMS `{CD}` environment be sufficient? `{CD}` does not supply diagonal arrows, which would make it much, much easier to write than something like `\xymatrix` or Paul Taylor's package.

A note to other KaTeX contributors. I'm pretty booked up right now and I do not volunteer to make the additions to `array.js` that would be necessary to parse the `{CD}` environment. But if someone else takes that on, I can supply SVGs for vertical arrows and edit `stretchy.js` for the rendering of those arrows.

aziis98:
https://codepen.io/aziis98/pen/KJYqZV

I tried making a prototype of this (at the moment this is all an enormous hack). I wanted to see if it was easily possible to parse the xymatrix syntax and compute all the arrow positions and sizes.
I don't really know if it is possible to create a parser plugin for Katex that does more then adding some simple macros but this may be a starting point.

### Notes on the Code

I tried to comment the code as better as I could. For now I only support simple arrows "->" and double ones "=>" and the support for arrow labels was just too bad to include here (it is commented out).

The parsing part completely bypasses Katex's parsing and macro system (but I think that it might be relatively simple to convert a Katex Macro/Function - for now I just splitting strings and call Katex's "render(expression, node)" on each sub expression)

Probability the most useful part might be the arrow creation. I compute the position of start & end of the arrow and it's angle and then rotate and translate it's container with CSS. For porting to Katex one would just need a function for creating an arrow of given length - I haven't really understood how `stretchy.js` works (?) for doing this.

ronkok:
Hi @aziis98. This looks really interesting. I noticed that you had some very insightful comments:

```
// Without the node already in the DOM it looks like
// it doesn't work. Probably because I can't compute
// all the bounding boxes necessary for this to work.
// Tecnically Katex already knows all the bounds
// of its objects (?)
```

[...]

PR #2396, opened 2020-08-09
feat: Support {CD}

ronkok:
This PR implements the AMS `{CD}` environment and therefore it is a partial response to #1834. Authors really want something more capable than `{CD}`, but this is doable now.  Here's a screenshot showing the various arrows and labels.
CD

~~I have not yet done type annotations. If PR #2326 is going to merge soon, I may wait and try my hand at TypeScript instead of Flow.~~


ronkok:
I have now added type annotations to this PR. It ought to be ready for review. Unfortunately, I have somehow also merged the linebreak branch (PR #2175) into this PR. I'll see if I can find a way to separate them. I don't enjoy rebase.

Or, if someone wants to review both in one go, that would be great.

ronkok:
Okay, I've added type annotations and resolved my little merge problem. This PR is ready for review.

src/environments/cd.js
```diff
@@ -0,0 +1,262 @@
...
+                    // Get the arrow labels.
+                    for (let k = j + 2; k < nodes.length; k++) {
+                        if (isLabelEnd(nodes[k], arrowChar)) {
+                            j = k;
+                            break;
+                        }
+                        firstLabel.body.push(nodes[k]);
+                    }
+                    for (let k = j + 1; k < nodes.length; k++) {
+                        if (isLabelEnd(nodes[k], arrowChar)) {
```
kevinbarabash, line 87: What happens if `isLabelEnd` never return true?  Is that valid with the `CD` environment?

src/environments/cd.js
```diff
@@ -0,0 +1,262 @@
...
+                        break;
+                    default:
+                        arrow = {type: "textord", text: " ", mode: "math"};
+                }
+                arrow = {
+                    type: "styling",
+                    body: [arrow],
+                    mode: "math",
+                    style: "display",
+                };
```
kevinbarabash, line 146: There's a lot going on in these `for` loops.  This seems like a good candidate to extract into a helper function.

src/environments/cd.js
```diff
@@ -0,0 +1,262 @@
...
+        type: "array",
+        mode: "math",
+        body,
+        arraystretch: 1,
+        addJot: true,
+        rowGaps: [null],
+        cols,
+        colSeparationType: "CD",
+        hLinesBeforeRow: new Array(body.length + 1).fill([]),
+        isCD: true,
```
kevinbarabash, line 183: `colSeparationType` and `isCD` seem redundant.  Is there anyway to use `CD` as the `colSeparationType` outside of the `CD environment?

src/environments/cd.js
```diff
@@ -0,0 +1,262 @@
...
+    };
+}
+
+// The function below are not available for general use.
+// They are here only for internal use by the {CD} environment in placing labels
+// next to vertical arrows.
```
kevinbarabash, line 189: The reason we don't have special code for positioning labels above/below horizontal arrows is that the existing arrow commands already support that so we reuse that existing functionality, right?  If that's the case, it might be worth adding a comment about that.

src/stretchy.js
```diff
@@ -54,11 +54,15 @@ const stretchyCodePoint: {[string]: string} = {
     xrightleftarrows: "\u21c4",
     xrightequilibrium: "\u21cc",  // Not a perfect match.
     xleftequilibrium: "\u21cb",   // None better available.
+    cdrightarrow: "\u2192",
+    cdleftarrow: "\u2190",
+    cdlongequal: "=",
 };
 
 const mathMLnode = function(label: string): mathMLTree.MathNode {
+    const key = label.replace(/^\\+/, "");
```
kevinbarabash, line 63: I think we could run into an issue here if someone defines a macro like `\\cdrightarrow` which wouldn't overwrite `\\\\cdrightarrow`, but both would have the same `key` here.

src/stretchy.js
```diff
@@ -182,7 +189,7 @@ const svgSpan = function(
         height: number,
     } {
         let viewBoxWidth = 400000;  // default
-        const label = group.label.substr(1);
+        const label = group.label.replace(/^\\+/, "");
```
kevinbarabash, line 192: This may have the same issue as `key` up above.

kevinbarabash (commented):
These diagrams are really cool!  Thanks for adding support for `CD`.  I'm not very familiar with this environment so forgive any silly questions in the comments.

I think it would be good to have some unit test cases that exercise some the failure cases.  Also, having at least one snapshot test for the MathML that's generated would be good too.

src/environments/cd.js
```diff
@@ -0,0 +1,262 @@
...
+    parser.gullet.beginGroup();
+    while (true) {  // eslint-disable-line no-constant-condition
+        // Get the parse nodes for the next row.
+        parseNodes.push(parser.parseExpression(false, "\\\\"));
+        parser.gullet.endGroup();
+        parser.gullet.beginGroup();
+        const next = parser.fetch().text;
+        if (next === "&" || next === "\\\\") {
+            parser.consume();
+        } else if (next === "\\end") {
```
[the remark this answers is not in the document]
edemaine: I believe [`\end` does that already](https://github.com/KaTeX/KaTeX/blob/8c08a3c8a9e886585d1d86fa612f29e2787b195b/src/functions/environment.js#L46); you don't need to do so within a specific environment.

src/environments/cd.js
```diff
@@ -0,0 +1,262 @@
...
+const isLabelEnd = (node: AnyParseNode, endChar: string): boolean => {
+    return ((node.type === "mathord" || node.type === "atom") &&
+        node.text === endChar);
+};
+
+export function parseCD(parser: Parser): ParseNode<"array"> {
+    // Get the array's parse nodes with \\ temporarily mapped to \cr.
+    const parseNodes = [];
+    parser.gullet.beginGroup();
+    parser.gullet.macros.set("\\cr", "\\\\\\relax");
```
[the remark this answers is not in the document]
edemaine: Yes, `beginGroup` starts a new group which the definition will be limited to (because [the last argument of `set` defaults to `false`](https://github.com/KaTeX/KaTeX/blob/8c08a3c8a9e886585d1d86fa612f29e2787b195b/src/Namespace.js#L91)). A similar definition is made in the array environment.

ronkok:
@kevinbarabash Thank you for an excellent review. It really helped me to clarify what the code is doing. I've picked up comments in the code but I have not yet written the new tests. More to come. 

Also, there's a good description of `CD` at https://www.jmilne.org/not/Mamscd.pdf.

docs/supported.md
```diff
@@ -83,17 +83,24 @@ $( \big( \Big( \bigg( \Bigg($ `( \big( \Big( \bigg( \Bigg(`
...
 
 </div>
 
-KaTeX also supports `darray`, `dcases`, and `drcases`, which apply `displaystyle`.
+KaTeX also supports:
+* `darray`, `dcases`, and `drcases`, which render in `displaystyle`
+* `align*`, `alignat*`, `gather*`, and `equation*`, which omit equation numbers
+* `aligned`, `alignedat`, and `gathered`, which LaTeX will render while in math mode
+
+Environments `align`, `alignat`, `CD`, `equation`, and `gather` (and their starred versions) may be called from display mode. This usually means that they may be called from within `$$…$$` delimiters or `\[…\]` delimiters. They may not be called from within `$…$` or `\(…\)` delimiters. In the KaTeX auto-render extension, as in LaTeX, these environments are themselves are a way to change from text mode to math display mode, and `\[…\]` delimiters are not necessary.
```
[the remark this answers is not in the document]
ronkok: PR #2700 revises this section, so there are merge conflicts to come. But its worth it.

src/environments/array.js
```diff
@@ -1016,6 +1024,20 @@ defineEnvironment({
...
+    names: ["CD"],
+    props: {
+        numArgs: 0,
+    },
+    handler(context) {
+        validateAmsEnvironmentContext(context);
+        return parseCD(context.parser);
+    },
+    htmlBuilder,
+    mathmlBuilder,
```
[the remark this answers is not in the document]
ronkok: I had to be careful to make this PR emit a parse tree that looks just like the one from `parseArray`. That constrained some of the parse nodes.

src/environments/cd.js
```diff
@@ -0,0 +1,262 @@
...
+    "=": "\\\\cdlongequal",
+    "A": "\\uparrow",
+    "V": "\\downarrow",
+    "|": "\\Vert",
+    ".": "no arrow",
+};
+
+const newCell = () => {
+    return {type: "styling", body: [], mode: "math", style: "display"};
+};
```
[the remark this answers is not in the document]
ronkok: All the values are fixed. Flow can infer the types.

src/environments/cd.js
```diff
@@ -0,0 +1,262 @@
...
+
+        for (let j = 0; j < nodes.length; j++) {
+            if (!(nodes[j].type === "textord" && nodes[j].text === "@")) {
+                cell.body.push(nodes[j]);
+            } else {
+                // Parse node j is an "@", the start of an arrow.
+                row.push(cell);
+                const arrowChar = assertSymbolNodeType(nodes[j + 1]).text;
+                const firstLabel = {type: "ordgroup", mode: "math", body: []};
+                const secondLabel = {type: "ordgroup", mode: "math", body: []};
```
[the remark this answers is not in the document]
ronkok: Flow can infer the types here, too. All the values are fixed.

src/environments/cd.js
```diff
@@ -0,0 +1,262 @@
...
+};
+
+const isLabelEnd = (node: AnyParseNode, endChar: string): boolean => {
+    return ((node.type === "mathord" || node.type === "atom") &&
+        node.text === endChar);
+};
+
+export function parseCD(parser: Parser): ParseNode<"array"> {
+    // Get the array's parse nodes with \\ temporarily mapped to \cr.
+    const parseNodes = [];
```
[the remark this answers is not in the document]
ronkok: Good idea. Done.

src/environments/cd.js
```diff
@@ -0,0 +1,262 @@
...
+                // Parse node j is an "@", the start of an arrow.
+                row.push(cell);
+                const arrowChar = assertSymbolNodeType(nodes[j + 1]).text;
+                const firstLabel = {type: "ordgroup", mode: "math", body: []};
+                const secondLabel = {type: "ordgroup", mode: "math", body: []};
+                if ("=|.".indexOf(arrowChar) > -1) {
+                    j += 1;
+                } else if ("<>AV".indexOf(arrowChar) > -1) {
+                    // Get the arrow labels.
+                    for (let k = j + 2; k < nodes.length; k++) {
```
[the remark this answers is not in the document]
ronkok: Take a look at the revised code. I've it a bit and taken much more care to explain what is happening.

src/environments/cd.js
```diff
@@ -0,0 +1,262 @@
...
+                };
+                row.push(arrow);
+                cell = newCell();
+            }
+        }
+        if (i % 2 === 0) {
+            row.push(cell);
+        } else {
+            row.shift();
+        }
```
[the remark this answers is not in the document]
ronkok: That is a bit cryptic, isn't it? Comments are done..

src/environments/cd.js
ronkok: Bad things happen. AMS LaTeX writes many unhelpful comments. I've just now figured out how to trap for this condition and write a good ParseError. Should be better now.

src/environments/cd.js
ronkok: Done.

src/environments/cd.js
ronkok: I did have a reason to have them both, but it was a bad reason. I've eliminated `isCD`. I have to keep `colSeparationType` for the MathML builder.

src/environments/cd.js
ronkok: Correct, and done.

src/stretchy.js
ronkok: Ah yes, I didn't notice that. Fixed now.

test/__snapshots__/mathml-spec.js.snap
```diff
@@ -204,6 +204,211 @@ exports[`A MathML builder normal spaces render normally 1`] = `
 </math>
 `;
 
+exports[`A MathML builder should build the CD environment properly 1`] = `
+<math xmlns="http://www.w3.org/1998/Math/MathML"
+      display="block"
+>
+  <semantics>
+    <mtable rowspacing="0.24999999999999992em"
```
kevinbarabash, line 212: nit: why not `0.25em`?

test/__snapshots__/mathml-spec.js.snap
```diff
@@ -204,6 +204,211 @@ exports[`A MathML builder normal spaces render normally 1`] = `
...
+          <mstyle scriptlevel="0"
+                  displaystyle="true"
+          >
+            <mrow>
+              <mrow>
+                <mstyle displaystyle="false"
+                        scriptlevel="1"
+                >
+                  <mpadded width="0"
+                           lspace="-1width"
```
kevinbarabash, line 273: Is `width` supposed to be a number here?

test/__snapshots__/mathml-spec.js.snap
```diff
@@ -204,6 +204,211 @@ exports[`A MathML builder normal spaces render normally 1`] = `
...
```
kevinbarabash, line 322: same

test/katex-spec.js
```diff
@@ -2809,6 +2809,25 @@ describe("AMS environments", function() {
...
+describe("The CD environment", function() {
+    it("should fail if not is display mode", function() {
+        expect(`\\begin{CD}A @X<a<< B @>>b> C @>>> D\\\\@. @| @AcAA @VVdV \\\\@. E @= F @>>> G\\end{CD}`).not.toParse(
+            new Settings({displayMode: false})
+        );
+    });
+    const displaySettings = new Settings({displayMode: true});
+    it("should fail if the character after '@' is not in <>AV=|.", function() {
+        expect(`\\begin{CD}A @X<a<< B @>>b> C @>>> D\\\\@. @| @AcAA @VVdV \\\\@. E @= F @>>> G\\end{CD}`).not.toParse(displaySettings);
+    });
```
kevinbarabash, line 2821: It would be good to have a successfully test here as well.  It looks like the display mode test also has an issue with the `X` coming after `@`.  Having a successfully test case would help when determine which part of the other test cases are causing them not parse.

src/environments/cd.js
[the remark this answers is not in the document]
kevinbarabash: I've been using TypeScript for another project recently and it requires more of the types to be explicitly specified.

src/environments/cd.js
```diff
@@ -0,0 +1,317 @@
...
+        // The parent gets position: relative. The child gets position: absolute.
+        // So CSS can locate the label correctly.
+        const parent = buildCommon.wrapFragment(
+            html.buildGroup(group.fragment, options), options
+        );
+        parent.classes.push("cd-vert-arrow");
+        return parent;
+    },
+    mathmlBuilder(group, options) {
+        return  new mathMLTree.MathNode("mrow",
```
kevinbarabash, line 314: super style nit: extra space

kevinbarabash (approved):
Thanks for refactoring `parseCD` and adding all sorts of detailed comments.  It makes it very easy to understand what's going on. 🎉

test/__snapshots__/mathml-spec.js.snap
ronkok: It's a computed number from https://github.com/KaTeX/KaTeX/blob/master/src/environments/array.js#L514-L519.

I'll add code that will round it to 4 places after the decimal.

test/__snapshots__/mathml-spec.js.snap
ronkok: It's correct as is. The "-" prefix tells a [lspace](https://developer.mozilla.org/en-US/docs/Web/MathML/Element/mpadded#attr-lspace) attribute that the value is an increment. And `lspace` accepts "1width" as a variable, which is a good thing because we don't know the numeric width.

So this `<mpadded>` creates a right-justified label.

test/katex-spec.js
ronkok: Will do.

test/__snapshots__/mathml-spec.js.snap
kevinbarabash: I think it's fine to leave this as is for now and handle it as part of #2460.

kevinbarabash:
This looks good to go.  I'll merge it once the automated checks complete.

edemaine:
Awesome work guys! Very cool to have commutative diagrams of some form in KaTeX!

PR #2400, opened 2020-08-10
ci: fix Dependabot autofix

ylemkimon:
It turns out that `@actions/checkout` fetches with the depth 1, so it cannot get `HEAD^`.

Also renamed `dependabot.js.yml` to `dependabot.yml` (`.js` was in the filename because I used `node.js.yml` template).


ylemkimon:
@kevinbarabash It works: #2392 :smile: Thank you for the review!

PR #2405, opened 2020-08-10
docs: add link to Discussions

ylemkimon:


ylemkimon:
@kevinbarabash I've added two more links.

ylemkimon:
I wonder Gitter has a migration feature. Spectrum.chat, for instance, supports redirecting community to another page.

PR #2417, opened 2020-08-12
ci: migrate to GitHub Actions from CircleCI, allow running Browserstack on forked repo via label

ylemkimon:
From #2415:
> Pros:
> * Nice integration with GitHub
> * Single codebase/infrastructure
> * Able to test on MacOS and Windows
> * Easier to view results
> * Annotation support
> * Doesn't have problem with forked repo enabling CI
> 
> Cons:
> * Not possible to reuse steps. The amount of code duplication is a bit high (https://github.com/actions/runner/issues/646)
> * Harder to view artifacts, e.g., screenshot diffs and new screenshots. They are zipped and have to be downloaded
> * Cannot skip if code is not changed, with required status checks enabled (but they only take <2 min. and GitHub Actions hour is unlimited)
> * [ci skip] or [skip ci] doesn't work for pull requests

- [ ] Disable CircleCI
- [x] Disable Screenshotter bot
- [ ] Remove Screenshotter bot GitHub App
- [x] Remove CircleCI from required status checks
- [x] Add GitHub Actions to required status checks
- [ ] Update other pull requests


.github/workflows/ci.yml
```diff
@@ -0,0 +1,130 @@
+name: CI
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
```
kevinbarabash, line 7: Should we be using `pull_request_target` instead of `pull_request` given the there are secrets in this file?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,130 @@
...
+
+    - name: Run Flow
+      run: yarn test:flow
+      if: always()
+
+    - name: Run Jest tests
+      run: yarn test:jest --coverage
+      if: always()
+
+    - uses: codecov/codecov-action@v1
```
kevinbarabash, line 50: For some reason I thought that this action required a secret of some sort to work correctly.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,130 @@
...
+      if: matrix.browserstack && github.repository_owner == 'KaTeX'
+      env:
+        BROWSERSTACK_USER: ${{ secrets.BROWSERSTACK_USER }}
+        BROWSERSTACK_ACCESS_KEY: ${{ secrets.BROWSERSTACK_ACCESS_KEY }}
+
+    - uses: actions/upload-artifact@v2
+      if: failure()
+      with:
+        name: new
+        path: test/screenshotter/new
```
kevinbarabash, line 124: I think it'd be good to have one commit in this PR that temporarily introduces a change so that this can be tested.

.circleci/config.yml
```diff
@@ -1,146 +0,0 @@
...
-      - when:
-          condition: << parameters.browserstack >>
-          steps:
-            - screenshot:
-                flags: --browserstack --selenium-capabilities '<< parameters.browserstack >>'
-      - unless:
-          condition: << parameters.browserstack >>
-          steps:
-            - screenshot:
-                flags: --selenium-ip localhost
```
kevinbarabash, line 114: Now I know why I couldn't find documentation for `when` and `unless`... I was looking at the GitHub actions docs instead of the circleci docs. 🙃

kevinbarabash:
Why did the coverage report change so drastically?  I wouldn't expect any change in those numbers from this PR.

.github/workflows/ci.yml
ylemkimon: @kevinbarabash AFAIK, secrets aren't and shouldn't be passed to forked repository.

.github/workflows/ci.yml
ylemkimon: @kevinbarabash Codecov works tokenless on the public repo. It seems they have verification logic on their side, checking whether it's from real CI job.

.github/workflows/ci.yml
ylemkimon: @kevinbarabash I've opened #2418.

ylemkimon: You're right. Artifacts with the same name from different jobs overwrite each other. I'll update the PR.

dockers/screenshotter/screenshotter.js
```diff
@@ -277,6 +281,8 @@ function buildDriver() {
         ffProfile.setPreference("browser.startup.page", 0);
         const ffOptions = new firefox.Options().setProfile(ffProfile);
         builder.setFirefoxOptions(ffOptions);
+    } else if (opts.browser == "chrome") {
```
ylemkimon, line 284:
```suggestion
    } else if (opts.browser === "chrome") {
```

ylemkimon:
@kevinbarabash 
> Why did the coverage report change so drastically? I wouldn't expect any change in those numbers from this PR.

I'm not sure. Maybe because the branch is not up-to-date.

ylemkimon:
It seems the codecov-action is uploading `coverage/coverage-final.json`, which was not uploaded with codecov-node. This file seems to have different line counts, so coverage seems to have changed.

.github/workflows/ci.yml
ylemkimon: It's now possible to run Browserstack tests, e.g., Safari screenshotter, upon maintainer's approval via adding `test screenshots` label.

kevinbarabash:
@ylemkimon do you need my help with any of the checkbox items at the top?

ylemkimon:
@kevinbarabash No, they are to-dos after this PR is merged.

kevinbarabash:
In that case I better review this. 🙂

.github/workflows/ci.yml
kevinbarabash: That's super cool.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,172 @@
...
+    branches: [ master ]
+  pull_request_target:
+    branches: [ master ]
+    types: [ labeled ]
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    if: |
+      github.event_name != 'pull_request_target' &&
```
kevinbarabash, line 16: This is to avoid running tests twice for every pull request, right?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,172 @@
...
+    - uses: codecov/codecov-action@v1
+      with:
+        directory: ./coverage/
+
+  screenshotter_dispatcher:
+    runs-on: ubuntu-latest
+    if: |
+      (github.event_name != 'pull_request_target' ||
+        (github.event.pull_request.head.repo.full_name != 'KaTeX/KaTeX' &&
+          contains(github.event.pull_request.labels.*.name, 'test screenshots'))) &&
```
kevinbarabash, line 63: This is so that we use `pull_request_target` for PRs from forks and `pull_request` for PRs from `KaTeX/KaTeX`, right?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,172 @@
...
+      !contains(toJSON(github.event.commits.*.message), '[skip ci]') &&
+      !contains(toJSON(github.event.commits.*.message), '[ci skip]')
+    outputs:
+      matrix: ${{ steps.set-matrix.outputs.result }}
+    
+    steps:
+    - id: set-matrix
+      uses: actions/github-script@v2
```
kevinbarabash, line 71: This docs for this action say that it's possible to require external scripts.  If we did that we could lint the script.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,172 @@
...
+
+    steps:
+    - uses: actions/checkout@v2
+      if: github.event_name != 'pull_request_target'
+      with:
+        submodules: recursive
+    - uses: actions/checkout@v2
+      if: github.event_name == 'pull_request_target'
+      with:
+        ref: refs/pull/${{ github.event.pull_request.number }}/head
```
kevinbarabash, line 131: Why do we have to specify the ref when using `pull_request_target`?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,172 @@
...
+
+    - name: Cache dependencies
+      uses: actions/cache@v2
+      with:
+        path: |
+          .yarn/cache
+          .pnp.js
+        key: yarn-deps-v1-${{ hashFiles('yarn.lock') }}
+        restore-keys: |
+          yarn-deps-v1-
```
kevinbarabash, line 147: It's too bad there isn't a way to share steps between jobs.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,172 @@
...
+      if: matrix.browserstack
+      env:
+        BROWSERSTACK_USER: ${{ secrets.BROWSERSTACK_USER }}
+        BROWSERSTACK_ACCESS_KEY: ${{ secrets.BROWSERSTACK_ACCESS_KEY }}
+
+    - uses: actions/upload-artifact@v2
+      if: failure()
+      with:
+        name: new-${{ matrix.browser }}
+        path: test/screenshotter/new
```
kevinbarabash, line 166: Hopefully GitHub will provide a way to browse artifacts in the future.

dockers/screenshotter/screenshotter.js
```diff
@@ -153,7 +154,10 @@ function guessDockerIPs() {
     }
     // Native Docker on Linux or remote Docker daemon or similar
     const gatewayIP = cmd("docker", "inspect",
-      "-f", "{{.NetworkSettings.Gateway}}", opts.container);
+        "-f", "{{.NetworkSettings.Gateway}}", opts.container)
+      || cmd("docker", "inspect",
+        "-f", "{{range .NetworkSettings.Networks}}{{.Gateway}}{{end}}",
+        opts.container);
```
kevinbarabash, line 160: Why do we need two different commands now?

dockers/screenshotter/screenshotter.js
```diff
@@ -277,6 +281,8 @@ function buildDriver() {
         ffProfile.setPreference("browser.startup.page", 0);
         const ffOptions = new firefox.Options().setProfile(ffProfile);
         builder.setFirefoxOptions(ffOptions);
+    } else if (opts.browser === "chrome") {
+        builder.setChromeOptions(new chrome.Options().setProxy(null));
```
kevinbarabash, line 285: This seems reasonable, but did something necessitate this change?

.github/workflows/ci.yml
ylemkimon: @kevinbarabash Yes, `test` doesn't require secrets so it can be run on `pull_request` event.

.github/workflows/ci.yml
ylemkimon: @kevinbarabash Yes, PRs from fork run Safari screenshotter on `pull_request_target` `labeled`.

.github/workflows/ci.yml
ylemkimon: @kevinbarabash `pull_request_target` is run [in the context of the base repository of the pull request](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#pull_request_target), so the default ref is `master` branch, not the head of the PR.

.github/workflows/ci.yml
ylemkimon: @kevinbarabash The repo has to be checked out in order to run the script on the repo. However, this job doesn't check out the repo, and adding `checkout` will increase the time by about 5 seconds, where this job takes only 1-2 seconds.

.github/workflows/ci.yml
ylemkimon: @kevinbarabash I think it'll be added soon. GitHub is already working on a PR: https://github.com/actions/runner/issues/646.

dockers/screenshotter/screenshotter.js
ylemkimon: @kevinbarabash If Docker is configured to use the non-default network interface, the IP address of the gateway is available at not `.NetworkSettings.Gateway` but `.NetworkSettings.Networks.*.Gateway`.

dockers/screenshotter/screenshotter.js
ylemkimon: @kevinbarabash This was in attempt to fix Chrome screenshotter taking too long (~20 mins) occasionally. [Some posts](https://stackoverflow.com/questions/58606020/selenium-webdriver-chromedriver-slow-to-launch-why) suggested it's a proxy issue, so I tried disabling proxy, but it turns out the issue is still reproducible. Then I've looked at Docker logs and it showed `Timed out receiving message from renderer`. [This post](https://stackoverflow.com/questions/48450594/selenium-timed-out-receiving-message-from-renderer) suggested adding `--disable-gpu` argument, and the issue is not reproducible for about 15 retries.

.github/workflows/ci.yml
kevinbarabash: I didn't realize that.  That's good to know.

.github/workflows/ci.yml
kevinbarabash: That makes sense.  Could you add a quick comment here saying that?

dockers/screenshotter/screenshotter.js
kevinbarabash: Can you add a comment saying that here before merging this PR?

.github/workflows/ci.yml
ylemkimon: @kevinbarabash I'm creating a ESLint plugin, which lints JavaScript in GitHub Actions workflow. I'll enable it later.

ylemkimon:
@kevinbarabash Thank you for the review! I've confirmed they work as intended in #2429. Could you stop CircleCI builds in https://app.circleci.com/settings/project/github/KaTeX/KaTeX?

kevinbarabash:
I've stopped the CircleCI builds.

edemaine:
Exciting! Nice work.

PR #2424, opened 2020-08-13
ci: enable Dependabot for website, submodules, and GitHub Actions

ylemkimon:
Dependabot turned out to be quite nice at keeping dependencies up to date and pull requests seem to be manageable. This enables it for three more: website, submodules (katex-fonts), and GitHub Actions.


kevinbarabash:
It'll be interesting to see how many dependabot PRs this results in.

.github/dependabot.yml
```diff
@@ -7,3 +7,17 @@ updates:
...
       interval: "daily"
+  - package-ecosystem: "npm"
+    directory: "/website/"
+    schedule:
+      interval: "daily"
+
+  - package-ecosystem: "gitsubmodule"
+    directory: "/" # Location of .gitmodules
+    schedule:
+      interval: "daily"
```
kevinbarabash, line 18: Will this try to update the submodules to `master`?

.github/dependabot.yml
```diff
@@ -7,3 +7,17 @@ updates:
...
+
+  - package-ecosystem: "github-actions"
+    directory: "/" # Location of .github
```
kevinbarabash, line 21: Thanks for adding these comments.

.github/dependabot.yml
ylemkimon: @kevinbarabash Yes, it'll.

PR #2425, opened 2020-08-14
Discussions

ylemkimon:


ylemkimon:
This PR was merged due to a workflow error. I've made changes to prevent workflows overriding branch protection and this won't happen again in the future. I apologize for the mess.

ylemkimon:
@edemaine From https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/about-pull-request-reviews:
> When several open pull requests each have a head branch pointing to the same commit, you won’t be able to merge them if one or both have a pending or rejected review.

image

It seems a pull request review is associated with the commit. i.e., if another PR containing a commit was approved, all PRs with the same commit as head is assumed to be approved. These commits were in #2405.

kevinbarabash:
@ylemkimon if there's a commit we want to undo I think doing a revert commit is probably the best way forward.

ylemkimon:
@kevinbarabash 69be3362fb128215f0aa600d3a78d7be47b0364b is an empty commit, so if you're okay with the commit history, I think we can leave it as is.

kevinbarabash:
I see.  I think it's fine to leave it.

PR #2438, opened 2020-08-18
ci: use personal access token in Dependabot autofix

ylemkimon:
A commit pushed using the default `GITHUB_TOKEN` will not create a new workflow run, i.e., run tests.

cf. https://docs.github.com/en/actions/reference/events-that-trigger-workflows#triggering-new-workflows-using-a-personal-access-token

### DIFF
--- a/.github/workflows/dependabot.yml
+++ b/.github/workflows/dependabot.yml
@@ -12,6 +12,7 @@ jobs:
     - uses: actions/checkout@v2
       with:
         fetch-depth: 2
+        token: ${{ secrets.GH_TOKEN }}
     - name: Use Node.js 12.x
       uses: actions/setup-node@v1
       with:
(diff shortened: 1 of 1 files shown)


.github/workflows/dependabot.yml
```diff
@@ -37,11 +39,19 @@ jobs:
 
         # deduplicate lockfile
         yarn deduplicate
+      env:
+        YARN_ENABLE_SCRIPTS: 0 # disable postinstall scripts
+    - name: Config Git
+      run: |
+        # use personal access token to allow triggering new workflow
```
kevinbarabash, line 46: Should this be `to avoid triggering a new workflow`?

ylemkimon: @kevinbarabash No, that is the exact purpose. Currently, autofix commits don't run the tests automatically.

ylemkimon:
Closed in favor of Renovate.

PR #2444, opened 2020-08-21
build: make vscode work with PnP

ylemkimon:
These files are generated using [Editor SDKs](https://yarnpkg.com/advanced/editor-sdks).


ylemkimon:
@kevinbarabash I've added `flow-for-vscode` support to Yarn in https://github.com/yarnpkg/berry/pull/1732, so Yarn needs to be updated before running the command.

PR #2450, opened 2020-08-21
ci: don't persist credentials and don't run scripts

ylemkimon:


kevinbarabash (approved):
Thanks for making these scripts more secure.

PR #2452, opened 2020-08-23
feat: Support \vcenter and \hbox

ronkok:
This PR adds support for `\vcenter` ~~`\lower`, and `\raise`~~.  It provides each function in two versions. 

The strict version follows TeX syntax and can act only on a box. The function `\hbox` is provided for this purpose. As in TeX, `\hbox` puts its contents into text mode and textstyle.

The non-strict version ~~of each function~~ follows MathJax syntax and will accept any math or text as an argument. It does not change the argument’s style.

`\hbox` is a valid stand-alone function. By itself, it doesn’t do anything except prevent a soft line break.

~~`\raise` and `\lower` do what you would expect. The first argument is a size and the second argument is what gets raised.~~

`\vcenter` sets the vertical alignment of its argument to center on the math axis. This is handy for something within extensible delimiters because the delimiters will extend symmetrically to the math axis. Thus, `\vcenter` can be used to change the first fraction below into the second one:

vcenter

This PR is a refresh on earlier PR #1167, which was set aside until a `strict` setting was available.


src/functions/raisebox.js
```diff
@@ -17,31 +42,41 @@ defineFunction({
...
-        node.setAttribute("voffset", dy);
-        return node;
+    handler({parser, funcName}, args) {
+        const amount = assertNodeType(args[0], "size").value;
+        const body = args[1];
+        return {
+            type: "raisebox",
+            mode: parser.mode,
+            funcName,
+            dy: amount,
```
kevinbarabash, line 76: If the following change is made here we won't have to repeat the logic in both the HTML and MathML builders.
```suggestion
            dy: funcName === "\\lower" ? -amount : amount,
```

src/functions/raisebox.js
```diff
@@ -8,7 +8,32 @@ import {calculateSize} from "../units";
...
 
-// Box manipulation
+// \raisebox, \raise, and \lower
+
+const htmlBuilder = (group, options) => {
+    const body = html.buildGroup(group.body, options);
+    let dy = calculateSize(group.dy, options);
+    if (group.funcName === "\\lower") {
+        dy *= -1;
+    }
```
kevinbarabash, line 18: If we do the `\lower` check in the handler this can be simplified to:
```suggestion
    const dy = calculateSize(group.dy, options);
```

src/functions/raisebox.js
```diff
@@ -8,7 +8,32 @@ import {calculateSize} from "../units";
...
+    }, options);
+};
+
+const mathmlBuilder = (group, options) => {
+    const node = new mathMLTree.MathNode(
+        "mpadded", [mml.buildGroup(group.body, options)]);
+    let dy = group.dy.number;
+    if (group.funcName === "\\lower") {
+        dy *= -1;
+    }
```
kevinbarabash, line 32: If we do the `\lower` check in the handler, this can be simplified to:
```suggestion
    const dy = group.dy.number;
```

src/functions/hbox.js
```diff
@@ -0,0 +1,38 @@
...
+    },
+    handler({parser}, args) {
+        return {
+            type: "hbox",
+            mode: parser.mode,
+            body: ordargument(args[0]),
+        };
+    },
+    htmlBuilder(group, options) {
+        const elements =  html.buildExpression(group.body, options, false);
```
kevinbarabash, line 30: nit
```suggestion
        const elements = html.buildExpression(group.body, options, false);
```

docs/supported.md
```diff
@@ -232,11 +232,14 @@ In display math, KaTeX does not insert automatic line breaks. It ignores display
...
-|$_u^o $ `_u^o `|$\underset{!}{=}$ `\underset{!}{=}` | $$\sum_{\substack{0<i<m\\0<j<n}}$$ `\sum_{\substack{0<i<m\\0<j<n}}`
+|$x_n$ `x_n` |$\stackrel{!}{=}$ `\stackrel{!}{=}`| $a\raisebox{0.25em}{$b$}c$ `a\raisebox{0.25em}{$b$}c`
+|$e^x$ `e^x` |$\overset{!}{=}$ `\overset{!}{=}`  | $M\raise2pt{M^2}M$ `M\raise2pt\hbox{$M^2$}M`
+|$_u^o $ `_u^o `| $\underset{!}{=}$ `\underset{!}{=}` | $a+\left(\vcenter{\frac{\frac a b}c}\right)$ `a+\left(\vcenter{\hbox{$\frac{\frac a b}c$}}\right)`
+|$a \atop b$ `a \atop b` | $M\lower2pt{M^2}M$ `M\lower2pt\hbox{$M^2$}M` | $$\sum_{\substack{0<i<m\\0<j<n}}$$ `\sum_{\substack{0<i<m\\0<j<n}}`
 
-The second argument of `\raisebox` can contain math if it is nested within `$…$` delimiters, as in `\raisebox{0.25em}{$\frac a b$}`
+`\raisebox` and `\hbox` put their argument into text mode. To raise math, nest `$…$` delimiters inside the argument as shown above. 
+
+`\vcenter`, `\raise`, and `\lower` can be written without an `\hbox` if the `strict` rendering option is *false*. In that case, omit the nested `$…$` delimiters.
```
kevinbarabash, line 242: Thanks for adding this comment.

docs/support_table.md
```diff
@@ -1171,8 +1174,9 @@ use `\ce` instead|
 |\vartriangleright|$\vartriangleright$||
 |\varUpsilon|$\varUpsilon$||
 |\varXi|$\varXi$||
-|\vcentcolon|$\vcentcolon$||
-|\vcenter|<span style="color:firebrick;">Not supported</span>||
+|\vcentcolon|$\mathrel{\vcentcolon =}$|`\mathrel{\vcentcolon =}`|
+|\vcenter|$a+\left(\vcenter{\frac{\frac a b}c}\right)$|`a+\left(\vcenter{\hbox{$\frac{\frac a b}c$}}\right)`<br>TeX (strict) syntax|
+|\vcenter|$a+\left(\vcenter{\frac{\frac a b}c}\right)$|`a+\left(\vcenter{\frac{\frac a b}c}\right)`<br>non-strict syntax|
```
kevinbarabash, line 1179: I really like that we're showing both strict and non-strict syntax for these new commands.

kevinbarabash (approved):
LGTM.  Thanks for adding these commands.  It's nice that they build on an existing command.

ronkok:
This PR currently succeeds at rendering `M\raise3pt{$M^2$}M` but it fails at rendering `M\raise3pt\hbox{$M^2$}M`. The second syntax is there to provide `strict` compatibility with LaTeX.

The difference is `\hbox`. I originally was able to make this syntax work by setting `\raise` to be less greedy than `\hbox`. Now that the `greediness` attribute no longer exists, I have been unable to get `\hbox` to work. I've tried to set `primitive` and `allowedInArgument` to reasonable values. No success.

`M\raise3pt\hbox{$M^2$}M` parses in LaTeX. It ought to work in KaTeX. 

@ylemkimon Do you have any ideas on how to make this happen?

ronkok:
Correction: It also fails when the dimension is not braced. That also used to work.

ronkok:
I withdraw `\raise` and `\lower` from this PR. It now supports only `\vcenter` and `\hbox`. 

I could not get a `strict` version of `\raise` or `\lower` to work and I suspect that such support may require parser changes. That's more than I signed up to do for this PR. `\vcenter` is the most valuable part of this PR. I'm glad that it works.

ronkok:
The screenshotter tests aren ow listed as failures, but the diff shows no error. Furthermore, the previous commit showed success for Chrome and Firefox screenshotter tests and this commit made no change to those files.

This seems to me to be a screenshotter error.

ylemkimon:
Issues with size argument will be mitigated with #2139. Sorry for not clearing with existing PRs.

> The screenshotter tests aren ow listed as failures, but the diff shows no error. Furthermore, the previous commit showed success for Chrome and Firefox screenshotter tests and this commit made no change to those files.

The version of browsers used in screenshotter is updated in #2543 and #2544, and there may be subpixel changes.

src/functions/hbox.js
```diff
@@ -0,0 +1,40 @@
...
+
+defineFunction({
+    type: "hbox",
+    names: ["\\hbox"],
+    props: {
+        numArgs: 1,
+        argTypes: ["text"],
+        allowedInText: true,
+        primitive: true,
+        allowedInArgument: true,
```
ylemkimon, line 22:
```suggestion
        allowedInArgument: false,
```
`e^\hbox{2}` or `\sqrt\hbox{2}` don't work.

src/functions/hbox.js
```diff
@@ -0,0 +1,40 @@
...
+import * as mml from "../buildMathML";
+
+// \hbox is provided for compatibility with LaTeX \vcenter.
+// In LaTeX, \vcenter can act only on a box, as in
+// \vcenter{\hbox{$\frac{a+b}{\dfrac{c}{d}}$}}
+// This function by itself doesn't do anything but prevent a soft line break.
+
+defineFunction({
+    type: "hbox",
+    names: ["\\hbox"],
```
ylemkimon, line 16: Does `\hbox` do anything different than `\text` in KaTeX?

ronkok:
Safari screenshotter is not emitting an artifact, so I have no way to update that particular file.

ylemkimon:
Thank you for the PR!

PR #2460, opened 2020-08-26
fix: round dimensions to 4 places

ronkok:
This PR revives PR #1747, which has stalled. It rounds distances in emitted KaTeX HTML to 4 decimal places.

**Edited by @ylemkimon:** closes #3170.


src/mathMLTree.js
```diff
@@ -189,15 +189,15 @@ class SpaceNode implements MathDomNode {
...
+        if (width >= 0.0555 && width <= 0.0556) {
             this.character = "\u200a";           // &VeryThinSpace;
         } else if (width >= 0.1666 && width <= 0.1667) {
             this.character = "\u2009";           // &ThinSpace;
         } else if (width >= 0.2222 && width <= 0.2223) {
             this.character = "\u2005";           // &MediumSpace;
         } else if (width >= 0.2777 && width <= 0.2778) {
             this.character = "\u2005\u200a";     // &ThickSpace;
-        } else if (width >= -0.05556 && width <= -0.05555) {
+        } else if (width >= -0.0556 && width <= -0.0555) {
```
kevinbarabash, line 200: nit: these numbers probably don't need rounding since they don't end up in the markup we produce.

ronkok: No problem. I've reverted these numbers.

ronkok:
> found way more than are addressed in this PR. Those should probably be wrapped in calls to utils.round as well.

No doubt that's true, but I wasn't really trying to reach rounding perfection in this PR. I was just trying to not lose the ground gained by the original PR. I'm hoping that the other dimensions can be left to be done later.

src/utils.js
```diff
@@ -100,6 +100,16 @@ export const protocolFromUrl = function(url: string): string {
...
+/**
+ * Round `n` to 4 decimal places, or to the nearest 1/10,000th em. The TeXbook
+ * gives an acceptable rounding error of 100sp (which would be the nearest
+ * 1/6551.6em with our ptPerEm = 10):
+ * http://www.ctex.org/documents/shredder/src/texbook.pdf#page=69
+ */
+const round = function(n: number): number {
+    return +n.toFixed(4);
+};
+
```
ylemkimon, line 112: I wonder why this file fails to be syntax-highlighted.

ylemkimon:
I think we can create a util function, `makeEm`, which automatically calls `round` and appends `em` unit.

src/utils.js
ronkok: Dunno. It's syntax highlighted when I open it in VS Code, and it's syntax highlighted when I hit the "Edit file" menu selection here on GitHub. But not these two lines in the viewer. Strange.

ylemkimon (approved):

ylemkimon:
Thank you for the PR and sorry for the delayed review. Please find attached the screenshot diffs: [diff.zip](https://github.com/KaTeX/KaTeX/files/7446232/diff.zip). I don't see significant changes, but requesting another set of eyes.

src/units.js
```diff
@@ -95,5 +94,15 @@ export const calculateSize = function(
...
         }
     }
-    return utils.round(Math.min(sizeValue.number * scale, options.maxSize));
+    return Math.min(sizeValue.number * scale, options.maxSize);
+};
+
+/**
+ * Round `n` to 4 decimal places, or to the nearest 1/10,000th em. The TeXbook
+ * gives an acceptable rounding error of 100sp (which would be the nearest
+ * 1/6551.6em with our ptPerEm = 10):
```
edemaine, line 103:
```suggestion
 * 1/6553.6em with our ptPerEm = 10):
```

src/units.js
```diff
@@ -95,5 +94,15 @@ export const calculateSize = function(
...
+ * http://www.ctex.org/documents/shredder/src/texbook.pdf#page=69
```
edemaine, line 104: This argument makes me a little uncomfortable, because that same passage says that TeX uses `1sp` as its unit, which is `1/65536 pt` or `1/655360 em` with our `ptPerEm = 10`, which would suggest we need 6 digits (instead of 4) to "match" TeX. The literal text is "the wavelength of visible light is approximately 100 sp, rounding errors of a few sp make no difference to the eye."  I think 100 sp *could* make a difference to the eye &mdash; that's like moving 1 "eye pixel". 🙂 

Then I wondered about device pixels. `1em` is `1/16 px`, and there are `devicePixelRatio` actual pixels per `px`.  I've typically seen `devicePixelRatio` ~ 2 for 4K displays.  `0.0001 = 1/(16*625)`, so we'd need a `devicePixelRatio` of `625` before 4-digit rounding moved by an entire pixel.  Said another way, with `devicePixelRatio = 2`, we shift things by ~0.3% of a pixel, which probably has minimal impact on antialiasing.

Makes sense: probably print is more sensitive than screens. Let's think about printing a KaTeX document with `1pt` mapped to `1pt` of the page. The highest resolution printer I could find is 5760 dpi, which is around 80 dots per pt, or 800 dots per em (at `ptPerEm = 10`).  This would require 3 digits of precision to get within a dot (roughly).  (I assume anti-aliasing doesn't happen at this resolution?)

Said another way, 4 digits of em precision = 3 digits of pt precision translates to 1000 dots per pt = 72270 dpi.  This does seem like an acceptable loss; even if we zoomed 10x, it's better than all existing printers.

edemaine (approved):
> I see no significant changes and acceptable results for #3170, but requesting another set of eyes.

There are visible (whole-pixel) changes in DisplayStyle-chrome, Gathered-safari, Smash-safari, Sqrt-chrome, Sqrt-firefox, SqrtRoot-chrome.  The change to DisplayStyle-chrome actually brings it into alignment with Firefox, and the change to SqrtRoot-chrome makes the spacing look more uniform.

Nice that this PR now includes/replaces #3170. The change in LowerAccent-safari is a big improvement, but LowerAccent-firefox got slightly worse.

Overall, I think we're getting burned more by the unpredictable nature of browser rounding, not our own precision. So I'm fine with merging this.

I also simulated the effect of this PR on @universemaster's large example. It reduced the minified HTML by 5.7%, which is something, though fairly small.

src/units.js
ylemkimon: I've replaced it with the link to this PR for future reference.

ylemkimon:
Thank you!

> Also note that this should be labeled `BREAKING CHANGE`.

I don't think this would be a breaking change, since this doesn't change how KaTeX is used.

edemaine:
> I don't think this would be a breaking change, since this doesn't change how KaTeX is used.

All inputs generate different outputs, and it's not exactly fixing a bug... I'd say it could break things for people. But too late now. 🙂

PR #2464, opened 2020-08-27
refactor: Delete obsolete comment re: limsup

ronkok:


edemaine:
Was the `test screenshots` label to give the Safari screenshotter a kick? It seemed to stall out, and this got it to restart? Ideally we could improve the reliability... 🙁

ronkok:
Yes, that was exactly the purpose for the `test screenshots` label. I got that tip from @ylemkimon in [PR 2459 comment](https://github.com/KaTeX/KaTeX/pull/2459#issuecomment-680975113). My guess is that Safari screenshots are not automatic so that they do not run with every `caniuse-lite` update and run up the BrowserStack time. But that is just my speculation. @ylemkimon Can you confirm?

ylemkimon:
@edemaine The Safari screenshotter is not run on PRs from a fork by default, due to security implications. Instead, a label is used to trigger it.

ronkok:
That information should probably be written into the [contributing page](https://github.com/KaTeX/KaTeX/blob/master/CONTRIBUTING.md).

edemaine:
Ah, I see. I guess it's finally time to start making branches on KaTeX's KaTeX instead of our personal forks. (Unless we add a whitelist of members' forks.)

PR #2465, opened 2020-08-27
feat: Add a targetable HTML id to \tag

ronkok:
This PR adds an HTML id to the same element to which KaTeX assigns `class="tag"`. This enables an author to write a link elsewhere in the document. It may someday be useful in the implementation of `\ref`.

The id takes the form `eqn-<tagText>`.

There are a couple of exceptions. Any `\tag` that contains math will not get an id. The characters `<>&'"/` will be omitted from the id.

### DIFF
--- a/docs/supported.md
+++ b/docs/supported.md
@@ -220,6 +220,8 @@ $$\tag{hi} x+y^{2x}$$
 `\tag*{hi} x+y^{2x}`
 $$\tag*{hi} x+y^{2x}$$
 
+If the argument to `\tag` or `\tag*` contains all text (no math), KaTeX will insert a targetable HTML id. Example: `tag{hi}` will result in `id="eqn-hi"`.
+
 ### Line Breaks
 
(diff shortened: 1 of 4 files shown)


ronkok:
I have a bad test in this PR. I wrote the Jest test:

```
it("should include a targetable HTML id that omits unsafe characters", () => {
    const built = getBuilt(r`\tag{<hi} x+y`, displayMode);
    expect(built).toMatchSnapshot();
});
```

The resulting snapshot is not what I would expect. It does not include the id. It does not even include the `class="tag"` that has been in KaTeX for some time. I get good results when I run `yarn start` so I don't know what causes the snapshot problem.

ronkok:
MathJax's id has a prefix. It follows the form `mjx-eqn-<tagText>`. I can write a prefix into this tag, too, if that is what others prefer.

pdmosses:
> If the argument to `\tag` or `\tag*` contains all text (no math), KaTeX will insert a targetable HTML id. Example: `tag{hi}` will result in `id="eqn-hi"`.

Automatic insertion of HTML ids could result in non-unique ids, breaking existing links to targets in web pages (and producing non-conforming HTML).

An option to use a prefix other than `eqn-` would allow users to ensure that tag-inserted ids don't clash with other ids on their pages. However, different equations on the same page might have the same tag text, so perhaps there should also be an option to disable automatic insertion of ids altogether?

PR #2477, opened 2020-08-31
build: upgrade Yarn to 2.2.0

ylemkimon:


kevinbarabash (approved):
I'm impressed they were able to yarn executable size so much.  Thanks for upgrading.

Issue #2491, opened 2020-09-10
Zoom levels stopping KaTeX/LaTeX fractions rendering correctly 


AEadie:
Hi, I am new to GitHub and coding generally, so thank you in advance for the assistance. 

**Describe the bug:**
I am building a maths education website that uses LaTeX code heavily. For some unknown reason, all sections of the site have suddenly stopped showing the thin horizontal dividing line in fractions at 100% or 110% browser zoom levels on my PC. What's particularly strange is that at 90% zoom or below, or 125% zoom and above, the fraction lines appear as normal. This issue seems to apply to Chrome and Microsoft browsers, but not Opera. On my Mac or on mobile there are no issues at any zoom levels, however my friend's Mac experiences the same issue at 125% zoom on Chrome. We have not updated KaTeX or any other plugin - there appears to be no cause for this to suddenly become an issue today. 

**(La)TeX code:**
Not relevant (I believe), just standard fractions. 

**Screenshots:**
90% zoom on PC (renders correctly):
![image](https://user-images.githubusercontent.com/71097132/92795858-9d757f80-f3a8-11ea-8496-bf18981dc442.png)

100% zoom on PC (fraction lines disappear):  
![image](https://user-images.githubusercontent.com/71097132/92795901-a8301480-f3a8-11ea-993d-f737be03d467.png)

**Environment:**
 - WP-KaTeX Version: 1.11.0

Many thanks!

edemaine:
I've been noticing this some lately too -- perhaps a recent change to how Chrome implements zoom? 🙁

ylemkimon:
@AEadie @edemaine Could you let me know the version of Chrome?

Also, could you try the following two versions and compare results?
**Windows x64**
https://www.googleapis.com/download/storage/v1/b/chromium-browser-snapshots/o/Win_x64%2F770054%2Fchrome-win.zip?generation=1589885407916667&alt=media
https://www.googleapis.com/download/storage/v1/b/chromium-browser-snapshots/o/Win_x64%2F770069%2Fchrome-win.zip?generation=1589887462069918&alt=media

**Linux x64**
https://www.googleapis.com/download/storage/v1/b/chromium-browser-snapshots/o/Linux_x64%2F770063%2Fchrome-linux.zip?generation=1589884575809384&alt=media
[...]

PR #2516, opened 2020-09-15
ci(renovate): update screenshotter browser, .flowconfig, editor SDKs, and Actions

### LINKED ISSUE #2483: Avoid unnecessary netlify builds
I got an email from netlify saying we'll "need an allowance increase soon".  I think we'll still be under this billing cycle but it would be nice to avoid unnecessary netlify builds if possible, e.g. dependabot builds.

ylemkimon:
Enable Renovate to
- update screenshotter browser image
  * #2491 proved it's important to test on latest browsers
  * Example: https://github.com/ylemkimon/KaTeX/pull/48
- update versions in `.flowconfig` and Yarn 2 editor SDKs
  * Example: https://github.com/ylemkimon/renovate-flow/pull/2
- update GitHub Actions in workflows
  * Turns out https://github.com/KaTeX/KaTeX/pull/2493#discussion_r487341345 was wrong and Renovate updates GitHub Actions more reliably

and fixes #2483, netlify skip.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,9 +1,0 @@
-# Please see the documentation for all configuration options:
-# https://help.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
-
-version: 2
-updates:
-  - package-ecosystem: "github-actions"
-    directory: "/" # Location of .github
-    schedule:
-      interval: "daily"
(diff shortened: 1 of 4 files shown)


kevinbarabash (approved):
That's pretty cool that `renovate` can regex various files in our repo to update versions of things even if the versioning is outside of a standard config file.  Thanks for figuring out that `[skip netlify]` needs to be in the commit message.

PR #2545, opened 2020-09-24
chore: add devcontainer.json

ylemkimon:
Configuration for GitHub Codespaces and Visual Studio Codespaces.


webpack.dev.js
```diff
@@ -19,6 +19,7 @@ katexConfig.devServer = {
     disableHostCheck: true,
     host: '0.0.0.0',
     port: PORT,
+    sockPort: 'location',
```
kevinbarabash, line 22: The docs do say anything about `"location"`.  What does this setting do?

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,14 @@
+{
+  "name": "KaTeX Codespaces",
+  "image": "mcr.microsoft.com/vscode/devcontainers/javascript-node:14",
+  "extensions": [
+    "dbaeumer.vscode-eslint",
+    "flowtype.flow-for-vscode",
+    "stylelint.vscode-stylelint",
+    "arcanis.vscode-zipfs"
```
kevinbarabash, line 8: What is this used for?

kevinbarabash (commented):
I'm excited to give this a try.  Once this PR is merged with the "Open codespace" button just appear in GitHub?

webpack.dev.js
ylemkimon: @kevinbarabash It makes `webpack-dev-server` to use the browser's location port as the socket port: https://github.com/webpack/webpack-dev-server/pull/2341. For instance, when developing locally in `http://localhost:7936`, it uses `7936`, and using GitPod or codespaces, they are tunneled using HTTPS, so it uses `443`. This eliminates the need to set `--sockPort 443` manually.

.devcontainer/devcontainer.json
ylemkimon: @kevinbarabash Yarn 2 uses a zip-based file system, so this extension allows us to browse into a zip file in vscode.

ylemkimon:
Since it's still in limited beta, it can be started from https://github.com/codespaces. Maybe when it's general availability, a badge or button to start it may be added.

kevinbarabash (approved):
Thanks for adding support for Codespaces.

PR #2570, opened 2020-10-07
docs: Fix documentation typo in operatorname*

ronkok:


docs/support_table.md
```diff
@@ -785,7 +785,7 @@ use `\ce` instead|
 |\omicron|$\omicron$||
 |\ominus|$\ominus$||
 |\operatorname|$\operatorname{asin} x$|\operatorname{asin} x|
```
edemaine, line 787: While we're here, could we fix the missing back-ticks in the rightmost column of `\operatorname`?

docs/support_table.md
```diff
@@ -785,7 +785,7 @@ use `\ce` instead|
...
-|\operatorname\*|$\operatorname{asin}\limits_y x$|`\operatorname{asin}\limits_y x`|
+|\operatorname\*|$\operatorname*{asin}\limits_y x$|`\operatorname*{asin}\limits_y x`|
```
edemaine, line 788: This example seems to work just as well without `\limits`. Is it worth including `\limits` in the example? (I'm not sure I understand the difference.)

edemaine (commented):
Thanks for spotting this!

docs/support_table.md
ronkok: Done.

docs/support_table.md
ronkok: > This example seems to work just as well without `\limits`.

To see the difference, you have to be in text mode, not display mode. So the difference will be visible in the documentation page. Let's leave it in.

edemaine:
Great! https://deploy-preview-2570--katex.netlify.app/docs/next/support_table.html now looks perfect.

However, I notice that https://deploy-preview-2570--katex.netlify.app/docs/support_table.html has the bug (unsurprisingly). Given that this is a fix to the documentation (not an update to functionality), and that is the page most people will see for a quite, would you mind updating `website/versioned_docs/version-0.12.0/support_table.md` too? (and possibly older ones, but that one seems most important, as it's the current version)

edemaine (approved):

Issue #2593, opened 2020-10-15
Wrong convertion LaTEX-code to MathML


Futyn-Maker:
**Describe the bug:**
The LATEX-code are converted to MathML wrong, although by visually all OK. The correct MathML is needed to read math formulas by screen-readers.

**To Reproduce:**
1. Try to type the difficult LaTEX-formula which contains, for example, log or trigonometric function. My example is:
```$${\log_{\sqrt{11}} 121}^2$$```
2. see the source MathML-code (not a visual picture of the formula).

**Expected behavior:**
I should get something about this:
```<math display = "block"><msup ><mrow ><msub ><mi>log</mi><msqrt ><mn>11</mn></msqrt></msub><mo>⁡</mo><mn>121</mn></mrow><mn>2</mn></msup></math>```
But I get such MathML:
```<math ><msup ><mrow ><msub ><mo></mo><msqrt ><mn>11</mn></msqrt></msub><mn>121</mn></mrow><mn>2</mn></msup><mrow encoding = "application/x-tex"></mrow></math>```
**Environment (please complete the following information):**
 - KaTeX Version: latest (the problem is relevant even or your demo)
 - Device: laptop, but the problem is relevant for all devices.
 - OS: Windows 10, iOS 14
 - Browser: Chrome, mobile safari
 - Version: Chrome 86, Safari 14

**Additional context:**

The wrong convertion of LaTEX-code breaks an accessibility. I'm sure that the problem is in a KaTEX, because with MathJax these formulas are converted correctly. The problems almost at all formulas with log or trigonometric functions. In some situations the functions are completely disapeared like in the example above. In other situations, for example, there is an excess subscript, and so on.

**Comment by ronkok** [comment]
> the problem is relevant even or your demo
> functions are completely disapeared like in the example above

I cannot reproduce your result in the demo. Your example input is `{\log_{\sqrt{11}} 121}^2`. And you report that a `<mo>` element is empty which should contain `log`. When I examine the output, I find that the same `<mo>` element is 
```
<mo>
   <mi>log</mi><mo>&#x2061;</mo>
<mo>
```

Now one could argue that the KaTeX should not wrap that `<mi>` element with a `<mo>`. But that would be a different discussion. In this discussion, it it incorrect to say that the `<mo>` is empty and the the `log` term is missing.

Futyn-Maker:
@ronkok How do You read output MathML in katex.org demo page?

ronkok:
[...]

PR #2596, opened 2020-10-16
fix: Improve MathML for math operators with subscripts

ronkok:
This PR is an accessibility advance and a MathML visual regression. It relates to math operators like `\log` or `\lim` that are written with subscripts or limits.  In Issue #2593, @Futyn-Maker reports that that the `log` is omitted in three different screen readers. This PR will eliminate that problem.

In so doing, this PR will also introduce a spacing regression for visible rendering viewed in a MathML-capable browser like Firefox. This may concern you, so I'll explain why I did it.

Spacing is usually easier in MathML  than in HTML. You just put content into the proper `<mi>` or `<mo>` elements and the browser does the spacing for you. The exception is math operators like `\log` or `\lim` or anything inside an `\operatorname`. MathML has no direct analogue to a TeX math operator. MathJax puts them into a `<mi>` element, but `<mi>` elements get the  spacing of a variable, not an operator. 

To get the correct spacing, I have previously tried wrapping the `<mi>` inside an `<mo>`. That is what caused issue #2593. I also tried to skip the `<mi>` and place the text directly in an `<mo>`. That caused Firefox to improperly display `\lim\limits`, i.e. MathML `<munder>` element, as if it were a subscript, not a limit. So that does not work either.

So I believe the math operator has to be in a `<mi>` with no `<mo>`. That's the way to enable screen readers to work properly. That is what this PR does.

There is a way to regain proper visual spacing. On the HTML side, the `buildHTML.traverseNonSpaceNodes()` function makes a pass through the HTML and inserts hard spacing elements where necessary. Such a thing could also be done on the MathML side. It might be slightly easier in MathML because spacing elements would be necessary only adjacent to math operators. MathML takes care of the other spacing issues.

Easy or not, it's more than I have time to do right now. I think we should get this accessibility fix into the next release and deal with the visible spacing later.


ronkok:
We now have [confirmation](https://github.com/KaTeX/KaTeX/issues/2593#issuecomment-710767752) that this PR fixes the accessibility problem.

kevinbarabash:
@ronkok thanks for fixing the accessibility issue.  I'm not to concerned about the visual regression.  I think it might be fixable with some CSS, but I'll have to investigate that and we don't really use much CSS with our MathML.

kevinbarabash (approved):
Thanks for fixing this.  The change looks reasonable especially given the changes to the snapshots.  I wish we had more MathML snapshots though since I think our coverage in this area is that good.  I'll spend some time later figuring out what snapshots to add, but this PR can go forward as is.

kevinbarabash:
Since this only makes changes to a `mathmlBuilder` function I think it's safe to merge even if we don't hear back from the Safari screenshotter.

Futyn-Maker:
Many thanks for fixing this problem and sorry for my long silence, I was really busy in last days.

Andrey

PR #2618, opened 2020-10-28
fix: Support Armenian characters

### LINKED ISSUE #2617: No character metrics for Armenian letters
**Describe the bug:**
Discovered by Khan Academy Armenian translators. When `\cancel{\text{g}}` works fine, but putting in a character from armenian alphabet (outside of KaTeX fonts) inside the `\text` breaks it.

https://katex.org/?data=%7B%22displayMode%22%3Atrue%2C%22leqno%22%3Afalse%2C%22fleqn%22%3Afalse%2C%22throwOnError%22%3Atrue%2C%22errorColor%22%3A%22%23cc0000%22%2C%22strict%22%3A%22warn%22%2C%22output%22%3A%22htmlAndMathml%22%2C%22trust%22%3Afalse%2C%22macros%22%3A%7B%22%5C%5Cf%22%3A%22%231f(%232)%22%7D%2C%22code%22%3A%22%5C%5Ccancel%7B%5C%5Ctext%7B%D5%A3%7D%7D%22%7D

However, this works `\cancel{\text{aգ}}`, suggesting that the issue is caused by the lack of vertical metrics for the argument.
As a workaround, one could do something like `\cancel{\text{\vphantom{a}գ}}`, but that's not a viable solution for our translators because changes to LaTeX markup get rejected by our linter.

**To Reproduce:**
Link to katex.org

https://katex.org/?data=%7B%22displayMode%22%3Atrue%2C%22leqno%22%3Afalse%2C%22fleqn%22%3Afalse%2C%22throwOnError%22%3Atrue%2C%22errorColor%22%3A%22%23cc0000%22%2C%22strict%22%3A%22warn%22%2C%22output%22%3A%22htmlAndMathml%22%2C%22trust%22%3Afalse%2C%22macros%22%3A%7B%22%5C%5Cf%22%3A%22%231f(%232)%22%7D%2C%22code%22%3A%22%5C%5Ccancel%7B%5C%5Ctext%7B%D5%A3%7D%7D%22%7D

**Expected behavior:**
I guess `\cancel` should use some reasonable height default for characters without vertical metrics.

**Screenshots:**

obrazek

 - KaTeX Version: version on katex.org
 - Device: Laptop
 - OS: Linux
 - Browser: latest Firefox, latest Chrome

ronkok:
Fix #2617.


danielhollas:
Thanks @ronkok! While we're at it, it might be nice to add other possible missing scripts from Khan Academy languages. I'll look into which might be missing tomorrow or Friday.

test/unicode-spec.js
```diff
@@ -80,6 +80,11 @@ describe("unicode", function() {
         expect`გამარჯობა`.not.toParse(strictSettings);
     });
 
+    it("should build Armenian both inside and outside \\text{}", function() {
+        expect`ԱԲԳաբգ`.toBuild(nonstrictSettings);
```
danielhollas, line 84: Based on rendering on katex.org, I'd think this test was passing even before this change?
Would it be possible to add a test that was failing before?

danielhollas:
@ronkok I went through all current languages supported by Khan Academy. Would it be possible to also add the [Khmer script](https://en.wikipedia.org/wiki/Khmer_script) and [Arabic scripts](https://en.wikipedia.org/wiki/Arabic_script)?

ronkok:
Kmer and Arabic scripts are way outside of my experience and I am reluctant to charge in and do something that might make matters worse.

For instance, @HosseinAgha has developed a plug-in that has support for Persian and at least partial support for Arabic. I don't want to make a change that could interfere with that plug-in.

I know nothing about Kmer, but its Wikipedia page alludes to many diacritics and ligatures. Before we make a blunt change, I'd like to get advice from somebody who can tell us if it is a positive change.

danielhollas:
I agree with keeping Arabic out of this PR. To properly support it at KA, we would need to use that plugin anyway.

It would be nice to add Khmer, since they already have a KA site so we'll hit the same problem as with Armenian sooner or later.
Not an expert here, somebody else might chime in, I am happy to test things out. 

That said, the existing supported unicode blocks already contain various diacritics so while rendering might not be optimal, probably better than not including them, since each get M vertical metrics anyway? If this also influences horizontal metrics, that would be problematic, since many diacritics do not take extra horizontal space.

ylemkimon (approved):

Issue #2637, opened 2020-11-13
globalGroup adversely affects \color and tall right delimiters


ronkok:
PR #2091 has an unintended consequence. When the rendering option `globalGroup` is set to `true`, then `\color` escapes its group and affects tall right delimiters.

The KaTeX function`\color` works by creating a macro `\current@color` and assigning a color to it. When `globalGroup` is true, then this macro escapes its group. Tall right delimiters are then given that color in line https://github.com/KaTeX/KaTeX/blob/21c4bc355c04cfa13bb79dd9dd06627f5b04f54d/src/functions/delimsizing.js#L155. I don't know why tall right delimiters get that treatment, but there it is.

The stated reason for PR #2091 was to give global scope to `\newcommand`.  But its effect is more sweeping than that. I think that @edemaine's [comment](https://github.com/KaTeX/KaTeX/pull/2091#issuecomment-527243164) was correct. We need some mechanism that gives global scope to `\newcommand`, not to all macros.

cirosantilli:
I wonder if there is a way to instead just prevent the color escaping side effect.

But yes, if it is hard, then I suppose it does not matter much as users can use gdef instead of def, and those are less important that newcommand anyawys.

I would keep the option name unchanged I think in any case, as it is more descriptive/specific than `strict`. E.g. in normal LaTeX, putting those definitions in global scope would make them visible to other `$$` equations, so it is not strictly strict (i.e a LaTeX extension) in that sense, but rather a way of saying "this goes in global scope". `\def` not being global could be considered a known incompatibility instead.

ylemkimon:
Could you give an example?

edemaine:
I'm surprised that this happens. While the original #2091 made all defs global, which would be very bad (it would also mess with array's temporary redefinition of `\\`), the #2091 that got merged only makes the top-level group that math gets evaluated in the global group. I don't think this should affect how `\current@color` works, but I might be missing something. Perhaps #1845 in general?

ronkok:
@ylemkimon For an example, I've created a minimal HTML document. It's entire `<body>` is:

```html
<p data-tex="\color{#0000FF} AaBb123"></p>
<p data-tex="\left.\frac a b \right)"></p>
<script>
  const ps = document.getElementsByTagName("p")
  const macros = {};
  [...ps].forEach(el => {
    const tex = el.dataset.tex
    katex.render(tex, el, {globalGroup: true, macros})
  });
</script>
```
The resulting rendering is:
[...]

PR #2640, opened 2020-11-14
fix: Revert \def to local scope regardless of rendering options.

ronkok:
This PR addresses issue #2637. It partially undoes PR #2091. The resulting behavior is that `\def` once again is always confined to local scope. `\gdef` is of course still available for global macros.

Part of PR #2091 is kept. That is, `\newcommand`, `\renewcommand`, and `\providecommand` can be made to have global scope by setting a rendering option. 

In order to minimize the complexity of KaTeX rendering options, the previous `globalGroup` option no longer exists. Instead, `\newcommand` becomes global whenever `strict` is not `true`.


ronkok:
@cirosantilli Is this PR acceptable to you?

PR #2679, opened 2020-12-12
fix: rewrite of splitAtDelimiters.js -- new fix for #2523

### LINKED ISSUE #2523: problem with parsing text with mixed $$ and $ delimiters
> Before reporting a bug

**Describe the bug:**
In LaTeX source texts it is common to see inline math like `$x$$y$` which is equivalent to `$xy$`
mixed with display math.

However, such math does not render correctly with the autorender extension, when the delimiter option is set to
```
{delimiters: [
  {left: '$$', right: '$$', display: true},
  {left: '$', right: '$', display: false},
  {left: '\\(', right: '\\)', display: false},
  {left: '\\[', right: '\\]', display: true}
]}
```

**To Reproduce:**

display this file in a supported browser
```html
<!DOCTYPE html>
<!-- KaTeX requires the use of the HTML5 doctype. Without it, KaTeX may not render properly -->
<html>
  <head>
    <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/katex@0.12.0/dist/katex.min.css" integrity="sha384-AfEj0r4/OFrOo5t7NnNe46zW/tFg\
W6x/bCJG8FqQCEo3+Aro6EYUG4+cU+KJWu/X" crossorigin="anonymous">

    <!-- The loading of KaTeX is deferred to speed up page rendering -->
    <script defer src="https://cdn.jsdelivr.net/npm/katex@0.12.0/dist/katex.min.js" integrity="sha384-g7c+Jr9ZivxKLnZTDUhnkOnsh30B4H0rpLUpJ4j\
AIKs4fnJI+sEnkvrMWph2EDg4" crossorigin="anonymous"></script>

    <!-- To automatically render math in text elements, include the auto-render extension: -->
    <script defer
    src="https://cdn.jsdelivr.net/npm/katex@0.12.0/dist/contrib/auto-render.min.js"
    integrity="sha384-mll67QQFJfxn0IYznZYonOWZ644AWYC+Pt2cHqMaRhXVrursRwvLnLaebdGIlYNa"
            crossorigin="anonymous" onload="renderMathInElement(document.body, {delimiters: [
  {left: '$$', right: '$$', display: true},
  {left: '$', right: '$', display: false},
  {left: '\\(', right: '\\)', display: false},
  {left: '\\[', right: '\\]', display: true}
]});"></script>
  </head>
  <body>
    <h1>A note on discrete $$\mathcal{R}$$ symmetries in $\mathbb{Z}$$_{6}$-II orbifolds with Wilson lines</h1>

  </body>
</html>
```

**Expected behavior:**
both the display math and the inline math should be rendered. It is not, but the `$$` in the middle of  `$\mathbb{Z}$$_{6}$` disappears. 

Omitting the `$$` (which is a NOOP) in the middle results in correct rendering.
Introducing a space `$\mathbb{Z}$ $_{6}$` also renders the inline math, albeit with a gratuitous space.

**Screenshots:**
katex-autorender

 - KaTeX Version: v0.12.0
 - Device: Linux desktop
 - OS: Fedora 32
 - Browser: Chrome
 - Version: 85.0.4183.102 (Official Build) (64-bit)

**Additional Context:**

We use KaTeX for rendering in a large repository of scientific papers. We don't control the TeX source, and hence can't avoid this mix of delimiters.

pzinn:
Resolves #2523 as well as other related auto-render problems (mentioned in the comments of the issue).
The text is only parsed once for all delimiters at once (this is presumably faster than the old, sequential way).
Everything works fine now according to my limited testing, but this is not my project so probably requires more testing.


pzinn:
oh yeah, I might need to update the tests. not sure what the semantic stuff is about.

kevinbarabash:
@pzinn the semantic stuff is referring to the formatting of use of `semantic-release` to help with publishing releases.  It looks for specific strings in the PR title when deciding how to bump the version.  In this case I think prefixing the title with `fix: ` should resolve it.

Could you sign our [Contributor License Agreement](https://docs.google.com/forms/d/e/1FAIpQLSdyXYrc8ogVoA46J9KXyIj5nKlZzNkOnQG-4A1R7X_BWGTShQ/viewform)?  Please ping me once you've done so (our CLAbot is out-of-order at the moment).

kevinbarabash:
@pzinn thanks for signing the CLA.

contrib/auto-render/splitAtDelimiters.js
```diff
@@ -27,76 +27,72 @@ const findEndOfMath = function(delimiter, text, startIndex) {
...
+                data: text.slice(delimiters[i].left.length, index),
+                rawData: text.slice(0, index + delimiters[i].right.length),
+                display: delimiters[i].display || false,
             });
-        } else {
-            finalData.push(startData[i]);
+            text = text.slice(index + delimiters[i].right.length);
         }
+
+        lookingForLeft = !lookingForLeft;
```
kevinbarabash, line 85: I think it'd be a bit easier to grok if the `if` and `else` blocks above set this explicitly, e.g.
```
if (lookingForLeft) {
   ...
   lookingForLeft = false;
} else {
   ...
   lookingForLeft = true;
}
```

contrib/auto-render/splitAtDelimiters.js
```diff
@@ -27,76 +27,72 @@ const findEndOfMath = function(delimiter, text, startIndex) {
...
+    let index;
+    const data = [];
 
-            nextIndex = text.indexOf(leftDelim);
-            if (nextIndex !== -1) {
-                currIndex = nextIndex;
-                finalData.push({
+    while (true) {
+        if (lookingForLeft) {
+            index = text.search(regexLeft);
```
kevinbarabash, line 53: Using `match` could save a bit of work later since it returns both the index of the delimiter was well as which delimiter was matched although you'd still have to search through `delimiters` to find the matching right delimiter.

contrib/auto-render/splitAtDelimiters.js
```diff
@@ -27,76 +27,72 @@ const findEndOfMath = function(delimiter, text, startIndex) {
...
-                        display: display,
-                    });
-
-                    currIndex = nextIndex + rightDelim.length;
-                }
-
-                lookingForLeft = !lookingForLeft;
+        } else {
+            let i = 0;
+            while (!text.startsWith(delimiters[i].left)) { i++; }
```
kevinbarabash, line 66: This could be:
```
const delim = delimiters.findIndex(delim => text.startsWith(delim.left));
```
It might be worth adding a comment here as well saying why this should always succeed.

contrib/auto-render/splitAtDelimiters.js
kevinbarabash: The more I look at this I wonder if the `if` and `else` blocks are even necessary.  We start out with `lookingForLeft` set to `true` and once we found the "left" we always switch to finding the "right".  Maybe we can get rid of `lookingForLeft` and the `if`/`else` blocks altogether.  That would also mean that we could get rid of some of the variables defined before the `while` loop.

contrib/auto-render/splitAtDelimiters.js
```diff
@@ -27,76 +27,72 @@ const findEndOfMath = function(delimiter, text, startIndex) {
     return -1;
 };
 
-const splitAtDelimiters = function(startData, leftDelim, rightDelim, display) {
-    const finalData = [];
+const escapeRegex = function(string) {
+    return string.replace(/[-/\\^$*+?.()|[\]{}]/g, "\\$&");
+};
```
kevinbarabash, line 32: What if there's a `\$` in some TeX`, won't it end up getting modified?

contrib/auto-render/auto-render.js
```diff
@@ -107,11 +96,7 @@ const renderMathInElement = function(elem, options) {
         {left: "\\(", right: "\\)", display: false},
         // LaTeX uses $…$, but it ruins the display of normal `$` in text:
         // {left: "$", right: "$", display: false},
-
-        //  \[…\] must come last in this array. Otherwise, renderMathInElement
-        //  will search for \[ before it searches for $$ or  \(
-        // That makes it susceptible to finding a \\[0.3em] row delimiter and
-        // treating it as if it were the start of a KaTeX math zone.
+	// $ must come after $$
```
kevinbarabash, line 99: style nit: inconsistent indentation

contrib/auto-render/auto-render.js
```diff
@@ -76,8 +65,8 @@ const renderElem = function(elem, optionsCopy) {
             const className = ' ' + childNode.className + ' ';
             const shouldRender = optionsCopy.ignoredTags.indexOf(
                 childNode.nodeName.toLowerCase()) === -1 &&
-                    optionsCopy.ignoredClasses.every(
-                        x => className.indexOf(' ' + x + ' ') === -1);
+                  optionsCopy.ignoredClasses.every(
+                      x => className.indexOf(' ' + x + ' ') === -1);
```
kevinbarabash, line 69: style nit: inconsistent indentation

contrib/auto-render/test/auto-render-spec.js
```diff
@@ -157,14 +169,14 @@ describe("A delimiter splitter", function() {
...
-            ]);
+           expect("hello \\( ( world ) boo").toSplitInto(
+           "(", ")",
+           [
+           {type: "text", data: "hello \\( "},
+           {type: "math", data: " world ",
+           rawData: "( world )", display: false},
+           {type: "text", data: " boo"},
+           ]);
         */
```
kevinbarabash, line 180: I think we should have a test like this with a mismatched delimiter, but it should probably throw.

contrib/auto-render/test/auto-render-spec.js
```diff
@@ -191,44 +204,68 @@ describe("A delimiter splitter", function() {
...
+                                     {left:"$", right:"$", display:false},
+            ])).toEqual(
+            [
+                {type: "math", data: "\\fbox{$hi$}",
+                    rawData: "\\(\\fbox{$hi$}\\)", display: false},
             ]);
     });
+
+
+    // the two tests below are no longer relevant
```
kevinbarabash, line 229: Let's get rid of them.

contrib/auto-render/test/auto-render-spec.js
```diff
@@ -191,44 +204,68 @@ describe("A delimiter splitter", function() {
...
-                {type: "math", data: " world ",
-                    rawData: "( world )", display: false},
-                {type: "text", data: " boo"},
-                {type: "math", data: "math", rawData: "(math)", display: true},
-                {type: "text", data: "hello "},
-                {type: "math", data: " world ",
-                    rawData: "( world )", display: false},
-                {type: "text", data: " boo"},
+                                     {left:"\\(", right:"\\)", display:false},
+                                     {left:"$", right:"$", display:false},
```
kevinbarabash, line 211: style nit: indentation

contrib/auto-render/test/auto-render-spec.js
```diff
@@ -191,44 +204,68 @@ describe("A delimiter splitter", function() {
...
+        expect(splitAtDelimiters("\\(\\fbox{$hi$}\\)",
             [
-                {type: "text", data: "hello "},
-                {type: "math", data: " world ",
-                    rawData: "( world )", display: false},
-                {type: "text", data: " boo"},
-                {type: "math", data: "hello ( world ) boo",
-                    rawData: "(hello ( world ) boo)", display: true},
+                                     {left:"\\(", right:"\\)", display:false},
+                                     {left:"$", right:"$", display:false},
```
kevinbarabash, line 220: style nit: indentation

contrib/auto-render/test/auto-render-spec.js
```diff
@@ -6,17 +6,16 @@ beforeEach(function() {
         toSplitInto: function(actual, left, right, result) {
             const message = {
                 pass: true,
-                message: "'" + actual + "' split correctly",
+                message: () => "'" + actual + "' split correctly",
```
kevinbarabash, line 9: Thanks for fixing this.

contrib/auto-render/test/auto-render-spec.js
```diff
@@ -146,6 +145,19 @@ describe("A delimiter splitter", function() {
             ]);
     });
 
+    it("correctly processes sequences of $..$", function() {
+        expect("$hello$$world$$boo$").toSplitInto(
```
kevinbarabash, line 149: There should be a test case where both `$` and `$$ are detected correctly, e.g. `"$hello$world$$boo$$"` or `"$hello$$world$$$boo$$"` or something like that.

contrib/auto-render/test/auto-render-spec.js
kevinbarabash: I see that we don't have a test case for `\$` within TeX so it'd be good to add one, maybe something like:
```
"$x = \$5$"
```

kevinbarabash (changes requested):
@pzinn thanks for the PR.  The new algorithm looks good.  The implementation could be cleaned up a bit (see inline comments).  I've also suggested a few additional test cases that should be added as well as some general cleanup of the tests.

contrib/auto-render/splitAtDelimiters.js
```diff
@@ -27,76 +27,72 @@ const findEndOfMath = function(delimiter, text, startIndex) {
...
+    let regexLeft;
+    for (let i = 0; i < delimiters.length; i++) {
+        if (i === 0) {
+            regexLeft = "(";
+        } else {
+            regexLeft += "|";
+        }
+        regexLeft += escapeRegex(delimiters[i].left);
+    }
+    regexLeft = new RegExp(regexLeft + ")");
```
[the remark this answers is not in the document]
pzinn: I agree with not reassigning. however I find the line above unreadable, and I'm not convinced using `` is good coding.

contrib/auto-render/splitAtDelimiters.js
```diff
@@ -27,76 +27,72 @@ const findEndOfMath = function(delimiter, text, startIndex) {
...
+        } else {
+            let i = 0;
+            while (!text.startsWith(delimiters[i].left)) { i++; }
+            index = findEndOfMath(
+                delimiters[i].right,
+                text,
+                delimiters[i].left.length
+            );
+            if (index === -1) {
+                break;
```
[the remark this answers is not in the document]
pzinn: currently autorender does not throw an error if it finds a loney opening delimiter e.g. `$a=1`. I haven't changed this behavior.

contrib/auto-render/splitAtDelimiters.js
pzinn: sure, I just kept the existing code.

contrib/auto-render/splitAtDelimiters.js
pzinn: I disagree. `match` is slower and produces lots of useless information. and as you pointed out a search is still needed.

contrib/auto-render/test/auto-render-spec.js
pzinn: weird, I thought I'd run `--whitespace=fix`. am I missing something?

pzinn:
thanks for reviewing will take a look soon

contrib/auto-render/splitAtDelimiters.js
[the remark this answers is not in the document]
kevinbarabash: I guess if people are happy with the current behavior let's keep it.

contrib/auto-render/test/auto-render-spec.js
kevinbarabash: What command are you passing `--whitespace=fix` to?

pzinn: `git rebase`. what is the canonical way to fix indentation in KaTeX code?

pzinn: emacs `untabify` and `yarn eslint . --fix` seem to do the trick

pzinn:
done, I think.
if we're going to fully rewrite `splitAtDelimiters.js`, one could also improve `findEndOfMath`. It's probably possible to speed it up quite a bit with some tricky regex, though there may be compatibility issues (e.g., Internet Explorer is missing advanced regex features).

contrib/auto-render/test/auto-render-spec.js
kevinbarabash: I didn't know about that git command.  I was going to setup prettier at one point but there's a bunch of code/comments in this repo that prettier ends up making worse.

kevinbarabash (approved):
LGTM.  Thanks for adding some more tests and refining `splitAtDelimiters`.

contrib/auto-render/auto-render.js
kevinbarabash: When I went to edit this file the indentation looks fine.

PR #2680, opened 2020-12-13
docs: delete stray backtick

brianstrauch:


edemaine (approved):
Thanks for the typo fix!

PR #2683, opened 2020-12-13
chore: Verify semantic messages via Pull Request title, not commits

edemaine:
KaTeX always squash-merges PRs, so this seems like the preferred way to ensure that final commits support semantic-release.  I'm hoping it will make fixing nonsemantic PRs easier as well (like #2680); see https://github.com/zeke/semantic-pull-requests/issues/112

### DIFF
--- a/.github/semantic.yml
+++ b/.github/semantic.yml
@@ -1,0 +1,2 @@
+# Always validate the PR title, and ignore the commits
+titleOnly: true


kevinbarabash (approved):
Let's give it a try.

ylemkimon (changes requested):
AFAIK, `semantic-pull-requests` already does that. It automatically detects the final commit title:
- If there is a single commit, it checks for the commit message
- If there are multiple commits, it checks for the PR title

edemaine:
@ylemkimon Thanks for the explanation. I guess the issue then is that single-commit PRs don't look at the PR title, so we can't change the PR title to satisfy the check, making it annoying to merge single-commit PRs. Perhaps this change will make all PRs act like the multi-commit case?

edemaine:
Could we try merging this? [This comment](https://github.com/zeke/semantic-pull-requests/issues/112#issuecomment-709389100) suggests that it might work...  And it does seem that `titleOnly` is set to `false` by default.

ylemkimon:
However, in the case of a single commit, GitHub automatically fills the title of the merge commit with its title. So the red light means the merger should _manually_ change the commit title.

edemaine:
My goal was to be able to manually change the PR title (which is easy in the GitHub interface), rather than having to manually change the commit title (which requires checking out the branch and force-pushing).

ylemkimon:
Currently, `semantic-pull-requests` is not a required status check, so a pull request can be merged even if it fails. I think failed status would serve as a reminder to change the merge commit title via the GitHub interface.

edemaine:
Ah, I see, the intent is to edit the commit title in the "squash" interface.  That makes sense.

I somehow thought the check was required, but you're right that it isn't.  Should it be?

ylemkimon:
@edemaine Unfortunately, updating the PR title doesn't affect the title of merge commit if there is a single commit. Since a non-compliant commit message doesn't break KaTeX itself, I think making it not required is okay.

PR #2698, opened 2020-12-23
fix: Use SVGs to avoid gaps in tall delimiters.

### LINKED ISSUE #2697: incorrect rendering of very large delimiters
I have this vague feeling that I've already reported this. if so, apologies but I couldn't find among the issues.
```
\left\{\{\raisebox{-30cm}{}\right\}
```
will produce this (note the gap)
katex

ronkok:
Fixes #2697.

When issue #1975 cropped up, @kevinbarabash suggested using SVGs for the middle of tall stretchy delimiters. I should have followed his advice. This PR belatedly does use SVGs, replacing the work done in PR #1986. SVGs have served as reliable stretchy horizontal elements. I expect them to be an improvement over what we've seen when the browser does vertical alignment with font glyphs.


ronkok:
In the screenshots, some of the delimiters are a little shorter than they were. TeX makes tall delimiters by stacking font glyphs and it doesn't overlap the glyphs, so an increase in height proceeds in increments of the repeat glyph height. This PR can stretch SVGs to any height, so it makes each delimiter the minimum height.

ronkok:
Hey @ylemkimon: Thank you for working on the screenshotter procedure. It's much better now.

kevinbarabash:
@ronkok thanks for creating stretchy SVGs for the delimiters.

> In the screenshots, some of the delimiters are a little shorter than they were. TeX makes tall delimiters by stacking font glyphs and it doesn't overlap the glyphs, so an increase in height proceeds in increments of the repeat glyph height. This PR can stretch SVGs to any height, so it makes each delimiter the minimum height.

This definitely feels like an improvement, but it is different from TeX.  I'm curious how other people feel about making this the default behavior even though it differs from TeX.

Screen Shot 2020-12-24 at 12 39 54 PM

Why don't the delimiters match the height of the array in this screenshot test?

ronkok:
I've reverted the height determining logic to what it was before this PR.

kevinbarabash:
@ronkok the changes in https://github.com/KaTeX/KaTeX/pull/2698/commits/57351df2b696ba0e437f2d5673400d99b23596ca are pretty isolated so maybe this behavior could be added behind a setting in the future if you want.  It's not necessary for this PR though.

ronkok:
I'm good with 57351df.  Let's keep it simple.

src/delimiter.js
```diff
@@ -152,11 +152,24 @@ const makeLargeDelim = function(delim,
...
+    // advance width of glyphs in the KaTeX fonts.
+    "\u2016": 0.778,
+    "\u2223": 0.333,
+    "\u2225": 0.556,
+    "\u239c": 0.875,
+    "\u239f": 0.875,
+    "\u23a2": 0.667,
+    "\u23a5": 0.667,
+    "\u23aa": 0.889,
+    "\u23d0": 0.667,
```
kevinbarabash, line 165: We should probably import fontMetricsData and look up the values from there directly, that way these numbers can't get out of sync if the fonts get updated.

src/delimiter.js
```diff
@@ -331,74 +373,44 @@ const makeStackedDelim = function(
...
-        const pathName = repeat === "\u239c" ? "leftParenInner" : "rightParenInner";
-        const innerSpan = buildCommon.staticSvg(pathName, options);
-        inners.push({type: "elem", elem: innerSpan});
-        inners.push({type: "kern", size: -overlap});
-    } else {
-        inners.push({type: "kern", size: shiftOfExtraElement});
-        inners.push(makeInner(repeat, font, mode));
-        inners.push(lap);
+        stack.push(lap);
+        stack.push(makeCorner(middle, font, mode));
```
kevinbarabash, line 399: It feels a little weird call `makeCorner` here but I don't have a suggestion for a better name for the function.  It might be worth add a short comment here saying the `makeCorner` is used to create `VListElem`s for all the non-stretchy parts of delimiters.

src/delimiter.js
```diff
@@ -331,74 +373,44 @@ const makeStackedDelim = function(
...
-            inners.push(lap);
-            inners.push(makeInner(repeat, font, mode));
-        }
-        // Insert one extra repeat element.
-        inners.push({type: "kern", size: shiftOfExtraElement});
-        inners.push(makeInner(repeat, font, mode));
-        inners.push(lap);
+        const innerHeight = (realHeightTotal - topHeightTotal - bottomHeightTotal -
+            middleHeightTotal) / 2 + 0.016;
+        stack.push(makeInner(repeat, innerHeight, options));
```
kevinbarabash, line 396: It would be good to have a constant for the overlap to avoid having `const lap = {type: "kern", size: -0.008};` up above and the `+ 0.016` here get out of sync.

src/svgGeometry.js
```diff
@@ -151,11 +151,35 @@ export const sqrtPath = function(
     return path;
 };
 
-export const path: {[string]: string} = {
-    // Two paths that cover gaps in built-up parentheses.
-    leftParenInner: `M291 0 H417 V300 H291 z`,
-    rightParenInner: `M457 0 H583 V300 H457 z`,
+export const innerPath = function(name: string, height: number ): string {
+    // The inner part of stretchy tall delimiters
```
kevinbarabash, line 155: super style nit: extra space after `height: number`.

src/svgGeometry.js
```diff
@@ -151,11 +151,35 @@ export const sqrtPath = function(
...
+    switch (name) {
+        case "\u239c":
+            return `M291 0 H417 V${height} H291z M291 0 H417 V${height} H291z`;
```
kevinbarabash, line 158: It looks like the same path is being repeated twice.  Am I misunderstanding what this SVG path is doing?

kevinbarabash (commented):
Thanks for adding stretchy SVG connectors.

src/delimiter.js
ronkok: I'll do that.

But keep in mind that if the fonts change, then the paths in `svgGeomtry.js` will also be invalid. This PR, and some other SVG tactics, bind us pretty tightly to the existing font glyphs.

src/delimiter.js
ronkok: Yep, I had the same thought as I was naming the variable.

src/delimiter.js
ronkok: I'll tie them together.

src/delimiter.js
ronkok: I ended up renaming that function to `makeGlyphSpan`.

src/svgGeometry.js
ronkok: > It looks like the same path is being repeated

Correct. See https://github.com/KaTeX/KaTeX/blob/master/src/stretchy.js#L96-L106

ronkok:
Comments are picked up. This should be ready to go.

kevinbarabash (approved):
LGTM

PR #2701, opened 2020-12-23
feat: Set Auto-render to recognize AMS environments without $$…$$ delimiters.

ronkok:


kevinbarabash:
I haven't reviewed #2679, but I think there might be a conflict between this PR and that one.  Hopefully it won't be too hard to resolve.

ronkok:
I tried to write this PR in a way that would avoid a conflict with #2679. I hope I succeeded.

ronkok:
The merge conflict in this PR is resolved. It's ready to go.

kevinbarabash (commented):
Thanks for the PR.  It's nice that people will be able to use `align`, `equation`, etc. in their documents while avoiding having to deviate from LaTeX's handling of math mode.

It looks like `renderMathInText` doesn't actually use the environment delimiters when rendering the math. As a result, the content between these delimiters doesn't get rendered within that environment which is what I was expect to happen.

ronkok:
Well, that was embarrassing. Fixed now.

kevinbarabash (approved):
LGTM. 🎉

edemaine:
Congrats, this is awesome! As it's one of the most requested features I've heard of KaTeX ("where's align?"), we should think about a release soon

kevinbarabash:
It would be nice to get the pending `fix` PRs reviewed before a release.

Issue #2513, opened 2020-09-14
Macros do not persist despite gdef and globalGroup


Recmo:
> Before reporting a bug
- [x] Check [common issues](https://katex.org/docs/issues.html).
- [x] Check the bug is reproducible in [the demo](https://katex.org). If not, check KaTeX is up-to-date and installed correctly.
- [x] Search for [existing issues](https://github.com/KaTeX/KaTeX/issues).

**Describe the bug:**
Macros are forgotten between invocations of `renderToString` (and possibly `render`), even when `globalGroup` is set and `\gdef` is used.

```javascript
katex.renderToString(String.raw`\gdef\foo{1}`);
katex.renderToString(String.raw`\foo`); // Fails
```

```javascript
katex.renderToString(String.raw`\gdef\foo{1}`, { globalGroup: true });
katex.renderToString(String.raw`\foo`); // Fails
```

**To Reproduce:**
1. Go to 'https://katex.org.'
2. Open javascript console
3. Enter either of the two examples above

**Expected behavior:**
Foo macro to expand because it was defined globally.

**Screenshots:**

**Environment (please complete the following information):**
 - KaTeX Version: 0.12.0
 - Device: Desktop
 - OS: MacOS 10.15.6 (19G2021)
 - Browser: Firefox
 - Version: 80.0.1 (64-bit)

**Additional context:**
I'm trying to define some globally available macros that require arguments. Doing this through the `macros` option seems extremely complicated and I can not find a working example, I'd much prefer parsing some `\gdef` statements.

Recmo:
After digging some more, the following seems to work:

```javascript
var macros = {};
katex.renderToString(String.raw`\gdef\foo{1}`, { macros });
katex.renderToString(String.raw`\foo`, { macros })
```

edemaine:
Right, that's the intended behavior -- sorry it took you hours to figure out! Do you have a suggestion on where we (or you via PR) could mention this in the documentation that would have helped you find it sooner?

PR #2702, opened 2020-12-23
docs: Explain how to make macros persist.

ronkok:
My attempt to address #2513.


docs/api.md
```diff
@@ -30,6 +30,21 @@ var html = katex.renderToString("c = \\pm\\sqrt{a^2 + b^2}", {
 // '<span class="katex">...</span>'
 ```
 
+## Persistent Macros
+
+KaTeX documentation tells the author that `\gdef` will create a macro that persists between KaTeX fields. In order to fulfill that promise, you must make certain provisions. Say that a parser has dropped from the sky with an array of DOM elements that contain math. You could then write this code:
```
edemaine, line 35: Here's a proposed second draft:

```suggestion
If you want to define macros within one math element that can be used in other math elements, you need to do two things:
1. Create one shared `macros` object that you pass into every call to `katex.render` or `katex.renderToString`. (Do not create a fresh `macros` object for each call.)
2. In your math content, use `\gdef` to define your macros (so that they are defined in the global group, not just the local group of that particular math mode).

For example, suppose that you have an array `mathElements` of DOM elements that contain math. Then you could then write this code:
```

docs/api.md
```diff
@@ -30,6 +30,21 @@ var html = katex.renderToString("c = \\pm\\sqrt{a^2 + b^2}", {
...
+const macros = {};
+for (let element of mathElements) {
+    katex.render(element.textContent, element, {
+        throwOnError: false,
+        macros
+    };
+```
+
+Notice that you create the `macros` object outside the loop. That object will contain macros that the user defines using `\gdef` and the macros will persist between calls to `katex.render`.
+
```
edemaine, line 47: It's probably worth warning about security:

```suggestion

### Security of Persistent Macros
Persistent macros can change the behavior of KaTeX (e.g. redefining standard commands), so for security, such a setup should be used only for multiple elements of common trust.  For example, you might enable persistent macros within a message posted by a single user (by creating a `macros` object for that message), but you probably should not enable persistent macros across multiple messages posted by multiple users.
```

edemaine (commented):
Thanks for working on this!  I took a stab at a second draft, and an additional notice about security &mdash; feel free to take from it what you like.  Love the example.

docs/api.md
ronkok: > **If** you want to define macros within one math element that can be used in other math elements

I'd like to keep consistency between what we say here and what we say in  `supported.md`. `supported.md` does not have a qualifying "if". It says that `\gdef` will persist, full stop.  Since security makes that uncertain, I want to revise `supported.md` before this PR goes anywhere.

 I'll use much of your draft here, but I may revise parts of it.  Revision coming later today.

docs/api.md
ronkok: Good point. I'll add it.

docs/api.md
edemaine: Sounds good.  I just found it weird to be referring to another part of the documentation, without even a link, as it's assuming a lot of the reader (what if they read this part before the other part?).  Perhaps the documentation that talks about `\gdef` should also refer to this section, so the proviso is linked directly from the potential offender.

ronkok: > without even a link

Excellent point. I will definitely write a link into this PR.

Not a reverse link, though. An author in, say [Observable](https://observablehq.com/), does not need to be bothered with the machinery behind the scenes. That's the sort of person I'm targeting in `supported.md`

ronkok:
I've made a revision to the docs. The relevant sections now contain a mixture of your words and mine.

docs/api.md
```diff
@@ -41,3 +41,25 @@ hover text.  Without this option, invalid LaTeX will cause a
 
 The last argument to `katex.render` and `katex.renderToString` can contain
 [a variety of rendering options](options.md).
+
+## Persistent Macros
+
+KaTeX’s [macro documentation](supported.html#gdef) tells the author that `\gdef` will create a macro that persists between KaTeX elements. In order to enable that persistence, you must create one shared `macros` object that you pass into every call to `katex.render` or `katex.renderToString`. (Do not create a fresh `macros` object for each call.)
+
+For example, suppose that you have an array `mathElements` of DOM elements that contain math. Then you could then write this code:
```
edemaine, line 49:
```suggestion
For example, suppose that you have an array `mathElements` of DOM elements that contain math. Then you could write this code:
```
(double "then")

docs/supported.md
```diff
@@ -323,7 +323,9 @@ Macros can also be defined in the KaTeX [rendering options](options.md).
 
 Macros accept up to nine arguments: #1, #2, etc.
 
-`\gdef`, `\xdef`, `\global\def`, `\global\edef`, `\global\let`, and `\global\futurelet` will persist between math expressions.
+<div id="gdef"></div>
+
+`\gdef`, `\xdef`, `\global\def`, `\global\edef`, `\global\let`, and `\global\futurelet` will persist between math expressions. (Exception: macro persistence may be disabled for security reasons.)
```
edemaine, line 328: 1. For correctness, should this sentence be prefixed by `Macros defined by`?  (It's the macros that persist, not `\gdef` itself.)
2. I would personally reword the second sentence to something like `(assuming this feature has been [properly setup](api.html#persistent-macros) and not disabled for security reasons)`. But up to you whether to reword; I understand you want to keep this simple.

edemaine (commented):
Looks great! Nice improvements. I just have a couple of small possible tweaks and then this should be good to go.

docs/supported.md
ronkok: Comments picked up. I revised one of them to avoid implying incompetence on the part of the web site administrator.

edemaine (approved):
Nice, thanks!

PR #2703, opened 2020-12-23
fix: Prevent global group from adversely affecting color.

### LINKED ISSUE #2637: globalGroup adversely affects \color and tall right delimiters

ronkok:
Fix #2637. Close #2640.

This PR is an alternative to PR #2640. Either this one or that one, but not both, should be approved and merged. I would prefer PR#2640, but it seems unlikely that anyone will approve it. We should get this bug fixed before the next release so this PR is the next best thing.


kevinbarabash:
> I would prefer PR#2640, but it seems unlikely that anyone will approve it.

I'll review that one sometime this week.

ylemkimon (approved):

ylemkimon:
I'll leave #2637 for further discussions.

Issue #2381, opened 2020-08-03
Support outer $...$ or $$...$$ in topEnv: true


edemaine:
Small request: could we make `topEnv: true` also work for enclosing `$...$` or `$$...$$` at the top level?

This would make it easier for auto-render (and the [many](https://github.com/edemaine/coauthor/blob/master/lib/formats.coffee) similar calls-to-KaTeX that exist in other software) to just pass in the whole math expression, whether it's a special environment or not -- currently I assume we have to strip the enclosing `$`s.

_Originally posted by @edemaine in https://github.com/KaTeX/KaTeX/issues/2378#issuecomment-668200692_

PR #2709, opened 2020-12-28
fix: Replace topEnv with  check of input string.

ronkok:
When I wrote the PR for `{equation}`, `{align}`, etc., I included a parameter named `topEnv`. I was trying to follow up on suggestions in issue #2042, but I did not have a firm grasp on what `topEnv` was supposed to accomplish. I ended up using `topEnv` to try to ensure that AMS environments were written at the top level.

I believe I botched that part. Currently `master` will return a ParseError for any AMS environment if the rendering options do not include either (1) strict = false, or (2) topEnv = true.

That's a weird condition to impose on users. So this PR removes `topEnv` and instead checks the input string to see if the environment was defined at the very start. It's a better check. I was even able to strengthen some of the unit test by changing strict from false to "warn".

There is another issue, #2381, which suggests allowing KaTeX input strings to include the `$` delimiters that many editors use to delineate math. I think that is a pretty good idea, but it is a different thing and belongs in a different PR. Right now, `topEnv` is just an awkward, unnecessary check parameter and it should go. We can address #2381 later.

### DIFF
--- a/src/environments/array.js
+++ b/src/environments/array.js
@@ -48,11 +48,14 @@ function getHLines(parser: Parser): boolean[] {
 
 const validateAmsEnvironmentContext = context => {
     const settings = context.parser.settings;
+    const input = context.parser.gullet.lexer.input.trim();
     if (!settings.displayMode) {
-        throw new ParseError(`{${context.envName}} cannot be used inline.`);
-    } else if (settings.strict && !settings.topEnv) {
-        settings.reportNonstrict("textEnv",
-            `{${context.envName}} called from math mode.`);
+        throw new ParseError(`{${context.envName}} can be used only in` +
+            ` display mode.`);
+    } else if (!/^\\begin{/.test(input)) {
+        // Some other math exists in this KaTeX element before this environment.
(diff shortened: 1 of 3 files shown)


src/environments/array.js
```diff
@@ -48,11 +48,14 @@ function getHLines(parser: Parser): boolean[] {
...
     const settings = context.parser.settings;
+    const input = context.parser.gullet.lexer.input.trim();
     if (!settings.displayMode) {
-        throw new ParseError(`{${context.envName}} cannot be used inline.`);
-    } else if (settings.strict && !settings.topEnv) {
-        settings.reportNonstrict("textEnv",
-            `{${context.envName}} called from math mode.`);
+        throw new ParseError(`{${context.envName}} can be used only in` +
+            ` display mode.`);
+    } else if (!/^\\begin{/.test(input)) {
```
ylemkimon, line 55: There may be edge cases, like the environment used via macro, e.g., `\def\foo#1{\begin{equation}#1\end{equation}}`

ronkok: Thank you for catching that. I am closing this PR and will attack the problem in a new PR.

PR #2710, opened 2020-12-28
feat: Add {CD} to auto-render.

ronkok:


kevinbarabash (approved):
Thanks for the PR.  I'm sure there'd be questions about this if we didn't make this change.

PR #2712, opened 2020-12-29
fix: Remove topEnv parameter.

ronkok:
In the current KaTeX master, the new AMS environments are each checked for two conditions:

1. KaTeX rendering option must have `displayMode` = true.
2. A new KaTeX rendering option, `topEnv` must be true or `strict` must be false.

I have come to regret the second condition, `topEnv`. Here is what I had in mind: In LaTeX, these environments are not written inside `\[…\]` delimiters. Rather, these AMS environments are themselves a way to change from text mode to math mode. I wanted some way for KaTeX to enforce the placement of these environments as a stand-alone item in `displayMode`.

I would still like to do that, if it were feasible, but I have not found any good way to enforce that condition. I no longer want to make web sites pass `topEnv` = true. Markdown editors are not set up to check for AMS environments. Their consoles will fill up with console errors. Even the KaTeX auto-render extension is not currently set up in a way that enables one to pass `topEnv` = true.

On top of the Markdown issue, I do not look forward to writing the documentation for a `topEnv` parameter. It would be a deep dive into constraints from backwards compatibility to LaTeX quirks. Complicated. And I don't want to answer all the issues that will be raised by confused users about this odd requirement.

I tried to get the check done without `topEnv` by revising `parseArray` such that it would ensure a top-level environment by checking the text of the input, but there doesn't seem to be any way to reliably do that and avoid edge cases.

So I offer two options: (a) This PR removes `topEnv`. (b) Keep `topEnv` but only raise a console error when `strict` = true, not the default value of `strict` = "warn".

My preference is for this PR and the removal of `topEnv`.  It will allow some input that would cause an error in LaTeX. That is not ideal but it better than the alternatives.


test/katex-spec.js
```diff
@@ -2782,41 +2782,41 @@ describe("AMS environments", function() {
...
-        expect`\begin{gather*}a+b\\c+d\end{gather*}`.toBuild(nonStrictDisplay);
-        expect`\begin{align}a&=b+c\\d+e&=f\end{align}`.toBuild(nonStrictDisplay);
-        expect`\begin{align*}a&=b+c\\d+e&=f\end{align*}`.toBuild(nonStrictDisplay);
-        expect`\begin{alignat}{2}10&x+ &3&y = 2\\3&x+&13&y = 4\end{alignat}`.toBuild(nonStrictDisplay);
-        expect`\begin{alignat*}{2}10&x+ &3&y = 2\\3&x+&13&y = 4\end{alignat*}`.toBuild(nonStrictDisplay);
-        expect`\begin{equation}a=b+c\end{equation}`.toBuild(nonStrictDisplay);
-        expect`\begin{equation}\begin{split}a &=b+c\\&=e+f\end{split}\end{equation}`.toBuild(nonStrictDisplay);
-        expect`\begin{split}a &=b+c\\&=e+f\end{split}`.toBuild(nonStrictDisplay);
-        expect`\begin{CD}A @<a<< B @>>b> C @>>> D\\@. @| @AcAA @VVdV \\@. E @= F @>>> G\end{CD}`.toBuild(nonStrictDisplay);
+    const displayMode = new Settings({displayMode: true, strict: false});
```
kevinbarabash, line 2785: Is `strict: false` still needed here?

ronkok: It is not needed. Good catch. I'll take it out.

ronkok: Done.

edemaine:
While this approach doesn't feel ideal, I agree that it seems better than the current state; `topEnv` ends up just being an extra layer of complexity/confusion that doesn't really help a user who probably doesn't care that `$$\begin{align}` is wrong while `\begin{align}` is correct.

It feels weird that we can't distinguish between the two, but I think the "right" way forward might be #2381 (which I'd forgotten about until @ronkok mentioned it in #2709): callers could specify that the input *must* have a top-level environment, which might include `$` or `$$`, via an option that might be called `topEnv` or `startInTextOrOuterParMode` (though it's a very limited form of text/outer par mode).

If we wanted to be **bold**, we could change this "modern content input mode" (which requires `$`s or other top-level environments) to be the actual default interface in KaTeX 1.0.0 (say), as a replacement for `displayMode`.  Or perhaps less bold: have a backward-compatible `displayMode` that wraps the contents in `$` or `$$` accordingly.  Then, to use the old interface, one only needs to specify `displayMode: false` if they haven't already; and to switch to the new interface, they just omit `displayMode`.  In the latter case, we could again throw a warning (when `strict: 'warn'`) if there's a top-level environment within another math environment like `$$`.

kevinbarabash:
It looks like `topEnv` was added since the most recent release so removing it before the upcoming release make sense and we can figure out a better solution after that.

src/environments/array.js
```diff
@@ -49,10 +49,8 @@ function getHLines(parser: Parser): boolean[] {
 const validateAmsEnvironmentContext = context => {
     const settings = context.parser.settings;
     if (!settings.displayMode) {
-        throw new ParseError(`{${context.envName}} cannot be used inline.`);
-    } else if (settings.strict && !settings.topEnv) {
-        settings.reportNonstrict("textEnv",
-            `{${context.envName}} called from math mode.`);
+        throw new ParseError(`{${context.envName}} can be used only in` +
+            ` display mode.`);
```
kevinbarabash, line 53: In addition to checking `!settings.displayMode` here maybe would should also check `settings.useStrictBehavior()` as well like so:
```
if (!settings.displayMode) {
   const message = `{${context.envName}} can be used only in display mode.`;
   if (settings.useStrictBehavior("invalidEnvironment", message)) {
      throw new ParserError(message);
   }
}
```

test/katex-spec.js
```diff
@@ -2783,41 +2783,41 @@ describe("AMS environments", function() {
...
-        expect`\begin{gather*}a+b\\c+d\end{gather*}`.toBuild(nonStrictDisplay);
-        expect`\begin{align}a&=b+c\\d+e&=f\end{align}`.toBuild(nonStrictDisplay);
-        expect`\begin{align*}a&=b+c\\d+e&=f\end{align*}`.toBuild(nonStrictDisplay);
-        expect`\begin{alignat}{2}10&x+ &3&y = 2\\3&x+&13&y = 4\end{alignat}`.toBuild(nonStrictDisplay);
-        expect`\begin{alignat*}{2}10&x+ &3&y = 2\\3&x+&13&y = 4\end{alignat*}`.toBuild(nonStrictDisplay);
-        expect`\begin{equation}a=b+c\end{equation}`.toBuild(nonStrictDisplay);
-        expect`\begin{equation}\begin{split}a &=b+c\\&=e+f\end{split}\end{equation}`.toBuild(nonStrictDisplay);
-        expect`\begin{split}a &=b+c\\&=e+f\end{split}`.toBuild(nonStrictDisplay);
-        expect`\begin{CD}A @<a<< B @>>b> C @>>> D\\@. @| @AcAA @VVdV \\@. E @= F @>>> G\end{CD}`.toBuild(nonStrictDisplay);
+    const displayMode = new Settings({displayMode: true});
```
kevinbarabash, line 2786: If we end up calling `settings.useStrictBehavior()` as suggested above, that'll probably result in a bunch of warning messages in the test logs so we may want to go back to using `strict: false`.

kevinbarabash (approved):
LGTM.  One minor suggestion in the comments about how errors/warning are reported, but it's not blocking.

src/environments/array.js
ronkok: I dunno if this is a good idea. Most of the AMS environments have automatic numbering. That isn't likely to combine well with inline math.  I'm usually the person here who advocates permissive behavior, but I think this is a case where a `ParseError` is better than math rendered in a strange way.

ronkok: I'm going to merge this PR as is. We can still edit `validateAmsEnvironmentContext()` in a separate PR if we decide such an edit is necessary.

PR #2725, opened 2021-01-07
fix:  add namespace for svg, making output XHTML+SVG+MathML compatible

roadupcc:
for sometime svg-renderer need namespace attributes


kevinbarabash:
@oft thanks for the PR.  Could you sign our Contributor License Agreement?  Please ping me once you've done so (our CLAbot is out-of-order at the moment).

oft:
@kevinbarabash I had signed

edemaine:
@oft Have you seen a context where this actually fails, or doesn't pass a check or something? I know raw SVG needs namespaces specified, but I thought namespaces were implicit in HTML5 (but I'd have to double check that).

If they are indeed required, it's also a bit inefficient to put them in every svg element. I believe they can be put at the top level.

oft:
@edemaine yet , the namespace is implicit in HTML5, but this HTML fragment is not only used in HTML5. For example, iOS, or WeChat miniProgram.

Even if it is used in a browser, it may not be displayed normally without a namespace, such as dataURL, so I think the standard practice is to add a namespace.


image


As a comparison


image

kevinbarabash:
@oft signature looks good. Thanks!

kevinbarabash:
@edemaine how would putting the namespace at the KaTeX's root node work?  I thought `xml:ns` needed to be on the `<svg>` nodes since the namespace was for all of the XML within the `<svg>`.

edemaine:
Indeed, I think I was incorrect.  I thought `xmlns=...` meant "load this namespace", but [reading the spec](https://www.w3.org/TR/xml-names/), it also makes it the default namespace, so it would in theory make HTML inaccessible. So never mind that comment.

I just [found confirmation](https://stackoverflow.com/a/18468348/7797661) that you don't need to specify the SVG namespace in HTML5.

I still don't follow @roadupcc's argument for adding a namespace, but perhaps I'm missing a useful context here. HTML (or MathML) is required for KaTeX to render; I can't imagine these SVG elements being used in another context. I'm not strictly against the idea, but it would be nice to know we're not just adding extra characters when there's no need for them.

kevinbarabash:
> I can't imagine these SVG elements being used in another context.

@edemaine that's a good point.  I think adding the namespace would make more sense if everything was rendered into an SVG.

> For example, iOS, or WeChat miniProgram.

@roadupcc could you say more about how KaTeX would be used in these scenarios?

roadupcc:
> > I can't imagine these SVG elements being used in another context.
> 
> @edemaine that's a good point. I think adding the namespace would make more sense if everything was rendered into an SVG.
> 
> > For example, iOS, or WeChat miniProgram.
> 
> @roadupcc could you say more about how KaTeX would be used in these scenarios?

@kevinbarabash 
On develop WeChat miniProgram, I need convert `svg` to base64 format, then as an image . if the svg has no namespace attribute, the image will display error.

other, I think add namespace is a standard,  respect a standard is a good idea. or we can use an option to config the namespace.

edemaine:
I feel like, if you're transforming SVG into base-64 images, it makes sense to add the xmlns at that point. Perhaps adding some documentation about this would make more sense?

How are you rendering the HTML that you cannot include `<svg>` tags?

Regarding standards: The HTML5 standard offers SVG tags without any `xmlns`, so this is standard-compliant. I'm not convinced that adding `xmlns` would actually be standard-compliant, given the death of XHTML.

roadupcc:
I did not understand the standard of svg in html5 carefully, I think it would be better to give an option or provide some other solutions

src/domTree.js
```diff
@@ -505,7 +505,8 @@ export class SvgNode implements VirtualNode {
     }
 
     toMarkup(): string {
-        let markup = "<svg";
+
+        let markup = `<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink"`;
```
edemaine, line 509:
```suggestion
        let markup = `<svg xmlns="http://www.w3.org/2000/svg"`;
```
We don't use `xlink:href` in any generated SVG. We embed images only via HTML ``.

edemaine:
#2881 seems like good motivation for including the SVG namespace.

edemaine (approved):
Approving this for XHTML compatibility, as discussed in #2881

PR #2807, opened 2021-02-24
Docs: Updated Environments to use \Rightarrow instead of copied character

arid-mangoss:
Found this typo in the environments section of the [supported section of the docs](https://katex.org/docs/supported.html). Went ahead and made the fix

The right arrow should be `\Rightarrow` but it looks like it got copied and pasted instead
image


ronkok:
That's not a typo. In KaTeX, `⇒` is a valid alias for `\Rightarrow.`  The KaTeX docs section on [arrows](https://katex.org/docs/supported.html#arrows) even lists `⇒` in the "Direct Input" section just below the table.

arid-mangoss:
Ok, interesting. Thanks for letting me know

PR #2766, opened 2021-02-01
docs: Tamasha app was added to KaTex users list

kookma:
Added Tamasha presentation app

PR #2825, opened 2021-03-08
docs: Tamasha app was added to KaTex users list

kookma:
This corrected #2766 PR

Tamasha app was added to KaTex users list

https://kookma.github.io/TW-Tamasha/


kevinbarabash (approved):
LGTM.  Thanks for adding your app to the users page.

PR #2846, opened 2021-03-17
chore: update Renovate configuration

ylemkimon:
- remove GitHub Actions regex (it's now supported out of the box)
- change browser Docker versioning to major to prevent auto-merge (the new image is not used in the workflow, so the screenshots should be tested and updated manually)


edemaine (approved):
I admit that I don't understand this configuration very well, but the changes seem OK to me.

PR #2854, opened 2021-03-22
docs: fix braket rows in support table

ruiokada:
The rows displaying the braket macros in the support table are not rendering properly due to problems with escaping pipes (see ket/Ket under https://katex.org/docs/support_table.html#jk for example). I replaced the "Rendered" columns for braket with the actual macros (which I think is suppose to be used) so the pipes are no longer needed.


ylemkimon (approved):

ylemkimon:
Thank you for the PR!

PR #2862, opened 2021-03-28
docs: better explanation of when to use mhchem extension

deining:


edemaine (approved):
Thanks for the edit! This should help people using the mhchem extension.

PR #2866, opened 2021-03-29
fix(fonts): update fonts dependencies

ylemkimon:
- Update base image to Ubuntu 20.04
- Use precompiled `ttfautohint` (1.8.3)
  - Remove build-related tools: `wget`, `man-db`, `pkg-config`
- Use Python 3
- Update `fonttools` to 4.21.1

There are subpixel changes to the screenshots: [diff.zip](https://github.com/KaTeX/KaTeX/files/6229538/diff.zip).


ylemkimon:
This PR is ready for review.

edemaine:
@ylemkimon I'm happy to take a look at this now. Could you resolve the conflict / update the screenshots?

edemaine:
@ylemkimon It would be nice to get font building working again. Do you have some time to look at this, or should I try to fix it myself?

ylemkimon:
@edemaine Sorry for the delay. This PR is ready for review.

src/fontMetricsData.js
```diff
@@ -1133,17 +1133,17 @@ export default {
         "8901": [-0.05555, 0.44445, 0, 0, 0.27778],
         "8902": [-0.03472, 0.46528, 0, 0, 0.5],
         "8904": [0.005, 0.505, 0, 0, 0.9],
-        "8942": [0.03, 0.9, 0, 0, 0.278],
-        "8943": [-0.19, 0.31, 0, 0, 1.172],
-        "8945": [-0.1, 0.82, 0, 0, 1.282],
+        "8942": [0.03, 0.903, 0, 0, 0.278],
+        "8943": [-0.19, 0.313, 0, 0, 1.172],
+        "8945": [-0.1, 0.823, 0, 0, 1.282],
```
edemaine, line 1138: These metric changes are presumably why there are subpixel differences in the screenshots.  They seem to be becoming more accurate, though, so presumably an improvement.

edemaine (approved):
I finally reviewed this, testing on a Linux machine.  Everything looks great, in particular generating the exact same font files, except for two commits which I added already.  Could you review them and then merge?  Here is what I changed:

* The code did `cd "$(dirname "$0")/../.."` and then referred to the path `$(dirname $0)` again.  This doesn't work when the path is relative (e.g. `./buildFonts.sh` or even `dockers/font/buildFonts.sh`). I replaced the latter occurrences with `dockers/fonts`.
* There was an uncalled function that claimed to deal with arguments, but no arguments seem to actually be supported.  (We could support `-h`, but why bother if no arguments are supported. 🙂 )

In either case, feel free to replace with more suitable code.  Thanks for these improvements!

edemaine:
It seems that `buildFonts.sh` doesn't run `buildMetrics.sh`.  Do we not have a Dockerized `buildMetrics.sh`?  It would be nice to include, perhaps in a future PR.

edemaine:
Another update: Should we add scripts for `build:fonts` and `build:metrics` or something along those lines?  Having them in the `scripts` list makes it more obvious that they exist.

ylemkimon:
@edemaine Thank you very much!

edemaine:
Oops, maybe this should have been `chore` instead of `fix`, but not a big deal. 🙂 We'll have a bunch more releases with actual font-based features soon.

ylemkimon:
@edemaine Since fonts are updated, a patch release makes sense.

PR #2870, opened 2021-03-30
fix: Protect fraction bars from CSS border-color

### LINKED ISSUE #2731: KaTeX with Tailwind.css
**Describe the bug:**
When using tailwind.css as stylesheet, the fraction line and other related "border-color" are rendered greyed.

This is due to the fact that tailwind.css starts with:

```
*,
::before,
::after {
  box-sizing: border-box; /* 1 */
  border-width: 0; /* 2 */
  border-style: solid; /* 2 */
  border-color: #d2d6dc; /* 2 */
}
```

The simple solution is to add the followind line in a custom css:

```
.frac-line{border-color: black !important;}
```

Don't know if there is a cleaner or better way to do this.

thecodingwizard:
`border-color` is not inheritable (https://drafts.csswg.org/css-backgrounds-3/#propdef-border-color), so the previous fix (#2292), which applied `border-color` to `.katex`, did not work. This PR applies `border-color: currentColor` to `.katex *`.

closes #2731

### DIFF
--- a/src/katex.less
+++ b/src/katex.less
@@ -17,12 +17,14 @@
     // Prevent a rendering bug that misplaces \vec in Chrome.
     text-rendering: auto;
 
-    // Insulate fraction bars and rules from CSS that sets border-color.
-    border-color: currentColor;
-
-    // Prevent background resetting on elements in Windows's high-contrast
-    // mode, while still allowing background/foreground setting on root .katex
-    * { -ms-high-contrast-adjust: none !important; }
+    * {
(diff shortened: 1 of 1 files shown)


edemaine (approved):
Thanks for catching and fixing this!  I confirmed that the old rule does not work, and the new rule works, with the [Tailwind CSS definitions](https://github.com/KaTeX/KaTeX/issues/2236#issuecomment-610639777).

ronkok:
@thecodingwizard Thank you for fixing this. I was unaware of the nuance that you described.

PR #2877, opened 2021-04-03
fix: Correctly parse \ followed by whitespace

### LINKED ISSUE #2860: Error with single backslash at end of line
**Describe the bug:**
(Copying from [a fixed pandoc issue](https://github.com/jgm/pandoc/issues/4134):)

> In LaTeX you can use a backslash `\` followed by any whitespace to produce a regular space, as opposed to an inter-sentence space. The whitespace needn't be a space; it could be a line break.

KaTeX treats a single `\` at the end of the line as an error.

**To Reproduce:**
1. Go to https://katex.org
2. Type a single `\` followed immediately by a line break.
3. See `KaTeX parse error: Undefined control sequence: \ at position 1: \̲ ̲`.

**Expected behavior:**
It should produce a regular space instead of a parse error.

**Screenshots:**

Screenshot

**Environment:**
 - KaTeX Version: 0.13.0

**Additional context:**
MathJax allows a single backslash at the end of the line.

edemaine:
LaTeX parses `\` followed by whitespace including up to one newline as equivalent to `\ `.  (With multiple newlines, you get paragraph breaks. These aren't supported by KaTeX, but I figured it might one day...)

Fix #2860.


ronkok:
This looks correct to me.  I've also done some testing, which agrees with LaTeX behavior.

If I may pick the tiniest of nits, there is a comment on line 23 of Lexer.js which asserts that the RegEx matches whitespace using its *first* group. That should be amended to say it uses its *first two groups*.

edemaine:
@ronkok Thanks for taking a look at this. I had missed that block comment, and have updated it.

While looking at this one more time, I realized that the control word case can be simplified to avoid doing a second RegExp match, using groups.  This doesn't seem to have any impact on `test:perf` but I think it's cleaner.

src/Lexer.js
```diff
@@ -112,12 +110,6 @@ export default class Lexer implements LexerInterface {
             return this.lex();
         }
 
-        // Trim any trailing whitespace from control word match
-        const controlMatch = text.match(controlWordWhitespaceRegex);
-        if (controlMatch) {
-            text = controlMatch[1];
-        }
-
         return new Token(text, new SourceLocation(this, pos,
```
ronkok, line 113: It's nice that you could remove this.

src/Lexer.js
```diff
@@ -28,8 +28,9 @@ import type Settings from "./Settings";
  * - does not match bare surrogate code units
  * - matches any BMP character except for those just described
  * - matches any valid Unicode surrogate pair
- * - matches a backslash followed by one or more letters
- * - matches a backslash followed by any BMP character, including newline
+ * - matches a backslash followed by one or more whitespace characters
+ * - matches a backslash followed by one or more letters then whitespace
+ * - matches a backslash followed by any BMP character
  * Just because the Lexer matches something doesn't mean it's valid input:
```
ronkok, line 34: These comment changes are good, but line 23 is still incorrect.

edemaine: I think it's correct.  At least, I don't think `\ ` is a whitespace character; it's a command whose body is space, now covered by line 31.  Anyway, these comments are about the character stream, and "backslash followed by whitespace" (line 31) is different from "whitespace" (line 23).

ronkok: > "backslash followed by whitespace" (line 31) is different from "whitespace"

Point taken.

Sometimes a complicated RegEx pattern comes with comments that lay out in detail the capturing groups. As an example:

```
// Capturing groups:
//    [1] sign
//    [2] hexadecimal integer
//    [3] integer part
//    [4] numerator of a mixed fraction
//    [5] denominator of a mixed fraction
//    [6] decimal fraction of significand, not including decimal point
//    [7] exponent of a number in scientific notation
//    [8] percentage sign
```

The original comments for this RegEx included some of the same ordering, written less formally. I think that has value.

The pattern that you have introduced is in capturing group 2. Could you humor me and move the corresponding comments to be positioned just after line 23?

edemaine: Documenting the capturing groups seems like a great idea!  I've added that.

I hesitate a little to re-organize the top list because it follows a somewhat logical order of what we *want* to match.  Hopefully documenting the capturing groups makes the order we actually match things clearer?

I'm working on a second PR that will hopefully clean up the regexp order, but it requires a different way of handling `\operatorname` so I think it's best left for a separate PR.

ronkok (approved):
This looks great.

edemaine:
@ronkok Thanks for the reviews!

PR #2900, opened 2021-04-05
ci(release): do not include dist in release commit

ylemkimon:
They should not be included (#2385).


edemaine (approved):
Thanks for fixing!

Too bad this made it into the permanent Git history (unless we do a force-push).

ylemkimon:
@edemaine Actually, we were doing this prior to #2385, but just as a detached commit. Thank you for the review!

PR #2901, opened 2021-04-05
build(website): fix build script for Netlify

ylemkimon:


website/lib/build.js
```diff
@@ -44,7 +44,7 @@ if (process.env.NETLIFY_API_KEY) {
 
 // use KaTeX from CDN on the main page for Netlify production deploy
 if (process.env.CONTEXT === 'production') {
-    const version = require('../versions.json')[0];
+    const {version} = require('../../package.json');
```
edemaine, line 47: We used to set the `version` to the next version `-pre`. Is this no longer the case?  (It seems not, but wanted to check.)  Otherwise LGTM.

edemaine (approved):

PR #2902, opened 2021-04-05
chore: add pull request template

ylemkimon:


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,13 @@
+<!-- The title should folllow Angular Commit Message Conventions (https://github.com/angular/angular.js/blob/master/DEVELOPERS.md#-git-commit-guidelines) -->
```
edemaine, line 1: This is news to me, though I see it's in [semantic-release readme](https://github.com/semantic-release/semantic-release#commit-message-format). 

Do we actually use the footer aspect? (I had never seen it before.) I thought deciding the version number was done via the commit type, not BREAKING CHANGE. It is nice to have a place for `fixes #...` though.

I also feel like the scope list isn't great, but I'm not sure what scopes we want to list.

ylemkimon: @edemaine 
> Do we actually use the footer aspect? (I had never seen it before.) I thought deciding the version number was done via the commit type, not BREAKING CHANGE.

The footer decides whether the major version (minor for `v0.x.x`) should be bumped.

> I also feel like the scope list isn't great, but I'm not sure what scopes we want to list.

I think scope usually is the name of the file that is updated. IMHO, it wouldn't matter much since our repo is small, yet.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,13 @@
...
+
+**What is the previous behavior?**
+
+
+**What is the new behavior?**
```
edemaine, line 6: I wonder if we should have a general Summary / What this PR does before these two more specific questions?  Not sure, maybe the title is enough...

edemaine (approved):
Looking at this again, I think the before/after should suffice for summary. We can always add more fields later, but I think this is a good start.

PR #2903, opened 2021-04-05
docs: update CHANGELOG for v0.13.0

ylemkimon:


edemaine (approved):
Thanks for fixing this omission. Nice that it shouldn't happen anymore with automatic releases!

PR #2943, opened 2021-04-19
docs(support): remove excess backslash in CD example

edemaine:
Just noticed this when copy/pasting the example.  The example in `support_table` is already correct.


ronkok (approved):
Looks good. Thank you for catching this.

PR #2952, opened 2021-04-22
feat: \nonumber/\notag support, \tag per row of {align}

### LINKED ISSUE #2379: Enable \tag within AMS environments
The following AMS environments allow `\tag`(s) to be placed inside the environment, one tag for each line.

* {align}
* {align*}
* {alignat}
* {alignat*}
* {equation}
* {equation*}
* {gather}
* {gather*}

It would be good if KaTeX could enable the same thing.

### LINKED ISSUE #2950: Support \nonumber and \notag
Now that we have equation numbering, it would be nice to add `\nonumber` (or the AMS [equivalent](https://tex.stackexchange.com/questions/48965/what-is-the-difference-between-notag-and-nonumber-in-align-environment), `\notag`), so that we can only number some rows of an `{align}` environment (for example).

Hopefully not too hard: `\nonumber` just prevents the `eqn-num` generation for that row.

It's in AMSMath, as well as base LaTeX (for `eqnarray`, which we don't yet support).

edemaine:
Support `\nonumber` (and equivalent `\notag`) using a global macro `\@eqnsw` to track whether one occurs in each row, similar to how amsmath uses global `\@eqnswtrue`/`\@eqnswfalse`.  Fix #2950.

I also added support for `\tag` on each row of an `{align}` or `{align*}` or similar environment, so each row is either automatically numbered (represented by `true`), not automatically numbered (represented by `false`), or has an explicit tag (represented by an array of parse nodes). Fix #2379.

This required a more general mechanism for expanding a macro (`\df@tag`) that doesn't actually appear in the source code.  We previously did this in `parseTree.js` only after parsing everything, but now we need it mid-parse too.  The new `Parser.subparse` does the job.

I also had to add a mechanism for undefining a macro value.  TeX represents this as setting a macro to `\relax`, but `undefined` was already used in some places in `Namespace.js` so I used that.

I added basic documentation and tests.

Example:

image

The tags are center-justified instead of the correct right-justified, but that was an existing bug that we can look at in another PR.


edemaine:
Hmm, perhaps I should try to implement `\tag`-per-row at the same time; this should be doable with a similar (if not the same) mechanism.

edemaine:
OK, I've added support for `\tag` in rows. This was harder than expected, but not too bad. See root post for details. Now ready for review.

edemaine:
I fixed the merge conflicts here, so this is ready to review again. FWIW, we've had a recent explicit request for this feature (#3183), and #2379 has 5 thumbs-up. PR #3214 also currently depends on this one. Let me know if you'd rather I split this up into multiple PRs.

src/environments/array.js
```diff
@@ -633,7 +675,11 @@ const alignedHandler = function(context, args) {
         {
             cols,
             addJot: true,
-            addEqnNum: context.envName === "align" || context.envName === "alignat",
+            autoTag:
+              context.envName === "align" || context.envName === "alignat" ?
+              true :
+              context.envName === "align*" || context.envName === "alignat*" ?
+              false : undefined,
```
ylemkimon, line 682:
```suggestion
            autoTag: getAutoTag(context.envName),
```

src/environments/array.js
```diff
@@ -982,7 +1028,8 @@ defineEnvironment({
             }],
             addJot: true,
             colSeparationType: "gather",
-            addEqnNum: context.envName === "gather",
+            autoTag: context.envName === "gather" ? true :
+                     context.envName === "gather*" ? false : undefined,
```
ylemkimon, line 1032:
```suggestion
            autoTag: getAutoTag(context.envName),
```

src/environments/array.js
```diff
@@ -1015,7 +1062,7 @@ defineEnvironment({
     handler(context) {
         validateAmsEnvironmentContext(context);
         const res = {
-            addEqnNum: context.envName === "equation",
+            autoTag: context.envName === "equation",
```
ylemkimon, line 1065:
```suggestion
            autoTag: getAutoTag(context.envName),
```

src/environments/array.js
```diff
@@ -4,10 +4,12 @@ import Style from "../Style";
...
 import {parseCD} from "./cd";
 import defineFunction from "../defineFunction";
+import defineMacro from "../defineMacro";
 import mathMLTree from "../mathMLTree";
 import ParseError from "../ParseError";
 import {assertNodeType, assertSymbolNodeType} from "../parseNode";
 import {checkSymbolNodeType} from "../parseNode";
+import {Token} from "../Token";
 import {calculateSize} from "../units";
 import utils from "../utils";
```
ylemkimon, line 15:
```suggestion

const getAutoTag = name => name.indexOf("ed") === -1 ? name.indexOf("*") === -1
    : undefined;
```
(below `validateAmsEnvironmentContext`)

src/Parser.js
```diff
@@ -150,6 +150,27 @@ export default class Parser {
...
+     * Tokens should be specified in reverse order, as in a MacroDefinition.
+     */
+    subparse(tokens: Token[]): AnyParseNode[] {
+        // Save the next token from the current job.
+        const oldToken = this.nextToken;
+        this.consume();
+
+        // Run the new job, terminating it with an excess '}'
+        this.gullet.feedTokens([new Token("}")]);
+        this.gullet.feedTokens(tokens);
```
ylemkimon, line 164:
```suggestion
        this.gullet.pushToken(new Token("}"));
        this.gullet.pushTokens(tokens);
```

src/MacroExpander.js
```diff
@@ -53,6 +53,16 @@ export default class MacroExpander implements MacroContextInterface {
...
+    /**
+     * Feed additional input as a set of token, specified in reverse order
+     * (e.g. from a MacroDefinition).
+     */
+    feedTokens(tokens: Token[]) {
+        for (let i = 0; i < tokens.length; i++) {
+            this.stack.push(tokens[i]);
+        }
+    }
+
```
ylemkimon, line 65:
```suggestion
```

ylemkimon (changes requested):
LGTM. Sorry for the delayed review. I think `subparse` logic can be improved later.

edemaine:
@ylemkimon Thanks for the nice improvements! Should be ready to look at again.

ylemkimon (approved):

PR #2960, opened 2021-04-24
ci: set necessary token permissions

ylemkimon:
I've changed [the default permission of the token to read-only](https://github.blog/changelog/2021-04-20-github-actions-control-permissions-for-github_token/) and these are the only places we need write access.


edemaine (approved):
LGTM. I guess if we missed some places where we need write access, we'll notice. 🙂

PR #2972, opened 2021-05-01
docs: mention wrapping in inline mode

Artoria2e5:
KaTeX allows line-wrap in inline mode, which is actually a really nice behavior when longer inline equations are involved. I think it's a feature worth advertising, especially since MathJax doesn't have it (wink).

The text proposed here is pretty rough, so any copyediting suggestions are welcome. I added the parenthetical about CSS customization because... well I did it in the [infamous Wikipedia-KaTeX script](https://phabricator.wikimedia.org/T99369#6970581) and found it useful, mainly for matching Wikipedia's own block equation style.


docs/options.md
```diff
@@ -4,7 +4,9 @@ title: Options
 ---
 You can provide an object of options as the last argument to [`katex.render` and `katex.renderToString`](api.md). Available options are:
 
-- `displayMode`: `boolean`. If `true` the math will be rendered in display mode, which will put the math in display style (so `\int` and `\sum` are large, for example), and will center the math on the page on its own line. If `false` the math will be rendered in inline mode. (default: `false`)
+- `displayMode`: `boolean`. If `true` the math will be rendered in display mode. If `false` the math will be rendered in inline mode. (default: `false`) Differences between the two modes include:
+  - Display mode puts math in display style, so `\int` and `\sum` are large, for example;
```
edemaine, line 8:
```suggestion
  - Display mode starts in `\displaystyle`, so `\int` and `\sum` are large, for example; while inline mode starts in `\textstyle`, where subscripts and superscripts usually don't stack on top of operators like `\sum`. You can always manually switch between `\displaystyle` and `\textstyle` using those commands.
```

docs/options.md
```diff
@@ -4,7 +4,9 @@ title: Options
...
+  - Display mode centers math on its on line and disables line-wrap (you can customize this behavior with CSS).
```
edemaine, line 9:
```suggestion
  - Display mode centers math on its on line and disables line-wrap (though you can customize this behavior with custom CSS). In inline mode, KaTeX allows line breaks after outermost relations (like `=` or `<`) or binary operators (like `+` or `\times`), the same as TeX.
```

docs/options.md
```diff
@@ -4,7 +4,9 @@ title: Options
 ---
 You can provide an object of options as the last argument to [`katex.render` and `katex.renderToString`](api.md). Available options are:
 
-- `displayMode`: `boolean`. If `true` the math will be rendered in display mode, which will put the math in display style (so `\int` and `\sum` are large, for example), and will center the math on the page on its own line. If `false` the math will be rendered in inline mode. (default: `false`)
+- `displayMode`: `boolean`. If `true` the math will be rendered in display mode. If `false` the math will be rendered in inline mode. (default: `false`) Differences between the two modes include:
```
edemaine, line 7:
```suggestion
- `displayMode`: `boolean` (default: `false`). If `true` the math will be rendered in display mode. If `false` the math will be rendered in inline mode. Differences between the two modes include:
```

edemaine (changes requested):
Nice idea to better document this!  Your text was actually quite well-written, but I made a few suggested changes that clarify things a bit and are a bit more symmetric I think.

It's a little weird to mention a custom CSS hack and not give it.  Perhaps it would be worth adding to [Common Issues](https://katex.org/docs/issues.html)?  I think another reasonable option to mention in that context is `.katex-display { overflow: auto hidden }` which makes individual display equations scrollable horizontally (as discussed in #2942).

docs/issues.md
```diff
@@ -49,3 +49,32 @@ the version of the JavaScript file (katex.js), which is defined in `katex.versio
...
+.katex-display {
+  margin-left: 1.6em;
+  text-align: initial;
+  text-align: inherit;
+}
+
+.katex-display > .katex {
+  text-align: inherit;
+}
+```
```
edemaine, line 80: I don't think it makes sense to include the last example.  KaTeX has  the `fleqn` option which achieves almost exactly this (2em left margin according to https://katex.org/docs/options.html).

Artoria2e5: Oh!

edemaine (approved):

edemaine:
@Artoria2e5 Thanks again for the PR!

PR #2976, opened 2021-05-02
fix: binom delimiter size in scriptscriptstyle.

### LINKED ISSUE #2973: \binom parentheses too big in \scriptscriptstyle
> Before reporting a bug

**Describe the bug:**
The `\binom` in ` \sum_{\vec a\in \mathbb{N}^{\binom {d}{2}}}` is way too tall.

**(La)TeX code:**
The code of (La)TeX you tried to render:
```latex
\displaystyle \sum_{\vec a\in \mathbb{N}^{\binom {d}{2}}}
```

The snippet is minimized from https://en.wikipedia.org/wiki/Solid_angle#Solid_angles_in_arbitrary_dimensions. (Okay it can actually get smaller.)

**Expected behavior:**
quicklatex

(Oops, GitHub made it transparent! Switch to day mode please. Or just open image in new tab.)

**Screenshots:**
Screen Shot 2021-05-01 at 18 22 29

 - KaTeX Version: "0.13.3"
 - Device: Desktop
 - OS: macOS 11
 - Browser: Cr 
 - Version: 90.0.4430.93

ronkok:
Fixes #2973.

Fun fact: This bug has existed since version 0.1.0.


edemaine:
This looks like a great improvement!  Here's the example from #2973 rendered with this PR:

image

Because I wasn't sure what "correct" behavior was, I ran our old texcmp script to see how the rendering compared to LaTeX (green):

image

It looks like LaTeX uses slightly shorter brackets in both script and scriptscript styles.  However, it does seem to use the same brackets for the two modes, so I believe this PR is correct.

ronkok:
This PR has now passed all tests and should be ready to go.

edemaine:
@ronkok Did you see my post above? Are you suggesting we save fixing the paren size in `\scriptstyle` for another time?

ronkok:
Yes, I did see your post and I am indeed suggesting that the paren in scriptstyle is a different problem, to be solved with a different PR. That is, if anybody bothers with it. It isn't much of a problem.

I think this PR is ready to go.

edemaine (approved):
Sounds good. I'll open a separate issue.

Issue #2876, opened 2021-04-03
Meta-request: mediawiki Math parity


Artoria2e5:
**Is your feature request related to a problem? Please describe.**
I am trying to [use KaTeX on Wikipedia for fun](https://phabricator.wikimedia.org/T99369#6970581). They have kicked away MathJax because of the server load and the speed issues, but I figured a KaTeX userscript won't hurt. After all, what's not to love about having actual text rendered instead of some SVG or even PNG image?

**Describe the solution you'd like:**
KaTeX should add support for everything TeX mentioned in the current [Wikipedia formula help page](https://en.wikipedia.org/wiki/Help:Displaying_a_formula). The errors are:
```
KaTeX parse error: Limit controls must follow a math operator at position 58: …ratorname {sn} \̲l̲i̲m̲i̲t̲s̲ ̲_{b>c}(b+c)
KaTeX parse error: Undefined control sequence: \S at position 63: …Re ,\circledS ,\̲S̲ ̲,\P ,\mathrm {\…
KaTeX parse error: Undefined control sequence: \P at position 9:  \amalg \̲P̲ ̲\S \%\dagger \d…
KaTeX parse error: Undefined control sequence: \sideset at position 2:  \̲s̲i̲d̲e̲s̲e̲t̲ ̲{_{1}^{2}}{_{3}…
KaTeX parse error: Undefined control sequence: \iiiint at position 2:  \̲i̲i̲i̲i̲n̲t̲ ̲\limits _{F}dx\…
KaTeX parse error: Undefined control sequence: \definecolor at position 2:  \̲d̲e̲f̲i̲n̲e̲c̲o̲l̲o̲r̲ ̲{myorange}{rgb}…
```

**Link to or name of a (La)TeX package that provides the same feature:**
* I have no idea what the \limits devil is.
* \S and \P are base LATEX 2ε, I think. They can be mapped to unicode chars.
* \sideset is from amsmath.
* \iiiint from esint is covered in #2059. 
* \definecolor from xcolor is covered in #750.

**Describe alternatives you've considered:**
Changing the existing formulae is unfortunately not acceptable.

**Additional context:**

edemaine:
Neat request.  Is Wikipedia `<math>` running actual `latex` to produce SVG, or some other parser like MathJax? MathJax doesn't seem to support `\P` so I doubt it's that, but LaTeX should support `\operatorname*` so I'm confused.

[...]

PR #2977, opened 2021-05-02
fix: Support \S and \P in math mode.

ronkok:
A fix to part of #2876.

Unicode characters "§" and "¶" now return math mode atoms, not text mode.


edemaine (approved):
Nice fix.  I tested that e.g. `§¶\S\P+\text{§¶\S\P}` renders as

image

And this matches LaTeX's output for `\S\P\S\P+\text{\S\P\S\P}`

PR #2979, opened 2021-05-02
fix: Avoid crash when \operatorname has \limits.

ronkok:
Fixes part of #2876.


edemaine (approved):
Thanks for fixing this!

PR #2984, opened 2021-05-04
feat: \operatornamewithlimits (and clean up \operatorname support)

edemaine:
My main motivation for this PR is to simplify the lexer to not care about `\operatorname`, by using KaTeX's existing `\@ifstar` macro to define `\operatorname`.

As a byproduct, it was natural to add support for `\operatornamewithlimits` (an alias to `\operatorname*` defined in `amsmath/amsopn.sty`).

~~I built this on top of #2877 (see just [the last commit](https://github.com/KaTeX/KaTeX/pull/2984/commits/59d9be519c85acd28951462e661125386b71bd73)) so that should be merged first.~~ Now that #2877 is merged, I've rebased this and it should be good to go.


edemaine:
Now that #2877 is merged, I've rebased this and it should be good to go.

This is really more of a cleanup than a feature release, so I wonder whether the `feat` heading is appropriate...

docs/support_table.md
```diff
@@ -801,6 +801,7 @@ use `\ce` instead|
 |\ominus|$\ominus$||
 |\operatorname|$\operatorname{asin} x$|`\operatorname{asin} x`|
 |\operatorname\*|$\operatorname*{asin}\limits_y x$|`\operatorname*{asin}\limits_y x`|
+|\operatornamewithlimits$\operatornamewithlimits{asin}\limits_y x$|`\operatornamewithlimits{asin}\limits_y x`|
```
ronkok, line 804: This does not display properly. It needs one more pipe to separate the cells.

src/functions/operatorname.js
```diff
@@ -159,3 +160,5 @@ defineFunction({
     mathmlBuilder,
 });
 
+defineMacro("\\operatorname",
+  "\\@ifstar\\operatornamewithlimits\\operatorname@");
```
ronkok, line 164: It had not occurred me that it is possible to call `defineMacro()` from a file other than `macros.js`. Intriguing.

Is it wise, though? When I look for a macro, I always look in `macros.js`. I'm not sure that it would occur to me to grep all the source code for `defineMacro`. And thus I could be misled about whether some macro exists.

ronkok (changes requested):

src/functions/operatorname.js
ronkok: On the other hand, this approach certainly aligns with the [spirit of the functions folder](https://github.com/KaTeX/KaTeX/pull/766#issuecomment-322648814). Maybe I just need a day to think and let it become familiar.

edemaine: For what it's worth, I did a similar thing in not-yet-merged #2952 (though maybe it feels slightly more comfortable in environments over functions?). In both cases, I feel like it's worth keeping logically clustered content together. 

A smaller example is that `environments/array.js`/`cd.js` redefines the `\cr` macro (though not using `defineMacro`).

In general I feel like `macros.js` is getting rather large and would probably be nicer to split it into logical components. The first chunk about defining macros for the sake of macro definitions makes sense, but the definition of various symbols, `\not`, `\dots`, etc. feel like they could belong elsewhere.  What if we moved those to the `functions` directory?  Or maybe it would be better to rename `functions` to `commands` and make explicit that commands can be defined by functions, macros, or a mixture thereof?

docs/support_table.md
```diff
@@ -801,7 +801,7 @@ use `\ce` instead|
 |\ominus|$\ominus$||
 |\operatorname|$\operatorname{asin} x$|`\operatorname{asin} x`|
 |\operatorname\*|$\operatorname*{asin}\limits_y x$|`\operatorname*{asin}\limits_y x`|
-|\operatornamewithlimits$\operatornamewithlimits{asin}\limits_y x$|`\operatornamewithlimits{asin}\limits_y x`|
+|\operatornamewithlimits$\operatornamewithlimits{asin}\limits_y x$|`\operatornamewithlimits{asin}\limits_y x`||
```
ronkok, line 804: This still needs work. It now displays with the function name and the rendered version in the same cell. What you need is:

```
|\operatornamewithlimits | $\operatornamewithlimits{asin}\limits_y x$|`\operatornamewithlimits{asin}\limits_y x`|
```

edemaine: Oops, sorry, I was clearly looking at this too quickly!

ronkok (approved):
This looks good.

PR #2993, opened 2021-05-06
fix: Update Safari screenshot.

ronkok:
Two pending PRs are failing because Screenshotter does not like a particular Safari image. This PR adopts the preferred image in hopes that this will enable the other two PRs to pass.


ronkok:
No good. It was worth a try.

edemaine:
@ronkok I find that I just have to rerun the screenshotter tests one or two times before the Safari tests pass. ☹️

ronkok:
How do I force screenshotter to rerun?

ronkok:
Ah, I now see the rerun button. Thanks for the tip.

PR #3006, opened 2021-05-11
fix: \char support for >16-bit Unicode characters

### LINKED ISSUE #3004: Unicode characters in astral code points are incorrectly rendered
**Describe the bug:**
The `\char` command incorrectly truncates to the lowest 16 bits when given an astral code point (that is, a Unicode code point of the form U+1hhhh).

**To Reproduce:**
1. Go to https://katex.org/#demo.
2. Input the TeX string `\char"1d7d9`.

The typeset character is actually U+D7D9 ([ퟙ](https://codepoints.net/U+D7D9)), as can be confirmed by inputting the TeX string `\char"d7d9` instead.

**Expected behavior:**
The Unicode character at code point U+1D7D9 (𝟙) to be typeset.

**Screenshots:**
image

**Additional context:**
Support for astral code points would allow for a workaround for https://github.com/KaTeX/KaTeX/issues/473, as "𝟙" and other double-struck digits have been assigned Unicode code points.

edemaine:
TIL [`String.fromCharCode`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/fromCharCode) only supports Unicode characters up to 16-bit. For example, `String.fromCharCode(0x1d7d9) === String.fromCharCode(0xd7d9)` (issue #3004).

[`String.fromCodePoint`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/fromCodePoint) from ECMAScript 2015 supports all Unicode code points: `String.fromCodePoint(0x1d7d9) !== String.fromCodePoint(0xd7d9)`.

Unfortunately, IE doesn't support the latter, so this PR uses the former as fallback. We could implement/include a [polyfill](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/fromCodePoint#polyfill), but IE seems sufficiently unimportant to not fix this bug for...?  However, note that `src/wide-character.js` uses a workaround for `codePointAt` so perhaps these are worth implementing and moving to `utils`.

Fixes #3004


edemaine:
Hmm, `\char"1d7d9` still doesn't actually render, because it's out of range of what we allow:

image

Should we add another block to `src/unicodeScripts.js`? Maybe all of [Supplementary Multilingual Plane](https://www.compart.com/en/unicode/plane/U+10000)?

ronkok:
Dropping support for IE would be okay with me. I believe that @ylemkimon has already advocated that step previously.

I wrote `wideCharacterFont()` specifically to support wide characters that have glyphs in the KaTeX fonts. But the branch to `wideCharacterFont()`  captures other characters, as well. We could narrow the criterion that branches to `wideCharacterFont()` and then I think those characters could be treated like other Unicode characters.

edemaine:
System fonts should automatically act as a fallback for *rendering* (and we could give an explicit list of backup system fonts in the CSS if we wanted).

The issue here is in computing the *vertical metrics* of this character. By specifying a font of `""`, the font metric lookup crashes. It should probably generate a warning (and assume `M` height) instead.

ylemkimon (approved):

Issue #3012, opened 2021-05-12
Safari screenshotter often breaks


edemaine:
Most PRs (and when I merge to master branch) requires clicking "re-run tests" 2-3 times to get the Safari screenshotter tests to work.

It seems to fail on the `Includegraphics` test most often, perhaps because of the web fetch it needs to do. Perhaps we should remove that test, ~~or make the URL it loads a local file on the tester~~? Oops, it's already a local file. I'm not sure why Safari fails at this -- perhaps a race condition?

ylemkimon:
I think we should disable `Includegraphics` test on Safari.

edemaine:
Ah, that would be a simpler solution. 🙂 Screenshotter already has an `-x`/`--exclude` option, so this should be easy to do.

PR #3015, opened 2021-05-12
chore(screenshotter): Improve screenshotter for Safari

edemaine:
This PR does two things, and fixes or improves #3012:

1. ~~Upgrade to Safari 14.0; seems good to track modern Safari~~ I accidentally did this only locally, not in CI. I reverted this change because I didn't want to deal with changed screenshots in this PR.
2. Fix running Docker on IPv6 system. This is an aside I guess, but was necessary in my Windows environment; unfortunately, I still couldn't get the screenshotter working on Windows; it seems to just hang while trying to connect to the Selenium docker.
3. Ensure images load before screenshotting, the same way we do with fonts. This is presumably what was causing the `Includegraphics` test to fail on Safari (dunno why it works on other browsers; maybe they load local files fast).

CI doesn't crash on `Includegraphics` tests anymore.  I've seen Safari crash with a "failure to find function `load_fonts_and_images`" error and a timeout error.  I still haven't seen Chrome or Firefox crash, so this PR doesn't seem to have broken anything. I think Safari tests succeed more often now but there might be work left to do.

### DIFF
--- a/test/screenshotter/test.html
+++ b/test/screenshotter/test.html
@@ -31,8 +31,8 @@
     <span id="math"></span>
     <span id="post"></span>
     <script type="text/javascript">
-      function load_fonts(callback) {
-        if (!document.fonts || !document.fonts.load) {
+      function load_fonts_and_images(callback) {
+        if (!document.fonts || !document.fonts.load || !document.images || !document.images.load) {
           callback();
           return;
         }
@@ -55,6 +55,17 @@
         }
         load_font("Mincho");
         load_font("Batang");
+
+        document.images.forEach(function (img) {
+          results.push(new Promise(function (done, error) {
+            if (img.complete) {
+              done();
+            } else {
+              img.addEventListener('load', done);
+              img.addEventListener('error', error);
+            }
+          }));
+        });
 
         Promise.all(results).then(callback)
       }
--- a/test/screenshotter/test.html
+++ b/test/screenshotter/test.html
@@ -32,6 +32,11 @@
     <span id="post"></span>
     <script type="text/javascript">
       function load_fonts_and_images(callback) {
+        if (typeof Promise === "undefined") {
+          callback();
+          return;
+        }
+
         var results = [];
 
         function load_font(name, style) {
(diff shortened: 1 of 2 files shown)


test/screenshotter/test.html
```diff
@@ -31,8 +31,8 @@
     <span id="math"></span>
     <span id="post"></span>
     <script type="text/javascript">
-      function load_fonts(callback) {
-        if (!document.fonts || !document.fonts.load) {
+      function load_fonts_and_images(callback) {
+        if (!document.fonts || !document.fonts.load || !document.images || !document.images.load) {
```
ylemkimon, line 35:
```suggestion
        if (!document.fonts || !document.fonts.load) {
```
`document.images` is available on all browsers and there is no `document.images.load`.

test/screenshotter/test.html
```diff
@@ -56,6 +56,17 @@
         load_font("Mincho");
         load_font("Batang");
 
+        document.images.forEach(function (img) {
```
ylemkimon, line 59: AFAIK, there is no `forEach` method on `document.images` (`HTMLCollection`).

edemaine:
@ylemkimon Thanks for spotting those bugs which caused the code to never run! (and broke font loading) I tweaked the code to always do the image loading checks, correctly this time, and no longer break font loading.

I've run the tests three times, and while Safari doesn't always get the Includegraphics test in the first try, it so far hasn't actually failed within the 5 attempts. Anyway, I think this PR should help and can't hurt, even if we remove the test for Safari.

test/screenshotter/test.html
```diff
@@ -31,14 +31,11 @@
...
     <span id="post"></span>
     <script type="text/javascript">
-      function load_fonts(callback) {
-        if (!document.fonts || !document.fonts.load) {
-          callback();
-          return;
-        }
-
+      function load_fonts_and_images(callback) {
         var results = [];
```
ylemkimon, line 35:
```suggestion
        if (typeof Promise === "undefined") {
          callback();
          return;
        }

        var results = [];
```

ylemkimon (approved):

ylemkimon:
Thank you for the PR! The Safari screenshotter always bothered me.

edemaine:
@ylemkimon Thanks for the review and improvement!

PR #3030, opened 2021-05-18
test(screenshotter): update Includegraphics-safari

### LINKED ISSUE #3012: Safari screenshotter often breaks

ylemkimon:
Fixes #3012.


edemaine (approved):
Nice catch! Looks like the old version had a low-resolution partial load of the image. This was accidentally introduced in #2976 when the test changed.

PR #3032, opened 2021-05-19
test(screenshotter): update screenshots

ylemkimon:
They were updated in the wrong directory in #2999.


edemaine (approved):
Weird. A bug in our CI?

Issue #1366, opened 2018-05-27
Incorrect rendering of \^ in text mode


pzinn:
Try
```
\text{a\^{}b}
```
in KaTeX vs LaTeX. In the latter `\^{}` is a proper caret character, whereas in the former.. well, I'm not quite sure what's going on, maybe it's only a spacing issue, but the result certainly looks wrong.
(Replacing `\text` with `\texttt` may help see the problem)

edemaine:
Interesting, thanks for spotting. I need to understand when accented characters take the width of the accent (perhaps empty base character is a special case?) and when they take the width of the base character (as in `\^\i`).

pzinn:
any progress on this?

kevinbarabash:
@edemaine it looks like it's a `text` mode thing.  Here's what quicklatex.com produces:

```\text{a^{}b}```
<img width="123" alt="Screen Shot 2020-12-11 at 5 05 29 PM" src="https://user-images.githubusercontent.com/1044413/101959367-30ee6480-3bd3-11eb-921b-8b9bfaa67842.png">

```a^{}b```
<img width="81" alt="Screen Shot 2020-12-11 at 5 05 18 PM" src="https://user-images.githubusercontent.com/1044413/101959378-38ae0900-3bd3-11eb-97d7-ccb560fe36b2.png">

kevinbarabash:
The second one is actual an error.  I didn't notice it at first because quicklatex.com doesn't enable error reporting by default.

edemaine:
I hadn't realized that `\^` is equivalent to `\textasciicircum`. I'll make a quick PR!

edemaine:
`\^` is a text-mode accent command. The issue is with accent commands having empty arguments. In KaTeX, a blank argument is treated like a zero-width box, so the accent gets centered on top.  In LaTeX, a blank argument seems to behave the same as something like a space, but I'm not sure by what mechanism it does so (minimum width? special case of `{}` argument?). Perhaps this is done at the TeX level? Without knowing the mechanism, it's difficult to implement...

edemaine:
[...]

PR #3036, opened 2021-05-20
feat: text-mode cedilla accent via \c

edemaine:
Add text-mode accent function `\c` and corresponding Unicode support.

This started as PR #1127 by @kevinbarabash but for technical reasons got closed for a future replacement PR. I happened to bump into it today (via #638), and revitalized the code into a new PR. Conveniently, the recently merged PR #3009 resolved the technical issue that was holding this back.

Solves part of #638.

Demo:

image

This PR changes the rendering of the Unicode glyphs `çÇ` to use the correct `cC` glyphs plus an accent.  Screenshot diff:

image


kevinbarabash:
@edemaine thanks for tackling this.  Something's not quite right though.  The new glyphs look quite different from the base glyphs.

**BEFORE**
Screen Shot 2021-08-28 at 12 10 52 PM

**AFTER**
Screen Shot 2021-08-28 at 12 10 57 PM

QuickLatex rendering of base glyphs beside those using `\c`:
Screen Shot 2021-08-28 at 12 12 24 PM

edemaine:
Are you sure you didn't reverse those?  I just tested katex.org vs. the netlify deploy preview, and this is what I get for `\huge\text{cCçÇ}`:

Old / katex.org: 

image

New / deploy preview:

image

kevinbarabash:
> Are you sure you didn't reverse those?

My bad.  I had them reversed, sorry about that. 😞

src/functions/accent.js
```diff
@@ -74,10 +74,14 @@ export const htmlBuilder: HtmlBuilderSupSub<"accent"> = (grp, options) => {
...
+
     // calculate the amount of space between the body and the accent
-    let clearance = Math.min(
-        body.height,
-        options.fontMetrics().xHeight);
+    let clearance = accentBelow
+        ? body.height + body.depth
+        : Math.min(
+            body.height,
+            options.fontMetrics().xHeight);
```
kevinbarabash, line 84: The rendering looks good, but the fact that we have kern of size `-clearance` below is a bit confusing.  No action req'd since it's been like this for a while but something we might want to refactor in the future to be less confusing.

test/unicode-spec.js
```diff
@@ -9,6 +9,12 @@ describe("unicode", function() {
             .toBuild();
     });
 
+    it("should build Latin-1 inside \\text{} like accent commands", function() {
+        expect`\text{ÀÁÂÃÄÅÈÉÊËÌÍÎÏÑÒÓÔÕÖÙÚÛÜÝàáâãäåèéêëìíîïñòóôõöùúûüýÿÇç}`
+            .toParseLike`\text{\`A\'A\^A\~A\"A\r A\`E\'E\^E\"E\`I\'I\^I\"I\~N\`O\'O\^O\~O\"O\`U\'U\^U\"U\'Y\`a\'a\^a\~a\"a\r a\`e\'e\^e\"e\`ı\'ı\^ı\"ı\~n\`o\'o\^o\~o\"o\`u\'u\^u\"u\'y\"y\c C\c c}`;
+        // TODO(edemaine): A few characters don't have analogs yet.
+    });
```
kevinbarabash, line 16: Nice test.

kevinbarabash (approved):
LGTM.  I don't see where the mapping is from the unicode chars to the `\c C`, `\c c` expansion.  Is there some code that was automatically handling this for all entries in unicodeAccents.js?

edemaine:
Thanks for the review!

> I don't see where the mapping is from the unicode chars to the `\c C`, `\c c` expansion. Is there some code that was automatically handling this for all entries in unicodeAccents.js?

Yep, that is `src/unicodeSymbols.js` (which is inlined via `/*preval*/` into `Parser.js`).

src/functions/accent.js
edemaine: I agree that it's confusing; I can't say I fully understand this code. Happy to review a PR that cleans it up. 🙂 There are some actual bugs here to fix, e.g., #1366 (though that has to do with horizontal rather than vertical spacing).

edemaine:
@ylemkimon This got merged but didn't trigger a release (apparently the job got canceled somehow). The next merge triggered a release, but this PR isn't mentioned in the changelog.  Any ideas what happened?

PR #3053, opened 2021-05-28
fix: Correct invalid box-sizing property

### LINKED ISSUE #3052: Invalid box-sizing property in .katex .angl css class
**Describe the bug:**

As seen [here](https://github.com/KaTeX/KaTeX/blob/2f62c0d8ee8135f4d5b7fe727add6ad25d5c86a0/src/katex.less#L596) the `.angl` class contains a non-existent property value for `border-box`. Judging from the context it should be `content-box`. I'm happy to open a PR if anyone can confirm that it should indeed be `content-box` and not `border-box`.

```
.angl {
    // from package actuarialangle, which is always used in a subscript.
    box-sizing: border-content;       // NON-EXISTENT css property
    border-top: 0.049em solid;        // defaultRuleThickness in scriptstyle
    border-right: 0.049em solid;      // ditto
    margin-right: 0.03889em;          // 1 mu
}
```

**To Reproduce:**
NA

**Expected behavior:**
Stylesheets to include [W3C valid properties](https://drafts.csswg.org/css-ui-3/#propdef-box-sizing): `content-box | border-box`.

klittlepage:
Fixes #3052


edemaine (approved):
Thanks for finding and fixing this! I took care of the Safari screenshot issue (which requires actions by one of the maintainers &mdash; nothing you could have done on your own).

PR #3114, opened 2021-07-12
fix: remove local macros upon parse error

edemaine:
Close all groups after parse, in particular in case of parse error, completing `Namespace`'s simulation of local definitions.

Fixes #3122 as closing groups deletes any local-to-group macros.


ylemkimon (approved):

edemaine:
@ylemkimon Thanks for all the reviews!  I assume you're holding releases somehow (manually canceling the release jobs?).

PR #3133, opened 2021-07-24
fix: stop line wrap immediately after inline math

### LINKED ISSUE #1233: Line breaks sometimes happen immediately after an inline KaTeX block

ekzhang:

Line breaks would occur immediately after inline math, even if there is punctuation, which cuts it off at the next line. This is stylistically undesirable. For example, @edemaine mentions using a parser hack to resolve this problem by moving punctuation into the math block.

image

Lines no longer soft-wrap immediately after an inline math block. To verify the new behavior, you can use the following example in the development server after running `yarn start`:

[http://localhost:7936/?display=0&text=a-b-c-d-e-f-g&before=this%20is%20a%20paragraph%20where%20&after=,%20and%20something%20else.](http://localhost:7936/?display=0&text=a-b-c-d-e-f-g&before=this%20is%20a%20paragraph%20where%20&after=,%20and%20something%20else.)

Open the page, then resize your browser window gradually until the comma is at the right edge of the screen. You'll see that as your screen gets just small enough, the comma will not wrap to the next line by itself, which was the previous behavior. Instead, it will take the last math character "g" with it.

Closes #1233.


ekzhang:
It looks like screenshot tests aren't working for Firefox and Safari - could you please advise on what to do here? @edemaine

edemaine:
@ekzhang If you click on [the check](https://github.com/KaTeX/KaTeX/pull/3133/checks?check_run_id=3151989980), you'll see 4 artifacts in the top right. Download those. You'll have the new screenshots for the changed tests (`Linebreak`) and diffs to show what's changed. You'll also want to explain why these tests have changed -- ideally it's an improvement in behavior, not a regression -- if it's a regression, you probably want to tweak your code more to get these tests to pass.

ekzhang:
Thanks for the help. Here are the diffs.

### Firefox

It looks like a slight change in where the line break happens. In the old render, the equation was very slightly too small for the margin, but in the new render, it barely is able to fit. I think both behaviors seem reasonable, offset by very little?

image

### Safari

I'm less sure about what's going on here. There's two diffs, Boxed and Units.

image

image

It looks like everything is just shifted down by one line.

ekzhang:
Does this screenshotter diff look acceptable?

jonsterling:
I wonder if we can bring some attention to this PR again? I am having this problem and it would be great to have a fix merged.

ekzhang:
I'm happy to look at this PR and fix any issues. Unfortunately I don't have the knowledge of the codebase to understand what's the issue with the screenshot tests right now and how I should fix it.

fweth:
According to [this](https://stackoverflow.com/a/61631199/1020237) answer, another thing that could be done is to wrap all text in spans (as far as I understand, currently inline-math gets placed inside a span while the text around not). But I don't know if that would be feasible for the use-case of KaTeX?

edemaine:
I just learned about the [`<wbr>` HTML element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/wbr). I wonder if this would be better/cleaner than a span with a zero-width space inside.

ajakaja:
What's the status on this? I'd be interested in picking it up and getting it landed if it has otherwise stagnated, although it'll be my first time contributing to KaTeX so... no promises, I guess.

fweth:
Not claiming that this is a good solution, but I fixed this on my personal website with this line of CSS:

```css
.katex .base:last-child {
  display: inline;
}
```

Hamarkhis:
Hello everyone,

I've been facing this issue and have now time to try to find a solution.

I first tried the CSS solution submitted by fweth and others but that's not working for me.

I cloned the [ekzhang](https://github.com/ekzhang) fork entirely with git, and then I build the files with yarn. I put manually katex on my site this way. That's not working.

I tried to put the latest version of Katex 0.16.11, the issue is still there.

At last I've just tried Katex 0.16.11 and manually modiying the files src/buildHTML.js and src/buildTree.js like ekzhang, but it's not working. Surprisingly, I don't have the file src/katex.less in my distribution.

On my phone, I still have an ugly comma at the start of a line (see on Safari [https://snipboard.io/9QZwSX.jpg](url)), just after an interval in math mode.

fweth:
Do you maybe have a test case for this? I tried with your example in JSFiddle, seems to work for me: https://jsfiddle.net/qs3n21or/6/

Hamarkhis:
> Do you maybe have a test case for this? I tried with your example in JSFiddle, seems to work for me: https://jsfiddle.net/qs3n21or/6/

Thanks for your reply. It solved partially the issue I was talking about : we don't have the comma alone.

But in some cases, we have this, see [https://snipboard.io/Bqvxet.jpg](url), e.g bracket and comma together at the beginning of a new line. I don't know how to handle this.

fweth:
Ah, but that's not part of the initial topic then (line wrap after *inline* math). I suggest you use normal brackets for open intervals, like `[0, 1)`. BTW, I initially wanted to post my CSS "fix" on the issue itself, not here on the pull request. But I also don't want to spam, I just thought somebody might find it helpful. Would it OK to post also on the issue page?

edemaine:
@Hamarkhis Maybe try `\nolinebreak` before `[` as usual in LaTeX? (I haven't tested so not sure whether this would work.) Any further discussion on this should be elsewhere though (maybe start a new discussion).

@fweth Sure.

Hamarkhis:
> @Hamarkhis Maybe try `\nolinebreak` before `[` as usual in LaTeX? (I haven't tested so not sure whether this would work.) Any further discussion on this should be elsewhere though (maybe start a new discussion).
> 
> @fweth Sure.

Thanks for your answers and your propositions.

But the problem with the last one is that `\nolinebreak` does not seem to be supported by KaTeX.
[see screenshot there](https://snipboard.io/KIvBJ7.jpg) and [KaTeX supported commands](https://katex.org/docs/support_table#n).

This issue could be fixed nicely if KaTeX could support `\mbox` but that's not already the case for now. I hope it will be!

edemaine:
Ah, my apologies. It's been a while since I implemented line breaking. 😅 
Try wrapping in braces, e.g. `{c_1[}`.

Hamarkhis:
@edemaine Thanks for this idea, I did not think about it. Just tried now, but no effect on this issue. I think that `mbox` command must be implemented.

edemaine:
You could try `\hbox` which is implemented, and roughly the same as `\mbox`.

Hamarkhis:
@edemaine Hello, I have tested your idea, but the `hbox` command goals consists in using text and maths together.

We could think it could help. 

Unfortunately, the created characters inside `hbox` are not inside a real `mbox`, so I still have a break inside the interval.

Issue #2994, opened 2021-05-06
Where should macro definitions go?


edemaine:
Traditionally, macro definitions go in `src/macros.js`. But there are exceptions:
* #2984 introduced a macro in `src/functions/operatorname.js`.
* I did a similar thing in an environment in #2952. (In both cases, I feel like it's worth keeping logically clustered content together. )
* A smaller older example is that `environments/array.js`/`cd.js` redefines the `\cr` macro (though not using `defineMacro`).

In general I feel like `macros.js` is getting rather large and would probably be nicer to split it into logical components. The first chunk about defining macros for the sake of macro definitions makes sense, but the definition of various symbols, `\not`, `\dots`, etc. feel like they could belong elsewhere.

What if we moved the latter macro definitions to the `functions` directory?  Or maybe it would be better to rename `functions` to `commands` and make explicit that commands can be defined by functions, macros, or a mixture thereof?

_Originally posted by @edemaine in https://github.com/KaTeX/KaTeX/pull/2984#discussion_r627681165_

PR #3157, opened 2021-08-07
fix(macros): resolve circular dependency

ylemkimon:
Resolve circular dependency introduced in #2984.


edemaine:
Thanks for catching this. We should probably add the cycle-checking eslint rule.

This is related to #2994. If we want to enable macro definitions like this, we could move `macros` and `defineMacro` into a new `defineMacro.js`, like we already do in `defineFunction.js`.  What do you think?

Issue #3168, opened 2021-08-13
the line of \left\vert is disconnected


lesonli:
`\dfrac{\left\vert\dfrac{12\sqrt{3+211}}{a}\right|}{1}`

![image](https://user-images.githubusercontent.com/17868761/129369227-bd345ae0-1bf8-4cb4-a2fd-4a7ac0b6b9ed.png)

How can we solve this problem?

ronkok:
Well, that's annoying. I'll turn in a PR that removes a `.toFixed(3)` from the code. That rounding changes the width of an SVG by a minute amount. It might be enough for a browser to occasionally misplace the SVG.

edemaine:
@lesonli The latest version of KaTeX makes some changes to this behavior. Can you confirm whether you're still seeing an issue?

narisuzu:
> @lesonli The latest version of KaTeX makes some changes to this behavior. Can you confirm whether you're still seeing an issue?
![aaa](https://user-images.githubusercontent.com/16162768/139618236-bc5494a2-258c-459f-b9ea-c1fe7f8dcd4f.png)
Still has this issue in 0.15.1

edemaine:
Definitely looks like an improvement.  I don't think we have an easy way to do better.

For the specific case, of `|`, I could imagine doing the whole symbol with one SVG, but I imagine this is pretty different from how LaTeX renders large symbols.

mbourne:
This is still causing issues (recently, #3669).

> I could imagine doing the whole symbol with one SVG, but I imagine this is pretty different from how LaTeX renders large symbols.

Seems to me this issue would be resolved if one element was used. I had a play around and offer 2 possible solutions, which have the added benefit of reducing the number of DOM elements should one or other be implemented.

The first suggestion is to render it with a single SVG (as @edemaine mentioned), and the second is to draw a border. (I'm just tinkering with clones of the KaTeX output.)

https://bourne2learn.com/math/katex/large-absolute-values-signs-proposal.php

Screenshot of the extended SVG approach:

![image](https://user-images.githubusercontent.com/7541940/178136035-91529bd9-0dda-479a-8411-28dd68552e7c.png)


I still feel at times pure LaTeX approaches don't make sense in HTML. My guess is 99% of users just want it to look right, and don't care if if fails LaTeX purity tests, as long as it is valid HTML.

rben01:
I'm also seeing this. It definitely seems to depend on the specifics how browsers render fonts.

## Safari

[...]

PR #3170, opened 2021-08-14
fix: rounding in middle of a tall delimiter.

ronkok:
My best effort at a fix for #3168.


edemaine:
Funny, I thought this might have been caused by #2460, but now I see that was never merged. Incidentally, I see that [font metrics are rounded to 5 decimals](https://github.com/KaTeX/KaTeX/blob/3064ae218e3e4f73472e41afefa7b78b78384621/src/metrics/extract_tfms.py) so maybe that could be the target for #2460.

But I've never been certain about rounding, and the issue that this fixes renews that worry.  So I'm still not sure about #2460, but this PR seems good. I see that the other uses of `toFixed` are in `\includegraphics` and `array`'s MathML `rowspacing`, neither of which seems likely to be impacted much by rounding, so this is probably enough unrounding for now.

(Waiting for screenshots to get updated before actual review though.)

ronkok:
Looking at the screenshots, I think they look worse now than they did before. I'll try again with rounding to 5 decimal places.

ronkok:
The screenshots still show small misalignments here and there. But the interior geometry of the SVG exactly matches the font glyph, and I've done my best shot at the rounding calculation. I can think of nothing more that I can do.

You know, in the KaTeX fonts, the glyph internal geometry numbers are all 1000x one em. In most fonts these days, that ratio is some even multiple of two, such as 2048 or 1024. I suppose that would make this sort of floating point rounding error go away.

I'm going to update the screenshots. You can approve this PR or not. I think it makes little difference.

Issue #3161, opened 2021-08-08
Full Support of Braket Notation


Yue-Zhengyuan:
**Is your feature request related to a problem? Please describe.**
Not a problem.

**Describe the solution you'd like:**
Is there any plan to provide full support of the brakes notation? Currently \Bra, \bra, \Ket, \ket, \braket are supported, but \Braket is not; this one handles expressions like <a|b|c> and can adjust its size to the contents. 

**Link to or name of a (La)TeX package that provides the same feature:**
[The Braket package](https://www.ctan.org/pkg/braket)

ronkok:
This seems like a reasonable request. I'll write a PR. Some time may go by before it gets reviewed.

PR #3171, opened 2021-08-14
feat: Support \Braket, \set, and \Set

ronkok:
Address #3161.

Also support `\set` and `\Set`. KaTeX now supports all of the Braket package.


ronkok:
Withdrawn in favor of PR #3214.

Issue #3189, opened 2021-08-26
yarn start / webpack-cli fails


edemaine:
I'm getting the following error message when trying to `yarn start`.  I don't seem to be able to figure out where there's a `running` property anywhere.  Any ideas for fixing?  Updating webpack (#3182) didn't help.

```
[webpack-cli] Invalid configuration object. Object has been initialized using a
configuration object that does not match the API schema.
 - configuration has an unknown property 'running'. These properties are valid:
   object { bonjour?, client?, compress?, dev?, firewall?, headers?, historyApiFallback?, host?, hot?, http2?, https?, injectClient?, injectHot?, liveReload?, onAfterSetupMiddleware?, onBeforeSetupMiddleware?, onListening?, open?, openPage?, overlay?, port?, proxy?, public?, setupExitSignals?, static?, stdin?, transportMode?, useLocalIp? }
```

I've reproduced the error on Windows and Linux, and using Yarn or NPM/`npx`.  On the other hand, `yarn build` works fine, so it's specific to the dev server.

PR #3190, opened 2021-08-26
feat: optional vertical alignment [t|b|c] for array and aligned environments

### LINKED ISSUE #2172: Is array vertical positioning supported?
This is a question about the optional vertical positioning argument of an array. When I tried
`\begin{array}[t]{c|}1\\ \,\end{array} \begin{array}[t]{rrrr}1& 2& -5& 2\\ & 1& 3& -2\\ \hline 1& 3& -2& 0\end{array}`
I did not get an error message and it seems to work, but I did not find the [position] optional argument mentioned in the official documentation. Can I assume that it is officially supported?

### LINKED ISSUE #2760: Support the optional vertical positioning argument of gathered, aligned, and alignedat
> Before requesting a feature, please search for [existing issues](https://github.com/KaTeX/KaTeX/issues).

Issue #2172 requests support for the corresponding vertical positioning argument of the `array` environment (see also a related [comment](https://github.com/KaTeX/KaTeX/issues/269#issuecomment-379936635) in #269).

Currently (v0.12.0) KaTeX renders `\begin{aligned}[t]...\end{aligned}` as a vertically-centred box, e.g.:
Screenshot 2021-01-26
LaTeX (with the amsmath package) reads the `[t]` as an optional parameter, and renders the same markup as a top-aligned box:
Screenshot 2021-01-26
This difference can hinder use of KaTeX on existing LaTeX documents with amsmath markup.

Support the optional vertical positioning argument of `gathered`, `aligned`, and `alignedat`, as in amsmath.

For `gathered` and `aligned` **this is a breaking change!** When the amsmath markup is the default, instances of `\begin{gathered}[...\end{gathered}` and `\begin{aligned}[...\end{aligned}` in existing KaTeX documents will need editing, to avoid the `[` being read as the start of an optional argument. A rendering option to restore the v0.12.0 behaviour might be needed.

[amsmath](https://ctan.org/pkg/amsmath)

The [amsmath user guide](http://mirrors.ctan.org/macros/latex/required/amsmath/amsldoc.pdf) (§3.7) explains that, for the `gathered`, `aligned`, and `alignedat` environments:

> Like the `array` environment, these `-ed` variants also take an optional `[t]`, `[b]` or the default `[c]` argument to specify vertical positioning.

These optional arguments are also supported by MathJax.

N/A

**Additional context:**
Support for the optional arguments of `\raisebox` (issue #2140) and `\smash` would further ease use of KaTeX on existing LaTeX documents with amsmath markup.

edemaine:
`\begin{array}` and `\begin{aligned}` rendered any optional argument like `[t]`.

Matching LaTeX, `\begin{array}`, `\begin{darray}`, `\begin{aligned}`, `\begin{alignedat}`, `\begin{gathered}` (but not `\begin{align}`, `\begin{alignat}`, `\begin{equation}`, `\begin{gather}`, `\begin{split}` and starred variants, `\begin{matrix}` and variants, `\begin{smallmatrix}`, `\begin{cases}`) support custom vertical alignment via `[t]`, `[b]`, or `[c]` optional argument.

~~At least, that's the goal.  I haven't been able to test this yet, or write tests, because of #3189, but I theorize that it or a slight variation will work.~~

**Optional arguments in environments:** Along the way, I had to add support for optional arguments in environments, which was silently ignored before.  ~~This only gets used in `\begin{alignedat}[t]{2}` because the optional argument is in the "correct" position before the required argument. We can't use it (without some modification) in `\begin{array}{cc}[t]` because the optional argument comes later.~~  Turns out I was wrong, and the correct order is `\begin{array}[t]{cc}`, so I can use the optional argument parser in all cases.

**Demo:** 

image

image

**BREAKING CHANGE:** Optional arguments used to be rendered in aligned environments, but now they are correctly processed (as they would in LaTeX).  This is arguably a bug fix, but could break existing code like
```latex
\begin{aligned}
[hello] & [world] \\
\end{aligned}
```

Fixes #2172, fixes #2760


ronkok:
This is good stuff. It should get recognized in the documentation. Suggest modification of `supported.md` file:

In the _Other KaTeX Environments_ table, to the row labeled `gathered`, etc., insert a line .".. take an optional argument `[t|b|c]` to set vertical alignment."

Revise one example environment to read
```
\def\arraystretch{1.5}
\begin{array}[c]{r|c:l}
   ab & c & de \\ \hline
   f  & g & h \\ \hdashline
   i  & j & k
\end{array}
```

There is a sentence that begins: "The `{array}` environment supports `|` ... . It's an unusual sentence for this document, which doesn't do much explanation. It just enumerates every KaTeX function and shows a working example of each. So that sentence is out of sync with the rest of the document. It should be either deleted or augmented with the `{darray}` environment and the new optional argument. I would be okay with either choice.

pzinn:
it would be nice if this got approved!

pzinn:
actually, I applied the patch and it doesn't seem to work too well for me:
```
$X=\begin{array}[t]{cc} 1 & 2 \\ 3 & 4\end{array}$
```
produces this:
x

PR #3199, opened 2021-08-27
chore(lint): add import rule checks

edemaine:
`import` statements were unchecked, and cycles of imports were undetected

`import` statements are checked, and cycles of imports are detected

Indeed, this PR won't pass CI until #3198 or #3157 gets merged.

`eslint-plugin-import` required upgrading eslint to parse ECMAScript 2017 (from the default of v5!), which triggered some new instances of `comma-dangle` in function calls and arguments.  Some of them look really ugly (e.g. when calling a function with an inline function as sole argument), so I've made `comma-dangle` optional for functions.


ylemkimon (approved):
I've disabled no-cycle check for now.

PR #3204, opened 2021-08-27
ci(screenshotter): use macos runner for Safari

ylemkimon:

Safari screenshots were tested using Browserstack which:
- relied on an external service
- was slow due to remote connection
- has limited number of simultaneous tests
- was flaky and failed often
- external contribtuors don't have access to
- uses Actions secrets, which required `pull_request_target` event and elevated permission
- the semantic release step was run even if failed

This PR uses [GitHub Actions macOS runner](https://github.com/actions/virtual-environments/blob/main/images/macos/macos-10.15-Readme.md) for Safari screenshots. It's free for public repositories and even faster than `ubuntu` runner as it uses a real machine.

CI hierachy

It now uses `pull_request` event and included in the `ci.yml`. The release step will only run if both test and screenshotter pass.

As it runs on macOS, it also serves as a build test on Mac.

One caveat would be we cannot lock the Safari version, so I've setup a bot to notify if the Safari is updated in the image.

There are subpixel changes as the Safari version has changed and diffs are available in https://github.com/KaTeX/KaTeX/actions/runs/1178042047.


edemaine:
@ylemkimon This is a great idea!  Any idea why the CI Safari screenshotter is timing out, though? (Just tried running it twice.) Is there a time limit or something? Or a race condition when using Safari?

It also looks like the screenshots need updating again, but hard to do with the CI process crashing. I'm a bit concerned that the difference from Browserstack (in the current PR anyway) seems to include a big vertical shift. With the current setup, I am able to generate Safari screenshots locally (by adding my Browserstack credentials to the environment), but relying on CI would prevent this (unless I got a Mac Mini I guess).  It'd be nice if they generated the same screenshots, but maybe it's impossible.

ylemkimon:
The failing `Screenshotter / screenshotter (safari)` is the old screenshotter (Browserstack) and the `CI / screenshotter (safari)` is the new screenshotter (macOS runner).

ylemkimon:
> With the current setup, I am able to generate Safari screenshots locally (by adding my Browserstack credentials to the environment), but relying on CI would prevent this (unless I got a Mac Mini I guess).

I think the GitHub Actions would be more accessible, as it's free for personal/open-source usage.

edemaine:
Replaced by #4206

PR #3214, opened 2021-08-30
feat: Support \Braket, \set, and \Set

### LINKED ISSUE #3161: Full Support of Braket Notation

edemaine:
This PR is a possible replacement for #3171, building on @ronkok's code to address [my comment/suggestion](https://github.com/KaTeX/KaTeX/pull/3171#discussion_r696223463) to try use a temporary macro definition of `|`.

Support for `\bra`, `\ket`, `\braket`, `\Bra`, `\Ket` from `braket` package.

Complete `braket` package support by adding `\Braket`, `\Set`, `\set`.

Fixes #3161.

**Comparison to #3171**
This approach turned out not to be a lot shorter than #3171, but I do think it's better behaved, as it more closely matches the LaTeX implementation.  For example, consider the following code:
```latex
\def\foo{|}
\Set{x \foo y}
```
LaTeX and this PR will treat `\foo` just like `|`, whereas #3171 doesn't expand `\foo` in its processing, so it won't notice the `|`.

Another example:
```latex
\Set{x \| y}
```
LaTeX and this PR will not touch the `\|` because that's a completely different token than `|`.  #3171 does a regular expression substitution on the string, so it will modify `\|`, and ends up rendering something as image which is confusing.

I also conjecture that this code is more efficient. #3171 consumes the tokens, rewrites them via a substitution, and converts back them to a string, only for them to be parsed again.

But I don't want to assume that this is a better approach without review, so I didn't push directly to #3171. The commits include all of #3171, so @ronkok should be correctly indicated as a coauthor.

**Dependency**
~~To pass Flow type checking, this PR assumes some small functionality I added in #2952 for undefining macros.  So ideally that PR would be reviewed and merged first.  Alternatively, I could split that functionality off into a separate PR (or part of this one).~~ #2952 now merged, so this isn't an issue.

**Demo**
```latex
\Braket{\frac{a}{b} | y | z} \\
\Set{\frac{a}{b} | y | z} \\
\set{\frac{a}{b} | y | z} \\
```
As written in [`braket.sty`](https://ctan.math.washington.edu/tex-archive/macros/latex/contrib/braket/braket.sty), "use the vertical bar character "|" ...  In \Braket these vertical lines will expand to match the arguments, and in \Set [and \set] the first vertical will expand."
image


edemaine:
In coming up with my `\|` bad example, I noticed that both `||` and `\|` are handled specially within `\Braket` and `\Set`.  I've implemented those as well.

Also I found that the spacing around braces and | was a bit off compared to `braket.sty`.  They should now be aligned.

ronkok:
I agree that this is a better approach than PR#3171. I may not do a detailed review for a few days yet, but I will get there eventually.

ronkok:
@ylemkimon Thank you for reviewing PR #2952. To lighten the burden on you, I'll review this PR as soon as PR #2952 lands.

edemaine:
OK, #2952 is now merged into main which is merged into here, so this PR (after squashing) should only involve Braket/set/Set-specific changes &mdash; ready for review.

ronkok (approved):
A program that advances primarily by expanding macros, instead of by nesting functions, ends up with an entirely different set of re-directions in logic. I suppose I will never become proficient at creating it, but at least I can follow the work done by another.

This looks good to me.

edemaine:
Whoops, looks like this got approved and then never merged. Merging after I double-check everything.

PR #3215, opened 2021-08-30
chore(tests): fix use of jest-diff

edemaine:
[Jest 27 changed](https://github.com/facebook/jest/blob/master/CHANGELOG.md#2700) the signature for jest-diff to named exports only.

This bug only arises from failing tests, so it wasn't caught by CI.  But I noticed it when debugging #3214.

### DIFF
--- a/test/helpers.js
+++ b/test/helpers.js
@@ -3,7 +3,7 @@ import ParseError from "../src/ParseError";
 import parseTree from "../src/parseTree";
 import Settings from "../src/Settings";
 
-import diff from 'jest-diff';
+import {diff} from 'jest-diff';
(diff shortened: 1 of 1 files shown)


ylemkimon (approved):

ylemkimon:
I wonder how it didn't print import warnings. Maybe #3199 would check this?

edemaine:
Good point: #3199 ought to have helped, but I ended up turning off warnings about `import foo from 'foo'` because we do that a lot in the codebase.  And ESLint doesn't check types.  I would think that Flow would catch it, but I guess it didn't have types for that module (maybe we need to import some?).  Switching to TypeScript may help.  (I'm guessing Jest has TypeScript types.)

PR #3222, opened 2021-09-01
fix(fonts): remove hints from unknown symbols

### LINKED ISSUE #3219: Wrong position of \in in subscript
image

edemaine:
This bug is pretty serious, so I'd appreciate a quick review.

`\small x \in y` rendered like this:

image

`\small x \in y` renders like this:

image

**Approach:**
As Latin letters seem to be working fine, and [the only difference I've detected](https://github.com/KaTeX/KaTeX/issues/3219#issuecomment-910435731) is the enabling of font hints, I turned on a feature of `ttfautohint` (`-f none -S`) to not hint characters that it doesn't know about (which seems to include all mathematical symbols). This seems to fix the problem.

**Worry:**
[`ttfautohint` 1.8.4](https://sourceforge.net/projects/freetype/files/ttfautohint/1.8.4/) apparently apparently fixed a bug in exactly this feature. (This is actually how I discovered the feature.) It's not released on Ubuntu yet, so it's more difficult to use, but we could roll back more of #2866 to install it in the Dockerfile. But if it's working now, perhaps we could wait on that?

Fixes #3219


edemaine:
As I read more of `ttfautohint`'s documentation, I wonder whether we should use [the `--reference` option](https://www.freetype.org/ttfautohint/doc/ttfautohint.html#blue-zone-reference-font) so that our bold fonts align better with our regular fonts.  Here's an example of the current height discrepancy:

```latex
\mathrm{x}\mathbf{x}
```
image

Admittedly not *that* big a deal, but something to consider...

ylemkimon (approved):

ylemkimon:
Could you add a regression screenshot test on another PR?

edemaine:
@ylemkimon Thanks for the quick review!

Unfortunately, I just tested in the screenshotter, and the bug doesn't seem to arise there. From reading [the documentation of ttfautohint](https://www.freetype.org/ttfautohint/#introduction), I believe that this may have been a Windows-only issue: other operating systems (at least, Linux and MacOS from our tests) seem to ignore the hints.  We could add BrowserStack tests on Windows, but that might be somewhat overkill...?  Or maybe it's worth adding one test that runs on Windows?

PR #3223, opened 2021-09-01
docs: add GitHub Sponsors to FUNDING.yml

ylemkimon:

N/A

@KaTeX has been approved for GitHub Sponsors and is live on https://github.com/sponsors/KaTeX.


edemaine (approved):
Thanks for adding this!

edemaine:
Do you think we should add a sponsor link somewhere to katex.org?

Issue #3038, opened 2021-05-21
\dv and \pdv support


visika:
These two commands speed up significantly the insertion of derivatives and partial derivatives.

**Link to or name of a (La)TeX package that provides the same feature:**
They are included in the physics package in LaTeX, and documented at page 5 of https://ctan.mirror.garr.it/mirrors/ctan/macros/latex/contrib/physics/physics.pdf

TimeTravelPenguin:
As mentioned in #3184:
> I wouldn't recommend using any of the genfrac code. Instead, I would recommend using the macro system; see `src/macros.js`. You can return a sequence of tokens that calls `\frac` instead of redefining `\frac`.

I second this. The macro system is able to handel the various use-cases. As an example, in the [Derivatives](https://ctan.org/pkg/derivative?lang=en) LaTeX package, there are various cases for partial and ordinary derivatives, given the \pdv and \odv commands respectively.

As an example (see page 4 of the Derivatives documentation):
```latex
\odv{x}          % Basically \dfac{d}{dx}
\odv{f}{x}      % Basically \dfrac{df}{dx}
\odv{f}{x,y}   % Basically \dfrac{d^{2}f}{dxdy}
```

As an important note to title rename, I would like to suggest `\odv` rather than `\dv`, or both. Thanks :)

TimeTravelPenguin:
How would you recommend implementing this feature? My JS is limited, but I'm happy to help with any latex related problems, where able, on this feature. 

I have located [macros.js](https://github.com/KaTeX/KaTeX/blob/145af05a4ec8d7594f0db50cbc2332020439010d/src/macros.js), and wonder if you plan on implementing it purely via latex, or by some other means involving JS? 

I've done a little looking into it, but the package is pretty dense with Expl syntax, and am interested as to the approach.

Thanks.

visika:
I tried to make a minimal working version of \odv and \pdv. As I discovered working on this, arguments are referenced as `#1`, `#2`, ... So it recognizes only two parameters, numerator and denominator.

It is possible in theory to make it recognize higher order derivatives, adjusting the exponent of the differential at the numerator accordingly, like in the original derivative package, but I must admit my JS-fu is too weak for that, and my break from studying is over :)

It should have something to do with arglist's lenght and some if statements sprinkled here and there.

PR #3273, opened 2021-09-22
feat: add basic \odv and \pdv

visika:
Two new commands are available: \odv{y}{x} and \pdv{y}{x}, respectively for ordinary derivatives and partial derivatives.
These are macros based on \frac and only support first order derivatives.

See #3038 for discussion.

### DIFF
--- a/docs/support_table.md
+++ b/docs/support_table.md
@@ -787,6 +787,7 @@ use `\ce` instead|
 |\O|$\text{\O}$|`\text{\O}`|
 |\o|$\text{\o}$|`\text{\o}`|
 |\odot|$\odot$||
+|\odv|$\odv{y}{x}$|`\odv{y}{x}`|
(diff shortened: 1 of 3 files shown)


ylemkimon (changes requested):
Thank you for the PR! Could you let me know which package the `\odv` is from? Furthermore, I'm reluctant to merge partial support for forward compatibility.

visika:
@ylemkimon That's from the [derivative](https://ctan.org/pkg/derivative) package. The original package supports many additional options that I don't know how to implement. I appeal to anyone with more expertise if you really want to support those right now. In my opinion, this pull request is enough to improve the experience of anyone using derivatives in KaTeX, more can come later.
The forward compatibility is guaranteed, as the options only add features on top of the default cases.

visika:
I'm closing this pull request, as it is superseded by #3703 with more features.

PR #3285, opened 2021-09-27
Update node.md

isamu:
Fixed sample code that doesn't work

Just fix typo.


edemaine (approved):
Thanks for spotting this!

Issue #2194, opened 2020-02-01
Reduce redundant DOM elements


mbourne:
Google's Pagespeed Insights complains about the number of dome elements on my pages. There are usually more than 1,000 which appears to be the trigger for the message.

For example, given the simple base and exponent expression, 5^3, there are 17 spans created, 3 of which have class "mord" and 2 of them repeat the class "mord mtight".

    <span class="katex">
     <span class="katex-mathml">
      <math>
       <semantics>
        <mrow>
          <mstyle scriptlevel="0" displaystyle="true">
           <msup>
            <mn>5</mn>
            <mn>3</mn>
          </msup>
         </mstyle>
        </mrow>
        <annotation encoding="application/x-tex">
          \displaystyle{5}^{3}
        </annotation>
       </semantics>
      </math>
     </span>
     <span class="katex-html" aria-hidden="true">
      <span class="base">
       <span class="strut" style="height: 0.864108em; vertical-align: 0em;">
       </span>
       <span class="mord">
        <span class="mord">
         <span class="mord">
          5
         </span>
        </span>
        <span class="msupsub">
         <span class="vlist-t">
          <span class="vlist-r">
           <span class="vlist" style="height: 0.864108em;">
            <span class="" style="top: -3.113em; margin-right: 0.05em;">
             <span class="pstrut" style="height: 2.7em;">
             </span>
             <span class="sizing reset-size6 size3 mtight">
              <span class="mord mtight">
               <span class="mord mtight">
                3
               </span>
              </span>
             </span>
            </span>
           </span>
          </span>
         </span>
        </span>
       </span>
      </span>
     </span>
    </span>

The following appears to produce the same result, but has been stripped of (seemingly) redundant spans:

    <span class="katex">
     <span class="katex-mathml">
      <math>
       <semantics>
        <mrow>
          <mstyle scriptlevel="0" displaystyle="true">
           <msup>
            <mn>5</mn>
            <mn>3</mn>
          </msup>
         </mstyle>
        </mrow>
        <annotation encoding="application/x-tex">
          \displaystyle{5}^{3}
        </annotation>
       </semantics>
      </math>
     </span>
     <span class="katex-html" aria-hidden="true">
      <span class="base">
       5
       <span class="msupsub">
        <span class="vlist-t">
         <span class="vlist-r">
          <span class="vlist" style="height: 0.864108em;">
           <span class="pstrut" style="top: -3.113em; margin-right: 0.05em;">
[...]

PR #3364, opened 2021-10-27
WIP: Use CSS flexbox to set vertical alignment.

ronkok:
This PR does a complete rewrite of function `makeVList()` which KaTeX uses to set vertical alignment. The proposed method uses CSS flexbox as suggested in #1409. It achieves a significant reduction in the number of visible DOM elements that KaTeX creates, thus partially addressing #2194.

This is a work in progress. I am getting anomalous results ~~which I believe are due to a problem in the KaTeX fonts.  I really cannot go further with this PR until that font problem is resolved, because I cannot be sure if problems are due to my code or to the font. I'll open a separate issue for the font problem.~~

~~After the font problem is resolved, I'll come back and finish this PR.~~

I think we have a font solution, so I'm no longer blocked


ronkok:
I have failed to make this PR work. To make this idea work, the code must have access to (1) the TeX height and depth of every symbol, and (2) the line box ascent and descent of every symbol. There will be cases where it will be impossible to know the line box dimensions.

When I wrote this code, I did not have a clear view of the full picture. I was narrowly focused on the characters in the KaTeX fonts. For them, this idea would have worked. But KaTeX can render a very wide range of Unicode symbols, not just the characters in the KaTeX fonts. Those symbols come from fonts that are unknown to KaTeX. So their line box dimensions are also unknown, and the calculations in this PR cannot be done.

I can't see a way around that problem.

PR #3370, opened 2021-10-29
fix(fonts): fix the timestamp of fonts to the epoch

ylemkimon:

The timestamp of fonts was updated every time, causing all fonts to be modified and merge conflicts.

Fix the timestamp of fonts to the epoch.


ylemkimon:
@edemaine I cannot find concrete documentation on how they are used. However, a comment in `fonttools` source states that some fonts have bogus timestamps, so it should not matter: https://github.com/fonttools/fonttools/blob/8ab6af03c89726cf80ca3c4b755ae1bd0038b5da/Lib/fontTools/ttLib/tables/_h_e_a_d.py#L45-L46

edemaine:
By the way, fonttools now outputs the following warning (from the code near where you linked):
```
'created' timestamp seems very low; regarding as unix timestamp
'modified' timestamp seems very low; regarding as unix timestamp
```

I assume this is fine, just mentioning.

PR #3377, opened 2021-10-30
feat: conditionally export ECMAScript modules

ylemkimon:

Conditionally export ECMAScript module via [the `exports` field](https://nodejs.org/api/packages.html#conditional-exports).

**BREAKING CHANGE:** With module loaders that support conditional exports and ECMAScript modules, `import katex from 'katex';` will import the ECMAScript module.


docs/migration.md
```diff
@@ -3,28 +3,51 @@ id: migration
...
 spacing, `\\` should be used.
 
-## `\cfrac`, `\color`, `\textcolor`, `\colorbox`, `\fcolorbox`
+### `\cfrac`, `\color`, `\textcolor`, `\colorbox`, `\fcolorbox`
 They are no longer allowed as an argument to primitive commands, such as `\sqrt`
 (without the optional argument) and super/subscript. For example,
 `\sqrt\textcolor{red}{x}` no longer works and should be changed to
 `\sqrt{\textcolor{red}{x}}`.
+
+## v0.14.0
```
edemaine, line 31: Perhaps we should put most recent versions at the top of this file, like CHANGELOG?  I'm not sure though.

package.json
```diff
@@ -3,6 +3,33 @@
...
+    },
+    "./contrib/mathtex-script-type": {
+      "require": "./dist/contrib/mathtex-script-type.js",
+      "import": "./dist/contrib/mathtex-script-type.mjs"
+    },
+    "./contrib/render-a11y-string": {
+      "require": "./dist/contrib/render-a11y-string.js",
+      "import": "./dist/contrib/render-a11y-string.mjs"
+    },
+    "./*": "./*"
```
edemaine, line 31: Could you clarify that/why this is a breaking change?  I just tried installing this branch into a project (via `npm link`) and I could successfully do `require('katex/dist/contrib/auto-render.js')` as well as `require('katex/dist/contrib/auto-render')`.  Perhaps the `"./*"` rule is recursive, so it enables all old paths to still work?

If this doesn't actually work, perhaps it's worth adding rules like `"./dist/*"`, `"./dist/contrib/*"`, `"./src/*"` for backward compatibility?

ylemkimon: I just noticed `require('katex/contrib/auto-render')` wouldn't work before as well. The only breaking change is that `import katex from 'katex'` will now import the ECMAScript module (`dist/katex.mjs`), where it imported the CommonJS module (`dist/katex.js`) before, if supported. It'd not affect normal use, as module loaders/bundlers are smart enough to handle them, but the ECMAScript module uses newer syntax and it's technically a breaking change.

edemaine: Ah, right, that makes sense. I can't imagine what that would break, but it's certainly different behavior.

I wonder if it still makes sense to mention the new ways to import modules in `docs/migration.md` like you had before, but described as features instead of breaking changes?

ylemkimon: I think the current wording sounds like a feature:
```
With module loaders that support conditional exports and ECMAScript modules,
`import katex from 'katex';` will import the ECMAScript module.
```
WDYT?

edemaine: I agree; that text is good. I was referring to the following text, and wondering whether some of it should be included.  Maybe with a prefix like "As an alternative to importing `.js` or `.mjs` files directly, you can now..."

https://github.com/KaTeX/KaTeX/pull/3377/commits/25ff5b6f5cbfee86ca2afe00f72fef6ee917c471#diff-758ba107a728fccfdf0d098b2c06ded59127dd53ea287d630f72fb51040bc72dR34-R53

ylemkimon: I've added a table.

edemaine (approved):
Looks great!

ylemkimon:
Thank you for the review!

PR #3379, opened 2021-10-30
ci(fonts): remove fetch-depth: 0

ylemkimon:
There is no need to fetch entire Git history (`fetch-depth: 0`) as we're no longer using the last commit date in font building (#3370).


edemaine (approved):
Nice improvement. I tried running the `build fonts` label and it worked.

PR #3381, opened 2021-10-30
fix(fonts): Set and use typo(graphy) metrics in fonts

### LINKED ISSUE #3365: KaTeX fonts should set USE_TYPO_METRICS flag.
I write this issue to request that someone will modify the KaTeX fonts.

### Background

PR #3364 (WIP) changes KaTeX’s method of vertical alignment. The proposed method uses CSS flexbox and achieves a significant reduction in the number of visible DOM elements produced by KaTeX. In order to properly align elements per TeX specifications, the  code accounts for the difference in glyph metrics and [line box][1] metrics.

Line box ascent and descent should be set by the font’s values for _sTypoAscender_ and _sTypoDescender_. (Ref: https://drafts.csswg.org/css-inline-3/#ascent-descent.) But I am finding that this is untrue for the KaTeX fonts.  The fonts should have a flag set that directs the application to “USE_TYPO_METRICS”.  When that flag is set, all applications will use the same line box dimensions. Otherwise Windows applications may go their own way.  ([Ref][line spacing]).

The flexbox PR is failing. I believe that it is failing because it is using the sTypoAscender value for the assumed line box, but the actual line box, at least in Windows, may have a different ascent value. I find that a `\Huge` italic denominator overlaps the fraction bar and I attribute this error to the mismatch in ascent values.

### Request

Modify the KaTeX fonts. Set bit #7 in the [fsSelection][3] field of the OS/2 metrics table. In order to make it possible to set the flag properly, it may be necessary to use version 4 of the OS/2 table instead of version 2.

### Follow up

After the fonts have been modified, I’ll finish PR #3364.

[1]: https://iamvdo.me/en/blog/css-font-metrics-line-height-and-vertical-align

[line spacing]: https://www.high-logic.com/font-editor/fontcreator/tutorials/font-metrics-vertical-line-spacing

[3]: https://docs.microsoft.com/en-us/typography/opentype/spec/os2#fsselection

edemaine:
* `sTypoAscender` and `sTypoDescender` were 800 and -200 respectively, which I believe are just defaults.
* The `USE_TYPO_METRICS` flags wasn't set, which I think means that some operating systems would use these metrics while others would use the other ascender/descender metrics such as `usWinAscent` and `usWinDescent` (which were set differently).

* Sets `sTypoAscender` and `sTypoDescender` to the actual bounding box (same as we set other ascender/descender metrics, such as `usWinAscent` and `usWinDescent`)
* Sets the `USE_TYPO_METRICS` flag to encourage cross-platform compatibility.

@ronkok Can you confirm that you had in mind actually changing `sTypoAscender` and `sTypoDescender`, in addition to setting the `USE_TYPO_METRICS` flag?

The screenshots apparently change from this, though I haven't looked at the changes yet. We should probably merge #2460 before this PR.

Fixes #3365


ylemkimon:
I'm surprised only Firefox screenshots change.

edemaine:
That is curious. I wonder if we'd see more changes if we did a matrix with different operating systems. On the plus side, the changes all look minimal.

@ronkok Perhaps it makes sense for you to merge this into your PR #3364 to see whether it fixes your problem, before we decide to merge this one into `main`?

ronkok:
@edemaine Thanks for the quick action!

> Perhaps it makes sense for you to merge this into your PR #3364 to see whether it fixes your problem, before we decide to merge this one into main?

Agree 100%.

>Can you confirm that you had in mind actually changing sTypoAscender and sTypoDescender...?

No. My hope was that all the KaTeX fonts would continue to use sTypoAscender=800 and sTypoDescender=200. Otherwise, I have to track any change that could possibly switch the font before I know the bounding box dimensions.

If PR #3364 works the way I hope, then it won't matter what the bounding box dimension is, only that I know what it is. The proposed `makeVList()` function makes a correction on every vertical shift so that we get TeX alignment, not bounding box alignment.

Sorry if this is a bother. Can you change the fonts again?

edemaine:
@ronkok No problem. I just changed the metrics back back (though didn't double check, so let me know if I missed something).

ronkok:
Great. Now I can get to work. There is a substantial amount of work left to do it PR #3364, but with this font change, I at least have hope that it will not be a giant time sink.

I'll let you know how things turn out.

edemaine:
FYI, this change caused a significant reduction in our line widths in the screenshots. For example, here's the change for `BoldSymbol-chrome`:

image

But... I think the new line spacing matches LaTeX's spacing better.  Here's a texcmp of the previous example, which differs mostly in horizontal spacing (issue with tight?), but the bottom two lines seem spaced the exact same:

image

Hmm, maybe that was a fluke.  Here's texcmp for a different test, `VerticalSpacing` (green is LaTeX, red is this PR):

image

Without this PR, the alignment is closer:

image

Any ideas what's going on here?

ronkok:
In most of the KaTeX fonts, usWinAscent is larger than 800. In Main_Italic, usWinAscent is smaller than 800. But there are many different values and I find it hard to tease out which line box is controlling the spacing at any one point.

To me, it reinforces the point that this PR and PR #3364 should be released as a package, not separately.

edemaine:
> In most of the KaTeX fonts, usWinAscent is larger than 800. In Main_Italic, usWinAscent is smaller than 800.

Ah right, that would explain the weirdness.

For a more controlled experiment, I made a very simple example testing characters that are short (x), descend (y), and tall (M):
```latex
\begin{array}{l}
  \rm xxx \\
  \rm xxx \\
  \rm yyy \\
  \rm yyy \\
  \rm MMM \\
  \rm MMM \\
\end{align}
```

The result I found is that the texcmp is exactly the same, and very good, both before and after this PR:

image

So maybe the differences are all to do with vertical stacking, which hopefully #3364 can make some nice progress on. 🤞 🙂

ronkok:
I now see a problem. Not all vertical alignment happens via `makeVList()`. For instance, KaTeX has soft line breaks after relations and binary operators. After such a break, the spacing to the second line will be controlled by the line box, not by `makeVList()`.

If those lines contain a Size_4 radical, we'll have a problem. A Size_4 radical is 3em tall. The proposed line spacing is a 1em line box with 0.2 em leading. That would be a collision.

So @edemaine's first impulse was correct to set sTypoAscender to the value of usWinAscent.

And PR #3364 will have to include some way of keeping track of which font is in use, so that it can know the line box dimensions.  I'm not sure if I can do that. I'll give it shot.

edemaine:
I trust your judgement, @ronkok, but are you certain that TeX behaves in one way or the other?  I find our current setting of `usWinAscent` pretty hacky: the highest point of any glyph in the font.

ronkok:
For `makeVList` purposes, sTypeAscender can be any arbitrary value, as long as it is known. So there may be some wiggle room there. I'll try to think of some way to avoid the line wrap problem w/o adding a lot of complexity to PR 3364. 

Don't know yet what that would be. I'm back in research mode, nowhere close to code-writing mode.

ronkok:
Since the purpose of this PR is to support PR #3364, and that PR has failed, I am also closing this one.

@edemaine Thank you for working on the fonts and creating this PR.  You're a pro.

PR #3384, opened 2021-10-31
feat: implement \relax as no-op function

ylemkimon:

`\relax` was just ignored in the `MacroExpander` level, not being able to be consumed.

Thanks to `internal` parseNode type, we can implement it as a no-op function. It now stops expansions such as `\kern2\relax em`. No existing tests break, but it'll break edge cases like `\relax` being used in the middle of an expansion, though this wouldn't have worked in LaTeX as well. Therefore, I'm not sure this would count as a breaking change.

Related: #2123.

### DIFF
--- a/docs/migration.md
+++ b/docs/migration.md
@@ -3,6 +3,12 @@ id: migration
 title: Migration Guide
 ---
 
+## v0.15.0
+
+`\relax` is now implemented as a function. It'll stop expansions and parsing,
+so the behavior around `\relax` may change. For example, `\kern2\relax em` will
+no longer work.
+
 ## v0.14.0
 
 With module loaders that support conditional exports and ECMAScript modules,
--- a/src/MacroExpander.js
+++ b/src/MacroExpander.js
@@ -20,7 +20,6 @@ import type Settings from "./Settings";
 // List of commands that act like macros but aren't defined as a macro,
 // function, or symbol.  Used in `isDefined`.
 export const implicitCommands = {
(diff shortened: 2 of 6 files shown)


edemaine (approved):
Nice parser improvement! I didn't realize that `\relax` is a primitive that lasts so long in the parsing cycle.  I found [this discussion](https://comp.text.tex.narkive.com/rgBmmMPE/what-is-the-difference-between-relax-and-empty) helpful to confirm that this is the correct behavior. Convenient that you already had `internal` nodes set up for this!

One thing I noticed is that `\relax` can now be removed from the `implicitCommands` list at the top of `MacroExpander.js`.

I also noticed one macro definition missing a `\relax` at the end: `\colon`. All the others seem to have it.

Regarding breaking changes: I agree this is like a bug fix relative to LaTeX. But I also think it's helpful to signal that existing code may break. This case is minor though. I'm finding the lack of any signals annoying while we're in major version 0, where both fix and feat only increment the patch number with semantic-version, but I think it's helpful to increment the minor number in this scenario. Alternatively, if we can change the behavior of semantic-version to increment minor during all feat commits, that would match our old version numbering better.

ylemkimon:
@edemaine Thank you for the review! It's just it feels like we're doing a release a lot recently, but now I think that's the very purpose of the semantic release. I've added a migration guide.

edemaine:
Changes look good!

I could see semantic-release waiting up to 24 hours or something, to combine some PRs together... if that's an easy option? But it's also nice not to think about it. 😄

PR #3385, opened 2021-10-31
feat(macro): add basic conditional support

ylemkimon:
`\iftrue`, `\iffalse`, and `\ifmmode`.

More conditionals can be defined using `defineConditional`.

Fixes partially #1003.

### DIFF
--- a/src/MacroExpander.js
+++ b/src/MacroExpander.js
@@ -11,7 +11,7 @@ import {Token} from "./Token";
 import type {Mode} from "./types";
 import ParseError from "./ParseError";
 import Namespace from "./Namespace";
-import macros from "./macros";
+import macros, {conditionals} from "./macros";
 
 import type {MacroContextInterface, MacroDefinition, MacroExpansion, MacroArg}
     from "./defineMacro";
@@ -33,6 +33,7 @@ export default class MacroExpander implements MacroContextInterface {
     macros: Namespace<MacroDefinition>;
     stack: Token[];
     mode: Mode;
+    conditions: boolean[];
 
     constructor(input: string, settings: Settings, mode: Mode) {
         this.settings = settings;
@@ -42,6 +43,7 @@ export default class MacroExpander implements MacroContextInterface {
         this.macros = new Namespace(macros, settings.macros);
         this.mode = mode;
         this.stack = []; // contains tokens in REVERSE order
+        this.conditions = [];
     }
 
     /**
@@ -151,6 +153,30 @@ export default class MacroExpander implements MacroContextInterface {
                 this.stack.pop();
             } else {
                 break;
+            }
+        }
+    }
+
+    /**
+     * Skips ahead until finding the \else or \fi that ends the skipped text
+     * keeping track of \if...\fi nesting.
+     */
+    skipConditionalText() {
+        const condition = this.conditions.pop();
+        let depth = 1;
+        while (depth) {
+            const token = this.popToken();
+            if (conditionals.hasOwnProperty(token.text)) {
+                depth++;
+            } else if (token.text === "\\fi" ||
+                    (depth === 1 && !condition && token.text === "\\else")) {
+                depth--;
+                if (token.text === "\\else") {
+                    this.conditions.push(condition);
+                }
+            } else if (token.text === "EOF") {
+                throw new ParseError("Incomplete conditional! " +
+                    "End of input while skipping conditional text.", token);
             }
         }
     }
--- a/src/defineMacro.js
+++ b/src/defineMacro.js
@@ -17,6 +17,12 @@ export interface MacroContextInterface {
     macros: Namespace<MacroDefinition>;
 
     /**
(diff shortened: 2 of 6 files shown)


src/defineMacro.js
```diff
@@ -116,3 +128,24 @@ export const _macros: MacroMap = {};
...
+) {
+    _conditionals[name] = true;
+    // When an \if... is expanded, TeX reads ahead as far as necessary
+    // to determine whether the condition is true or false; and if false,
+    // it skips ahead until finding the \else, \or, or \fi that ends the
+    // skipped text.
+    defineMacro(name, function(context) {
+        const condition = evaluate(context);
+        context.conditions.push(condition);
+        if (!condition) {
```
edemaine, line 146: I believe this is the only place where `conditions`' boolean values are used.  Which suggests that you just need a depth counter, not an actual stack, because here it's known from line 144.

src/defineMacro.js
```diff
@@ -116,3 +128,24 @@ export const _macros: MacroMap = {};
...
+            context.skipConditionalText();
+        }
+        return '';
```
edemaine, line 149: Would `return []` be more efficient here, because it doesn't engage the tokenizer?  Not sure.  (Also applies to a lot of other macros.)

Alternatively, I wonder if it makes sense to check for `undefined` as a return value from a macro, and expand it to nothing in that case.  Then this line could be omitted, here and elsewhere.

src/macros.js
```diff
@@ -214,6 +216,29 @@ defineMacro("\\show", (context) => {
...
+    if (context.conditions.length === 0) {
+        throw new ParseError("Extra \\fi.");
+    }
+    context.conditions.pop();
+    return '';
+});
+
+defineConditional("\\iftrue", () => true);
+defineConditional("\\iffalse", () => false);
+defineConditional("\\ifmmode", (context) => context.mode === 'math');
```
edemaine, line 240: I seem to recall when implementing `\TextOrMath` that this wasn't possible before. I'm guessing it works now though.

Here's an attempt at `\newif`:

```js
defineMacro("\\newif", function(context) {
    let value = false;
    const name = context.popToken();  // don't expand following token
    // LaTeX silently strips off the first two characters, intending them to be 'if'. This is a more strict check:
    if (!name.text.startsWith('\\if')) {
        throw new ParseError("\newif followed by something other than \if...");
    }
    const cond = name.text.slice(2);
    defineConditional(name.text, () => value);
    defineMacro(`\\${cond}false`, () => {
        value = false;
        return '';
    });
    defineMacro(`\\${cond}true`, () => {
        value = true;
        return '';
    });
    return '';
});
```

test/katex-spec.js
```diff
@@ -3593,6 +3593,24 @@ describe("A macro expander", function() {
...
+            expect`\iffalse{\fi xyz\iffalse}\fi`.toParseLike`xyz`;
+            expect`\iftrue xyz`.toParse();
+            expect`\iffalse\else xyz`.toParse();
+            expect`\iffalse xyz \else \else abc \fi`.not.toParse();
+            expect`xyz \fi`.not.toParse();
+        });
+
+        it("\\ifmmode should work", function() {
+            expect`\def\foo{\ifmmode math\else text\fi}\foo`.toParseLike`math`;
+            expect`\def\foo{\ifmmode math\else text\fi}\textit{\foo}`.toParseLike`\textit{text}`;
```
edemaine, line 3611: Maybe add ``expect`\def\foo{\ifmmode math\else text\fi}\text{$\foo$}`.toParseLike`\text{$math$}`;``

Or these tests could be combined into something like `\foo\text{\foo$\foo$\foo}\foo` to make sure math mode is both entered and exited correctly.

src/MacroExpander.js
```diff
@@ -155,6 +157,30 @@ export default class MacroExpander implements MacroContextInterface {
...
+                depth++;
+            } else if (token.text === "\\fi" ||
+                    (depth === 1 && !condition && token.text === "\\else")) {
+                depth--;
+                if (token.text === "\\else") {
+                    this.conditions.push(condition);
+                }
+            } else if (token.text === "EOF") {
+                throw new ParseError("Incomplete conditional! " +
+                    "End of input while skipping conditional text.", token);
```
edemaine, line 179: LaTeX says `Incomplete \iffalse; all text was ignored after line <n>.` or `Incomplete \iftrue; all text was ignored after line <n>.` (in the case of `\iftrue ... \else ...`).  Perhaps `` `Incomplete \\if${condition}` `` would be appropriate and more informative?

edemaine (commented):
Wow, this is great!

As you'll see in the review, it took me a little while to understand how it all works, partly because of the file order presented in GitHub.  A few extra comments would help with that.

There are also a couple of design decisions that are at least worth thinking about, but which you can deem worse than the current approach if you like.

And I (maybe) implemented `\newif`, though I didn't test it.  And I reimplemented `\TextOrMath` in a better way thanks to `\ifmmode` (also untested).

PR #3422, opened 2021-11-19
fix(auto-render): concatenate content of successive text nodes

marcoesters:
Webkit-based browsers like Chrome, Edge, or Safari do not parse very large math expressions (limit appears to be around 48k characters) because they split large text nodes into smaller ones so that auto-render misses delimiters. This PR fixes this by concatenating all successive text nodes.

Tested on:

* Windows 10 with Chrome 96.0.4664.45, Edge 95.0.1020.53, and Firefox 93.0
* MacOS 12.0.1 with Chrome 96.0.4664.55 and Safari 15.1
* MacOS 10.13.6 with Safari 11.1.2
* Android 12 with Chrome 95.0.4638.74 Firefox 94.1.2

Webkit-based browsers (Chrome, Edge, Safari) split large text nodes into smaller ones. This can cause delimiters to be in separate child nodes, which the auto-render parser will miss.

All successive text modes will be concatenated into one node and the concatenated textContent will be used to render math.


contrib/auto-render/auto-render.js
```diff
@@ -55,7 +55,17 @@ const renderElem = function(elem, optionsCopy) {
...
             // Text node
-            const frag = renderMathInText(childNode.textContent, optionsCopy);
+            // Concatenate all sibling text nodes.
+            // Webkit browsers split very large text nodes into smaller ones,
+            // so the delimiters may be split across different nodes.
+            let textContentConcat = childNode.textContent;
+            let sibling = childNode.nextSibling;
+            while (sibling && (sibling.nodeType == 3)) {
+                textContentConcat += sibling.textContent;
+                sibling.remove();
```
edemaine, line 65: This looks like it will remove most text nodes that have no math in them; notice the `if (frag)` below.

edemaine (changes requested):
https://stackoverflow.com/questions/67738121/in-what-cases-do-browsers-create-multiple-adjacent-text-nodes seems to confirm this issue; do you have other documentation, or an example where this is failing?

Overall, this seems like a good idea, though potentially slow/inefficient, better than not working. But we need to not touch such text if it has no math expressions in it; see comment.

contrib/auto-render/auto-render.js
marcoesters: Thank you! You are right, if Chrome, etc., encountered a large text node, that would happen. I tried to address your comment.

As for efficiency, I'm wondering if testing if `childNode.textContent` contains any of the left delimiters first would improve the algorithm.

I attached an example that renders fine on Firefox but not on Chrome or Edge.

[katex_large_math_mwe.zip](https://github.com/KaTeX/KaTeX/files/7573075/katex_large_math_mwe.zip)

edemaine (commented):
Thanks -- this is looking good! Would you be able to add tests to `contrib/auto-render/test/auto-render-spec.js`? If not, I can try.

contrib/auto-render/auto-render.js
edemaine: > As for efficiency, I'm wondering if testing if childNode.textContent contains any of the left delimiters first would improve the algorithm.

Probably... but I like the simplicity of this strategy (for now), and it only impacts the performance when there are multiple sibling text nodes. If we wanted to be efficient, we'd only work with two siblings at once (or more if a math section actually spans more than two text nodes), but that's quite a bit more complicated and probably best left for a future PR if it ends up being an issue for someone.

> I attached an example that renders fine on Firefox but not on Chrome or Edge.

Thanks! I confirmed this.

I also confirmed that your changes fix the problem on Chrome (by running `yarn build` and using the resulting `auto-render.js` instead).

marcoesters:
> Thanks -- this is looking good! Would you be able to add tests to `contrib/auto-render/test/auto-render-spec.js`? If not, I can try.

I'm working on a test with two strings long enough to break on Chrome (including and not including math to render). I'm not familiar enough with jest to know if it's possible to reproduce the issue in a unit test since the problem is browser-specific.

My tests pass with the current version of katex as well, so I could provide a much smaller string if a browser-specific test is not possible.

edemaine:
I'm not sure what DOM the tests use, but it might be virtual. I had in mind tests where you manually make multiple adjacent text nodes, not necessarily large, and make sure the right substitutions happen (or don't happen).

marcoesters:
That makes a lot more sense than what I was trying to do.

The current tests should do the trick. The math parser fails in the current KaTeX version but passes with my mods.

contrib/auto-render/auto-render.js
```diff
@@ -55,10 +55,29 @@ const renderElem = function(elem, optionsCopy) {
...
         if (childNode.nodeType === 3) {
             // Text node
-            const frag = renderMathInText(childNode.textContent, optionsCopy);
+            // Concatenate all sibling text nodes.
+            // Webkit browsers split very large text nodes into smaller ones,
+            // so the delimiters may be split across different nodes.
+            let textContentConcat = childNode.textContent;
+            let sibling = childNode.nextSibling;
+            let nSiblings = 0;
+            while (sibling && (sibling.nodeType === 3)) {
```
ylemkimon, line 64:
```suggestion
            while (sibling && (sibling.nodeType === Node.TEXT_NODE)) {
```

ylemkimon (approved):

ylemkimon:
Thank you for the PR!

Issue #3500, opened 2022-01-11
Invalid markup error for \braket inside \frac element


pranabdas:
**Describe the bug:**
Invalid markup error for `\braket` inside `\frac` element in Firefox. Consider following equation:

$$
\frac{1}{\braket{\Phi|\Phi}}
$$

This throws `Invalid markup: Incorrect number of children for <mfrac/> tag.` in the browser console, nonetheless the equation is correctly displayed.

I am using KaTeX v0.13.24, Node 16.13.2, Mozilla Firefox 95.0.2 (64bit) on Windows 10.

**To Reproduce:**
1. Go to [katex.org](https://katex.org) using Firefox on Windows
2. Open browser console
3. Paste `\frac{1}{\braket{\Phi|\Phi}}` in the demo box

**Environment (please complete the following information):**
 - KaTeX Version: 0.13.24 and also on katex.org
 - Device: Desktop
 - OS: Windows 10
 - Browser: Firefox
 - Version: 95.0.2

**Additional context:**
The error does not appear on Safari on macOS, or Edge on Windows 10.

ronkok:
Thank you for raising this issue. I've found the bug and it is in the MathML part of KaTeX. That's why the HTML can still render properly.

I'll turn in a PR to fix it.

PR #3501, opened 2022-01-12
fix: \mathinner when invoked as a denominator.

ronkok:
Currently, the MathML version of `\mathinner` returns a fragment, not a group. That results in invalid MathML when `\mathinner` becomes the denominator of a fraction, because in MathML a denominator can be only a single element, not several of them.

This PR revises `\mathinner` so that it returns a single `<mpadded>` element instead of a fragment. It thereby fixes issue #3500, which invokes `\mathinner` via the macro for `\braket`.

### DIFF
--- a/src/functions/mclass.js
+++ b/src/functions/mclass.js
@@ -22,7 +22,7 @@ function mathmlBuilder(group: ParseNode<"mclass">, options) {
     const inner = mml.buildExpression(group.body, options);
 
     if (group.mclass === "minner") {
-        return mathMLTree.newDocumentFragment(inner);
+        node = new mathMLTree.MathNode("mpadded", inner);
     } else if (group.mclass === "mord") {
         if (group.isCharacterBox) {
             node = inner[0];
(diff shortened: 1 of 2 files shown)


ronkok:
Screenshotter tests have failed to run. In other news, the BrowserStack account for `ronkok/KaTeX` has expired due to inactivity. I suspect those two facts are related.

Since this is a parsing issue, I did not create any new screenshotter tests for this PR. I think it would be okay to merge it as is.

ronkok:
I wrote a better test -- one that captures the problem with the previous MathML output. This should be good to go.

edemaine (approved):
Thanks @ronkok for the quick fix! This looks like a nice improvement, presumably also slightly improving MathML layout as it captures the spacing that should surround `\mathinner`. Admittedly I don't quite understand when that space is supposed to be present/absent, but the simple strategy in this PR is probably good enough for now (certainly better than "no spacing ever").

I also confirmed that `\frac{1}{\braket{\Phi|\Phi}}` no longer gives an error on Firefox.

Also, the tests just needed a rerun. I think GitHub Actions was temporarily down.

Issue #3503, opened 2022-01-13
\pmb with math operator renders awkward


arctouch-danielarenhart:
**Describe the bug:**
`\pmb` with math operator inside renders awkward (looks like it's rendering double)
It renders ok with text, Letters, or Unicode

**(La)TeX code:**

```latex
\pmb{\sin}
```

**Screenshots:**
![Screen Shot 2022-01-13 at 9 19 46 AM](https://user-images.githubusercontent.com/78916643/149329120-ccef611b-1338-4803-ab28-382be53f2b93.png)

PR #3505, opened 2022-01-13
Fix: Implement \pmb via CSS text-shadow

ronkok:
Currently, `\pmb` is implemented by rendering multiple copies of the argument with small offsets. It's a kludge, as indicated by (1) the complex exceptions in the original AMS macro, and (2) actual results.

This PR implements `\pmb` via CSS `text-shadow`. The rendering is better, particularly in large sizes and in math operators. The improvement resolves issue #3503.

### DIFF
--- a/contrib/render-a11y-string/render-a11y-string.js
+++ b/contrib/render-a11y-string/render-a11y-string.js
@@ -392,6 +392,11 @@ const handleObject = (
             break;
         }
 
+        case "pmb": {
+            a11yStrings.push("bold");
+            break;
+        }
+
         case "phantom": {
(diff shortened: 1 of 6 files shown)


ronkok:
Don't review this PR yet. There is an issue in MathML.

ronkok:
Update: The issue that I have just found is unrelated to this PR. I was thinking that I might add the correction here, but the other issue should really be in its own PR.

So this PR should be good to go.

ylemkimon (approved):
Thank you for the PR, as always. Sorry for the late review.

PR #3515, opened 2022-01-23
docs: update support_table.md to reflect the fact that \leftarrowtail is supported

BettyJJ:
…ct it.

\leftarrowtail is supported. It can be used in the live demo on https://katex.org/. It is also listed in supported.md/https://katex.org/docs/supported.html. However, in support_table.md, it is listed as "not supported". 

update support_table.md


edemaine (approved):
Thanks for spotting this, and sorry for the delay.

PR #3516, opened 2022-01-23
feat: add the \iiiint symbol ⨌

BettyJJ:

### DIFF
--- a/src/symbols.js
+++ b/src/symbols.js
@@ -657,6 +657,7 @@ defineSymbol(math, main, op, "\u222b", "\\int");
 defineSymbol(math, main, op, "\u222b", "\\intop");
 defineSymbol(math, main, op, "\u222c", "\\iint");
 defineSymbol(math, main, op, "\u222d", "\\iiint");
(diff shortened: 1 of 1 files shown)


ronkok:
It takes more than this to get a good `\iiiint` function. The `symbols.js` module is written to call glyphs from the KaTeX fonts. But the KaTeX fonts do not contain a glyph for Unicode U+2a0c. So the result of this PR will be a console error message and a glyph from a system font that may be taller than what KaTeX expects. 

The next thing an `\iiiint` function needs is proper treatment of `\limits`, which would require modification to KaTeX file `op.js`.

If you just need something quick for your own work, you could try a macro: `\def\iiiint{\iint\mathchoice{\mkern-10mu}{\mkern-10mu}{\mkern-7mu}{\mkern-6mu}\iint}`. That will not deal properly with `\limits`, but it will look okay otherwise.

If you want this PR to be approved, you probably need first to persuade @edemaine or @ylemkimon to revise the KaTeX fonts and add a glyph at code point U+2a0c.

BettyJJ:
@ronkok OK, I'm new to this. Thank you.

PR #3583, opened 2022-03-13
feat(array): add support for repetition in format description

aaruni96:
It is now possible to describe an array environment like
\begin{array}{* {35}{l}}
This completes task number 3 on
[#269](https://github.com/KaTeX/KaTeX/issues/269)

When encountering an array environment described with the repetition format ( `\begin{array}{* {7}{cl}` ), KaTeX would throw an error and not render the equation.

Equation now renders as expected.

Contribtues to #269

### DIFF
--- a/src/environments/array.js
+++ b/src/environments/array.js
@@ -788,12 +788,23 @@ defineEnvironment({
             symNode ? [args[0]] : assertNodeType(args[0], "ordgroup").body;
         let myColAlign = [];
         if (colalign[0].text === "*") {
-            const numStart = colalign[1].loc.start + 1;
-            const numEnd = colalign[1].loc.end - 1;
-            const iRepeat = colalign[0].loc.lexer.input.slice(numStart, numEnd);
-            for (let i = 0; i < iRepeat; i++) {
-                for (let j = 0; j < colalign[2].body.length ; j++) {
-                    myColAlign.push(colalign[2].body[j]);
+            const repeatCol = assertNodeType(colalign[1], "ordgroup");
+            let repeatString = "";
+            for (let i = 0; i < repeatCol.body.length ; i++) {
+                const iRepeat = assertNodeType(repeatCol.body[i], "textord");
+                repeatString += iRepeat.text ;
+            }
+            const nRepeat = parseInt(repeatString);
+            if (isNaN(nRepeat)) {
+                throw new ParseError("Repeat specification not a number: "
+                + repeatString, repeatCol);
+            }
+            const ca = assertNodeType(colalign[2], "ordgroup");
+            const caLength = ca.body.length;
+            for (let i = 0; i < nRepeat; i++) {
+                for (let j = 0; j < caLength ; j++) {
+                    const jca = assertNodeType(ca.body[j], "mathord");
+                    myColAlign.push(jca);
                 }
             }
         } else {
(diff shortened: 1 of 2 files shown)


edemaine:
> Can someone explain why the test for flow is breaking ?

Because `loc?: ?SourceLocation` says that `loc` is optional. I don't remember exactly why, but I imagine this can happen with macros, e.g. `*\foo`.

Instead of reading the text directly from the lexer, I think it would be better to iterate over the contents of the node and extract the `text` from them, verifying that each is `/[0-9]/`. Check out how this is done in `functions/char.js`, for example.

aaruni96:
I have made the changes and and my commit is awaiting approval before it will be tested.

aaruni96:
Hello, I have added a test for this feature so that codecov is also happy.

aaruni96:
Hello, I have fixed all the things I could figure out how to fix.

The deploy checks for netlify are currently failing, but as far as I can tell, its not because of and of my changes ?

aaruni96:
Hello,
The PR now appears to pass all checks. Is it ready to merge?

aaruni96:
*bump*

aaruni96:
@edemaine does this PR require any more work from me to get merged?

test/katex-spec.js
```diff
@@ -1260,6 +1260,10 @@ describe("A begin/end parser", function() {
         expect`\begin{array}{cc}a&b\\c&d\end{array}`.toParse();
     });
 
+    it("should parse an environment with argument in repetition format", function() {
+        expect`\begin{array}{*{35}{l}} x+y\le 12, \ 2x-y\ge 0, \ x-2y\le 0\end{array}`.toParse();
```
edemaine, line 1264: I think a `toParseLike` test would be more helpful here. E.g. `\begin{array}{*{10}{l}}...` should parse like `\begin{array}{llllllllll}...`.

src/environments/array.js
```diff
@@ -786,7 +786,31 @@ defineEnvironment({
         const symNode = checkSymbolNodeType(args[0]);
         const colalign: AnyParseNode[] =
             symNode ? [args[0]] : assertNodeType(args[0], "ordgroup").body;
-        const cols = colalign.map(function(nde) {
+        let myColAlign = [];
+        if (colalign[0].text === "*") {
```
edemaine, line 790: This seems to only support `*` at the beginning of `colalign`. But `*` can be used anywhere, e.g. `\begin{array}{ll*{5}{lcr}rr*5{rrcl}}`. I think ideally you'd do a `map` to expand all instances of `*`.

src/environments/array.js
```diff
@@ -786,7 +786,31 @@ defineEnvironment({
...
+            const repeatCol = assertNodeType(colalign[1], "ordgroup");
```
edemaine, line 791: This seems to only support `*{123}` with braces. But for single-character numbers, you can omit the braces. For example, `*5` should work.

src/environments/array.js
```diff
@@ -786,7 +786,31 @@ defineEnvironment({
...
+            let repeatString = "";
+            for (let i = 0; i < repeatCol.body.length ; i++) {
+                const iRepeat = assertNodeType(repeatCol.body[i], "textord");
+                repeatString += iRepeat.text ;
+            }
+            const nRepeat = parseInt(repeatString);
```
edemaine, line 797: Should use `Number` (or `parseInt(..., 10)`) to force decimal parsing.

I see that there's similar code (which uses `Number`) when computing `numMaths`. Could this be factored out into a helper function? (with the comment above too) Converting text to numbers like this is quite useful... In fact, I see a totally different implementation of it (which I maybe wrote?) in `macros.js` under `defineMacro("\\char"`. 😮

edemaine (changes requested):
Sorry this took so long to review. The main comment here is we should support `*` anywhere in the `colalign` string, not just at the beginning.

PR #3586, opened 2022-03-20
fix(copy-tex): Use JS (instead of CSS) to select full equation, solving display glitches

fonsp:
When selecting a subrange of a KaTeX equation, we [currently use `user-select: all` to ensure that the entire equation is selected](https://github.com/KaTeX/KaTeX/tree/main/contrib/copy-tex#copy-tex-extension), which is needed to get the tex code correctly. 

I found this behaviour to be unstable on Chrome (not just on Safari, like mentioned in the README), and I would often get a partial selection, leading to strange HTML content:

https://user-images.githubusercontent.com/6933510/159169941-cf4aaf8e-bc40-4182-af40-0aa0588dc5ab.mov

This PR uses JS to guarantee that when a subrange of an equation is selected, the parent equation is used for the copying logic. We use `.closest('.katex')` to find this parent.

When the selection is not fully contained inside a katex equation element, the old behaviour is used, which was working great!

This means that the `user-select: all` class (and in fact, the entire CSS file), can be be removed 🎉  This fixes the display bugs:

https://user-images.githubusercontent.com/6933510/159170237-08f5ae44-b76b-48e1-bd5c-c4d039a0f150.mov

Removing the CSS file should not be breaking, as importing it from a future version will just give a 404, which is fine. (Right?) If not, we can put the file back with empty contents.

Let me know what you think!


contrib/copy-tex/copy-tex.js
```diff
@@ -1,24 +1,41 @@
...
+
+    if (closest_katex) {
+        // Then we are fully inside an equation. We expand the selection to be the entire equation.
+
+        fragment = /** @type Element */ (closest_katex.cloneNode(true));
+
+        if (!fragment.querySelector('.katex-mathml')) {
+            return; // default action OK if no .katex-mathml elements
+        }
+        html_contents = fragment.innerHTML;
```
edemaine, line 26: Should this be `html_contents = fragment.outerHTML`? Otherwise you're missing the `<span class="katex">` wrapper, right?

edemaine (commented):
This looks nice (but I still need to test it). I have one comment about the HTML output though.

contrib/copy-tex/copy-tex.js
fonsp: True, thanks!

fonsp:
```suggestion
        html_contents = fragment.outerHTML;
```

edemaine (approved):
Sorry for the delay. I cleaned up your code (in particular, lint was failing, and we weren't using Flow, but are now).

In experimenting with use of this rewrite, I found the undesirable behavior that, if your selection starts within one equation and ends within a different equation, we get ugly text again. So I rewrote the code to separately extend both the start and end of the range, instead of just looking at the common ancestor. This should restore the intended behavior.

Could you take a look and make sure you're good with my changes, before I merge and release?

fonsp:
Looks great, thanks for the help!

fonsp:
Ooh and an instant release after the merge, that's awesome! 🌟

Mister-Hope:
Hi @fonsp (contributor) and @edemaine (approver), curiously to find out that this PR is made a long time ago.

Currently the file is written in flow (I guess?), but the file is obviously not a valid js file for most browsers and it's publishing to npm as-is.

Can anyone try to fix it by compiling it to a standard JS file when publishing?

See main branch content and you will get what I am talking:

- https://github.com/KaTeX/KaTeX/blob/main/contrib/copy-tex/copy-tex.js
- https://github.com/KaTeX/KaTeX/blob/main/contrib/copy-tex/katex2tex.js

related issue https://github.com/KaTeX/KaTeX/issues/3948

PR #3525, opened 2022-01-27
docs: fix broken link

HallowedDust5:
The previous link is deprecated, I'm pretty sure the new link directs to what is intended, but I was not here before that link broke.

Fixes #3524

PR #3588, opened 2022-03-22
feat: Add `\strictif` and `\strictfi` symbols

SnO2WMaN:

There's no symbols for `/strictif` (⥽) and `/strictfi` (⥼) for strict implication

Added them.

ps. some of the links in CONTRIBUTING.md were dead (https://github.com/KaTeX/KaTeX/pull/3525), and I didn't find correct way to contribute (adding symbols), so I referenced #2283 . Please let me know if there are any problems.

### DIFF
--- a/docs/support_table.md
+++ b/docs/support_table.md
@@ -1005,6 +1005,8 @@ use `\ce` instead|
 |\Stigma|<span style="color:firebrick;">Not supported</span>||
 |\stigma|<span style="color:firebrick;">Not supported</span>||
 |\strut|<span style="color:firebrick;">Not supported</span>||
+|\strictif|$\strictif$||
+|\strictfi|$\strictfi$||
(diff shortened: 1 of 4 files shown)


edemaine:
We can't add symbols to KaTeX (correctly) without adding them to the fonts and metrics. In the console, you'll see a warning that the character metrics aren't defined, and the symbol you're seeing is rendered from an arbitrary system font instead of a KaTeX font.

#3525 isn't relevant because MathJax doesn't have this symbol. What LaTeX package does? Presumably it's not in the standard Computer Modern fonts, which means you'd need to modify the font extraction code in `src/fonts` to look at other LaTeX fonts.

SnO2WMaN:
`\strictif` and `\strictfi` are included in the package `txfonts` or `pxfonts`. Sorry, but I'm not very familiar with TeX itself or KaTeX, and I'm not quite sure what I'm supposed to do, should I open an Issue?

PR #3590, opened 2022-03-25
refactor: replace deprecated String.prototype.substr()

CommanderRoot:
[String.prototype.substr()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr) is deprecated so we replace it with [String.prototype.slice()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/slice) which works similarily but isn't deprecated.
.substr() probably isn't going away anytime soon but the change is trivial so it doesn't hurt to do it.

Same as new

Same as before, just without a deprecated function


ylemkimon (approved):
Thank you for the PR!

PR #3633, opened 2022-05-16
feat: Support Unicode (sub|super)script characters

ronkok:
Math-mode Unicode (sub|super)script characters will now render as if you had written regular characters in a subscript or superscript. For instance, `A²⁺³` will render the same as `A^{2+3}`.

Resolves #1218.


ronkok:
I'm having some issues.

1. `yarn start` will cause my browser tab to hang. It's waiting for some `msn.assets`, which seems odd.
2.  Screenshotter is failing. This is suspicious because I have not altered any screenshotter tests and the test definitions do not contain any (sub|super)script characters. The screenshotter tests should be unchanged.
3. Netlfy preview is also hanging.

This PR changes only the Parser and the unit tests indicate that the Parser is working just fine. So I don't know what is causing these errors.

Any suggestions would be welcome.

ronkok:
Agree completely regarding the performance penalty to the Lexer. That concerned me, too. And now that I think about it, the `fetch` and repeat approach would fit nicely into the proposed code block that begins with `} else if (unicodeSubsAndSups[lex.text[0]]) {`.

That approach would not have fit so nicely in an earlier draft of my code. Hence the revision to the Lexer. 

I'll have a go at the `fetch` and repeat approach. It might be a few days until I get to it.

ronkok:
This PR now acquires tokens by repeated `fetch`. The Lexer is reverted to its previous state.

I now get good behavior from `yarn start`, but I still get no response from `yarn run test:jest`. That Jest problem occurs both in this branch and in my local main branch. A `yarn install` did not help. Odd.

ronkok:
The Jest test has apparently run successfully in GitHub and the code is executing well in the Netlify preview. I think this PR is ready to go.

universemaster:
This question should not hold up any pending reviews but `unicodeSupOrSub.js` does not contain the subscripts `ₕ , ₖ , ₗ , ₘ , ₙ , ₚ , ₛ , ₜ`. 
Is there a reason for that?  They are in the same `U+209x` range as 

`ₐ,ₑ ,ₒ, ₓ`.

`ₔ` and latin-extended-C block ⱼ, and Phonetic Extensions ᵢ ᵣ ᵤ ᵥ and Greek `ᵦ ᵧ ᵨ ᵩ ᵪ` are  also missing, but that's more understandable.

I'm getting these from the [wikipedia](https://en.wikipedia.org/wiki/Unicode_subscripts_and_superscripts#Superscripts_and_subscripts_block) page.  There are some other ones on that page, too.

universemaster:
unicode-math has a much larger list with the unicode values [here]( https://github.com/wspr/unicode-math/blob/d28e48e46392516bae43bd9cb15660f7037e47ac/um-code-sscript.dtx#L109).

ronkok:
Characters such as `ₕ  ₖ ᵦ ᵧ` will not render properly in NotePad++ with the Consolas font. Yes, there are many editors in which these characters will appear fine. Even in NotePad++, they will render properly if the JuliaMono font is installed and used.

But I think that the  NotePad++ / Consolas combination is a frequent entry point for beginning coders and I am a little reluctant to encourage the creation of documents that some authors will not be able to read.

edemaine:
I feel like matching unicode-math's list makes sense. Presumably those who want to use this will use appropriate fonts.

ronkok:
> Presumably those who want to use this will use appropriate fonts.

I have in mind the beginner who looks at existing code as a means to learn something.

I do not hold this opinion strongly. I'm going to wait a day before I take any action.

universemaster:
In my experience, people generally understand what's going on when a character doesn't display correctly and know to switch to a font that supports it.

edemaine:
I understand your concern, but I'd lean heavily toward LaTeX compatibility here. Also, the characters render in the web browser (at least my Chrome on Windows), so it's natural to try to copy/paste them. Even if they don't render in the editor, I would expect them to render on the website.

ronkok:
Okay, I've matched the list in unicode-math. This should be ready to go.

src/unicodeSupOrSub.js
```diff
@@ -0,0 +1,111 @@
...
+    '\u2092': 'o',
+    '\u209A': 'p',
+    '\u1D63': 'r',
+    '\u209B': 's',
+    '\u209C': 't',
+    '\u1D64': 'u',
+    '\u1D65': 'v',
+    '\u2093': 'x',
+    '\u1D66': 'β',
+    '\u1D67': 'β',
```
edemaine, line 42:
```suggestion
    '\u1D67': 'γ',
```

src/Parser.js
```diff
@@ -399,6 +400,29 @@ export default class Parser {
                 }
                 // Put everything into an ordgroup as the superscript
                 superscript = {type: "ordgroup", mode: this.mode, body: primes};
+            } else if (uSubsAndSups[lex.text]) {
+                // A Unicode subscript or superscript character.
+                // We treat these similarly to the unicode-math package.
+                // So we render a string of Unicode (sub|super)scripts the
+                // same as a (sub|super)script of regular characters.
+                let str = uSubsAndSups[lex.text];
+                const scriptType = unicodeSubRegEx.test(lex.text) ? SUB : SUPER;
```
edemaine, line 409: The `SUB`/`SUPER` codes seem a bit overly complicated. How about just an `isSub` boolean?

edemaine (changes requested):
This is looking great now! I found what I think is one typo, and a possible code simplification.

src/unicodeSupOrSub.js
ronkok: Done.

src/Parser.js
ronkok: Done.

edemaine (approved):

PR #3635, opened 2022-05-19
stlye: improve links (http -> https)

deining:

Several links were called with `http` protocol.

Links are now called with `https` protocol.


edemaine (approved):
Thanks!

Issue #2707, opened 2020-12-27
Update tests to use toThrowErrorMatchingInlineSnapshot in unit tests


kevinbarabash:
We currently use `.not.toParse()` which doesn't actually provide any information about what the failure mode is.  Switching to `toThrowErrorMatchingInlineSnapshot()` shouldn't be too much work.  It's probably something that could be done with a find-and-replace regex followed by running `jest` with the `-u` to update all of the snapshots.

docs: https://jestjs.io/docs/en/expect#tothrowerrormatchinginlinesnapshotinlinesnapshot

3r3n-n:
Hi ! 
Can I try to work on this ?

kevinbarabash:
@3r3n-n go for it.  Maybe try converting a couple uses of `.not.toParse()` and put up a PR to see what other people think about the change before converting the rest.

jakeoeding:
Hello - I'm interested in working on this. It looks like there is a similar issue (#2068) from 2019. The only difference is that one also mentions changing instances of `.not.toBuild()`. However, if I'm not mistaken there is only one instance of that at the moment.

I don't see any open pull requests referencing either of these issues so I'll put together a WIP that changes a few instances like you suggested to gather feedback.

jakeoeding:
Ok I've prepared a WIP showing what this could look like in #3636.

I found this a little more challenging to get right than I originally anticipated. The tests I changed were initially just failing because they threw the `ParseError`. The only way I found to get around this was to wrap the parsing with `expect(() => ... )`.

Please let me know what you think. I see that you already use non-inline snapshots in other places in this file so I wasn't sure if you would prefer to continue following that convention or not. I opted to go with inline for now because that's what you mentioned when initially opening this issue.

jakeoeding:
@kevinbarabash Do you have any feedback on the approach I've taken in #3636?

Green00101:
Hi, I’m interested in this issue and would like to work on it. Please let me know if it’s available.

PR #3636, opened 2022-05-22
test: include more info on parse failure tests

jakeoeding:
WIP showing potential option for #2707

Previously, tests that check for parsing failure do not include any information about what the failure mode is.

This attempts to add context to these tests by utilizing Jest snapshots.

### DIFF
--- a/test/katex-spec.js
+++ b/test/katex-spec.js
@@ -445,7 +445,9 @@ describe("A function parser", function() {
     });
 
     it("should not parse 1 argument functions with no arguments", function() {
-        expect`\blue`.not.toParse();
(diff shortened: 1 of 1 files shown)


jakeoeding:
Can anyone provide some comments/feedback on this approach for #2707 ?

grigoriy-reshetniak:
@jakeoeding sorry for delay on this.

It seems that [`toFailWithParseError`](https://github.com/KaTeX/KaTeX/blob/main/test/setup.ts#L59) should be used instead plain `toThrowErrorMatchingInlineSnapshot`.

PR #3638, opened 2022-05-23
feat(css): configurable margin for display math

nschloe:
See https://github.com/KaTeX/KaTeX/discussions/3637.


edemaine (approved):
Sorry this took so long to review. Hopefully it's still useful to you.

PR #3646, opened 2022-06-04
docs: Fix \set documentation

edemaine:
Work around `|` in tables

See https://github.com/facebook/docusaurus/issues/2004

https://katex.org/docs/supported.html

image

image

Similarly in support table.

image

image

Working around [this bug](https://github.com/facebook/docusaurus/issues/2004) was tricky. I had to `\gdef\VERT{|}` outside any table to make `|` effectively accessible within a Markdown table. This required making macros re-usable within docs, which may be useful in the future too.


edemaine:
@ronkok Would you mind reviewing this small doc fix PR?

ronkok (approved):
Looks good. Thank you for making this revision.

I wish that Markdown would accept grid tables similar to reStructuredText. That would make pipes easy to handle, in addition to other goodness.

PR #3649, opened 2022-06-06
docs: Cleanup changelog

edemaine:
Last release changelog was a little off because of a breaking change message in the middle of the log instead of the end.


kevinbarabash (approved):
Looks like the full commit message got added by accident.  Thanks for cleaning this up.

PR #3672, opened 2022-07-16
fix: @font-face Adds the font-display attribute to optimize FOIT in SEO

jerry-yo:
When I used Chrom's Lighthose tool to test site SEO, Lighthose suggested this
image
Here's what Google suggests
https://web.dev/font-display/?utm_source=lighthouse&utm_medium=devtools#how-to-avoid-showing-invisible-text
image
image

Ensure text remains visible during webfont load

### DIFF
--- a/src/fonts.less
+++ b/src/fonts.less
@@ -37,6 +37,7 @@
         .use-ttf(@family, @suffix);
         font-weight: @weight;
         font-style: @style;
+        font-display: swap;
     }
 }
 


ylemkimon:
IMHO, this should be up to the user to decide as swapping the math font to a system font may not be ideal.

edemaine:
What would be nice is if we could add a Less variable so that others could easily compile our `fonts.less` file with custom `font-display` option, without having to modify the `fonts.less` file.

PR #3681, opened 2022-08-15
fix: Use SVGs for built-up stretchy delimiters

ronkok:
This PR will eventually write a single SVG for a built-up delimiter, replacing the current stack of glyphs and thereby evading browser imprecision of glyph location. I have not yet got very far with this PR because I cannot get anything to run in my local repository. Even the main branch is inoperative.

If I try to run `yarn start`, the error message is:

```
(node:7640) ExperimentalWarning: Custom ESM Loaders is an experimental feature. This feature could change at any time
(Use `node --trace-warnings ...` to show where the warning was created)
node:internal/errors:477
    ErrorCaptureStackTrace(err);
    ^

Error [ERR_LOADER_CHAIN_INCOMPLETE]: "file:///C:/Users/Ronal/OneDrive/Documents/GitHub/KaTeX/.pnp.loader.mjs 'resolve'" did not call the next hook in its chain and did not explicitly signal a short circuit. If this is intentional, include `shortCircuit: true` in the hook's return.
    at new NodeError (node:internal/errors:387:5)
    at ESMLoader.resolve (node:internal/modules/esm/loader:860:13)
    at async ESMLoader.getModuleJob (node:internal/modules/esm/loader:439:7)
    at async Promise.all (index 0)
    at async ESMLoader.import (node:internal/modules/esm/loader:541:24)
    at async loadESM (node:internal/process/esm_loader:91:5)
    at async handleMainPromise (node:internal/modules/run_main:65:12) {
  code: 'ERR_LOADER_CHAIN_INCOMPLETE'
}
```

I have no idea how to overcome this. Any help would be appreciated. @edemaine? @ylemkimon? Help?

PR #3682, opened 2022-08-15
chore: Upgrade to yarn 3.2.2, Node 16

edemaine:
Adds support for Node 18

As documented by @ronkok in #3681, `yarn start` (and most `yarn` operations) crashes on Node 16 as follows:

```
(node:7640) ExperimentalWarning: Custom ESM Loaders is an experimental feature. This feature could change at any time
(Use `node --trace-warnings ...` to show where the warning was created)
node:internal/errors:477
    ErrorCaptureStackTrace(err);
    ^

Error [ERR_LOADER_CHAIN_INCOMPLETE]: "file:///C:/Users/Ronal/OneDrive/Documents/GitHub/KaTeX/.pnp.loader.mjs 'resolve'" did not call the next hook in its chain and did not explicitly signal a short circuit. If this is intentional, include `shortCircuit: true` in the hook's return.
    at new NodeError (node:internal/errors:387:5)
    at ESMLoader.resolve (node:internal/modules/esm/loader:860:13)
    at async ESMLoader.getModuleJob (node:internal/modules/esm/loader:439:7)
    at async Promise.all (index 0)
    at async ESMLoader.import (node:internal/modules/esm/loader:541:24)
    at async loadESM (node:internal/process/esm_loader:91:5)
    at async handleMainPromise (node:internal/modules/run_main:65:12) {
  code: 'ERR_LOADER_CHAIN_INCOMPLETE'
}
```

I followed the workaround in https://github.com/yarnpkg/berry/issues/4629. `yarn start` etc. are functional again.

I had to upgrade CI from Node 12 to 16 (or anything 14+) for Yarn to keep working.  Node 12 is deprecated, and 16 is the current active version.

**Thoughts for consideration:**
I'm not really convinced we should be using `yarn` anymore. I'd be tempted to use NPM or PNPM. But we might want to merge this anyway so at least we're functional again.

@ronkok Can you test that this works for you?


ronkok:
@edemaine Thank you. The Yarn upgrade got me back in business.

Do you know why some of the checks are failing? I'd like to approve this PR, but I hesitate to do so.

edemaine:
> Do you know why some of the checks are failing? I'd like to approve this PR, but I hesitate to do so.

Indeed, the website is failing to build. The issue seems to be https://github.com/yarnpkg/berry/issues/2470 but the workaround described there doesn't work. ☹️ 

Unless we can get some Yarn expertise from @ylemkimon, I'd be tempted to move to PNPM. I could try that in a separate PR.

merceyz:
@edemaine add `enableGlobalCache: false` to your `.yarnrc.yml` file.

edemaine:
Thanks @merceyz! Switching to stable Yarn 3.2.2 makes sense, and it looks like all tests are passing now! @ronkok could you test/approve/merge?

ronkok (approved):
Terrific. Back on track.

Issue #3647, opened 2022-06-04
Unable to build website on Windows


edemaine:
Within `website` I can `yarn start` on Linux on Node 17 and 18, but on Node 18 Windows I get the following error:

```
~/Projects/KaTeX/website ⟫ yarn start
(node:49280) ExperimentalWarning: Custom ESM Loaders is an experimental feature. This feature could change at any time
(Use `node --trace-warnings ...` to show where the warning was created)
(node:53936) ExperimentalWarning: Custom ESM Loaders is an experimental feature. This feature could change at any time
(Use `node --trace-warnings ...` to show where the warning was created)
Browserslist: caniuse-lite is outdated. Please run:
npx browserslist@latest --update-db

Why you should do it regularly:
https://github.com/browserslist/browserslist#browsers-data-updating
C:\Users\edemaine\Projects\KaTeX\website\.yarn\cache\@babel-core-npm-7.13.10-57ebf35dd5-9b3362fd02.zip\node_modules\@babel\core\lib\transformation\normalize-file.js:132
  t().traverseFast(ast, node => {
      ^

TypeError: t(...).traverseFast is not a function
    at extractComments (C:\Users\edemaine\Projects\KaTeX\website\.yarn\cache\@babel-core-npm-7.13.10-57ebf35dd5-9b3362fd02.zip\node_modules\@babel\core\lib\transformation\normalize-file.js:132:7)
    at normalizeFile (C:\Users\edemaine\Projects\KaTeX\website\.yarn\cache\@babel-core-npm-7.13.10-57ebf35dd5-9b3362fd02.zip\node_modules\@babel\core\lib\transformation\normalize-file.js:70:27)
    at normalizeFile.next (<anonymous>)
    at run (C:\Users\edemaine\Projects\KaTeX\website\.yarn\cache\@babel-core-npm-7.13.10-57ebf35dd5-9b3362fd02.zip\node_modules\@babel\core\lib\transformation\index.js:31:50)
    at run.next (<anonymous>)
    at Function.transform (C:\Users\edemaine\Projects\KaTeX\website\.yarn\cache\@babel-core-npm-7.13.10-57ebf35dd5-9b3362fd02.zip\node_modules\@babel\core\lib\transform.js:19:41)
    at transform.next (<anonymous>)
[...]

PR #3698, opened 2022-08-29
fix: update dependencies

ylemkimon:

Outdated dependencies

Updated dependencies.


package.json
```diff
@@ -173,7 +174,9 @@
     "testMatch": [
       "**/test/*-spec.js"
     ],
-    "testURL": "http://localhost/",
+    "testEnvironmentOptions": {
+      "url": "http://localhost/"
+    },
```
ylemkimon, line 179: https://jestjs.io/docs/28.x/upgrading-to-jest28#testurl

edemaine (approved):
I realize this is now out of date, but it brings us closer to up-to-date dependencies. I tested what I could (website still doesn't build for me because of #3647), and tests look good.

PR #3717, opened 2022-09-20
Update where to report vulnerabilities to in the docs

kevinbarabash:
This is a minor change to security.md to link to the Security Advisories within in the KaTeX repo instead of Khan Academy's hackerone.com account.  This should streamline the reporting of vulnerabilities and make it easier for KaTeX maintainers to interact with reporters.

### DIFF
--- a/docs/security.md
+++ b/docs/security.md
@@ -18,4 +18,4 @@ with untrusted inputs; refer to [Options](options.md) for more details.
 The error message thrown by KaTeX may contain unescaped LaTeX source code.
 See [Handling Errors](error.md) for more details.
 
-> If you discovered a security issue, please let us know via https://hackerone.com/khanacademy
+> If you discovered a security issue, please let us know via https://github.com/KaTeX/KaTeX/security/advisories


edemaine:
Hi @kevinbarabash! Unfortunately, regular users (and even I) cannot create security advisories at that link. It seems we need to receive email, and then someone with Admin privileges needs to make the security advisory and invite the user to report. We could use the maintainer mailing list that I created years ago, but I'm not sure whether it's up-to-date.

kevinbarabash:
@edemaine I didn't realize it was so restrictive.  How would using the mailing list work?  If a report emails the list, but isn't on the list does that work?

kevinbarabash:
I'm going to close this PR since it doesn't make sense.

edemaine:
[GHSA docs](https://docs.github.com/en/code-security/repository-security-advisories/about-github-security-advisories-for-repositories) recommend contacting the maintainers of a package to request that they (we) create a GHSA and invite the reporter to. I was suggesting listing the mailing list as the contact point, so we'd all get the notification and then whoever has time first can create the GHSA.

edemaine:
P.S. @kevinbarabash Are you receiving email from me? I sent a couple of messages in the past couple months. No worries if you're just busy; just want to make sure they're not going to the wrong email address.

PR #3719, opened 2022-09-22
Update "No character metrics" documentation

jameshfisher:
1. This page didn't appear for me when searching for "No character metrics". I then stumbled onto it, and it's quite useful
2. metrics/extract_ttfs.py does not exist, and it's not the "root" of the metric-generation stuff


edemaine (approved):
Thanks for the docs improvement!

PR #3732, opened 2022-10-04
docs(support_table): fix escaping of accent macros

little-cindy:
Some KaTeX codes such as `\text{\\={a}}` are wrong.

Some of these types of errors are fixed.


edemaine (approved):
Thank you for spotting and fixing these!

PR #3735, opened 2022-10-10
fix: \hline after \cr

### LINKED ISSUE #3734: `\hline` not working after `\cr`
* `\\` is OK
```katex
\begin{array}{}
  a & b \\ \hline
  c & d
\end{array}
```
hline_true

* `\newline` makes a parsing error
```katex
\begin{array}{}
  a & b \newline \hline
  c & d
\end{array}
```
hline_false

https://katex.org/#demo

ronkok:
Support `\hline` when it occurs after a `\cr` within an environment. Fixes #3734.

### DIFF
--- a/src/environments/array.js
+++ b/src/environments/array.js
@@ -39,6 +39,11 @@ function getHLines(parser: Parser): boolean[] {
     const hlineInfo = [];
     parser.consumeSpaces();
     let nxt = parser.fetch().text;
+    if (nxt === "\\relax") { // \relax is an artifact of the \cr macro below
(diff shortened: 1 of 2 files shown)


edemaine (approved):
Thanks for fixing this!

At first I thought this felt a bit hacky, but I think it's a general property of how we're simulating `\cr` and `\hline`, which are fairly different from how LaTeX does it. In the end, I can't think of a better way, except possibly adding a `consumeRelax()` feature to the parser. But it seems we don't actually (need to) do this anywhere else.

PR #3738, opened 2022-10-22
fix: Support `\let` via `macros` option

### LINKED ISSUE #3737: Equivalent of `\let` via `macros` options
**Describe the bug:**
It's certainly not convenient to do the equivalent of `\let\Oldint=\int` via a user-constructued `macros` option. In fact, it may not be possible.

**To Reproduce:**
Steps to reproduce the behavior, in Node or browser console:
1. `macros = {};`
2. `katex.renderToString("\\let\\Oldint=\\int\\def\\int{\\Oldint\\limits}", {macros, globalGroup: true});`
3. `Object.values(macros).forEach((val) => val.tokens.forEach((token) => delete token.loc))` — remove `loc` pointers which prevent serialization
4. `console.log(JSON.stringify(macros))`
    &rarr; `{"\\Oldint":{"tokens":[{"text":"\\int","noexpand":true}],"numArgs":0,"unexpandable":false},"\\int":{"tokens":[{"text":"\\limits"},{"text":"\\Oldint"}],"numArgs":0,"delimiters":[[]]}}`
5. `katex.renderToString("\\int", {macros})` works fine
6. But `katex.renderToString("\\int", {macros: {"\\Oldint":{"tokens":[{"text":"\\int","noexpand":true}],"numArgs":0,"unexpandable":false},"\\int":{"tokens":[{"text":"\\limits"},{"text":"\\Oldint"}],"numArgs":0,"delimiters":[[]]}}})` causes an infinite loop (!).

**Expected behavior:**
Manually constructing the `macros` object in the same shape [as produced by `\let`](https://github.com/KaTeX/KaTeX/blob/c5ac4c769588761fb68d89890998234e83fc1c1a/src/functions/def.js#L37) should simulate the behavior of `\let`.

We might also want to document this shape, or otherwise make it easier to request `\let` behavior for built-in commands.

Also, an infinite loop should not be possible, even with invalid `macros` input. I would expect `maxExpand` to hit a limit here if nothing else.

**Screenshots:**

 - KaTeX Version: 0.16.2
 - Device: Desktop
 - OS: Windows 10
 - Browser: Chrome or Node
 - Version: 106.0.5249.119 or v18.7.0

**Additional context:**
From #3736

edemaine:
As described in #3737, `katex.renderToString("\\int", {macros: {"\\Oldint":{"tokens":[{"text":"\\int","noexpand":true}],"numArgs":0,"unexpandable":false},"\\int":{"tokens":[{"text":"\\limits"},{"text":"\\Oldint"}],"numArgs":0,"delimiters":[[]]}}})` results in an infinite loop.

Fixes #3737

The issue turned out to be how we handled the return value of `expandOnce`. We assumed that, if the return value isn't an `Array`, it's an `instanceof Token`.  This isn't necessary true with a user's `macros` object, and given that we don't currently export `Token`, it's pretty difficult to bypass.

Given that we never actually use the array return values from `expandOnce`, I changed the return value for `expandOnce` to either a `number` (to indicate the number of expanded tokens, so you could still look up the tokens in the stack if you wanted to) or `false` (to indicate no expansion happened).  We can't use `0` for the latter because an actual expansion might result in zero tokens. The resulting code is arguably cleaner.

I also documented that `macros` can have object expansions, and specified how to simulate `\let`.


ronkok:
Would you be willing to consider an alternate approach? In the [Temml](https://temml.org/) library, I have exposed an additional function, `definePreamble(<str>)`. A web site administrator would run this function before any math is rendered, as in this example:

```
// Optional preamble.
const macros = temml.definePreamble(
    `\\let\\Oldint=\\int
    \\def\\int{\\Oldint\\limits}
    \\newcommand\\d[0]{\\operatorname{d}\\!}
    \\def\\foo{x^2}`
);
```

As you can see, the argument is a string containing macros written as they would be written in a document. The function returns a `macros` object. This approach gives an author a way to write macros in a syntax that they already know. No need to learn how the `macros` object is internally constructed.

`definePreamble()` was not that hard to code. In the Temml.js source file, I added and exposed a new function:

```
/**
 * Take an expression which contains a preamble.
 * Parse it and return the macros.
 */
const definePreamble = function(expression, options) {
  const settings = new Settings(options);
  settings.macros = {};
  if (!(typeof expression === "string" || expression instanceof String)) {
    throw new TypeError("Temml can only parse string typed expression")
  }
  const parser = new Parser(expression, settings, true)
  // Blank out any \df@tag to avoid spurious "Duplicate \tag" errors
  delete parser.gullet.macros.current["\\df@tag"]
  const macros = parser.parse()
  return macros
};
```

As you can see, the function uses the existing parser to do the work. No need to reinvent a new wheel. If you look very closely, you can see that the call to `new Parser()` includes a new, third, optional argument. It’s a boolean named `isPreamble`.

After the parser has parsed the top-level expression, it checks if it is engaged in a preamble. If so, it extracts the macros and executes a quick return.

```
if (this.isPreamble) {
    const macros = Object.create(null)
    Object.entries(this.gullet.macros.current).forEach(([key, value]) => {
        macros[key] = value
      })
    this.gullet.endGroup();
    return macros
}
```

That’s about it. Temml does not have a `globalGroup` rendering option, which may complicate a KaTeX version of `definePreamble`. Hopefully not too much.

edemaine:
@ronkok KaTeX already has the equivalent of `definePreamble` (though uglier):

```js
const macros = {}
katex.renderToString(`...\let\foo=bar...`, {macros, globalGroup: true})
```

(The main motivation for `globalGroup`, I believe, is running in the preamble.)

It may be nice to add a helper for this, but at least it's possible. Probably we should document this approach at least.

I still think this PR is an improvement over the current code, making the `macros` object a bit more robust. (It's currently very easy to cause KaTeX to crash by passing in bad `macros`.) But perhaps I should remove the documentation of how to do `\let` this way, and instead add documentation of the above approach. What do you think?

docs/options.md
```diff
@@ -15,7 +15,11 @@ You can provide an object of options as the last argument to [`katex.render` and
...
+  - Each macro is a property with a name like `\name` (written `"\\name"` in JavaScript) which maps to a string that describes the expansion of the macro, or a function that accepts an instance of `MacroExpander` as first argument and returns the expansion as a string. `MacroExpander` is an internal API and subject to non-backwards compatible changes. See [`src/defineMacro.js`](https://github.com/KaTeX/KaTeX/blob/main/src/defineMacro.js) for its usage.
+  - Single-character keys can also be included in which case the character will be redefined as the given macro (similar to TeX active characters).
+  - *This object will be modified* if the LaTeX code defines its own macros via `\gdef` or `\global\let` (or via `\def` or `\newcommand` or `\let` when using `globalGroup`), which enables consecutive calls to KaTeX to share state.
+  - Expansions can also be objects matching [an internal `MacroExpansion` specification](https://github.com/KaTeX/KaTeX/blob/main/src/defineMacro.js), which is what results from global `\def` or `\let`. For example, you can simulate the effect of `\let\realint=\int` via `{"\\realint": {tokens: [{text: "\\int", noexpand: true}], numArgs: 0}}`.
```
ronkok, line 22: I'd like to propose an edited version of the `macros` entry, with an eye to toward less experienced webpage administrators.

- `macros`: `object`. A collection of custom macros.

  - Each macro is a key-value pair in which the key is a new KaTeX function name and the value is the expansion of the macro. Example: `macros: {"\\R": "\\mathbb{R}"}`.

  - To be more precise, each macro is a property with a name like `\name` (written `"\\name"` in JavaScript) which maps to a string that describes the expansion of the macro, or a function that accepts an instance of `MacroExpander` as first argument and returns the expansion as a string. `MacroExpander` is an internal API and subject to non-backwards compatible changes. See [`src/defineMacro.js`](https://github.com/KaTeX/KaTeX/blob/main/src/defineMacro.js) for its usage.

  - Single-character keys can also be included in which case the character will be redefined as the given macro (similar to TeX active characters).

  - The `macros` object _will be modified_ **if** the LaTeX code defines its own macros via `\gdef` or `\global\let` (or via `\def` or `\newcommand` or `\let` when using `globalGroup`). This enables macro definitions to [persist](https://katex.org/docs/api.html#persistent-macros) between calls to KaTeX.

  - Expansions can also be objects matching [an internal `MacroExpansion` specification](https://github.com/KaTeX/KaTeX/blob/main/src/defineMacro.js), which is what results from global `\def` or `\let`. For example, you can simulate the effect of `\let\realint=\int` via `{"\\realint": {tokens: [{text: "\\int", noexpand: true}], numArgs: 0}}`.

ronkok (changes requested):
All the code changes look good. I've proposed some re-wording of the docs for `macros.`

ronkok (approved):
It's good code. Let's go with it.

docs/options.md
edemaine: Thanks for the comments, and sorry for taking so long to get to this. I just took a revision pass inspired by your feedback. Let me know if you have further comments, or I'll merge later today. (We can always edit further in another PR.)

ronkok: The macros entry looks much better. This is good to go.

PR #3746, opened 2022-11-01
fix: space should prevent optional argument to \\

### LINKED ISSUE #3745: The error "KaTeX parse error: Invalid size" occurred when using `\\ [`
**Describe the bug:**

When using `\\[` such as:
```
a+b=c\\
[f(x)]^2
```
The error `KaTeX parse error: Invalid size: 'f(x)' at position 9: a+b=c\\ [̲f̲(̲x̲)̲]̲^2` occurred.

**To Reproduce:**
Can be reproduced at [https://katex.org/](https://katex.org/).

 - KaTeX Version: 0.16.3

edemaine:
Fix `src/functions/cr.js`'s definition of `\\` to manually look for an optional argument via `future()` instead of `numOptionalArgs`, so that it does *not* skip over spaces.  This matches the existing behavior of `\\` in `src/environments/array.js` and AMSMath's behavior of `\math@cr` via `\new@ifnextchar`.

Fixes #3745; see that for details.

### DIFF
--- a/src/functions/cr.js
+++ b/src/functions/cr.js
@@ -13,13 +13,13 @@ defineFunction({
     names: ["\\\\"],
     props: {
         numArgs: 0,
-        numOptionalArgs: 1,
-        argTypes: ["size"],
+        numOptionalArgs: 0,
         allowedInText: true,
     },
 
     handler({parser}, args, optArgs) {
-        const size = optArgs[0];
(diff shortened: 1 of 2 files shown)


ronkok (approved):
Looks good to me.

ronkok:
@edemaine Thank for following up on this.

PR #3777, opened 2023-01-18
feat: Support bold Fraktur

### LINKED ISSUE #3776: "Font metrics not found for font" error
Try typesetting the randomly chosen unicode symbol 𝕭 (this can be tested e.g. in https://katex.org/)
KaTeX will produce an error:
```
Font metrics not found for font: .
```
This is a bit brutal. Maybe KaTeX could handle this more gracefully?

ronkok:
Support bold Fraktur. Resolves #3776.

Note: KaTeX has a font for bold Fraktur, but no font metrics. That is why I left bold Fraktur out of the original PR for wide-character Unicode support. In this PR, I evade the lack of metrics by using the font metrics for _regular_ Fraktur, not bold Fraktur.


ronkok:
The second commit makes this PR more general in addressing the complaint of issue #3776. The results of function `wideCharacterFont()` are no longer used when inapplicable. That avoids all the bad error messages that would come from inapplicable use.

edemaine (approved):
Sorry for the long delay in review here. Slowly getting back into things...

Overall, this seems like a good fix to bold Fraktur, so this PR should be probably be merged.  However, I noticed one thing more generally about Fraktur support:

I just tested and `\text{𝔄-ℨ}` generates an error, because `\u212D` (Z), `\u210C` (H), and `\u2128` (Z) are macro'd in `src/macros.js` to use `\mathfrak` but sometimes we want `\textfrak`. (Not sure why this doesn't generate an error when building the documentation. Perhaps it does and it's being silently ignored.)

Perhaps this is legacy code and can simply be removed?

https://github.com/KaTeX/KaTeX/blob/5dd9bc4a47c508b064ce341a7fc92e752fb3f2ab/src/macros.js#L256-L258

edemaine:
P.S. I also looked into generating actual metrics for the bold Fraktur font. (Part of the delay.) My attempts failed, though, so I think this should go as is; we can consider adding them later if I figure out how metrics get built.

src/buildCommon.js
```diff
@@ -165,9 +165,11 @@ const makeOrd = function<NODETYPE: "spacing" | "mathord" | "textord">(
     // Math mode or Old font (i.e. \rm)
     const isFont = mode === "math" || (mode === "text" && options.font);
     const fontOrFamily = isFont ? options.font : options.fontFamily;
-    if (text.charCodeAt(0) === 0xD835) {
+    const [wideFontName, wideFontClass] = text.charCodeAt(0) !== 0xD835
+      ? ["", ""]
+      : wideCharacterFont(text, mode);
+    if (wideFontName.length > 0) {
```
edemaine, line 171: It would be nice to avoid the array creation in the negative case. How about this:
```js
let wideFontName, wideFontClass;
if (text.charCodeAt(0) === 0xD835) {
  [wideFontName, wideFontClass] = wideCharacterFont(text, mode);
}
if (wideFontName) {
```

ronkok: I like it.

edemaine:
> Perhaps this is legacy code and can simply be removed?

@ronkok Do you agree with this assessment? If so, I can remove the code and make the other requested change.

ronkok:
> Perhaps this is legacy code and can simply be removed?

I don't think that is correct. Most Fractur characters are outside the Unicode Basic Multilingual Plane (BMP).  They are addressed by the wide character function modified in this PR.

But a few Fractur characters are in the BMP. KaTeX defines them in these macros. 
https://github.com/KaTeX/KaTeX/blob/ad03d1e9dc1325a95072f58cf061b579a81bb9a9/src/macros.js#L247-L258

I suppose I could have instead defined them in `symbols.js`. That would probably be better than the macro method. It would be easier in `symbols.js` to define them both as a `mathord` and as a `textord`. The lack of a `textord` definition is causing the error.

I think a `symbols.js` definition of the BMP characters should be located below the loop where wide character Frakturs are defined.
https://github.com/KaTeX/KaTeX/blob/ad03d1e9dc1325a95072f58cf061b579a81bb9a9/src/symbols.js#L815-L817

Then the BMP definitions would overwrite the first definition.

Issue #437, opened 2016-02-10
[auto-render] add an option to ignore \$ outside of latex expressions


PaulSohn:
If it's possible for escaping \$ to normal $, then latex $ (expressions) $ would not be any problem.
I wanted to make my own hack for this, but I'm not used to string so far.
Is there any plan to support dollar sign escaping?

spicyj:
We already have logic that should skip over backslashes:

https://github.com/Khan/KaTeX/blob/c79fb5893686ac0b8c4d8ec33ea1b940d2e4a0c4/contrib/auto-render/splitAtDelimiters.js#L16

spicyj:
Maybe it's not working?

xymostech:
That only works for escaping `$` inside of math, but not outside of it. @PaulSohn can you give an example of what you're trying to do?

decademoon:
If this is the delimiter:

``` js
{ left: '$', right: '$', display: false }
```

and I want to render this text:

``` text
I will have you \$2 if you solve $y = x^2$.
```

The first $ is ignored but the `\` is still output.

decademoon:
Rendering this text:

``` text
I will give you \$2 if you can solve $y = x^2$.
```

with this delimiter:

``` js
{ left: '$', right: '$', display: false }
```

results in

``` text
<#text>I will give you \</#text>
<span>KATEX MATH</span>
<#text>y = x^2</#text>
<#text>$.</#text>
```

where `<#text>` denotes a text DOM node.

erhnsn:
Agreed, can not write dollar sign outside of math equations...

kevinbarabash:
@PaulSohn can you use different delimiters, e.g. `\(`,`\)` and `\[`,`\]`?

kevinbarabash:
I'm going to close this since it's going to be very difficult to implement something that handles the different behaviors people want.  Also, there is a workaround by using different delimiters.

ronkok:
There is more than one way to do this. If you are using the KaTeX [auto-render extension](https://khan.github.io/KaTeX/docs/autorender.html), then you can delimit your math with `\(…\)` instead of `$…$`. Or you can nest the whole phrase in `<noscript>…</noscript>` tags.

Some markdown parsers only recognize the closing `$` as a math delimiter if it is **not** followed by a number. Sometimes I think we should do that too.

shazrazdan:
[...]

PR #3798, opened 2023-03-30
feat(auto-render): add option to ignore escaped special characters outside of LaTeX math mode (#437)

lnasrc:
feat(auto-render): add option to ignore escaped special characters outside of LaTeX math mode (#437)

Some escapable characters were not properly escaped in math LaTeX mode in auto-render.
For instance, in the following LaTeX string: 
$$
\\$100 + \\$100 = \\$200
$$
the "\\$" should be parsed as "$" symbols.

This PR introduces a new option to auto-render:
- `supportEscapedSpecialCharsInText`: `boolean` (default: `false`). If `true`, `\$` are ignored outside of LaTeX math expressions.
  For example: `Please enjoy this 2\$ coffee and I'll explain why $e = mc^2$.`

When set to `true`, it fixes the issue and properly escape those LaTeX escapable characters.
When set to `false`, the existing behavior is kept to avoid breaking existing uses.

The tests have been updated accordingly, and some tests added to show those fixed behaviors.

### DIFF
--- a/contrib/auto-render/auto-render.js
+++ b/contrib/auto-render/auto-render.js
@@ -7,18 +7,30 @@ import splitAtDelimiters from "./splitAtDelimiters";
  * API, we should copy it before mutating.
  */
 const renderMathInText = function(text, optionsCopy) {
-    const data = splitAtDelimiters(text, optionsCopy.delimiters);
+    const data = splitAtDelimiters(text, optionsCopy.delimiters,
+        optionsCopy.supportEscapedSpecialCharsInText);
     if (data.length === 1 && data[0].type === 'text') {
         // There is no formula in the text.
         // Let's return null which means there is no need to replace
         // the current text node with a new one.
-        return null;
+        if (!optionsCopy.supportEscapedSpecialCharsInText) {
+            return null;
+        }
     }
 
     const fragment = document.createDocumentFragment();
 
     for (let i = 0; i < data.length; i++) {
         if (data[i].type === "text") {
+            if (optionsCopy.supportEscapedSpecialCharsInText) {
+                data[i].data = data[i].data.replace(/\\\$/g, '$');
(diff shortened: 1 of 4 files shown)


lnasrc:
> I disagree with this premise (assuming I understand it correctly):
> 
> > For instance, in the following LaTeX string:
> > ```
> > $$
> > $100 + $100 = $200
> > $$
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > the "$" should be parsed as "$" symbols.
> 
> In fact, `$`s have a specific meaning inside LaTeX math, which is to re-enter math mode inside `\text`. For example:
> 
> ```
> $$
> \{x \mid x \text{ is divisible by $y$ or $z$}\}
> $$
> ```
> 
> So you can't just escape them.
> 
> `Please enjoy this 2\$ coffee` should already be ignored by auto-render, though #3775 is a possible improvement to that.

Thanks for your reply. There was an issue with the formatting in my previous answer, where the backslashes had been removed.
I have edited my answer to fix the formatting.
Sorry for that!!

Moreover, the PR I propose does not bring any breaking change, and solve this issue mentioned in #3775: 

> The only way existing users are affected by this is if they have a backslash immediately (no whitespace or anything) followed by a delimiter that enters math mode.

edemaine:
Ahh, now I understand. Thanks for explaining! That does sound useful.

LNA-h:
This PR is ready for final review, thanks in advance.
How can I update the Codecov report, based on the latest commit?

jyc:
Thanks for making this! Would also be interested in helping get this over the finish line if any reviewers are willing to take a look.

lnasrc:
This PR is ready for final review, thanks in advance 🙏 
cc @edemaine

PR #3805, opened 2023-04-11
feat(api): __defineFunction API exposing internal defineFunction

### LINKED ISSUE #3756: expose defineFunction
We would like to be able to add a custom command with a custom render. I don't see a way right now to add a custom render function to convert a command to HTML.

It would be great if we could call defineFunction and provide our our own html render function.

defineSymbol and defineMacro won't work in this case because we want to render custom html. We've considered adding a new symbol and then replacing that symbol with custom html after it renders, but there is no way to determine function arguments.

**Additional context:**
We want to create a drag and drop question interface with complex expressions. We want to add a custom command called `\drop{id}`. Having a custom function that we could define ourselves would allow us to render an html drop target within the LaTeX render.

edemaine:
KaTeX libraries should be allowed to define functions.

We offered `__defineMacro` and `__defineSymbol`, but no `__defineFunction`.

Expose `defineFunction` via `__defineFunction`.  Like `__defineMacro` and `__defineSymbol`, the API isn't documented.

Fixes #3756.


ronkok (approved):
For this feature, I suspect that the number of potential users is one.

In which case, yay!, go for it.

If you dream up any particularly useful functions, let me know about them. Maybe I'll add them to Temml.

PR #3814, opened 2023-04-27
fix(docs/support_table.md): delete redundant "varPsi"

untitledunrevised:
In page [support_table](https://katex.org/docs/support_table.html#v), "varPsi" repeated twice.

"varPsi" will display only once.


edemaine (approved):
Thanks for spotting this.

Issue #2384, opened 2020-08-04
Should katex.org support IE 11?


ylemkimon:
I'm considering migrating the website to Docusaurus 2. However, Docusaurus 2 doesn't support IE 11.

kevinbarabash:
What are the benefits of moving to Docusaurus 2?

kevinbarabash:
I had a look at https://docusaurus.io/blog/2018/09/11/Towards-Docusaurus-2 and it seems like thing we'd benefit from the most right now is the changes to how versioned docs are being handled.

ylemkimon:
@kevinbarabash I don't have a specific in mind for now.

grigoriy-reshetniak:
IE 11 support for KaTeX [was dropped](https://github.com/KaTeX/KaTeX/pull/4168). There's no point for katex.org support it either.

PR #3820, opened 2023-05-15
feat: expose error length and raw error message on ParseError

mangerlahn:
Add length and rawError as public parameters of `ParseError` to allow for richer error UIs in editors. This information was already there, but only encoded into the errors message. This change is purely additional, the existing parameters have not changed.

`ParseError` consists of a position (integer) and message (string) that describe the error. Inside the message string, the position, length (through underline) and actual cause are encoded.

The existing behavior stays the same, `ParseError` gains two new properties, `length` (integer) and `rawError` (string). `length` together with the existing `position` describes the range affected by the error. `rawError` exposes the underlying error message, without the formatting that is added within `ParseError`.

**Reason**
We currently implement formula editing and rendering in our app and found this information useful yet missing from the current implementation.

This is what we are currently doing. Showing the raw error below the code while underlining the affected text.
Inline - Error

**Testing**
Since this only exposes existing values, the behavior has not changed. I am happy to update the tests in error-specs.js to both check for position and length of the error. However, I did not do that yet because both values are indirectly tested through the message string (rendering the underline).

### DIFF
--- a/src/ParseError.js
+++ b/src/ParseError.js
@@ -12,15 +12,21 @@ import type {AnyParseNode} from "./parseNode";
  * about where in the source string the problem occurred.
  */
 class ParseError {
+    name: "ParseError";
     position: number | void;
-        // Error position based on passed-in Token or ParseNode.
+        // Error start position based on passed-in Token or ParseNode.
+    length: number | void;
+        // Length of affected text based on passed-in Token or ParseNode.
+    rawMessage: string | void;
+        // The underlying error message without any context added.
 
     constructor(
         message: string,               // The error message
         token?: ?Token | AnyParseNode, // An object providing position information
-    ): Error {
+    ): ParseError {
         let error = "KaTeX parse error: " + message;
         let start;
+        let end;
 
         const loc = token && token.loc;
         if (loc && loc.start <= loc.end) {
@@ -31,7 +37,7 @@ class ParseError {
 
             // Prepend some information
             start = loc.start;
-            const end = loc.end;
+            end = loc.end;
             if (start === input.length) {
                 error += " at end of input: ";
             } else {
@@ -60,12 +66,16 @@ class ParseError {
 
         // Some hackery to make ParseError a prototype of Error
         // See http://stackoverflow.com/a/8460753
-        const self = new Error(error);
+        // $FlowFixMe
+        const self: ParseError = new Error(error);
         self.name = "ParseError";
         // $FlowFixMe
         self.__proto__ = ParseError.prototype;
-        // $FlowFixMe
         self.position = start;
+        if (start != null && end != null) {
+            self.length = end - start;
+        }
+        self.rawMessage = message;
         return self;
     }
 }
--- a/test/katex-spec.js
+++ b/test/katex-spec.js
@@ -5,6 +5,7 @@ import buildTree from "../src/buildTree";
 import katex from "../katex";
 import parseTree from "../src/parseTree";
 import Options from "../src/Options";
+import ParseError from "../src/ParseError";
 import Settings from "../src/Settings";
 import Style from "../src/Style";
 import {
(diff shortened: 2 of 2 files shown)


src/ParseError.js
```diff
@@ -12,15 +12,17 @@ import type {AnyParseNode} from "./parseNode";
  * about where in the source string the problem occurred.
  */
 class ParseError {
-    position: number | void;
-        // Error position based on passed-in Token or ParseNode.
+    position: number | void;    // Error position based on passed-in Token or ParseNode.
```
edemaine, line 15: Maybe mention "start position" in this comment? Thanks for adding a comment!

edemaine (changes requested):
Sorry for the delay in reviewing this. This looks great, and useful, and the changes are minimal.

Please tweak the comment placement (perhaps on their own line) to satisfy the line length lint.

Regarding testing, please add one or two tests that check that these properties are properly set on the returned error (perhaps where `position` is already tested, if any). I don't think we need to check the properties for every erroring test, just in a couple of examples so that we ensure that these properties accidentally don't get set in a future regression.

src/ParseError.js
```diff
@@ -66,6 +71,8 @@ class ParseError {
         self.__proto__ = ParseError.prototype;
         // $FlowFixMe
         self.position = start;
+        self.length = end - start;
```
edemaine, line 74: It occurs to me that this computes `NaN` if `start` and `end` are undefined. Indeed, Flow caught this as well. Perhaps this line should be wrapped in `if (start != null && end != null)`?

Apparently we also need `// $FlowFixMe` because `self` is typed as `Error` here. But maybe this could be fixed via type casting? (In TypeScript it'd be `as`.)

(You can run `yarn test` to run Lint, Jest, and Flow yourself.)

test/katex-spec.js
```diff
@@ -3060,6 +3061,53 @@ describe("A parser that does not throw on unsupported commands", function() {
...
+            katex.renderToString("\\verb|hello\nworld|");
+
+            // Render is expected to throw, so this should not be called.
+            fail();
+        } catch (error) {
+            expect(error).toBeInstanceOf(ParseError);
+            expect(error.message).toBe("KaTeX parse error: \\verb ended by end of line instead of matching delimiter");
+            expect(error.rawMessage).toBe("\\verb ended by end of line instead of matching delimiter");
+            expect(error.position).toBeUndefined;
+            expect(error.length).toBeUndefined;
```
edemaine, line 3106: I think these need to be `toBeUndefined()`, which is why the `length` NaN bug wasn't detected.

edemaine (changes requested):

src/ParseError.js
mangerlahn: Ah okay, I always just used `yarn test:jest`, makes more sense now 😅

mangerlahn: I tried to cast this error, but that does not seem to help. I get syntax errors, so I assume that this is not TypeScript?
Sorry, I am super unexperienced with JS.

Using `// $FlowFixMe` works fine.

test/katex-spec.js
```diff
@@ -3060,6 +3061,53 @@ describe("A parser that does not throw on unsupported commands", function() {
     });
 });
 
+describe("ParseError properties", function() {
+    it("should contain affected position and length information", function() {
+        try {
+            katex.renderToString("1 + \\fraq{}{}");
+
+            // Render is expected to throw, so this should not be called.
+            fail();
```
mangerlahn, line 3070: The linter throws an error on the `fail` command I used in the tests, because this function appears to be defined in a different package called Jasmine. I could add that to the `.eslintrc` file (appears to work), but I'm not sure if that's desired. 

Alternatively I could replace fail with some other failing check like `expect(true).toBe(false)`. What do you prefer?

src/ParseError.js
mangerlahn: Thanks for the hint! However this all does not appear to work. As far as I can see, `ParseError` and `Error` are distinct classes, having nothing to do with each other. So if I cast `self` to `ParseError`, I just get errors when returning `self` from the constructor, which expects `Error`. So it's just the other way around. I followed the StackOverflow link at the top of the class declaration. The top answer provides an alternative available since ES6 by just extending `Error`. This would allow to remove some of the hackery around `ParseError`.

However, lint is unhappy with this, because extending classes appears to be prohibited in the `eslintrc` file (`ClassDeclaration[superClass]`).

If I remove this restriction, the tests appear to run fine on my machine 😅

edemaine:
> I just updated the PR, maybe it's better to see some code rather than discussing this in the comments. ParseError now extends Error, not sure if that's desired. I isolated this change in its own commit for now.

This code is so much cleaner... However, I read https://stackoverflow.com/q/1382107 a bunch (linked from the source here and from https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error) and it looks like `class ParseError extends Error` cannot be reasonably transpiled for IE 11. Currently, IE 11 is listed as an officially supported platform of KaTeX.  That said, there's some good reason to remove that support, mentioned in #2384, and given that IE 11 has been end-of-life for over a year (though you can still download it).  Thoughts from anyone on this?  I do like how much cleaner this code is, but the code was also OK as it was before...

edemaine:
By the way, I took a crack at making the old code compatible with Flow. Here's the best I came up with, which uses "only" two `$FlowFixMe`s.

```js
class ParseError {
    name: "ParseError";
    ...
    constructor(
        message: string,               // The error message
        token?: ?Token | AnyParseNode, // An object providing position information
    ): ParseError {
        ...
        // $FlowFixMe
        const self: ParseError = new Error(error);
        self.name = "ParseError";
        // $FlowFixMe
        self.__proto__ = ParseError.prototype;
        self.position = start;
        if (start != null && end != null) {
            self.length = end - start;
        }
        self.rawMessage = message;
        return self;
    }
}
```
I'm still not sure which version is best though, as described in my previous message.

mangerlahn:
Either is fine by me (again having little experience with web dev). If IE11 support is blocking the first solution, I don't see a reason to not use your second version. Could still be changed when that support is removed, right?

edemaine:
Yes. Perhaps we could merge the IE 11 compatible version now, and make  separate cleanup PR for dropping IE 11?

edemaine (approved):

mangerlahn:
Thanks for your patience! 😃

edemaine:
Sorry, one more thing: can you update against the current main branch? Merge is fine; we squash PRs. (You don't seem to have given me permission to do so.)

mangerlahn:
The "test" check was not successful, anything I need to do here?

PR #3821, opened 2023-05-17
docs: add dark-mode logo

adamlui:
Adds a light logo to header to appear in dark mode so it doesn't appear almost empty.

Logo virtually disappears if user is browsing in dark mode:

Screenshot_2023-05-17-07-16-25-80_3aea4af51f236e4932235fdada7d1643

Logo is clearly visible if user is browsing in dark mode:

Screenshot_2023-05-17-07-29-11-10_3aea4af51f236e4932235fdada7d1643


edemaine (approved):
This looks great — thanks for fixing!

PR #3833, opened 2023-06-13
docs(libs.md): add Sphinx extension

hagenw:
Adds the [sphinxcontrib-katex](https://github.com/hagenw/sphinxcontrib-katex) Sphinx extension to the "Extensions & Libraries" section of the documentation.

At the moment the entries under "Libraries" follow no particular order, maybe we order them alphabetically?

### DIFF
--- a/docs/libs.md
+++ b/docs/libs.md
@@ -47,7 +47,7 @@ you'll need to first convert AsciiMath into LaTeX input, then call KaTeX.
 
 - [asciimath2tex](https://github.com/christianp/asciimath2tex): Converts AsciiMath to TeX, with KaTeX in mind
 
-### Canvas LaTeX
+### Canvas
 
 - [canvas-latex](https://github.com/CurriculumAssociates/canvas-latex): Renders mathematical expressions on HTML5's canvas element. Supports popular libraries like: CreateJS, and PIXI.
 
--- a/docs/libs.md
+++ b/docs/libs.md
@@ -15,17 +15,32 @@ These extensions are provided by KaTeX.
 
 These libraries are maintained by third-parties.
 
+### AsciiMath
+
+If you want to render math written in [AsciiMath](http://asciimath.org/),
+you'll need to first convert AsciiMath into LaTeX input, then call KaTeX.
+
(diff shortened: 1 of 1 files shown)


docs/libs.md
```diff
@@ -60,3 +60,6 @@ you'll need to first convert AsciiMath into LaTeX input, then call KaTeX.
 
 - [@rojer/katex-mini](https://github.com/rojer95/katex-mini): A Wechat Mini Program library that uses KaTeX for TeX math rendering.
```
edemaine, line 61: Accidental deletion? This should be kept, unless there's a reason to delete it.

docs/libs.md
```diff
@@ -60,3 +60,6 @@ you'll need to first convert AsciiMath into LaTeX input, then call KaTeX.
...
 
+### Sphinx Extension
```
edemaine, line 63: I'd maybe remove Extension here. (I think "LaTeX" should also be removed in "Canvas LaTeX".)

edemaine (commented):
Thanks for the addition!

Organizing is a good idea. Perhaps we should divide Libraries into implicit sections (maybe documented in comments), or replace the current headings with these and turn each section into a bulleted list (but I see web components has two items already, so maybe this doesn't work as well...)? I'm thinking Mobile (Android/iOS/Wechat), Web (Angular/Canvas/Vue), and CLI/misc (AsciiMath/Ruby/Rust), and alphabetical within? Or just alphabetical like you suggest. :-)

hagenw:
> CLI/misc (AsciiMath/Ruby/Rust)

I updated the entries by sorting them alphabetically.
By not grouping them to "Mobile" ... sub-groups, we still have the advantage that you can easily find a language/util in the right navigation menu:

image

docs/libs.md
hagenw: I removed " Extension" from "Sphinx Extension" and also " LaTeX" from "Canvas LaTeX".

docs/libs.md
hagenw: I just deleted the empty line at the bottom of the file, or should this also be kept?

edemaine (approved):
Looks great, thanks!

PR #3846, opened 2023-08-07
feat: copied d.ts types from DefinitelyTyped

### LINKED ISSUE #3531: Distribute typescript definitions
When using `katex` in typescript, one has to use the `@types/katex` package to get some type interference. However, these type definitions are quite minimal.

Provide full typescript definitions automatically.

Using `@types/katex`.

**Additional context:**
https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html

kelvinsjk:

Using KaTeX in TypeScript meant having to install the types from @types/katex

The types are now bundled together

Closes #3531 

Notes: I couldn't get the types to work correctly with the `"exports"` section in the `package.json` file, so I didn't include the types for the auto-render extension that are on DefinitelyTyped. However, have tested locally that the `"types"` key in the `package.json` file works as intended and provides types to my TypeScript imports of the main KaTeX library

### DIFF
--- a/types/katex.d.ts
+++ b/types/katex.d.ts
@@ -1,0 +1,162 @@
+// Type definitions for KaTeX 0.16
+// Project: https://katex.org/
+// Definitions by: Michael Randolph <https://github.com/mrand01>
+//                 Kevin Nguyen <https://github.com/knguyen0125>
+//                 bLue <https://github.com/dreamerblue>
+//                 Sebastian Weigand <https://github.com/s-weigand>
+//                 sapphi-red <https://github.com/sapphi-red>
+//                 Stefaans <https://github.com/Stefaans>
+// Definitions copied from: https://github.com/DefinitelyTyped/DefinitelyTyped
+// on 2023-08-07 by Kelvin Soh <https://github.com/kelvinsjk>
+
+export interface TrustContext {
+	command: string;
+	url: string;
+	protocol: string;
+}
+
+/** Documentation: https://katex.org/docs/options.html */
+export interface KatexOptions {
+	/**
+	 * If `true`, math will be rendered in display mode
+	 * (math in display style and center math on page)
+	 *
+	 * If `false`, math will be rendered in inline mode
+	 * @default false
+	 */
+	displayMode?: boolean | undefined;
+	/**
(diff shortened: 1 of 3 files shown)


joshuawwright:
@edemaine Any chance this can be approved?

types/katex.d.ts
```diff
@@ -0,0 +1,162 @@
...
+	 * If set to Infinity, users can make elements and space
+	 * arbitrarily large
+	 *
+	 * @default Infinity
+	 */
+	maxSize?: number | undefined;
+	/**
+	 * Limit the number of macro expansions to specified number
+	 *
+	 * If set to `Infinity`, marco expander will try to fully expand
```
edemaine, line 103:
```suggestion
	 * If set to `Infinity`, macro expander will try to fully expand
```

types/katex.d.ts
```diff
@@ -0,0 +1,162 @@
...
+	 * as in LaTex
+	 *
+	 * @default 1000
+	 */
+	maxExpand?: number | undefined;
+	/**
+	 * If `false` or `"ignore"`, allow features that make
+	 * writing in LaTex convenient but not supported by LaTex
```
edemaine, line 111:
```suggestion
	 * writing in LaTeX convenient but not supported by LaTeX
```

This clear typo makes me wonder whether we should instead copy descriptions from the official documentation, which are likely more precise.

edemaine (commented):
Sorry for the long delay. Have you tested this in both ESM and CJS modes? I intend to but it would be good for both of us to check. I thought getting both to work would require more effort than what I see here.

StefaniOSApps:
can we merge this?
@edemaine

types/katex.d.ts
StefaniOSApps: please apply this changes @kelvinsjk

edemaine:
@StefaniOSApps As in my original review, this needs testing before it gets merged. I don't think this has happened yet. If you are able to test, please report back.

kelvinsjk:
Sorry for taking forever. I've taken another stab at this in #3992 (opening a new PR since it's a rewrite from scratch)

PR #3854, opened 2023-09-01
hide character metrics console warnings on production - fixes #3853

### LINKED ISSUE #3853: Don't render character metrics console warnings on production
Currently the character metrics warnings appears on all `NODE_ENV` environments.

I would like to hide them in `NODE_ENV`: `production`

Example:
```
No character metrics for '€' in style 'Main-Regular' and mode 'text'
```

borriglione:

It logs console warnings if no metric was found for the character in the current font and mode.

It logs console warnings only in non-`production` `NODE_ENV`

 Fixes #3853


edemaine (commented):
Sorry for the delay in reviewing this. In addition to the style issues now highlighted below, there seems to a problem: `process.env.NODE_ENV` is getting evaluated and replaced at build time instead of at runtime.  I see the following in `dist/katex.js` after running `yarn build`:

```js
    const nodeEnv = typeof process !== 'undefined' && "production";
```

Presumably this is a feature of our build process (Webpack), and there's a way around it. Are you able to explore workarounds?

kalvdans:
I think we can get rid of the warning completely, regardless of environment.

edemaine:
@kalvdans The warning serves a purpose: it tells you that the height of the character is unknown, so there's a decent chance it will be placed incorrectly. (A good example is [within a `\sqrt`](https://github.com/KaTeX/KaTeX/issues/3718).)

kalvdans:
> @kalvdans The warning serves a purpose: it tells you that the height of the character is unknown, so there's a decent chance it will be placed incorrectly. (A good example is [within a `\sqrt`](https://github.com/KaTeX/KaTeX/issues/3718).)

Yes but what should the user (document writer) do about it? 

If the message is aimed for font developers, a full list of missing metrics is more useful.

PR #3859, opened 2023-09-17
specify `position: relative` for `.katex`

mikeguo99:

The class `.katex-mathml` inside `.katex` has specified `position: absolute`, but since `.katex` has not specified `position: relative`, `.katex-mathml` will keep looking up layer by layer, potentially causing layout chaos.

After specifying `position: relative` for `.katex`, the position of `.katex-mathml` has stabilized.

### DIFF
--- a/src/katex.less
+++ b/src/katex.less
@@ -8,6 +8,8 @@
 @version: "";
 
 .katex {
+    position: relative;
+
     font: normal 1.21em KaTeX_Main, Times New Roman, serif;
     line-height: 1.2;
 


edemaine:
Could you clarify in what scenario the layout of `.katex-mathml` matters, given that it's clipped to be invisible? Do you have an example of resulting "layout chaos"?

moesoha:
@edemaine If using KaTeX within scrollable content on a page, it will break the outer height.

```html
<body>
<div style="max-height: 80vh; overflow: auto; border: 1px solid #000;">
	<div style="height: 200vh;"></div>
	<span class="katex"><span class="katex-mathml"></span></span>
</div>
</body>
```

image

shinexyt:
Same issue as @moesoha

enzo2:
Also encountered. It does cause layout chaos and break my outer height, and adding .katex {position: relative;} fixes it.

will961123:
I'm having the same issue and hope it can be merged as soon as possible.

edemaine:
Replaced by #4170

PR #3863, opened 2023-09-27
feat: add KaTeX options menu for the test server

### LINKED ISSUE #1225: GUI for test server (main.js) settings
The test server (`npm start`) supports lots of settings by manually modifying the URL. It would be nice to have an actual GUI with checkboxes for boolean options (e.g. displayMode) and text boxes for others (e.g. adding macros). I think it would also make sense to always keep the URL in and with the current state, instead of requiring a press of a Permalink button.

nnethaji:
- modify settings manually by editing the URL. 
- lack of GUI support for both boolean options and options that need text input

- show KaTeX options menu for the test server below the permalink button

Closes #1225


static/index.html
```diff
@@ -1,17 +1,75 @@
 <!DOCTYPE html>
 <html>
-    <head>
-        <meta charset="UTF-8">
-        <title>KaTeX Test</title>
-        <script defer src="/main.js" type="text/javascript"></script>
-    </head>
-    <body>
-        <textarea id="input" rows="5">
+<!-- referenced from issue #1615 -->
```
edemaine, line 3: It would be better to refer to `website/pages/index.html` in this comment. And maybe it should move down 

You also need to consistently use either indentation of 4 (as before in this file) or 2 (as in `website/pages/index.html`), and `<head>` and `<body>` should be indented within `<html>`.

static/main.js
```diff
@@ -85,15 +85,80 @@ function init() {
...
 
+    // This sets default values
+    // Initialize an empty array to store the options as strings
+    const optionElements = [];
+
+    // Init array with ids of each option
+    const optionIds = ["displayMode", "leqno", "fleqn", "throwOnError",
+        "errorColor", "strict", "output", "trust", "macros"];
+
+    // Loop through each option ID
```
edemaine, line 96: This and some other comments are redundant with the code and should be removed.

static/main.js
```diff
@@ -85,15 +85,80 @@ function init() {
...
+        // Add the processed element to the optionElements array
+        optionElements.push(element);
+    }
+
+
     reprocess();
 
+    function changeOption() {
+        // Init empty object options
+        const options = {};
```
edemaine, line 127: I don't see why you need to make a new `options` object, instead of modifying the original one. I think the latter approach would be more robust, in case `options` has some hidden options that are not reflected by the UI.

static/main.js
```diff
@@ -14,7 +14,7 @@ function init() {
     input.addEventListener("input", reprocess, false);
     permalink.addEventListener("click", setSearch);
 
-    const options = {displayMode: true, throwOnError: true, trust: true};
+    let options = {displayMode: true, throwOnError: true, trust: true};
```
edemaine, line 17: This can change back to `const` if you implement the comment below about modifying `options` in-place.

static/main.js
```diff
@@ -85,15 +85,80 @@ function init() {
...
+                // parse the JSON input entered into the macro
+                val = element.value ? JSON.parse(element.value) : undefined;
+            } else {
+                // For rest of the elements, set val to the value of the element
+                val = element.value;
+            }
+
+            // Set id of each element in the options object to its current val
+            options[element.id] = val;
+        }
```
edemaine, line 148: You also need to modify `query`, so that the permalink changes to reflect the new options. Currently the permalink doesn't update. This is the main show-stopper for this PR.

edemaine (changes requested):
Thanks for working on this, and sorry for the very slow review. The UI looks good but there are a couple of style and functionality issues that need fixing.

edemaine:
Replaced by #4072. Thanks for getting this started!

PR #3896, opened 2023-11-04
doc(users): add bearbei

Edan0615:
Add bearbei to user list

### DIFF
--- a/website/siteConfig.js
+++ b/website/siteConfig.js
@@ -170,6 +170,11 @@ const users = [
         image: 'https://github.com/zzllrr/mather/raw/master/img/logo.jpg',
         infoLink: 'https://github.com/zzllrr/mather',
     },
+    {
+        caption: 'BearBei 貝貝',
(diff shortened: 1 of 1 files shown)


edemaine (changes requested):
Could you please move this before CindyJS so that the list is alphabetical (except for Khan Academy in front)?

Also, if there's a smaller logo file to link to, that might be preferable. (The current one is over 200kB.)

Edan0615:
> Could you please move this before CindyJS so that the list is alphabetical (except for Khan Academy in front)?
> 
> Also, if there's a smaller logo file to link to, that might be preferable. (The current one is over 200kB.)

Sure no problem, working on it. Thank you.

Edan0615:
> Could you please move this before CindyJS so that the list is alphabetical (except for Khan Academy in front)?
> 
> Also, if there's a smaller logo file to link to, that might be preferable. (The current one is over 200kB.)

Hi I just updated it with another pull request, thank you.

PR #3897, opened 2023-11-04
docs(users): add bearbei

Edan0615:
Update Bearbei to the list and update it in the order and with a smaller picture.


edemaine (approved):
Thank you!

PR #3903, opened 2023-11-12
Fix parising of `\,'`

### LINKED ISSUE #3856: Regression when rendering `$a\,'`
> Before reporting a bug

**Describe the bug:**
Using version prior to `0.15.0` rendering latex like `$a\,'` (`'` after `\,`) works just fine (which is consistent with pdflatex), but since `0.15.0` and in `0.16.x` versions it throws `ParseError: KaTeX parse error: Got group of unknown type: 'internal'` error.

Note that using anything other than `'` doesn't trigger the bug, so things like `a\,b` work just fine

**To Reproduce:**
```html
<!DOCTYPE html>
<html>
  <head>
    <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/katex@0.15.0/dist/katex.min.css" crossorigin="anonymous">
    <script src="https://cdn.jsdelivr.net/npm/katex@0.15.0/dist/katex.min.js" crossorigin="anonymous"></script>
  </head>
  <body>
    <script>
      katex.render("a\\,'", window.document.body, {
        throwOnError: false
      });
    </script>
  </body>
</html>
```
If you change `0.15.0` to `0.14.1` in the snippet above it is rendered correctly as expected.

**Expected behavior:**
fragment gets rendered

**Screenshots:**

 - KaTeX Version: 0.15.0 and any later
 - Device: Desktop
 - OS: NixOS
 - Browser: Chrome
 - Version: 116.0.5845.110

**Additional context:**
Reported initially [here](https://github.com/logseq/logseq/issues/10153)

t4ccer:
Closes #3856

### DIFF
--- a/src/macros.js
+++ b/src/macros.js
@@ -529,7 +529,7 @@ defineMacro("\\DOTSX", "\\relax");
 // Spacing, based on amsmath.sty's override of LaTeX defaults
 // \DeclareRobustCommand{\tmspace}[3]{%
 //   \ifmmode\mskip#1#2\else\kern#1#3\fi\relax}
-defineMacro("\\tmspace", "\\TextOrMath{\\kern#1#3}{\\mskip#1#2}\\relax");
+defineMacro("\\tmspace", "\\TextOrMath{\\kern#1#3\\relax}{\\mskip#1#2}");
 // \renewcommand{\,}{\tmspace+\thinmuskip{.1667em}}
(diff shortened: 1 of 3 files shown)


edemaine (commented):
This is not the correct solution to this problem. The `\relax` is needed to terminate the length, so should stay where it is. The root problem is that `\relax` cannot be followed by sub/superscripts, whereas it is allowed in LaTeX.

PR #3940, opened 2024-03-21
fix(css): fefault to font-display: swap; closes #2242

### LINKED ISSUE #2242: Use font-display: swap to prevent invisible text on load
Google PageSpeed Insights suggests that I should "ensure text remains visible during webfont load".

For KaTeX this would simply involve including [`font-display: swap;`][mdn-font-face] in all of the `@font-face` entries in `katex.css`.

I think this a good idea, but how would we then cater for people who *don't* want to temporarily show a fallback font?

[mdn-font-face]: https://developer.mozilla.org/en-US/docs/Web/CSS/@font-face/font-display#Values

FreyaHolmer:
Google fonts seem to default to swap, and mixing them with katex makes it so that katex is invisible briefly on page load, unlike google fonts which make use of fallbacks while loading. This would unify their behavior.

Ideally there'd be an option for this, and a ?display=swap in the cdn url (like google fonts: https://developers.google.com/fonts/docs/css2#use_font-display), but I don't know if jsdelivr supports that!

Katex fonts would be invisible until loaded

Katex fonts will now render a fallback until fonts have been loaded

BREAKING CHANGE: changes font loading behavior to font-display: swap,
which means equations will render a fallback before loaded
instead of being invisible

Closes #2242

### DIFF
--- a/src/fonts.less
+++ b/src/fonts.less
@@ -2,6 +2,7 @@
 @use-ttf: true;
 @use-woff: true;
 @use-woff2: true;
+@font-display: swap;
 
 .use-woff2(@family, @family-suffix) when (@use-woff2 = true) {
     src+: url('@{font-folder}/KaTeX_@{family}-@{family-suffix}.woff2') format('woff2')
@@ -37,6 +38,7 @@
         .use-ttf(@family, @suffix);
         font-weight: @weight;
         font-style: @style;
+        font-display: @font-display;
     }
 }
 


yao-genius:
Why hasn't this fix been approved yet? I had the same problem.

edemaine:
This PR needs to be ported from less to scss, though I suspect that would be easy. There's also a question of what's the best default.

Actually, we could offer multiple `.css` builds with different settings. I think `block` and `swap` are the two natural choices here, depending on whether you want to flash empty content (current behavior I guess) or flash incorrectly styled content.

edemaine (approved):
Thanks for submitting this, and apologies for the long delay.

I modified the PR to work as follows:
* `katex.css` and `katex.min.css` use `font-display: block`, roughly the same as before. I believe this is still a useful configuration to [avoid "flash of unstyled text" (FOUT), but instead have "flash of invisible content" (FOIT)](https://css-tricks.com/fout-foit-foft/).
  * In particular, I hesitate to modify the default here, because so many people use KaTeX and this might be a significant deviation from expected behavior.
* `katex-swap.css` and `katex-swap.min.css` use `font-display: swap` to enable "flash of unstyled text" (FOIT) instead of "flash of invisible content" (FOUT)
* Variable `$font-display` is still overridable (now via SCSS instead of Less)

TLDR: if you want `font-display: swap`, you use `katex-swap.css` instead of `katex.css`. But default remains roughly the same.

I'd appreciate any feedback on this approach (from original poster or anyone else interested in this feature) before merging.

grigoriy-reshetniak:
Hello @edemaine. 
I've discovered this PR while browsing open ones and I feel like it's a shame it hadn't been merged.  Looking from the OP's perspective:

> Ideally there'd be an option for this

I think since your changes actually provided that option this PR deserves to be merged.

edemaine:
Thanks Grigoriy! That's the feedback/confirmation I was looking for. I'll merge it.

PR #3949, opened 2024-04-25
feat: \boxslash support

FernandoChu:
`\boxslash` not supported.

`\boxslash` is supported.

### DIFF
--- a/docs/support_table.md
+++ b/docs/support_table.md
@@ -192,6 +192,7 @@ $\gdef\VERT{|}$
 |\boxminus|$\boxminus$||
 |\boxplus|$\boxplus$||
 |\boxtimes|$\boxtimes$||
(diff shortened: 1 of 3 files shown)


edemaine:
I get the warning `No character metrics for '⧄' in style 'AMS-Regular' and mode 'math'` when trying your branch (via the Netlify link above). This says that the font doesn't have metrics for this symbol. I'm guessing that it means that the font doesn't actually have this character, sadly. (Did you check? https://opentype.js.org/ is a good tool for inspect the font.) Adding it to the font would be required to accept this, but unfortunately this process is a bit involved.

FernandoChu:
Thanks for the quick answer! Honestly I don't know what's going on so I'll just close this PR and make a macro :)

PR #3954, opened 2024-06-05
chore: use tagged literals in tests

grigoriy-reshetniak:
https://github.com/KaTeX/KaTeX/pull/3902 looks abandoned, so I've decided to try to finish it.

`expect("\\mathinner{\\langle{\\psi}\\rangle}").toParse();`

``expect`\mathinner{\langle{\psi}\rangle}`.toParse();``

Fixes https://github.com/KaTeX/KaTeX/issues/1717

### DIFF
--- a/test/katex-spec.js
+++ b/test/katex-spec.js
@@ -812,7 +812,7 @@ describe("A text parser", function() {
 describe("A texvc builder", function() {
     it("should not fail", function() {
         expect`\lang\N\darr\R\dArr\Z\Darr\alef\rang`.toBuild();
-        expect("\\alefsym\\uarr\\Alpha\\uArr\\Beta\\Uarr\\Chi").toBuild();
+        expect`\alefsym\uarr\Alpha\uArr\Beta\Uarr\Chi`.toBuild();
         expect`\clubs\diamonds\hearts\spades\cnums\Complex`.toBuild();
         expect`\Dagger\empty\harr\Epsilon\hArr\Eta\Harr\exist`.toBuild();
         expect`\image\larr\infin\lArr\Iota\Larr\isin\Kappa`.toBuild();
@@ -2211,7 +2211,7 @@ describe("An accent parser", function() {
         expect`\vec{x^2}`.toParse();
         expect`\vec{x}^2`.toParse();
         expect`\vec x`.toParse();
-        expect("\\underbar{X}").toParse();
+        expect`\underbar{X}`.toParse();
     });
 
     it("should produce accents", function() {
@@ -2299,20 +2299,20 @@ describe("A stretchy MathML builder", function() {
 
 describe("An under-accent parser", function() {
     it("should not fail", function() {
-        expect("\\underrightarrow{x}").toParse();
-        expect("\\underrightarrow{x^2}").toParse();
-        expect("\\underrightarrow{x}^2").toParse();
-        expect("\\underrightarrow x").toParse();
+        expect`\underrightarrow{x}`.toParse();
+        expect`\underrightarrow{x^2}`.toParse();
+        expect`\underrightarrow{x}^2`.toParse();
+        expect`\underrightarrow x`.toParse();
     });
 
     it("should produce accentUnder", function() {
-        const parse = getParsed("\\underrightarrow x")[0];
+        const parse = getParsed`\underrightarrow x`[0];
 
         expect(parse.type).toEqual("accentUnder");
     });
 
     it("should be grouped more tightly than supsubs", function() {
-        const parse = getParsed("\\underrightarrow x^2")[0];
+        const parse = getParsed`\underrightarrow x^2`[0];
 
         expect(parse.type).toEqual("supsub");
     });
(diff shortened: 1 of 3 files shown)


grigoriy-reshetniak:
@edemaine could you please review and start CI job? Tests are passing locally.
image

test/katex-spec.js
```diff
@@ -1671,42 +1670,42 @@ describe("A font parser", function() {
 
 describe("A \\pmb builder", function() {
     it("should not fail", function() {
-        expect("\\pmb{\\mu}").toBuild();
-        expect("\\pmb{=}").toBuild();
-        expect("\\pmb{+}").toBuild();
-        expect("\\pmb{\\frac{x^2}{x_1}}").toBuild();
-        expect("\\pmb{}").toBuild();
-        expect("\\def\\x{1}\\pmb{\\x\\def\\x{2}}").toParseLike("\\pmb{1}");
+        expect`\pmb{\\mu}`.toBuild();
```
edemaine, line 1673:
```suggestion
        expect`\pmb{\mu}`.toBuild();
```

test/katex-spec.js
```diff
@@ -1671,42 +1670,42 @@ describe("A font parser", function() {
...
+        expect`\pmb{=}`.toBuild();
+        expect`\pmb{+}`.toBuild();
+        expect`\pmb{\\frac{x^2}{x_1}}`.toBuild();
```
edemaine, line 1676:
```suggestion
        expect`\pmb{\frac{x^2}{x_1}}`.toBuild();
```

test/katex-spec.js
```diff
@@ -1523,18 +1522,18 @@ describe("A TeX-compliant parser", function() {
...
+        expect`\iiint_i^n`.toBuild();
+        expect`\int\nolimits_i^n`.toBuild();
+        expect`\iint\nolimits_i^n`.toBuild();
+        expect`\iiint\nolimits_i^n`.toBuild();
+        expect`\oint_i^n`.toBuild();
+        expect`\oiint_i^n`.toBuild();
+        expect`\oiiint_i^n`.toBuild();
+        expect`\oint\nolimits_i^n`.toBuild();
+        expect`\oiint\nolimits_i^n`.toBuild();
+        expect`\oiiint\nolimits_i^n`.toBuild();
```
edemaine, line 1536: A lot of these tests were buggy before with `\n`! Nice to fix them.

test/katex-spec.js
```diff
@@ -3410,27 +3409,27 @@ describe("A macro expander", function() {
...
-        expect("\\char'").not.toParse();
-        expect('\\char"').not.toParse();
-        expect("\\char'a").not.toParse();
-        expect('\\char"g').not.toParse();
-        expect('\\char"g').not.toParse();
+        expect("\\char`\\%").toParseLike`\char37`;
+        expect("\\char`\\%").toParseLike`\char'45`;
+        expect("\\char`\\%").toParseLike`\char"25`;
+        expect`\char`.not.toParse();
+        expect`\char\``.not.toParse();
```
edemaine, line 3416: Does this convert to `"\\char`"` or `"\\char\\`"`? I'm not sure...

test/katex-spec.js
```diff
@@ -811,15 +811,15 @@ describe("A text parser", function() {
...
-        expect("\\alefsym\\uarr\\Alpha\\uArr\\Beta\\Uarr\\Chi").toBuild();
-        expect("\\clubs\\diamonds\\hearts\\spades\\cnums\\Complex").toBuild();
-        expect("\\Dagger\\empty\\harr\\Epsilon\\hArr\\Eta\\Harr\\exist").toBuild();
-        expect("\\image\\larr\\infin\\lArr\\Iota\\Larr\\isin\\Kappa").toBuild();
-        expect("\\Mu\\lrarr\\natnums\\lrArr\\Nu\\Lrarr\\Omicron").toBuild();
-        expect("\\real\\rarr\\plusmn\\rArr\\reals\\Rarr\\Reals\\Rho").toBuild();
-        expect("\\text{\\sect}\\sdot\\sub\\sube\\supe").toBuild();
-        expect("\\Tau\\thetasym\\weierp\\Zeta").toBuild();
+        expect`\lang\N\darr\R\dArr\Z\Darr\alef\rang`.toBuild();
+        expect`\alefsym\uarr\Alpha\uArr\Beta\Uarr\Chi`.toBuild();
```
edemaine, line 815: I thought `\u` didn't work in raw strings because it's a Unicode escape. See e.g.

https://github.com/KaTeX/KaTeX/blob/bedce93260cc413abb7730a5693dc27222297178/test/katex-spec.js#L2919

Is that incorrect? (It might be; I'm not sure.)

grigoriy-reshetniak: No, you right. I'm changing it to double backslash 
image

edemaine: This isn't the right experiment though. I think ``String.raw`...` `` is closer, but I couldn't immediately find how we're providing template strings creation functions.

grigoriy-reshetniak:
I have fixed 2 issues with double backslashes and reverted every test that includes `\u` or backtick. While backtick test passes,  the string that expect resolves to has changed. Some test with `\u`, like `An under-accent parser` work fine with backticks and single backslash, but I'm not sure why  and I better move it to separate PR after I figure it out.

edemaine:
I found where we handle tagged literals:

https://github.com/KaTeX/KaTeX/blob/bedce93260cc413abb7730a5693dc27222297178/test/helpers.js#L17-L20

This gets called e.g. in `getBuilt`. So I think the correct test is:

```js
const r = x => x != null && x.hasOwnProperty('raw') ? x.raw[0] : x;
r`\url`
```

which correctly generates `"\\url"`. So I think all the `\u` tests can in fact be restored to use backticks. Sorry for the runaround!

Maybe you can update these comments too:

https://github.com/KaTeX/KaTeX/blob/bedce93260cc413abb7730a5693dc27222297178/test/katex-spec.js#L2919

https://github.com/KaTeX/KaTeX/blob/bedce93260cc413abb7730a5693dc27222297178/docs/api.md#L16

I think what is still problematic is `${` (because of interpolations, and `\$` won't fix it) and `` ` `` that isn't preceded by a backslash (again `` \` `` would produce something different than `` ` ``). But otherwise things template strings should be OK: `\u` and `\x` do not seem to pose any issue *when using `String.raw` (or `r` as in our tests)*.

In the future, can you add commits instead of rebasing? Otherwise I can't really review the new changes. We squash commit when accepting the PR, so the commit history doesn't matter except for review.

edemaine (approved):
Thank you!

PR #3963, opened 2024-06-30
feat: add \emph

### LINKED ISSUE #3566: add \emph
`\text{a}` and `\textit{a}` compile find with KaTeX, `\emph{a}` doesn't. Perhaps that'd be a worthwhile addition.

grigoriy-reshetniak:
`\emph` wasn't supported

`\emph` is supported

image

Implements default behavior of making text italic if it's not and making text upright if it's italic. Please let me know ff more complex rules are required.

Closes #3566

### DIFF
--- a/src/functions/text.js
+++ b/src/functions/text.js
@@ -19,6 +19,7 @@ const textFontWeights = {
 const textFontShapes = {
     "\\textit": "textit",
     "\\textup": "textup",
+    "\\emph": "textit",
 };
 
 const optionsWithFont = (group, options) => {
@@ -30,9 +31,11 @@ const optionsWithFont = (group, options) => {
         return options.withTextFontFamily(textFontFamilies[font]);
     } else if (textFontWeights[font]) {
         return options.withTextFontWeight(textFontWeights[font]);
-    } else {
-        return options.withTextFontShape(textFontShapes[font]);
+    } else if (options.fontShape === "textit" && font === "\\emph") {
+        return options.withTextFontShape("textup");
     }
+
+    return options.withTextFontShape(textFontShapes[font]);
 };
 
 defineFunction({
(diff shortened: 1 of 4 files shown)


src/functions/text.js
```diff
@@ -19,6 +19,7 @@ const textFontWeights = {
 const textFontShapes = {
     "\\textit": "textit",
     "\\textup": "textup",
+    "\\emph": "textit",
```
edemaine, line 22: I think it would be better to avoid including this as a shape (see below).

src/functions/text.js
```diff
@@ -30,9 +31,11 @@ const optionsWithFont = (group, options) => {
         return options.withTextFontFamily(textFontFamilies[font]);
     } else if (textFontWeights[font]) {
         return options.withTextFontWeight(textFontWeights[font]);
-    } else {
-        return options.withTextFontShape(textFontShapes[font]);
+    } else if (options.fontShape === "textit" && font === "\\emph") {
```
edemaine, line 34: I'd suggest changing this to `if (font === "\emph")` and using `options.withTextFontShape` (enforcing text mode) with `"textit"` or `"textup"` according to `options.fontShape`.

edemaine (changes requested):
I didn't realize that `\emph` works in math mode.  Your implementation seems to exactly match LaTeX behavior:

```latex
x+\emph{foo \emph{bar}}+\textit{foo \emph{bar}}
```

image

LaTeX:
image

In addition to the implementation comments below, can you add some tests? Something like ``expect`\emph{foo \emph{bar}}`.toParseLike`\textit{foo \textup{bar}}` ``.

grigoriy-reshetniak:
My bad, I should've added test from the beginning

edemaine (approved):
Looks good! I re-organized the Fonts docs table to include `\emph`. Could you double check it looks good before I merge?

grigoriy-reshetniak (commented):
Looks good to me.

PR #3970, opened 2024-08-05
Fix svg if katex in bootstrap's modal in katex.scss

hotamago:
add z-index: 1;

image

image

### DIFF
--- a/src/styles/katex.scss
+++ b/src/styles/katex.scss
@@ -442,7 +442,7 @@ $version: "" !default;
     svg {
         display: block;
         position: absolute; // absolute relative to parent
-        z-index: 1;
+        z-index: auto;
         width: 100%;
         height: inherit;
 
--- a/src/styles/katex.scss
+++ b/src/styles/katex.scss
@@ -442,6 +442,7 @@ $version: "" !default;
     svg {
         display: block;
         position: absolute; // absolute relative to parent
+        z-index: 1;
         width: 100%;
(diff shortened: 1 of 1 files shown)


edemaine:
Could you give more details about when you run into this issue? Does Bootstrap have a rule about `svg` specifically? I don't see why it would need a z-index, unless it has something to do with `position: absolute`.

Your description (`z-index: 1`) also does not match your PR (`z-index: auto`). The latter seems helpful only if there's some other `z-index` rule we're overriding...

edemaine:
I also wonder if #3859 would fix your issue. Could you try adding `position: relative` to `.katex` to see whether that also fixes your problem?

hotamago:
> I also wonder if #3859 would fix your issue. Could you try adding `position: relative` to `.katex` to see whether that also fixes your problem?

Katex uses `position: absolute`, which can cause it to be obscured by other elements. This is why the square root symbol (sqrt) is not visible in my case.

I've tried using `z-index: auto` instead of `z-index: 1` because the latter failed the `Screenshotter / screenshotter (firefox)` test. `z-index: auto` seems to work well for my specific situation.

I also experimented with `position: relative` but it resulted in incorrect positioning of the radical sign.

grigoriy-reshetniak:
@hotamago could you please confirm that your issue is resolved after https://github.com/KaTeX/KaTeX/releases/tag/v0.16.40 was released?

PR #3992, opened 2024-12-01
feat: ESM exports, TypeScript types

### LINKED ISSUE #3531: Distribute typescript definitions

kelvinsjk:

Using KaTeX in TypeScript meant having to install the types from @types/katex

The types are now bundled together.

Closes #3531

As suggested in #3846, I have rewritten the dts from scratch, referring to the official documentation rather than simply copying what was done in the DefinitelyTyped repo.

I have also split the `d.ts` files into the esm and cjs counterparts, following [the lint rule on publint](https://publint.dev/rules#export_types_invalid_format).

 I have done my own testing for those two modes at https://github.com/kelvinsjk/katex-types. It works for my system and setup, so definitely looking to see how it fares for others.


StefaniOSApps:
please take a look @edemaine

types/katex.d.ts
```diff
@@ -0,0 +1,212 @@
+// Adapted from
```
edemaine, line 1: Do we really need a separate type definition file for CJS? I think `types/katex.d.mts` should work for CJS as well, possibly with slight tweaking (but maybe not even?). I'd really like to avoid having two copies of the type definitions, because it makes it easy to update definitions inconsistently.

edemaine (changes requested):
Thanks for picking this back up! I have comments for removing use of `any`, and removing the second file copy `types/katex.d.ts` (or vice versa).

types/katex.d.mts
```diff
@@ -0,0 +1,218 @@
...
+     * @default [1000]
+     *
+     * Limit the number of macro expansions to the specified number, to prevent
+     * e.g. infinite macro loops. `\edef` expansion counts all expanded tokens.
+     * If set to `Infinity`, the macro expander will try to fully expand as in
+     * LaTeX.
+     */
+    maxExpand?: number;
+    /**
+     * @type {boolean|string|(errorCode: any, errorMsg: any, token: any)=>any}
```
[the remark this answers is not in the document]
kelvinsjk: Will remove all the `@type`s in an upcoming commit

types/katex.d.ts
kelvinsjk: Have updated `types/katex.d.mts` to import type definitions from `types/katex.d.ts`

edemaine (approved):
Thanks for implementing the changes!  I made a couple more:

1. I modified the KaTeX source to use ESM non-default exports. So now you can `import { renderToString } from 'katex'`, whereas before you had to use `import katex from 'katex'; const { renderToString } = katex`. This is what your types indicated, and is natural behavior. The default export remains the same.
2. I merged `katex.d.ts` and `katex.d.mts` into one file. The exports are identical on both sides, so this seems natural. According to my testing, autocomplete works great in both CJS and ESM.
3. I added a root `tsconfig.json` so that we can declare the used libs (for `Record` and `HTMLElement`). I'm not totally sure this is necessary, but it certainly makes VSCode happier when editing `katex.d.ts`.

Let me know if you approve of these. If so, I think we're done!

kelvinsjk:
> Thanks for implementing the changes! I made a couple more:
> 
>     1. I modified the KaTeX source to use ESM non-default exports. So now you can `import { renderToString } from 'katex'`, whereas before you had to use `import katex from 'katex'; const { renderToString } = katex`. This is what your types indicated, and is natural behavior. The default export remains the same.
> 
>     2. I merged `katex.d.ts` and `katex.d.mts` into one file. The exports are identical on both sides, so this seems natural. According to my testing, autocomplete works great in both CJS and ESM.
> 
>     3. I added a root `tsconfig.json` so that we can declare the used libs (for `Record` and `HTMLElement`). I'm not totally sure this is necessary, but it certainly makes VSCode happier when editing `katex.d.ts`.
> 
> 
> Let me know if you approve of these. If so, I think we're done!

Yes, have tested the new changes on my own machine and it works great!

Mister-Hope:
Appreciate for the pr, but the katex types are not accurate enough, especially for the `token: any` defined in logger, where `[1,2]` can also be an `object`

edemaine:
Are you referring to `strict`? Would you like to send in a PR that improves the typing?

Mister-Hope:
https://github.com/KaTeX/KaTeX/pull/4009

Mister-Hope:
Hi, it's been a while since I last dig in to it.

I believe `"type": "module"` in package.json should be added in this PR to ensure it's working in all ESM env, is there any reason that is blocking it?

edemaine:
@Mister-Hope That shouldn't be necessary. This PR was already merged, and in our testing works in an ESM project.

PR #3998, opened 2024-12-09
feat: italic sans-serif in math mode via `\mathsfit` command

### LINKED ISSUE #3993: Support for Sans Serif Italic font in math mode
Currently, there is no way to enable Sans Serif and Italic while in math mode. 
[According to the documentation](https://katex.org/docs/supported#style-color-size-and-font), it is possible to stack commands in text mode like `\textsf{\textit{X}}`, but this is not possible in math mode `\mathsf{\mathit{X}}`. 

Would it be possible to implement stacking of family, weight and shape for math mode as well? Or would it be possible to add a `\mathsfit` command that combine Sans Serif and Italic?

olavhaugen:
There was no way to make math sans serif and italic at the same time.

Using the `\mathsfit` command, math will be both sans serif and italic 

Closes #3993


test/screenshotter/ss_data.yaml
```diff
@@ -253,6 +253,7 @@ MathNormal: \mathnormal{Ax2k\breve{a}\omega\Omega\imath+\KaTeX}
 MathOp: a\mathop+b\mathop:c\mathop{\delta}e\mathop{\textrm{and}}f\mathrel{\mathop{:}}=g\sin h
 MathRm: \mathrm{Ax2k\breve{a}\omega\Omega\imath+\KaTeX}
 MathSf: \mathsf{Ax2k\breve{a}\omega\Omega\imath+\KaTeX}
+MathSfit: \mathsfit{Ax2k\breve{a}\omega\Omega\imath+\KaTeX}
```
edemaine, line 256: This screenshot test is good, but needs corresponding image files. It looks like there's a bug in the CI that this isn't erroring or outputting the necessary files. I'll investigate.

docs/supported.md
```diff
@@ -606,7 +606,7 @@ For color definition, KaTeX color functions will accept the standard HTML [pred
 |$\mathit{Ab0}$ `\mathit{Ab0}`  |$\mathtt{Ab0}$ `\mathtt{Ab0}`  |$\mathcal{AB0}$ `\mathcal{AB0}`
 |$\textit{Ab0}$ `\textit{Ab0}`  |$\texttt{Ab0}$ `\texttt{Ab0}`  |$\cal AB0$ `\cal AB0`
 |$\it Ab0$ `\it Ab0`            |$\tt Ab0$ `\tt Ab0`            |$\mathscr{AB}$ `\mathscr{AB}`
-|$\emph{Ab0}$ `\emph{Ab0}`      |                               ||
+|$\emph{Ab0}$ `\emph{Ab0}`      |$\mathsfit{Ab0}$ `\mathsfit{Ab0}`||
```
edemaine, line 609: I think I'd rather see this in the third column just below `\sf`. (Sorry maintaining these tables in text is a pain.)

This example doesn't work on the Netlify preview, but I think that's because it's using the released CSS instead of the branch CSS.

edemaine (changes requested):
It seems that `\mathsfit` is in at least some packages and common definitions — see https://tex.stackexchange.com/questions/434391/how-do-i-get-italic-sans-serif-in-math-mode especially [mention of isomath](https://tex.stackexchange.com/a/584582/245104) — so this function name seems reasonable.

src/styles/katex.scss
```diff
@@ -152,6 +152,11 @@ $display-margin: 1em 0 !default;
         font-family: KaTeX_SansSerif;
     }
 
+    .mathsfit {
+        font-family: KaTeX_SansSerif;
+        font-style: italic;
+    }
+
```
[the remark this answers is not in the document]
olavhaugen: Added `.mathsfit` to the existing CSS rule

test/screenshotter/ss_data.yaml
olavhaugen: Renamed

edemaine (approved):

test/screenshotter/ss_data.yaml
edemaine: I was able to generate the screenshots by hand (running `yarn test:screenshots:update` with a browserstack key), so we're all set here!

PR #4009, opened 2024-12-20
fix(types): improve `strict` function type

Mister-Hope:


types/katex.d.ts
```diff
@@ -20,6 +20,43 @@ export type TrustContext =
...
+  errorCode:
+    | "unknownSymbol"
+    | "unicodeTextInMathMode"
+    | "mathVsTextUnits"
+    | "commentAtEnd"
+    | "htmlExtension"
+    | "newLineInDisplayMode",
+  errorMsg: string,
+  token: KatexToken,
+) => boolean | "error" | "warn" | "ignore" | undefined | null | void;
```
Mister-Hope, line 57: Actually I think `boolean | "error" | "warn" | "ignore" | void` is enough

types/katex.d.ts
```diff
@@ -20,6 +20,43 @@ export type TrustContext =
...
+  catcodes: KatexCatcodes;
+}
+
+interface KatexSourceLocation {
+  start: number;
+  end: number;
+  lexer: KatexLexerInterFace;
+}
+
+export interface KatexToken {
```
Mister-Hope, line 39: BTW, to safely fit all users even they are patching katex for they own usage, we can let katexToken extends `Record<string, unknown>`, so other properties hacked by users can be added without causing any type issues.

Requesting for a feedback here.

types/katex.d.ts
```diff
@@ -20,6 +20,43 @@ export type TrustContext =
     | { command: "\\htmlStyle", style: string }
     | { command: "\\htmlData", attributes: Record<string, string> }
 
+
+type KatexCatcodes = Record<string, number>;
```
edemaine, line 24: I don't think you need to prefix all the types and interfaces with `Katex`. And probably they should all be `export`ed too.

types/katex.d.ts
```diff
@@ -20,6 +20,43 @@ export type TrustContext =
...
+
+interface KatexLexerInterFace {
```
edemaine, line 26: I also don't think you need `Interface` in the name of this `Interface`. Matching the internal `Lexer` seems fine.

types/katex.d.ts
```diff
@@ -20,6 +20,43 @@ export type TrustContext =
...
+
+interface KatexSourceLocation {
+  start: number;
+  end: number;
+  lexer: KatexLexerInterFace;
+}
+
+export interface KatexToken {
+  text: string;
+  loc: KatexSourceLocation;
```
edemaine, line 41: According to `Token.js`, `loc` should have `| undefined` as well.

types/katex.d.ts
```diff
@@ -20,6 +20,43 @@ export type TrustContext =
...
+  noexpand: boolean | undefined;
+  treatAsRelax: boolean | undefined;
```
edemaine, line 43: Actually, in TypeScript notation, I think these should be
```ts
  noexpand?: boolean;
  treatAsRelax?: boolean;
```

types/katex.d.ts
edemaine: I don't think `| void` means what you think it means... I definitely wouldn't write that.

I agree that the docs don't allow `undefined | null` but the code explicitly says that it does. I could go either way on including them; probably `boolean` is enough. Or adding just `undefined` is maybe helpful.

types/katex.d.ts
```diff
@@ -20,6 +20,43 @@ export type TrustContext =
...
+}
+
+
+export type KatexLoggerFunction = (
```
edemaine, line 47: I think this should be called `StrictFunction`

types/katex.d.ts
edemaine: I don't think that's necessary; the user can always [merge interfaces](https://www.typescriptlang.org/docs/handbook/declaration-merging.html#merging-interfaces) to extend anything we provide. (You'd need to wrap it in `declare module "katex" { ... }` I believe.)

edemaine (changes requested):
Looks like a good start! Here are some suggested changes.

Mister-Hope:
All done

edemaine (approved):
Thanks!

PR #4026, opened 2025-01-29
chore: bump GitHub actions to their latest versions

deining:
This PR bumps GitHub actions to their latest versions, thus preventing warnings as seen e.g. [here](https://github.com/KaTeX/KaTeX/actions/runs/13015000604).
In order to keep workflow actions up-to-date automatically, a dependabot script was added.


edemaine (approved):
Thanks!

PR #4035, opened 2025-02-21
chore: spelling fixes

deining:
This PR fixes two typos I spotted in the project.

### DIFF
--- a/src/functions/supsub.js
+++ b/src/functions/supsub.js
@@ -192,7 +192,7 @@ defineFunctionBuilders({
             options);
     },
     mathmlBuilder(group, options) {
-        // Is the inner group a relevant horizonal brace?
+        // Is the inner group a relevant horizontal brace?
         let isBrace = false;
         let isOver;
         let isSup;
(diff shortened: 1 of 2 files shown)


ronkok (commented):
“diaresis” is not incorrect; it’s just different. “diaresis” is the American spelling. “diaeresis” is the British spelling.

deining:
> “diaresis” is not incorrect; it’s just different.
> “diaresis” is the American spelling.
> “diaeresis” is the British spelling.

I agree that there are both American and British spelling. But the American spelling is `dieresis`, not `diaresis`. At least that's what [wiktionary](https://en.wiktionary.org/wiki/dieresis) tells.

Let's have a look at the context: this is a comment to unicode point [U+0308 ◌̈](https://en.wikipedia.org/wiki/Diaeresis_(diacritic)) COMBINING DIAERESIS. My idea was to stick to the description and use `diaeresis` here.

edemaine (approved):

PR #4045, opened 2025-04-09
fix: \relax in base or exponent of super/subscript

### LINKED ISSUE #3878: Space before exponent subscript, superscript: Got group of unknown type: 'internal'
Putting an explicit horizontal space right before exponentiation `^` is valid in LaTeX
```latex
\[
a^t
a\,^t
a\:^t
a\>^t
a\;^t
a\quad^t
a\qquad^t
\]
```

screenshot_2023-10-18-145326

but doesn't work in KaTeX:

```
KaTeX parse error: Got group of unknown type: 'internal'
```

Same with `_`.

_Edit:_ The same error message occurs with

```
.\!''
```

### LINKED ISSUE #4044: Parse Error: "Got group of unknown type: 'internal'" with complex superscript/subscript expressions
> Before reporting a bug
- [ ] Check [common issues](https://katex.org/docs/issues.html).
- [ ] Check the bug is reproducible in [the demo](https://katex.org). If not, check KaTeX is up-to-date and installed correctly.
- [ ] Search for [existing issues](https://github.com/KaTeX/KaTeX/issues).

**Describe the bug:**
I encountered a parsing error in KaTeX when rendering certain mathematical expressions involving nested superscripts and subscripts. The error message is: ```KaTeX parse error: Got group of unknown type: 'internal'```.
Here are the problematic expressions:
```x^{y_z},\quad x^{(n)},\quad ^{16}O^{2-}```
This formula can be correctly parsed using MathJax.

Image

**To Reproduce:**

**Expected behavior:**

**Screenshots:**

 - KaTeX Version: [Can be checked via `katex.version`]
 - Device: [e.g. Desktop, iPhone 8]
 - OS: [e.g. Windows 10, iOS 11]
 - Browser: [e.g. Chrome, Safari]
 - Version: [e.g. 72]

**Additional context:**
Add any other context about the problem here. If you're using a library to use KaTeX, please include its name and setup.

ronkok:
Fixes #4044. Fixes #3878.


ronkok:
Looking at your code, I cannot tell if a `\relax` in the superscript will cause the same issue. It's worth checking that.  If the superscript is not a problem, then I think your code is a good idea.

edemaine:
> I cannot tell if a `\relax` in the superscript will cause the same issue. It's worth checking that.

Good idea! Both your code and my code failed at that, with this error:
`Got function '\relax' with no arguments as superscript`

This is generated by `parseGroup`, which is also how arguments to macros get parsed. I played with LaTeX a bit and discovered:
* `\foo \relax` gets treated like `\foo{}` (empty argument) — `\relax` is *not* skipped. So `parseGroup` needs to return the `\relax`.
* `x^\relax y` gets treated like `x^y` — `\relax` *is* skipped. So `handleSupSubscript` needs to detect and skip `\relax`.

The latest commit should fix this issue, and add some tests for the above. I also noticed that the text you wrote wasn't working correctly, because `\\` gets literalized here, so was really doing two backslashes, not a `\relax`.

ronkok:
It all looks good to me.

PR #4054, opened 2025-05-29
feat(website): add autofocus to demo-input textarea

tom-huntington:
Improve experience using website as LaTeX editor

I'm using your website as a TeX editor for creating LLM prompts. I have to move my hands of the keyboard to the mouse to focus the `textarea`.

`textarea` auto focuses


grigoriy-reshetniak:
Hello.
> I have to move my hands of the keyboard to the mouse to focus

Or you can use Tab button. Also, I believe auto focusing something while skipping main navigation links would be an anti-pattern.

PR #4058, opened 2025-06-07
feat: support optional args with \newcommand

### LINKED ISSUE #2228: Support OPTIONAL ARGUMENT in user defined macro by \newcommand
It seems that optional arguments are not supportes by user defined `\newcommand`.
Just see the following code.
```tex
\newcommand \foo [2] [bar] {{#1}^{#2}}
\foo {1}
\foo [baz]{2}
```
image

Would this optional argumant feature be implemented just like `\sqrt`macro? this is a predefined macro in `KaTeX` with optional argument support.

N/A

N/A

**Additional context:**
N/A

xplat:
A second optional argument to `\newcommand`, `\renewcommand`, or `\providecommand` causes parse failure.

A second optional argument to `\newcommand`, `\renewcommand`, or `\providecommand` causes the defined macro's first argument to become optional with the provided argument as default value.

Fixes #2228

### DIFF
--- a/src/MacroExpander.js
+++ b/src/MacroExpander.js
@@ -219,10 +219,31 @@ export default class MacroExpander implements MacroContextInterface {
     }
 
     /**
-     * Consume the specified number of (delimited) arguments from the token
-     * stream and return the resulting array of arguments.
+     * Consume the specified number of (delimited *or* defaulted) arguments
+     * from the token stream and return the resulting array of arguments.
      */
-    consumeArgs(numArgs: number, delimiters?: string[][]): Token[][] {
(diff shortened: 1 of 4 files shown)


edemaine (changes requested):
Sorry for the delay in reviewing this. This looks like a great feature addition! `defaultArgs` is a decent mechanism that allows a prefix of arguments to be optional. To support the future of `\NewDocumentCommand` (see e.g. [this post](https://www.reddit.com/r/LaTeX/comments/1igbj99/can_you_make_a_command_with_two_optional/)) perhaps we could make `defaultArgs` into an object that specifies which arguments are optional (have defaults)? In the case of `\newcommand` it would just be `{0: ...}`.

It also looks like you're using your `main` branch. I believe this prevents me from pushing to your branch. It would be better to use a different branch.

PR #4061, opened 2025-07-08
chore: remove polyfill for Array.prototype.includes

grigoriy-reshetniak:
I think of this MR as continuation of https://github.com/KaTeX/KaTeX/pull/1645. There was a custom function, but native `Array.includes()` is widely adopted now and based on config in .browserlistrc polyfill is not needed anymore.

`utils.contains()` was used to find if array contains an element.

Native `Array.includes()` is used for this purpose. **`Array.includes() `presented in the output artifact.**


grigoriy-reshetniak:
Hello @edemaine. I see your reaction on the first comment, but I'm not sure of my next steps. Is this PR in 'review pending' or you expect me to fix codecov check first? I'm asking because it's a newly introduced check that failed after merging `main` into `grigoriy-reshetniak:replace-contains-with-includes` and I have a doubts about it:

- In lists lines of `src/three.js` as missing but they were missing before.
- It lists `src/mathMLTree.js` as dropping test coverage, but that file is not even included in commit.
- It states that coverage for `src/utils.js` has been reduced by 0.32%, but I've only deleted lines from there.
- It is behind 40 commits behind the `main`.

I have a strong feeling that it will fail on majority of PR's that touching the existing code and I'm looking for your advise.

PR #4064, opened 2025-07-12
refactor(Style): replace default export with named

grigoriy-reshetniak:
This PR is done in assumption, that only some files require all for exported Style types and for most of them only subset they use would be sufficient, and unused instances wouldn't be imported, reducing bundle size. It is indeed **slightly** smaller, from 608.07 KB to 607.53 KB based on `yarn analyze` (parsed). 

I feel like maybe if this done at the scale it could reduce bundle more ~~significantly~~, but I'm interested in maintainers opinion about this, is it worth checking all default exports or do you prefer them over the named?

`export default` was used

`export const` is used now

### DIFF
--- a/src/Style.js
+++ b/src/Style.js
@@ -122,9 +122,7 @@ const cramp = [Dc, Dc, Tc, Tc, Sc, Sc, SSc, SSc];
 const text = [D, Dc, T, Tc, T, Tc, T, Tc];
 
 // We only export some of the styles.
-export default {
-    DISPLAY: (styles[D]: Style),
-    TEXT: (styles[T]: Style),
-    SCRIPT: (styles[S]: Style),
-    SCRIPTSCRIPT: (styles[SS]: Style),
-};
+export const DISPLAY = (styles[D]: Style);
+export const TEXT = (styles[T]: Style);
+export const SCRIPT = (styles[S]: Style);
+export const SCRIPTSCRIPT = (styles[SS]: Style);
--- a/src/buildHTML.js
+++ b/src/buildHTML.js
@@ -7,7 +7,7 @@
  */
 
 import ParseError from "./ParseError";
-import Style from "./Style";
+import {DISPLAY, TEXT, SCRIPT, SCRIPTSCRIPT} from "./Style";
 import buildCommon from "./buildCommon";
 import {Span, Anchor} from "./domTree";
 import utils from "./utils";
@@ -29,10 +29,10 @@ const binLeftCanceller = ["leftmost", "mbin", "mopen", "mrel", "mop", "mpunct"];
 const binRightCanceller = ["rightmost", "mrel", "mclose", "mpunct"];
 
 const styleMap = {
-    "display": Style.DISPLAY,
-    "text": Style.TEXT,
-    "script": Style.SCRIPT,
-    "scriptscript": Style.SCRIPTSCRIPT,
+    "display": DISPLAY,
+    "text": TEXT,
+    "script": SCRIPT,
+    "scriptscript": SCRIPTSCRIPT,
 };
 
 type Side = "left" | "right";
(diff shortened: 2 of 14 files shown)


edemaine:
Sorry for the delay on this. I generally like named exports over default exports. However, in this case, I kind of like how `Style` acts as a namespace, similar to TypeScript `enum`s. It's a bit sad to lose that... But I could go either way. What other contexts do you think this idea could be applied, and would it lead to further namespace degredation?

grigoriy-reshetniak:
I see that it could apply to almost every case with `export default {...}`. Let's take a look at the examples:

1. `mathMLTree.js` includes `TextNode` in [default export](https://github.com/KaTeX/KaTeX/blob/main/src/mathMLTree.js#L262) and [named one](https://github.com/KaTeX/KaTeX/blob/main/src/mathMLTree.js#L156). Then in `buildMathML.js` it was imported [as part of default](https://github.com/KaTeX/KaTeX/blob/main/src/buildMathML.js#L10) and separately [named](https://github.com/KaTeX/KaTeX/blob/main/src/buildMathML.js#L14). After that both `TextNode` and `mathMLTree.TextNode` are used and I believe that in that case we rather talking about namespace cleansing rather than degradation.
2. The opposite situation with `delimeter.js` [default export](https://github.com/KaTeX/KaTeX/blob/main/src/delimiter.js#L828). It is nicely encapsulated, but from 3 imports only 1 really benefits from grouping them: `genfrac.js` and `sqrt.js` both using only 1 function.

There's total 6  `export default {...}` besides one in `Style.js`. I would rather go case by case with them, as `mathMLTree.js` looks like it's easy to make a decision about it, but `buildCommon.js` doesn't seem so definitive, `fontMetricsData.js` should rather be untouched and `utils.js` would benefit from some attention in general prior to working with exports.

grigoriy-reshetniak:
Hello @edemaine. Is there any plans going forward with this?

edemaine:
@grigoriy-reshetniak Could you please refrain from automatically merging main into your PRs? I can do it myself when reviewing, and it leads to lots of email spam.

grigoriy-reshetniak:
Sorry, I was trying to keep branches up to date. Won't do it again.

edemaine:
> 1. `mathMLTree.js` includes `TextNode` in [default export](https://github.com/KaTeX/KaTeX/blob/main/src/mathMLTree.js?rgh-link-date=2025-10-09T18%3A21%3A37Z#L262) and [named one](https://github.com/KaTeX/KaTeX/blob/main/src/mathMLTree.js?rgh-link-date=2025-10-09T18%3A21%3A37Z#L156). Then in `buildMathML.js` it was imported [as part of default](https://github.com/KaTeX/KaTeX/blob/main/src/buildMathML.js?rgh-link-date=2025-10-09T18%3A21%3A37Z#L10) and separately [named](https://github.com/KaTeX/KaTeX/blob/main/src/buildMathML.js?rgh-link-date=2025-10-09T18%3A21%3A37Z#L14). After that both `TextNode` and `mathMLTree.TextNode` are used and I believe that in that case we rather talking about namespace cleansing rather than degradation.

Agreed, we should definitely move to just named exports here!

> 2\. The opposite situation with `delimeter.js` [default export](https://github.com/KaTeX/KaTeX/blob/main/src/delimiter.js?rgh-link-date=2025-10-09T18%3A21%3A37Z#L828). It is nicely encapsulated, but from 3 imports only 1 really benefits from grouping them: `genfrac.js` and `sqrt.js` both using only 1 function.

I'd support moving to non-default exports here.

So based on these two examples, I think this is a good cleanup project.

I'm still not sure about `Style`; as I said, the enum feel, and clarify of what `Style.DISPLAY` means, feels sad to lose. So maybe start a new PR for the others and close this one?

grigoriy-reshetniak:
Thank you for your feedback, I'll proceed with suggested changes.

PR #4072, opened 2025-08-24
test: add gui for the test server

### LINKED ISSUE #1225: GUI for test server (main.js) settings

grigoriy-reshetniak:
It feels like https://github.com/KaTeX/KaTeX/pull/3863 is abandoned so I decided to take it.

Test server was configurable via URL.

Test server is configurable via gui. URL is now for storing permalink, url configuration parameters like
` display=0` or `displayMode=0` (or `=f`/`=false`/`=n`/`=no`)` were removed as now it is possible to click checkbox. Permalink format is changed and is similar to one in the website.

BREAKING CHANGE: removed different flag parameters for test server. Permalink format is changed.

Closes #1225


grigoriy-reshetniak:
Hi @edemaine could you please review this PR?

grigoriy-reshetniak:
Hello @edemaine. Thank you for your feedback and sorry for the delay. I've addressed most of you points:
> I'm not a fan of how much screen real estate this takes

Button to collapse options panel added.
>This crashes when testing

Fixed, tested with example you have provided, as well as with default one from katex.org.
>I would prefer to see the URL update all the time

Done.
>I don't like replacing the previous nice URL format

To be honest, I thought that whole purpose of https://github.com/KaTeX/KaTeX/issues/1225 is to get rid of modifying url by hand and move to UI. Having three different flags to control the same thing doesn’t seem very efficient. By changing this format, it would also become compatible with katex.org, which I think is worth reconsidering.
>modern version is [URLSearchParams](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams) and you should use that

Done.

edemaine:
Thanks for the changes! This is looking better, but the sidebar is very cropped on my display on initial load:

image

After clicking on "displayMode":

image

No scrollbars so I can't scroll it. Even middle clicking won't let me scroll?

I also can't find the collapse button. Ah, that's because of Dark Reader, which makes the gear invisible. Maybe we should just use text instead?

So I guess the CSS still needs work.

I had forgotten that katex.org uses a JSON blob like this. I agree it's better to be compatible with that format.

edemaine:
I ended up working on this, with help of Codex. Will push some commits for a fancier GUI tomorrow.

grigoriy-reshetniak:
Okay, thanks for telling me before I've start fixing those. Also please let me know in case you want pass it back.

edemaine:
OK, I pushed my changes, and rebased to remove the conflict. Yes, could you please review to see what you think? Here's a summary:

* Generally tighter sidebars I think
* Adjustable partition for the sidebar (also remembered in localStorage)
* Switches to options on top in narrow view
* If there's an error, it's displayed instead of just in console (this is a long-standing issue I wanted to fix)
* A little more dark mode and a little color

It's maybe a little overkill, but this isn't part of the distribution, so hopefully not too much.

grigoriy-reshetniak:
Okay, so after reviewing it I can say that:
>Generally tighter sidebars I think
>Adjustable partition for the sidebar (also remembered in localStorage)

Looks nice.
>Switches to options on top in narrow view

I cannot achieve this one.
image
Also, I'm not even sure it's necessary, I doubt anyone will run dev server in the phone-sized screen environment.
>If there's an error, it's displayed instead of just in console (this is a long-standing issue I wanted to fix)

Really good improvement, close to how build warnings become full page errors for dev server. Maybe render errors deserve same treatment, but I'm not sure. 

>A little more dark mode and a little color

Improved it a little. This is how options toggle button looks like with Dark Reader on.

image

I've fixed in in last commit, now it looks like 
image

-------------------
Also, I've noticed some weird behavior for drop-down for Chromium-based browsers.
image

It happens only while in dev mode for Chrome and Edge while triggering smaller screen resolutions, but does not happen in Firefox. I guess it's related to https://issues.chromium.org/issues/441008122 and should be fixed on browser side.

edemaine:
> I cannot achieve this one.
> Also, I'm not even sure it's necessary, I doubt anyone will run dev server in the phone-sized screen environment.

Weird!  I get it as soon as I open a dev panel:

image

I often open Chrome on the left or right half of my display, so I have more of a portrait-sized window (taller than wide), and I think it's a useful view for that. Obviously not important, but I think it's somewhat helpful. (Though maybe an explicit toggle between the two layouts would be more helpful?)

> I've fixed in in last commit, now it looks like

Nice, thanks!

> Also, I've noticed some weird behavior for drop-down for Chromium-based browsers.

FWIW, I don't get this behavior. I'm running regular Chrome on Windows. So maybe it's Linux-specific. Anyway, at least it's still functional.

grigoriy-reshetniak:
Okay, it looks good to me!

edemaine (approved):

PR #4079, opened 2025-09-22
docs: update Delimiters table to add `/` and correct `\lt`, `\gt` rendering.

marcianx:
The PR corrects the following: https://katex.org/docs/supported#delimiters is missing `/` and the `\lt~\gt` preview shows < and > rather than angle brackets.


marcianx:
I've not figured out how to actually generate the HTML as shown at https://katex.org/docs/supported#delimiters. Any insights?

marcianx:
Ohhh, got it! I didn't realize there was a separate `package.json` in `website/`. Thanks for the directions!

PR #4091, opened 2025-10-07
feat: support alpha transparency in hex colors (#4067)

### LINKED ISSUE #4067: Support Alpha in Hex Colors
For using colors, [currently HTML color names or hex colors can be provided](https://katex.org/docs/supported.html#style-color-size-and-font), e.g.:

`\textcolor{blue}{F=ma}` -> $\textcolor{blue}{F=ma}$
`\textcolor{#228B22}{F=ma}` -> $\textcolor{#228B22}{F=ma}$

For the latter, it would be great if also _alpha_, i.e. transparency values could be supported. Typically, these would just be added to the hex color code, e.g. #228B22**80**
Latex also has the feature to specify transparency for colors like this:
```latex
\colorbox{blue!80}{Hello World}
```

KaranUnique:

4-digit hex colors (e.g., #RGBA) were not supported.
8-digit hex colors without a # prefix were incorrectly allowed.
The regex for color validation was repetitive and less maintainable.
Test suite included redundant parse tests.

Adds proper support for both 4-digit and 8-digit hex colors with alpha.
Requires a leading # for 8-digit hex codes (consistent with CSS spec).
Simplifies and optimizes the regex for better readability.
Updates the test suite to include valid and invalid alpha color cases.
Removes redundant parse tests and keeps the structure cleaner.

Fixes #4067

### DIFF
--- a/src/Parser.js
+++ b/src/Parser.js
@@ -680,7 +680,8 @@ export default class Parser {
         if (res == null) {
             return null;
         }
-        const match = (/^(#[a-f0-9]{3}|#?[a-f0-9]{6}|[a-z]+)$/i).exec(res.text);
+        const match = (/^(#[a-f0-9]{3,4}|#[a-f0-9]{6}|#[a-f0-9]{8}|[a-f0-9]{6}|[a-z]+)$/i)
+            .exec(res.text);
         if (!match) {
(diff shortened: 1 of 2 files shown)


edemaine:
Please don't open duplicate PRs.

PR #4098, opened 2025-10-31
chore: modernize commander API

parth-117:
### Summary
This PR updates the CLI import for `commander` to be compatible with newer Commander.js versions and fixes minor lint spacing so the pre-commit hook passes.

**Changes**
- Replace `require("commander").version(version)` with:
  ```js
  const { Command } = require("commander");
  const program = new Command();
  program.version(version);


edemaine (approved):
I'm not sure what lint wasn't passing for you, but this does seem like a reasonable future-proofing change.

I switched to the simpler `{program}` API, and fixed two other uses. If you could give a look over, that would be appreciated.

PR #4099, opened 2025-10-31
Fix: Use fixed decimal precision in unit outputFix decimal precision

parth-117:
This pull request fixes the issue where numerical outputs in KaTeX unit rendering showed excessive decimal digits.
The updated code ensures that numbers are rounded to a fixed precision (for example, 2 decimal places), improving readability and consistency across rendered math expressions.


grigoriy-reshetniak:
Hello @parth-117.
Sorry for the delay. I think we can't proceed because:

- Unrelated code should remain untouched.
- The precision point is agreed to be 4 digits after decimal separator. Please refer to https://github.com/KaTeX/KaTeX/pull/2460 for more info.

PR #4109, opened 2025-11-21
feat: support equals signs in \htmlData attributes

### LINKED ISSUE #4108: Data attributes via `\htmlData` cannot contain equals signs
The parser complains if you specify a data attribute whose value contains an equals sign. Which makes sense, given the overall syntax. But I could not find any way to escape them either. Is that something that could be added?

AljoschaMeyer:

`\htmlData{foo=<value>}{}` results in a parse error if `<value>` contains equals signs.

All equals signs but the first (which marks where the attribute name ends and the value starts) are ignored. Values can contain equals signs now.

Closes #4108

---

Did this in the Github WebUI, I'm unable to run the yarn tests locally =/

### DIFF
--- a/src/functions/html.js
+++ b/src/functions/html.js
@@ -53,18 +53,18 @@ defineFunction({
                 const data = value.split(",");
                 for (let i = 0; i < data.length; i++) {
                     const indexFirstEq = data[i].indexOf("=");
-                    const keyVal = [
-                            data[i].substring(0, indexFirstEq),
-                            data[i].substring(indexFirstEq + 1),
-                        ];
-                    const keyVal = data[i].split(/=(.*)/s);
-                    if (keyVal.length !== 3) {
-                        // if data[i] contains no equals sign, `keyVal` is `[data[i]]`
+
+                    if (indexFirstEq === -1) {
                         throw new ParseError(
                             "Error parsing key-value for \\htmlData, there was no equals sign at all");
                     }
+                    
+                    const keyVal = [
+                        data[i].substring(0, indexFirstEq),
+                        data[i].substring(indexFirstEq + 1),
+                    ];
+            
                     attributes["data-" + keyVal[0].trim()] = keyVal[1].trim();
-                    // keyVal[2] is the empty string
                 }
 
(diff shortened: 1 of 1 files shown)


src/functions/html.js
```diff
@@ -52,12 +52,14 @@ defineFunction({
             case "\\htmlData": {
                 const data = value.split(",");
                 for (let i = 0; i < data.length; i++) {
-                    const keyVal = data[i].split("=");
-                    if (keyVal.length !== 2) {
+                    const keyVal = data[i].split(/=(.*)/s);
+                    if (keyVal.length !== 3) {
```
edemaine: This seems like a weird use of `split` to me, though it does appear to work, provided there's no newline in the value (`.` doesn't match newline). Perhaps it would make more sense to do a regular expression match? Something like `const match = data[i].match(/^([^=]+)=([^]*)$/)`. This also makes it easy to correctly reject an empty key.

AljoschaMeyer: I had assumed an empty key should be allowed, since data attributes are allowed to be empty. Do you prefer a different syntax for setting empty data attributes?

Newlines are matched already of the [`/s` modifier](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/dotAll#description).

As for using `split` like this, that was simply the first result of a web search for splitting a string into two parts in js: https://stackoverflow.com/questions/4607745/split-string-only-on-first-instance-of-specified-character/4607799#4607799

I'll wait for your response and then implement whichever variant you prefer =)

edemaine: Thanks, TIL `/s` option!

I guess while we're debating the best form, the `indexOf` plus two `slice`s would actually be much more efficient (as discussed on that StackOverflow page), so maybe we should go for that. Allowing an empty key (so `data-` attribute) is fine I guess; seems we currently allow it, so probably better to still allow it.

AljoschaMeyer:
Apologies, I'm doing things with git that are not very clever... Will open a new PR from a clean branch with the commits squashed.

PR #4112, opened 2025-11-25
feat: support equals sign and surrounding whitespace in \htmlData attribute values

### LINKED ISSUE #4108: Data attributes via `\htmlData` cannot contain equals signs

AljoschaMeyer:

`\htmlData{foo=<value>}{}` results in a parse error if `<value>` contains equals signs.

All equals signs but the first (which marks where the attribute name ends and the value starts) are ignored. Values can contain equals signs now.

Closes #4108

---

@edemaine Continuing from #4109, with apologies for failing to git properly.

New changes compared to #4109:

- Implemented splitting at the first `"="` via `indexOf`.
- Added explicit (but primitive) tests ensuring that empty keys and empty values are not rejected by the parser.
- ["Squashed"](https://xkcd.com/1597/) everything into a single commit.

As before, I am unable to get the dev setup working locally; I have run none of the tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,13 @@
 # Changelog
 All notable changes to this project will be documented in this file. This CHANGELOG roughly follows the guidelines from [www.keepachangelog.com](https://keepachangelog.com/en/1.0.0/).
 
+## [0.16.26](https://github.com/KaTeX/KaTeX/compare/v0.16.25...v0.16.26) (2025-12-07)
+
+
+### Bug Fixes
+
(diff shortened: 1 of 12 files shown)


AljoschaMeyer:
Yes, I like keeping whitespace for the value, too 👍

Thank you for bearing with my technical restrictions that led to some subpar commits =)

Issue #4111, opened 2025-11-24
`\mathop  \int` shows a parse  error while parsed by other latex implementations


nightwing:
**Describe the bug:**
`\mathop  \int` shows a parse  error 

**To Reproduce:**
in https://katex.org/ `\mathop  \int` prints `KaTeX parse error: Got function '\int' with no arguments as argument to '\mathop' at position 10: \mathop \̲i̲n̲t̲ ̲`

**Expected behavior:**
 should render the same as `\mathop  {\int }`
 
**Environment (please complete the following information):**
 - KaTeX Version: '0.16.25'

edemaine:
Fixed by #4117

nightwing:
@edemaine should the default be changed to allowedInArgument? The issue is present with almost all other operators e.g.
```latex
${\mathop \bigcup} {\mathop \sum} {\mathop \bigwedge}$
```
and even
```latex
${\mathop \det} {\mathop \sin} $
```

edemaine:
@nightwing I agree that it's worth testing various operators, but I found that `\mathop \sin` and `\mathop \max` do not work in LaTeX, and I stopped there.

However, it does look like `\mathop \sum` and `\mathop \bigcup` work in LaTeX. Would you like to do a review of what operators are supported by LaTeX in this way and send a PR to change `allowedInArgument` accordingly?

Changing the default value of `allowedInArgument` could be considered, but not without reviewing every instance of where it isn't specified...

(Out of curiosity, why are you wrapping these symbols in `\mathop`? I've never seen them used as binary operators.)

nightwing:
indeed `{\mathop \sin}` doesn't work, i've made a mistake comparing with mathjax behavior. I'll do a more careful review and make a pull request.

The reason i found this issue, was that i have used Katex to render latex generated by MathType and it generates that strange construct in some cases.

PR #4117, opened 2025-12-06
fix: \mathop followed by integral symbol

ronkok:
Fixes issue #4111.


edemaine (approved):
Thanks! A surprisingly simple fix. I wondered about some of the other definitions in this file, but tested e.g. `\mathop\max` and it doesn't work in LaTeX.

PR #4125, opened 2026-01-22
fix(type): add missing types definition path to package.json

Sepush:

since v0.16.18 actually publish TypeScript type definitions but we did not export the top level `package.json.types`

ref https://github.com/KaTeX/KaTeX/releases/tag/v0.16.18 

add the missing `package.json.types` for compatibility

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,6 +3,7 @@
   "version": "0.16.27",
   "description": "Fast math typesetting for the web.",
   "main": "dist/katex.js",
+  "types": "types/katex.d.ts",
   "exports": {
     ".": {
       "require": {


Sepush:
seems release failed 
https://github.com/KaTeX/KaTeX/actions/runs/21261595421/job/61190330872

edemaine:
Yes, we're trying to figure out the fix here. If you have ideas, let me know. I think the tokens expired, but it still doesn't work with new tokens, perhaps because it entered a broken state, or perhaps because of policy changes...

Sepush:
> Yes, we're trying to figure out the fix here. If you have ideas, let me know. I think the tokens expired, but it still doesn't work with new tokens, perhaps because it entered a broken state, or perhaps because of policy changes...

I suspect this might be related to the new npm Access Token and 2FA policy changes. If that's the case, we could migrate to npm Trusted Publishing to solve this cleanly.
Here's a working example I found: [sxzz/workflows/release.yml](https://github.com/sxzz/workflows/blob/main/.github/workflows/release.yml) implements exactly this approach using GitHub OIDC authentication instead of long-lived tokens.

[npm Trusted Publishers](https://docs.npmjs.com/trusted-publishers)
[Publishing packages with provenance](https://docs.npmjs.com/publishing-packages-with-provenance)

edemaine:
I got releases working again, via fresh GitHub + NPM tokens, manually deleting the failed release branch (`git push origin --delete v0.16.28`), and force-pushing the merge commit (removing the failed release commit). I tried several times to get npm Trusted Publishing working, via the latest semantic-release which supports them (see https://github.com/KaTeX/KaTeX/tree/new-semantic-release), but I ran into issues summarized in https://github.com/semantic-release/npm/issues/1081 . Hopefully we can revisit once those are resolved, as trusted publishing sounds like a nice improvement.

PR #4136, opened 2026-02-16
refactor: switch utils to named exports

grigoriy-reshetniak:
Default export was used in `src/utils.js`

Named exports used in `src/utils.js`. Utility functions brought to same baseline: arrow functions. Some of the functions were inlined and removed.

Bundle size reduction:

- katex.js size is 1817 bytes smaller.
- katex.min.js size is 1435 bytes smaller.

Derived from https://github.com/KaTeX/KaTeX/pull/4064

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -52,7 +52,7 @@ export const getBaseElem = (group: AnyParseNode): AnyParseNode => {
     }
 };
 
-const nodes = new Set(["mathord", "textord", "atom"]);
+const characterNodesTypes = new Set(["mathord", "textord", "atom"]);
 
 /**
  * TeXbook algorithms often reference "character boxes", which are simply groups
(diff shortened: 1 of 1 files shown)


src/utils.js
```diff
@@ -6,20 +6,10 @@
 
 import type {AnyParseNode} from "./parseNode";
 
-/**
- * Provide a default value if a setting is undefined
- * NOTE: Couldn't use `T` as the output type due to facebook/flow#5022.
- */
-const deflt = function<T>(setting: T | void, defaultIfUndefined: T): * {
```
grigoriy-reshetniak, line 13: Replaced with `??`

src/utils.js
```diff
@@ -63,33 +52,22 @@ const getBaseElem = function(group: AnyParseNode): AnyParseNode {
...
-const isCharacterBox = function(group: AnyParseNode): boolean {
-    const baseElem = getBaseElem(group);
-
-    // These are all they types of groups which hold single characters
-    return baseElem.type === "mathord" ||
-        baseElem.type === "textord" ||
-        baseElem.type === "atom";
-};
-
-export const assert = function<T>(value: ?T): T {
```
grigoriy-reshetniak, line 80: Used only once, inlined in place.

edemaine:
P.S. I realize we should use `refactor:` commit messages for this project instead of `chore:`.

PR #4137, opened 2026-02-17
fix: `\*frac` sizing

### LINKED ISSUE #4080: \tfrac does not work when we have frac in the power
**Describe the bug:**
When we use \tfrac in power, it does not work. 

**To Reproduce:**
Use this expression to render and you won't see any difference with \tfrac here. 
`\dfrac{1}{2} + x^{\tfrac{a}{b}} = \dfrac{1}{2} + \sqrt[3]{x}`

**Expected behavior:**
When We used \tfrac, even in power, it should be formatted like normal fraction. 
**Screenshots:**

edemaine:

`\dfrac`, `\tfrac`, `\cfrac` used some custom code for forcing size to display, text, display, but it was incorrect: it didn't stack well with subscripts etc.

Here's a bad example:

```tex
\dfrac{a}{b}x_{\dfrac{a}{b}y_{\dfrac{a}{b}z_{\dfrac{a}{b}}}}
+
\tfrac{a}{b}x_{\tfrac{a}{b}y_{\tfrac{a}{b}z_{\tfrac{a}{b}}}}
+
\frac{a}{b}x_{\frac{a}{b}y_{\frac{a}{b}z_{\frac{a}{b}}}}
```

image

Notice how `\dfrac` gets progressively bigger, beyond actual display size. 😱 

I changed the code to simply return `type: "styling"` nodes, which already deal with sizing correctly. Admittedly this requires a type hack because we expect the builder to return a `type: "genfrac"` node, but it works well. Also the code is a lot simpler because we're no longer duplicating styling functionality.

image

Now the `\dfrac` set of examples are all the same size, as is `\tfrac`, while regular `\frac` is unchanged.

Fixes #4080

### DIFF
--- a/src/functions/genfrac.js
+++ b/src/functions/genfrac.js
@@ -1,7 +1,7 @@
 // @flow
 import defineFunction, {normalizeArgument} from "../defineFunction";
-import buildCommon from "../buildCommon";
-import delimiter from "../delimiter";
+import {makeLineSpan, makeSpan, makeVList} from "../buildCommon";
+import {makeCustomSizedDelim} from "../delimiter";
 import {MathNode, TextNode} from "../mathMLTree";
 import type {ParseNode} from "../parseNode";
 import Style from "../Style";
@@ -41,9 +41,9 @@ const htmlBuilder = (group, options) => {
     if (group.hasBarLine) {
         if (group.barSize) {
             ruleWidth = calculateSize(group.barSize, options);
-            rule = buildCommon.makeLineSpan("frac-line", options, ruleWidth);
+            rule = makeLineSpan("frac-line", options, ruleWidth);
         } else {
-            rule = buildCommon.makeLineSpan("frac-line", options);
+            rule = makeLineSpan("frac-line", options);
         }
         ruleWidth = rule.height;
         ruleSpacing = rule.height;
@@ -86,7 +86,7 @@ const htmlBuilder = (group, options) => {
             denomShift += 0.5 * (clearance - candidateClearance);
         }
 
-        frac = buildCommon.makeVList({
+        frac = makeVList({
             positionType: "individualShift",
             children: [
                 {type: "elem", elem: denomm, shift: denomShift},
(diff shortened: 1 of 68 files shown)


edemaine:
This PR changes two screenshots. Here are the diffs on Chrome:

| Test | Diff |
|-----|-----|
| FractionTest | FractionTest-chrome-diff |
| Raisebox | Raisebox-chrome-diff |

Did these minor spacing changes improve or worsen things? To evaluate, I ran our texcmp script to compare against LaTeX's rendering.

| Test | Before PR | After PR |
|-----|------------|----------|
| FractionTest | FractionTest_before | FractionTest_after |
| Raisebox | Raisebox_before | Raisebox_after |

Nothing is perfect, but my sense is that the changes from this PR are improvements:
* FractionTest has a significant horizontal shift in the bottom half, and that brings things into more alignment with LaTeX
* Raisebox has a significant vertical shift in the bottom half, and that brings things into more alignment with LaTeX

edemaine:
@nabeel-hussain @grigoriy-reshetniak If either of you have time to review, it would be appreciated! You can check out the new behavior at https://deploy-preview-4137--katex.netlify.app/

src/functions/genfrac.js
```diff
@@ -289,60 +282,31 @@ defineFunction({
...
-            case "\\dbinom":
-                size = "display";
-                break;
-            case "\\tfrac":
-            case "\\tbinom":
-                size = "text";
-                break;
+        const continued = funcName === "\\cfrac";
+        let style = null;
+        if (continued || funcName.startsWith("\\d")) {
```
grigoriy-reshetniak, line 287: I don't like `switch` with two `case`s, but prior to this change it was clearer which `case` related to which function.

grigoriy-reshetniak:
Besides that comment ,which comes to personal preference, I think PR looks great.

src/functions/genfrac.js
edemaine: Fair point. I was more thinking about efficiency here than clarity, trying to avoid lots of string comparisons for what is fundamentally "does it start with `\d`(isplay) or `\t`(ext)?" I'm inclined to keep this version if you think it's OK.

grigoriy-reshetniak: I think it's fine.

edemaine:
@greptileai review this

PR #4138, opened 2026-02-17
refactor: switch buildCommon to named exports

grigoriy-reshetniak:
Default export was used in `src/buildCommon.js`

Named exports used `src/buildCommon.js`

Bundle size reduction:

- katex.js size is 2.2 Kb smaller.
- katex.min.js size is 1.8 Kb smaller.

Derived from https://github.com/KaTeX/KaTeX/pull/4064

### DIFF
--- a/src/functions/mclass.js
+++ b/src/functions/mclass.js
@@ -1,6 +1,6 @@
 // @flow
 import defineFunction, {ordargument} from "../defineFunction";
-import buildCommon from "../buildCommon";
+import {makeSpan} from "../buildCommon";
 import {isCharacterBox} from "../utils";
 import {MathNode} from "../mathMLTree";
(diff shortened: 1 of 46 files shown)


grigoriy-reshetniak:
It seems that Screenshotter for safari failed for reasons unrelated to content of the PR. Would it be possible to restart it?

src/functions/mclass.js
```diff
@@ -10,8 +10,6 @@ import * as mml from "../buildMathML";
 
 import type {ParseNode} from "../parseNode";
 
-const makeSpan = buildCommon.makeSpan;
-
```
edemaine, line 14: The fact that we did this before a couple of times is a good sign we're on the right track here. 🙂

edemaine (approved):
Nice, much cleaner!

PR #4140, opened 2026-02-17
fix: no line breaks after `\not`

### LINKED ISSUE #4077: \not\in wraps incorrectly in inline formulas
As demonstrated in [NotIn.html](https://github.com/user-attachments/files/22301828/NotIn.html), when `\not\in` wraps across a line in an inline formula, the slash often appears on its own at the end of the line, while the `\in` symbol appears on its own at the start of the next line. It looks like this:

Image

 - KaTeX Version: "0.16.22" 
 - Device: Desktop
 - OS: Ubuntu 20.04
 - Browser: Firefox 136.0 (64-bit) and Chrome Version 139.0.7258.127 (Official Build) (64-bit)

edemaine:
Automatic line breaking could happen after `\not`, preventing it from sticking to the symbol after it. This was somewhat common because `\not` has negative size.

Automatic line breaking is forbidden after `\not`, just by adding `\nobreak`.

In addition to the automatic tests, I tested in our test environment (`yarn start`) on Chrome for Windows, and dragging the window size back and forth. I always get something like this:

image

Fixes #4077

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,13 @@
 # Changelog
 All notable changes to this project will be documented in this file. This CHANGELOG roughly follows the guidelines from [www.keepachangelog.com](https://keepachangelog.com/en/1.0.0/).
 
+## [0.16.29](https://github.com/KaTeX/KaTeX/compare/v0.16.28...v0.16.29) (2026-02-22)
+
+
+### Bug Fixes
+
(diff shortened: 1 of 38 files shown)


edemaine:
@OsePedro Could you test that this fixes your issue? You can test the new behavior at https://deploy-preview-4140--katex.netlify.app/ - or maybe download the katex.js from there to test your example.

OsePedro:
Thanks for the fix! I've tested it on my example with various div widths, and it seems to work correctly.

edemaine:
@greptileai review this

PR #4141, opened 2026-02-17
refactor: remove default exports from stretchy and delimiter

grigoriy-reshetniak:
Default export was used in `src/delimiter.js` and `src/stretchy.js`

Named exports used `src/delimiter.js` and `src/stretchy.js`. One function was inlined.

Bundle size reduction:

- katex.js size is 579 bytes smaller.
- katex.min.js size is 224 bytes smaller.

Derived from https://github.com/KaTeX/KaTeX/pull/4064, last of this kind.

### DIFF
--- a/cli.js
+++ b/cli.js
@@ -20,7 +20,9 @@ try {
 const {version} = require("./package.json");
 const fs = require("fs");
 
-const program = require("commander").version(version);
+const {program} = require("commander");
+program.version(version);
+
 for (const prop in katex.SETTINGS_SCHEMA) {
     if (katex.SETTINGS_SCHEMA.hasOwnProperty(prop)) {
         const opt = katex.SETTINGS_SCHEMA[prop];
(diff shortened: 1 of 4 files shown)


src/functions/accent.js
```diff
@@ -195,7 +195,7 @@ export const htmlBuilder: HtmlBuilderSupSub<"accent"> = (grp, options) => {
 const mathmlBuilder: MathMLBuilder<"accent"> = (group, options) => {
     const accentNode =
         group.isStretchy ?
-            stretchy.mathMLnode(group.label) :
+            mathMLnode(group.label) :
```
edemaine: Without the `stretchy.` namespace, `mathMLnode` feels rather generic. Maybe `stretchyMathML`? Or `stretchyMathNode`? Should it start with `make`?

I also wonder whether we should rename `svgSpan` to `stretchySvg` (it's confusingly close to `makeSvgSpan`) andf `encloseSvg to `stretchyEnclose`... What do you think?

edemaine (approved):

PR #4145, opened 2026-02-22
chore: port from Flow to TypeScript

### LINKED ISSUE #2110: Consider using TypeScript instead of Flow
Related: #1938 

@kevinbarabash It seems you own https://github.com/Khan/flow-to-ts, how difficult is it to migrate to TS?

grigoriy-reshetniak:
It was impossible to have .ts files in the project.

It is possible now.

I've seen https://github.com/KaTeX/KaTeX/pull/2326 and came to the conclusion that there was an intent to migrate to TypeScript at some point. However I don't see how it could be done at once with sufficient confidence.
So this is an attempt to migrate only one .ts file to solve all dependency issues now and deal with migration itself later, in chunks.

IMHO weakest point is [src/ParseError.d.ts](https://github.com/KaTeX/KaTeX/compare/main...grigoriy-reshetniak:add-typescript-support?expand=1#diff-37bb1f75ffba7ddce7439c3647a6d4f8044fa8323907d86e7faa10957b216d5d). It is a temporary file, but it's there to stay for a while and in case of gradual migration there would more like it.

---

This PR has since grown to a full port to TypeScript.

Fixes #2110

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,7 +33,7 @@ jobs:
       run: yarn test:lint
 
     - name: Run TypeScript typecheck
-      run: yarn test:typecheck
+      run: yarn test:ts
       if: always()
 
     - name: Run Jest tests
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -141,7 +141,7 @@ files using ESLint and stylesheets using stylelint. They must pass to commit
 the changes.
 
 TypeScript files can be checked for type errors by running
-`yarn test:typecheck`.
+`yarn test:ts`.
 
 ## Pull Requests
 
--- a/package.json
+++ b/package.json
@@ -137,7 +137,7 @@
   },
   "bin": "cli.js",
   "scripts": {
-    "test": "yarn test:lint && yarn test:typecheck && yarn test:jest",
+    "test": "yarn test:lint && yarn test:ts && yarn test:jest",
     "test:lint": "yarn test:lint:js && yarn test:lint:css",
     "test:lint:js": "eslint .",
     "test:lint:css": "stylelint src/styles/katex.scss static/main.css website/static/**/*.css",
@@ -148,7 +148,7 @@
     "test:screenshots": "yarn test:screenshots:update --verify",
     "test:screenshots:update": "dockers/screenshotter/screenshotter.sh",
     "test:perf": "NODE_ENV=test node -e \"require('@babel/register')({extensions:['.js','.ts']}); require('./test/perf-test.ts')\"",
-    "test:typecheck": "tsc --noEmit",
+    "test:ts": "tsc --noEmit",
     "clean": "rm -rf dist/ node_modules/",
     "clean-install": "yarn clean && yarn",
     "start": "webpack serve --config webpack.dev.js",


edemaine:
This is exciting! Yes, I think we should move to TypeScript; it's far more popular, and tooling is far better in my experience. Doing the migration in pieces reminds me of when we migrated to Flow, which was also done in a long series of PRs.

However, this time we have the complication of two type-checking systems co-existing. I like your `src/ParseError.d.ts` workaround so that TypeScript can understand the types of the Flow subsystem. Is it possible to do something in the other direction, where Flow can get be given the types of the TypeScript code? Otherwise, we're not going to be able to typecheck the system as a whole — potentially, there could be bugs at the boundary between TypeScript and Flow.

If some kind of temporary glue isn't possible, I see a few possibilities:
* Pause other work while we finish the migration. Not ideal, but until you and I started working together recently, things have been slow, so it wouldn't be the end of the world.
* Work on another branch. Might lead to conflicts.
* Try to do the migration faster. Given your start, I'm going to see if Codex can finish it...

grigoriy-reshetniak:
> Is it possible to do something in the other direction, where Flow can get be given the types of the TypeScript code?

I don’t have much experience with Flow. Maybe it could be done with `.js.flow` files in the same way, how it's done with `d.ts`, but this could have several problems:

- Migration would require x2 temporary files. Looking at imaginary singular PR supporting two-way type system may take more effort than migration itself.
- To ease the support of this two way system types would become simpler. I feel like this migration is great opportunity to touch many of `//$FlowFixMe` without making them `as any` instead. But if TS types would become somewhat complex translating them to Flow may require simplification.

Coming to the other options:

- Pause could be a solution, especially combined with option 3. But tbh I don't see it as hard stop, if things are not touching both `.ts` and `.js` files they _should be_ safe.
- I'm not in favor of this idea, and merge conflicts are not even my main concern. A PR of this scale would be very difficult to review and maintain. If the author gets pulled into other work, it risks being abandoned -- and given its size, it would become very hard to recover. This is what happen to https://github.com/KaTeX/KaTeX/pull/2326 per my understanding.

I’m strongly in favor of a chunked approach to migration. Even for this PR I haven't tried actually dig in type system of `wide-character,` because it aims to fix the config. It allows a little refactorings there and there, it creates smaller pressure in terms of releasing changes.

edemaine:
Yeah, if we move relatively quickly, there probably won't be issues at the boundary. 

> I'm going to see if Codex can finish it...

About 10M tokens later, here's the result: https://github.com/KaTeX/KaTeX/tree/flow-to-ts in particular https://github.com/KaTeX/KaTeX/commit/32b761a47aa808265a74b49c21218cf47e739934

There's "only" about 1000 lines to review. I've gone through it at a high level, and I think it's not bad for a first pass. Most of the typing workarounds were in the old code as well, and we could try to fix them later. Along with other things like maybe automatically computing `types/katex.d.ts` (if that's feasible).

I leave it up to you to decide whether to pull in that commit (and presumably add your desired edits), or submit pieces of it in staged PRs. It's looking relatively clean and reviewable though. (Admittedly I had to work pretty hard to get Codex to minimize the diff, but I think I largely succeeded.)

grigoriy-reshetniak:
I've reviewed https://github.com/KaTeX/KaTeX/commit/32b761a47aa808265a74b49c21218cf47e739934 and I believe it needs to be included in this PR. Although this approach has disadvantages like concessions in `.eslintrc` and `tsconfig.json`, I feel like it'sbetter to be 20%-40%-60% migrated and still have Flow. I wont comment on types workarounds, because it will just delay the PR and could be indeed done separately.

However, I have a few questions moving forward:

1. How to deal with `undefined` in functions parameters? I see both `key?: value` and `key: Type | undefined` and even `key?: Type | undefined`. Could we please have `key?: value` as a standard?
2. I see a favor for tuples over object-like types. Is this historical artifact or is there any performance reason or strong personal preference behind it? How do you feel about converting some of the tuples into named objects?

P.S. After merging `main` and `flow-to-ts` into current branch there are some conflicts. Also `flow-to-ts` fails to build on it's own. I’m looking into it, but new issues keep popping up.

edemaine:
It looks like `test/screenshotter/ss_data.js` got converted to TypeScript but needs to be reverted. 

1. I guess I'm used to `exactOptionalPropertyTypes: true` where these have different meanings. But I'm fine with standardizing on something simple like `foo?: T`. Happy to get Codex to do the heavy lifting once we decide.

2. Maybe better to keep data structures the same for this PR, but I'm not against a refactor. We should consider V8 efficiency (this is complicated and I am no expert) and verbosity of output (large tables take fewer characters as arrays).

grigoriy-reshetniak:
Hello @edemaine.

PR is finally ready for review. I had to change a couple of things from your migration, please take closer look on those parts.

test/__snapshots__/katex-spec.ts.snap
```diff
@@ -1046,6 +1046,7 @@ exports[`An includegraphics builder should render with trust setting 1`] = `
     ],
     "depth": 0,
     "height": 0.9,
+    "maxFontSize": 0,
```
edemaine, line 1049: Should we worry about this addition?

edemaine (approved):
Looks great!

One general annoyance I have is all the `| null | undefined`. I wonder if we could shift to `undefined` as the default value, and get rid of all the `null`s. Probably not for this PR; it's big enough!

I have one question but I think we're otherwise good to go.

edemaine:
Oops, I forgot to mention that I pushed a few changes, mostly reducing the diff size, but also resolving the `any`s in auto-render. Could you review those?

test/__snapshots__/katex-spec.ts.snap
grigoriy-reshetniak: I've investigated this and it seems that this comes from https://github.com/KaTeX/KaTeX/pull/4145/changes/BASE..d82f10c4c4eb2b722e4f109ece7ddb6b448f2f51#diff-0c0f91820ce934947c04bb2d92efa46f99c07e266c87262a55a5990272f6b443R36. Previously default value was `undefined` and was stripped during snapshot serialization, but now it became `0`. 

In this particular case I don't see problem with image having `font-size` as 0.

grigoriy-reshetniak:
> Could you review those?

LGTM

edemaine:
Sorry, a few more commits to review:
1. I grepped for `flow` and found some comments and json that needed changing.
2. Realized `unicodeSupOrSub` can be ported from js to ts
3. Restored some Flow comments into JSDoc, although we can't easily typecheck `webpack*.js` because TypeScript and Yarn PnP don't play well together. 😢 But JSDoc also let me remove some of the `.d.ts` files, which is cleaner.

webpack.common.js
grigoriy-reshetniak, line 1: I feel like it's not worth to type-check config files at this stage. This file has 8 `@ts-ignore` -- for me it's good sign it can be omitted for now.

I like rest of the changes.

edemaine:
@grigoriy-reshetniak I noticed https://github.com/KaTeX/KaTeX/pull/2326#issuecomment-685801308 - maybe we should move to mhchem.ts in the future. There are some other comments in that PR that might be worth picking up (e.g. removing no-longer-relevant comments about Flow typing).

But I'm very happy we have a TypeScript codebase now!!

grigoriy-reshetniak:
Okay, I'll go through comments for that PR.

Thank you for your collaboration!

Issue #3972, opened 2024-08-20
Scriptstyle font size is too big


amblafont:
The ratio between the normal font size and the scriptstyle font size differs from standard LaTeX

1) In Latex (the second line is script style)
![image](https://github.com/user-attachments/assets/d53d095a-3289-40a1-99b8-fb5bffd46be8)

2) Same code in KaTeX
![image](https://github.com/user-attachments/assets/cd8984bb-52ed-4381-bf0b-c62ed89cf6ad)

edemaine:
You seem to be right, though the results differ substantially depending on LaTeX's size mode (10pt, 11pt, or 12pt).

Reproduction:
```latex
\begin{array}{c}
\displaystyle
abcdefghijklmnopqrstuvwxyz\\
\scriptstyle
abcdefghijklmnopqrstuvwxyz
\end{array}
```

LaTeX 10pt:

![image](https://github.com/user-attachments/assets/7b77cea8-3d10-4a70-ad07-58dca9c09ef8)

LaTeX 11pt:

![image](https://github.com/user-attachments/assets/4999cafd-7a9a-4635-85c1-96bdce48f8d6)

LaTeX 12pt:

![image](https://github.com/user-attachments/assets/65e3cfd9-21bf-4074-8c10-fe91fa48cd9d)

KaTeX:

![image](https://github.com/user-attachments/assets/59e0043f-7c87-4351-a234-5f5fced85ee2)


I'm surprised, because I thought the size constants were taken directly from LaTeX measured at 10pt.

edemaine:
I took a look at this again, and I think KaTeX does match LaTeX's 10pt font heights pretty well:

<img width="1241" height="276" alt="Image" src="https://github.com/user-attachments/assets/b239f98a-9594-4b22-a83d-1f02d581d3b9" />

The main issue is that TeX's 10pt subscript font (cmmi7) has substantially wider characters than a scaled-down 10pt point (cmmi10), and KaTeX does the latter. This makes our subscripts feel smaller, but the heights are correct. (There's also clearly a kerning or italic correction issue in `jk`.)

Maybe someday we could vectorize the smaller fonts too, but I also worry that this would make for a substantially larger network payload.

I'm going to close this for now, unless someone has a clear direction to proceed.

PR #4147, opened 2026-02-24
feat: \underbracket and \overbracket

### LINKED ISSUE #4103: Add support for `\underbracket` and `\overbracket`
Hi, I have an old document that I'd like to digitalize, and it uses essentially `\underbracket` that is not currently supported. I want to faithfully reproduce the document so alternatives like `\undergroup` are less desirable. I suppose implementing it should be almost the same as `\underlinesegment`, only slightly shifted, yes? Please add support for it, thanks!

grigoriy-reshetniak:
Adds support for `\underbracket `and `\overbracket`.

I wasn't able to update screenshots due to images incompatibility.

Closes #4103


grigoriy-reshetniak:
Thanks for the feedback, I'll work on the dimensions.

grigoriy-reshetniak:
I've addressed review comments:
- Thickness and height are closer to LaTeX now (compared with `mathtools`)
image

- Also it seems to me that `mathtools` puts some kern around brackets regardless of the surroundings, I've tried to replicate this too.
- There are fewer unit tests.

src/functions/horizBrace.js
```diff
@@ -32,6 +32,8 @@ export const htmlBuilder: HtmlBuilderSupSub<"horizBrace"> = (grp, options) => {
         group = assertNodeType(grp, "horizBrace");
     }
 
+    const atomClass = /bracket/.test(group.label) ? "minner" : "mord";
```
edemaine: I think `String.prototype.includes` is slightly faster than `RegExp.prototype.test`.

Also, are you sure it's just brackets? I'm seeing some space in `\underbrace` in vanilla LaTeX, though I didn't check the class yet.

grigoriy-reshetniak: `RegExp.prototype.test` replaced with `String.prototype.includes`.

No, it's not just brackets. There's also a space for `\underbrace` and `\underline`, but it's inconsistent for surroundings. This is output by `pdflatex` for 

`test\underline{hello}test\underbracket{hello}_{world}test\underbrace{hello}_{world}test`

image

`\underbrace` has selectable space before it. At the same time, with input

`\underline{hello}\underbracket{hello}_{world}\underbrace{hello}_{world}`

Output looks like this:
image
Kern is present, but the selectable space is gone. 

I feel like looking into this could also involve `\underline` and over-* functions as well, and it's out of scope for `\underbracket` and `\overbracket `support.

edemaine: Your screenshots look like you're starting in text mode, and LaTeX is crashing and automatically entering math mode. You should start in math mode.

Here's your examples:

image

image

It's relevant because I guess this line of code should just treat underline specially... but it could use more investigation. https://github.com/KaTeX/KaTeX/wiki/Examining-TeX should help.

ronkok:
I am commenting just to say that this PR is cool. @grigoriy-reshetniak, you are the first contributor other than me to use SVGs to extend KaTeX. I'm happy that someone has been able to follow and extend the methods that I used.

@edemaine has been maintaining KaTeX by himself for a while now, and I'm glad that you have stepped up to give him some support.

src/functions/horizBrace.js
grigoriy-reshetniak: `\underbrace` indeed had a space as `\underbracket`, same with `\overbrace`. Updated, changes are ready for review. I had to force-push, as it was easier to do after migration to TS.

src/functions/horizBrace.ts
```diff
@@ -31,6 +33,8 @@ export const htmlBuilder: HtmlBuilderSupSub<"horizBrace"> = (grp, options) => {
         group = assertNodeType(grp, "horizBrace");
     }
 
+    const atomClass = new Set(functionNames).has(group.label) ? "minner" : "mord";
```
edemaine: We should probably build the Set once globally. Or else use `functionNames.includes`.

test/katex-spec.ts
```diff
@@ -2612,16 +2612,55 @@ describe("A horizontal brace builder", function() {
...
+    });
+
+    it("should be grouped more tightly than supsubs", function() {
+        const parse = getParsed`\overbracket x^2`[0];
+
+        expect(parse.type).toEqual("supsub");
+    });
+});
+
+describe("A horizontal bracket builder", function() {
```
edemaine, line 2666: Got duplicated?

grigoriy-reshetniak: Sorry, I'm a bit lost. I don't see any duplicates, previous `describe` is for "parser" and this one is for "builder". 
It's the same way at https://github.com/KaTeX/KaTeX/pull/4147/changes#diff-aef84d06a1cee8255e4fe8d1af62697de550d03685b5b43c5cc34d655bd55cabR2607 and https://github.com/KaTeX/KaTeX/pull/4147/changes#diff-aef84d06a1cee8255e4fe8d1af62697de550d03685b5b43c5cc34d655bd55cabR2584

src/functions/horizBrace.ts
grigoriy-reshetniak: Replaced with `Array.includes()`

edemaine: Actually, why do we need this test at all? Isn't it always true?

grigoriy-reshetniak: It is. I thought maybe it makes sense because this `htmlBuilder` is used in `supsub`, but it seems that this check is redundant now.

grigoriy-reshetniak:
> Can you add the screenshots from the Screenshotter actions' uploaded artifacts?

Unfortunately, I'm not able to find them. I'm able to see artifacts from failed builds, like https://github.com/KaTeX/KaTeX/actions/runs/22549692911, but its not available for successful ones (I believe because no diff is introduced), like https://github.com/KaTeX/KaTeX/actions/runs/22796025639?pr=4147.

This is KaTeX output in dev server for 
```
\text{test}\underline{hello}\text{test}\underbracket{hello}_{world}\text{test}\underbrace{hello}_{world}\text{test}
\\[1em]
\text{test}\overline{hello}\text{test}\overbracket{hello}^{world}\text{test}\overbrace{hello}^{world}\text{test}
```
image

This is LaTeX output for the 

```
\begin{align*}
  \text{test}\underline{hello}\text{test}\underbracket{hello}_{world}\text{test}\underbrace{hello}_{world}\text{test} \\[1em]
  \text{test}\overline{hello}\text{test}\overbracket{hello}^{world}\text{test}\overbrace{hello}^{world}\text{test}
\end{align*}
```

image

It seems to me that it's pretty close now.

edemaine:
> (I believe because no diff is introduced

Actually I think this was #4158! Hopefully it works now.

edemaine:
I ran `texcmp` (from `dockers/texcmp`) on the screenshotter example and the two example lines from your comment, and they indeed look good:

HorizontalBrackets

PR4147CommentOver

PR4147CommentUnder

If I were to nit pick, I'd say the green (KaTeX) under/overbracket is slightly thicker. Did you get the line thickness measurement from LaTeX somewhere, or did you eyeball it? If it's from LaTeX, I'm happy to keep it as is; maybe just a browser rendering thing.

The horizontal spacing looks perfect! And ignore the sub/superscripts looking so different; that's #3972.

grigoriy-reshetniak:
I did it by eye. However, I've checked documentation for [mathtools](https://texdoc.org/serve/mathtools/0) and in `3.3.2 Braces and brackets` it says:

>The default rule thickness is equal to that of \underbrace (app. 5/18 ex)

KaTeX `\underbrace` is has thickness of 120 (SVG path), but current implementation of `\underbracket` has thickness of 100. 

Per my understanding of documentation `\underbracket` should be even thicker (or `\underbrace` thinner to match). At the same time I made it thinner before and it seems to be too thick anyway.

What do you think?

edemaine:
I think I must have red and green backwards. Indeed, the texcmp README says green is LaTeX. (I was confused because LaTeX is the red channel, but light is additive, so...)

So overbracket is indeed too thin, and I'd up it to 120 like the documentation says.
Indeed, I think the right answer is approximately 120, which according to `stretchy.js` corresponds to 0.120em. This code:

```latex
\documentclass{article}
\begin{document}
$
\newlength\LEN
\setlength\LEN{1em}
\showthe\LEN
\setlength\LEN{0.277777ex}
\showthe\LEN
$
\end{document}
```

measures 1em at 10.00002pt and 5/18 ex at 1.19595pt, so 5/18 ex = 0.119595 em. Very close to 0.120em.

I agree it looks thick, but compatibility is a clearer target. In the future, we could try to make the thickness configurable — if that's even possible...

test/katex-spec.ts
edemaine: Oops, I didn't realize. Sorry!

edemaine:
Here are the texcmp's with 120:

HorizontalBrackets

PR4147CommentOver

PR4147CommentUnder

I think this is a bit better: there's now both red and green of roughly equal amount. So the thickness seems right, just vertical offset is slightly off.

ronkok:
I agree that `\underbracket` thickness should be 0.12 em, per mathtools docs and KaTeX `fontMetrics` definition of x-height.

To define `\underbrace` geometry, I converted a KaTeX font into an SVG font, copied a brace glyph, rotated it 90°, and used that path.

If you really want to make the thickness configurable, there is a go-by in current KaTeX settings. `delimiter.js` reads `minRuleThickness` from the KaTeX settings and passes a value to `svgGeometry`, which modifies the thickness of a `sqrt` vinculum.  It's a lot of trouble for something used very infrequently. I recommend sticking with 0.12 em.

src/svgGeometry.ts
```diff
@@ -259,11 +259,11 @@ v40h399900v-40zM0 241v40h399900v-40zm0 0v40h399900v-40z`,
...
 M40 281 V428 H0 V94 H40 V241 H400000 v40z`,
 
-    leftbracketunder: `M0 0 h100 V310 H400000 v100 H0z
-M0 0 h100 V310 H400000 v100 H0z`,
+    leftbracketunder: `M0 30 h100 V340 H400000 v120 H0z
+M0 30 h100 V340 H400000 v120 H0z`,
 
-    leftbracketover: `M0 410 h100 V100 H400000 v-100 H0z
-M0 410 h100 V100 H400000 v-100 H0z`,
+    leftbracketover: `M0 460 h100 V150 H400000 v-120 H0z
```
edemaine: Shouldn't the 100s be 120 here? I think the vertical segments didn't thicken.

grigoriy-reshetniak: Thanks, fixed it.

grigoriy-reshetniak:
Next thing I need to do -- update docker images for screenshotter. I'm not able to run it on arm architecture so I have to take image from artifacts.

edemaine:
> Here are the texcmp's with 120:
> 
> HorizontalBrackets

I reran texcmp and the offsets seem slightly worse:

HorizontalBrackets

The top red text seems slightly higher, and the bottom red text seems slightly lower, than before, but should be the other way around.

grigoriy-reshetniak:
Here's fresh diff from `texcmp`:

image

PR #4150, opened 2026-02-28
Fix #4128: Ensure MathML scripts apply to full parens

### LINKED ISSUE #4128: Parens in LaTeX not converted to MathML properly
When converting LaTeX to MathML, KaTeX is not handling parenthesized expressions properly.

1. Go to the KaTeX homepage, https://katex.org.
2. In the "Type an expression:" box, type (x)^2
3. Open the Elements tab of DevTools to inspect the MathML that KaTeX has added to the DOM.
```
<math xmlns="http://www.w3.org/1998/Math/MathML" display="block">
  <semantics>
    <mrow>
      <mo stretchy="false">(</mo>
      <mi>x</mi>
      <msup>
        <mo stretchy="false">)</mo>
        <mn>2</mn>
      </msup>
    </mrow>
    <annotation encoding="application/x-tex">(x)^2</annotation>
  </semantics>
</math>
```

=> Note: in the `<msup>` the base is `<mo stretchy="false">)</mo>` So the exponent applies to just the close paren.

Expected: the base in the `<msup>` to be the whole parenthesized expression wrapped in an `<mrow>`:

```
<math xmlns="http://www.w3.org/1998/Math/MathML" display="block">
  <msup>
    <mrow>
      <mo stretchy="false">(</mo>
      <mi>x</mi>
      <mo stretchy="false">)</mo>
    </mrow>
    <mn>2</mn>
  </msup>
</math>
```

It is important for the MathML to have the proper semantic structure for compatibility with other apps.

**Environment**
 - KaTeX Version: 0.16.28
 - Device: Desktop PC
 - OS: Windows 11 25H2
 - Browser: Edge
 - Version: 144.0.3719.82

Image

agentic-sanyam:
Hey team!

This PR addresses #4128 where parenthesized LaTeX expressions like `(x)^2` weren't getting converted to semantically correct MathML. Previously, the `msup` (superscript) would incorrectly attach its exponent only to the closing parenthesis `)`.

I've updated `src/buildMathML.js` to correctly handle these cases. Here's the gist:
*   We now identify closing delimiters that are acting as the base of a script (`msup`, `msub`, `msubsup`).
*   When such a delimiter is found, we backtrack through the already built MathML nodes to find its matching opening delimiter.
*   Once a balanced pair is identified, all the MathML nodes from the opening delimiter to the closing delimiter (inclusive) are wrapped into a new `<mrow>`. This `<mrow>` then becomes the proper base for the script node, ensuring the exponent or subscript applies to the entire parenthesized expression.

To prevent regressions, I've added a dedicated test suite in `test/mathml-spec.js` covering simple and nested parenthesized scripts.

This should make KaTeX's MathML output more accurate and compatible with other tools!

Cheers,

Fixes #4128

### DIFF
--- a/src/buildMathML.js
+++ b/src/buildMathML.js
@@ -21,6 +21,29 @@ import type {FontVariant, Mode} from "./types";
 const noVariantSymbols = new Set(["\\imath", "\\jmath"]);
 const rowLikeTypes = new Set(["mrow", "mtable"]);
 
+const closeToOpen = {
+    ")": "(",
+    "]": "[",
+    "}": "{",
+    "\u27e9": "\u27e8", // \rangle, \langle
+    "\u230b": "\u230a", // \rfloor, \lfloor
+    "\u2309": "\u2308", // \rceil, \lceil
+    "\u27ef": "\u27ee", // \rgroup, \lgroup
+    "\u23b1": "\u23b0", // \rmoustache, \lmoustache
+};
+
+/**
+ * Helper to get the text content of a MathNode if it's an <mo> with a single
+ * TextNode child.
+ */
+const getmoText = (node: ?MathDomNode): ?string => {
+    if (node instanceof MathNode && node.type === "mo" &&
+        node.children.length === 1 && node.children[0] instanceof TextNode) {
+        return node.children[0].text;
+    }
+    return null;
+};
+
 /**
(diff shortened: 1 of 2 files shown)


grigoriy-reshetniak:
@agentic-sanyam could you please resolve conflicts?

PR #4151, opened 2026-03-05
fix: emoji with color variation selector

### LINKED ISSUE #3842: Rainbow flag emoji not rendered correctly
If the [rainbow flag emoji](https://emojipedia.org/rainbow-flag/) (🏳️‍🌈) is present in a KaTeX expression, it will render as "🏳®🌈". For instance, on [katex.org](https://katex.org/):

image

edemaine:
U+FE0F ([VARIATION SELECTOR](https://en.wikipedia.org/wiki/Variation_Selectors_(Unicode_block))-16) was treated like `\textregistered`, causing weird behavior reported in #3842.

It looks like I introduced this in #1454, specifically https://github.com/KaTeX/KaTeX/pull/1454/changes/1c801e5b5fe8fa1bdb2667dbb96ed9f823760014 . I think the reason was a page like https://emojiterra.com/registered/ which mentions U+00AE U+FE0F is a registered trademark emoji ®️, and I interpreted the sequence as two individual characters. In fact, ®️ used to render as two registered trademarks in a row, which is incorrect.

U+FE0F remains as is, enabling color selection in emoji like 🏳️‍🌈.

®️ also renders (somewhat) correctly as a single `\textregistered`, though the "color emoji" part becomes nonfunctional and there's a warning about missing character metrics for U+FE0F.

Fixes #3842

### DIFF
--- a/src/macros.ts
+++ b/src/macros.ts
@@ -335,7 +335,6 @@ defineMacro("\u231E", "\\llcorner");
 defineMacro("\u231F", "\\lrcorner");
 defineMacro("\u00A9", "\\copyright");
 defineMacro("\u00AE", "\\textregistered");
-defineMacro("\uFE0F", "\\textregistered");
 
 // The KaTeX fonts have corners at codepoints that don't match Unicode.
 // For MathML purposes, use the Unicode code point.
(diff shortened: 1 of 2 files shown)


grigoriy-reshetniak:
LGTM

PR #4161, opened 2026-03-09
feat: \qedsymbol and \openbox

### LINKED ISSUE #3843: Add \qedsymbol command support
Hello, I would like to end some math proofs with the nice \qedsymbol command.

But it does not seem to work, see the screen capture there.
Capture d’écran 2023-07-31 à 23 18 27

edemaine:
`\qedsymbol` and `\openbox` from amsthm were not supported

**Interlude**
Originally I planned to implement `\qedsymbol` and `\openbox` as aliases for the `\square` symbol, but (like LaTeX), working in both text and math mode (unlike `\square`).

Here's a comparison of how LaTeX renders `\openbox` versus `\square`:

red `\openbox` over green `\square`
image

red `\openbox` under green `\square`
image

Zoomed in a ton like this, there are some slight dimension differences, and perhaps more notably, `\square` has rounded corners.

Instead I tried to write `\openbox` using macros, close to how amsthm does it. Here's how it looks with the PR:

```latex
\openbox \text{ vs. } \square
```

image

Here's what texcmp renders:

Openbox

I *think* this is just rasterization error.

Fixes #3843

### DIFF
--- a/docs/support_table.md
+++ b/docs/support_table.md
@@ -803,6 +803,7 @@ use `\ce` instead|
 |\Omicron|$\Omicron$||
 |\omicron|$\omicron$||
 |\ominus|$\ominus$||
+|\openbox|$\openbox$||
 |\operatorname|$\operatorname{asin} x$|`\operatorname{asin} x`|
 |\operatorname\*|$\operatorname*{asin}\limits_y x$|`\operatorname*{asin}\limits_y x`|
 |\operatornamewithlimits|$\operatornamewithlimits{asin}\limits_y x$|`\operatornamewithlimits{asin}\limits_y x`|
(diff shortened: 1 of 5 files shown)


edemaine:
By the way, I think we could implement this simpler via `\fbox`, but not until `\fboxsep` is configurable. At 0.3em, the minimum width and height of an `\fbox` is 0.6em, whereas `\openbox` wants an inner height of 0.595em. It's pretty close. I'm not sure our `\fbox` has the right horizontal padding, because I needed the following to get a good approximation, whereas the inner kern should be more like 0:

```tex
\kern0.04889em%
\raisebox{0.34em}{\fbox{\kern .076em}}%
\kern0.04889em%
```

red `\fbox` approach stacked above/below green `\rule` approach of this PR

image

Arguably this may be close enough... Let me know what you think. But we should probably at least fix `\fbox`'s horizontal padding before switching to this system.

grigoriy-reshetniak:
I like `\fbox` approach because macros itself is simpler. Current implementation seems more fragile, as if any of those functions will _slightly_ change their behavior it may impact output of `\qedsymbol` and `\openbox`.

At the same time current implementation is working already and migration to `\fbox` could be a good candidate for a separate issue.

edemaine:
I'd also be fine with switching to `\fbox` now and fixing the fbox spacing shortly thereafter, and updating this macro as the same time. I agree it feels less fragile (worst case the box has wrong dimensions), and also less code bloat.

Issue #3366, opened 2021-10-28
Around 31% of KaTeX `<span>`s have no inner content - restricting the use of html minifier tools


universemaster:
Work is underway in https://github.com/KaTeX/KaTeX/pull/3364 that will likely improve this, but I want to document this in relation to #2194 and https://github.com/KaTeX/KaTeX/issues/3356 .

**Describe the bug:**
- In #3353 and #3344, I examine various cases of KaTeX `<span>`s.
- In both those previous issues, I have only considered spans with no attributes, or with attributes of only unstyled classes.
- Once these unnecessary `<span>`s are removed, I found that 31% of the remaining katex `<span>`s are without inner content.
- But in this case, not all are redundant (in fact, maybe all are **necessary** for vertical alignment - I haven't yet investigated further).
- I raise this issue in particular, because it means that users cannot make use the of `removeEmptyElements` feature of [html-minifier-terser](https://github.com/terser/html-minifier-terser).
- This means that other empty elements on the page cannot be removed with this tool without further effort.
- This is particularly relevant for users of KaTeX because KaTeX produces a large DOM tree, and users might want to try minifying their HTML.

**To Reproduce:**
1. First create some html from a KaTeX source, it's likely that anything with a subscript or superscript will be suitable here, as (I think) all these `<span>s` relate to vertical alignment.
2. Open the html in a browser.
3. To view these elements, run this in the browser console:

```javascript
document.querySelectorAll('.katex span').forEach(function (currentNode) {

  if (currentNode.attributes.length != 0 && currentNode.childNodes.length === 0) {
    console.log(currentNode);
  }
});
```

**Expected behavior:**
Ideally, KaTeX output should have no elements with empty inner content, so that tools like [html-minifier-terser](https://github.com/terser/html-minifier-terser) can be safely run.
I realise this may be an impossible goal, but I wanted to document the motivation.

adil14788:
can I work on this issue

Issue #3353, opened 2021-10-25
`<span></span>` can be removed from html emitted from KaTeX with no visual change (about 3% of all katex `<span>`s)


universemaster:
In #3344, I say:
> (Note that I did not remove the 4% of katex spans with no attributes at all that @ronkok indicated [here](https://github.com/KaTeX/KaTeX/issues/2194#issuecomment-581064708) as being necessary for correct rendering, especially in Safari.)

Again this relates to #2194 .

Here, I analyse these a little further.

**Describe the bug:**
- There are two types of `<span>` with no attributes.
- Namely those with, and those without text nodes (the text nodes seem to be shapes for large curly brackets?)
- The vast majority are without text nodes and are simply `<span></span>`, with no inner content.
- I removed these, and noticed no visual difference, even in (desktop) Safari (13.1.3)
- The `<span></span>` represent about 3% of all katex `<span>`s emitted.

**To Reproduce:**
1. Produce some html with KaTeX.
2. Open the html in a browser.
3. Run this in the console:

```javascript
//Let's try deleting spans with no attributes, thus no classes, and also have nochild nodes including no text nodes.  This represents about 20,000 nodes from about 16,000 katex equations.  This is a hacky method, don't use in production.
document.querySelectorAll('.katex span').forEach(function (currentNode) {
  if (currentNode.attributes.length === 0 && currentNode.childNodes.length === 0) {
    currentNode.outerHTML = currentNode.innerHTML;
  }
});
}
```

4. Wrap this in a while loop, or just run it multiple times and then check you get nothing back from a query.
5. Try to find any visual change.

**Expected behavior:**
- `<span></span>` should not be emitted.
- For performance reasons, this should be fixed at the point of generation, and not "tidied up" afterwards.

**Environment (please complete the following information):**
 - KaTeX Version: 0.13.18.

As always, I'm happy to be corrected, or told I am wrong, it's very possible that there are things I am not considering.
Note, I was not able to check in mobile safari.

PR #4164, opened 2026-03-10
reduce DOM size: strip build-time classes, unwrap and omit empty spans

### LINKED ISSUE #3344: 20% of katex spans can be removed without visual change and can significantly improve page speed and DOM size
**Describe the bug:**
- This relates to https://github.com/KaTeX/KaTeX/issues/2194 .

- In my server-side-rendered file, I removed the 20% of katex spans whose only attributes were classes that are not styled by the katex.css stylesheet.  These are described by @ronkok [here](https://github.com/KaTeX/KaTeX/issues/2194#issuecomment-584703052) as being previously used by css for horizontal spacing, but importantly, not now being used by css - and so not necessary in the final html.

- The unstyled classes (that I found with [html-inspector](https://github.com/philipwalton/html-inspector) ) are:
`'mord', 'mbin', 'mrel', 'mtight', 'mop', 'mclose', 'mopen', 'mpunct', 'minner', 'delimsizinginner', 'allowbreak'`

- (Note that I did not remove the 4% of katex spans with no attributes at all that @ronkok indicated [here](https://github.com/KaTeX/KaTeX/issues/2194#issuecomment-581064708) as being necessary for correct rendering, especially in Safari.)

- After removing these 20% of katex spans whose only attributes were these classes, I did not observe any visual changes when using the latest versions of chrome (desktop), firefox (desktop) or safari desktop (13.1.3).

- For my own purposes, I plan to always remove these spans, but can anyone tell me what visual difference I might see: what should I look for?

- I found the following significant improvements to page speed metrics on my large page of math notes (as reported by lighthouse on chrome):

| Lighthouse Metric | Decrease After Removal |
|---|---|
| Total Blocking Time (ms) | 35% |
| Parse HTML & CSS (ms) | 31% |
| Rendering (ms) | 28% |
| Maximum DOM Depth | 25% |
| Time To Interactive (s) | 22% |
| Total DOM Elements | 19% |
| Style and Layout (ms) | 18% |
| Uncompressed File Size | 17% |
| Brotli Compressed File Size | 5% |

**To Reproduce:**
1.  Generate some server side rendered html using katex (with option output: html)
2. Open the html in the browser.
3. In the browser console copy and paste this [gist](https://gist.github.com/universemaster/5c7ecd506ef5c6d59e3ede4c32217598#file-removeunstyledclassesinkatexhtml-js) .
4. Review the messages in the console (some may take several seconds to generate).
5. Try to find any visual difference.
6. To generate lighthouse scores first take a baseline with your current html.
7. Then create a new static html page with puppeteer in your build script with the contents of the gist in a [page.evaluate](https://pptr.dev/#?product=Puppeteer&version=v10.4.0&show=api-pageevaluatepagefunction-args).
8. Then use lighthouse in chrome on the new page.

**Expected behavior:**
- Emitted html from katex should be minimal and contain no content not necessary for correct rendering.
- For performance reasons, these unnecessary elements should not be removed after generation, but instead, not generated at all.

 - KaTeX Version: 0.13.18

---

- I am open to being challenged on any of this: If I have missed something or have got the wrong idea, please let me know. I massively respect the work the contributors to KaTex have put in, and think KaTex is nothing less than revolutionary.
- It's quite possible that there are visual changes that this removal produces that I have not noticed and I am happy to be told of them.

alexander-turner:
## Summary

Reduces the size of `renderToString()` output by stripping CSS classes that exist only for build-time spacing calculations and eliding wrapper/empty spans that carry no information after filtering. `toNode()` is unchanged, so CSS child selectors and existing DOM traversal code are unaffected.

Closes #3344. Addresses #3353 and a subset of #3366 / #2194.

## What changes

1. **Strip build-time-only classes from `toMarkup()`.** `mord`, `mbin`, `mrel`, `mop`, `mopen`, `mclose`, `mpunct`, `minner`, `mtight`, `nobreak`, `allowbreak`, and `delimsizinginner` are used only during the in-memory build phase (spacing, binary cancellation, line breaking) and are not referenced by any CSS rule. They now get filtered before serialization.
2. **Unwrap wrapper spans whose classes all drop out.** If a span's classes were all build-time-only and it has no styles/attributes, its children are emitted directly instead of inside a pointless `<span>` wrapper.
3. **Elide truly empty spans.** After filtering, any span that would serialize as `<span></span>` (no children, no styles, no attributes, no surviving classes) is omitted outright. `html-minifier-terser --remove-empty-elements` is now a no-op on KaTeX output.
4. **Preserve the Chrome contenteditable workaround in `makeVList`.** The intentional empty span inside `depthStrut` gains a `\u200b` child so it isn't elided (same idiom as the adjacent `topStrut`).

### `toNode()` / `toMarkup()` divergence is intentional

All of the above apply **only** to `toMarkup()`. `toNode()` keeps the full class list and every wrapper/empty span so CSS selectors that walk exact child indexes (e.g. `.mfrac > span > span`) keep working against rendered DOM. KaTeX does not support SSR hydration, so the two paths producing structurally different trees is safe. The divergence is documented inline on `Span.canUnwrap()` / `Span.canOmit()` in `src/domTree.ts`.

## Measurements

Across the 126 screenshotter fixtures (`test/screenshotter/ss_data.yaml`), rendered with `displayMode: true, output: "html"`:

| Revision | Bytes | `<span>` opens | `<span></span>` |
|---|---:|---:|---:|
| `main` | 683,802 | 14,653 | 486 |
| This PR | 627,074 | 12,910 | **0** |
| **Δ** | **−56,728 (−8.3%)** | **−1,743 (−11.9%)** | **−486** |


alexander-turner:
@greptile another review?

alexander-turner:
(Closed due to insufficient context on repo, but hopefully there are useful code fragments here for others in the future)

PR #4165, opened 2026-03-11
refactor: rework typing for wide char

grigoriy-reshetniak:
I was working on https://github.com/KaTeX/KaTeX/issues/1456 when I stumbled upon `wide-character.ts` again. I've tried to rework how it stores data and work with types, please tell me what you think of this approach. If you would find it suitable I will start using this types when I touch something like https://github.com/KaTeX/KaTeX/blob/main/src/buildCommon.ts#L135 or https://github.com/KaTeX/KaTeX/blob/main/src/buildCommon.ts#L104.

Arrays were used to look up character data.

Data structure changed, behavior should remain the same.

Requires https://github.com/KaTeX/KaTeX/pull/4160 (Edit: I've fixed it in couple iterations, but I would love to be able to check screenshots locally)


grigoriy-reshetniak:
@greptile another review?

src/types.ts
```diff
@@ -33,3 +27,35 @@ export type BreakToken = "]" | "}" | "\\endgroup" | "$" | "\\)" | "\\\\" | "\\en
...
+    ScriptRegular: "Script-Regular",
+    FrakturRegular: "Fraktur-Regular",
+    AMSRegular: "AMS-Regular",
+    SansSerifRegular: "SansSerif-Regular",
+    SansSerifBold: "SansSerif-Bold",
+    SansSerifItalic: "SansSerif-Italic",
+    TypewriterRegular: "Typewriter-Regular",
+} as const;
+
+export type FontName = (typeof FontName)[keyof typeof FontName];
```
edemaine: Maybe defining a `ValueOf` helper would be useful?

```ts
type ValueOf<T> = T[keyof T];
```

grigoriy-reshetniak:
@greptile can you do another review with score?

edemaine (commented):
I think this approach is overall a good one; I just haven't had time for a careful review. It also feels a bit repetitive, but I think it's not more repetitive than the previous system.

src/wide-character.ts
```diff
@@ -7,100 +7,97 @@
...
-    ["mathsf", "textsf", "SansSerif-Regular"],        // 0-9 sans-serif
-    ["mathboldsf", "textboldsf", "SansSerif-Bold"],   // 0-9 bold sans-serif
-    ["mathtt", "texttt", "Typewriter-Regular"],       // 0-9 monospace
-] as const satisfies readonly (readonly [string, string, string])[];
+    boldUpright,      // 0-9
+    noFont,           // 0-9 double-struck. No KaTeX font.
+    sansSerif,        // 0-9
+    boldSansSerif,   // 0-9
+    monospace,       // 0-9
+] as const;
```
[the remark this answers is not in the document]
grigoriy-reshetniak: I feel like adding `satisfies` and `readonly` seems too verbose. I've added it initially to check if features from modern TS will work properly, but when I came back to it as not an experiment, I see that this is a local variable which not modified anywhere and tends just to live there quietly. It could be typed more _properly_, but I'm not sure if benefit is worth the verbosity.

src/wide-character.ts
```diff
@@ -7,100 +7,97 @@
  */
 
 import type {Mode} from "./types";
+import {FontName, FontClass} from "./types";
 import ParseError from "./ParseError";
 
+type WideChar = {
+    readonly class: (typeof FontClass)[keyof typeof FontClass];
+    readonly font: FontName | "";
+};
```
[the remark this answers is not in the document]
grigoriy-reshetniak: This replicates original behavior and I think that empty string is valid value. I don't like `undefined` as an assigned value, I would prefer a `null` instead, but it would still require me to guard against it.

src/types.ts
grigoriy-reshetniak: It is, thank you!

src/types.ts
```diff
@@ -33,3 +29,35 @@ export type BreakToken = "]" | "}" | "\\endgroup" | "$" | "\\)" | "\\\\" | "\\en
...
+    MathItalic: "Math-Italic",
+    MainBoldItalic: "Main-BoldItalic",
+    ScriptRegular: "Script-Regular",
+    FrakturRegular: "Fraktur-Regular",
+    AMSRegular: "AMS-Regular",
+    SansSerifRegular: "SansSerif-Regular",
+    SansSerifBold: "SansSerif-Bold",
+    SansSerifItalic: "SansSerif-Italic",
+    TypewriterRegular: "Typewriter-Regular",
+} as const;
```
edemaine: I don't think we need this object. The string values seem to be just as good as the key names. We can replace the `WideCharFontName` type with a union of strings.

src/wide-character.ts
```diff
@@ -6,101 +6,98 @@
  * the font information necessary to render it properly.
  */
 
-import type {Mode} from "./types";
+import type {Mode, ValueOf} from "./types";
+import {WideCharFontName, FontClass} from "./types";
 import ParseError from "./ParseError";
 
+type WideChar = {
+    readonly class: ValueOf<typeof FontClass>;
```
edemaine:
```suggestion
    readonly class: FontClass;
```

src/wide-character.ts
```diff
@@ -6,101 +6,98 @@
...
+const boldItalic: WideChar = {class: FontClass.boldItalic, font: WideCharFontName.MainBoldItalic};
+const script: WideChar = {class: FontClass.script, font: WideCharFontName.ScriptRegular};
+const noFont: WideChar = {class: FontClass.noFont, font: ""};
+const fraktur: WideChar = {class: FontClass.fraktur, font: WideCharFontName.FrakturRegular};
+const doubleStruck: WideChar = {class: FontClass.doubleStruck, font: WideCharFontName.AMSRegular};
+const boldFraktur: WideChar = {class: FontClass.boldFraktur, font: WideCharFontName.FrakturRegular};
+const sansSerif: WideChar = {class: FontClass.sansSerif, font: WideCharFontName.SansSerifRegular};
+const boldSansSerif: WideChar = {class: FontClass.boldSansSerif, font: WideCharFontName.SansSerifBold};
+const italicSansSerif: WideChar = {class: FontClass.italicSansSerif, font: WideCharFontName.SansSerifItalic};
+const monospace: WideChar = {class: FontClass.monospace, font: WideCharFontName.TypewriterRegular};
```
edemaine: It seems these are one-to-one with `FontClass` which is used only in this file, so why not combine this into one object with three keys? I guess that's closer to the original 3-element array design, so you probably did this for a reason. Could you explain?

grigoriy-reshetniak: My idea was that class names stored separately from font names. So the `WideChar` would know only about those class names which are correspond to the current mode, not possible variations for both modes.

> so why not combine this into one object with three keys

Also output is the object with two properties (class and font) contributed too.

src/wide-character.ts
grigoriy-reshetniak: It's not the same as in [types.ts](https://github.com/KaTeX/KaTeX/pull/4165/changes#diff-c54113cf61ec99691748a3890bfbeb00e10efb3f0a76f03a0fd9ec49072e410aR63). There it's narrowed by `Mode` and result in the union of strings `"mathbf" | "textbf" | "mathnormal" | "textit" |...`, but here it's union of objects

```
  { readonly math: "mathbf"; readonly text: "textbf" } |
  { readonly math: "mathnormal"; readonly text: "textit" } |
  { readonly math: "boldsymbol"; readonly text: "boldsymbol" }...
```

src/types.ts
grigoriy-reshetniak: Done

src/wide-character.ts
edemaine: I guess I don't see the advantage of this separation. All three properties feel like they jointly describe "the font". I took a stab at implementing this alternative in https://github.com/KaTeX/KaTeX/tree/refactor-wide-char-alt so that we have a concrete option to consider. If you like it, feel free to merge into this branch.

I could see this table (currently a bunch of `const`s but could be in one object) being defined outside wide-character too, ~~though it seems they're not used elsewhere~~... actually, perhaps `boldSymbol`, `mathsym`, and especially `fontMap` could use them too, with a bit of refactoring and addition of `variant`. What do you think? For starters, I've added [a commit](https://github.com/KaTeX/KaTeX/commit/1bfce8bf93bae086f973f3c11a8a8bcf62a43f9d) to https://github.com/KaTeX/KaTeX/tree/refactor-wide-char-alt that uses `FontName` (new name for `WideCharFontName`) in `fontMetricsData.d.ts` which seems like an improvement.

grigoriy-reshetniak: > All three properties feel like they jointly describe "the font"

I thought of it more like "wide char in math mode doesn't need to know what class it has in text mode. But I like that output type is `WideChar` and class managing become simpler, so I've added this commit.

However, I'd like to have `Flesh out FontNames...` in other PR. This is exactly the direction I want to go in next PRs, but for this one :
- I want to keep this PR smaller and related to `WideChar` only.
- It will require `Omit` for `WideChar` because it doesn't support added fonts.

edemaine: Sounds good. Let's aim to unify all the font tables in future PRs!

src/wide-character.ts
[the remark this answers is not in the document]
edemaine: I think adding some `: WideChar[]` typing (or `satisfies`) would be good to ensure we don't accidentally add something invalid. Unless the length of array is important to capture in types? This applies to a bunch of declarations above.

edemaine (approved):

PR #4182, opened 2026-03-27
feat(a11y): add `role="math"` and `contrib/a11y-tabindex` for keyboard-accessible scrollable math

alexander-turner:

- The `.katex` span in HTML-only output mode has no ARIA role, so screen readers don't identify it as mathematical content.
- In HTML-only mode, `.katex-html` is marked `aria-hidden="true"` even though there is no MathML fallback, making the content invisible to screen readers.
- Scrollable (overflowing) math elements are not keyboard-focusable, failing [WCAG 2.1 SC 2.1.1](https://www.w3.org/WAI/WCAG21/Understanding/keyboard.html) and the axe [scrollable-region-focusable](https://dequeuniversity.com/rules/axe/4.4/scrollable-region-focusable) rule.

- Adds `role="math"` to the `.katex` span in HTML-only output mode (`output: "html"` and `__renderToHTMLTree`). Skipped in combined HTML+MathML mode (the `<math>` element already provides this semantic) and MathML-only mode.
- Removes the incorrect `aria-hidden="true"` from `.katex-html` in HTML-only mode so screen readers can access the rendered content.
- Adds a new `contrib/a11y-tabindex` module that uses `ResizeObserver` to detect overflowing `.katex` elements and applies `tabindex="0"` (with `role="math"` when not already set by core). A `MutationObserver` handles dynamically added math. When an element stops overflowing, the attributes are cleaned up automatically.

(Narrowed from #4162.)


alexander-turner:
@greptile another review?

src/buildTree.ts
```diff
@@ -41,12 +41,23 @@ export const buildTree = function(
         return buildMathML(tree, expression, options, settings.displayMode, true);
     } else if (settings.output === "html") {
         const htmlNode = buildHTML(tree, options);
+        // buildHTML sets aria-hidden="true" because in combined mode the
+        // visual HTML is a duplicate of the MathML.  In HTML-only mode
+        // there is no MathML, so the HTML must remain visible to screen
+        // readers.
+        delete htmlNode.attributes["aria-hidden"];
```
grigoriy-reshetniak: This attribute was set here https://github.com/KaTeX/KaTeX/blob/main/src/buildHTML.ts#L398 and now it's removed in two places. I feel like it's better to add it in single place which needs it and change `buildHTML`.

src/buildTree.ts
```diff
@@ -41,12 +41,23 @@ export const buildTree = function(
...
         katexNode = makeSpan(["katex"], [htmlNode]);
+        // HTML-only mode has no MathML <math> element to convey semantics,
+        // so we add role="math" explicitly.
+        katexNode.setAttribute("role", "math");
     } else {
         const mathMLNode = buildMathML(tree, expression, options,
             settings.displayMode, false);
         const htmlNode = buildHTML(tree, options);
         katexNode = makeSpan(["katex"], [mathMLNode, htmlNode]);
+        // In combined mode, the MathML <math> element already provides
```
grigoriy-reshetniak: I propose to remove this comment.

src/buildTree.ts
```diff
@@ -41,12 +41,23 @@ export const buildTree = function(
         return buildMathML(tree, expression, options, settings.displayMode, true);
     } else if (settings.output === "html") {
```
grigoriy-reshetniak, line 42: Changes in this branch changes output in VoiceOver from
> group

for `main` branch to 

> with 33 items, math

for `alexander-turner:claude/refocus-a11y-attributes-d8zsa` for default dev server equation. I'm not confident this is an improvement.

src/buildTree.ts
```diff
@@ -59,7 +70,11 @@ export const buildHTMLTree = function(
 ): DomSpan {
     const options = optionsFromSettings(settings);
     const htmlNode = buildHTML(tree, options);
+    // buildHTMLTree is always HTML-only — remove aria-hidden so screen
```
grigoriy-reshetniak: > buildHTMLTree is always HTML-only

As the name suggests. I propose to remove this comment.

alexander-turner:
@grigoriy-reshetniak: I removed `role="math"` from HTML-only output. 

1. **MathJax deliberately removed `role="math"` in v2.7** ([release notes](https://github.com/mathjax/MathJax/wiki/Draft-for-what's-new-page)) and never brought it back.
2. As you found, without `aria-label`, `role="math"` causes VoiceOver to announce "with N items" for internal spans — pure noise ([Apple Community thread](https://discussions.apple.com/thread/254913325)). OTOH, #4162 discussed that adding meaningful `aria-label` would require bundling `render-a11y-string` into core (+2.3KB gzipped), which is out of scope. #4183 addresses this with a console warning instead.

edemaine:
I think the `aria-hidden` change should be a separate PR. It sounds like a good idea, though I'm not an ARIA expert. Does part of the spec make this clearly correct? What does MathJax do in HTML-only mode?

`contrib/a11y-tabindex` seems a rather complicated solution. Is it really bad to include `tabindex` in nonscrollable display math? Can you find documentation either way? This solution doesn't work with server-side rendering, and I suspect it adds a rather large client-side overhead because `MutationObserver` is expensive.

alexander-turner:
@edemaine 

> I think the aria-hidden change should be a separate PR. It sounds like a good idea, though I'm not an ARIA expert. Does part of the spec make this clearly correct? What does MathJax do in HTML-only mode?

Sure. I split this into [alexander-turner:KaTeX#15](https://github.com/alexander-turner/KaTeX/pull/15) (LMK if I should turn this into a PR here). The spec is clear — [MDN's `aria-hidden` docs](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-hidden) state:

> Only when identical or equivalent meaning and functionality is exposed to assistive technologies can removing visible content from the accessibility API be considered.

And [WAI-ARIA 1.0 States and Properties](https://www.w3.org/TR/wai-aria-1.0/states_and_properties):

> Authors using aria-hidden to hide visible content from screen readers **MUST** ensure that identical or equivalent meaning and functionality is exposed to assistive technologies.

In HTML-only mode there's no MathML alternative, so `aria-hidden="true"` hides math from screen readers with no fallback.

MathJax always provides an accessible alternative alongside its visual output — in v4 via `aria-label` with [speech strings generated by the Speech Rule Engine](https://docs.mathjax.org/en/latest/basic/accessibility.html). It never ships HTML-only output without an accessible alternative.

> `contrib/a11y-tabindex` seems a rather complicated solution. Is it really bad to include `tabindex` in nonscrollable display math? Can you find documentation either way? This solution doesn't work with server-side rendering, and I suspect it adds a rather large client-side overhead because MutationObserver is expensive.

Here's what I'm thinking. For cases like SSR and because most consumers won't include an optional script, let's go with your suggestion of including `tabindex=0` for all `.katex-display` (but not `.katex`, as it apparently cannot be a scroll container). However, to precisely conform to accessibility guidelines, "all `.katex-display`" is overly broad by [`a11yproject.com`](https://www.a11yproject.com/posts/how-to-use-the-tabindex-attribute/#scrollable-overflow-containers): 

> Apply `tabindex="0"` to elements used as containers for content where the containing element includes the CSS `overflow` property. 

However, since (without the script) we can't check that, better to allow keyboard users the ability to tab through display math at all (and have a few too many tabbed display equations) than to allow no tabs. This does add extra tab stops on math-heavy pages, but that's preferable to no keyboard scrolling at all. I'll also add `role="group"` since [WCAG 4.1.2](https://www.w3.org/WAI/WCAG21/Understanding/name-role-value.html) requires focusable elements to have a role. WCAG 4.1.2 also calls for an `aria-label`, but in #4162 I explained why that's out of scope for `"html"` mode.

alexander-turner:
@edemaine Update on this?

alexander-turner:
@edemaine Circling back here :)

grigoriy-reshetniak:
Sorry for delay @alexander-turner.

For `aria-hidden` I suggest to move this part in separate PR as @edemaine suggested. This could serve as intermediate solution while we decide how to improve SR output in https://github.com/KaTeX/KaTeX/issues/820.

For `contrib/a11y-tabindex` it feels a bit like using a sledgehammer to crack a nut. If I use VoiceOver I can navigate via math blocks using `CapsLock` + `->`. Does it really needs `tabindex`? Even if it does, does it overflow on it's own, or this should it be responsibility of container above `katex` and therefore KaTeX consumers? It seems like demo of problem could help to address it from different perspectives.

PR #4198, opened 2026-04-20
Siunitx compatible extension

### LINKED ISSUE #817: per

krtko1:
- **siunitx: add most of the functionality**
- **siunitx: implement declare commands**
- **siunitx: allow literal units**
- **external-siunitx: move siunitx outside of core**

Fixes #817

### DIFF
--- a/contrib/siunitx/README.md
+++ b/contrib/siunitx/README.md
@@ -1,0 +1,36 @@
+# siunitx extension
+
+This extension adds siunitx-compatible commands to KaTeX, including:
+`\sisetup`, `\num`, `\si`, `\unit`, `\SI`, `\qty`, `\ang`, and related list/range forms.
+
+## Usage
+
+This extension is not part of core KaTeX and must be loaded separately.
+
+### Browser
+
+Load it after `katex.js`:
+
+```html
+<script defer src="https://cdn.jsdelivr.net/npm/katex@0.16.45/dist/katex.min.js" crossorigin="anonymous"></script>
+<script defer src="https://cdn.jsdelivr.net/npm/katex@0.16.45/dist/contrib/siunitx.min.js" crossorigin="anonymous"></script>
+```
+
+### Node / bundlers
+
+```js
+const katex = require("katex");
+require("katex/contrib/siunitx");
+```
+
+Or with ESM:
+
+```js
+import katex from "katex";
+import "katex/contrib/siunitx";
+```
+
+## Notes
+
+- The `siunitx` render option (for example `siunitx: "group-separator={,}"`) requires this extension to be loaded first.
+- The extension registers its commands on the KaTeX instance when imported.
--- a/contrib/siunitx/README.md
+++ b/contrib/siunitx/README.md
@@ -34,3 +34,4 @@ import "katex/contrib/siunitx";
 
 - The `siunitx` render option (for example `siunitx: "group-separator={,}"`) requires this extension to be loaded first.
 - The extension registers its commands on the KaTeX instance when imported.
+- KaTeX exposes `__siunitxInternals` so this extension can register into core internals. This is an unstable internal bridge for `contrib/siunitx` only, not a public/general extension API.
--- a/contrib/siunitx/register.ts
+++ b/contrib/siunitx/register.ts
@@ -1,14 +1,29 @@
-import defineFunction from "../defineFunction";
-import {makeSpan} from "../buildCommon";
-import {MathNode} from "../mathMLTree";
-import ParseError from "../ParseError";
+import katex from "katex";
 
-import * as html from "../buildHTML";
-import * as mml from "../buildMathML";
+import type Options from "../../src/Options";
+import type Parser from "../../src/Parser";
+import type {AnyParseNode, ParseNode} from "../../src/parseNode";
 
-import type Options from "../Options";
-import type Parser from "../Parser";
-import type {AnyParseNode, ParseNode} from "../parseNode";
+type KaTeXWithSiunitxInternals = typeof katex & {
+    __siunitxInternals?: {
+        defineFunction: typeof import("../../src/defineFunction").default;
+        makeSpan: typeof import("../../src/buildCommon").makeSpan;
+        MathNode: typeof import("../../src/mathMLTree").MathNode;
+        html: typeof import("../../src/buildHTML");
+        mml: typeof import("../../src/buildMathML");
+        ParseError: typeof import("../../src/ParseError").default;
+    };
+};
+
+const k = katex as KaTeXWithSiunitxInternals;
+if (!k.__siunitxInternals) {
+    throw new Error(
+        "KaTeX siunitx internals are unavailable. " +
+        "Load the matching core KaTeX build first.",
+    );
+}
+
+const {defineFunction, makeSpan, MathNode, html, mml, ParseError} = k.__siunitxInternals;
 
 /**
  * Practical siunitx-compatible support for:
@@ -4037,15 +4052,15 @@ const textToMathNodesWithMode = (
             }
         }
         if (SUPERSCRIPT_REVERSE_MAP[text[i]]) {
-            const last = nodes[nodes.length - 1];
-            if (last && last.type === "supsub") {
-                let j = i;
-                let exponent = "";
-                while (j < text.length && SUPERSCRIPT_REVERSE_MAP[text[j]]) {
-                    exponent += SUPERSCRIPT_REVERSE_MAP[text[j]];
-                    j++;
-                }
-                if (exponent) {
+            let j = i;
+            let exponent = "";
+            while (j < text.length && SUPERSCRIPT_REVERSE_MAP[text[j]]) {
+                exponent += SUPERSCRIPT_REVERSE_MAP[text[j]];
+                j++;
+            }
+            if (exponent) {
+                const last = nodes[nodes.length - 1];
+                if (last && last.type === "supsub") {
                     const supGroup: ParseNode<"ordgroup"> = last.sup &&
                             last.sup.type === "ordgroup"
                         ? last.sup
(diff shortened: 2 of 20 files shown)


krtko1:
@edemaine Is there chance that you can look at this? Mainly interface between siunitx and katex should be checked. Even if it doesn't get merged I would like some opinion or guidance. Thanks.

test/siunitx-spec.ts
```diff
@@ -0,0 +1,677 @@
+/* eslint max-len:0 */
+
+import "../contrib/siunitx/siunitx";
```
edemaine, line 3: Can we make this modification apply just to these tests? Perhaps by `register`ing on a deep copy of `katex`.

src/parseTree.ts
```diff
@@ -20,16 +52,37 @@ const parseTree = function(
...
+            "The `siunitx` option requires loading `katex/contrib/siunitx` first.",
+        );
+    }
+    if (settings.siunitx && !hasBalancedSiunitxBraces(settings.siunitx)) {
+        throw new ParseError(
+            "Invalid `siunitx` option: unbalanced braces in settings.siunitx.",
+        );
+    }
+    const expression = settings.siunitx
+        ? `\\sisetup{${settings.siunitx}}${toParse as string}`
```
edemaine, line 66: This ends up shifting all column numbers in a way that's counterintuitive to the end-user receiving errors.

Instead of an siunitx-specific configuration, it seems like an option (maybe `preamble`?) that gives arbitrary KaTeX code that prefixes the input would be more generally helpful (especially when using auto-render).

Actually, I'm a little confused why you wouldn't just use `katex.renderToString("\\sisetup{...}", {globalGroup: true})` to set global settings like this. Wouldn't this be much more efficient? And then we could just document this approach in the extension.

contrib/siunitx/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+This extension is not part of core KaTeX and must be loaded separately.
+
+### Browser
+
+Load it after `katex.js`:
+
+```html
+<script defer src="https://cdn.jsdelivr.net/npm/katex@0.16.45/dist/katex.min.js" crossorigin="anonymous"></script>
+<script defer src="https://cdn.jsdelivr.net/npm/katex@0.16.45/dist/contrib/siunitx.min.js" crossorigin="anonymous"></script>
```
edemaine, line 16: These should include `integrity` so that our release scripts will automatically update them.

docs/browser.md
```diff
@@ -127,3 +127,15 @@ for more details.
 project. Then bundle using bundlers like [webpack](https://webpack.js.org/) or
 [rollup.js](https://rollupjs.org/). Note that you have to bundle the stylesheet
 (`katex.css`) or include it manually.
+
+## Using siunitx extension
```
edemaine, line 131: Not sure this belongs here; other extensions don't get this treatment. Maybe it makes sense to link from here to `extensions.md`, and then that file should be updated to link to the new extension. We could include a general "using extensions" that mentions loading after KaTeX, but then maybe include an example that loads all relevant extensions (mhchem + siunitx).

contrib/siunitx/siunitx.ts
```diff
@@ -0,0 +1,6 @@
+import katex from "katex";
+import {registerSiunitx} from "./register";
+
+registerSiunitx();
```
edemaine, line 4: I find the file naming confusing here. I would more naturally let `siunitx.ts` have the main code, and `register.ts` do the registration.

I also wonder whether `registerSiunitx` should take `katex` as an argument, instead of importing directly. This would be somewhat similar to how `mhchem.js` works, which uses an implicit `katex` global. Have you tested your setup in a web environment, or only on Node?

src/parseNode.ts
```diff
@@ -431,6 +430,109 @@ type ParseNodeTypes = {
         width: Measurement;
         height: Measurement;
     };
+    "siunitx": {
```
edemaine, line 433: Can we do these type additions within `contrib/siunitx` instead? I'd rather not put them in the base code.

katex.ts
```diff
@@ -152,6 +155,19 @@ const __domTree = {
...
+// same core function registry instance, but it is NOT a supported extension API.
+// Treat as unstable internals; shape/behavior may change without notice.
+const __siunitxInternals = {
+    defineFunction,
+    makeSpan,
+    MathNode,
+    html,
+    mml,
+    ParseError,
+};
```
edemaine, line 169: I think we need to consider how to best export these. I'd be more tempted to export what we need with double underscore names. Also I believe `defineFunction` and `ParseError` are already exported. If more are helpful, we should just export them, with the understanding that the APIs are unstable.

src/parseTree.ts
```diff
@@ -6,10 +6,42 @@
 import Parser from "./Parser";
 import ParseError from "./ParseError";
 import {Token} from "./Token";
+import functions from "./functions";
 
 import type Settings from "./Settings";
 import type {AnyParseNode} from "./parseNode";
 
+const hasBalancedSiunitxBraces = (siunitxSettings: string): boolean => {
```
edemaine, line 14: This should be omitted.

contrib/siunitx/register.ts
```diff
@@ -0,0 +1,5373 @@
...
+
+const splitListItems = (src: string): string[] => {
+    const parts = src.split(";").map((item) => item.trim());
+    if (parts.length <= 1) {
+        return src.trim() ? [src.trim()] : [];
+    }
+    return parts.filter(Boolean);
+};
+
+const unwrapBraces = (value: string): string => {
```
edemaine, line 315: I'm not sure `raw` arguments followed by manual unwrapping etc. is the right approach. This probably forbids macro expansion inside arguments, which I assume siunitx supports. Did you try a more token-by-token approach?

contrib/siunitx/register.ts
```diff
@@ -0,0 +1,5373 @@
...
+    let out = value;
+    // Convert common TeX-ish spacing/markup commands used in siunitx setup.
+    out = out.replace(/\\allowbreak/g, "");
+    out = out.replace(/\\mskip\+?3mu\\relax/g, " ");
+    out = out.replace(/\\ensuremath\{([^}]*)\}/g, "$1");
+    out = out.replace(/\\text\{([^}]*)\}/g, "$1");
+    out = out.replace(/\\mathrm\{([^}]*)\}/g, "$1");
+    out = out.replace(/\\cdot/g, "⋅");
+    out = out.replace(/\\times/g, "×");
+    out = out.replace(/\\,/g, " ");
```
edemaine, line 354: Is this stuff that siunitx does?? I find it unlikely... which is a sign we shouldn't be doing it.

contrib/siunitx/register.ts
```diff
@@ -0,0 +1,5373 @@
...
+    "3": "³",
+    "4": "⁴",
+    "5": "⁵",
+    "6": "⁶",
+    "7": "⁷",
+    "8": "⁸",
+    "9": "⁹",
+    "+": "⁺",
+    "-": "⁻",
+};
```
edemaine, line 987: Why are we converting to Unicode?

contrib/siunitx/register.ts
```diff
@@ -0,0 +1,5373 @@
...
+
+    defineFunction({
+        type: "siunitx",
+        names: ["\\sisetup"],
+        props: {
+            numArgs: 1,
+            argTypes: ["raw"],
+            allowedInText: true,
+            primitive: true,
+        },
```
edemaine, line 4665: The API for `defineFunction` changed in the latest KaTeX; the `props` wrapper must now be removed.

Did you try to implement these using `defineMacro`? I would have that would be a better fit, and probably involve less use of KaTeX internals; the given `parser` ought to have what you need to implement like siunitx.sty does.

test/siunitx-spec.ts
```diff
@@ -0,0 +1,677 @@
...
+import katexOrig from "../katex";
+import {r} from "./helpers";
+
+const katex: any = katexOrig;
+
+describe("siunitx-compatible commands", function() {
```
edemaine, line 9: I asked GPT 5.5 to review the tests. What’s good: `test/siunitx-spec.ts` has broad smoke coverage across the command surface, many option combinations, MathML output, declarations, fraction modes, expression evaluation, uncertainty formatting, and a few integration cases around `settings.siunitx`.

Main gaps:

- Many tests are `toParse()` / `toBuild()` smoke tests. Those catch crashes, not semantic correctness.
- There is no oracle comparison against LaTeX/siunitx output or a curated compatibility matrix, so “compatible” is mostly asserted by selected examples.
  - [human again] This is perhaps a big point; I'm not sure the cleanest way to ensure this other than running latex tests and documenting in comments.
- Browser/CDN/package-loading behavior is only lightly covered through build verification, not runtime UMD/script-tag behavior.
- State/scoping behavior deserves more targeted tests: `\sisetup` grouping, `\DeclareSI*` persistence, cross-render behavior with shared `macros`, and interaction with `globalGroup`.
- Error behavior needs tests, especially `ParseError.position` with `settings.siunitx`, malformed options, malformed units, and `throwOnError: false`.
- The extension is 5k+ lines; coverage is broad but still example-driven, not systematic over parser/tokenizer edge cases.

edemaine (commented):
Thanks for working on this, and sorry for the delay in looking at it. Overall, I think there's room for an siunitx extension, though it should definitely be an extension. Partly this is an interesting test-case for whether our API suffices for extension making.

This is a *lot* of code, so I've just done a high-level pass so far, with various comments throughout. I do think that the integration should be weaker (e.g. no dedicated option). I could also see it becoming stronger, e.g., offering a `parseNumber` helper in `Parser`. (But I haven't read in detail to see exactly what you need.)

I also have a question about approach. `defineFunction` is one way, but `defineMacro` might be simpler. Did you try that? I also question whether 5k lines are necessary, though for comparison `siunitx.sty` is 10k lines. Still, I wonder whether the code could be cleaned up some.

Happy to (eventually) look at it more once you respond to these.

PR #4200, opened 2026-04-23
refactor: improve typing for fonts

grigoriy-reshetniak:
This PR is successor of https://github.com/KaTeX/KaTeX/pull/4165. It targets type system for fonts and characters.


src/buildCommon.ts
```diff
@@ -672,13 +678,13 @@ const retrieveTextFontName = function(
         fontStylesName = "BoldItalic";
     } else if (fontWeight === "textbf") {
         fontStylesName = "Bold";
-    } else if (fontWeight === "textit") {
+    } else if (fontShape === "textit") {
```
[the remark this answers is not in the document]
grigoriy-reshetniak: I've updated this part and test are green, but I haven't checked why it was working before.

src/delimiter.ts
```diff
@@ -129,7 +129,7 @@ const mathrmSize = function(
     mode: Mode,
     options: Options,
 ): SymbolNode {
-    return makeSymbol(value, "Size" + size + "-Regular",
+    return makeSymbol(value, "Size" + size + "-Regular" as FontName,
         mode, options);
```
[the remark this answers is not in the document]
grigoriy-reshetniak: I went with more strict type for delimiter sizes.

src/symbols.ts
```diff
@@ -32,9 +58,9 @@ export default symbols;
 /** `acceptUnicodeChar = true` is only applicable if `replace` is set. */
 export function defineSymbol(
     mode: Mode,
-    font: Font,
+    font: SymbolFont,
     group: Group,
```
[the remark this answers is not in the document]
grigoriy-reshetniak: It was an artifact from migration to TS and I feel like it's safe, but would like to hear other opinions.

src/buildCommon.ts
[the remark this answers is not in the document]
edemaine: Maybe it "worked" because we generally only use vertical metrics (height and depth) for layout, and italics might not differ (much) in that respect. And/or we didn't have tests that used `\textit` in a context where vertical matters.

It looks like `i` and `j` have the maximum height difference between Main-Regular (0.66786) and Main-Italic (0.65536) font metrics. And I can indeed see a difference in layout when it matters:

```tex
\overline{\textit{i}}
```

| Before PR | After PR | Overlay |
|---|---|---|
| before | after | before_after_overlay |

The overline is a bit lower after the PR. (At this scale, the image is 2 pixels shorter.)

src/functions/font.ts
```diff
@@ -101,12 +102,11 @@ defineFunction({
     handler: ({parser, funcName, breakOnTokenText}, args) => {
         const {mode} = parser;
         const body = parser.parseExpression(true, breakOnTokenText);
-        const style = `math${funcName.slice(1)}`;
 
         return {
             type: "font",
             mode: mode,
-            font: style,
+            font: `math${funcName.slice(1)}` as Exclude<MathFont, "">,
```
edemaine: It's too bad `String.prototype.slice` doesn't have better type behavior. If it did, we could remove the type cast by typing `handler` to guarantee `funcName` is among the `names` (which should be possibly by tweaking `defineFunction`'s types, I think). As it is, probably not worth it. Just mentioning as a possible future idea.

src/types/fonts.ts
```diff
@@ -0,0 +1,73 @@
...
+    | "SansSerif-Italic"
+    | "Script-Regular"
+    | "Size1-Regular"
+    | "Size2-Regular"
+    | "Size3-Regular"
+    | "Size4-Regular"
+    | "Typewriter-Regular";
+
+//[depth, height, italic, skew, width]
+export type CharacterMetricsTuple = [number, number, number, number, number];
```
edemaine:
```suggestion
export type CharacterMetricsTuple = [depth: number, height: number, italic: number, skew: number, width: number];
```

This allows TypeScript to give better hints, along the lines of your comment (which is probably no longer necessary).

src/buildCommon.ts
```diff
@@ -168,13 +174,13 @@ export const makeOrd = function<NODETYPE extends "spacing" | "mathord" | "textor
...
     const text = group.text;
-
     const classes = ["mord"];
+    const {font, fontFamily, fontWeight, fontShape} = options;
 
     // Math mode or Old font (i.e. \rm)
-    const isFont = mode === "math" || (mode === "text" && options.font);
-    const fontOrFamily = isFont ? options.font : options.fontFamily;
-    let wideFontName = "";
+    const useMathFont = mode === "math" || (mode === "text" && !!font);
```
edemaine: `useMathFont` seems like the wrong (re)name here, whereas the comment is accurate. Maybe `useFont` is what you mean?

src/buildCommon.ts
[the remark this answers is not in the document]
edemaine: P.S. Great to see typing fix a bug!!

src/buildCommon.ts
```diff
@@ -644,11 +650,11 @@ export const makeGlue = (measurement: Measurement, options: Options): DomSpan =>
...
-    fontFamily: string,
-    fontWeight: string,
-    fontShape: string,
-): string {
+const retrieveTextFontName = (
+    fontFamily: TextFont,
+    fontWeight: FontWeight,
+    fontShape: FontShape,
+): FontName => {
     let baseFontName = "";
```
edemaine:
```suggestion
    let baseFontName: "AMS" | "Main" | "SansSerif" | "Typewriter" | TextFont;
```

src/buildCommon.ts
edemaine:
```suggestion
    let fontStylesName: "BoldItalic" | "Bold" | "Italic" | "Regular";
```

src/buildCommon.ts
```diff
@@ -672,13 +678,13 @@ const retrieveTextFontName = function(
...
         fontStylesName = "Bold";
-    } else if (fontWeight === "textit") {
+    } else if (fontShape === "textit") {
         fontStylesName = "Italic";
     } else {
         fontStylesName = "Regular";
     }
 
-    return `${baseFontName}-${fontStylesName}`;
+    return `${baseFontName}-${fontStylesName}` as FontName;
```
edemaine, line 688: With the above typing, this typecast can *almost* be removed, except that `TextFont` allows `""` which makes an invalid font name `"-Foo"`. This suggests that the `retrieveTextFontName` parameter type should be narrowed. Should `TextFont` exclude `""` and only add it when needed, like you already do with `FontName`? Or should we use `Exclude<TextFont, "">`?

src/symbols.ts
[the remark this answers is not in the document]
edemaine: I'm less sure about this change; we used `null` below and you changed it to `""`. There are other packages out there that use `defineSymbol` so maybe it's worth keeping the `null` compatibility? Actually, looking at how it was used, `""` is a more natural replacement value... So I'm fine either way.

src/wide-character.ts
```diff
@@ -6,13 +6,25 @@
...
+        | Exclude<
+              FontName,
+              | "Caligraphic-Regular"
+              | "Main-Italic"
+              | "Main-Regular"
+              | "Math-BoldItalic"
+              | "Size1-Regular"
+              | "Size2-Regular"
+              | "Size3-Regular"
+              | "Size4-Regular"
```
edemaine: I feel like a positive list would make more sense here... Imagine we add a new font like small-caps.

edemaine (commented):
Lots of nice cleanup here!

src/functions/font.ts
grigoriy-reshetniak: Yeah, I've thought about it too, but after realizing  `slice` will turn it into string anyway I end up casting. I could make a map for it, but it felt too much of an overhead.

edemaine: I guess we could do a somewhat more "obviously correct" type cast like

```ts
type Slice1<S extends string> =
  S extends `${string}${infer Rest}` ? Rest : "";

foo.slice(1) as Slice1<S>;

// or if we want a helper, but maybe too much...
function slice1<S extends string>(s: S): Slice1<S> {
  return s.slice(1) as Slice1<S>;
}

File: src/symbols.ts (hunk 3 above)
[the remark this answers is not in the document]
Comment (grigoriy-reshetniak) [reply src/symbols.ts:43]: I've considered keeping `null`, but there would be no difference in runtime. `defineSymbol` is not typed in `katex.d.ts` so consumers wont notice this difference. I think it shouldn't be typed _yet_, migration to TS was recently and interfaces may change.

File: src/buildCommon.ts (hunk 8 above)
Comment (grigoriy-reshetniak) [reply src/buildCommon.ts:688]: I would prefer `Exclude<TextFont, "">`. 

`FontName` is not used in `Options`, it has to include `""` only while dealing with `WideChar` while representing absents of data. But for `TextFont` empty string is default value in `Options` context. 

Also, exluding `""` from `TextFont` would require more `TextFont |  ""` than `Exclude<TextFont, "">` and including `""` into `FontName` would require more `Exclude<FontName, "">` than `FontName |  ""`.

Comment (grigoriy-reshetniak) [reply src/buildCommon.ts:688]: Ah, unfortunately, this is what I've got
> Type "Typewriter-BoldItalic" is not assignable to type FontName

Shall I make a lookup table or keep cast for now?

Edit: I've checked additionally and it seems that `Exclude<TextFont, "">` wont be correct, because this parameter comes from options where it could be `""`.

File: src/buildCommon.ts (hunk 6 above)
Comment (grigoriy-reshetniak) [reply src/buildCommon.ts]: Per my understanding what `isFont` meant to be is:
> Should the math font (`options.font`) be used rather than the text font family (`options.fontFamily`)?

Is this incorrect?

Comment (edemaine) [reply src/buildCommon.ts]: Hmm, I didn't realize that `options.font` is only used for math fonts and `options.fontFamily` is only used for text fonts, but the new typings do make that clear. In my mind, the difference is something else (`font` replaces everything about the font, while `fontFamily` replaces just the rm/it/etc family), but it happens to also be the case that only text mode font commands have the "family switching" commands. Anyway, it's a bit weird to call it "mathFont" here when the rest of the naming is font vs fontFamily. So how about `useFont`?

File: src/buildCommon.ts (hunk 8 above)
Comment (edemaine) [reply src/buildCommon.ts:688]: Keeping a cast for now sounds good.

If `""` can be passed in, isn't that a problem? Won't this function return an invalid font? Or do we not get to the last line in this case?

I'm still not super happy with `TextFont` including `""` while `FontName` excludes it. But maybe there's a good reason?

Comment (grigoriy-reshetniak) [reply src/buildCommon.ts:688]: > If "" can be passed in

Empty string is a valid value in `Options`, but `retrieveTextFontName` it's called only in `makeOrd` in a way that guards against `""` being passed as value. I guess correct way to adequately represent this would require refactoring of `makeOrd` which I would prefer to do in other round.

`TextFont` is a value from `Options` where `""` is a default value. Same behavior applies to `FontWeight`, `FontShape`, and `MathFont`. 

`FontName` doesn't have empty string because it's the internal representation of what font names are available and extended only for `WideChar`, where absence of value is possible.

Maybe the `TextFont` and `MathFont` are just incorrect names for possible options passed and changing that would help?

Comment (edemaine) [reply src/buildCommon.ts:688]: Hmm, could we type `Options` so that `fontFamily` is forbidden (forced `""`) when font exists (not `""`)? Or is that not true? Anyway, fine to leave this for the future.

Comment (grigoriy-reshetniak) [reply src/buildCommon.ts:688]: This is true, but afaik discriminated unions are not allowed for class properties.

I would consider refactor `Options` into a function rather than class, but in future.

File: src/functions/font.ts (hunk 4 above)
Comment (grigoriy-reshetniak) [reply src/functions/font.ts]: `funcName` is `string` (coming from `FunctionContext`), so for `funcName.slice(1) as Slice1<typeof funcName>` generic value would be `string`. Because of that TS considers possibility of `funcName.slice(1)` being empty string and checks are broken downstream. It's fixable with code like

```
type OldFontCommands = "\\rm" | "\\sf" | "\\tt" | "\\bf" | "\\it" | "\\cal";
...
font: `math${funcName.slice(1) as Slice1<OldFontCommands>}`,
```

but I'm not sure it's worth it. If you think that this is an improvement -- please let me know and I'll fix it here an above.

edemaine: I think it might be worth it. We might have use for `Slice1` elsewhere too, given all the `\`s.

src/buildCommon.ts
edemaine: Agreed that we should push this to the future. I'm not sure I want to move away from classes.

src/functions/font.ts
grigoriy-reshetniak: Updated!

src/functions/font.ts
edemaine: Can we add `satisfies FontCommands | keyof typeof fontAliases` here, to ensure the two lists are the same? Type of `fontAliases` would need to change...

edemaine (approved):

src/functions/font.ts
grigoriy-reshetniak: Done!

PR #4201, opened 2026-05-01
chore: update stylelint and adjust styles for updated set of rules

grigoriy-reshetniak:
This PR updates stylelint and related dependency. Unsupported rules were removed, code was adjusted to the new rules. Some minor cleanup was done.

Partially replaces https://github.com/KaTeX/KaTeX/pull/3555


src/styles/katex.scss
```diff
@@ -49,7 +49,7 @@ $display-margin: 1em 0 !default;
         /* Accessibility hack to only show to screen readers
          Found at: http://a11yproject.com/posts/how-to-hide-content/ */
         position: absolute;
-        clip: rect(1px, 1px, 1px, 1px);
+        clip-path: inset(50%);
```
edemaine, line 52: Have you tested that this has the same accessibility behavior? (I.e., it's still visible to screen readers, but invisible on the page?)

src/styles/katex.scss
```diff
@@ -349,7 +349,9 @@ $display-margin: 1em 0 !default;
         > .root {
             /* These values are taken from the definition of `\r@@t`,
              `\mkern 5mu` and `\mkern -10mu`. */
+            /* stylelint-disable-next-line declaration-property-value-no-unknown */
             margin-left: calc(5*$mu);
+            /* stylelint-disable-next-line declaration-property-value-no-unknown */
             margin-right: calc(-10*$mu);
```
edemaine, line 355: Why do these (but not other uses of `$mu`) trigger [declaration-property-value-no-unknown](https://stylelint.io/user-guide/rules/declaration-property-value-no-unknown/)? I guess because it uses SCSS variables? I see that the docs recommend disabling this rule for SCSS.

.stylelintrc
```diff
@@ -6,7 +6,6 @@
     "rules": {
         "at-rule-no-vendor-prefix": null,
         "font-family-name-quotes": "always-where-required",
-        "indentation": 4,
```
edemaine, line 9: Is it worth adding the [stylelint-stylistic](https://github.com/stylelint-stylistic/stylelint-stylistic) plugin so that we can enforce style? It would probably be good for enforcing style of outside contributions.

edemaine (commented):
Looks good to me - just a few questions/ideas.

src/styles/katex.scss
grigoriy-reshetniak: The behavior is the same.

src/styles/katex.scss
grigoriy-reshetniak: Because these are nested `calc()`. Also, there are no other uses of `$mu`, `$muspace`, `$thinspace`, etc. — these are constants. I could inline `$mu` here, but if the `$mu` declaration is removed, the comments for the mentioned variables would need to be adjusted to make them clearer.

.stylelintrc
grigoriy-reshetniak: Yes, but I think I'll handle this in Prettier upgrade. Stylelint team [mentioned it](https://stylelint.io/migration-guide/to-15) as a reason they deprecated this rule.
Would you prefer a plugin instead?

edemaine: Oh, I thought the plan was to remove prettier. But eslint may also handle this... Either way, happy to leave this to a future PR.

edemaine (approved):

.stylelintrc
grigoriy-reshetniak: > remove prettier

I'm considering this, but not sure yet.

PR #4205, opened 2026-05-03
chore: update linters

grigoriy-reshetniak:
This PR handles linters update and cleanup.

Changelog:

- @stylistic/indentation added to handle `indentation`, because Prettier was removed.
- Prettier was removed. I've decided to do it, because it was never used as a part of CI, and only served a source of VS Code extension which would (possibly) work on editor actions. But I'm not sure it was actually used, because it either replaced by or contradicted eslint config. Prettier rules were:
    - `tabWidth` -- covered by `indent-legacy`
    - `semi`-- covered by `semi`
    - `bracketSpacing` -- covered by `object-curly-spacing`
    - `trailingComma` -- conflicted with eslint, eslint definition remains.
    - `singleQuote` -- no equivalent rule in eslint. This is problematic, because if I enable it in eslint, I get 1168 errors. I propose to handle this in dedicated PR.
- Eslint was migrated. This led to some code changes, most notably ones with removing unused default values because of `no-useless-assignment`.
- `eslint-plugin-import` was removed.`import/no-unresolved`, `import/named`, `import/namespace`, `import/namespace`, `import/namespace`, `import/default`, `import/export ` -- TS will cover this for `.ts` files, but leftover `.js` files might be impacted. `import/no-named-as-default`, `import/no-named-as-default-member`, `import/no-duplicates` were disabled anyway.
- Some dependencies around module resolutions are no longer needed.
- I've migrated comments even if they seems outdated, I propose to handle it in same PR as double quotes.

Replaces https://github.com/KaTeX/KaTeX/pull/3555


src/Settings.ts
```diff
@@ -94,7 +94,7 @@ type SchemaMetadata<K extends keyof SettingsOptions> = {
      * more information. Signature varies per setting (e.g. parseFloat,
      * or (def, defs) => defs.push(def) for macros).
      */
-    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    // @typescript-eslint/no-explicit-any
     cliProcessor?: (...args: any[]) => SettingsValue;
```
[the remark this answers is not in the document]
grigoriy-reshetniak: Updated Eslint warns against this:
> ESLint: Unused eslint-disable directive (no problems were reported from '@typescript-eslint/no-explicit-any').

This is the reason this part of comment was removed.

contrib/mathtex-script-type/mathtex-script-type.js
```diff
@@ -14,8 +14,8 @@ scripts.forEach(function(script) {
                               display ? "equation" : "inline-equation");
     try {
         katex.render(script.text, katexElement, {displayMode: display});
+        // eslint-disable-next-line no-unused-vars
     } catch (err) {
-        //console.error(err); linter doesn't like this
```
edemaine, line 18: It seems like we should be outputting the error...?

dockers/screenshotter/screenshotter.js
```diff
@@ -129,6 +129,7 @@ function isDockerDesktop() {
     try {
         const operatingSystem = cmd("docker", "info", "-f", "{{.OperatingSystem}}");
         return /Docker Desktop/i.test(operatingSystem);
+        // eslint-disable-next-line no-unused-vars
     } catch (e) {
```
edemaine: I believe `} catch {` is valid JS.

src/Settings.ts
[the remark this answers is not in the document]
edemaine: Maybe just remove the comment altogether? I see we're ignoring rest arguments in config.

contrib/mathtex-script-type/mathtex-script-type.js
grigoriy-reshetniak: Yes, I've found out original comment for linter's complain wasn't related to `no-console`. It's fixed.

edemaine:
@greptileai revise your review

test/katex-spec.ts
```diff
@@ -4298,7 +4298,7 @@ describe("Unicode", function() {
         wideCharText += String.fromCharCode(0xD835, 0xDC00);   // bold A
         wideCharText += String.fromCharCode(0xD835, 0xDC68);   // bold italic A
         wideCharText += String.fromCharCode(0xD835, 0xDD04);   // Fraktur A
-        wideCharStr += String.fromCharCode(0xD835, 0xDD6C);    // bold Fraktur A
+        wideCharText += String.fromCharCode(0xD835, 0xDD6C);    // bold Fraktur A
```
edemaine, line 4301: Nice to see a bug found via linting!

edemaine (approved):

PR #4218, opened 2026-05-16
Fix \htmlData not allowing commas in values

### LINKED ISSUE #4204: htmlData values cannot contain commas
> Before reporting a bug

**Describe the bug:**

**To Reproduce:**

```tex
% this breaks

% this works
```

**Expected behavior:**

 - KaTeX Version: 0.16.45

lloyd-c137:
## Summary

Fixes #4204.

The `\htmlData` command splits its key=value pairs by commas, but values containing commas (e.g. `[a,b]`) would be incorrectly split and cause a confusing parse error.

## Changes

**`src/functions/html.ts`**
- Replaced naive `value.split(",")` with a manual parser that handles backslash escaping
- Commas can now be escaped with `\,` to include them literally in values
- Empty entries from trailing data are skipped gracefully
- Backward compatible: all existing usage without commas continues to work unchanged

**Test URLs**
Before: `\htmlData{annotation_text=[a,b] denotes...}{[a, b]}` → Parse Error
After: `\htmlData{annotation_text=[a,b] denotes...}{[a, b]}` → Works
Escape: `\htmlData{text=comma\,separated}{content}` → Works

## Validation
- `yarn test:jest` — 1287 tests passed, 125 snapshots (no changes needed) ✅
- `yarn test:ts` ✅
- `yarn test:lint` ✅

### DIFF
--- a/src/functions/html.ts
+++ b/src/functions/html.ts
@@ -52,8 +52,6 @@ defineFunction({
             case "\\htmlData": {
                 // Split on unescaped commas only. Backslash escapes the
                 // following character so it is included literally.
-                // This avoids the need for negative lookbehind, which has
-                // limited Safari support (16.4+).
                 const data = [];
                 let current = "";
                 let escaped = false;
--- a/src/functions/html.ts
+++ b/src/functions/html.ts
@@ -50,16 +50,32 @@ defineFunction({
                 };
                 break;
             case "\\htmlData": {
-                // Split on unescaped commas using negative lookbehind.
-                // Escaped commas (\,) become literal commas in the value.
-                // Negative lookbehind is well-supported in modern browsers
-                // (Chrome 62+, Firefox 78+, Safari 16.4+, Node 8.10+).
-                const data = value.split(/(?<!\\),/)
-                    .map(s => s.replace(/\\,/g, ","));
-                for (const item of data) {
-                    if (item.length === 0) {
-                        continue;
+                // Split on unescaped commas only. Backslash escapes the
+                // following character so it is included literally.
+                // This avoids the need for negative lookbehind, which has
+                // limited Safari support (16.4+).
+                const data = [];
+                let current = "";
+                let escaped = false;
+                for (const char of value) {
+                    if (escaped) {
+                        current += char;
+                        escaped = false;
+                    } else if (char === "\\") {
+                        escaped = true;
+                    } else if (char === ",") {
+                        data.push(current);
+                        current = "";
+                    } else {
+                        current += char;
                     }
+                }
+                if (escaped) {
+                    current += "\\";
+                }
+                data.push(current);
+
+                for (const item of data) {
                     const firstEquals = item.indexOf("=");
                     if (firstEquals < 0) {
                         throw new ParseError(`\\htmlData key/value '${item}'` +
--- a/src/functions/html.ts
+++ b/src/functions/html.ts
@@ -50,8 +50,8 @@ defineFunction({
                 };
                 break;
             case "\\htmlData": {
-                // Split on unescaped commas only (\\, → literal comma)
-                // Manual parse avoids negative lookbehind compatibility issues
+                // Split on unescaped commas only (\\, → literal comma).
+                // Manual parse avoids negative lookbehind compatibility issues.
                 const data = [];
                 let current = "";
                 let escaped = false;
@@ -68,11 +68,15 @@ defineFunction({
                         current += char;
                     }
                 }
+                // Emit a dangling backslash literally rather than dropping it
+                if (escaped) {
+                    current += "\\";
+                }
                 data.push(current);
 
                 for (const item of data) {
                     if (item.length === 0) {
-                        continue; // Skip trailing/empty entries
+                        continue;
                     }
                     const firstEquals = item.indexOf("=");
                     if (firstEquals < 0) {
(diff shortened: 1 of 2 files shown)


lloyd-c137:
All three points addressed in 8dfad10:

1. ✅ Removed the redundant `.replace()` — the parser loop already produces the correct output, no post-processing needed
2. ✅ Dangling trailing backslash now emitted literally (added `if (escaped) { current += "\\" }` before `data.push(current)`)
3. ✅ Empty entries still skip (consistent with how the original `split(",")` handled trailing commas), but actual key=value parsing errors are still caught

src/functions/html.ts
```diff
@@ -50,17 +50,43 @@ defineFunction({
                 };
                 break;
             case "\\htmlData": {
-                const data = value.split(",");
-                for (let i = 0; i < data.length; i++) {
-                    const item = data[i];
+                // Split on unescaped commas only (\\, → literal comma).
+                // Manual parse avoids negative lookbehind compatibility issues.
```
grigoriy-reshetniak: Based on [what](https://caniuse.com/js-regexp-lookbehind) I see in Caniuse, Safari being the main concern for compatibility. But in [browserlist](https://browsersl.ist/#q=Safari+%3C+16.4) I see there are very few people would be affected by this, so I think it's safe to use lookbehind and avoid manual parse.

grigoriy-reshetniak:
I'd also appreciate a unit test.

lloyd-c137:
Updated per review: replaced the manual parser with a negative lookbehind regex `/(?<!\\),/` + `string.replace(/,/g, ",")`. Also added unit tests for multiple key=value pairs and edge cases. All 1290 tests pass.

src/functions/html.ts
```diff
@@ -50,17 +50,38 @@ defineFunction({
...
+                    } else {
+                        current += char;
+                    }
+                }
+                data.push(current);
+
+                for (const item of data) {
+                    if (item.length === 0) {
+                        continue; // Skip trailing/empty entries
+                    }
```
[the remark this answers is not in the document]
grigoriy-reshetniak: @lloyd-c137 I believe greptile is right here, original `Array.split()` handled it differently:

image

I think that this check should be removed.

src/functions/html.ts
```diff
@@ -50,17 +50,24 @@ defineFunction({
                 };
                 break;
             case "\\htmlData": {
-                const data = value.split(",");
-                for (let i = 0; i < data.length; i++) {
-                    const item = data[i];
+                // Split on unescaped commas using negative lookbehind.
```
grigoriy-reshetniak: Please remove this comment.

edemaine:
I doubt this PR works as is. `\htmlData{text=\,}{content}` already "works"/has a meaning, which is `data-text="\mskip+3mu\relax"` because macros expand before `\htmlData` applies and `\,` is a thin space implemented as a macro. `\htmlData` will never see `\,` unless I missed something.

I wonder if ``\htmlData{text=foo\char`,bar}{content}`` should be the correct way to escape `,` in LaTeX?

lloyd-c137:
Addressed both review points:

1. Removed the empty-item skip — the original `value.split(",")` let empty items through and threw a ParseError, so the new parser should too. Matching existing behavior.

2. Removed the long comment block about browser compatibility. Instead, kept a brief note about why the manual parser is used over lookbehind.

Regarding @edemaine's question about `\,` escaping: you're right that `\,` (thin space) gets expanded by the tokenizer before reaching the raw argument. In practice `\,` serves as a backslash-escape of the comma, which the parser correctly handles as: `\\` → literal backslash + `,` → separator. The escape mechanism handles `\,` (single backslash before comma) defensively if it somehow appears — even though under normal KaTeX parsing `\,` never does.

A more practical escaping path for users who need literal commas in \htmlData values would be to use `\\,` in the LaTeX source (double backslash + comma), which produces a literal backslash that the parser keeps. Happy to explore alternative approaches if there are specific requirements.

edemaine:
I think `{,}` might be a more reasonable way to escape a comma (and generally anything inside braces). If it can be distinguished.

lloyd-c137:
@edemaine That makes sense — using `{,}` is far more natural in TeX than backslash escaping, and it avoids the `\,` thin space collision entirely.

If I understand correctly: the parser should be brace-aware — treat `{...}` as a single token so commas inside braces are not treated as separators. This would mean:
- `\htmlData{text={a,b}, key=val}{x}` → `data-text=a,b` (comma inside braces preserved)
- Users write `{,}` for a literal comma, which is standard TeX notation

Would you like me to update the PR to use a brace-depth-aware parser instead of the current backslash approach? I can keep the backslash escape as a fallback too.

lloyd-c137:
Closing this PR as stale — been open with no maintainer activity. Happy to re-open if there's renewed interest.

grigoriy-reshetniak:
@lloyd-c137 it feels like this PR is closed to soon, it isn't stale by KaTeX standard. I think it would be worth to reopen it.

I believe suggestions from @edemaine was only related to 
> Users write {,} for a literal comma, which is standard TeX notation

part, not 

> `\htmlData{text={a,b}, key=val}{x} `→ `data-text=a,b` (comma inside braces preserved)

As for me `{,}` looks more reasonable, because `{a,b}` looks like a group, not an escaped comma.

PR #4219, opened 2026-05-20
refactor: tighten ParseNode types

grigoriy-reshetniak:
This PR is a continuation of https://github.com/KaTeX/KaTeX/pull/4145. It tightens the typing of different sub-types of `ParseNode`, to remove `TYPE | undefined | null` where it's possible. 

Big change is breaking `ParseNodeTypes` map into separate types. I feel like it's worth it, because it allows to use discriminated unions.

Also there's some minor type extractions or improvements on the way.


src/Parser.ts
```diff
@@ -525,18 +517,18 @@ export default class Parser {
         funcData: FunctionSpec<NodeType> | EnvSpec<NodeType>,
```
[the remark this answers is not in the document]
grigoriy-reshetniak: Fixed

src/types/nodes.ts
```diff
@@ -0,0 +1,456 @@
...
+    type: "op";
+    limits: boolean;
+    alwaysHandleSupSub?: boolean;
+    suppressBaseShift?: boolean;
+    parentIsSupSub: boolean;
+}
+
+type OperatorNode =
+    | OperatorNodeBase & {symbol: true; name: string; body?: never;}
+    | OperatorNodeBase & {symbol: false; body: AnyParseNode[]; name?: never;}
```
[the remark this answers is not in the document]
grigoriy-reshetniak: This variant is derived from actual usage.

src/Parser.ts
```diff
@@ -2,24 +2,22 @@
...
 import {isAtom} from "./atoms";
 import {validUnit} from "./units";
 import {supportedCodepoint} from "./unicodeScripts";
 import ParseError from "./ParseError";
 import {combiningDiacriticalMarksEndRegex} from "./Lexer";
 import Settings from "./Settings";
 import SourceLocation from "./SourceLocation";
-import {uSubsAndSups, unicodeSubRegEx} from "./unicodeSupOrSub";
+import {unicodeSubRegEx, uSubsAndSups} from "./unicodeSupOrSub";
 import {Token} from "./Token";
```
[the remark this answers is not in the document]
grigoriy-reshetniak: Must have been ~~the wind~~ IDE auto-format.
Fixed.

grigoriy-reshetniak:
@greptile revise your review

src/Namespace.ts
```diff
@@ -82,8 +82,8 @@ export default class Namespace<Value> {
      * to `false` in JavaScript.  Use `if (namespace.get(...) != null)` or
      * `if (namespace.has(...))`.
      */
-    get(name: string): Value | null | undefined {
-        if (this.current.hasOwnProperty(name)) {
+    get(name: string): Value | undefined {
+        if (name in this.current) {
```
edemaine: This change to `in` feels unsafe. `toString` and `constructor` match. Inconsistent with `has`.

src/Parser.ts
```diff
@@ -484,7 +477,7 @@ export default class Parser {
         } else if (this.mode === "text" && !funcData.allowedInText) {
             throw new ParseError(
                 "Can't use function '" + func + "' in text mode", token);
-        } else if (this.mode === "math" && funcData.allowedInMath === false) {
+        } else if (this.mode === "math" && !funcData.allowedInMath) {
```
edemaine, line 480: This is functionally quite different: it changes the default (`undefined`) behavior. Is it correct?

src/types/nodes.ts
```diff
@@ -0,0 +1,456 @@
...
+type ParseNodeMap = {
+    [K in NodeType]: Extract<AnyParseNode, {type: K}>;
+};
+
+type BaseNode = {
+    mode: Mode;
+    loc?: SourceLocation | null;
+};
+
+type ArrayNode = BaseNode & {
```
edemaine, line 92: Would `interface ... extends` be better? I'm not sure.  Maybe not because of e.g. `SupSubNode`.

edemaine (commented):
Lots of great improvements here! Thanks for the cleanup!

I pushed a few small changes that you hopefully agree with (feel free to revert if not), that seemed easy to do than to ask for:

- Slice-based assertions for `mclass`, CD label `side`, and old font command typing, using old `Slice1` and new `Slice4` and `Slice5` helper types
  - Added `satisfies` checks for related command name lists.
- Replaced `binrelClass` string concatenation/cast with a template string.
- Changed `color: color` to shorthand `color`.
- Added a leading `|` to the first `AnyParseNode` union member.
- Added a comment that spacing nodes render as `textord`.

I left some comments on some functionality changes that I *think* should be reverted but maybe you had reasons I'm missing.

src/Namespace.ts
edemaine: Actually we should probably use `Object.hasOwn` in case a `macros` object accidentally adds its own `hasOwnProperty`. Admittedly unlikely given that macros normally start with `\`, but better practice I think.

src/Parser.ts
grigoriy-reshetniak: I thought that undefined is not possible value for `funcData.allowedInText` at this point as this case is [already covered](https://github.com/KaTeX/KaTeX/blob/main/src/defineFunction.ts#L177) at `defineFunction`.

src/types/nodes.ts
grigoriy-reshetniak: I would like to avoid using `interface` here. I like to declare as `type` to describe an entity and as `interface` things which are just needed to pass by, like function arguments. 

It also seems that `type` is more future-proof. Maybe someday we'd like to do something like 

```
type NodesWithBody = {
    [K in NodeType]: ParseNode<K> extends { body: any } ? K : never;
}[NodeType];
```

or

```
type HtmlBuilders = {
    [K in NodeType]?: (group: ParseNode<K>, options: Options) => HtmlDomNode;
};
```

and `type`s would give more opportunities.

src/Namespace.ts
grigoriy-reshetniak: This was a miss from my part, rolled back.

grigoriy-reshetniak:
Thanks for the added improvements, I like them!

I explored `Slice4` and `Slice5` during my implementation and wasn't happy with either, so I ended up using type casts. But if you find them preferable -- I don't mind keeping them.

src/Namespace.ts
edemaine: What do you think about switching to `hasOwn`?

src/Parser.ts
edemaine: Ah, I forgot; that is OK then.

I do wonder whether it's worth removing that normalization step though. It seems a little unnecessary to rebuild the object every time we call `defineFunction`, especially for an internal API which we can expect reasonable use of (e.g. correctly following types). If we instead had `defineFunction(names, data)` simply set `_functions[names[i]] = data` (including builders and everything), then we would probably save a fair amount of initialization time and memory thrashing. Something to think about...

I profiled and `defineFunction` gets called 77 times and takes about 0.15 ms. Not nothing. But the bigger culprit is `defineSymbol` which gets called 1881 times and takes about 0.89 ms. `defineMacro` gets called 337 times and takes about 0.13 ms.

Anyway, this change is fine for this PR, but we might revert it if we end up removing normalization.

edemaine:
> I explored `Slice4` and `Slice5` during my implementation and wasn't happy with either, so I ended up using type casts. But if you find them preferable -- I don't mind keeping them.

Yeah, maybe it's just me, but I find it a bit more reassuring that they check that the names didn't get mangled or aren't missing a case, whereas a typecast can hide these issues.

Things would be a *lot* better if `defineFunction` guaranteed that the `name` parameter was typed to be among `names`. Then we could use `as const` in `names`, and use `typeof name` in the argument to `SliceN`. Maybe something to try fixing.

src/Namespace.ts
grigoriy-reshetniak: It's not allowed for supported Chrome version in [.browserlistrc](https://github.com/KaTeX/KaTeX/blob/main/.browserslistrc). 

Should I update it anyway or should `.browserlistrc` and related code upgrade be handled separately (there are 38 `hasOwnProperty` alone)?

edemaine: Ah, if there are that many, should be a separate PR. I think we could safely increase Chrome beyond 60. Should probably look at percentages to know how far to increase it, but hopefully within `hasOwn`'s support.

edemaine (approved):

PR #4222, opened 2026-05-22
perf: simplify `defineFunction` to avoid destructuring, improve typing

edemaine:

1. `defineFunction` normalized each function registration by constructing a new parser-facing function spec, applying defaults, and storing HTML/MathML builders through separate helper tables. This added work during module initialization, even though most defaults can be handled at use sites.

2. The `names` field on function registrations was typed broadly, so handlers saw `funcName` as `string` even when a registration listed a fixed set of command names. Some files used separate command-name union types with `satisfies` to recover that information manually.

1. `defineFunction` now stores the registration data directly and registers builders from the same object, avoiding the previous normalization/wrapping step during initialization. Parser call sites now apply defaults where needed for optional fields.
   * Local benchmark runs showed about a 2-2.5% median improvement in minified bundle require time (from 1.5398ms to 1.5046ms, a 0.0352ms improvement), with a small minified bundle-size reduction (from 272,537 to 271,179 bytes, a 1,358-byte or 0.5% improvement, or about half that when gzipped).
   * Should also reduce memory thrashing a little. Does result in `names` getting permanently stored, instead of garbage collected. 
   * Another argument for this change is that the code is a lot simpler.
   * Also removed a few redundant specifications of duplicate builders on shared types, which was unnecessary.
   * Originally inspired by https://github.com/KaTeX/KaTeX/pull/4219#discussion_r3282426680

2. Function registrations now infer the handler `funcName` type from `names`. For example, `names: ["\\foo", "\\bar"]` gives the handler `funcName` type `"\\foo" | "\\bar"`. This removes several redundant command-name union types (Slices are getting cleaner!) and lets supporting maps/arrays, such as delimiter sizes and sizing commands, be checked more precisely.

BREAKING CHANGE: The internal API for `__defineFunction` changed: you should no longer wrap properties in `props`.

BUT: I'm not totally sure this should be considered a breaking change. It would affect any third-party modules that use `__defineFunction`. But looking at https://github.com/search?q=katex+%22__defineFunction%22&type=code I'm not sure there are any. Whereas I know of examples of `__defineMacro` and `__defineSymbol`. I'd like feedback on whether we should bump minor version for this.

### DIFF
--- a/src/functions/delimsizing.ts
+++ b/src/functions/delimsizing.ts
@@ -92,10 +92,9 @@ defineFunction({
         "\\bigm", "\\Bigm", "\\biggm", "\\Biggm",
         "\\big",  "\\Big",  "\\bigg",  "\\Bigg",
     ],
-    props: {
-        numArgs: 1,
-        argTypes: ["primitive"],
-    },
+    numArgs: 1,
+    argTypes: ["primitive"],
+
     handler: (context, args) => {
         const delim = checkDelimiter(args[0], context);
 
@@ -107,6 +106,7 @@ defineFunction({
             delim: delim.text,
         };
     },
+
     htmlBuilder: (group, options) => {
         if (group.delim === ".") {
             // Empty delimiters still count as elements, even though they don't
@@ -157,10 +157,9 @@ function assertParsed(group: ParseNode<"leftright">) {
 defineFunction({
     type: "leftright-right",
     names: ["\\right"],
-    props: {
-        numArgs: 1,
-        primitive: true,
-    },
+    numArgs: 1,
+    primitive: true,
+
     handler: (context, args) => {
         // \left case below triggers parsing of \right in
         //   `const right = parser.parseFunction();`
@@ -183,10 +182,9 @@ defineFunction({
 defineFunction({
     type: "leftright",
     names: ["\\left"],
-    props: {
-        numArgs: 1,
-        primitive: true,
-    },
+    numArgs: 1,
+    primitive: true,
+
     handler: (context, args) => {
         const delim = checkDelimiter(args[0], context);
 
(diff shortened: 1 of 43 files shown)


src/functions/delimsizing.ts
```diff
@@ -35,7 +32,10 @@ const delimiterSizes: Record<string, {
     "\\Big"  : {mclass: "mord",     size: 2},
     "\\bigg" : {mclass: "mord",     size: 3},
     "\\Bigg" : {mclass: "mord",     size: 4},
-};
+} as const satisfies Record<string, {
+    mclass: "mopen" | "mclose" | "mrel" | "mord";
```
grigoriy-reshetniak: Would it be worth to be [`MathClass`](https://github.com/KaTeX/KaTeX/blob/main/src/types/index.ts#L39) with `Exclude`?

src/functions/delimsizing.ts
```diff
@@ -35,7 +32,10 @@ const delimiterSizes: Record<string, {
...
+    size: 1 | 2 | 3 | 4;
```
grigoriy-reshetniak: I think it should be [`DelimiterSize`](https://github.com/KaTeX/KaTeX/blob/main/src/types/index.ts#L26).

grigoriy-reshetniak (commented):
This looks amazing, thank you!

Regarding the breaking change: I think this definitely deserves a bump. Those search results are not very reliable, they touch only public GitHub repositories. For example, GitLab itself uses KaTeX (I've seen `ParseError`s) and there's no guarantee that `__defineFunction` is not extended there.

src/functions/delimsizing.ts
edemaine: Maybe just `MathClass` makes sense? I'm not sure it's important that the other options are excluded, but I wanted to enforce some kind of verification against typos. `MathClass` looks good to me.

grigoriy-reshetniak (approved):

Issue #3722, opened 2022-09-23
Expose some version of buildTree


wu000168:
**Is your feature request related to a problem? Please describe.**
Currently both `parseTree` and `renderToDomTree` are exposed but there is no exposed way to convert an internal parse tree to a DOM/HTML tree. I understand that none of these APIs are expected to be stable or for public use. However, this would improve the consistency of internal functions exposed and especially useful in my case as described below.

**Additional context:**
I am working on an experimental project that require selecting segments of KaTeX output. Currently I am hooking into the custom macros in the options and parsing the groups myself in order to find the correct selections and inserting custom tokens (as strings) in order to find the correct ranges in the final HTML output. Exposing `buildTree` (as well as `Token`) would dramatically improve the extensibility of KaTeX. This might not be a common use case but should require minimal effort to implement and would likely be worth considering if these functions are to continue to be exposed once the API is stable.

PR #4226, opened 2026-05-31
feat: add opt-in source locations to HTML output

chaoxu:
## Motivation

This adds an opt-in way to recover approximate TeX source ranges from KaTeX's HTML output. When `outputSourceLocations: true` is enabled, rendered HTML nodes receive `data-katex-source-start` and `data-katex-source-end` attributes. Consumers can use those attributes to map clicks or selections in rendered math back to the corresponding TeX source range.

The main use case is editor integration: notebooks, Markdown editors, live preview panes, and WYSIWYG-like math tools often show rendered KaTeX beside or instead of the TeX source. When a user clicks part of the rendered formula, the editor wants to move the source cursor to the relevant part of the input. Today this requires downstream patches, custom macro tricks, or DOM heuristics.

This has come up before in related discussions:

- #809 discusses WYSIWYG/editor use cases, including mapping rendered math interactions back to the original input.
- #3722 describes a project that needs to select segments of KaTeX output and currently works around the lack of source-position metadata by inserting custom tokens.
- #358 is related to cursor/editor rendering, though this PR focuses only on source ranges.

This PR does not attempt to make KaTeX a full editor model. It only exposes source locations that KaTeX already has internally, and only when explicitly requested.

Demo: http://chaoxu.prof/katex-source-locations-demo/
Demo source: https://github.com/chaoxu/katex-source-locations-demo

KaTeX HTML output did not expose a supported way to map rendered HTML nodes back to source offsets in the input expression. Editor integrations that need click-to-source behavior have to keep a local patch or use rough geometry-only heuristics.

This adds an opt-in `outputSourceLocations` option. When enabled, HTML output includes `data-katex-source-start` and `data-katex-source-end` attributes on rendered nodes that can be mapped back to input source ranges. The offsets are zero-based UTF-16 offsets into the expression passed to `render` or `renderToString`.

The option defaults to `false`, so existing output is unchanged. Wrapper nodes that do not have their own parse-node location inherit the smallest range covering their children, which makes the attributes useful for editor click targets without requiring every render node to come directly from a parse node.

Tests/docs/types were added for the new option.

**Validation**

- `yarn test:jest`
- `yarn test:ts`
- `yarn test:lint:js`
- `yarn test:lint:css`
- `yarn build`
- `yarn test:build`


test/katex-spec.ts
```diff
@@ -2499,6 +2499,41 @@ describe("A markup generator", function() {
...
+        expect(markup).toContain(
+            'data-katex-source-start="5" data-katex-source-end="11"');
+        expect(markup).toContain(
+            'data-katex-source-start="6" data-katex-source-end="7"');
+        expect(markup).toContain(
+            'data-katex-source-start="9" data-katex-source-end="10"');
+    });
 });
 
 describe("A parse tree generator", function() {
```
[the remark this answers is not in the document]
chaoxu: Addressed in e25cfbe and tightened in ca4ce17: the default htmlAndMathml path is covered, and the test now verifies source-location attributes are present in the HTML subtree but absent from the MathML portion.

src/buildHTML.ts
```diff
@@ -251,6 +251,67 @@ export const makeNullDelimiter = function(
...
+            for (let i = current.children.length - 1; i >= 0; i--) {
+                stack.push(current.children[i]);
+            }
+        }
+    }
+
+    return minStart !== Infinity && maxEnd !== -Infinity
+        ? [minStart, maxEnd]
+        : null;
+};
```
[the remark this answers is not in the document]
chaoxu: Addressed in e25cfbe: inheritSourceLocationRange now starts from the node children instead of including the wrapper node itself.

src/domTree.ts
```diff
@@ -376,6 +376,7 @@ export class SymbolNode implements HtmlDomNode {
     width: number;
     maxFontSize: number;
     classes: string[];
+    attributes?: Record<string, string>;
```
grigoriy-reshetniak: It looks like `attributes` are missing in `constructor`.

src/domTree.ts
```diff
@@ -419,6 +420,11 @@ export class SymbolNode implements HtmlDomNode {
         return this.classes.includes(className);
     }
 
+    setAttribute(attribute: string, value: string) {
+        this.attributes = this.attributes || {};
```
grigoriy-reshetniak: I believe this line is redundant if `|| {}` part moved to `constructor`.

src/buildHTML.ts
```diff
@@ -251,6 +251,68 @@ export const makeNullDelimiter = function(
...
+    setAttribute?: (attribute: string, value: string) => void;
+};
+
+const setSourceLocationAttributes = function(
+    node: HtmlDomNode,
+    start: number,
+    end: number,
+) {
+    const attributedNode = node as AttributedDomNode;
+    if (typeof attributedNode.setAttribute !== "function") {
```
grigoriy-reshetniak:
```suggestion
    if (typeof attributedNode.setAttribute === "function") {
         attributedNode.setAttribute(sourceStartAttribute, String(start));
         attributedNode.setAttribute(sourceEndAttribute, String(end));
    }
```

src/buildHTML.ts
```diff
@@ -251,6 +251,68 @@ export const makeNullDelimiter = function(
...
+    attributes?: Record<string, string>;
+    children?: ReadonlyArray<HtmlDomNode>;
+    setAttribute?: (attribute: string, value: string) => void;
+};
+
+const setSourceLocationAttributes = function(
+    node: HtmlDomNode,
+    start: number,
+    end: number,
+) {
```
grigoriy-reshetniak: I believe `void` as return type would be optimal here.

src/buildHTML.ts
```diff
@@ -251,6 +251,68 @@ export const makeNullDelimiter = function(
...
+/**
+ * When a builder node does not have its own parse-node location, merge child
+ * source locations so wrapper spans still act as useful click targets.
+ */
+const inheritSourceLocationRange = function(
+    node: HtmlDomNode,
+): [number, number] | null {
+    let minStart = Infinity;
+    let maxEnd = -Infinity;
+    const children = (node as AttributedDomNode).children;
```
grigoriy-reshetniak: `inheritSourceLocationRange` takes `node`, but only looks at it's `children` and that means that if `node` itself has any attributes they will be missed.

grigoriy-reshetniak:
Thank you, this looks promising!

I'm a little concerned about `SymbolNode` may now create `span` solely for attributes, this will impact  text nodes in output.

grigoriy-reshetniak:
@greptile revise your review

grigoriy-reshetniak (approved):

edemaine:
GPT reports these issues. Can you verify and fix?

- **[P1] Macro expansions emit offsets from the macro body, not the input.** [src/buildHTML.ts:347](https://github.com/KaTeX/KaTeX/blob/d777b7338e54e135d75bf9e2690a685ddfb71f38/src/buildHTML.ts#L347-L349) serializes every `group.loc` without checking its lexer. String macros are tokenized using a separate lexer ([MacroExpander.ts:429](https://github.com/KaTeX/KaTeX/blob/d777b7338e54e135d75bf9e2690a685ddfb71f38/src/MacroExpander.ts#L429-L435), so `renderToString("\\neq", …)`—an input of length 4—emits ranges including `26–45`. Built-in and configured macros therefore violate the documented “offsets into the input expression” contract. Locations need original-input provenance or must be omitted when their lexer differs.

- **[P2] Character coalescing leaves stale ranges.** Locations are attached in `buildGroup`, after which `buildExpression` invokes [tryCombineChars](https://github.com/KaTeX/KaTeX/blob/d777b7338e54e135d75bf9e2690a685ddfb71f38/src/buildHTML.ts#L77-L78). The combiner concatenates text without merging or preserving attributes. `\text{ab}` consequently produces one `ab` span labeled `6–7`, covering only `a`, instead of `6–8`. Either merge ranges when combining or prevent combining nodes with different source locations.

PR #4228, opened 2026-06-09
chore(screenshotter): migrate from safari to webkit

grigoriy-reshetniak:
Safari was used to take screenshots to compare with previous version.

Webkit is used to take screenshots to compare with previous version.

**Outdated**
There's is a problem with this approach: 
 WebKitGTK rendering is not deterministic. In current pipeline only `DeepFontSizing` is failing, but I've seen `Includegraphics`, `Pmb`, `SizingBaseline`, `Sizing`, `VerticalSpacing`, `Arrays`, `Colors`, `LowerAccent`, `MathAtom`, `OverUnderline`, `Phantom`, `Phase`, `StretchyAccent`, `StretchyAccentColor`, `StrikeThrough`, `StrikeThroughColor`, `UnsupportedCmds`, and `VerticalSpacing` also failed (or passed) during different runs without any code chages. This tends to happen for certain tests more often than for others, but whole list of possible range is unknown.

<details>
<summary>AI opinion of the reason</summary>

- Font rendering timing — WebKitGTK renders fonts asynchronously. The \Huge, \big, \Big size commands in DeepFontSizing produce glyphs at multiple sizes. If fonts haven't fully loaded/rendered when the screenshot is taken, the output differs by a few pixels.
- Subpixel / antialiasing differences — WebKitGTK's text rasterizer can produce slightly different pixel values across runs, especially for large/scaled glyphs (which this test heavily uses with \Huge, \big, \Big, \dfrac). Unlike Chrome/Firefox Selenium images (which use a consistent Xvfb + Mesa stack), your WebKit container uses xvfb-run with software rendering (libEGL warning: DRI3 error) which is less deterministic.

</details>

There's an option to increase `--wait` and  `--attempts`, narrow list down as much as possible and then do something like [`Lap` hack](https://github.com/KaTeX/KaTeX/blob/main/dockers/screenshotter/screenshotter.js#L523) but this looks fragile and could easily because obsolete, because now `Lap` is consistent even without that code.

**This was fixed by adding Mesa software renderer to webkit image, but still could be an issue**.


grigoriy-reshetniak:
@greptile, review every file except `.png` ones. Also do same in the following PRs.

dockers/screenshotter/screenshotter.sh
```diff
@@ -13,24 +13,42 @@ cleanup() {
 container=
 trap cleanup EXIT
 status=0
-for browserTag in "firefox:128.0-20260222" "chromium:145.0-20260222"; do
+#use webkit:1.2 for x86_64 and webkit:1.1 for arm
+for browserTag in "firefox:128.0-20260222" "chromium:145.0-20260222" "webkit:1.2"; do
```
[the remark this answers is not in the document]
grigoriy-reshetniak: > workflow builds a multi-arch image

It supposed to build image in future, but for now it hasn't build anything yet. Manual run is unavailable until workflow merged to main, so it reflects current state of things.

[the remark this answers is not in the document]
grigoriy-reshetniak: No, don't make a rule out of it.

grigoriy-reshetniak:
@greptile revise your review

dockers/screenshotter/README.md
```diff
@@ -22,14 +19,39 @@ docker containers with the selenium setups running.  Essentially you
...
 manually.  Example run for Firefox:
 
-    container=$(docker run -d -P selenium/standalone-firefox:2.46.0)
+    container=$(docker run -d -P selenium/standalone-firefox:128.0-20260222)
     node dockers/screenshotter/screenshotter.js -b firefox -c ${container}
     # possibly repeat the above command as often as you need, then eventually
     docker stop ${container}
     docker rm ${container}
 
 For Chrome, simply replace both occurrences of `firefox` with `chrome`.
```
edemaine: Is this correct? Does version number change?

dockers/screenshotter/screenshotter.js
```diff
@@ -216,11 +224,21 @@ if (seleniumURL) {
...
         driver = await getProxyDriver();
     } else {
-        if (seleniumIP) {
+        if (seleniumURL) {
+            const urlObj = new URL(seleniumURL);
+            if (!seleniumIP) {
+                seleniumIP = urlObj.hostname;
+            }
+            seleniumPort = parseInt(urlObj.port, 10) || seleniumPort;
+            await pRetry(tryConnect, {
```
edemaine: Could we move this outside the `if` (and drop `else`) to avoid repetition with below?

dockers/screenshotter/screenshotter.sh
```diff
@@ -13,24 +13,39 @@ cleanup() {
...
+        extra_args="--selenium-url http://localhost:${port} --wait 0.5"
+    else
+        image=selenium/standalone-${browserTag}
+        echo "Starting container for ${image}"
+        container=$(docker run -d --shm-size=2g -P ${image})
+        [[ ${container} ]] || continue
+        echo "Container ${container:0:12} started"
+        sleep 5
+        extra_args="--container=${container}"
+    fi
```
edemaine, line 43: I understand the image overriding - but why the other differences / why can't the two branches share the code?

edemaine (commented):
Sorry for the delay on this.

Overall, looks good! In addition to my comments, GPT has these findings:

1. **[P1] WebKit screenshotter CI is still timing-flaky.**  
   The CI WebKit path uses `ghcr.io/katex/katex-webkit:2.0`, but its Safari/WebKit args omit the stabilizing wait that the local script adds: workflow args are set at [`.github/workflows/screenshotter.yml:55`](https://github.com/KaTeX/KaTeX/blob/ba1a1c0c21846a9eca05ebe9e5e517db351ab630/.github/workflows/screenshotter.yml#L55) and invoked at [line 69](https://github.com/KaTeX/KaTeX/blob/ba1a1c0c21846a9eca05ebe9e5e517db351ab630/.github/workflows/screenshotter.yml#L69), while local WebKit gets `--wait 0.5` in [`dockers/screenshotter/screenshotter.sh:34`](https://github.com/KaTeX/KaTeX/blob/ba1a1c0c21846a9eca05ebe9e5e517db351ab630/dockers/screenshotter/screenshotter.sh#L34). The latest successful PR run already had WebKit retries: `DeepFontSizing` failed 4 times before passing on the 5th/default final attempt, plus retries for `Includegraphics` and `SizingBaseline`. Since default attempts are only 5 in [`screenshotter.js:68-69`](https://github.com/KaTeX/KaTeX/blob/ba1a1c0c21846a9eca05ebe9e5e517db351ab630/dockers/screenshotter/screenshotter.js#L68), this can fail unrelated PRs with a tiny timing variation. I’d pass the same WebKit wait in CI or otherwise fix the nondeterminism before merging.

1. **[P2] The image publishing workflow likely lacks checkout permission.**  
   [`webkit-image.yml:12-15`](https://github.com/KaTeX/KaTeX/blob/ba1a1c0c21846a9eca05ebe9e5e517db351ab630/.github/workflows/webkit-image.yml#L12) sets job permissions to only `packages: write`, then immediately runs `actions/checkout@v6`. With explicit `permissions`, unspecified scopes are removed, so this workflow should declare `contents: read` as well. Otherwise the documented update flow in [`dockers/screenshotter/README.md:47-53`](https://github.com/KaTeX/KaTeX/blob/ba1a1c0c21846a9eca05ebe9e5e517db351ab630/dockers/screenshotter/README.md#L47) may fail when maintainers try to rebuild/publish the next WebKit image.

1. **[P2] The PR silently keeps reporting this as Safari even though macOS Safari coverage is removed.**  But I think the plan was to rename from `safari` to `webkit` after review.

---

GPT regarding timeouts:

I think the font-rendering nondeterminism is still worth addressing before merge. The latest green WebKit run still had transient screenshot mismatches: `DeepFontSizing` failed four times before passing on the last default attempt, and `Includegraphics` / `SizingBaseline` also retried before passing. So Mesa seems to make the job pass, but not fully deterministic.

One thing to try: the screenshotter already calls `document.fonts.load(...)` for the KaTeX fonts, but it does not await `document.fonts.ready`, which is closer to “the document’s font set has settled after layout.” For images, `img.decode()` may also be a better readiness signal than only `complete`/`load`.

Possible follow-up in `load_fonts_and_images`:

```js
if (document.fonts && document.fonts.ready) {
  results.push(document.fonts.ready);
}

if (img.decode) {
  results.push(img.decode().catch(function() {}));
}
```

It may also help to wait one or two `requestAnimationFrame`s after fonts/images resolve, since there is no standard browser API that promises glyph rasterization has fully completed. I’d still mirror the local WebKit `--wait 0.5` in CI as a pragmatic guard, but adding these awaits would make the wait less arbitrary.

dockers/screenshotter/screenshotter.sh
grigoriy-reshetniak: Webkit has different `extra_args` and selenium requires `sleep 5`. I could've make it in something like 

```
    if [ "${browser}" = "webkit" ]; then
        image=ghcr.io/katex/katex-webkit:${browserTag#*:}
        browser="safari"
    else
        image=selenium/standalone-${browserTag}
    fi

    echo "Starting container for ${image}"
    container=$(docker run -d --shm-size=2g -P ${image})
    [[ ${container} ]] || continue
    echo "Container ${container:0:12} started"

    if [ "${browser}" = "safari" ]; then
        port=$(docker port "${container}" 4444 | head -1 | sed 's/.*://')
        extra_args="--selenium-url http://localhost:${port} --wait 0.5"
    else
        sleep 5
        extra_args="--container=${container}"
    fi
```
    
but IMO this doesn't look clearer and  `if [ "${browser}" = "safari" ];` is especially confusing, because it looks like it handles `safari` but really it's same `webkit`, just renamed. 

However, I thinks this part could be addressed at the same time when job and screenshots would be renamed and this would be less confusing.

dockers/screenshotter/README.md
grigoriy-reshetniak: This is indeed incorrect, now versions are different.

grigoriy-reshetniak:
Thanks for the review!

I've added checks for `document.fonts.ready `and `img.decode()`, and moved wait to `screenshotter.js` level. `requestAnimationFrame` would replicate `--wait` so I've omitted it for now.

grigoriy-reshetniak:
I see that job failed again. I'll investigate it further, but this isn't reproducible on my local so it will require some pushes.

grigoriy-reshetniak:
It looks like for large glyphs the WebKitGTK rendering is not deterministic, but only byte-to-byte. If images themselves being compared they are identical. Latest changes introduced two packages as dev dependencies, but they seem to be worth to consider as main comparison approach.

Edit: there's also an alternative approach with macOs runner https://github.com/KaTeX/KaTeX/pull/4245

grigoriy-reshetniak:
@greptile revise your review

grigoriy-reshetniak:
Closed in favor of https://github.com/KaTeX/KaTeX/pull/4245. It's worth to keep a branch for a while.

PR #4230, opened 2026-06-13
refactor: remove direct hasOwnProperty call

grigoriy-reshetniak:
This PR removes `Object.prototype.hasOwnProperty()` in favor of `Object.hasOwn()` or other methods which guarantee presence of  only enumerable properties. 

Sometimes during this I've noticed possibilities for type tightening and went down the rabbit hole.

Also, fixed comment which I've missed before: https://github.com/KaTeX/KaTeX/pull/4165#discussion_r3079357772


src/parseNode.ts
```diff
@@ -36,8 +36,8 @@ export function assertSymbolNodeType(node: AnyParseNode | null | undefined): Sym
  * Returns the node more strictly typed if it is of the given type. Otherwise,
  * returns null.
  */
-export function checkSymbolNodeType(node: AnyParseNode | null | undefined): SymbolParseNode | null | undefined {
-    if (node && (node.type === "atom" || NON_ATOMS.hasOwnProperty(node.type))) {
+export function checkSymbolNodeType(node: AnyParseNode): SymbolParseNode | null {
+    if (node.type === "atom" || node.type in NON_ATOMS) {
```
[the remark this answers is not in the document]
grigoriy-reshetniak: It was intentional.

src/functions/symbolsSpacing.ts
edemaine: Would this be better as a `Map`? Same question above.

Alternatively, we could use `Object.create` and use `in`. Not sure what's fastest, especially with one tiny object, so I ran another two benchmarks:

**regularSpace** Values are objects; lookup returns `value.className || ""`.

| Case | Object.hasOwn + lookup | hasOwnProperty.call + lookup | cached call + lookup | `in` + lookup | Map.has + get | Map.get sentinel |
| --- | ---: | ---: | ---: | ---: | ---: | ---: |
| hit with class `"~"` | 5.24 ns | 6.95 ns | 6.54 ns | 2.89 ns | 7.14 ns | 5.85 ns |
| hit without class `"\\space"` | 11.99 ns | 11.91 ns | 11.22 ns | 13.00 ns | 8.36 ns | 5.71 ns |
| miss | 7.60 ns | 7.26 ns | 6.91 ns | 11.99 ns | 5.80 ns | 5.94 ns |
| mixed hits | 13.44 ns | 12.62 ns | 13.82 ns | 12.98 ns | 9.68 ns | 6.56 ns |
| mixed hits/misses | 13.51 ns | 12.82 ns | 12.96 ns | 13.12 ns | 9.58 ns | 6.79 ns |

**cssSpace**  Values are strings; lookup returns the string.

| Case | Object.hasOwn + lookup | hasOwnProperty.call + lookup | cached call + lookup | `in` + lookup | Map.has + get | Map.get sentinel |
| --- | ---: | ---: | ---: | ---: | ---: | ---: |
| hit `"\\nobreak"` | 7.66 ns | 6.94 ns | 6.54 ns | 3.13 ns | 7.96 ns | 5.64 ns |
| hit `"\\allowbreak"` | 12.05 ns | 11.25 ns | 10.88 ns | 11.62 ns | 6.52 ns | 4.94 ns |
| miss | 7.42 ns | 6.57 ns | 6.36 ns | 12.15 ns | 6.73 ns | 7.07 ns |
| mixed hits | 11.48 ns | 10.93 ns | 11.73 ns | 11.23 ns | 6.93 ns | 5.11 ns |
| mixed hits/misses | 9.83 ns | 9.84 ns | 9.53 ns | 12.00 ns | 6.42 ns | 5.85 ns |

So `Map.get` with `undefined` as the absence sentinel is the winner, and feels the most logical.

<details>
<summary>Benchmark code</summary>

```js
const {performance} = require("node:perf_hooks");

const cssSpace = {
  "\\nobreak": "nobreak",
  "\\allowbreak": "allowbreak",
};

const regularSpace = {
  " ": {},
  "\\ ": {},
  "~": {className: "nobreak"},
  "\\space": {},
  "\\nobreakspace": {className: "nobreak"},
};

const cssMap = new Map(Object.entries(cssSpace));
const regularMap = new Map(Object.entries(regularSpace));
const hop = Object.prototype.hasOwnProperty;

let sink = 0;

function objectHasOwnRegular(k) {
  if (Object.hasOwn(regularSpace, k)) return regularSpace[k].className || "";
  return "ERR";
}

function hasOwnCallRegular(k) {
  if (Object.prototype.hasOwnProperty.call(regularSpace, k)) {
    return regularSpace[k].className || "";
  }
  return "ERR";
}

function cachedCallRegular(k) {
  if (hop.call(regularSpace, k)) return regularSpace[k].className || "";
  return "ERR";
}

function inRegular(k) {
  if (k in regularSpace) return regularSpace[k].className || "";
  return "ERR";
}

function mapHasGetRegular(k) {
  if (regularMap.has(k)) return regularMap.get(k).className || "";
  return "ERR";
}

function mapGetRegular(k) {
  const value = regularMap.get(k);
  return value === undefined ? "ERR" : value.className || "";
}

function objectHasOwnCss(k) {
  if (Object.hasOwn(cssSpace, k)) return cssSpace[k];
  return "ERR";
}

function hasOwnCallCss(k) {
  if (Object.prototype.hasOwnProperty.call(cssSpace, k)) return cssSpace[k];
  return "ERR";
}

function cachedCallCss(k) {
  if (hop.call(cssSpace, k)) return cssSpace[k];
  return "ERR";
}

function inCss(k) {
  if (k in cssSpace) return cssSpace[k];
  return "ERR";
}

function mapHasGetCss(k) {
  if (cssMap.has(k)) return cssMap.get(k);
  return "ERR";
}

function mapGetCss(k) {
  const value = cssMap.get(k);
  return value === undefined ? "ERR" : value;
}

function bench(fn, values, iters = 10_000_000) {
  for (let i = 0; i < iters / 20; i++) {
    sink += fn(values[i & (values.length - 1)]).length;
  }

  const samples = [];

  for (let r = 0; r < 13; r++) {
    const t0 = performance.now();
    let local = 0;

    for (let i = 0; i < iters; i++) {
      local += fn(values[i & (values.length - 1)]).length;
    }

    const t1 = performance.now();
    sink += local;
    samples.push((t1 - t0) * 1e6 / iters);
  }

  samples.sort((a, b) => a - b);
  return samples[6];
}

function printTable(title, cases, variants) {
  console.log(`\n${title}`);
  console.log(
    "| Case | Object.hasOwn + lookup | hasOwnProperty.call + lookup | " +
    "cached call + lookup | `in` + lookup | Map.has + get | Map.get sentinel |"
  );
  console.log("| --- | ---: | ---: | ---: | ---: | ---: | ---: |");

  for (const [name, values] of cases) {
    const row = variants.map(([, fn]) => bench(fn, values));
    console.log(`| ${name} | ${row.map(v => `${v.toFixed(2)} ns`).join(" | ")} |`);
  }
}

console.log(`Node ${process.version}`);

printTable(
  "regularSpace: presence check + value lookup",
  [
    ['hit with class `"~"`', ["~"]],
    ['hit without class `"\\\\space"`', ["\\space"]],
    ["miss", ["\\unknown"]],
    ["mixed hits", [" ", "\\ ", "~", "\\space", "\\nobreakspace"]],
    ["mixed hits/misses", [" ", "\\unknown", "~", "nope", "\\space"]],
  ],
  [
    ["Object.hasOwn", objectHasOwnRegular],
    ["hasOwnProperty.call", hasOwnCallRegular],
    ["cached call", cachedCallRegular],
    ["in", inRegular],
    ["Map.has+get", mapHasGetRegular],
    ["Map.get", mapGetRegular],
  ]
);

printTable(
  "cssSpace: presence check + value lookup",
  [
    ['hit `"\\\\nobreak"`', ["\\nobreak"]],
    ['hit `"\\\\allowbreak"`', ["\\allowbreak"]],
    ["miss", ["\\unknown"]],
    ["mixed hits", ["\\nobreak", "\\allowbreak"]],
    ["mixed hits/misses", ["\\nobreak", "\\unknown", "\\allowbreak", "nope"]],
  ],
  [
    ["Object.hasOwn", objectHasOwnCss],
    ["hasOwnProperty.call", hasOwnCallCss],
    ["cached call", cachedCallCss],
    ["in", inCss],
    ["Map.has+get", mapHasGetCss],
    ["Map.get", mapGetCss],
  ]
);

console.error("sink", sink);
```

</details>

src/MacroExpander.ts
```diff
@@ -449,19 +449,22 @@ export default class MacroExpander implements MacroContextInterface {
...
 
     /**
      * Determine whether a command is expandable.
      */
     isExpandable(name: string): boolean {
         const macro = this.macros.get(name);
-        return macro != null ? typeof macro === "string"
-                || typeof macro === "function" || !macro.unexpandable
-            : functions.hasOwnProperty(name) && !functions[name].primitive;
+        if (macro !== undefined) {
```
edemaine: I'd prefer keeping the more permissive check `macro != null`.

src/parseNode.ts
edemaine: Would `ATOMS`/`NON_ATOMS` be better as `Set`s? (I think so.)

test/helpers.ts
```diff
@@ -15,11 +15,11 @@ export class ConsoleWarning extends Error {
...
 /**
- * Return the first raw string if x is tagged literal. Otherwise return x.
+ * Return the first raw string if x is tagged literal. Otherwise, return x.
  */
-export const r = (x: any): any =>
-    x != null && Object.prototype.hasOwnProperty.call(x, 'raw')
-    ? (x as {raw: string[]}).raw[0] : x;
+// eslint-disable-next-line @typescript-eslint/no-wrapper-object-types
+export const r = (x: TemplateStringsArray | string | String): string =>
+    typeof x === "object" && "raw" in x ? x.raw[0] : String(x);
```
edemaine: Adding the `String(...)` cast means boxed `String` inputs passed through the test helpers are converted to primitive strings before parsing, so the `expect(new String(...)).toParseLike...` tests no longer exercise boxed-string handling. Can we keep boxed strings unchanged here, add  `| String` output, and maybe add overloads so primitive `string` input still returns `string` while `String` input returns `String`?

edemaine (commented):
I know I suggested `hasOwn`, but there's a compatibility issue:
* `Object.hasOwn` was not present in KaTeX’s pre-PR source or built `dist/` output.
* Babel leaves `Object.hasOwn(obj, key)` unchanged, so it would ship directly in the next build.
* KaTeX’s current Browserslist output still includes unsupported targets:

| Unsupported target currently included | Why it matters |
| --- | --- |
| `ios_saf 11.0-11.2` | No `Object.hasOwn` |
| `kaios 2.5` | No `Object.hasOwn` |
| `kaios 3.0-3.1` | No `Object.hasOwn` |
| `op_mini all` | Unknown/no reliable `Object.hasOwn` support |

A minimum Browserslist change would be:

```json
[
  "defaults",
  "not ios_saf &lt; 15.4",
  "not op_mini all",
  "not kaios &lt;= 3.1"
]
```

but that seems like a large compatibility change for a refactor.

I think the easy fix is to replace `Object.hasOwn` with `Object.prototype.hasOwnProperty.call`. It's wordier, but easy compatibility, and seems to be roughly free in terms of performance (see below).

---

`Object.entries` is cleaner, especially type-wise, but I was worried about performance. Here's a microbenchmark showing the difference:

Own props | for..in + Object.hasOwn | for..in + hasOwnProperty.call | Object.keys + lookup | Object.entries
-- | -- | -- | -- | --
0 | 2.1 | 4.2 | 5.8 | 6.3
1 | 10.6 | 5.6 | 8.0 | 13.3
2 | 37.2 | 15.9 | 44.8 | 46.1
4 | 65.2 | 19.2 | 74.4 | 72.1
8 | 124.3 | 29.4 | 137.6 | 129.5
32 | 1793.7 | 1785.5 | 1177.1 | 5994.2

This is sensitive to JIT noise, but it shows the expected allocation cost: `Object.entries` can get expensive as object size grows because it creates pair arrays.

<details>
<summary>Synthetic object-loop microbenchmark</summary>

```js
const {performance} = require("node:perf_hooks");

function makeObj(n) {
  const o = {};
  for (let i = 0; i < n; i++) o["k" + i] = "value" + i;
  return o;
}

let sink = 0;

function objectHasOwn(o) {
  let s = 0;
  for (const k in o) {
    if (Object.hasOwn(o, k)) s += o[k].length + k.length;
  }
  return s;
}

function hasOwnCall(o) {
  let s = 0;
  for (const k in o) {
    if (Object.prototype.hasOwnProperty.call(o, k)) {
      s += o[k].length + k.length;
    }
  }
  return s;
}

function keysLoop(o) {
  let s = 0;
  for (const k of Object.keys(o)) {
    s += o[k].length + k.length;
  }
  return s;
}

function entriesLoop(o) {
  let s = 0;
  for (const [k, value] of Object.entries(o)) {
    s += value.length + k.length;
  }
  return s;
}

function bench(fn, obj, iters) {
  for (let i = 0; i < iters / 10; i++) sink += fn(obj);

  const samples = [];
  for (let r = 0; r < 9; r++) {
    const t0 = performance.now();
    let local = 0;
    for (let i = 0; i < iters; i++) local += fn(obj);
    const t1 = performance.now();
    sink += local;
    samples.push((t1 - t0) * 1e6 / iters);
  }

  samples.sort((a, b) => a - b);
  return samples[Math.floor(samples.length / 2)];
}

console.log("| Own props | Object.hasOwn | hasOwnProperty.call | Object.keys | Object.entries |");
console.log("| ---: | ---: | ---: | ---: | ---: |");

for (const n of [0, 1, 2, 4, 8, 32]) {
  const obj = makeObj(n);
  const iters = n <= 2 ? 5_000_000 : n <= 8 ? 2_000_000 : 500_000;

  const a = bench(objectHasOwn, obj, iters);
  const b = bench(hasOwnCall, obj, iters);
  const c = bench(keysLoop, obj, iters);
  const d = bench(entriesLoop, obj, iters);

  console.log(`| ${n} | ${a.toFixed(1)} ns | ${b.toFixed(1)} ns | ${c.toFixed(1)} ns | ${d.toFixed(1)} ns |`);
}

console.error("sink", sink);
```

</details>

Luckily, this looping cost seems to be a minor part of `toMarkup` cost. Here's a more realistic benchmark for `toMarkup()` in already-built MathML trees:

Case | MathNode count | Attribute count | hasOwnProperty.call | Object.hasOwn | Object.keys | Object.entries
-- | -- | -- | -- | -- | -- | --
simple x | 5 | 2 | 0.537 us | 0.572 us | 0.552 us | 0.552 us
moderate expression | 32 | 2 | 2.621 us | 2.634 us | 2.746 us | 2.724 us
attr-heavy expression | 21 | 6 | 2.269 us | 2.306 us | 2.344 us | 2.323 us
matrix | 34 | 25 | 4.043 us | 3.961 us | 4.044 us | 4.057 us

`Object.keys`, `Object.entries`, `Object.hasOwn`, and `hasOwnProperty.call` were all effectively tied.

<details>
<summary>KaTeX MathML <code>toMarkup()</code> benchmark</summary>

```js
const {performance} = require("node:perf_hooks");

require("@babel/register")({extensions: [".js", ".ts"]});

const buildMathML =
  require("./src/buildMathML").default || require("./src/buildMathML");
const parseTree =
  require("./src/parseTree").default || require("./src/parseTree");
const Settings =
  require("./src/Settings").default || require("./src/Settings");
const Options =
  require("./src/Options").default || require("./src/Options");
const Style =
  require("./src/Style").default || require("./src/Style");

const {MathNode} = require("./src/mathMLTree");
const {escape} = require("./src/utils");
const {createClass} = require("./src/domTree");

function install(kind) {
  MathNode.prototype.toMarkup = function() {
    let markup = "<" + this.type;

    if (kind === "entries") {
      for (const [attr, value] of Object.entries(this.attributes)) {
        markup += " " + attr + "=\"" + escape(value) + "\"";
      }
    } else if (kind === "keys") {
      for (const attr of Object.keys(this.attributes)) {
        const value = this.attributes[attr];
        markup += " " + attr + "=\"" + escape(value) + "\"";
      }
    } else if (kind === "hasOwn") {
      for (const attr in this.attributes) {
        if (Object.hasOwn(this.attributes, attr)) {
          markup += " " + attr + "=\"" + escape(this.attributes[attr]) + "\"";
        }
      }
    } else if (kind === "call") {
      for (const attr in this.attributes) {
        if (Object.prototype.hasOwnProperty.call(this.attributes, attr)) {
          markup += " " + attr + "=\"" + escape(this.attributes[attr]) + "\"";
        }
      }
    }

    if (this.classes.length > 0) {
      markup += ` class ="${escape(createClass(this.classes))}"`;
    }

    markup += ">";

    for (let i = 0; i < this.children.length; i++) {
      markup += this.children[i].toMarkup();
    }

    markup += "</" + this.type + ">";
    return markup;
  };
}

function makeTree(expr, settingsObj = {}) {
  const settings = new Settings(settingsObj);
  const options = new Options({
    style: settings.displayMode ? Style.DISPLAY : Style.TEXT,
  }, settings);

  return buildMathML(
    parseTree(expr, settings),
    expr,
    options,
    settings.displayMode
  );
}

function stats(node, out = {nodes: 0, attrs: 0}) {
  if (node instanceof MathNode) {
    out.nodes++;
    out.attrs += Object.keys(node.attributes).length;
  }
  if (node.children) {
    for (const child of node.children) stats(child, out);
  }
  return out;
}

let sink = 0;

function bench(tree, kind, iters) {
  install(kind);

  for (let i = 0; i < 1000; i++) sink += tree.toMarkup().length;

  const samples = [];
  for (let r = 0; r < 11; r++) {
    const t0 = performance.now();
    let local = 0;
    for (let i = 0; i < iters; i++) {
      local += tree.toMarkup().length;
    }
    const t1 = performance.now();
    sink += local;
    samples.push((t1 - t0) * 1e3 / iters);
  }

  samples.sort((a, b) => a - b);
  return samples[Math.floor(samples.length / 2)];
}

const cases = [
  ["simple x", "x", {}, 200000],
  [
    "moderate expression",
    String.raw`\frac{a+b}{c+d}+\sqrt{x^2+1}+\sum_{i=1}^n i^2`,
    {},
    50000,
  ],
  [
    "attr-heavy expression",
    String.raw`\href{https://example.com}{\color{red}{x}}+\textsf{ABC123}+\overbrace{a+b}^{note}`,
    {trust: true},
    50000,
  ],
  [
    "matrix",
    String.raw`\begin{array}{c|c:r} a&b&c\\ \hline d&e&f\\ g&h&i \end{array}`,
    {},
    50000,
  ],
];

console.log("| Case | MathNode count | Attribute count | hasOwnProperty.call | Object.hasOwn | Object.keys | Object.entries |");
console.log("| --- | ---: | ---: | ---: | ---: | ---: | ---: |");

for (const [label, expr, settings, iters] of cases) {
  const tree = makeTree(expr, settings);
  const s = stats(tree);

  const call = bench(tree, "call", iters);
  const hasOwn = bench(tree, "hasOwn", iters);
  const keys = bench(tree, "keys", iters);
  const entries = bench(tree, "entries", iters);

  console.log(
    `| ${label} | ${s.nodes} | ${s.attrs} | ` +
    `${call.toFixed(3)} us | ${hasOwn.toFixed(3)} us | ` +
    `${keys.toFixed(3)} us | ${entries.toFixed(3)} us |`
  );
}

console.error("sink", sink);
```

</details>

So I think `Object.keys` and `Object.entries` are fine to use, but `Object.hasOwn` should probably be replaced by `Object.prototype.hasOwnProperty.call`.

src/parseNode.ts
grigoriy-reshetniak: It feels natural, but I had to cast `ReadonlySet<string>`, otherwise `Set.has(<string>)` complained about type incompatibility. Not sure it's clean, but still looks better.

src/MacroExpander.ts
grigoriy-reshetniak: I've changed it, but may I know why do you prefer it this way? `macro` is `MacroDefinition | undefined` so it felt safe to narrow.

test/helpers.ts
grigoriy-reshetniak: Good catch, thank you! Fixed it with overloads, without them, some tests were breaking.

grigoriy-reshetniak:
Thanks for review and benchmarks!

I went through this PR again, looking at every instance of `hasOwn` with idea
> if it's a loop guard -- I'll modify the loop itself

and now `Object.prototype.hasOwnProperty.call` has only 14 calls across code base.

src/MacroExpander.ts
edemaine: Old habits die hard? Partly I come from CoffeeScript where `macro?` tests `macro != null`. Partly I don't understand the intended difference between `undefined` and `null`, so I'd rather treat them identically (except for types where I'm happy to be stricter).

For macros specifically, I think there's a much stronger argument: existing callers of KaTeX might provide a `macros` object with `null` values, and it used to work, but this change would cause it to crash (I think). There doesn't seem to be a good reason to break such code.

src/functions/symbolsSpacing.ts
```diff
@@ -45,10 +44,10 @@ defineFunctionBuilders({
                     [mathsym(group.text, group.mode, options)],
                     options);
             }
-        } else if (Object.hasOwn(cssSpace, group.text)) {
+        } else if (cssSpaceClass !== undefined) {
```
edemaine: Personally I'd just use truthiness tests here and above (for shorter code, and empty classes would also not be great), but either way is fine.

src/atoms.ts
```diff
@@ -4,30 +4,30 @@
...
+    "op-token",
+    "spacing",
+    "textord",
+] as const;
+
+const Atoms: ReadonlySet<string> = new Set(atomList);
+export const NonAtoms: ReadonlySet<string> = new Set(nonAtomList);
+
+export type Atom = typeof atomList[number];
+type NonAtom = typeof nonAtomList[number];
```
edemaine, line 28: Not particularly sure why `Atoms` and `NonAtom` aren't exported... might be useful later?

edemaine (approved):
Nice improvements, thanks!

src/atoms.ts
grigoriy-reshetniak: I prefer not to export local variables, this gives an instant understanding that they aren't used anywhere else.

PR #4233, opened 2026-06-23
fix(auto-render, parser): add shouldRender callback and handle braced commas in \htmlData

### LINKED ISSUE #691: Suggestion: add a property called shouldRender to renderMathInElement options

### LINKED ISSUE #4204: htmlData values cannot contain commas

ShyamKumar1:
## Summary

- Fixes #691: `renderMathInElement` now accepts an optional `shouldRender` callback to conditionally skip rendering.
- Fixes #4204: `\\htmlData` now correctly parses key-value pairs within brace-delimited groups so that values can contain literal commas.

### #691: shouldRender callback

- Added optional `shouldRender?: (node: HTMLElement) => boolean` property to `renderMathInElement` options
- Callback is invoked after the `ignoredTags`/`ignoredClasses` check, so excluded elements are never passed to it
- If the callback returns `false`, the element and its descendants are skipped
- Backward compatible: no callback means unchanged behavior

### #4204: Braced values in \\htmlData

- Replaced naive `value.split(',')` with a brace-depth-aware parser that only splits on commas at depth 0
- Values wrapped in balanced outer braces have those braces stripped automatically
- Example: `\\htmlData{annotation_text={[a,b]}}{[a, b]}` sets `data-annotation_text="[a,b]"`
- All existing behavior preserved

## Test plan

- [x] Added 5 unit tests for `shouldRender` (true/false/not provided/child skipping/ignoredTags interaction)
- [x] Added 3 integration tests for `\\htmlData` with brace-wrapped commas (single pair, multiple pairs, unbalanced braces)
- [x] All 1295 tests pass locally
- [x] TypeScript type check passes
- [x] ESLint passes
- [x] Build succeeds (Rollup + Webpack)

## Notes for reviewers

- The `shouldRender` callback is invoked at the start of `renderElem` for every element, including the root element. This differs from `ignoredTags` which only applies to child elements.
- For `\\htmlData`, the brace-depth splitter preserves backward compatibility with all existing usage patterns. Only values that start and end with balanced outer braces will have those braces stripped; all other values are passed through unchanged.

No breaking changes — both additions are backward-compatible.

### DIFF
--- a/contrib/auto-render/auto-render.ts
+++ b/contrib/auto-render/auto-render.ts
@@ -12,6 +12,13 @@ interface RenderMathInElementOptions {
     errorCallback?: (msg: string, err: Error) => void;
     displayMode?: boolean;
     macros?: Record<string, string>;
+
+    /**
+     * An optional callback invoked for each element that is not excluded by
+     * `ignoredTags` or `ignoredClasses`. Return `false` to skip rendering
+     * math inside that element and its descendants.
+     */
+    shouldRender?: (node: HTMLElement) => boolean;
 }
 
 interface RenderMathInElementOptionsCopy {
@@ -22,6 +29,7 @@ interface RenderMathInElementOptionsCopy {
     errorCallback: (msg: string, err: Error) => void;
     displayMode?: boolean;
     macros?: Record<string, string>;
+    shouldRender?: (node: HTMLElement) => boolean;
 }
 
 /* Note: optionsCopy is mutated by this method. If it is ever exposed in the
@@ -78,6 +86,10 @@ const renderElem = function(
     elem: HTMLElement,
     optionsCopy: RenderMathInElementOptionsCopy
 ) {
+    if (optionsCopy.shouldRender &&
+            !optionsCopy.shouldRender(elem)) {
+        return;
+    }
     for (let i = 0; i < elem.childNodes.length; i++) {
         const childNode = elem.childNodes[i];
         if (childNode.nodeType === 3) {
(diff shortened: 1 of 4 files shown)


edemaine:
This looks like it should be two separate PRs.

ShyamKumar1:
@edemaine Good point, happy to split this into two PRs. I also reviewed the Greptile analysis and agree with both issues it raised:

1. **`shouldRender` double-fires** on non-root child elements — I'll drop the guard from the child loop and let `renderElem`'s own check handle it.

2. **Nested brace depth tracking** breaks on input like `{a,{b,c},d}` — the `depth === 0` guard on `{` increment needs to be removed.

I'll fix both, split them into separate PRs, and re-submit. Thanks for the review.

ShyamKumar1:
As suggested, I've split this into two separate PRs and fixed the bugs Greptile flagged:

- **shouldRender callback** → #4237 — fixed the double-fire bug (guard removed from child loop)
- **\htmlData braced commas** → #4238 — fixed nested brace depth tracking

Closing this PR in favor of those two.

ShyamKumar1:
Replaced by #4237 and #4238

Issue #4234, opened 2026-06-23
KaTeX in Rustdoc


attack68:
If I render the following rust docs:

```rust
/// \\( \frac{\partial \mathbb{E^Q} [\bar{C}_t] }{\partial \xi} = \frac{Nd}{10000} ( Q(m_{a.e.}) V(m_t) + I_{pa} X_{pa} V(m_{a.e.}) ) \\)
///
/// =
///
/// \\( \frac{Nd}{10000} ( Q(m_{a.e.}) V(m_t) + I_{pa} X_{pa} V(m_{a.e.}) ) \\)
```` 

I get a valid render:

<img width="263" height="88" alt="Image" src="https://github.com/user-attachments/assets/4c200729-ce3a-4af0-8254-55e98116313e" />

If I combine the exact same symbols into a single line:

```rust
/// \\( \frac{\partial \mathbb{E^Q} [\bar{C}_t] }{\partial \xi} = \frac{Nd}{10000} ( Q(m_{a.e.}) V(m_t) + I_{pa} X_{pa} V(m_{a.e.}) ) = \frac{Nd}{10000} ( Q(m_{a.e.}) V(m_t) + I_{pa} X_{pa} V(m_{a.e.}) ) \\)
```
then it fails.
<img width="673" height="53" alt="Image" src="https://github.com/user-attachments/assets/07943d51-7b47-4306-acde-ea44ab3387a8" />

This happens frequently with many different equations in my project. I would like to understand the issue so that I can try to avoid it.

 I am using a KateX header file:

```
<!--<link rel="preconnect" href="https://jsdelivr.net">-->
<link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/katex@0.16.45/dist/katex.min.css" integrity="sha384-UA8juhPf75SzzAMA/4fo3yOU7sBJ0om7SCD2GHq0fZqZco6tr1UCV7nUbk9J90JM" crossorigin="anonymous">
<!-- The loading of KaTeX is deferred to speed up page rendering -->
<script defer src="https://cdn.jsdelivr.net/npm/katex@0.16.45/dist/katex.min.js" integrity="sha384-Tt7wBxLKwSzFVRET4O4U9H6v8MNaQ/CjN2FMP4xFm0ErrFu6aNqoonRVW5W40iGI" crossorigin="anonymous"></script>
<!-- To automatically render math in text elements, include the auto-render extension: -->
[...]

PR #4235, opened 2026-06-23
fix(auto-render): recover math split across <em> tags from Markdown

chaubenn:
Addressing #4234, where two equations that each render fine on their own fail the moment you combine them into a single `\(...\)` line.

### Tracing the issue

1. The TeX itself wasn't the problem — feeding the full expression straight to `katex.renderToString` works. The breakage happens earlier, in how the math reaches KaTeX. rustdoc (and other Markdown preprocessors) run **before** KaTeX and rewrite the source: an underscore subscript like `\(\bar{C}_t\)` reads as emphasis, so the underscores get consumed and the text between them turns into an `<em>`:

   ```html
   \(\bar{C}<em>t] ... Q(m</em>{a.e.}) ... \)
   ```

2. Now `\(` and `\)` sit in different DOM nodes. auto-render only stitches together adjacent **text nodes** (to undo WebKit splitting large ones), so it walks right past the `<em>` and never sees a complete delimiter pair. The single-line case fails because that longer run is the one that trips the `<em>` boundary; the split version survives only because each `\(...\)` happens to land inside one text node. And once the underscores are gone from the HTML, nothing downstream can rebuild the subscripts.

### New behaviour

1. In that same sibling-walk in `renderElem`, auto-render now also absorbs an adjacent `<em>` and wraps its text back in `_`, undoing exactly what the Markdown emphasis did. The delimiters reunite and the formula renders identically to clean LaTeX.
   - An `<em>` that already holds a full delimiter pair (e.g. `see <em>\(x^2\)</em> here`) is left untouched for the normal recursion, so deliberately emphasised math keeps working.

2. Tests and docs:
   - Regression tests use the exact rustdoc-style HTML from #4234 and assert it renders to the same `innerHTML` as the un-corrupted `\(...\)` source — red on `main`, green here. Full suite still passes.
   - [`docs/autorender.md`](docs/autorender.md) documents the behavior and its one caveat: the recovery assumes `_..._` emphasis, so asterisk emphasis (`*...*`) inside math would be restored with underscores too. Escaping `\_` or keeping each formula in its own `\(...\)` span is the workaround.

### DIFF
--- a/contrib/auto-render/README.md
+++ b/contrib/auto-render/README.md
@@ -5,4 +5,5 @@ searches all of the text nodes in a given element for the given delimiters, and
 renders the math in place.
 
 See [Auto-render extension documentation](https://katex.org/docs/autorender.html)
-for more information.
+for more information, including how math split into `<em>` tags by Markdown
+preprocessors such as rustdoc is recovered.
--- a/contrib/auto-render/auto-render.ts
+++ b/contrib/auto-render/auto-render.ts
@@ -82,20 +82,39 @@ const renderElem = function(
         const childNode = elem.childNodes[i];
         if (childNode.nodeType === 3) {
             // Text node
-            // Concatenate all sibling text nodes.
-            // Webkit browsers split very large text nodes into smaller ones,
-            // so the delimiters may be split across different nodes.
+            // Concatenate adjacent sibling text nodes. WebKit browsers split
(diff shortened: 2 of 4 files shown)


edemaine:
I don't think this kind of fix/hack makes sense for KaTeX.

chaubenn:
Thanks for the comment & fair point @edemaine - will keep this in mind for the future

PR #4238, opened 2026-06-24
fix(parser): handle brace-wrapped values with commas in \htmlData

### LINKED ISSUE #4204: htmlData values cannot contain commas

ShyamKumar1:
## Summary

Fixes #4204: `\htmlData` now correctly parses key-value pairs within brace-delimited groups so that values can contain literal commas.

## Changes

- Replaced naive `value.split(',')` with a brace-depth-aware parser that only splits on commas at depth 0
- Parser correctly handles **nested braces** (e.g. `{a,{b,c},d}`)
- Values wrapped in balanced outer braces have those braces stripped automatically
- Example: `\htmlData{annotation_text={[a,b]}}{[a, b]}` sets `data-annotation_text="[a,b]"`
- All existing behavior preserved

## Tests

- Added tests for: single brace-wrapped value, multiple pairs, unbalanced braces, nested braces
- All 1291 tests pass locally (this branch)

## Notes

- The brace-depth splitter preserves backward compatibility with all existing usage patterns. Only values that start and end with balanced outer braces will have those braces stripped; all other values are passed through unchanged.
- No breaking changes.

### DIFF
--- a/src/functions/html.ts
+++ b/src/functions/html.ts
@@ -49,7 +49,29 @@ defineFunction({
                 };
                 break;
             case "\\htmlData": {
-                const data = value.split(",");
+                // Split on commas at brace depth 0 so that users can wrap
+                // values containing literal commas in braces, e.g.
+                //   \htmlData{annotation_text={[a,b]}}{[a, b]}
+                const data = [];
+                let current = "";
+                let depth = 0;
+                for (let i = 0; i < value.length; i++) {
+                    const ch = value[i];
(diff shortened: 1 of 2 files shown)


grigoriy-reshetniak:
Resolved in #4236

PR #4243, opened 2026-07-07
feat: handle character metrics warnings via `strict`

### LINKED ISSUE #1950: Don`t print warnings in production
https://github.com/KaTeX/KaTeX/blob/a51dc4b0727dad43187e8cf006cf1aeb32cb4de3/src/buildCommon.js#L82

Have you plans to implement this functionality? There are a lot of warnings about symbols like `№` in production, so that would be very helpful :)

### LINKED ISSUE #3720: KaTeX logs warnings to console in production
**Describe the bug:**

My production server logs are littered with thousands of lines like:

```
[2022-09-22 07:24:52] No character metrics for '🐦' in style 'Main-Regular' and mode 'text'
[2022-09-22 07:24:52] No character metrics for '🦅' in style 'Main-Regular' and mode 'text'
```

**To Reproduce:**
1. Go to [katex.org](https://katex.org/)
2. Paste the LaTeX string: `\sqrt{\char"1F642} = \text{?}`
3. Open dev console
4. See warning: `No character metrics for '🙂' in style 'Main-Regular' and mode 'math'`

**Expected behavior:**
In production, problems should not result in console logs, at least not without some way to turn this behavior off.

**Screenshots:**

 - KaTeX Version: 0.16.2
 - Device: Desktop
 - OS: Microsoft Windows 11 Home
 - Browser: Chrome 
 - Version: 105.0.5195.127 (Official Build) (64-bit)

**Cause:**

Here's an example `console.warn`: https://github.com/KaTeX/KaTeX/blob/299584d/src/buildCommon.js#L74-L76

```js
        // TODO(emily): Figure out a good way to only print this in development
        typeof console !== "undefined" && console.warn("No character metrics " +
            `for '${value}' in style '${fontName}' and mode '${mode}'`);
```

But the problem is more general, as `emily` notes in the above comment.

### LINKED ISSUE #3853: Don't render character metrics console warnings on production

grigoriy-reshetniak:
`No character metrics...` was a `console.warn()` which occurred when character was missing a metrics and couldn't be disabled.

Created a new error code for `strict`, -- `symbolNotInFont`, which now could be controlled via `strict`:  disabled, `console.warn` or throw a `ParseError`. 
However, while implementing this some other things were touched:
- Existing error codes were missing from `Settings` type declarations and docs, so I've added them.
- I've basically merged and extracted `reportNonstrict` and `useStrictBehavior`, because otherwise I would need a `Settings` instance. I kept those methods in place so I wont have to touch call sites in this PR.
- I've implemented [this TODO](https://github.com/KaTeX/KaTeX/blob/main/src/buildCommon.ts#L62), but as result I had to unwrap [this condition](https://github.com/KaTeX/KaTeX/pull/4243/changes#diff-d80498b1a0a3f084055a8911268b445da16bb1148df8ac5271a14fc0d6e9b169L91). Before there were a clear distinction between call sites which are passing `Options` and `delimiter.ts` which didn't. Tests are green, screenshotter has passed, and Claude was convincing me that it won't have an impact, but this looks a little dangerous anyway.

It's also worth mentioning, that now input like € produces 2 warnings at once 
image

If `strict` is set to `error`, only `unknownSymbol` would be thrown.

**BREAKING CHANGE**: Now characters without metrics will throw `ParseError`s. There's edge cases like `\origof`, `\imageof` which are listed as supported functions, but would throw if strict is set to `error`.

Fixes #3853, fixes #3720, fixes #1950, replacement for #3854


package.json
```diff
@@ -152,7 +152,7 @@
     "clean-install": "pnpm clean && pnpm install",
     "start": "webpack serve --config webpack.dev.js",
     "analyze": "webpack --config webpack.analyze.js",
-    "build": "rimraf dist/ && mkdirp dist && cp README.md dist && rollup -c --failAfterWarnings && webpack && node update-sri.js package dist/README.md",
+    "build": "NODE_ENV=production rimraf dist/ && mkdirp dist && cp README.md dist && rollup -c --failAfterWarnings && webpack && node update-sri.js package dist/README.md",
```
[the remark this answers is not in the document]
grigoriy-reshetniak: `NODE_ENV` removed.

src/buildCommon.ts
```diff
@@ -82,9 +82,10 @@ export const makeSymbol = function(
             value, metrics.height, metrics.depth, italic, metrics.skew,
             metrics.width, classes);
     } else {
-        // TODO(emily): Figure out a good way to only print this in development
-        typeof console !== "undefined" && console.warn("No character metrics " +
-            `for '${value}' in style '${fontName}' and mode '${mode}'`);
+        if (typeof process !== "undefined" && process.env.NODE_ENV !== "production") {
+            console.warn("No character metrics " +
+                `for '${value}' in style '${fontName}' and mode '${mode}'`);
+        }
```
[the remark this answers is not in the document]
grigoriy-reshetniak: I would argue against it, because https://github.com/KaTeX/KaTeX/blob/main/CONTRIBUTING.md?plain=1#L36 and https://github.com/KaTeX/KaTeX/pull/3854#issuecomment-2105558527 suggest that the warning is targeting contributors, and they kind of expected to use `pnpm start`. At the same time, 
> context where KaTeX is loaded without a bundler (CDN <script> tag, browser ESM import without a Node-polyfill)

looks a lot like "production" environment, so omitting the warning here is the intent of this PR.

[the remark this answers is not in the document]
grigoriy-reshetniak: @greptile any counter-argument?

edemaine (commented):
I'm a little uncertain about this behavior, mainly the no-warnings-in-browser aspect. Most users probably don't understand the message, so I guess it makes some sense to remove it, but it has a useful meaning to users not just developers: when users type unsupported Unicode characters in a KaTeX environment (such as the demo — does this PR affect the demo?), it's maybe helpful to know that the characters are probably being rendered from an unknown font with unknown metrics, so bad layouts are likely.

Another approach I just thought of is to add a `warn` setting that gets called in case of warnings like this. It'd be similar to `strict` — maybe it should actually just be integrated into `strict`? Then the user could explicitly ignore, use built-in warn behavior, even reject unknown symbols, or (perhaps most useful) render warnings in their GUI using their own styling. What do you think?

Separately I guess we could imagine changing the default `strict` (or `warn`) setting according to `NODE_ENV`, though that feels a bit too magical.

grigoriy-reshetniak:
@edemaine thank you for feedback and sorry for delay on this.

I've looked into idea of creating new setting vs merging it with `strict`, and I'm leaning towards separate setting. Here's why:
- It's not related to LaTeX faithfulness mode as stated in [docs](https://github.com/KaTeX/KaTeX/blob/main/docs/options.md?plain=1#L30). Although, as I understand it, it's already the case for [`unknownSymbol`](https://github.com/KaTeX/KaTeX/blob/main/src/Parser.ts#L995), I believe new setting could be expanded to`unknownSymbol` as well .
- It would be a _breaking_ change. It wouldn't really break anything, but users who have `strict` set  to `true` would suddenly get more errors than before.

Binding `strict` (or a new setting) to `NODE_ENV` is indeed feels to magical and breaks the pattern for other default values.

WDYT should be new target of this PR, new setting and handling of `No character metrics` or I can expand it to `Unrecognized Unicode character` as well?

edemaine:
It is *related* to LaTeX faithfulness, similar to `unknownSymbol`: LaTeX (with `unicode-math.sty` etc. or XeLaTeX) defines a set of allowed Unicode symbols, and others will fail: [unicode-math](https://tug.ctan.org/macros/unicodetex/latex/unicode-math/unicode-math.pdf) has its own fallback behavior, while XeTeX emits explicit [missing-character warnings](https://tug.org/pipermail/xetex/2022-April/028026.html). In KaTeX, missing character metrics generally means unrecognized symbol (see below).

That said, we're not (currently) maintaining an exact list of what LaTeX supports and diffing against that, which would be the original spirit of `strict`: if it works with strict KaTeX, then it should work in LaTeX.

I was actually not sure of the difference between missing metrics and `unknownSymbol`. The main difference is that `unknownSymbol` is suppressed when the character is within “supported script” ranges (ASCII, Latin Extended, Cyrillic, CJK, etc.), whereas metrics is about whether it's actually in the font (assuming metrics are complete), though there are some exceptions like `Ø` in math mode. Some examples (via GPT):

Input | `unknownSymbol` | Missing metrics warning
-- | -- | --
Literal €, default settings | Yes | Yes
Literal €, strict: false | No | Yes
\char"1F642 | No | Yes
Literal ⊶ or ⊷ | No | Yes
\textit{£} | No | Yes
Literal Ø in math mode | Yes | No
Literal 試 | No; `unicodeTextInMathMode` instead | No

I still lean toward extending `strict` because it offers the desired `"ignore"`/`"warn"`/`"error"`/callback interface, and because `unknownSymbol` establishes precedent. The new case could be called `missingCharacterMetrics` (literal issue) or something bolder like `glyphlessSymbol` or `symbolWithoutGlyph` or `symbolNotInFont` (more user-facing, assuming that metrics tables match the KaTeX fonts).

I think we can "fix" the discrepancy on the meaning of `strict` by changing the documentation to mention strict symbol handling in addition to strict LaTeX compatibility.

Regarding breaking change: I didn't imagine many people use `strict: "error"`; I am probably one of the users most concerned with LaTeX compatibility, and I've never used it. 😅 I always imagined using it conditionally as a compatibility pass, but even then, I'd probably collect errors with a callback and display those. That said, there do seem to be several uses on GitHub: [`strict: "error"`](https://github.com/search?q=katex+%22strict%3A+%5C%22error%5C%22%22&type=code)  and [`strict: true`](https://github.com/search?q=katex+%22strict%3A+true%22&type=code). So it is probably indeed helpful to mark this as breaking change, if we go this route.

I'm still open to alternatives though. Do you want to flesh out what your new-option proposal would actually look like, with or without moving `unknownSymbol`? Note that moving `unknownSymbol` out of `strict` would also be a breaking change (a potentially bigger one).

Somewhat related, I think KaTeX's docs would benefit from these warnings turning into errors, as they probably indicate something is wrong, so should cause CI to fail. (I'm pretty sure we've had docs bugs in the past from the lack of this.) Maybe with some exceptions, which we could put in the `strict` callback.

GPT also spotted an astral-character bug: `unknownSymbol` uses `charCodeAt(0)`, so an emoji is reported as its leading surrogate rather than its full code point. Maybe we can fix this at the same time.

grigoriy-reshetniak:
@greptile, PR scope has changed drastically, please do a new code review.

grigoriy-reshetniak:
> moving unknownSymbol out of strict would also be a breaking change (a potentially bigger one)

This was a very convincing point. My whole argument was built around minimizing breaking changes, but if one is introduced either way, I'd rather reuse `strict`,  especially since the precedent `unknownSymbol` with  is already set. 

So I implemented it within `strict` and I'm looking forward to your feedback.

PS. For that astral character bug, I left it out for now. I would like to give it attention together with [this](https://github.com/KaTeX/KaTeX/blob/main/src/functions/char.ts#L22) and other places where `String.fromCodePoint()` could be used.

docs/options.md
edemaine: We need to update this. Maybe something like:

```suggestion
- `strict`: `boolean` or `string` or `function` (default: `"warn"`). Control strict LaTeX faithfulness and symbol handling. If `false` or `"ignore`", allow arbitrary Unicode symbols which might lay out poorly, and allow features that make writing LaTeX convenient but are not actually supported by (Xe)LaTeX (similar to MathJax). If `true` or `"error"`, throw an error for any such transgressions, enforcing symbols with known metrics and input that should render faithfully in LaTeX. If `"warn"` (the default), warn about such behavior via `console.warn`. Provide a custom function `handler(errorCode, errorMsg, token)` to customize behavior depending on the type of transgression (summarized by the string code `errorCode` and detailed in `errorMsg`); this function can also return `"ignore"`, `"error"`, or `"warn"` to use a built-in behavior.  A list of such features and their `errorCode`s:
```

src/buildCommon.ts
```diff
@@ -529,7 +529,7 @@ const getVListChildrenAndDepth = function(params: VListParam): {
     } else {
         const firstChild = params.children[0];
         if (firstChild.type !== "elem") {
-            throw new Error('First child must have type "elem".');
+            throw new Error("First child must have type \"elem\".");
```
edemaine: Are all these typographic changes intentional, or an editor gone haywire? This last one seems less than ideal, unless it's somehow enforced by our linting settings or something. Some of the spacing changes seem less pretty too. But not a big deal either way.

src/strict.ts
```diff
@@ -0,0 +1,76 @@
...
+                `${errorMessage} [${errorCode}]`);
+
+            if (!report) {
+                return false;
+            }
+            break;
+        case undefined:
+        case false:
+        case "ignore":
+        default:
```
edemaine, line 71: **Unrecognized strict settings now fail silently.** This `default` discards the previous warning for unexpected truthy values. For example, `strict: "erorr"` or a callback returning that value previously warned about the unrecognized setting; now it silently ignores violations. We should probably preserve the old fallback warning.

src/strict.ts
```diff
@@ -0,0 +1,76 @@
...
+import ParseError from "./ParseError";
+
+export type Strict = boolean | "ignore" | "warn" | "error" | StrictFunction;
+
+export type StrictFunction =
+    (errorCode: string, errorMsg: string, token?: Token | AnyParseNode) =>
+        (boolean | "ignore" | "warn" | "error") | undefined;
+
+export type StrictParameters<R extends boolean = boolean> = {
+    strictSetting: Strict;
```
edemaine, line 13: I expected this to be called `strict`...

src/buildCommon.ts
```diff
@@ -82,21 +81,22 @@ export const makeSymbol = function(
             value, metrics.height, metrics.depth, italic, metrics.skew,
             metrics.width, classes);
     } else {
-        // TODO(emily): Figure out a good way to only print this in development
-        typeof console !== "undefined" && console.warn("No character metrics " +
-            `for '${value}' in style '${fontName}' and mode '${mode}'`);
+        handleStrict({
```
edemaine, line 84: I guess we don't have `settings` here so we can't use the usual methods? Makes me wonder whether the methods should be removed in general and switch to `handleStrict` approach...

src/strict.ts
```diff
@@ -0,0 +1,76 @@
...
+    errorCode: string;
+    errorMessage: string;
```
edemaine: I expected this to be called `errorMsg` like elsewhere.

src/strict.ts
```diff
@@ -0,0 +1,76 @@
...
+    report: R;
+    token?: Token | AnyParseNode;
+};
+
+export function handleStrict(params: StrictParameters<true>): void;
```
edemaine, line 36: Needs a comment explaining what this does. Possibly replacing the method comments...

src/strict.ts
```diff
@@ -0,0 +1,76 @@
...
+        case "ignore":
+        default:
+            if (!report) {
+                return false;
+            }
+            break;
+    }
+}
+
+const invokeStrictFunction = (
```
edemaine: Should this be inlined? The body is almost as long as the prototype. 🙂

edemaine (commented):
This design is looking good! I'm thinking now that this approach of getting `strict` via `options` should maybe replace the `Settings` methods, for simplicity? What do you think?

Also need to add tests.

src/strict.ts
grigoriy-reshetniak: Old fallback in both methods [had](https://github.com/KaTeX/KaTeX/pull/4243/changes#diff-c70796c51613f50f45a1df8ff713d8265da3888efd6e8fc5247f424c0732102dL452) a very specific comment 
> won't happen in type-safe code

Looking at it's [origin](https://github.com/KaTeX/KaTeX/commit/7ab4f76e16e3d3950062ffd6efa4a4b50f7fa5fe) I see that it was added 6.5 years prior to [type definitions](https://github.com/KaTeX/KaTeX/commit/ea9c173a0de953b49b2ce5d131e88b785f5dffa1). I assume that it was needed at the time, but modern TS user is already warned with 
> Type "erorr" is not assignable to type `boolean | "ignore" | "warn" | "error" | StrictFunction | undefined`

and it feels like duplicating type features by hand.

Given the chance, I would rather remove `undefined` from the union, because it doesn't serve as a valid value and can be handled in `default`.

src/buildCommon.ts
grigoriy-reshetniak: `Settings` instance is not available here, that was the reason of extracting code to the separate function which could be used with value from options. However, I felt restrained of removing existing methods because of magnitude of changes (e.g. `Lexer` now doesn't need `Settings` instance, only `strict` value. After that change it looks like `MacroExpander` doesn't need a whole `Settings`, because it's just need to pass `strict` and consume `maxExpand`), but given the green light, I'm happy to proceed.

src/strict.ts
edemaine: Fair enough. Maybe we can debate which option should be default though. I'd lean toward the actual default behavior, warn.

src/buildCommon.ts
edemaine: I haven't looked everything over closely, but it seems like it would make the code simpler, so it's worth investigating.

I guess there's a worry that we might need the rest of settings in those contexts in the future. So instead of replacing all if settings with just strict, I might prefer just replacing the method calls with the function call.

docs/options.md
grigoriy-reshetniak: Thanks for drafting this, updated.

src/buildCommon.ts
grigoriy-reshetniak: The editor applied those and since it was per linting config I kept it. they're drawing attention as less pretty I reverted it.

src/strict.ts
grigoriy-reshetniak: Indeed, `warn` is looking much better as a branch for default.

src/strict.ts
grigoriy-reshetniak: This will conflict with [local variable](https://github.com/KaTeX/KaTeX/pull/4243/changes#diff-d7ade4730b9921ec90d7cf792c55191b3412df606346438e474a43b9f92493bcR40). Looks like something must have a different name, please tell me what you prefer, interface for consumers or local variable.

src/buildCommon.ts
grigoriy-reshetniak: Done, `Settings` context is still presented everywhere.

grigoriy-reshetniak:
I've also added tests, but skipped `warn`. console.warn is [set up](https://github.com/KaTeX/KaTeX/blob/main/test/setup.ts#L50) to throw in the test environment. Test case
```
it("should print in console when strict is 'warn'", () => {
    expect(() => report("warn")).toThrow(
        `LaTeX-incompatible input and strict mode is set to 'warn': ` +
        `${errorMsg} [${errorCode}]`
    );
});
```
 looks misleading so I decided to omit it.

src/Parser.ts
```diff
@@ -943,9 +944,12 @@ export default class Parser {
...
             // This behavior is not strict (XeTeX-compatible) in math mode.
             if (this.settings.strict && this.mode === "math") {
-                this.settings.reportNonstrict("unicodeTextInMathMode",
-                    `Accented Unicode text character "${text[0]}" used in ` +
-                    `math mode`, nucleus);
+                handleStrict({
+                    strictSetting: this.settings.strict,
+                    errorCode: "unicodeTextInMathMode",
+                    errorMsg: `Accented Unicode text character "${text[0]}" used in math mode`,
+                    report: true,
```
edemaine, line 951: Missing `token: nucleus`

src/strict.ts
edemaine: I'd prefer the interface to be cleaner. Thanks!

edemaine (commented):
Looking good! A couple small comments.

GPT also notes that the behavior for a function returning `undefined` has changed; it [used to act like `"ignore"`](https://github.com/KaTeX/KaTeX/blob/2d2a17763cb068e2d453553c3b106d3c0e07a690/src/Settings.ts#L401) but now acts like `"warn"`. A function returning `undefined` was [previously allowed by the types too](https://github.com/KaTeX/KaTeX/blob/2d2a17763cb068e2d453553c3b106d3c0e07a690/types/katex.d.ts#L57). So I'm not sure whether it's worth breaking this; perhaps we should add back the `case undefined` for ignore, and just keep the change in `default` behavior. I could go either way with removing `undefined` from types though. I guess one motivation for allowing `undefined` is if you want to log issues but not have KaTeX take any action.

For testing, GPT suggests:
* Temporarily replace `console.warn` using `jest.spyOn(...).mockImplementation(...)`, assert the calls, and restore it afterward.
* Add coverage for callback `undefined`, token preservation, and actual missing-metrics rendering under ignore/warn/error.

PR #4245, opened 2026-07-11
chore: use pixel diff as fallback for byte-to-byte

grigoriy-reshetniak:
This PR is an alternative for https://github.com/KaTeX/KaTeX/pull/4228. If it comes down to pixel comparison anyway, maybe macOs runner can stay and migration for `webkit` is not needed.


dockers/screenshotter/screenshotter.js
```diff
@@ -18,6 +18,8 @@ const selenium = require("selenium-webdriver");
 const firefox = require("selenium-webdriver/firefox");
 const chrome = require("selenium-webdriver/chrome");
 const seleniumHttp = require("selenium-webdriver/http");
+const PNG = require("pngjs").PNG;
+const pixelmatch = require("pixelmatch").default;
```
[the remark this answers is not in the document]
grigoriy-reshetniak: I've mentioned this in README.

edemaine (commented):
Thanks for putting together this alternative. I'm still not sure which approach is better, but here is some feedback from GPT 5.6 about tightening the comparison logic in this one:

---

1. **[P1] The “pixel-identical” fallback is perceptual, not exact.**

   [`screenshotter.js`](https://github.com/KaTeX/KaTeX/blob/53ecf2caa6a654c7b55534a22644af8e1b8ae95c/dockers/screenshotter/screenshotter.js#L547-L560) calls `pixelmatch` with `threshold: 0.1` and its default antialiasing exclusion, then treats `diffPixels === 0` as pixel identity. Pixelmatch can return zero for genuinely different RGBA data, so this weakens Chrome and Firefox comparisons too, even though only Safari passes `--threshold`.

2. **[P1] The Safari threshold masks a stable change and prevents retries or artifact review.**

   [The workflow passes `--threshold 0.001`](https://github.com/KaTeX/KaTeX/blob/53ecf2caa6a654c7b55534a22644af8e1b8ae95c/.github/workflows/screenshotter.yml#L80). For a 1024×768 screenshot, that permits up to 786 differing pixels.

   Both available Safari runs accepted exactly 267 differing pixels for `Boxed`: [[run 1](https://github.com/KaTeX/KaTeX/actions/runs/29164588443/job/86575399769)](https://github.com/KaTeX/KaTeX/actions/runs/29164588443/job/86575399769) and [[run 2](https://github.com/KaTeX/KaTeX/actions/runs/29165012638/job/86576520115)](https://github.com/KaTeX/KaTeX/actions/runs/29165012638/job/86576520115). The identical count strongly suggests a stable Safari/baseline change (update `Boxed`) rather than asynchronous rendering.

   Because the threshold path immediately returns success, `Boxed` is never retried and the changed screenshot is not uploaded for inspection. I would let it fail, review the artifact, and update `Boxed-safari.png` if the Safari change is legitimate. If tolerance remains necessary, it should only be considered after exact-pixel retries and should still produce a reviewable artifact.

3. **[P2] The comparison behavior needs focused tests.** [Erik's opinion: not important.]

   The screenshot workflow exercises successful comparisons, but it does not prove that real regressions are rejected. I would extract a comparison helper and cover at least:

   - identical pixels with different PNG encodings;
   - one or more changed pixels;
   - dimension mismatches;
   - antialiasing/color differences;
   - any retained threshold boundary.

4. **[P2] The ESM dependency creates an avoidable local Node requirement.** [Erik's opinion: not important.]

   `pixelmatch@7` is ESM-only, while the screenshotter uses `require("pixelmatch").default`. CI’s Node 24 supports this, but older local Node versions do not. The README caveat does not identify a minimum version. Exact decoded comparison using only `pngjs` would avoid this issue.

GPT's preferred approach is a narrower version of this PR:

- keep native Safari on macOS;
- compare decoded RGBA data exactly (via `pngjs`, no `pixelmatch`):
   ```js
   imgA.width === imgB.width &&
   imgA.height === imgB.height &&
   imgA.data.equals(imgB.data)
   ```
- port `document.fonts.ready` and `img.decode()` from #4228;
- retry every genuine pixel mismatch;
- inspect and update the stable `Boxed` baseline;
- add focused comparator tests.

This preserves real Safari coverage and solves the PNG-encoding problem without weakening the regression suite.

---

What do you think? I do like more accurately reproducing actual Safari behavior, which is the browser most people will use. I guess the fear is that we need to update screenshots every time the browser changes. But maybe it won't be too often?...

grigoriy-reshetniak:
> compare decoded RGBA data exactly (via `pngjs`, no `pixelmatch`)

This approach failed: https://github.com/KaTeX/KaTeX/actions/runs/29610755168/job/87984596753?pr=4245, going back to `pixelmatch`, but removing all threshold.

grigoriy-reshetniak:
`{threshold: 0.05}` is required for pipeline to succeed. It means that for every pixel, anti-aliasing/subpixel color shifts smaller than ~5% of max color distance are tolerated. If single pixel color shift would be more than 5% test will fail.
It seems pretty tight. 

Removing threshold or compare with `pngjs` seems to do same thing as byte-to-byte comparison and doesn't make sense as an additional check.

edemaine (changes requested):
Looks great! The removal of the whole-image difference allowance and the updated `Boxed` Safari baseline address the main concerns. In particular, it is reassuring that `threshold: 0.05` still rejected the old `Boxed` baseline as 267 differing pixels on every retry. The tolerance now seems appropriately tight for Safari’s small rasterization differences.

GPT has two final suggestions:

1. Limit the perceptual `pixelmatch` fallback to Safari. It currently runs in the shared screenshot path, so Chrome and Firefox could also accept perceptually similar images whenever their PNG bytes differ. Suggest accepting exact decoded RGBA equality for every browser, then allowing the thresholded comparison only for Safari:

   ```js
   if (imgA.data.equals(imgB.data)) {
       // Exact pixels, different PNG encoding.
       return;
   }

   if (
       opts.browser === "safari" &&
       pixelmatch(
           imgA.data,
           imgB.data,
           null,
           imgA.width,
           imgA.height,
           {threshold: 0.05}
       ) === 0
   ) {
       // Tolerated Safari rasterization difference.
       return;
   }
   ```

2. Rename `pixel-identical` in the success message to something like `perceptually equivalent`. With a nonzero Pixelmatch threshold and its default antialiasing exclusion, a zero difference count does not mean the decoded pixels are literally identical.

With those small changes, this approach looks ready to merge.

dockers/screenshotter/screenshotter.js
```diff
@@ -532,14 +534,32 @@ async function takeScreenshot(key) {
...
+                        imgA.width, imgA.height,
+                        {threshold: 0.05} // per-pixel color sensitivity
+                    ) === 0) {
+                    console.log(
+                        `* ok ${key}: byte mismatch, but perceptually equivalent`
+                    );
+                    return;
+                }
+            }
+            console.log(`error ${key}`);
```
edemaine: Should we still include the number of differing pixels in this case, by duplicating log inside `if` and in `else`? Not a big deal either way.

edemaine (approved):

grigoriy-reshetniak:
I've applied suggestions, but there's one thing I disagree with:

```
if (imgA.data.equals(imgB.data)) {
    // Exact pixels, different PNG encoding.
    return;
}
```

outside of check with `opts.browser === "safari"`. There's real small chance that this would ever work, because either byte-t-byte would `return`, or if it fails -- `imgA.data.equals(imgB.data)` would fail too.

P.S.: started this comment before the approval.

dockers/screenshotter/screenshotter.js
grigoriy-reshetniak: I went with GPT suggestion and that approach eliminated separate variable for pixel diff. But I liked that too and I will return it.

PR #4249, opened 2026-07-22
docs: add working math examples to browser starter template

### LINKED ISSUE #3662: Add basic examples to body of Starter Template
### Discussed in https://github.com/KaTeX/KaTeX/discussions/3660

<div type='discussions-op-text'>

<sup>Originally posted by **mbourne** June 15, 2022</sup>
It's been a long while since I used KaTeX "from scratch" so I made use of the Starter Template on this page: https://katex.org/docs/browser.html

I placed some equations in the `body` and loaded the page, but it just showed the equations as text. It turned out it was the issue I reported on the default delimiters - the environments I was using happened to not be in that default list. There was no error message since KaTeX just thought the page was devoid of math and I was stumped until I tried a few `$$ ... $$` equations, which did work OK.

It would have helped if the Starter Template was a bit more extensive and included a `body` with some examples that will actually work out of the box, and a note that if other environments were required, to add an `options` object, as per https://katex.org/docs/autorender.html (or, if all supported environments were included in the defaults, as I mentioned.)

This will help people who are brand new to the library, I believe.

I'm sorry if this is documented somewhere and I missed it.</div>

@mbourne feel free to submit a PR that replaces the `...` in https://katex.org/docs/browser.html#starter-template with a `<body>` that has some actual math in it!

Hashim1999164:
## Summary
- Replace the `...` placeholder in the browser starter template with a `<body>` that includes inline (`\(...\)`) and display (`$$...$$`) math examples that work with default auto-render delimiters.
- Note that single `$...$` is not enabled by default, and point to the auto-render docs for customizing delimiters via `options`.

Fixes #3662

## Test plan
- [ ] Open the Starter template section on the browser docs page and confirm the HTML example includes a body with math.
- [ ] Copy the starter HTML into a local file and open it in a browser; confirm both inline and display math render.
- [ ] Confirm the note about default delimiters and the link to autorender docs are correct.

### DIFF
--- a/docs/browser.md
+++ b/docs/browser.md
@@ -20,9 +20,21 @@ title: Browser
     <script defer src="https://cdn.jsdelivr.net/npm/katex@0.18.1/dist/contrib/auto-render.min.js" integrity="sha384-bjyGPfbij8/NDKJhSGZNP/khQVgtHUE5exjm4Ydllo42FwIgYsdLO2lXGmRBf5Mz" crossorigin="anonymous"
         onload="renderMathInElement(document.body);"></script>
   </head>
-  ...
+  <body>
+    <p>Inline math: \(e^{i\pi} + 1 = 0\)</p>
+    <p>Display math:</p>
+    $$
+    \int_{-\infty}^{\infty} e^{-x^2} \, dx = \sqrt{\pi}
+    $$
+  </body>
 </html>
 ```
+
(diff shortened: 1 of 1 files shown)


Hashim1999164:
Wrapped the display math example in a `<p>` so the starter template is valid HTML, matching the inline example.

PR #4255, opened 2026-07-29
fix: allow braced arguments for delimsizing

### LINKED ISSUE #3828: `\big{[}` not working
MWE:

```latex
\big{[}
```

LaTeX output: Same as `\big[`.

Apparently the argument of `\big` isn't correctly parsed here.

grigoriy-reshetniak:
`delimsizing` functions accepted only unbraced arguments, like `\big|`

`delimsizing` functions also accept braced arguments e.g. `\big{|}`

image

**Mismatch with LaTeX behavior**:
LaTeX silently fails on multi-symbol group delimiters like `\big{||}`, first delimiter would be _big_ while other remains regular size. In current implementation KaTeX will fail with 
> Invalid delimiter type 'ordgroup'

Which, to me, doesn't look that bad because it clearly signals that the argument format is wrong, but it's still a mismatch to consider.

Fixes #3828

### DIFF
--- a/test/katex-spec.ts
+++ b/test/katex-spec.ts
@@ -1050,6 +1050,7 @@ describe("A delimiter sizing parser", function() {
         const bare = getParsed(normalDelim)[0];
 
         expect(braced.delim).toEqual(bare.delim);
+        expect(normalDelim).toParseLike(bracedDelim);
     });
 
     it("should produce the correct direction delimiter", function() {
--- a/test/katex-spec.ts
+++ b/test/katex-spec.ts
@@ -1023,20 +1023,33 @@ describe("A delimiter sizing parser", function() {
     const normalDelim = r`\bigl |`;
     const notDelim = r`\bigl x`;
     const bigDelim = r`\Biggr \langle`;
+    const bracedDelim = r`\bigl{|}`;
 
     it("should parse normal delimiters", function() {
         expect(normalDelim).toParse();
         expect(bigDelim).toParse();
     });
 
+    it("should accept a single braced delimiter", function() {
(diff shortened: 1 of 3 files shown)


test/katex-spec.ts
```diff
@@ -1039,6 +1045,13 @@ describe("A delimiter sizing parser", function() {
         expect(parse.type).toEqual("delimsizing");
     });
 
+    it("should produce the same delim for braced and unbraced forms", function() {
+        const braced = getParsed(bracedDelim)[0];
+        const bare = getParsed(normalDelim)[0];
+
+        expect(braced.delim).toEqual(bare.delim);
```
edemaine, line 1052: I think `toParseLike` should also be true, right? Maybe an additional test like ``expect`\left{(}x\right{)}`.toParseLike`\left(x\right)` ``

edemaine (approved):
Nice and simple! I did not recall `normalizeArgument`.

test/katex-spec.ts
grigoriy-reshetniak: I've added `toParseLike` expect for `\big`, but neither `\left` nor `\right` is affected by this PR and accept braced arguments. I've checked and in `pdflatex` `\left{(}x\right{)}` fails with
> ! Missing delimiter (. inserted).

and produces different output than `\left(x\right)`. 

Could you please confirm that my conclusion is correct and I should leave `\left`, `\right`, and `\middle` untouched?

edemaine: Great, thanks! Yes, I'd rather keep LaTeX compatibility with that (unless we get strong requests otherwise  and we could add a non-strict deviation).

PR #4256, opened 2026-07-30
fix(environment): report invalid environment names as ParseError

### LINKED ISSUE #3760: Katex fails and crashes entirely!
> Before reporting a bug

**Describe the bug:**
Go to the [Katex demo](https://katex.org/#demo). Insert the following text: `\begin{\pmatrix}`. Disable "throw on error". 

Now notice that this will crash Katex. This is not an error message that is generated by katex that can be rendered. It actually throws an exception. This breaks for example `react-katex` completely, making the entire page go blank!

**Expected behavior:**
It should not crash. Instead, it should return a nicely formatted error messages etc.

The rest is already explained since it is reproducible on the current and latest Katex Demo!

AtsushSaito:

`\begin` and `\end` build the environment name by asserting that every node in the name group is a `textord` (`src/functions/environment.ts`). `assertNodeType` throws a plain `Error`, not a `ParseError`, so an environment name that doesn't parse to plain characters is reported as an internal failure.

With `throwOnError: false` that plain `Error` escapes the error handling entirely and crashes the caller instead of rendering error markup, which is the blank `react-katex` page described in #3760:

```js
katex.renderToString("\\begin{\\pmatrix}", {throwOnError: false});
// Uncaught Error: Expected node of type textord, but got node of type color
```

This isn't specific to `\pmatrix`. Under the default settings `\begin{\pmatrix}` fails earlier on the undefined control sequence, which is why it looks fine there, but a name that reaches the assertion by another route throws the plain `Error` under the default settings too:

```js
katex.renderToString("\\begin{\\text{a}}");
// Uncaught Error: Expected node of type textord, but got node of type text
```

`\end` is affected the same way, since one handler serves both: `\begin{matrix}a\end{\pmatrix}` with `throwOnError: false` throws instead of rendering.

The same assertion also rejects `\begin{ma trix}`, because a space parses to a `spacing` node rather than a `textord`. LaTeX lets an environment be defined with a space in its name, so that name should be read and then reported as missing, not refused for holding a space.

Two other argument readers have the same shape and the same bug: the column count of `\begin{alignedat}{n}` and the argument of `\@char`. The column count is then read with `Number()`, which takes anything numeric, so `\begin{alignedat}{2.5}` asks for five columns and `\begin{alignedat}{0}` silently falls back to the automatic column counting of `aligned`.

`assertCharacterGroup()` in `src/parseNode.ts` reads a group that should hold nothing but plain characters and throws a `ParseError` for anything else. `\begin`/`\end`, `\begin{alignedat}{n}` and `\@char` all read their argument through it, so all three render the usual error markup under `throwOnError: false` instead of crashing the caller, and report a parse error rather than an internal failure under the default settings:

- `\begin{\pmatrix}`, `\begin{\text{a}}`, `\end{\pmatrix}` → `ParseError: Environment name should contain only text characters and spaces`
- `\begin{alignedat}{\text{2}}` → `ParseError: Number of columns should be a positive integer`
- `\@char{\error}` → `ParseError: \@char has non-numeric argument`

A literal space counts as a character, so `\begin{ma trix}` now reads the name and reports `No such environment: ma trix`. Only a literal space does: `~` and `\ ` parse to `spacing` nodes as well but are not spaces in a name, so they stay rejected.

Whether a space is a character is the caller's decision rather than the helper's, so only `\begin` and `\end` opt in. `\@char` is allowed in text mode, where `\text{\@char{6 5}}` would otherwise reach `parseInt` as `6 5` and silently render U+0006.

The column count of `\begin{alignedat}{n}` now has to be a positive integer, so `{2.5}` and `{0}` are reported instead of being taken as five columns and as no column count at all.

The `ordgroup` check above the environment name keeps its own `Invalid environment name` message: that is a different failure, the argument not being a group at all.

`test/errors-spec.ts` covers the cases that reach these guards under the default settings. `test/katex-spec.ts` covers `\begin{\pmatrix}`, `\begin{\error}`, `\end{\pmatrix}`, `\begin{alignedat}{\error}` and `\@char{\error}` under `throwOnError: false`, which is the path this escaped. All of them fail without the source change. Full Jest suite, `tsc` and lint pass.

Two more argument readers have the same plain `Error`, and I've left both for a follow-up because neither has the shape this helper reads: `\genfrac` looks at only the first node of its style group, so reading the whole group through the helper would change what `\genfrac{}{}{}{01}{a}{b}` means, and the column spec of `{array}` goes through `assertSymbolNodeType` per node, where the existing `Unknown column alignment` `ParseError` is the report to reach for instead.

Fixes #3760

### DIFF
--- a/src/environments/array.ts
+++ b/src/environments/array.ts
@@ -6,8 +6,12 @@ import defineFunction from "../defineFunction";
 import defineMacro from "../defineMacro";
 import {MathNode} from "../mathMLTree";
 import ParseError from "../ParseError";
-import {assertNodeType, assertSymbolNodeType} from "../parseNode";
-import {assertCharacterGroup, checkSymbolNodeType} from "../parseNode";
+import {
+    assertCharacterGroup,
+    assertNodeType,
+    assertSymbolNodeType,
+    checkSymbolNodeType,
+} from "../parseNode";
 import {Token} from "../Token";
 import {calculateSize, makeEm} from "../units";
 
@@ -734,8 +738,12 @@ const alignedHandler = function(context: EnvContextLike, args: AnyParseNode[]) {
         body: [],
     };
     if (args[0] && args[0].type === "ordgroup") {
-        numMaths = Number(
-            assertCharacterGroup(args[0], "Invalid number of columns"));
+        const message = "Number of columns should be a positive integer";
+        const numColumns = assertCharacterGroup(args[0], message);
+        if (!/^[0-9]+$/.test(numColumns) || Number(numColumns) < 1) {
+            throw new ParseError(message, args[0]);
+        }
+        numMaths = Number(numColumns);
         numCols = numMaths * 2;
     }
     const isAligned = !numCols;
--- a/src/environments/array.ts
+++ b/src/environments/array.ts
@@ -7,7 +7,7 @@ import defineMacro from "../defineMacro";
 import {MathNode} from "../mathMLTree";
 import ParseError from "../ParseError";
 import {assertNodeType, assertSymbolNodeType} from "../parseNode";
-import {checkSymbolNodeType} from "../parseNode";
+import {assertCharacterGroup, checkSymbolNodeType} from "../parseNode";
 import {Token} from "../Token";
 import {calculateSize, makeEm} from "../units";
 
@@ -734,12 +734,8 @@ const alignedHandler = function(context: EnvContextLike, args: AnyParseNode[]) {
         body: [],
     };
     if (args[0] && args[0].type === "ordgroup") {
-        let arg0 = "";
-        for (let i = 0; i < args[0].body.length; i++) {
-            const textord = assertNodeType(args[0].body[i], "textord");
-            arg0 += textord.text;
-        }
-        numMaths = Number(arg0);
+        numMaths = Number(
+            assertCharacterGroup(args[0], "Invalid number of columns"));
         numCols = numMaths * 2;
     }
     const isAligned = !numCols;
--- a/src/functions/char.ts
+++ b/src/functions/char.ts
@@ -1,6 +1,6 @@
 import defineFunction from "../defineFunction";
 import ParseError from "../ParseError";
-import {assertNodeType} from "../parseNode";
+import {assertCharacterGroup, assertNodeType} from "../parseNode";
 
 // \@char is an internal function that takes a grouped decimal argument like
 // {123} and converts into symbol with code 123.  It is used by the *macro*
@@ -13,12 +13,8 @@ defineFunction({
 
     handler({parser}, args) {
         const arg = assertNodeType(args[0], "ordgroup");
-        const group = arg.body;
-        let number = "";
-        for (let i = 0; i < group.length; i++) {
-            const node = assertNodeType(group[i], "textord");
-            number += node.text;
-        }
+        const number = assertCharacterGroup(
+            arg, "\\@char has non-numeric argument");
         let code = parseInt(number);
         let text;
         if (isNaN(code)) {
--- a/src/functions/environment.ts
+++ b/src/functions/environment.ts
@@ -19,7 +19,9 @@ defineFunction({
             throw new ParseError("Invalid environment name", nameGroup);
         }
         const envName = assertCharacterGroup(
-            nameGroup, "An environment name should be text characters", true);
+            nameGroup,
+            "Environment name should contain only text characters and spaces",
+            true);
 
         if (funcName === "\\begin") {
             // begin...end is similar to left...right
--- a/src/functions/environment.ts
+++ b/src/functions/environment.ts
@@ -18,9 +18,16 @@ defineFunction({
         if (nameGroup.type !== "ordgroup") {
             throw new ParseError("Invalid environment name", nameGroup);
         }
+        // An environment name is plain characters. Anything else -- a macro,
+        // a space -- is user error, so report it as a ParseError rather than
+        // an assertion failure that would escape `throwOnError: false`.
         let envName = "";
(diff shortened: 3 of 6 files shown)


src/functions/environment.ts
```diff
@@ -18,9 +18,16 @@ defineFunction({
         if (nameGroup.type !== "ordgroup") {
             throw new ParseError("Invalid environment name", nameGroup);
         }
+        // An environment name is plain characters. Anything else -- a macro,
```
grigoriy-reshetniak: I believe this comment is redundant. Instead, the error message could be something like `An environment name should be text characters`.

grigoriy-reshetniak (changes requested):

grigoriy-reshetniak:
Thanks! 

> The same plain Error reaches user input from four other argument readers, which I've left alone to keep this reviewable: the column count of \begin{alignat} in display mode, the argument of \@char, the style argument of \genfrac, and the column spec of {array}. The first three are the same assertNodeType(…, "textord") loop and could share a helper; {array} goes through assertSymbolNodeType instead. Happy to follow up with those.

It sounds like a good idea. @edemaine do you agree?

AtsushSaito:
@grigoriy-reshetniak 
Thanks, that's better than what I had — the message is the right place for it. Dropped the comment and used your wording, pushed as 7953f0ff.

grigoriy-reshetniak (approved):

edemaine:
Thanks for this! Agreed, it'd be nice to similarly guard the other argument consumers, especially if we can share a helper among them.

One note: `katex.renderToString("\\begin{ma trix}")` shouldn't reject on the basis of space. Space nodes with content `" "` ought to be treated like a space character. Environments can be defined with spaces in LaTeX.

grigoriy-reshetniak:
@greptile revise your review

src/environments/array.ts
```diff
@@ -7,7 +7,7 @@ import defineMacro from "../defineMacro";
 import {MathNode} from "../mathMLTree";
 import ParseError from "../ParseError";
 import {assertNodeType, assertSymbolNodeType} from "../parseNode";
-import {checkSymbolNodeType} from "../parseNode";
+import {assertCharacterGroup, checkSymbolNodeType} from "../parseNode";
```
grigoriy-reshetniak: It would be worth to merge these imports with the ones above.

src/environments/array.ts
```diff
@@ -734,12 +734,8 @@ const alignedHandler = function(context: EnvContextLike, args: AnyParseNode[]) {
...
     };
     if (args[0] && args[0].type === "ordgroup") {
-        let arg0 = "";
-        for (let i = 0; i < args[0].body.length; i++) {
-            const textord = assertNodeType(args[0].body[i], "textord");
-            arg0 += textord.text;
-        }
-        numMaths = Number(arg0);
+        numMaths = Number(
+            assertCharacterGroup(args[0], "Invalid number of columns"));
```
grigoriy-reshetniak: This could be improved to accept only positive integers with corresponding error message.

src/functions/environment.ts
```diff
@@ -18,10 +18,8 @@ defineFunction({
         if (nameGroup.type !== "ordgroup") {
             throw new ParseError("Invalid environment name", nameGroup);
         }
-        let envName = "";
-        for (let i = 0; i < nameGroup.body.length; ++i) {
-            envName += assertNodeType(nameGroup.body[i], "textord").text;
-        }
+        const envName = assertCharacterGroup(
+            nameGroup, "An environment name should be text characters", true);
```
grigoriy-reshetniak:
```suggestion
            nameGroup, "Environment name should contain only text characters and spaces", true);
```

src/parseNode.ts
```diff
@@ -42,3 +43,33 @@ export function checkSymbolNodeType(node: AnyParseNode): SymbolParseNode | null
     }
     return null;
 }
+
+/**
+ * Returns the string spelled out by a group that should hold nothing but plain
```
grigoriy-reshetniak: This comment feels too verbose and explain things outside of context `assertCharacterGroup`

AtsushSaito:
@grigoriy-reshetniak Thanks! All four are done in 1ed8f261.

AtsushSaito:
@edemaine Both of your points are done in this PR.

- Spaces are kept: `\begin{ma trix}` now reports `No such environment: ma trix`.
- `\begin`/`\end`, `\begin{alignedat}{n}` and `\@char` now share a new helper,
  `assertCharacterGroup()`.

Two readers are still left: the style argument of `\genfrac` and the column spec of
`{array}`. They don't have the same shape, so guarding them changes a little more.
Should I add them here, or in a follow-up PR?

grigoriy-reshetniak (approved):

src/functions/char.ts
```diff
@@ -13,12 +13,8 @@ defineFunction({
...
         const arg = assertNodeType(args[0], "ordgroup");
-        const group = arg.body;
-        let number = "";
-        for (let i = 0; i < group.length; i++) {
-            const node = assertNodeType(group[i], "textord");
-            number += node.text;
-        }
+        const number = assertCharacterGroup(
+            arg, "\\@char has non-numeric argument");
         let code = parseInt(number);
```
edemaine, line 18: We should maybe validate this like the code above, requiring `/^[0-9]+$/` before converting with `Number`. (Should we make and share a helper?) `parseInt` accepts numeric prefixes, so `\@char{4.5}` is silently treated as character 4. On the other hand, this is an internal macro, so not a big deal either way.

edemaine (approved):
Looks great, thanks! Just one optional comment. 

Followup PR for genfrac and array would be appreciated!

PR #4257, opened 2026-08-01
fix(char): reject control, surrogate and noncharacter code points

### LINKED ISSUE #2549: \char can be used to write invalid raw characters
Using `\char`, KaTeX can be convinced to output raw characters that are invalid HTML5 and/or XML:

* `\char0`: NUL character (invalid HTML5, XML)
* `\char1`–`\char8`, `\char11`, `\char14`–`\char31`: C0 control codes (invalid HTML5)
* `\char12`: form feed (invalid XML)
* `\char127`: DEL character (invalid HTML5)
* `\char128`–`\char159`: C1 control codes (invalid HTML5)
* `\char55296`–`\char57343`: unpaired surrogates (invalid HTML5, XML)
* `\char64976`–`\char65007`: noncharacters (invalid HTML5)
* `\char65534`, `\char65535`: noncharacters (invalid HTML5, XML)

The number is interpreted modulo 65536, so `\char65536` also outputs a NUL character, etc.

```
> katex = require('katex'); katex.version
'0.12.0'
> katex.renderToString('\\char0').includes('\x00')  // NUL character
true
> katex.renderToString('\\char1').includes('\x01')  // C0 control code
true
> katex.renderToString('\\char12').includes('\f')  // form feed
true
> katex.renderToString('\\char127').includes('\x7f')  // DEL character
true
> katex.renderToString('\\char128').includes('\x80')  // C1 control code
true
> katex.renderToString('\\char55296').includes('\ud800')  // unpaired surrogate
true
> katex.renderToString('\\char64976').includes('\ufdd0')  // noncharacter
true
> katex.renderToString('\\char65534').includes('\ufffe')  // noncharacter
true
> katex.renderToString('\\char65536').includes('\x00')  // NUL character again
true
```

AtsushSaito:

`\@char` checked only that its argument was inside the Unicode codespace, then wrote the resulting character straight into both the HTML and the MathML output. Being a code point is weaker than being a character that may appear in a document, so several ranges escaped into the markup:

| Input | Emitted |
|---|---|
| `\char0`, `\char12`, `\char127`, `\char128` | raw C0/C1 control bytes |
| `\char55296` | a lone surrogate |
| `\char64976`, `\char65534`, `\char1114110` | Unicode noncharacters |

All of it is reachable from user input, which is the case that matters for a site rendering untrusted LaTeX.

Four groups are now rejected with a `ParseError`, for two distinct reasons.

*They cannot be represented.* The C0 controls other than tab, line feed and carriage return — form feed included — the surrogates, and `U+FFFE`/`U+FFFF` fall outside XML's `Char` production, so the emitted MathML was not well-formed and a strict XML consumer (EPUB, XHTML, a server-side pipeline) fails outright. A lone surrogate additionally has no UTF-8 encoding at all.

*Unicode does not permit them in interchange.* DEL, the C1 controls, `U+FDD0`–`U+FDEF`, and `U+nFFFE`/`U+nFFFF` for n ≥ 1 are legal XML 1.0, so the weaker claim is the honest one for these: HTML makes each of them a parse error, and a browser will not reproduce what was asked for.

```
\char0 → KaTeX parse error: \@char code point 0 is not allowed in the output
```

Because it is a `ParseError` rather than an assertion, `throwOnError: false` renders the usual error text instead of leaking the byte.

**This rejects input that previously rendered.** It is a visible behaviour change, so flagging it explicitly. I have left it as `fix:` rather than adding a `BREAKING CHANGE:` footer, on the grounds that the affected inputs never produced usable output — a browser substitutes U+FFFD or drops the character — so no working document should depend on them. Happy to add the footer if you would rather it went out as a major.

Still accepted, with tests now on **both** sides of every boundary: `\char9`, `\char10`, `\char13`, `\char32`, `\char160`, `\char55295`, `\char57344`, `\char64975`, `\char65008`, `\char65533`, `\char65536`, `\char1114109`, and the existing `\char"1d7d9`.

**Scope.** This guards `\@char` only, and is not an output-wide invariant. A literal C1 control or noncharacter placed directly in the input string still reaches the output, because `Lexer`'s `tokenRegex` admits them; literal C0 controls and lone surrogates are already rejected while lexing. `\char` is the path that lets plain ASCII input construct an arbitrary code point, which is why it is the one the issue reports, but enforcing the same invariant where the text node is serialized would cover the remaining cases. That is a larger change — happy to open a follow-up issue, or to widen this PR, whichever you prefer.

Two smaller notes:

- The bound on the existing range check was off by one (`code >= 0x10ffff` rejected `U+10FFFF`, which is a real code point). Corrected to `> 0x10ffff`. `U+10FFFF` is a noncharacter, so it is still rejected — only now with the accurate message, which a test pins.
- The issue also reports that "the number is interpreted modulo 65536, so `\char65536` also outputs a NUL character". That was true of 0.12.0 but is not true today — `\char65536` correctly produces `U+10000`, and there is now a test for it.

Docs updated, since `supported.md` promised that any character can be written with `\char`.

Fixes #2549

### DIFF
--- a/src/functions/char.ts
+++ b/src/functions/char.ts
@@ -2,6 +2,21 @@ import defineFunction from "../defineFunction";
 import ParseError from "../ParseError";
 import {assertNodeType} from "../parseNode";
 
+// A code point inside the Unicode codespace is not necessarily one the output
+// may contain.  Of those rejected below, the C0 controls, the surrogates and
+// U+FFFE/U+FFFF are outside XML's Char production, so emitting one would make
+// the MathML ill-formed; the rest are legal XML but are HTML parse errors.
+const isForbiddenInOutput = (code: number): boolean =>
+    // C0 controls; form feed is not exempted because it is not valid XML
+    (code < 0x20 && code !== 0x09 && code !== 0x0a && code !== 0x0d) ||
+    // DEL and the C1 controls
+    (code >= 0x7f && code <= 0x9f) ||
+    // surrogates
+    (code >= 0xd800 && code <= 0xdfff) ||
+    // noncharacters; the second test covers U+nFFFE and U+nFFFF in any plane
+    (code >= 0xfdd0 && code <= 0xfdef) ||
+    (code & 0xfffe) === 0xfffe;
+
 // \@char is an internal function that takes a grouped decimal argument like
 // {123} and converts into symbol with code 123.  It is used by the *macro*
 // \char defined in macros.js.
@@ -25,8 +40,11 @@ defineFunction({
             throw new ParseError(`\\@char has non-numeric argument ${number}`);
         // If we drop IE support, the following code could be replaced with
         // text = String.fromCodePoint(code)
-        } else if (code < 0 || code >= 0x10ffff) {
+        } else if (code < 0 || code > 0x10ffff) {
             throw new ParseError(`\\@char with invalid code point ${number}`);
+        } else if (isForbiddenInOutput(code)) {
+            throw new ParseError(
+                `\\@char code point ${number} is not allowed in the output`);
         } else if (code <= 0xffff) {
             text = String.fromCharCode(code);
         } else { // Astral code point; split into surrogate halves
--- a/docs/supported.md
+++ b/docs/supported.md
@@ -230,7 +230,7 @@ All unrecognized characters will be treated as if they appeared in text mode, an
 
 For Persian composite characters, a user-supplied [plug-in](https://github.com/HosseinAgha/persian-katex-plugin) is under development.
 
-Any character can be written with the `\char` function and the Unicode code in hex. For example `\char"263a` will render as $\char"263a$.
+Any character can be written with the `\char` function and the Unicode code in hex, except for surrogates, noncharacters, and control characters other than tab, line feed and carriage return. For example `\char"263a` will render as $\char"263a$.
 
 ## Layout
 
--- a/test/errors-spec.ts
+++ b/test/errors-spec.ts
@@ -282,6 +282,51 @@ describe("functions.js:", function() {
         });
     });
 
+    describe("\\char", function() {
+        it("rejects code points outside Unicode", function() {
+            expect`\char1114112`.toFailWithParseError(
+                   "\\@char with invalid code point 1114112");
+        });
+        it("rejects control characters", function() {
+            expect`\char0`.toFailWithParseError(
+                   "\\@char code point 0 is not allowed in the output");
+            expect`\char12`.toFailWithParseError(
+                   "\\@char code point 12 is not allowed in the output");
+            expect`\char31`.toFailWithParseError(
+                   "\\@char code point 31 is not allowed in the output");
+            expect`\char127`.toFailWithParseError(
+                   "\\@char code point 127 is not allowed in the output");
+            expect`\char128`.toFailWithParseError(
+                   "\\@char code point 128 is not allowed in the output");
+            expect`\char159`.toFailWithParseError(
+                   "\\@char code point 159 is not allowed in the output");
+        });
+        it("rejects surrogate code points", function() {
+            expect`\char55296`.toFailWithParseError(
+                   "\\@char code point 55296 is not allowed in the output");
+            expect`\char57343`.toFailWithParseError(
+                   "\\@char code point 57343 is not allowed in the output");
+        });
+        it("rejects noncharacters", function() {
+            expect`\char64976`.toFailWithParseError(
+                   "\\@char code point 64976 is not allowed in the output");
+            expect`\char65007`.toFailWithParseError(
+                   "\\@char code point 65007 is not allowed in the output");
+            expect`\char65534`.toFailWithParseError(
+                   "\\@char code point 65534 is not allowed in the output");
+            expect`\char65535`.toFailWithParseError(
+                   "\\@char code point 65535 is not allowed in the output");
+            expect`\char131070`.toFailWithParseError(
+                   "\\@char code point 131070 is not allowed in the output");
+            expect`\char1114110`.toFailWithParseError(
+                   "\\@char code point 1114110 is not allowed in the output");
+            // U+10FFFF is a real code point, so it must be rejected here and
+            // not by the range check above.
+            expect`\char1114111`.toFailWithParseError(
+                   "\\@char code point 1114111 is not allowed in the output");
+        });
+    });
+
 });
 
 describe("Lexer:", function() {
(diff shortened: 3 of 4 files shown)


AtsushSaito:
@grigoriy-reshetniak would you mind taking a look when you have a moment?

src/functions/char.ts
```diff
@@ -2,6 +2,21 @@ import defineFunction from "../defineFunction";
 import ParseError from "../ParseError";
 import {assertNodeType} from "../parseNode";
 
+// A code point inside the Unicode codespace is not necessarily one the output
+// may contain.  Of those rejected below, the C0 controls, the surrogates and
+// U+FFFE/U+FFFF are outside XML's Char production, so emitting one would make
+// the MathML ill-formed; the rest are legal XML but are HTML parse errors.
+const isForbiddenInOutput = (code: number): boolean =>
```
grigoriy-reshetniak, line 9: I'm not sure about this. It looks like it could be spitted into  `isForbiddenInHTML` and `isForbiddenInMathML` to throw errors with more precise messages and potentially not throw based on `output` in `Settings`, but I'm not sure about practical use cases.

grigoriy-reshetniak: @edemaine could you please advise on this?

edemaine (commented):
> They cannot be represented. The C0 controls other than tab, line feed and carriage return …, the surrogates, and `U+FFFE`/`U+FFFF` fall outside XML's `Char` production.

This is true specifically of [XML 1.0’s `Char` production](https://www.w3.org/TR/xml/#charsets). [XML 1.1 permits most controls as characters](https://www.w3.org/TR/xml11/#charsets) but requires most of them to be written as character references rather than literally. XML 1.1 still forbids NUL, surrogates, U+FFFE, and U+FFFF entirely, but permits the other noncharacters, as does XML 1.0.

It is less clear to me that guaranteeing XML-serializable output is necessarily KaTeX’s responsibility. The common use of KaTeX output is as HTML or DOM content. Directly treating the generated markup as strict XML—for example in an XHTML, EPUB, or server-side XML pipeline—seems relatively unusual. Such use cases may exist, but I am not sure they justify rejecting input for every KaTeX user (at least not without an option).

> Unicode does not permit them in interchange.

This is too strong. [The Unicode Standard says](https://www.unicode.org/versions/Unicode17.0.0/core-spec/chapter-23/#G12985) that noncharacters are not illegal in interchange and do not make Unicode text ill-formed. [The Unicode FAQ recommends](https://unicode.org/faq/private_use.html#nonchar8) that general-purpose APIs preserve them rather than arbitrarily filter them.

> DEL, the C1 controls, `U+FDD0`–`U+FDEF`, and `U+nFFFE`/`U+nFFFF` for n ≥ 1 are legal XML 1.0 … HTML makes each of them a parse error, and a browser will not reproduce what was asked for.

The latter part is generally not true. The HTML Standard says that [control characters are parsed as-is and usually reach the DOM](https://html.spec.whatwg.org/multipage/parsing.html#parse-error-control-character-in-input-stream), and [says the same about noncharacters](https://html.spec.whatwg.org/multipage/parsing.html#parse-error-noncharacter-in-input-stream).

I also tested this in Chrome via `innerHTML`. U+007F, U+0080–U+009F, U+FDD0, U+FFFE/U+FFFF, and supplementary-plane noncharacters were preserved in the resulting text node. NUL is dropped or replaced, while CR is normalized to LF. The acceptance of `\char13` therefore also seems inconsistent with browser round-tripping as the criterion.

> No working document should depend on them.

I do not think we can assume this. These values generally have no useful visible glyph, but JavaScript and DOM APIs can preserve many of them, and [Unicode explicitly permits internal uses of noncharacters](https://www.unicode.org/versions/Unicode17.0.0/core-spec/chapter-23/#G12985).

More fundamentally, I am not convinced KaTeX should decide which otherwise valid JavaScript strings downstream applications are allowed to receive. KaTeX needs to prevent markup injection, but these characters do not bypass its escaping of `<`, `>`, `&`, or attributes. Problems encountered when subsequently encoding the result as XML may belong to that serialization layer.

> This guards `\@char` only, and is not an output-wide invariant.

This makes me less convinced that the change belongs here. Literal C1 controls and noncharacters can still reach the output through the lexer, including literal U+FFFE and U+FFFF.

Consequently, the PR does not establish that KaTeX output is valid XML or free of HTML parse errors. It only gives `\char` different Unicode filtering rules from literal input.

> It could be split into `isForbiddenInHTML` and `isForbiddenInMathML`.

That would describe the standards more accurately. We could offer some kind of option that forces the output to be consistent with a particular serialization spec, maybe even specifying the version of XML. But I'd like to see a real situation where this is useful before we add the complexity. And if implemented, I think it should be a filter on the final output, not local processing during `\char`.

> The bound on the existing range check was off by one.

Agreed. Changing `code >= 0x10ffff` to `code > 0x10ffff` is clearly correct: U+10FFFF is inside the Unicode codespace. I think we should definitely extract this part of the PR and merge it.

At this point, I lean toward taking only that corrected bound. I am not yet convinced the additional Unicode filtering is KaTeX’s job, especially when it does not create an output-wide invariant and primarily benefits direct XML consumers, which feels a bit special.

AtsushSaito:
@edemaine @grigoriy-reshetniak 

Thank you for the detailed review — let me consider each point in turn.
The XML 1.0 / 1.1 distinction  were new to me, and reading them changed my view of the patch rather than just its wording.

As a first step, I have extracted the off-by-one fix into a separate PR: [#4265](https://github.com/KaTeX/KaTeX/pull/4265). You said it should be extracted and merged on its own, so I kept it independent of this branch, and it takes no position on the noncharacter question.

Could you take a look when you have a moment?

AtsushSaito:
@edemaine @grigoriy-reshetniak 
Thank you for the detailed review — the corrections all hold, and I have withdrawn the categorical filtering.

Two things I take from it. Fitting the output to a particular serialization belongs to that serialization rather than to \char; and if it is ever done inside KaTeX, the boundary is the finished string, not local processing during parsing. I agree with both.

That leaves one question I could not settle by measuring, and I think it is yours rather than mine:

Should one \char produce one valid character?

Today it need not. \char55296 on its own emits an unpaired surrogate. But \text{\char55296\char56320} emits U+10000 — the same character \char65536 produces — because adjacent textords share a text node in text mode. In math mode they do not, so the same input is ill-formed there. The invariant is held in one mode and not the other, and the direct spelling works in both.

I am closing #4257.

AtsushSaito:
Closing this, as stated above. The corrected upper bound went in as #4265; the lone-surrogate case is the only part I still think stands on its own, and I have left it out rather than reopen the wider question here. Thanks for the detailed review — the Corrigendum #9 point in particular.

PR #4260, opened 2026-08-05
fix: prevent prototype pollution in settings

grigoriy-reshetniak:
Prototype pollution could affect values in `Settings`

`hasOwnProperty` checks added in `Settings` to guard against prototype pollution


edemaine (approved):
Looks great! One small comment.

GPT found one additional lookup in `Namespace` that would enable adding custom macros via `Object.prototype` pollution. I pushed a fix for this which you should review.

grigoriy-reshetniak:
Looks great, thank you!

PR #4262, opened 2026-08-07
feat: add \reflectbox and \mapsfrom support

### LINKED ISSUE #3730: Adding the symbol `\mapsfrom`
I think it is reasonable to add support of `\mapsfrom`, which is the left version of `\mapsto`. This is implemented in the LaTeX package [stmaryrd](https://ctan.org/pkg/stmaryrd).

iridescent99:
- `\reflectbox` is supported
- `\mapsfrom` symbol is supported, implemented using `\reflectbox`
- support for `↤`

Closes #3730


edemaine (changes requested):
Nice idea to use reflection! 

I would rather define a general `\reflectbox` feature that does this CSS transform, and then define `\mapsfrom` via the macro ``\mathbin{\html@mathml{\reflectbox{$\mapsto$}}{\char`↤}}``.

Also need to update docs.

iridescent99:
> Nice idea to use reflection!
> 
> I would rather define a general `\reflectbox` feature that does this CSS transform, and then define `\mapsfrom` via the macro `` \mathbin{\html@mathml{\reflectbox{$\mapsto$}}{\char`↤}} ``.
> 
> Also need to update docs.

Thank you for the feedback. I added support for reflectbox, hope it's close to what you had in mind

docs/support_table.md
```diff
@@ -901,6 +902,7 @@ use `\ce` instead|
 |\Reals|$\Reals$||
 |\reals|$\reals$||
 |\ref|<span style="color:firebrick;">Not supported</span>|[Issue #350](https://github.com/KaTeX/KaTeX/issues/350)|
+|\reflectbox|$\reflectbox{$x^2$}$||
```
grigoriy-reshetniak: Could you please add a comment in the third column, it is not reflected on the [support table](https://deploy-preview-4262--katex.netlify.app/docs/support_table) (although it's reflected in the demo).

grigoriy-reshetniak:
Looks great!

Could you please add `screenshotter` test cases?

grigoriy-reshetniak (approved):

grigoriy-reshetniak:
Thank you, looks great to me!

edemaine (approved):
I reviewed, found some issues, and pushed proposed fixes. Can either of you review?

Specifically, I fixed spacing and sizing, and added `\mathreflectbox`:

- Both reflection commands now produce ordinary atoms (`mord`), restoring spacing around adjacent operators.
- `\mapsfrom` now uses `\mathrel` (not `\mathbin`), matching LaTeX’s relation spacing.
- `\mathreflectbox` shares the reflection node and builders but parses math directly, preserving the surrounding style. `\mapsfrom` uses it to scale correctly in scripts. `\reflectbox` retains LaTeX’s text-box behavior.

Updated documentation, regression tests, and screenshots (Safari once they build).

src/macros.ts
```diff
@@ -811,6 +811,12 @@ defineMacro("\\rBrace", "\\html@mathml{" +
     "\\mathclose{]\\mkern-3.2mu\\}}}" +
     "{\\mathclose{\\char`\u2984}}");
 
+defineMacro("↤", "\\mapsfrom");
+defineMacro(
+    "\\mapsfrom",
+    "\\mathrel{\\html@mathml{\\mathreflectbox{\\mapsto}}{\\char`↤}}",
```
grigoriy-reshetniak: This will produce `<mo><mi mathvariant="normal">↤</mi></mo>` for MathML. I see in other marco `\html@mathml` is wrapped if `\mathrel` to prevent this. Should it be done here as well?

```suggestion
    "\\html@mathml{\\mathrel{\\mathreflectbox{\\mapsto}}}{\\mathrel{\\char`↤}}",
```

edemaine: Good catch, thanks! Fixed.

PR #4265, opened 2026-08-12
fix(char): accept U+10FFFF

AtsushSaito:

`\@char`'s range check was `code < 0 || code >= 0x10ffff`. The upper bound is off by one: U+10FFFF is inside the Unicode codespace, so `\char"10ffff` failed to parse while `\char"10fffe` parsed fine. U+10FFFF was the only code point the bound wrongly excluded.

The bound becomes `> 0x10ffff`, and `\char"10ffff` parses.

U+10FFFF is a noncharacter, so it has no visible glyph and is of little use in a document. Whether `\char` should reject noncharacters as a class is a separate question, discussed in #2549 and in #4257; this PR takes no position on it — the bound is wrong either way. @edemaine agreed in [this review](https://github.com/KaTeX/KaTeX/pull/4257#pullrequestreview-4898604976) that this part should be extracted and merged on its own, so this is that extraction.

Tests cover both sides of the boundary — `\char"10ffff` parses and yields U+10FFFF, `\char"110000` does not parse. Neither side was covered before.

### DIFF
--- a/test/katex-spec.ts
+++ b/test/katex-spec.ts
@@ -3898,7 +3898,14 @@ describe("A macro expander", function() {
         expect(parsed[0].text).toEqual("𝟙");
     });
 
-    it("\\char accepts the last code point of the Unicode codespace", () => {
+    it("\\char accepts the first and last code point of every plane", () => {
+        for (let plane = 0; plane <= 0x10; plane++) {
+            for (const code of [plane * 0x10000, plane * 0x10000 + 0xffff]) {
+                const parsed = getParsed(`\\char"${code.toString(16)}`);
+                expect(parsed[0].type).toEqual("textord");
+            }
+        }
+    });
         const parsed = getParsed`\char"10ffff`;
         expect(parsed[0].type).toEqual("textord");
         // Written as a surrogate pair, since U+10FFFF has no visible glyph
--- a/test/katex-spec.ts
+++ b/test/katex-spec.ts
@@ -3906,11 +3906,6 @@ describe("A macro expander", function() {
             }
         }
     });
-        const parsed = getParsed`\char"10ffff`;
-        expect(parsed[0].type).toEqual("textord");
-        // Written as a surrogate pair, since U+10FFFF has no visible glyph
-        expect(parsed[0].text).toEqual("\udbff\udfff");
-    });
 
     it("\\char rejects code points outside the Unicode codespace", () => {
         expect`\char"110000`.toFailWithParseError();
--- a/test/katex-spec.ts
+++ b/test/katex-spec.ts
@@ -3898,6 +3898,17 @@ describe("A macro expander", function() {
         expect(parsed[0].text).toEqual("𝟙");
     });
 
+    it("\\char accepts the last code point of the Unicode codespace", () => {
+        const parsed = getParsed`\char"10ffff`;
+        expect(parsed[0].type).toEqual("textord");
+        // Written as a surrogate pair, since U+10FFFF has no visible glyph
+        expect(parsed[0].text).toEqual("\udbff\udfff");
+    });
+
+    it("\\char rejects code points outside the Unicode codespace", () => {
+        expect`\char"110000`.toFailWithParseError();
+    });
+
     it("should build Unicode private area characters", function() {
         expect`\gvertneqq\lvertneqq\ngeqq\ngeqslant\nleqq`.toBuild();
         expect`\nleqslant\nshortmid\nshortparallel\varsubsetneq`.toBuild();
--- a/src/functions/char.ts
+++ b/src/functions/char.ts
@@ -21,7 +21,7 @@ defineFunction({
             throw new ParseError(`\\@char has non-numeric argument ${number}`);
         // If we drop IE support, the following code could be replaced with
         // text = String.fromCodePoint(code)
-        } else if (code < 0 || code >= 0x10ffff) {
+        } else if (code < 0 || code > 0x10ffff) {
             throw new ParseError(`\\@char with invalid code point ${number}`);
         } else if (code <= 0xffff) {
             text = String.fromCharCode(code);


test/katex-spec.ts
```diff
@@ -3898,6 +3898,17 @@ describe("A macro expander", function() {
         expect(parsed[0].text).toEqual("𝟙");
     });
 
+    it("\\char accepts the last code point of the Unicode codespace", () => {
```
grigoriy-reshetniak: It feels like this test case is too specific and `\char rejects code points outside the Unicode codespace` is sufficient.

grigoriy-reshetniak: ~~I've also found https://github.com/KaTeX/KaTeX/blob/main/test/unicode-spec.ts and I feel like that test belongs there.~~

Edit: Other `\char` tests located in `katex-spec`, so I recall comment above.

AtsushSaito: @grigoriy-reshetniak 
Thanks for the review — one thing I'd like to confirm, sorry if I'm missing your point.

When you say \char rejects code points outside the Unicode codespace is sufficient, do you mean the accept-side check (that U+10FFFF itself parses) isn't needed at all? Or do you mean we should merge the two it() blocks into one, keeping both the accept and reject assertions together in that single block?

The reason I ask is that this fix corrects an off-by-one bug (>= → >), so I'd think we want a test that would have failed before the fix and passes after it — otherwise a future change could reintroduce the same off-by-one without anything catching it. Wouldn't it be worth keeping that accept-side assertion for that reason?

I will think of your second point: the location of the tests later

I always aprreciate your educational  comments!

grigoriy-reshetniak: > accept-side check (that U+10FFFF itself parses) isn't needed at all

Yes, that was the point. I understand that this is exact case which was uncovered and let to a bug, but this test case is too specific. It verifies, that that char accepts that specific character, which just happens to be last character on the last plane. It cannot be scaled for other characters, like `\char produces literal characters` or `\char handles >16-bit characters`.

I was thinking more about [this test](https://github.com/AtsushSaito/KaTeX/blob/main/test/unicode-spec.ts#L167) as a reference where it verifies every code point, but it may be too much for `\char`.

AtsushSaito: @grigoriy-reshetniak 
Fair point — the name pins one character, so it says nothing general.

How about your sweep idea, just cut down so it isn't too much? First and last code point of every plane, 34 parses:

```suggestion
    it("\\char accepts the first and last code point of every plane", () => {
        for (let plane = 0; plane <= 0x10; plane++) {
            for (const code of [plane * 0x10000, plane * 0x10000 + 0xffff]) {
                const parsed = getParsed(`\\char"${code.toString(16)}`);
                expect(parsed[0].type).toEqual("textord");
            }
        }
    });
```

U+10FFFF is no longer hand-picked, it's just the last plane's last one. The off-by-one stays covered as a side effect.

Happy to drop it and keep only the reject test if you'd rather — your call.

grigoriy-reshetniak: That works, thank you.

Looking at https://github.com/KaTeX/KaTeX/issues/2549 I don't think that `\char` _should_ accepts all those symbols, but at least this test will fixate current behavior.

grigoriy-reshetniak (approved):

grigoriy-reshetniak:
@AtsushSaito thank you!

@edemaine I've merged it because [that comment](https://github.com/KaTeX/KaTeX/pull/4257#pullrequestreview-4898604976)
> Agreed. Changing code >= 0x10ffff to code > 0x10ffff is clearly correct: U+10FFFF is inside the Unicode codespace. I think we should definitely extract this part of the PR and merge it.

felt like an approval.

test/katex-spec.ts
```diff
@@ -3898,6 +3898,19 @@ describe("A macro expander", function() {
         expect(parsed[0].text).toEqual("𝟙");
     });
 
+    it("\\char accepts the first and last code point of every plane", () => {
+        for (let plane = 0; plane <= 0x10; plane++) {
+            for (const code of [plane * 0x10000, plane * 0x10000 + 0xffff]) {
```
edemaine, line 3903: I'm curious: How long does this test take?

grigoriy-reshetniak: Short answer: about **8ms.** 
<details>
<summary>Generated script I've used for measurement</summary>

```
#!/usr/bin/env node
/*
 * Report how long individual Jest tests actually take, separated from the
 * fixed per-suite overhead (module imports, ts transform, describe bodies).
 *
 * Usage:
 *   node jest-times.js <test-file> [-t <name pattern>]
 *
 * Examples:
 *   node jest-times.js test/katex-spec.ts -t 'char accepts the first'
 *   node jest-times.js test/unicode-spec.ts
 *
 * Exactly one test file is required; no other Jest arguments are accepted.
 */

const {spawnSync} = require("child_process");
const fs = require("fs");
const os = require("os");
const path = require("path");

const usage = "usage: node jest-times.js <test-file> [-t <name pattern>]";

const argv = process.argv.slice(2);
let suite = null;
let pattern = null;

for (let i = 0; i < argv.length; i++) {
    const arg = argv[i];
    if (arg === "-t") {
        if (pattern !== null) { die("-t given more than once"); }
        pattern = argv[++i];
        if (pattern === undefined) { die("-t requires a name pattern"); }
    } else if (arg.startsWith("-")) {
        die(`unsupported argument: ${arg}`);
    } else if (suite === null) {
        suite = arg;
    } else {
        die(`only one test file is accepted (got "${suite}" and "${arg}")`);
    }
}

if (suite === null) { die("no test file given"); }
if (!fs.existsSync(suite)) { die(`no such file: ${suite}`); }

function die(message) {
    console.error(`jest-times: ${message}\n${usage}`);
    process.exit(2);
}

const outFile = path.join(
    fs.mkdtempSync(path.join(os.tmpdir(), "jest-times-")), "results.json");

// --runTestsByPath keeps Jest from resolving and transforming the other
// suites, which otherwise costs ~750 ms each even when their tests are
// filtered out.
const jestArgs = ["jest", "--runTestsByPath", suite,
    "--json", "--outputFile", outFile];
if (pattern !== null) { jestArgs.push("-t", pattern); }

const started = Date.now();
spawnSync("npx", jestArgs, {stdio: ["inherit", "ignore", "ignore"]});
const wall = Date.now() - started;

```

if (!fs.existsSync(outFile)) {
    console.error(`jest-times: Jest produced no output for ${suite}`);
    process.exit(1);
}

const report = JSON.parse(fs.readFileSync(outFile, "utf8"));
fs.rmSync(path.dirname(outFile), {recursive: true, force: true});

const result = report.testResults[0];

if (!result) { die(`no suite matched ${suite}`); }

const ms = (n) => `${String(Math.round(n)).padStart(7)} ms`;
const suiteWall = result.endTime - result.startTime;
const tests = result.assertionResults.filter((a) => a.status !== "pending");
const testTime = tests.reduce((sum, a) => sum + (a.duration || 0), 0);

console.log(`\n${suite}` + (pattern === null ? "" : `  -t '${pattern}'`));
for (const t of tests.sort((a, b) => (b.duration || 0) - (a.duration || 0))) {
    console.log(`${ms(t.duration || 0)}  ${t.status === "failed" ? "✕" : "✓"}` +
        ` ${t.fullName}`);
}
if (!tests.length) {
    console.log("  (no tests matched)");
}

console.log("\n" + "-".repeat(60));
console.log(`tests run        ${String(tests.length).padStart(7)}`);
console.log(`test time        ${ms(testTime)}   sum of the durations above`);
console.log(`suite overhead   ${ms(suiteWall - testTime)}   imports, ts ` +
    `transform, describe bodies`);
console.log(`suite wall       ${ms(suiteWall)}`);
console.log(`wall clock       ${ms(wall)}   plus jest startup and teardown`);

// Ignore report.success: with -t it is false purely because the filtered-out
// tests leave their snapshots unchecked.
process.exit(report.numFailedTests || report.numFailedTestSuites ? 1 : 0);

</details>

But when testing individual test cases spinning `jest` takes more time than the test case itself. Commenting it out doesn't impact overall time for `pnpm test:jest`. Also, the test only does 34 `getParsed` calls so the measurement looks correct.

PR #4267, opened 2026-08-17
fix(symbols): map \perp to U+27C2 PERPENDICULAR

### LINKED ISSUE #4011: \perp is mapped to the wrong Unicode code point
**Describe the bug:**
KaTeX maps the TeX macro \perp to the Unicode char U+22A5 which Unicode defines as "UP TACK {base, bottom}". It should be mapped to U+27C2 which Unicode defines as "PERPENDICULAR {orthogonal to}". Although they look similar, they are slightly different.

Beside the potential subtle difference in rendering, the bad mapping means that screen readers will say the wrong thing for \perp unless they (wrongly) map U+22A5 to "perpendicular". If they do that, then they won't speak "up tack" or "bottom" when they should.

The bug is in `src\symbols.js`:
```
defineSymbol(math, main, rel, "\u22a5", "\\perp");
```
Note that `src\macros.js` maps the proper Unicode to `\perp`:
```
defineMacro("\u27C2", "\\perp");
```

**To Reproduce:**
1. Go to https://katex.org/
2. Type `x \perp y` into the input area
3. One way to see what character is used is to right click on the `⊥` and choose "inspect". From there, you can copy the character used to see that it is U+22A5

**Expected behavior:**
 U+27C2  should be used for `\perp`, both for rendering and for MathML export.

 - KaTeX Version: katex.org seems to be using 0.16.19 when I tried it
 - Device:  Desktop  (shouldn't matter)
 - OS: Windows 10 (shouldn't matter)
 - Browser: Firefox (shouldn't matter)

tzh476:

`\perp` was defined as U+22A5 (UP TACK / bottom), the same code point as `\bot`. MathML export and copy/paste therefore used the "bottom" character, so assistive tech could not distinguish `\perp` from `\bot`.

`\perp` maps to U+27C2 (PERPENDICULAR). `\bot` stays on U+22A5. KaTeX fonts do not carry U+27C2, so HTML rendering substitutes the U+22A5 glyph — which is the same glyph — while MathML keeps the semantically correct U+27C2.

## How

- `src/symbols.ts`: `\perp` → U+27C2; `\bot` unchanged.
- `src/fontMetrics.ts`: new `glyphSubstitutionMap` + `getFontGlyphCharacter()`, holding **only genuine glyph equivalences**. This is deliberately *not* `extraCharacterMap`.
- `src/buildCommon.ts`: `lookupSymbol` consults the substitution **in math mode only**, before the metrics lookup. Text mode is left alone so the browser can still fall back to system fonts.

## On the earlier review (addressed)

@grigoriy-reshetniak was right that `extraCharacterMap` is a *metrics* table: "If there's no metric for `Ð`, **get metric** for `D`, not ... **render** `D`". My first attempt routed glyph selection through it, which silently turned every entry into a *rendering* substitution — `Ð`→`D`, and worse, `Ш`→`W`, `Д`→`A`, `Ж`→`K` in math mode. The demo link made that concrete.

Commit `7de5eb94` removes the `extraCharacterMap` entry entirely and puts the single true equivalence (U+27C2 ≡ U+22A5) in a separate map. `test/unicode-spec.ts` now asserts the metrics-only lookalikes are **never** substituted, so that regression cannot come back unnoticed.

## Verification

Checked against the real `src/fontMetricsData.js`: of KaTeX's 18 metric fonts, U+27C2 is absent from **all** of them, while U+22A5 is present in Main-Regular and Main-Bold — so the substitution premise holds rather than being assumed. After the patch `getCharacterMetrics(U+27C2)` returns `undefined`, but `lookupSymbol` substitutes to U+22A5 *first*, so metrics resolve normally and there is no metrics regression.

`Ш Д Ж Г Å Ð Þ å` were confirmed unchanged across all fonts.

## Testing

*(Updated 2026-09-02 — the numbers and the test inventory below drifted after the review changes in
`512ce0c0`; corrected against head `75a76a1d`.)*

- `pnpm jest` — 8 suites, **1308** tests, 125 snapshots passed
- `tsc --noEmit` — 0 errors
- `pnpm test:lint` (eslint + stylelint) — exit 0
- New in `test/unicode-spec.ts`:
  - `\perp` renders with the U+22A5 glyph in HTML while MathML keeps U+27C2 — moved here from
    `mathml-spec.ts` at the reviewer's request, since it asserts HTML output
  - the lookalike characters in `extraCharacterMap` are returned **unchanged** (the negative case;
    injecting a lookalike into `glyphSubstitutionMap` fails only this test)
- New in `test/mathml-spec.ts`: each command carries the character Unicode assigns it — `\perp` → U+27C2,
  `\bot` → U+22A5, `\top` → U+22A4, each absent from the others' output
- **Removed** since the first version of this description: the direct
  `getFontGlyphCharacter("⟂") === "⊥"` assertion in `unicode-spec.ts`, which the moved HTML test covers
  end-to-end through `lookupSymbol`

Fixes #4011


grigoriy-reshetniak:
@greptile revise your review

src/fontMetrics.ts
```diff
@@ -105,6 +105,10 @@ const sigmasAndXis: Record<string, [number, number, number]> = {
 // the fraction bar or root line from intersecting the glyph.
 // TODO(kevinb) allow union of multiple glyph metrics for better accuracy.
 const extraCharacterMap: Record<string, string> = {
```
grigoriy-reshetniak, line 107: The name of this map is a little misleading, but it serves as a mapping for characters _metrics_. The logic here is:

> If there's no metric for `Ð`, **get metric** for `D`.

not the

> If there's no metric for `Ð`, **render** `D`.

But in a way how `getFontGlyphCharacter` works it fallback when it shouldn't. Please take a look at [demo](https://katex.org/?data=%7B%22displayMode%22%3Atrue%2C%22leqno%22%3Afalse%2C%22fleqn%22%3Afalse%2C%22throwOnError%22%3Atrue%2C%22errorColor%22%3A%22%23cc0000%22%2C%22strict%22%3A%22warn%22%2C%22output%22%3A%22htmlAndMathml%22%2C%22trust%22%3Afalse%2C%22macros%22%3A%7B%22%5C%5Cf%22%3A%22%231f(%232)%22%7D%2C%22code%22%3A%22%5C%5Cboxed%7B%C3%90%5C%5Cquad%20%C3%9E%5C%5Cquad%20%C3%BE%7D%5C%5Cqquad%5C%5Cboxed%7B%5C%5Ctext%7B%C3%90%20%C3%9E%20%C3%BE%7D%7D%22%7D) and [deploy preview](https://deploy-preview-4267--katex.netlify.app/?data=%7B%22displayMode%22%3Atrue%2C%22leqno%22%3Afalse%2C%22fleqn%22%3Afalse%2C%22throwOnError%22%3Atrue%2C%22errorColor%22%3A%22%23cc0000%22%2C%22strict%22%3A%22warn%22%2C%22output%22%3A%22htmlAndMathml%22%2C%22trust%22%3Afalse%2C%22macros%22%3A%7B%22%5C%5Cf%22%3A%22%231f(%232)%22%7D%2C%22code%22%3A%22%5C%5Cboxed%7B%C3%90%5C%5Cquad%20%C3%9E%5C%5Cquad%20%C3%BE%7D%5C%5Cqquad%5C%5Cboxed%7B%5C%5Ctext%7B%C3%90%20%C3%9E%20%C3%BE%7D%7D%22%7D).

grigoriy-reshetniak (changes requested):

src/fontMetrics.ts
```diff
@@ -105,6 +105,10 @@ const sigmasAndXis: Record<string, [number, number, number]> = {
...
+    // Math symbols present in Unicode but missing from KaTeX fonts.
+    // U+27C2 PERPENDICULAR shares a glyph with U+22A5 UP TACK.
+    '⟂': '⊥',
```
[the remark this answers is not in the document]
junleealfonso45: U+22A5

tzh476:
You're right, and thank you for the demo link — it made the problem concrete. I reproduced it and the failure was worse than a misleading name.

`extraCharacterMap` is a *metrics* table, exactly as you describe: "if there's no metric for `Ð`, get the metric for `D`". By routing glyph selection through it I turned every entry into a rendering substitution, so in math mode the visible text became a different character. Running your input `\boxed{Ð\quad Þ\quad þ}` against the two versions:

| | `Ð` in `katex-html` | a literal `D` in the output |
|---|---|---|
| previous commit | present | **present** |
| this commit | present | absent |

And it was not limited to Latin-1 — `Ш → W`, `Д → A`, `Ж → K` are in that same map, so Cyrillic in math mode was being rewritten to Latin letters.

The fix separates the two concerns. A new `glyphSubstitutionMap` holds only genuine glyph equivalences and is the only map consulted when choosing what to render; `extraCharacterMap` keeps its original metrics-only role, untouched. U+27C2 moves to the new map because it is the same glyph as U+22A5, which the fonts do carry.

`\perp` still behaves as intended: MathML emits U+27C2, the HTML branch emits U+22A5, and no U+27C2 reaches the HTML. Rendering `\perp \bot x \perp y` under `strict: "warn"` emits no warning, so the metrics still resolve.

Verification on this commit:

- the full jest suite passed when I ran it while preparing this commit (8 suites,
  1307 tests); I could not re-run it just now because the toolchain does not
  reinstall on my current Node. CI cannot settle it either right now: the runs on
  `7de5eb9` sit at `action_required`, so they need a maintainer to approve the
  workflow before they execute. If you approve them, treat CI as the authority over
  my local figure
- `tsc --noEmit` and `eslint` on the changed files — clean
- new tests in `test/unicode-spec.ts` assert both directions: U+27C2 substitutes, and the metrics-borrowing characters (`Ш Д Ж Г`, `Å Ð Þ å`) do not. Reverting `fontMetrics.ts` to the previous implementation fails the second assertion, so the test pins the regression you found rather than restating the fix.

tzh476:
Small follow-up, and one thing I can now state with numbers rather than defer.

I got the toolchain working (Node 22 + `corepack pnpm@11.4.0`, matching `packageManager` in `package.json`), so the suite runs locally on this branch:

```
pnpm test:jest   Test Suites: 8 passed, 8 total
                 Tests:      1307 passed, 1307 total
                 Snapshots:  125 passed, 125 total
pnpm test:ts     tsc --noEmit — 0 errors
pnpm test:lint   0 problems
```

Running it turned up one thing that was mine: the `eslint-disable-next-line @typescript-eslint/no-var-requires` I had added above the `require` in the HTML-glyph test was unnecessary — the rule does not fire there, and `eslint .` reported it as an unused directive. Removed in `ce24a93`, which takes the lint run from 1 warning to 0 problems. No source or behaviour change.

Earlier I said I could not reproduce the test results and deliberately left them out. That was a toolchain problem on my side (I had tried an old `pnpm` under Node 20), not a property of the branch — correcting it here so the record is accurate.

The substantive question from the review stands as answered in [my earlier comment](https://github.com/KaTeX/KaTeX/pull/4267#issuecomment-5461749369): `extraCharacterMap` keeps its metrics-only role and is no longer consulted when choosing what to render, so `Ð`/`Þ`/`þ` and the Cyrillic entries no longer get rewritten to Latin lookalikes in math mode. Happy to adjust anything further you would like changed.

tzh476:
@grigoriy-reshetniak your distinction was right, and the code changed after your review — flagging
it rather than leaving the PR sitting on a stale `CHANGES_REQUESTED`.

You were reviewing the version that reused `extraCharacterMap`. That map borrows *metrics* from a
similarly-shaped character, and as you say, several entries are lookalikes rather than
equivalents — Cyrillic `Ш` borrowing Latin `W` metrics is exactly the case that makes
substitution wrong. `7de5eb9` ("render U+27C2 with the U+22A5 glyph without substituting
lookalikes", 2026-08-29) stopped using it.

Substitution now goes through a separate map that is deliberately not the metrics map, and it has
exactly one entry:

```ts
const glyphSubstitutionMap: Record<string, string> = {
    // U+27C2 PERPENDICULAR is the same glyph as U+22A5 UP TACK, which the
    // KaTeX fonts do carry.
    '⟂': '⊥',
};
```

So for your demo characters, `getFontGlyphCharacter` hits `substitute === undefined` and returns
them unchanged:

| char | in `glyphSubstitutionMap`? |
|---|---|
| `Ð` U+00D0 | no |
| `Þ` U+00DE | no |
| `þ` U+00FE | no |
| `Ш` U+0428 | no |

`Ð Þ þ` are therefore untouched by this PR — they keep whatever behaviour `main` has. The
`extraCharacterMap` metrics path is unchanged; only U+27C2 is substituted, and only when the font
genuinely lacks it (`metricMap[font]?.[ch] ? character : substitute`) and only in math mode
(`94f57c6`), so text mode still falls back to system fonts.

MathML keeps U+27C2 for semantics while HTML renders U+22A5 — covered by
`test/mathml-spec.ts:270` and `:281`.

**What I could not verify:** I wasn't able to run the jest suite locally — `pnpm install` failed
on my machine (corepack/Node 20 `ERR_VM_DYNAMIC_IMPORT_CALLBACK_MISSING`) and left `node_modules`
empty, so I'm not going to claim a green run I didn't see. The table above is read directly from
the source on this branch, not inferred from a test. Please do re-check the deploy preview against
your demo link — if `Ð Þ þ` still render wrong there, then something outside this map is
responsible and I'd want to know.

test/unicode-spec.ts
```diff
@@ -181,3 +182,22 @@ describe("unicodeScripts", () => {
         }
     });
 });
+
+describe("glyph substitution", () => {
+    it("renders U+27C2 PERPENDICULAR with the U+22A5 glyph the fonts carry", () => {
+        expect(getFontGlyphCharacter("\u27C2", "Main-Regular")).toBe("\u22A5");
+    });
+
+    it("does not substitute characters that only borrow metrics", () => {
```
grigoriy-reshetniak: This test case feels redundant. It reads as:
> Completely separate code path is not used by `getFontGlyphCharacter`.

And this is only natural, because now `getFontGlyphCharacter` doesn't have an intersect with `extraCharacterMap`.

test/unicode-spec.ts
```diff
@@ -181,3 +182,22 @@ describe("unicodeScripts", () => {
         }
     });
 });
+
+describe("glyph substitution", () => {
+    it("renders U+27C2 PERPENDICULAR with the U+22A5 glyph the fonts carry", () => {
```
grigoriy-reshetniak: This test case seems to be too close to the new cases in `matml-spec.ts`.

test/mathml-spec.ts
```diff
@@ -266,4 +266,26 @@ describe("A MathML builder", function() {
         expect(getMathML("\\copyright\\neq\\notin\u2258\\KaTeX"))
             .toMatchSnapshot();
     });
+
+    it("should map \\perp to U+27C2 and keep \\bot as U+22A5", () => {
```
grigoriy-reshetniak: This case is great, but I think it could be improved by focusing on character substitution in general, e.g. that \perp maps to U+27C2 and \bot stays as U+22A5, rather than implying this single mapping is the purpose of the test.

test/mathml-spec.ts
```diff
@@ -266,4 +266,26 @@ describe("A MathML builder", function() {
...
+        const bot = getMathML("x \\bot y");
+        expect(perp).toContain("\u27C2");
+        expect(perp).not.toContain("\u22A5");
+        expect(bot).toContain("\u22A5");
+        expect(bot).not.toContain("\u27C2");
+        expect`x \perp y`.toBuild();
+        expect`x \bot y`.toBuild();
+    });
+
+    it("should render \\perp in HTML using the \\bot font glyph", () => {
```
grigoriy-reshetniak: This test case doesn't belong in `mathml-spec.ts` and comment about "substitution in general" applies as well.

src/fontMetrics.ts
```diff
@@ -191,6 +191,38 @@ export function setFontMetrics(
     metricMap[fontName] = metrics;
 }
 
+/**
+ * Characters that KaTeX fonts do not carry, but which are typographically
+ * identical to a character the fonts do carry, so the substitute may be
+ * rendered in place of the original.
+ *
+ * This is deliberately separate from `extraCharacterMap`. That map exists to
```
grigoriy-reshetniak: I believe comment above `extraCharacterMap` about `glyphSubstitutionMap` is misleading and could be omitted.

src/fontMetrics.ts
```diff
@@ -191,6 +191,38 @@ export function setFontMetrics(
...
+ * the visible text, so only genuine glyph equivalences belong here.
+ */
+const glyphSubstitutionMap: Record<string, string> = {
+    // U+27C2 PERPENDICULAR is the same glyph as U+22A5 UP TACK, which the
+    // KaTeX fonts do carry.
+    '\u27C2': '\u22A5',
+};
+
+/**
+ * Returns the KaTeX-font glyph to render for `character`, substituting an
```
grigoriy-reshetniak: Comments on L195, L212 and L221 are describing same thing in different ways. Please combine them into single one.

src/buildCommon.ts
```diff
@@ -45,6 +45,17 @@ const lookupSymbol = function(
         }
     }
 
+    // In math mode, characters absent from this KaTeX font may have a known
+    // substitute glyph; render the substitute so HTML output is not blank.
+    // In text mode, leave the character as-is so the browser can fall back to
+    // system fonts (e.g. Cyrillic, CJK) instead of showing Latin lookalikes.
+    if (mode === "math") {
+        const glyph = getFontGlyphCharacter(value, fontName);
```
grigoriy-reshetniak, line 53: This approach is working, but there is an alternative to add data about this character to KaTeX font. I'm not sure if this should be done instead. @edemaine could you please advise?

tzh476:
Thanks for the detailed pass, @grigoriy-reshetniak — this made the tests better. Pushed `512ce0c0`,
which also merges `main` so the PR is no longer behind (that pulled in your `#4265` `U+10FFFF` fix;
I checked the PR diff still shows only my five files and doesn't revert it).

Taking your points in order.

**Moved the HTML test out of `mathml-spec.ts` (your `:281`)** — you're right, it asserts HTML output
and doesn't belong there. It now lives in `unicode-spec.ts` next to the substitution tests.

**Generalized the MathML test (your `:270`)** — done, and it reads as the general property now:
`\perp` → U+27C2, `\bot` → U+22A5, `\top` → U+22A4, each absent from the others' output. One
correction to my own first draft of this: I initially justified including `\top` as "pinning that the
substitution table isn't over-broad". That's wrong, and I checked rather than assumed — I made the
table genuinely over-broad (added `'⊤':'⊥'`, `'Ш':'W'`, `'Å':'A'`) and this MathML test still
**passed**. It can't observe `glyphSubstitutionMap`, because MathML never calls
`getFontGlyphCharacter`. The `\top` row does guard something real, just narrower: `\top`'s own
`symbols.ts` entry.

**Dropped the redundant positive case in `unicode-spec.ts` (your `:187`)** — agreed. The moved HTML
test covers that direction end-to-end through `lookupSymbol`, which is the only call site.

**I'd like to keep the negative test at `:191`**, and your premise for questioning it is correct —
`getFontGlyphCharacter` no longer intersects `extraCharacterMap` (it reads only
`glyphSubstitutionMap`; `extraCharacterMap` is now consulted solely by `getCharacterMetrics`). But
that makes it independent, not redundant. Three arms:

| arm | result |
|---|---|
| with the fix | 64/64 pass |
| revert `\perp` to U+22A5 | **2 fail** (both spec files) |
| leave the codepoint, add a lookalike to `glyphSubstitutionMap` | **1 fail — only this test** |

That third arm is the whole argument: an over-broad substitution table is invisible to every MathML
and HTML assertion here, because MathML doesn't call the function and the rendering tests only
exercise `\perp`. This is the only thing in the repo pinning that table's membership policy. Happy to
be overruled, but I didn't want to delete it on the strength of "looks redundant" without measuring.

**On the `extraCharacterMap` comment (your `:199`)** — I think there's a location mix-up: the comment
in question sits *above* `glyphSubstitutionMap` and refers *to* `extraCharacterMap`, not the reverse.
The comment above `extraCharacterMap` is the pre-existing `TODO(kevinb)` block, which this PR doesn't
touch. And the distinction it draws is load-bearing: if I fold the two maps together and let
`getFontGlyphCharacter` consult `extraCharacterMap`, then `Ð` renders as **D** and `Þ`/`þ` render as
**o** in math mode. So I'd rather keep it — though I'm glad to reword it if it's the phrasing rather
than the substance that reads oddly.

A correction I owe you here too: while checking this I found Cyrillic `Ш` is *not* actually at risk,
because it always reaches `lookupSymbol` with `mode === "text"` — even bare, and even under
`\mathrm`/`\mathbf`/`\mathit`/`\mathsf`/`\mathtt`/`\operatorname`. Only `Ð`/`Þ`/`þ`/`ð` reach the
math-mode branch. The comment's example is therefore illustrative rather than a live hazard, and I've
left the eth/thorn cases as the real ones.

**Combining the three comments (your `:212`)** — I haven't done this one yet; I want to make sure the
merged version keeps every distinct fact rather than reading more smoothly while losing one. Will
follow up.

**On the font route (your `:53`, for @edemaine)** — deferring to you both, but some specifics that
might make the decision cheaper, since the two halves differ a lot in cost:

- **The metrics half is cheap.** `src/fontMetricsData.js` is generated and marked DO NOT MODIFY, but
  for this character it comes from the TeX TFMs, not the TTF cmap — U+22A5 isn't in
  `extract_ttfs.py`'s `metrics_to_extract` at all. The entry producing `8869` is
  `src/metrics/mapping.pl:182`, `0x3F => 0x22A5, # \bot`. Adding `0x3F => 0x27C2` beside it and
  re-running `buildMetrics.sh` emits a `10178` entry from `cmsy10` char 63 — I ran the patched
  `mapping.pl` to confirm, and there's no duplicate-mapping error because the `die` only fires on a
  duplicate *target*. Needs TeX + kpathsea, but no Docker and no font rebuild.
- **The glyph half is the expensive one.** `src/fonts/makeFF:187` has the same `0x3F => 0x22A5`, and
  changing it means `buildFonts.sh` — Docker + texlive + fontforge per `dockers/fonts/Dockerfile` —
  and rewriting the 60 binaries under `fonts/`. There is precedent for giving one TeX char a second
  Unicode codepoint (`makeFF` already does it for `0x17`), so it's not unprecedented, just costly.

If you'd prefer the font route I'm happy to redo it that way; I went with the substitution map because
it needs no regenerated binaries.

Verification for this push: `tsc --noEmit` clean, `pnpm test:lint` exit 0 (eslint + stylelint), full
suite **1308 passed / 8 suites / 125 snapshots**, run against a tree extracted from the commit itself.

tzh476:
Followed up on the one I deferred — combining the three comments (`75a76a1d`). You were right that
they said the same thing three ways; I wanted to merge them without quietly dropping a fact, so here
is what each contributed and where it went:

| fact | was in | now |
|---|---|---|
| only same-glyph equivalences may be substituted | block above the map | kept, first sentence |
| why that differs from `extraCharacterMap` (metrics vs glyphs) | block above the map | kept, second paragraph |
| U+27C2 is the same glyph as U+22A5 | inline in the map | kept, trimmed |
| the return contract incl. the unchanged cases | doc above the function | kept, tightened |
| "only substitute when the font lacks the original" | inline before the metric check | **dropped** — the function's doc comment now says it and the code reads the same way |

I also swapped the abstract warning for the concrete consequence: instead of saying substituting
lookalikes "would change the visible text", it now names what actually happens — `Ð` renders as `D`
and `Þ` as `o`. That is the measurement from the earlier arm, so the comment states a fact I checked
rather than a caution I asserted.

`tsc --noEmit` clean, eslint clean, full suite **1308 passed / 8 suites**.

That is all seven points. The two I pushed back on are the negative substitution test and the
`extraCharacterMap` comment, both with the measurement inline above — happy to change either if you
read the evidence differently.

tzh476:
@grigoriy-reshetniak thanks — I went to delete this test, then measured it first, and the measurement
says keep it. Showing the numbers rather than just disagreeing.

**Your reading of the code is exactly right:** `getFontGlyphCharacter` returns early for anything not in
`glyphSubstitutionMap`, and no Cyrillic character is in that map, so on today's `main` the assertion
can only pass. A test that cannot fail is worth nothing, which is why the objection is a fair one.

**But it is not dead — it fails under the one mutation it exists to catch.** The lookalike bug is
reintroduced by a single `??`, i.e. by anyone later deciding the two maps should share a lookup:

```diff
-    const substitute = glyphSubstitutionMap[character[0]];
+    const substitute = glyphSubstitutionMap[character[0]] ??
+        extraCharacterMap[character[0]];
```

Same test file, same command, only that hunk applied:

|  | `renders \perp …` | `does not substitute … borrow metrics` |
|---|---|---|
| current branch | ✓ | ✓ |
| with the `??` mutation | **✓** | **✕** |

So the `\perp` test — the one you note is close to the MathML cases — passes straight through the
regression, and **this test is the only thing in the suite that catches it.** That asymmetry is the
argument for keeping it: it guards the invariant that separating the two maps was meant to establish,
not the current behaviour of `getFontGlyphCharacter`.

I have left it in place and not pushed a change for this thread. If you would still rather it go, say so
and I will drop it — it is your call and I will not re-argue it; I just did not want to remove a live
control on the strength of it *looking* redundant.

For completeness, the other five inline comments are addressed in `512ce0c0` and `75a76a1d` (the HTML
test moved out of `mathml-spec.ts`, the MathML case generalized to `\perp` vs `\bot` rather than the one
mapping, the misleading `extraCharacterMap` comment dropped, and L195/L212/L221 combined into a single
comment). GitHub marks those threads outdated rather than resolved, so they may still look open to you.

Full verification on `75a76a1d`, Node 22 + `corepack pnpm@11.4.0`:

```
jest    8 suites passed, 1308 tests passed, 125 snapshots passed
tsc     --noEmit, exit 0
eslint  ., exit 0, 0 problems
```

On the font question at `buildCommon.ts:53` — that one is genuinely for @edemaine and I have no
independent view worth adding. If adding U+27C2 to the fonts is preferred, the substitution map and its
two tests come out and I am happy to redo it that way.

tzh476:
@grigoriy-reshetniak — correcting a statement I made to you, because it was wrong.

In my summary comment I wrote that the changes included *"the misleading `extraCharacterMap` comment
**dropped**"*. **It was not dropped.** It is still there at `src/fontMetrics.ts:196-200` on head
`75a76a1d`, reworded rather than removed — and 41 minutes earlier in the same thread I had told you *"I'd
rather keep it"* and listed it as one of "the two I pushed back on". So my own summary contradicted my
own position, in the direction that made me look more compliant than I was.

To be unambiguous about where I actually stand on your seven points:

| your point | what I did |
|---|---|
| `:281` HTML test in the wrong file | **moved** to `unicode-spec.ts` |
| `:270` generalize the MathML test | **done** (`\perp`/`\bot`/`\top`) |
| `:187` redundant positive case | **deleted** |
| `:212` three comments say the same thing | **combined into one** |
| `:53` font-data alternative | **deferred to @edemaine**, with the cost of each half |
| **`:199` misleading `extraCharacterMap` comment** | **kept, reworded** — I pushed back |
| **`:191` negative test looks redundant** | **kept** — I pushed back, with the three-arm measurement |

Two kept, five actioned. If you would still rather the `:199` comment go, say so and I will remove it —
it is your review and I am not going to re-argue a point I have already made once.

I also corrected the PR description, which had drifted: it claimed 1307 tests (it is **1308**), still
listed the HTML assertion under `mathml-spec.ts` after `512ce0c0` moved it, and still advertised the
direct `getFontGlyphCharacter` assertion that the same commit deleted. The first thing you read should
describe the branch as it is, not as it was three commits ago.

grigoriy-reshetniak:
@tzh476 I appreciate your willingness to contribute to KaTeX, but it feels like your comments have been influenced _too much_ by an LLM. There are too many of them, and they’re too careless about tagging people. Please take a moment to review and trim them before posting, it would make the discussion more respectful of everyone’s time.

And there’s not really anything that can be done until [this comment](https://github.com/KaTeX/KaTeX/pull/4267#discussion_r3894401651) is answered.

tzh476:
Found two untested conditionals in my own diff and pinned them — `5627e90d`.

Both are the same class of argument I used to *keep* the negative test you questioned, so it would have
been inconsistent of me not to apply it to my own gaps. Neither guard was pinned by anything: mutating
each left the **entire suite green at 1308 tests**.

```
buildCommon.ts:52   if (mode === "math")  ->  if (mode === "math" || mode === "text")   1308 passed
fontMetrics.ts:221  return metricMap[font]?.[ch] ? character : substitute;
                    ->  return substitute;                                             1308 passed
```

The first is the design decision I argued for at length in this thread — that text mode must **not**
substitute, so a character the KaTeX fonts lack is left for the browser to resolve from a system font
rather than being replaced by a Latin lookalike. **Defending a decision in prose while nothing enforces
it in code is exactly the gap you were probing.**

The second is what makes the substitution conditional: a font that *does* carry U+27C2 should render
U+27C2, not the U+22A5 stand-in.

One test each, and both verified to be real controls by re-running the same mutations:

| | before | after |
|---|---|---|
| text-mode substitution mutant | survives (1308 passed) | **1 failed** |
| unconditional-substitute mutant | survives (1308 passed) | **1 failed** |

The second test uses the public `setFontMetrics` to register a font that carries U+27C2 — that is the
case which distinguishes "substitute when missing" from "substitute always", and it is the one that
kills the mutant.

**A side effect worth mentioning**, since it says something about the earlier test: removing a local
`require("../katex")` that shadowed the module import exposed a real type error. That local was typed
`any`, which hid that `renderToString`'s second argument is required — `tsc` flagged it the moment the
shadow was gone. So the HTML test had been passing through an untyped path.

Full suite **1310 passed / 8 suites / 125 snapshots**, `tsc --noEmit` 0, `eslint` 0.

tzh476:
@grigoriy-reshetniak — I went back to your `:191` "feels redundant" point with a measurement rather than another argument, because your premise turned out to be exactly right and I want to be precise about where I think the conclusion diverges.

**Your premise is correct.** `getFontGlyphCharacter` has no intersection with `extraCharacterMap` any more, and I verified that rather than asserting it: `glyphSubstitutionMap` contains exactly one entry (`⟂ → ⊥`), so all eight characters in that test hit the `substitute === undefined` early return. The assertion passes **by construction** on today's code. As a statement about *current* behaviour it proves nothing, which is what you said.

**Where I think it still earns its place** — I extracted the real function and both maps verbatim into a standalone harness (npm is broken on this machine, so I could not run vitest; the logic is pure and dependency-free, so this is the same code, not a paraphrase) and mutated it into the *specific* mistake your 2026-08-24 review was about — a future contributor treating the metrics map as a rendering fallback:

```diff
-const substitute = glyphSubstitutionMap[character[0]];
+const substitute = glyphSubstitutionMap[character[0]] ?? extraCharacterMap[character[0]];
```

Result:

```
Ш -> W    Д -> A    Ж -> K    Г -> F
Å -> A    Ð -> D    Þ -> o    å -> a
assertions failing under the mutant: 8/8
```

And critically: **`\perp` still renders correctly under that mutant** (`⟂ → ⊥`), so every `\perp`-based test in `mathml-spec.ts` and `unicode-spec.ts` stays green. The only thing that catches it is this test.

So it is not testing "a separate code path is not used". It is pinning the invariant that made your original `CHANGES_REQUESTED` valid — *metrics fallback must never become rendering fallback* — and it is the sole guard on it. A test that only passes by construction today is still the thing that fails the day the construction changes.

That said: **you have now raised this twice, and it is your review.** If you would still rather it go, say so plainly and I will delete it in the next push without further argument — I would rather have your approval on a slightly thinner test suite than keep relitigating one case. I have put the mutation result in the test's comment either way, so if it survives, the next reader sees why it exists instead of reaching your same conclusion.

The `:199` comment is the other one I kept, and my position there is weaker — it is a judgement call about prose, not an invariant. Say the word and both go.

tzh476:
@grigoriy-reshetniak — my correction yesterday was itself wrong, and it pointed you at the wrong place.
Sorry for the noise; here is the verified state so the thread is accurate.

I wrote that the misleading `extraCharacterMap` comment *"is still there at `src/fontMetrics.ts:196-200`."*
**That line range is not the comment you objected to.** It is the docblock on my *new*
`glyphSubstitutionMap`, which mentions `extraCharacterMap` in order to explain why the two maps are
separate.

The comment above `extraCharacterMap` itself — the one your note was about — is **untouched upstream
text**:

```
// These are very rough approximations.  We default to Times New Roman which
// should have Latin-1 and Cyrillic characters, but may not depending on the
// operating system.  ...
// TODO(kevinb) allow union of multiple glyph metrics for better accuracy.
const extraCharacterMap: Record<string, string> = {
```

No mention of `glyphSubstitutionMap`, because this PR never edited that block. Confirmed from the PR's
own file list rather than my memory: `src/fontMetrics.ts` is **+30/-0** — purely additive, so nothing
that was already there could have been reworded.

**So your item is satisfied, and was already satisfied when you raised it.** If what you would prefer is
that my new docblock *stop referring to* `extraCharacterMap` at all, say so and I will trim it to just
the invariant — I kept the reference only because the reason the maps are separate is the thing a future
reader is most likely to get wrong, and it is exactly the distinction you drew.

Remaining on this PR from my side: nothing. The one open question is the font alternative you raised for
@edemaine — whether U+27C2 should be added to the KaTeX fonts instead of substituted. If that is the
preferred direction, the substitution map and its two tests come out and I am happy to redo it that way.

tzh476:
@edemaine @grigoriy-reshetniak — on the open question of whether to add U+27C2 to the KaTeX fonts
instead of substituting: here is the font-side data, so the choice can be made on facts rather than my
preference. I have no stake in which way it goes.

**1. What the shipped fonts actually contain.** Parsed the `cmap` of the TTFs in `fonts/` directly:

| codepoint | `KaTeX_Main-Regular` | `KaTeX_AMS-Regular` |
|---|---|---|
| U+22A5 UP TACK | **present** | absent |
| U+22A4 DOWN TACK | **present** | absent |
| U+27C2 PERPENDICULAR | **absent** | absent |

(`Main-Regular` exposes 283 codepoints, `AMS-Regular` 257.)

**2. Why it is absent, and what adding it would mean.** The fonts are generated from the TeX sources by
`src/fonts/makeFF`, which maps TeX font slots to Unicode. The relevant lines:

```perl
0x3E => 0x22A4,         # \top
0x3F => 0x22A5,         # \bot
```

**There is no separate `\perp` glyph in the source font.** Slot `0x3F` is `\bot`, and TeX draws `\perp`
from that same slot — the two commands have always shared one glyph, which is precisely why U+22A5 was
serving both. So "add U+27C2 to the font" is not exposing an existing glyph under a second codepoint by
adding a mapping row; there is no distinct outline to point at. It would mean either:

- adding U+27C2 as an **alternate Unicode** (`altuni`) for the same glyph as U+22A5 — the honest
  encoding of "same glyph, two codepoints", and it makes the substitution map unnecessary; or
- drawing a genuinely distinct perpendicular glyph, which is a design decision about whether KaTeX
  wants `\perp` and `\bot` to *look* different. TeX's answer has always been no.

**3. What that implies for this PR.** The first option is strictly better than my substitution map if you
are willing to regenerate and ship the fonts, because it moves the fact "U+27C2 and U+22A5 are the same
glyph" from TypeScript into the font, where it belongs. The cost is a font rebuild through
`src/fonts/Makefile` (mftrace/fontforge toolchain) and new binary assets in the repo.

My substitution map is the no-font-rebuild version of the same statement. It is the cheaper change, not
the more correct one.

**If you prefer the font route, say so and I will close this PR** — the `symbols.ts` one-liner
(`⊥` → `⟂`) is the only part that survives, and it becomes correct on its own once the font
carries U+27C2. I would rather that than leave a workaround in `buildCommon.ts` for a problem the font
can express directly.

PR #4276, opened 2026-09-01
feat(array): add the {numcases} environment

### LINKED ISSUE #3962: Add support for {numcases}
{cases} and {dcases} are supported in katex. If can we also support for {numcases}? Sometimes I hope each formula has a number tag.

AtsushSaito:

A multi-case equation is written with `{cases}` (amsmath) or `{dcases}` / `{rcases}` (mathtools). All of them set the whole thing as one unnumbered object, so the individual cases cannot be referred to later — there is no way to say "by (2), the derivative is −1" about the second case, as you would about a line of an `{align}`.

`cases.sty` solves this with `{numcases}`, which differs from `{cases}` in exactly two ways: it takes the left-hand side as a mandatory argument, and it numbers every row. KaTeX has no equivalent — `\begin{numcases}` fails with `No such environment: numcases`. This is #3962.

```tex
\begin{numcases}{\lvert x\rvert=}
  x,  & \text{for $x \geq 0$} \\
  -x, & \text{for $x < 0$} \nonumber
\end{numcases}
```

{numcases} rendered by pdflatex with cases.sty above, and by KaTeX with this patch below

- **one equation number per row**, so prose can refer to a single case
- `\nonumber` / `\notag` suppress the number on one row; `\tag{…}` replaces it
- the counter is **shared with `{align}` and `{gather}`**, so numbers keep running across environments
- display-only, like the other numbered environments

Here `{numcases}` follows an `{align}` and picks up the counter at (3), and the brace grows with the number of rows:

{numcases} numbered (3) directly after an {align} numbered (1) and (2), and a four-row {numcases} numbered (4) to (7)

### Why the html builder hands over one node instead of three

The environment has three visible parts plus a tag, so the natural thing to return is four siblings:

```js
return makeFragment([lhs, brace, alignment, tagCol]);   // does not work
```

That renders without the numbers. `buildHTML` lifts a tag out of `.katex-base` only when the root expression has exactly two elements:

```js
// src/buildHTML.ts
if (expression.length === 2 && expression[1].hasClass("katex-tag")) {
    eqnNum = expression.pop();
}
```

With four, the tag stays inside the base — and the stylesheet places it relative to that base rather than the line, so it lands in the middle of the formula instead of at the right margin (measured in Chrome, 600px container: `572…600` as a direct child of `.katex-html`, `299…354` inside a base).

So the three parts are collapsed into one node, purely to make the count two:

```js
return makeFragment([makeSpan(["mord"], [lhs, brace, alignment]), tagCol]);
```

Wrapping the alignment in a `leftright`, the way `{cases}` does, is not an option either: the tag would end up inside the resulting `minner` and never reach the root at all.

That condition looks incidental rather than intended — it went in with `{gather}` (#2183) to match the only producer that existed at the time, and the stylesheet itself only requires the tag to be a *direct child* of `.katex-html`, not a second sibling. I've raised that separately as #4281, since it turns on a behaviour question of its own.

If #4281 is taken, this environment no longer needs the wrapper and the builder simplifies to:

```diff
-    const bodyNode = prefix.length > 0
-        ? makeSpan(["mord"], [...prefix, tableBody], options)
-        : tableBody;
-
     if (tagSpans.length === 0) {
-        return makeSpan(["mord"], [bodyNode], options);
+        return makeSpan(["mord"], [...prefix, tableBody], options);
     } else {
         ...
-        return makeFragment([bodyNode, tagCol]);
+        return makeFragment([...prefix, tableBody, tagCol]);
     }
```

Nothing else here changes, so the two are independent — this PR does not depend on that one.

The upside of the current shape is that **`buildHTML.ts` and the stylesheet are untouched** — the change is one optional field on `ArrayNode` plus a branch in each of the two array builders, reusing the existing `{cases}` column spec and the existing `autoTag` machinery.

### Notes

`\begin{numcases}` is **not** added to auto-render's default delimiters, so it needs `$$…$$` around it like `{cases}` does. That list is introduced as *"Render AMS environments even if outside `$$…$$` delimiters"* and every entry in it is an AMS environment, while `{numcases}` is from `cases.sty` — adding it is a question about that list rather than about this environment. Happy to open a follow-up if you want it there.

`test/screenshotter/test.tex` now loads `cases.sty`, which as a side effect widens amsmath's `{cases}` by `\hskip 2\p@` (29.74635pt → 31.74635pt measured), so the LaTeX reference for the existing `Cases` entry shifts slightly by its presence.

Fixes #3962


AtsushSaito:
@grigoriy-reshetniak  @edemaine 
This is ready for review whenever you have time.
One things in the description are worth a look in particular: why the html builder hands over one node rather than three.

PR #4277, opened 2026-09-01
fix(array): preserve tags on empty final rows

### LINKED ISSUE #3898: tag in empty `{align}` row not showing
MWE:

```latex
\begin{align}
\nonumber
a\\
\tag{1}
\end{align}
```

LaTeX output:

screenshot_2023-11-05-111814

KaTeX output:

screenshot_2023-11-05-111841

lprnmns:

An explicit `\tag` on an empty final row of an `align` environment is recorded, but the row is then removed as a trailing empty row. This leaves the row and tag arrays out of alignment, so the visible tag column is empty.

Final empty rows are retained when they carry a manual tag. Automatic empty trailing rows and rows disabled with `\nonumber` continue to be removed.

The regression test fails on the previous implementation because the visible tag span is absent and passes with the fix.

Validation:

- ESLint
- Stylelint
- TypeScript (`tsc --noEmit`)
- Jest: 8 suites, 1306 tests, 125 snapshots
- Full production build
- `git diff --check`

Fixes #3898

### DIFF
--- a/src/environments/array.ts
+++ b/src/environments/array.ts
@@ -210,12 +210,13 @@ function parseArray(
             endRow();
             // Arrays terminate newlines with `\crcr` which consumes a `\cr` if
             // the last line is empty.  However, AMS environments keep the
-            // empty row if it's the only one.
+            // empty row if it's the only one or has a manual tag.
             // NOTE: Currently, `cell` is the last item added into `row`.
             if (row.length === 1 && cell.type === "styling" &&
                 cell.body.length === 1 && cell.body[0].type === "ordgroup" &&
                 cell.body[0].body.length === 0 &&
-                (body.length > 1 || !emptySingleRow)) {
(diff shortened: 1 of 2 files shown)


grigoriy-reshetniak (approved):

grigoriy-reshetniak:
Thank you, looks good!

edemaine (approved):
Looks good! I confirmed against pdfLaTeX with `amsmath`: preserving an empty final row carrying `\tag` or `\tag*` matches LaTeX, including empty tags and tags combined with `\nonumber`. The `Array.isArray` check looks appropriate.

One other discrepancy I noticed: LaTeX distinguishes `align` from `aligned`. A bare trailing `\\` creates another row in `align`, including automatic numbering, but is discarded in `aligned`. KaTeX currently discards it in both. Something we might want to fix someday... though its kind of annoying behavior.

PR #4283, opened 2026-09-06
fix(array): preserve trailing rows in align

lprnmns:

A bare `\\` at the end of an automatically numbered `align` environment is parsed as an empty final row, but the array parser removes it before the HTML and MathML builders run. As a result, KaTeX omits the automatically numbered row that LaTeX preserves. The corresponding `aligned` behavior is intentionally unchanged.

Automatically numbered AMS environments retain an empty final row when it was explicitly ended with a bare `\\`. Non-numbered, nested, manually suppressed, and starred environments keep their existing trailing-row cleanup behavior.

**Tests**

- `pnpm exec jest test/katex-spec.ts -t "should keep a bare final row in align but not aligned" --runInBand` — passed
- `pnpm test` — passed (8 suites, 1,307 tests, 125 snapshots)
- `pnpm build` — passed
- `pnpm test:build` — passed
- `pnpm test:screenshots` — passed for the repository's local Firefox and Chromium screenshotter
- `git diff --check` — passed

Safari screenshots are generated by the repository's macOS CI job; they were not runnable on this Linux host.


grigoriy-reshetniak:
Thank you! 

I've run `pdflatex` on [align test suite](https://github.com/KaTeX/KaTeX/blob/main/test/screenshotter/ss_data.yaml#L25) and it's producing the following:
image
keeping the trailing row.

LGTM!

edemaine (approved):

grigoriy-reshetniak (approved):

Issue #1107, opened 2018-01-28
Add support for international monetary indicators (Euro, Rupee, Ruble, etc.)


kevinbarabash:
- [ ] Euro €
- [ ] Rupee ₹
- [ ] Ruble ₽

These don't exist in any of our fonts currently so we'll want to add them to the fonts first.  I've opened https://github.com/KaTeX/katex-fonts/issues/18 to track that part of the work.

ylemkimon:
- [ ] Won ₩
- [ ] Yuan 元
- [ ] Lira ₺
- [ ] Baht ฿

edemaine:
@ribrdb I would create a new issue, as that sounds much easier to fix. Please let us know what symbols specifically. Thanks!

kevinbarabash:
They feel roughly the same cost.  If we don't add real glyphs we still need to add fake metrics similar to what we do for Cyrillic.

edemaine:
Sorry, I meant: if there's an *existing* currency symbol such as `\pounds` that isn't working in text mode but isn't, then let us know in a separate issue. But yes, for these *new* symbols, adding to math and text mode are equally difficult and this is the issue for both.

PR #4288, opened 2026-09-15
feat: add \euro and \text{\texteuro} support

Yashuchirag:
## Summary
- Adds `\euro` (math and text mode) and `\text{\texteuro}` (text mode alias), closing the Euro-symbol part of #1107.
- The glyph is sourced from the `eurosym` CTAN package's `feymr10` font (the official Euro glyph, character `e`), which ships as a ready-made Type 1 font alongside its METAFONT source — no new font-pipeline tooling was needed, just wiring it in the same way `eufm10`/`eufb10` (Euler) are already imported as pre-built `.pfb` glyphs.
- `src/fonts/Makefile`, `src/fonts/makeFF`: import `feymr10.pfb`.
- `src/metrics/mapping.pl`, `src/metrics/extract_tfms.py`: mirror that mapping for metrics extraction.
- `src/symbols.ts`: define the symbols, following the existing `\pounds`/`\textsterling` pattern.
- `fonts/KaTeX_Main-Regular.*`, `src/fontMetricsData.js`: regenerated via `dockers/fonts`; verified byte-identical to the current files aside from the new glyph/metric.
- `docs/support_table.md`, `docs/supported.md`: updated per CONTRIBUTING.md.

## Known limitation
The imported glyph is visibly heavier than the rest of `Main-Regular`'s stroke weight — `eurosym`'s "medium" weight doesn't match Computer Modern's actual weight at this size. FontForge's automated `changeWeight()` couldn't cleanly thin it without introducing outline artifacts where the crossbars meet the curve. Matching it properly would need manual node-level editing by someone with type-design/FontForge experience — flagging for review rather than blocking on it.

## Scope note
Rupee (₹) and Ruble (₽), also requested in #1107, are not included here. Rupee has a working font source (`tfrupee`) but it's GPLv3-licensed with no font exception, which conflicts with KaTeX's MIT license — it needs a differently-licensed glyph source. Ruble has no usable typographic source in TeX Live at all (only an icon-font glyph, a poor stylistic fit). Leaving both for separate follow-up.

## Test plan
- [x] `pnpm test:jest` — all 1328 tests pass
- [x] `pnpm test:lint` — clean
- [x] `pnpm test:ts` — clean
- [x] Verified `\euro` and `\text{\texteuro}` render correctly through KaTeX's actual parser/renderer (checked HTML/MathML output)
- [x] Regenerated fonts/metrics diffed byte-for-byte against current files to confirm no unintended drift

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01D5fFGNNETk7gLdmwAFLCh3


Yashuchirag:
Thanks for the review, @greptile-apps — both fixed in the latest commit:

**P1 (styled Euro loses metrics):** confirmed this was real — `\mathbf{\euro}`/`\textbf{\texteuro}` fell back to a zero-sized box since Main-Bold had no glyph. Added the bold-weight glyph (`feybr10`, the eurosym package's bold companion to `feymr10`) to Main-Bold, mirroring how `\pounds` already has dedicated Main-Bold metrics. Regenerated fonts/metrics and diffed against `main` to confirm no unrelated drift.

**P2 (no regression coverage):** added focused tests in `test/katex-spec.ts` (math/text mode builds, correct character, and a test specifically for the bold-metrics fix using `.not.toWarn()`) and `test/mathml-spec.ts` (MathML output), following the pattern from #4262.

Note: Main-Italic/Main-BoldItalic still lack the glyph, same pre-existing gap as `\pounds` has today (neither has Italic coverage either). Happy to extend further if that's wanted, but wanted to flag it's consistent with the existing convention rather than a regression.

